PR #1233, opened 2022-03-18
[REF] popovers : improve popover component

hokolomopo:
The popover component have currently some problems :

- It's currently not possible to define the height of a popover.
The "childHeight" props is only used for positioning.
- flipHorizontalOffset/flipVerticalOffset props are confusing to use.
They should sometime be positive, sometime negatives, and are computed
by hand to use popovers. Ideally we'd want the logic of it to be
hidden in the component.
- The popovers are linked with the grid (with scrollbar/topbar size
used in computations), but we use them at other places

Refactored the popover component. Now :

- The logic of the popover being displayed in the left/right, depending
on if whether there is space or not, is hidden in the component
- The popovers are now independent of the grid, they can be used anywhere
- Now the popover is positioned based on the actual size of its child,
not around arbitrary childHeight/Width props (that were dropped).

The main props are now :
- positioning: position of the popover (TopLeft/BottomRight)
- anchorRect : rectangle around which the popover is positioned
(the cell rectangle in the grid, a single point for menus)
- containerRec : rectangle inside which the popover is displayed, and
shouldn't overflow out of (ie. the viewport in the grid)
- Dropped props marginTop, it is now implied in containerRec
- added props max height/width

Odoo task ID : [2796248](https://www.odoo.com/web#id=2796248&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/figures/figure_image/figure_image.ts
+++ b/src/components/figures/figure_image/figure_image.ts
@@ -9,6 +9,7 @@ import { Menu, MenuState } from "../../menu/menu";
 
 interface Props {
   figure: Figure;
+  onFigureDeleted: () => void;
 }
 
 export class ImageFigure extends Component<Props, SpreadsheetChildEnv> {
@@ -106,3 +107,8 @@ export class ImageFigure extends Component<Props, SpreadsheetChildEnv> {
     return this.env.model.getters.getImagePath(this.figureId);
   }
 }
+
+ImageFigure.props = {
+  figure: Object,
+  onFigureDeleted: Function,
+};
--- a/src/components/grid/grid.ts
+++ b/src/components/grid/grid.ts
@@ -1,4 +1,12 @@
-import { Component, onMounted, useEffect, useExternalListener, useRef, useState } from "@odoo/owl";
+import {
+  Component,
+  onMounted,
+  useChildSubEnv,
+  useEffect,
+  useExternalListener,
+  useRef,
+  useState,
+} from "@odoo/owl";
 import {
   AUTOFILL_EDGE_LENGTH,
   HEADER_HEIGHT,
@@ -21,6 +29,7 @@ import {
   HeaderIndex,
   Pixel,
   Position,
+  Rect,
   Ref,
   SpreadsheetChildEnv,
 } from "../../types/index";
@@ -109,6 +118,7 @@ export class Grid extends Component<Props, SpreadsheetChildEnv> {
     this.canvasPosition = useAbsolutePosition(this.gridRef);
     this.hoveredCell = useState({ col: undefined, row: undefined });
 
+    useChildSubEnv({ getPopoverContainerRect: () => this.getGridRect() });
     useExternalListener(document.body, "cut", this.copy.bind(this, true));
     useExternalListener(document.body, "copy", this.copy.bind(this, false));
     useExternalListener(document.body, "paste", this.paste);
@@ -332,6 +342,10 @@ export class Grid extends Component<Props, SpreadsheetChildEnv> {
 
   isCellHovered(col: HeaderIndex, row: HeaderIndex): boolean {
     return this.hoveredCell.col === col && this.hoveredCell.row === row;
+  }
+
+  private getGridRect(): Rect {
+    return { ...this.canvasPosition, ...this.env.model.getters.getSheetViewDimensionWithHeaders() };
   }
 
   // ---------------------------------------------------------------------------
--- a/src/components/grid/grid.xml
+++ b/src/components/grid/grid.xml
@@ -51,8 +51,8 @@
       </t>
       <GridPopover
         t-if="!menuState.isOpen"
-        gridPosition="canvasPosition"
         hoveredCell="hoveredCell"
+        gridRect="getGridRect()"
         onMouseWheel.bind="onMouseWheel"
         onClosePopover.bind="onClosePopover"
       />
--- a/src/components/link/link_display/link_display.ts
+++ b/src/components/link/link_display/link_display.ts
@@ -7,7 +7,7 @@ import { CellPopoverComponent, PopoverBuilders } from "../../../types/cell_popov
 import { css } from "../../helpers/css";
 import { Menu } from "../../menu/menu";
 
-const LINK_TOOLTIP_HEIGHT = 43;
+const LINK_TOOLTIP_HEIGHT = 32;
 const LINK_TOOLTIP_WIDTH = 220;
 
 css/* scss */ `
@@ -19,6 +19,9 @@ css/* scss */ `
     border-radius: 4px;
     display: flex;
     justify-content: space-between;
+    height: ${LINK_TOOLTIP_HEIGHT}px;
+    width: ${LINK_TOOLTIP_WIDTH}px;
+    box-sizing: border-box !important;
 
     img {
       margin-right: 3px;
@@ -63,7 +66,6 @@ interface LinkDisplayProps {
 export class LinkDisplay extends Component<LinkDisplayProps, SpreadsheetChildEnv> {
   static components = { Menu };
   static template = "o-spreadsheet-LinkDisplay";
-  static size = { width: LINK_TOOLTIP_WIDTH, height: LINK_TOOLTIP_HEIGHT };
 
   get cell(): EvaluatedCell {
     const { col, row } = this.props.cellPosition;
--- a/src/components/menu/menu.ts
+++ b/src/components/menu/menu.ts
@@ -1,11 +1,10 @@
 import { Component, onWillUpdateProps, useExternalListener, useRef, useState } from "@odoo/owl";
 import {
-  HEADER_HEIGHT,
   MENU_ITEM_DISABLED_COLOR,
   MENU_ITEM_HEIGHT,
   MENU_SEPARATOR_HEIGHT,
+  MENU_VERTICAL_PADDING,
   MENU_WIDTH,
-  TOPBAR_HEIGHT,
 } from "../../constants";
 import { FullMenuItem, MenuItem } from "../../registries";
 import { getMenuChildren, getMenuDescription, getMenuName } from "../../registries/menus/helpers";
@@ -13,7 +12,7 @@ import { DOMCoordinates, Pixel, SpreadsheetChildEnv } from "../../types";
 import { css } from "../helpers/css";
 import { isChildEvent } from "../helpers/dom_helpers";
 import { useAbsolutePosition } from "../helpers/position_hook";
-import { Popover } from "../popover/popover";
+import { Popover, PopoverProps } from "../popover/popover";
 
 //------------------------------------------------------------------------------
 // Context Menu Component
@@ -22,7 +21,10 @@ import { Popover } from "../popover/popover";
 css/* scss */ `
   .o-menu {
     background-color: white;
-    padding: 5px 0px;
+    padding: ${MENU_VERTICAL_PADDING}px 0px;
+    width: ${MENU_WIDTH}px;
+    box-sizing: border-box !important;
+
     .o-menu-item {
       display: flex;
       justify-content: space-between;
@@ -85,7 +87,6 @@ export interface MenuState {
 }
 export class Menu extends Component<Props, SpreadsheetChildEnv> {
   static template = "o-spreadsheet-Menu";
-  MENU_WIDTH = MENU_WIDTH;
 
   static components = { Menu, Popover };
   static defaultProps = {
@@ -117,23 +118,29 @@ export class Menu extends Component<Props, SpreadsheetChildEnv> {
   }
 
   get menuHeight(): Pixel {
-    return this.menuComponentHeight(this.props.menuItems);
+    const menuItems = this.props.menuItems;
+
+    let menuItemsHeight = this.getMenuItemsHeight(menuItems);
+
+    // We don't display separator at the end of a menu
+    if (menuItems[menuItems.length - 1].separator) {
+      menuItemsHeight -= MENU_SEPARATOR_HEIGHT;
+    }
+
+    return 2 * MENU_VERTICAL_PADDING + menuItemsHeight;
   }
 
-  get subMenuHeight(): Pixel {
-    return this.menuComponentHeight(this.subMenu.menuItems);
-  }
-
-  get popover() {
+  get popover(): PopoverProps {
     const isRoot = this.props.depth === 1;
-    // TODO: see if we could reformulate this margin
-    const marginTop = 6 + TOPBAR_HEIGHT + HEADER_HEIGHT;
-
     return {
-      // some margin between the header and the component
-      marginTop,
-      flipHorizontalOffset: MENU_WIDTH * (this.props.depth - 1),
-      flipVerticalOffset: isRoot ? 0 : MENU_ITEM_HEIGHT,
+      anchorRect: {
+        x: this.props.position.x - MENU_WIDTH * (this.props.depth - 1),
+        y: this.props.position.y,
+        width: isRoot ? 0 : MENU_WIDTH,
+        height: isRoot ? 0 : MENU_ITEM_HEIGHT,
+      },
+      positioning: "TopRight",
+      verticalOffset: isRoot ? 0 : MENU_VERTICAL_PADDING,
     };
   }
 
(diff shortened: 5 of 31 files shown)


src/components/popover/popover.ts
```diff
@@ -1,112 +1,303 @@
...
+  containerRect: { type: Object, optional: true },
+  positioning: { type: String, optional: true },
+  maxWidth: { type: Number, optional: true },
+  maxHeight: { type: Number, optional: true },
+  verticalOffset: { type: Number, optional: true },
+  onMouseWheel: { type: Function, optional: true },
+  slots: Object,
+};
+
+abstract class PopoverHelper {
```
hokolomopo, line 121: Name kinda sucks. I'm open to any suggestion.

src/components/link/link_display/link_display.ts
```diff
@@ -7,7 +7,7 @@ import { CellPopoverComponent, PopoverBuilders } from "../../../types/cell_popov
 import { css } from "../../helpers/css";
 import { Menu } from "../../menu/menu";
 
-const LINK_TOOLTIP_HEIGHT = 43;
+const LINK_TOOLTIP_HEIGHT = 32;
```
hokolomopo, line 10: This really highlight a problem we had before : this size was just an arbitrary size that was only used to know if we display the component above or below the cell, but had nothing to do with the actual size in the DOM.

src/components/popover/popover.ts
```diff
@@ -1,112 +1,303 @@
...
+      const el = this.popoverRef.el!;
+
+      const propsMaxSize = { width: this.props.maxWidth, height: this.props.maxHeight };
+
+      const elDims = {
+        width: el.getBoundingClientRect().width,
+        height: el.getBoundingClientRect().height,
+      };
+
+      const spreadsheetRect = el.parentElement!.getBoundingClientRect();
```
hokolomopo, line 72: This should be fine since we portal the popover directly as a child of the spreadsheet.

src/helpers/rectangle.ts
```diff
@@ -0,0 +1,28 @@
...
+import { intersection } from "./zones";
+
+/**
+ * Compute the intersection of two rectangles. Returns nothing if the two rectangles don't overlap
+ */
+export function rectIntersection(rect1: Rect, rect2: Rect): Rect | undefined {
+  return zoneToRect(intersection(rectToZone(rect1), rectToZone(rect2)));
+}
+
+function rectToZone(rect: Rect): Zone {
```
hokolomopo, line 11: Not technically a zone since we still are in pixels and not in col/row coordinates. But this allow me to use all the helpers of Zone.

hokolomopo (commented):
:partying_face:

tests/components/figure.test.ts
```diff
@@ -64,8 +68,20 @@ class TextFigure extends Component<Props, SpreadsheetChildEnv> {
...
+  "o-figure-menu-item": () => ({ top: 500, left: 500 }),
+});
+
 beforeAll(() => {
   figureRegistry.add("text", { Component: TextFigure });
+  chartComponentRegistry.add("line", ChartJsComponent);
+  chartComponentRegistry.add("bar", ChartJsComponent);
+  chartComponentRegistry.add("pie", ChartJsComponent);
+  chartComponentRegistry.add("gauge", ChartJsComponent);
+  chartComponentRegistry.add("scorecard", ScorecardChart);
```
hokolomopo, line 84: This is foireux, this wasn't needed before because we imported `charts.test` where, probably by luck, the registries were populated.
And I cannot import `charts.test` anymore, sicne I need to mockGetBoundingClientRect in both files.

We should do something about that, there is also a random `chartComponentRegistry.add("bar", ChartJsComponent);` in `grid.test.ts`

src/helpers/rectangle.ts
LucasLefevre: :man_shrugging:

src/components/popover/popover.ts
```diff
@@ -1,112 +1,303 @@
...
+  ): number;
+
+  protected abstract getLeftCoordinate(
+    actualElementWidth: number,
+    shouldRenderAtRight: boolean
+  ): number;
+
+  getCSSStyle(
+    elDims: DOMDimension,
+    spreadsheetOffset: DOMCoordinates,
```
LucasLefevre, line 195: maybe `spreadsheetOffset` should be a class attribute ? Not sure what is the logic between class attribute vs `getCSSStyle` params here.
One could see attributes (rects, propsMaxSize) as external values/constraints (not linked to the element itself). Spreadsheet offset is also something not related to the element

src/components/popover/popover.ts
```diff
@@ -1,112 +1,303 @@
...
+    actualElementHeight: number,
+    shouldRenderAtBottom: boolean
+  ): number;
+
+  protected abstract getLeftCoordinate(
+    actualElementWidth: number,
+    shouldRenderAtRight: boolean
+  ): number;
+
+  getCSSStyle(
```
LucasLefevre, line 193: maybe just `getCSS` ? or `getElementCSS` ?

Also, we should harmonize our naming. Either `...Css` or `...CSS`. We currently have both.

LucasLefevre (commented):
Nice one :)
The Popover API looks much better and flexible

src/components/menu/menu.ts
```diff
@@ -117,23 +119,29 @@ export class Menu extends Component<Props, SpreadsheetChildEnv> {
...
-      marginTop,
-      flipHorizontalOffset: MENU_WIDTH * (this.props.depth - 1),
-      flipVerticalOffset: isRoot ? 0 : MENU_ITEM_HEIGHT,
+      anchorRect: {
+        x: this.props.position.x - MENU_WIDTH * (this.props.depth - 1),
+        y: this.props.position.y,
+        width: isRoot ? 0 : MENU_WIDTH,
+        height: isRoot ? 0 : MENU_ITEM_HEIGHT,
+      },
+      positioning: isRoot ? "BottomLeft" : "TopRight",
```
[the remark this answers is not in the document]
hokolomopo: That's an excellent point 🙈

src/components/popover/popover.ts
```diff
@@ -1,112 +1,303 @@
...
-  }
-
-  private get shouldRenderBottom(): boolean {
-    const { y } = this.props.position;
+  private popoverRef = useRef("popover");
+
+  setup() {
+    // useEffect occurs after the DOM is created and the element width/height are computed, but before
+    // the element in rendered, so we can still set its position
+    useEffect(() => {
```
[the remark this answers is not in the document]
hokolomopo: Do you think having `el/el.parent.getBoundingClientRect()` as dependency is fine? I feel like it might mess up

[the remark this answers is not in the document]
hokolomopo: And (Owl noo here) is there a better way to create dependencies than that ? 

```suggestion
      () => [
        ...Object.values(this.props.anchorRect),
        ...Object.values(this.props.containerRect || {}),
        this.props.maxHeight,
        this.props.maxWidth,
        this.props.positioning,
        this.props.verticalOffset,
        ...Object.values(this.popoverRef.el?.getBoundingClientRect() || {}),
        ...Object.values(this.popoverRef.el?.parentElement?.getBoundingClientRect() || {}),
      ]
```
? I cannot put the object themselves in the dependancies since owl check if the references are the same, not the values

src/components/popover/popover.ts
hokolomopo: Soudns good to me

src/components/popover/popover.ts
```diff
@@ -1,112 +1,303 @@
...
+
+  protected get availableHeightUp() {
+    return this.anchorRect.y - this.containerRect.y;
+  }
+
+  protected get availableHeightDown() {
     return (
-      y + this.props.childHeight <
-      this.viewportDimension.height + (this.env.isDashboard() ? 0 : TOPBAR_HEIGHT)
+      this.containerRect.y +
```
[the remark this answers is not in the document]
hokolomopo: Slightly more readable i'd say, but all the `getAvailable*` are kinda hard to read anyway, that's some low level stuff that you have to think a bit to understand each of them.

src/components/popover/popover.ts
```diff
@@ -1,112 +1,303 @@
...
+    return (
+      this.containerRect.x + this.containerRect.width - (this.anchorRect.x + this.anchorRect.width)
+    );
+  }
+
+  protected get availableWidthLeft() {
+    return this.anchorRect.x - this.containerRect.x;
+  }
+
+  protected getTopCoordinate(actualElementHeight: number, shouldRenderAtBottom: boolean): number {
```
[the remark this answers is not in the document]
hokolomopo: "actual" makes the difference between the size of the elements with getElBoundingRect and its size limited with the maxHeight/Width we computed. It makes sense in `getCSSStyle()`, not so much here.

tests/components/popover.test.ts
```diff
@@ -0,0 +1,289 @@
...
+let fixture: HTMLElement;
+let model: Model;
+let app: App;
+
+interface MountPopoverArgs extends Partial<PopoverProps> {
+  childWidth: Pixel;
+  childHeight: Pixel;
+}
+
+async function mountTestPopover(args: MountPopoverArgs) {
```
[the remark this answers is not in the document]
hokolomopo: stonks

src/components/popover/popover.ts
hokolomopo: TBH the logic I had was just "Is it needed at several places or not". Putting `spreadsheetOffset` is fine by me.

src/components/popover/popover.ts
hokolomopo: I think that `CSS` makes a bit more sense, but `Css` is prettier with CamelCase so I'll use that I think :)

src/components/popover/popover.ts
```diff
@@ -80,7 +80,10 @@ export class Popover extends Component<PopoverProps, SpreadsheetChildEnv> {
         };
 
         const anchor = rectIntersection(this.props.anchorRect, containerRect);
-        if (!anchor) throw new Error("The popover anchorRect should intersect its containerRect");
+        if (!anchor) {
+          el.style.visibility = "hidden";
```
LucasLefevre: nice :)

LucasLefevre (approved):
maybe I'll let the others take a look

hokolomopo:
TODO : 
- bottom bar popover not working in odoo ? 
- sub-menu can overflow out of grid
- sub-menu isn't hidden when parent is hidden

src/components/menu/menu.xml
```diff
@@ -6,6 +6,8 @@
       containerRect="props.menuContainerRect"
       positioning="popover.positioning"
       verticalOffset="popover.verticalOffset"
+      onPopoverHidden="() => this.closeSubMenu()"
+      onPopoverMove="() => this.closeSubMenu()"
```
LucasLefevre, line 10: `onPopoverMoved` ? to be consistent with "hidden"

src/components/figures/figure_image/figure_image.ts
```diff
@@ -3,12 +3,13 @@ import { MENU_WIDTH } from "../../../constants";
...
 import { MenuItemRegistry } from "../../../registries/index";
 import { _lt } from "../../../translation";
-import { DOMCoordinates, Figure, SpreadsheetChildEnv, UID } from "../../../types";
+import { DOMCoordinates, Figure, Rect, SpreadsheetChildEnv, UID } from "../../../types";
 import { useAbsolutePosition } from "../../helpers/position_hook";
 import { Menu, MenuState } from "../../menu/menu";
 
 interface Props {
   figure: Figure;
+  gridRect: Rect;
```
rrahir, line 12: if we keep it, it should be visible in the static ImageFIgure.props

src/components/popover/popover.ts
rrahir: might be worth making a dedicade helper just like useSpreadsheetPosition to find this information ?

src/helpers/rectangle.ts
```diff
@@ -0,0 +1,28 @@
+import { Rect, Zone } from "../types";
+import { intersection } from "./zones";
+
+/**
+ * Compute the intersection of two rectangles. Returns nothing if the two rectangles don't overlap
+ */
+export function rectIntersection(rect1: Rect, rect2: Rect): Rect | undefined {
+  return zoneToRect(intersection(rectToZone(rect1), rectToZone(rect2)));
```
pro-odoo, line 8: It is useful to introduce two helpers to convert a rect to a non-really zone just to use `intersection` ? It's two `if` (overlap) and basic operations (math.Max/Min).
For now, `zoneToRect` and `rectToZone` are private, but could be easily exported. If it's the case, I could already see myself in 6 months trying to figure why the f* we convert a rect to a zone without the getters.
The only interest I see is that `intersection` is already tested.
What do you think ?

src/components/grid/grid.ts
```diff
@@ -334,6 +335,10 @@ export class Grid extends Component<Props, SpreadsheetChildEnv> {
     return this.hoveredCell.col === col && this.hoveredCell.row === row;
   }
 
+  get gridRect(): Rect {
```
pro-odoo, line 338: I'm not a great fan of this. We have to "trimballer" `gridRect` from the top component to every children and children of children to be sure that **IF** a child use popover, this popover will correctly placed. It makes me feel that we introduce here a coupling between all our components.
It's typically a good use case for `useChildSubEnv`, you can add a function `popoverContainerRect` in children env, it will be automatically propagated and use by the popover component directly. With that, intermediate components should not take care of "trimballer" the props.

src/components/popover/popover.ts
pro-odoo: Can `useSpreadsheetPosition` make the coffee ?

src/components/popover/popover.ts
```diff
@@ -1,112 +1,294 @@
...
+      const elDims = {
+        width: el.getBoundingClientRect().width,
+        height: el.getBoundingClientRect().height,
+      };
+
+      const spreadsheetRect = el.parentElement!.getBoundingClientRect();
+      const spreadsheetOffset = { x: spreadsheetRect.left, y: spreadsheetRect.top };
+
+      const containerRect = {
+        x: this.props.containerRect?.x || spreadsheetOffset.x,
```
pro-odoo, line 83: if `this.props.containerRect?.x` is `0`, it will take `spreadsheetOffset.x`.
Same below

src/components/popover/popover.ts
```diff
@@ -1,112 +1,294 @@
...
+      this.currentPosition = newPosition;
+    });
+  }
+
+  private resetPopoverElStyle() {
+    const el = this.popoverRef.el;
+    if (!el) return;
+    el.style["visibility"] = "visible";
+    el.style["max-height"] = el.style["max-width"] = "";
+    el.style["bottom"] = el.style["top"] = el.style["right"] = el.style["left"] = "";
```
pro-odoo, line 122: Pretty sure there is a more "owly" way to achieve that.

src/components/popover/popover.ts
```diff
@@ -1,112 +1,294 @@
...
+  maxWidth: { type: Number, optional: true },
+  maxHeight: { type: Number, optional: true },
+  verticalOffset: { type: Number, optional: true },
+  onMouseWheel: { type: Function, optional: true },
+  onPopoverHidden: { type: Function, optional: true },
+  onPopoverMove: { type: Function, optional: true },
+  slots: Object,
+};
+
+abstract class PopoverPositionContext {
```
pro-odoo, line 139: I like it !

src/components/popover/popover.ts
```diff
@@ -1,112 +1,294 @@
...
+  private getMaxWidth(elementWidth: number) {
+    const shouldRenderAtRight = this.shouldRenderAtRight(elementWidth);
+    const availableWidth = shouldRenderAtRight ? this.availableWidthRight : this.availableWidthLeft;
+
+    return this.propsMaxSize.width
+      ? Math.min(availableWidth, this.propsMaxSize.width)
+      : availableWidth;
+  }
+
+  getCss(elDims: DOMDimension, verticalOffset: number): CSSProperties {
```
pro-odoo, line 191:
```suggestion
  getCSS(elDims: DOMDimension, verticalOffset: number): CSSProperties {
```

pro-odoo (commented):
It's way better

src/components/grid/grid.xml
```diff
@@ -70,7 +68,7 @@
       <Menu
         t-if="menuState.isOpen"
         menuItems="menuState.menuItems"
-        menuContainerRect="gridRect"
+        menuContainerRect="getGridRect()"
```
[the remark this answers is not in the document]
hokolomopo: Yeah the menus in the topBar don't have the same container rect. They're not child of the grid at all anyway, so they won't have the same env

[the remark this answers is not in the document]
hokolomopo: Not needed if I use a hook `usePopoverContainer` like Pierre suggested tho.

pro-odoo (approved):
robodoo r+

PR #1447, opened 2022-06-16
[IMP] spreadsheet: Add data filters

hokolomopo:
## Description:

Add data filters to o_spreadsheet. It is now possible to create tables in
sheets that allow to filter the rows that contain certain values.

Add a Filter plugin that manage the data of the filters and adapt their zone
with the grid modification. A new UI plugin FilterEvaluation will keep track
of which rows are currently filtered by the defined filters.

Split the HeaderVisibility plugin in a core plugin and an UI plugin.
The core plugin will manage the col/row hidden by the user, and the UI plugin
will combine these with the filtered rows. Many visibility getters are thus
moved in UIGetters and cannot be accessed in core plugins.

Commit 46e3795 changed adaptZone to not extend the zone if a new col/row was
added at the start of the zone. This commit make adaptZone have the same
behaviour if a new col/row was added at the end of the zone, the zone won't
be extended.

Add 2 new options to the sort command:
- hasNoHeader: disable the detection of header in the sorted zone
- emptyCellAsZero: empty cells are treated as number cells with value of 0

Moved the Sort items of the top bar menu to a new category "Data" that also
contains menu items to create/remove filters.

Odoo task ID : [2585911](https://www.odoo.com/web#id=2585911&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


hokolomopo:
TODO : filter table border probably shouldn't overwrite border of cells

LucasLefevre (commented):
another batch of comments :)

src/plugins/core/filters.ts
```diff
@@ -0,0 +1,333 @@
...
+          cmd.dimension === "COL" ? "left" : "top",
+          cmd.base,
+          cmd.position,
+          cmd.quantity
+        );
+        filters.push(new Filter(filter.id, filterZone));
+      }
+
+      // Add filters for new columns
+      if (filters.length < zone.right - zone.left + 1) {
```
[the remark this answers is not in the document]
hokolomopo: Sure :)

src/plugins/core/filters.ts
```diff
@@ -0,0 +1,333 @@
...
+    }
+    return false;
+  }
+
+  private onAddColumnsRows(cmd: AddColumnsRowsCommand) {
+    for (const filterTable of this.getFilterTables(cmd.sheetId)) {
+      if (!filterTable) {
+        return;
+      }
+      const zone = expandZoneOnInsertion(
```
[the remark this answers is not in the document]
hokolomopo: I originally wanted to use ranges, but it doesn't work that well.

One issue is that every time a row is added/removed in the filter table, I have to add/remove items in the `filters[]` array of the table. So after adapting a range, I have to check if new cols were added/removed (I don't even know what command triggered the change in the adapt ranges), then rebuild the filters array.

The other issue is that I'm pretty sure I just cannot use ranges inside `Filters` atm, because inserting a cell above a table column would shift is down, which we really don't want. Or we'd need additional logic "only update Filter range if the FilterTabel range was updated".

In the end, I think using ranges could be done, but wouldn't really improve the code.

src/plugins/ui/filter_evaluation.ts
```diff
@@ -0,0 +1,165 @@
...
+
+  private filterValues: Record<UID, Record<FilterId, string[]>> = {};
+
+  hiddenRows: Set<number> = new Set();
+  isEvaluationDirty = false;
+
+  allowDispatch(cmd: Command) {
+    switch (cmd.type) {
+      case "UPDATE_FILTER":
+        if (!this.getters.getFilterId(cmd.sheetId, cmd.col, cmd.row)) {
```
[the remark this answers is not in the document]
hokolomopo: `UPDATE_FILTER` is a local command

src/plugins/core/filters.ts
```diff
@@ -0,0 +1,333 @@
...
+  /**
+   * Check if an UpdateCell command should cause the given table to be extended by one row.
+   *
+   * The table should be extended if all three condition are true:
+   * 1) The updated cell is right below the table
+   * 2) The command adds a content to the cell
+   * 3) No cell right below the table had any content before the command
+   *
+   */
+  private canUpdateCellCmdExtendTable(
```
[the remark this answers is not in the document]
hokolomopo: Yeah, same thing with merge, I'll fix this and add tests

src/plugins/core/filters.ts
```diff
@@ -0,0 +1,355 @@
...
+
+    const zone = table.zone;
+    if (!(zone.bottom + 1 === row && col >= zone.left && col <= zone.right)) {
+      return false;
+    }
+
+    for (const col of range(zone.left, zone.right + 1)) {
+      // Since this plugin is loaded before CellPlugin, the getters still give us the old cell content
+      const cellContent = this.getters.getCell(sheetId, col, row)?.content;
+      if (cellContent !== undefined) {
```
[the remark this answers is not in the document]
rrahir: > Correct me if i'm wrong, but the only case the content should be an empty string is if the cell is a formula returning an empty string no? In this case, there is an (invisible) content in the cell, and we shouldn't extend the table. Thus the condition is correct.

if the cell is a formula, then it's content will be `={formula stuff}`, it's not the evaluated value ;-)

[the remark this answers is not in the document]
hokolomopo: True, I switched to `if(cellContent)`

src/components/filters/filter_menu/filter_menu.ts
```diff
@@ -0,0 +1,352 @@
...
+
+    this.state.selectedValue =
+      selectedIndex !== undefined ? displayedValues[selectedIndex].string : undefined;
+    this.scrollListToSelectedValue(selectedIndex);
+  }
+
+  /** If the selected value is out of the list, scroll until it's displayed */
+  private scrollListToSelectedValue(selectedIndex: number | undefined) {
+    const listElement = this.valueListRef.el;
+    const listItem = listElement?.querySelector(".o-filter-menu-item") as HTMLElement;
```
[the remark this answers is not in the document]
rrahir: alternatively, , create a small component 'menuValue' that would scroll itself into view with the right prop

tests/collaborative/ot/ot_columns_added.test.ts
```diff
@@ -112,34 +114,46 @@ describe("OT with ADD_COLUMNS_ROWS with dimension COL", () => {
...
+        const result = transform(command, addColumnsAfter);
+        expect(result).toEqual(command);
+      });
+      test(`${cmd.type} with two targets, one before and one after`, () => {
+        const command = { ...cmd, target: [toZone("A1:A3"), toZone("M1:O2")] };
+        const result = transform(command, addColumnsAfter);
+        expect(result).toEqual({ ...command, target: [toZone("A1:A3"), toZone("O1:Q2")] });
+      });
+    }
+  );
```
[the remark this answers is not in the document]
hokolomopo: `UPDATE_FILTER` is a local command, so I do'nt think we need to transform anything.
I added transformations for merges and create_table :)

src/plugins/ui/sort.ts
```diff
@@ -259,13 +266,17 @@ export class SortPlugin extends UIPlugin {
     // Update in case of merges in the zone
     let cells = this.mainCells(sheetId, zone);
 
-    if (this.hasHeader(cells)) {
+    if (!options.hasNoHeader && this.hasHeader(cells)) {
```
rrahir, line 269: shoundl't this option be called  `ignoreHeader`?

src/helpers/filters.ts
```diff
@@ -0,0 +1,65 @@
...
+  /** Filtered zone, ie. zone of the filter without the header */
+  get filteredZone(): Zone | undefined {
+    const zone = this.zoneWithHeaders;
+    if (zone.bottom === zone.top) {
+      return undefined;
+    }
+    return { ...zone, top: zone.top + 1 };
+  }
+
+  clone(): Filter {
```
LucasLefevre, line 62: implementing `Cloneable` seems useless :)

src/collaborative/ot/ot_specific.ts
```diff
@@ -141,3 +151,24 @@ function freezeTransformation(
...
+function createTableTransformation(
+  cmd: CreateFilterTableCommand | AddMergeCommand,
+  executed: CreateFilterTableCommand
+): CreateFilterTableCommand | AddMergeCommand | undefined {
+  if (cmd.sheetId !== executed.sheetId) {
+    return cmd;
+  }
+
+  for (const cmdTarget of cmd.target) {
+    for (const executarget of executed.target) {
```
LucasLefevre, line 167: `executarget` :smile:

src/collaborative/ot/ot_specific.ts
```diff
@@ -141,3 +151,24 @@ function freezeTransformation(
   }
   return quantity > 0 ? { ...cmd, quantity } : undefined;
 }
+
+/**
+ * Cancel CREATE_FILTER_TABLE and ADD_MERGE commands if they overlap a filter
+ */
+function createTableTransformation(
```
LucasLefevre, line 158: it could be the exact same transformation as `mergeTransformation`, right ? (should be renamed if it's reused)
It wouldn't drop the entire command (just the overlapping zoes), but it's ok from a functional point of view I think

src/helpers/filters.ts
hokolomopo: No because it allow me to do a deepCopy at sheet duplication

src/collaborative/ot/ot_specific.ts
hokolomopo: It's different because we don't create a table for each target in `CREATE_FILTER_TABLE `, but we merge the targets together to create a single table (if the targets are continuous, else the allowdispatch prevent everythin).
So it make sense here to drop the whole command if a target overlap a merge/another table

src/helpers/filters.ts
LucasLefevre: `clone` of `FilterTable` already "clones" the `filters` by building new ones in its constructor :wink:

src/plugins/ui/sort.ts
hokolomopo: Not that my name is any good, but wouldn't `ignoreHeader` imply that there is headers, but that we leave them as it ?

LucasLefevre: `ignoreFirstRow` ? :man_shrugging:

src/plugins/ui/sheetview.ts
```diff
@@ -220,6 +228,10 @@ export class SheetViewPlugin extends UIPlugin {
   }
 
   finalize() {
+    for (const sheetId of this.sheetsWithDirtyViewports) {
```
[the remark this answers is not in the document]
hokolomopo: rip

src/plugins/ui/sort.ts
rrahir: > Not that my name is any good, but wouldn't `ignoreHeader` imply that there is headers, but that we leave them as it ?

ignoreHeader would imply that even if there are headers (which is determined by the function hasHeader) we would ignore them, which is exactly what you are doing.  I stand by my suggestion ^^ I also feel it is less cryptic than ignoreFirstRow

hokolomopo: How about `sortHeaders` ?

src/collaborative/ot/ot_specific.ts
```diff
@@ -46,6 +51,11 @@ otRegistry.addTransformation(
   ["FREEZE_COLUMNS", "FREEZE_ROWS"],
   freezeTransformation
 );
+otRegistry.addTransformation(
```
LucasLefevre, line 54: add transformation tests please :) :see_no_evil:

LucasLefevre (commented):
Almost ready to merge I think :partying_face:

src/plugins/core/filters.ts
```diff
@@ -0,0 +1,343 @@
...
+      }
+    }
+  }
+
+  export(data: WorkbookData) {
+    for (const sheet of data.sheets) {
+      for (const filterTable of this.getFilterTables(sheet.id)) {
+        sheet.filterTables.push({
+          range: zoneToXc(filterTable.zone),
+        });
```
LucasLefevre, line 335:
```suggestion
        sheet.filterTables.push(zoneToXc(filterTable.zone));
```
Any value to have this object with a `range` key?

src/components/filters/filter_menu/filter_menu.ts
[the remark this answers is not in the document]
hokolomopo: @rrahir added a commit in which I do that. I don't need a querySelector anymore. I save which value I should scroll to (and from which direction), then give it to the FilterMenuValueItem.  component as props.

src/collaborative/ot/ot_specific.ts
hokolomopo: I already added them in `tests\collaborative\collaborative.test.ts`. Or am I missing some ?

src/plugins/ui/sort.ts
rrahir: not bad

src/plugins/core/filters.ts
hokolomopo: Since I remove the filtered values from core, not really.
But having `{range : "xc"}` instead of a  simple string also meansa bit less work/migration when/if in the future we add things to Tables (styles, tables name, ...). 
I can change if you want, but this doesn't really affect anything.

src/collaborative/ot/ot_specific.ts
hokolomopo: Added pure ot tests in `tests\collaborative\ot\ot.test.ts` and `tests\collaborative\ot\ot_merged.test.ts`

src/plugins/core/filters.ts
LucasLefevre: I have mixed feeling.
On one hand you've got a point it's more future proof. On the other I like to keep things simple while we don't need it (YAGNI)
Neither Google nor Excel seem to have any options to filter table. And I don't think we will either
Hence I'd choose the simple version

hokolomopo: Excel definitevely have tons of options to add to tables. Nos so much for GSheet.
But I get your point, I'll change it.

LucasLefevre (approved):
robodoo r+ :partying_face: :tada:

PR #1613, opened 2022-08-31
[FIX] Chart Panel: Chart title is updated when typing

anhe-odoo:
## Task Description

When having two charts on a sheet, if we change the title of the first one (without typing enter) and then select the second chart, the new title is set for the second chart instead of the first one.

## Reproducibility
This issue can be reproduced following these steps:
1. Create two charts
2. Go to the settings panel of the first one
3. Type a new title (but don't click out yet)
4. Click out on the second chart

## Fix Description
This issue comes from the fact we linked the title update to the "onChange" event of the input textbox. Unfortunately,
this event is triggered when the textbox loose focus, which happen here when clicking on the second chart. But clicking
on the second chart also change the id of the figure we try to edit, meaning the onChange event is called only after
having changed the target figure.
In this fix, we use another event (onInput), leading us to change the title each time the value in the textbox is changed.

## Related task
Odoo task ID : [2961701](https://www.odoo.com/web#id=2961701&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## Review Checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/chart/main_chart_panel/main_chart_panel.ts
+++ b/src/components/side_panel/chart/main_chart_panel/main_chart_panel.ts
@@ -51,6 +51,7 @@ export class ChartPanel extends Component<Props, SpreadsheetChildEnv> {
   static template = "o-spreadsheet-ChartPanel";
 
   private state!: State;
+  private shouldUpdateChart: boolean = true;
 
   get figureId(): UID {
     return this.state.figureId;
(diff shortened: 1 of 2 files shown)


rrahir:
@robodoo r+

PR #1627, opened 2022-09-07
[FIX] Topbar: closes dropdown on right-click

anhe-odoo:
## Task Description

When opening a dropdown menu in the topBar (may it be in the menu or from an action button), and right-clicking somewhere else after (e.g. in a cell), you will get bot the context menu of the clicked component and the previously opened dropdown menu, while the last should be closed. This is fixed here by adding an eventListener on the contextMenu event

## Related Task
Odoo task ID : [2862746](https://www.odoo.com/web#id=2862746&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## Review Checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/color_picker/color_picker.xml
+++ b/src/components/color_picker/color_picker.xml
@@ -1,6 +1,9 @@
 <templates>
   <t t-name="o-spreadsheet-ColorPicker" owl="1">
-    <div class="o-color-picker" t-att-class="props.dropdownDirection || 'right'">
+    <div
+      class="o-color-picker"
+      t-att-class="props.dropdownDirection || 'right'"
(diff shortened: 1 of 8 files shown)


pro-odoo (changes requested):
I think this task requires a bigger refactoring, to allow only one menu/popover at the same time on the page. They find a good way to do that in Odoo, we should check that.
But we will discuss it in the near future ;-)

rrahir:
Task closed : will be addressed in a broader ref wit a menu service

PR #1699, opened 2022-10-07
[FIX] tests: fix in-determinist helper

pro-odoo:
Before this revision, we had one test non-determinist, but this test was actually false. The falsy part of the snapshot was the position of the autofill, which was not placed correctly because of the absence of re-rendering after a `RESIZE_SHEETVIEW`.

Here is how the rendering worked before this revision to render a spreadsheet:

- Render the Spreadsheet Component - Render the Grid - onMounted of Grid => dispatch of `RESIZE_SHEETVIEW` - trigger "update" - onMounted of Spreadsheet => register "update" event to re-render

With this process, "update" event was triggered **before** the registration of the event.
=> Spreadsheet was not re-rendered so the autofill was not correctly placed.

With this revision, we register the listener during the setup of Spreadsheet Component.

A second issue which explains why the test was non-determinist is that the test had a `await nextTick()` inside. This `nextTick` have to be in the helper `mountSpreadsheet`.

This revision highlight another incorrect test, which is fixed in this revision

Co-authored-by: Rémi Rahir <rar@odoo>

### DIFF
--- a/tests/components/__snapshots__/grid.test.ts.snap
+++ b/tests/components/__snapshots__/grid.test.ts.snap
@@ -63,7 +63,7 @@ exports[`Grid component simple rendering snapshot 1`] = `
   
   <div
     class="o-autofill"
-    style="top:19px;left:92px"
+    style="top:45px;left:140px"
   >
     <div
       class="o-autofill-handler"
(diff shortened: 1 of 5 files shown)


tests/components/__snapshots__/grid.test.ts.snap
```diff
@@ -63,7 +63,7 @@ exports[`Grid component simple rendering snapshot 1`] = `
   
   <div
     class="o-autofill"
-    style="top:19px;left:92px"
+    style="top:45px;left:140px"
```
pro-odoo, line 66: Note to reviewer: 19 (old value) + HEADER_HEIGHT = 45; 92 + HEADER_WIDTH = 140

LucasLefevre (approved):
robodoo r+

LucasLefevre:
robodoo r-
commit message ? `non deterministic` ?

LucasLefevre:
robodoo r+

PR #1715, opened 2022-10-10
[IMP] Charts: improve side-panel clarity

anhe-odoo:
## Task Description:

This task aims to add some clarity in the invalid labels/ series ranges handling. The confirm button is now hide when at least one range is invalid, and a new reset button appears in the same case, allowing the user to revert the ranges to the last valid ranges.

## Related task(s)
- Odoo task ID : [2877163](https://www.odoo.com/web#id=2877163&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## Review Checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/components/selection_input/selection_input.ts
```diff
@@ -245,19 +241,74 @@ export class SelectionInput extends Component<Props, SpreadsheetChildEnv> {
...
+            id: this.id,
+            rangeId: range.id,
+            value: zoneToXc({
+              left: zone.left,
+              right: zone.left,
+              top: zone.top,
+              bottom: zone.bottom,
+            }),
+          });
+          for (var j = zone.left + 1; j <= zone.right; ++j) {
```
LucasLefevre, line 267: Splitting the zone on columns is business logic related to charts.
Nothing tells *every use* of the `<SelectionInput/>` component has the same logic.
It doesn't belong here.

It should probably be something updated with the component `"range"` prop.
I don't know why its value is defined only at `setup` in the config panel component and why they aren't dynamic

src/components/selection_input/selection_input.ts
```diff
@@ -245,19 +241,74 @@ export class SelectionInput extends Component<Props, SpreadsheetChildEnv> {
...
+            id: this.id,
+            rangeId: range.id,
+            value: zoneToXc({
+              left: zone.left,
+              right: zone.right,
+              top: zone.top,
+              bottom: zone.top,
+            }),
+          });
+          for (var j = zone.top + 1; j <= zone.bottom!; ++j) {
```
LucasLefevre, line 289: please never use `var`
[4 Reasons Why ‘var’ is Considered Obsolete in Modern JavaScript](https://javascript.plainenglish.io/4-reasons-why-var-is-considered-obsolete-in-modern-javascript-a30296b5f08f)

LucasLefevre (changes requested):

src/components/selection_input/selection_input.ts
```diff
@@ -62,6 +78,7 @@ interface Props {
   class?: string;
   onSelectionChanged?: (ranges: string[]) => void;
   onSelectionConfirmed?: () => void;
+  postProcessSelection?: (ranges: (RangeInputValue & { isFocused: boolean })[], id: string) => void;
```
LucasLefevre, line 81: the SelectionInput API is already SOTG. This doesn't make it easier.
IMO we should check how `onSelectionChanged` and `onSelectionConfirmed` are used and clarify the situation.
Currently, the state is duplicated inside the plugin and in the parents (check `onSelectionChanged`)

I think we can explore this behavior:
1. when not focused: just display ranges from it's props (no internal state)
2. when focused: display it's internal state
3. when confirmed: send the changes to its parent (with onSelectionConfirmed), clean it's internal state and get back to 1.

That way, I think `spreadRange` would not be needed at all. We would just rely on the existing business logic of charts/CF which, I assume, already do the very same job of splitting ranges

src/components/selection_input/selection_input.ts
```diff
@@ -143,7 +168,8 @@ export class SelectionInput extends Component<Props, SpreadsheetChildEnv> {
 
   checkChange() {
     const value = this.env.model.getters.getSelectionInputValue(this.id);
-    if (this.previousRanges.join() !== value.join()) {
+    const valid = !this.isInvalid && this.ranges.every((range) => range.isValidRange);
```
LucasLefevre, line 171: why isn't `this.ranges.every((range) => range.isValidRange)` part of `isInvalid` then ?

src/components/selection_input/selection_input.ts
LucasLefevre: Ok, sorry, it can't work like what I proposed. From what I see now, the "Save" button in the CF side panel takes over the "Confirm" button, so the state must be held in the parent component.
This API requires a bit more thoughts then that

src/components/selection_input/selection_input.ts
anhe-odoo: For what I understand, isInvalid is coming from the parent (so it should be related to the parent) and is optionnal, so it can be totally not related to the fact a range is a valid range

src/helpers/range.ts
```diff
@@ -194,3 +195,43 @@ export function copyRangeWithNewSheetId(sheetIdFrom: UID, sheetIdTo: UID, range:
 export function createRange(getters: CoreGetters, sheetId: UID, range?: string): Range | undefined {
   return range ? getters.getRangeFromSheetXC(sheetId, range) : undefined;
 }
+
+/**
+ * Spread multiple colrows zone to one row/col zone and add a many new input range as needed.
+ * For example, A1:B4 will become [A1:A4, B1:B4]
+ */
+export function spreadRange(ranges: string[]): string[] {
```
rrahir, line 203: I think we introduced some behaviour recently that made the assumption that datasets would only be served as columns. This function still raises the question of unbounded zones, how should we handle them? I need to look further into it, we might have gone wrong with a recent spec

In any case, the docstring is incomplete, we don't really know how the spread will take place (at first glance ,I would have sworn it was spitting by column but turns out I wa wrong)

tests/components/charts.test.ts
```diff
@@ -566,31 +651,72 @@ describe("figures", () => {
...
         await simulateClick(".o-data-labels input");
-        setInputValueAndTrigger(".o-data-labels input", "this is not valid", "input");
+        setInputValueAndTrigger(".o-data-labels input", "A1", "input");
+        await nextTick();
+        expect(fixture.querySelectorAll(".o-data-labels .o-selection-ok").length).toBe(1);
+      }
+    );
+
+    test.each(["basicChart", "scorecard"])(
+      "update chart with invalid labels/baseline hide confirm button/show reset button",
```
rrahir, line 700: the confirm/show reset should also be tested in the selection_input component (actually, it makes more sense to test it without the chart context unless necessary)

src/components/icons/icons.xml
```diff
@@ -524,4 +524,14 @@
       />
     </svg>
   </t>
+  <t t-name="o-spreadsheet-Icon.ERROR" owl="1">
+    <svg class="error-icon" width="14" height="14" viewBox="0 0 512 512">
+      <!-- Font Awesome Free 5.15.4 by @fontawesome - https://fontawesome.com License - https://fontawesome.com/license/free (Icons: CC BY 4.0, Fonts: SIL OFL 1.1, Code: MIT License) -->
+      <path
+        fill="#000000"
```
rrahir, line 531: cs wizard striked again; https://github.com/odoo/o-spreadsheet/pull/2067/files
```suggestion
        fill="currentColor"
```

src/components/selection_input/selection_input.xml
```diff
@@ -7,23 +7,30 @@
...
+              'o-required': props.required,
+              'o-optional': !props.required,
+              'o-invalid': isInvalid || !range.isValidRange,
+              'text-decoration-underline': range.isFocused and state.mode === 'select-range'
+            }"
+            t-ref="{{range.isFocused ? 'focused' : 'unfocused'}}Input"
+          />
+          <span t-if="isInvalid || !range.isValidRange" class="icon" title="This range is invalid">
+            <t t-call="o-spreadsheet-Icon.ERROR">
+              <t t-set="color" t-value="'fill : #FF0000'"/>
```
rrahir, line 30: based on  https://github.com/odoo/o-spreadsheet/pull/2067/files, you can define a css rule to set the color on the parent span.

src/components/icons/icons.xml
```diff
@@ -524,4 +524,14 @@
...
+        t-att-style="color"
```
rrahir, line 532: you can remove this

rrahir (commented):
👍   just small comments

src/components/selection_input/selection_input.ts
```diff
@@ -58,12 +69,17 @@ css/* scss */ `
         font-size: 14px;
         height: 25px;
       }
+      .icon {
+        position: absolute;
+        right: 8px;
```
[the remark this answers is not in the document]
anhe-odoo: What do you mean ? Something like a computation in the selection_input ?

src/components/selection_input/selection_input.ts
```diff
@@ -58,12 +69,18 @@ css/* scss */ `
         font-size: 14px;
         height: 25px;
       }
+      .icon {
+        position: absolute;
+        right: 7px;
+        top: 7px;
+        color: #ff0000;
```
rrahir, line 76: you should use bootstrap class text-danger instead ;-)

src/helpers/range.ts
```diff
@@ -194,3 +195,43 @@ export function copyRangeWithNewSheetId(sheetIdFrom: UID, sheetIdTo: UID, range:
...
+  const postProcessedRanges: string[] = [];
+  for (const range of ranges) {
+    const sheetName = range.includes("!") ? range.split("!")[0] + "!" : "";
```
rrahir, line 206: Use `splitReference`,  the exclamantion mark may be part of a sheetname, which would break your code

rrahir (commented):
small generic comments :) 

- there is a typo in the first commit message
- the second commit title is not complete and it should mention that it is specifically about
  the charts(which is a good thing, we have other usecase where we don't want to split the  
  ranges by column  ;-) )
- `existingSelectionRange` should be plural, we copy pasted it a several places but its a list or
   ranges
- style could be more bootstrap oriented

there are still some inconsistencies in the selection input which I think are due to the use of `getSelectionInputValue` vs `getSelectionInput`.  
Example, in demo data, edit the line graph, add some empty ranges[1] then confirm. The empty ranges are still visible in the component but they are not forwarded to the parent callbacks. So unmounting an remounting the component -> the empty ranges disappear.
hitting 'confirm' should probably purge those.



[1] note that you can add them at any place since a recent introduction of dynamic inputs when we type ',' (comma) inside an input field of the selectioninput component.

src/components/selection_input/selection_input.ts
```diff
@@ -19,57 +18,58 @@ const uuidGenerator = new UuidGenerator();
...
+        border: 1px solid;
       }
       input.o-focused {
-        border-color: ${SELECTION_BORDER_COLOR};
+        border-width: 2px;
+        padding: 3px 5px;
       }
       input.o-invalid {
-        border-color: red;
+        /* le background-color mis ici est similaire à la classe bootstrap alert-danger mais a cause du commit 0358a76d,
```
LucasLefevre, line 40: coucou french :)

rrahir:
@robodoo r+

PR #1718, opened 2022-10-11
[IMP] xlsx: import and export data filters

hokolomopo:
This commit is a follow-up of 5d6eec44 and add xlsx import/export for data filters.

Some notes on the import/export:

- Filtered values are no imported as they are non-persistent in o_spreadsheet, contrarily to Excel

- Filtered rows as imported as standard hidden rows. Checking at import if a row was hidden by the user of by a data filter would be really hard. Note that Google Sheet does the same.

- Filtered Values are exported for Excel

- Tables are exported with a default table style (OpenXML TableStyleLight8, but note that Excel doesn't seems to follow the OpenXML specification for table styles)

Odoo task [3009053](https://www.odoo.com/web#id=3009053&cids=1&menu_id=4720&action=327&model=project.task&view_type=form)

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/xlsx/conversion/table_conversion.ts
```diff
@@ -46,7 +55,8 @@ function applyTableStyle(convertedData: WorkbookData, xlsxData: XLSXImportData)
 
   for (let xlsxSheet of xlsxData.sheets) {
     for (let table of xlsxSheet.tables) {
-      const sheet = convertedData.sheets.find((sheet) => sheet.name === xlsxSheet.sheetName)!;
+      const sheet = convertedData.sheets.find((sheet) => sheet.name === xlsxSheet.sheetName);
+      if (!sheet) continue;
```
LucasLefevre, line 59: :+1:

src/xlsx/functions/cells.ts
```diff
@@ -63,18 +63,19 @@ export function addFormula(cell: ExcelCellData): {
 
 export function addContent(
   content: string,
-  sharedStrings: string[]
+  sharedStrings: string[],
+  forceString = false
```
LucasLefevre, line 67: table header's content is already forced to strings in `exportForExcel`, right ?

src/xlsx/functions/table.ts
```diff
@@ -0,0 +1,91 @@
...
+}
+
+function addTableColumns(table: ExcelFilterTableData, sheetData: ExcelSheetData): XMLString {
+  const tableZone = toZone(table.range);
+  const columns: XMLString[] = [];
+  for (const i of range(0, zoneToDimension(tableZone).width)) {
+    const colHeaderXc = toXC(tableZone.left + i, tableZone.top);
+    const colName = sheetData.cells[colHeaderXc]?.content || `col${i}`;
+    const colAttributes: XMLAttributes = [
+      ["id", i + 1],
```
LucasLefevre, line 80: maybe deserve a comment saying +1 is normal ? :man_shrugging:

LucasLefevre (commented):
:+1:

src/xlsx/functions/cells.ts
hokolomopo: Well `cell.content` is always a string, but `addContent` use our `isNumber` helper that use regex to know if the content is a number or a string. 
Using `cell.value` doesn't work, because Error cells then will export their value (error message) instead of their content (formula in error). Maybe we can fix this (adding a cell.isError or setting cell.isFormula to true for the error cells), but I'm not sure it's better

src/xlsx/functions/table.ts
hokolomopo: Added a comment. The problem being that ids cannot be `0`

LucasLefevre (approved):
robodoo r+

PR #1733, opened 2022-10-18
[FIX] renderer: negative number with lot of decimal are overflowing

hokolomopo:
## Description

There was a problem with some negative numbers with decimals, the minus sign was overflowing and was displayed on the cell border. For example the result of CUMPRINC in the demo data.

There was actually 2 problems:

1) We define a default format for numbers with max 10 decimal digits. This don't take the length of the integer part (+ minus sign) into account. Replace this by having a default format that try to keep the whole number within 12 digits, by rounding the decimal part if possible.

2) The renderer didn't take into account the text margin when computing the clip boxes. This one was a bit tricky, because the alignment of the cell depends on whether the content is overflowing, but to know if the content overflows we need to know the margin... which depends on the alignment.

Worked around this issue by having all the cells have a margin equal to MIN_CELL_TEXT_MARGIN, no matter their alignment or content (with or w/o CF icon)

Odoo task ID : [2977643](https://www.odoo.com/web#id=2977643&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/helpers/format.ts
```diff
@@ -306,8 +312,24 @@ function formatJSTime(jsDate: Date, format: Format): FormattedValue {
...
+ * of digits in the integer part, keep the number as it is.
+ */
 export function createDefaultFormat(value: number): Format {
-  let { decimalDigits } = splitNumber(value, 10);
+  let { integerDigits, decimalDigits } = splitNumber(value);
+  if (!decimalDigits) return "0";
+
+  // If there's no space for at least the decimal separator + a decimal digit, don't clip the number
+  if (integerDigits.length + 2 > DEFAULT_FORMAT_NUMBER_OF_DIGITS) {
+    return "0." + "0".repeat(decimalDigits.length);
```
laa-odoo, line 328: functional question: why not return "0" ?

hokolomopo: I think that it's worth for the user to have the information that it's a decimal number, and not just an integer. 

But maybe not, I though about having "0" too, I wasn't too sure which one was better

src/helpers/cells/cell_types.ts
```diff
@@ -111,7 +111,7 @@ export class EmptyCell extends AbstractCell<EmptyEvaluation> {
 }
 
 export class NumberCell extends AbstractCell<NumberEvaluation> {
-  readonly content = formatValue(this.evaluated.value);
+  readonly content = String(this.evaluated.value);
```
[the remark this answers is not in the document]
laa-odoo: I think formatValue avoids this here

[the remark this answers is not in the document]
hokolomopo: Shouldn't a scientific _format_, or any decimal format, be stored inside `cell.formattedValue` ? I don't think it makes much sense to keep a formatted value inside `cell.content`

src/helpers/format.ts
laa-odoo: I understand,

In this particular case, there will be a lot of digits on the integer part. 
The decimal part will therefore be significantly not important.

Maybe keep the simple rule of a maximum number of digits by default, in all cases :)

hokolomopo: Changed to not display decimals if we have no space for the decimal separator + a decimal digit.

tests/helpers/format.test.ts
```diff
@@ -19,10 +18,34 @@ describe("formatValue on number", () => {
...
+    expect(formatValue(-0.000000001)).toBe("-0.000000001");
+    expect(formatValue(-0.0000000001)).toBe("-0");
+    expect(formatValue(-0.00000000001)).toBe("-0");
+    expect(formatValue(-0.000000000001)).toBe("-0");
+
+    expect(formatValue(10.123456789)).toBe("10.123456789");
+    expect(formatValue(100.123456789)).toBe("100.12345679");
+    expect(formatValue(1000.123456789)).toBe("1000.1234568");
+    expect(formatValue(10000.123456789)).toBe("10000.123457");
+    expect(formatValue(-10000.123456789)).toBe("-10000.12346");
```
laa-odoo, line 39: I wonder if we really want a difference in the number of digits between the positive case and the negative case.

hokolomopo: My reasoning was "display as much digits as possible, while keeping the number less than 13 characters so it fit nicely in a default cell width with default font size". So there it make sense to display one mroe digit if we don't have the minus sign.

But we could display always one less digit, leaving a space for the minus sign. I don't have a strong opinion on this.

src/helpers/cells/cell_types.ts
```diff
@@ -120,7 +120,15 @@ export class NumberCell extends AbstractCell<NumberEvaluation> {
     if (this.format?.includes("%")) {
       return `${this.evaluated.value * 100}%`;
     }
-    return super.composerContent;
+    return this.numberToString(this.evaluated.value);
+  }
+
+  /** Convert a number into a string, without scientific notation */
+  private numberToString(number: number): string {
```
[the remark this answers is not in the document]
hokolomopo: + used `this.numberToString` on the percentage too

tests/helpers/format.test.ts
hokolomopo: Removed a digit to always leave a space for the minus sign

pro-odoo (approved):
robodoo r+ rebase-ff

pro-odoo:
robodoo r-

src/helpers/cells/cell_types.ts
```diff
@@ -111,7 +111,7 @@ export class EmptyCell extends AbstractCell<EmptyEvaluation> {
 }
 
 export class NumberCell extends AbstractCell<NumberEvaluation> {
-  readonly content = formatValue(this.evaluated.value);
+  readonly content = this.numberToString(this.evaluated.value);
```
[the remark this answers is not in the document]
laa-odoo: Not sure what I'm saying,
However, do this change will bring us closer to the behavior of Google sheet:
if the content is given with a percentage, clear he format will not change the content. (this is not the case in master) 
--> discuss it with the PO

rrahir:
fix in master ?

laa-odoo:
> fix in master ?

IMO not necessarily

laa-odoo:
@robodoo r+

PR #1739, opened 2022-10-19
[FIX] composer: when changing sheet

hokolomopo:
# [FIX] composer: keep focus after changing sheet

Currently the composer lose focus when changing sheet. This is because
of the `useEffect()` in grid, that focus the grid when changing sheet.

Change it to only focus the grid when the grid was focused before
changing sheet.

# [FIX] composer: close text composer when changing sheet

The composer currently stays open when switching to another sheet.
This is useful when editing a formula, to allow the user to select a range
in another sheet, but perfectly useless when editing a text cell.

Odoo task ID : [2974535](https://www.odoo.com/web#id=2974535&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/grid/grid.ts
+++ b/src/components/grid/grid.ts
@@ -267,7 +267,10 @@ export class Grid extends Component<Props, SpreadsheetChildEnv> {
   };
 
   focus() {
-    if (!this.env.model.getters.getSelectedFigureId()) {
+    if (
+      !this.env.model.getters.getSelectedFigureId() &&
+      this.env.model.getters.getEditionMode() === "inactive"
+    ) {
       this.hiddenInput.el!.focus();
     }
   }
--- a/src/plugins/ui/edition.ts
+++ b/src/plugins/ui/edition.ts
@@ -124,7 +124,7 @@ export class EditionPlugin extends UIPlugin {
         break;
       case "STOP_EDITION":
         if (cmd.cancel) {
-          this.cancelEdition();
+          this.cancelEditionAndActivateSheet();
           this.resetContent();
         } else {
           this.stopEdition();
@@ -139,7 +139,7 @@ export class EditionPlugin extends UIPlugin {
         this.replaceSelection(cmd.text);
         break;
       case "SELECT_FIGURE":
-        this.cancelEdition();
+        this.cancelEditionAndActivateSheet();
         this.resetContent();
         break;
       case "ADD_COLUMNS_ROWS":
@@ -186,6 +186,10 @@ export class EditionPlugin extends UIPlugin {
         this.selectionEnd = this.currentContent.length;
         break;
       case "ACTIVATE_SHEET":
+        if (!this.currentContent.startsWith("=")) {
+          this.cancelEdition();
+          this.resetContent();
+        }
         if (cmd.sheetIdFrom !== cmd.sheetIdTo) {
           const { col, row } = this.getters.getNextVisibleCellPosition(cmd.sheetIdTo, 0, 0);
           const zone = this.getters.expandZone(cmd.sheetIdTo, positionToZone({ col, row }));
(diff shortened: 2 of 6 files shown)


rrahir:
this should target 16.0 and you have a small conflit apparently

src/plugins/ui/edition.ts
```diff
@@ -442,14 +446,14 @@ export class EditionPlugin extends UIPlugin {
...
-  private cancelEdition() {
+  private cancelEdition(activateEditionSheet = true) {
     if (this.mode === "inactive") {
       return;
     }
     this.mode = "inactive";
     this.selection.release(this);
     const sheetId = this.getters.getActiveSheetId();
-    if (sheetId !== this.sheetId) {
+    if (activateEditionSheet && sheetId !== this.sheetId) {
```
rrahir, line 456: I think it calls for a change of structure. Adding this parameter just to avoid an infinite loop is not that sturdy

hokolomopo: Created function `cancelEdition()` and `cancelEditionAndActivateSheet()`

src/plugins/ui/edition.ts
```diff
@@ -198,7 +202,7 @@ export class EditionPlugin extends UIPlugin {
         const sheetIdExists = !!this.getters.tryGetSheet(this.sheetId);
         if (!sheetIdExists && this.mode !== "inactive") {
           this.sheetId = this.getters.getActiveSheetId();
-          this.cancelEdition();
+          this.cancelEditionAndActivateSheet();
```
rrahir, line 205: this line  was shady already.  we first set this.sheetId, which means the activateSheet part will never occur...

Seems louche to me to set this.sheetId even though you're not editing anything? I think that cancelling should probably reset the value of this.sheetId as well. That being said, looks more like a refactoring than a fix. I wll add this in the llst of the stuff we need to address when refactoring the composer. The whole internal state is a mess.

src/components/grid/grid.ts
```diff
@@ -113,10 +113,6 @@ export class Grid extends Component<Props, SpreadsheetChildEnv> {
     useGridDrawing("canvas", this.env.model, () =>
       this.env.model.getters.getSheetViewDimensionWithHeaders()
     );
-    useEffect(
-      () => this.focus(),
-      () => [this.env.model.getters.getActiveSheetId()]
```
rrahir, line 118: removing this unfortunately does not work properly. most of the time it will work because a click on the bottom bar will bubble up to Spreadsheet component which will focus the grid but if the said click event is stopped at some point, then you are screwed.
Example of a broken behaviour now,
go to the bottom bar and duplicate the sheet
-> now the activeElement is body, grid no longer has the focus.

Might be worth adding a test for this

src/plugins/ui/edition.ts
```diff
@@ -442,12 +446,8 @@ export class EditionPlugin extends UIPlugin {
...
   }
 
-  private cancelEdition() {
-    if (this.mode === "inactive") {
-      return;
-    }
-    this.mode = "inactive";
-    this.selection.release(this);
+  private cancelEditionAndActivateSheet() {
+    this.cancelEdition();
```
rrahir, line 450: when the mode is inactive, this function should also return early. Again a inssue of design ...
I think taht's the main reason you have failing tests ;-)

rrahir (commented):
Tests are failing  :cry:

src/components/grid/grid.ts
rrahir: the focus function already has an exception if we are selecting a figure, you could add one if the edition mode is not `inactive`for example

hokolomopo: Note that every other option in the bototm bar (except duplicated sheet) doesn't give the focus back to the grid in master.  Added it in https://www.odoo.com/web#id=3094579&cids=1&menu_id=4720&action=333&active_id=2328&model=project.task&view_type=form

src/components/grid/grid.ts
```diff
@@ -267,7 +267,10 @@ export class Grid extends Component<Props, SpreadsheetChildEnv> {
   };
 
   focus() {
-    if (!this.env.model.getters.getSelectedFigureId()) {
+    if (
+      !this.env.model.getters.getSelectedFigureId() &&
+      this.env.model.getters.getEditionMode() === "inactive"
```
LucasLefevre, line 272: focus is :nauseated_face:

LucasLefevre (approved):
robodoo r+

LucasLefevre:
robodoo rebase-ff

PR #1755, opened 2022-10-26
[IMP] figure drag and drop: freeze pane and scrolling

hokolomopo:
Bassed on #1737 because rebase
## [IMP] figure: drag & drop and freeze panes

The current behaviour of figures with freeze panes seemed a bit arbitrary.
The figure would go under or over the freeze pane based on the position
of the mouse. This lead to some strange behaviours depending on the position
of the mouse at the start of the D&D, and the case where the figure start
on the frozen pane with the mouse outside of the pane wasn't handled at
all.

Improve this behaviour by making the figure always go over frozen panes
during d&d. This avoid all the issues of the figure blinking above or
below frozen panes.

## [IMP] figure: drag & drop and freeze panes

Previously scrolling during a figure d&d wasn't working, the grid scrolled
but the figure didn't move.

Now the figure will scroll alongside with the grid during a d&d. Scrolling
during the resizing of the figure will stop the resize.

## [FIX] scrollbar: wrong scroll value in onScroll

The scrollbars were using `useEffect()` to update their scroll to the
scroll value of the sheet. This caused an issue when scrolling fast,
sometime the `onScroll` was called before the `useEffect`, and thus
used incorrect scroll values.

Fixed this by replacing the `useEffect` by a `onWillUpdateProps`, that is
always called before the `onScroll`.

Odoo task ID : [3035531](https://www.odoo.com/web#id=3035531&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/figures/figure/figure.ts
+++ b/src/components/figures/figure/figure.ts
@@ -1,7 +1,8 @@
-import { Component, useEffect, useRef, useState } from "@odoo/owl";
+import { Component, useEffect, useRef } from "@odoo/owl";
 import {
   ComponentsImportance,
   FIGURE_BORDER_COLOR,
+  FIGURE_BORDER_WIDTH,
   SELECTION_BORDER_COLOR,
 } from "../../../constants";
 import { figureRegistry } from "../../../registries/index";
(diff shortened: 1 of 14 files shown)


hokolomopo:
Replaced by #2005

PR #1780, opened 2022-11-02
[FIX] model: do not replay core commands in ui plugins

pro-odoo:
This fix is a bit tricky, let's try to resume the situation:

In a collaborative context [1], commands are applied locally and then sent to the server. This is done on purpose in order to not block the user each time he performs an operation.
The server is responsible to order the operations it received. When at least two users perform concurrent operations, the operations are applied locally and then ordered par the server. After that, we have to ensure that in both clients the operations are applied in the order defined by the server. For the user who have applied it's operation locally and who received an operation that should be applied **before** its own operation, we have to revert the state to the state before its operation, then apply the received operation and then re-apply its operation (with a transformation if needed). The re-apply of the operation should be done **only** on core plugins.

Before this revision, if a command to replay itself dispatch another command, this command was also dispatched in the ui plugins.

With this revision, a such command is now only dispatched on core plugins.

[1] For more details on how the collaborative works, please see the documentation.

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/model.ts
+++ b/src/model.ts
@@ -104,6 +104,12 @@ export class Model extends owl.core.EventBus implements CommandDispatcher {
   private session: Session;
 
   /**
+   * In a collaborative context, some commands can be replayed, we have to ensure
+   * that these commands are not replayed on the UI plugins.
+   */
+  private isReplayingCommand: boolean = false;
+
+  /**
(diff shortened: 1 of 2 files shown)


rrahir:
@robodoo r+

PR #1817, opened 2022-11-17
[REF] models: separate ui plugins in three concepts

pro-odoo:
Before this commit, plugins were split in two: core and ui, based on whether a state should be shared in collaborative.

With this commit, we split the ui plugins in three categories:
- Plugins which have a state, but which should not be shared in collaborative (edition, selection) => stateful_ui_plugins
- Plugins which have a derived state from core data (evaluation) => core_views_plugins
- Plugins which handle a specific feature, without handling any core commands (autofill, sort) => feature_plugins

For now, the only difference is that a command replayed will be dispatched to core view plugins (typically to ensure the evaluation is correct even after a command replay).

But in the future, we should give the history to stateful ui plugins, in order to save the selection or the active sheet when a core command is dispatched.

Part of task-id 2813749

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/model.ts
```diff
@@ -298,7 +317,10 @@ export class Model extends EventBus<any> implements CommandDispatcher {
         this.state.recordChanges.bind(this.state),
         (command: CoreCommand) => {
           this.isReplayingCommand = true;
-          this.dispatchToHandlers([this.range, ...this.corePlugins], command);
+          this.dispatchToHandlers(
+            [this.range, ...this.corePlugins, ...this.coreViewsPlugins],
```
LucasLefevre, line 321: this can be tested:
A core command dispatching an `UPDATE_CELL` command should invalidate evaluation

LucasLefevre (approved):

src/model.ts
pro-odoo: Done: https://github.com/odoo/o-spreadsheet/pull/1817/files#diff-cf5735e699ad89c54032c224139a1cca3f60889cd5d5d30381d59043e3f47a63R201

src/plugins/index.ts
```diff
@@ -46,26 +50,30 @@ export const corePluginRegistry = new Registry<CorePluginConstructor>()
...
   .add("autofill", AutofillPlugin)
   .add("find_and_replace", FindAndReplacePlugin)
   .add("sort", SortPlugin)
   .add("automatic_sum", AutomaticSumPlugin)
   .add("format", FormatPlugin)
   .add("cell_popovers", CellPopoverPlugin)
-  .add("selection_multiuser", SelectionMultiUserPlugin)
+  .add("selection_multiuser", SelectionMultiUserPlugin);
+
+export const statefulUIPluginRegistry = new Registry<UIPluginConstructor>()
```
LucasLefevre, line 68: faut updater la doc :/

rrahir:
@robodoo r+

PR #1828, opened 2022-11-23
[FIX] InternalViewport: compute the actual visible rect of a zone

rrahir:
The function `getRect` would compute the rect of a zone only based on
its own zone without considering that some of its cells are not completely
visible in the grid.

Task 3079048

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/internal_viewport.ts
+++ b/src/helpers/internal_viewport.ts
@@ -203,18 +203,34 @@ export class InternalViewport {
     this.adjustViewportZoneY();
   }
 
+  /**
+   *
+   * @param zone
+   * @returns Computes the absolute coordinate of a given zone inside the viewport
+   */
   getRect(zone: Zone): Rect | undefined {
     const targetZone = intersection(zone, this.zone);
     if (targetZone) {
+      const x =
+        this.getters.getColRowOffset("COL", this.zone.left, targetZone.left) +
+        this.offsetCorrectionX;
+
+      const y =
+        this.getters.getColRowOffset("ROW", this.zone.top, targetZone.top) + this.offsetCorrectionY;
+
+      const width = Math.min(
+        this.getters.getColRowOffset("COL", targetZone.left, targetZone.right + 1),
+        this.width
+      );
+      const height = Math.min(
+        this.getters.getColRowOffset("ROW", targetZone.top, targetZone.bottom + 1),
+        this.height
+      );
       return {
-        x:
-          this.getters.getColRowOffset("COL", this.zone.left, targetZone.left) +
-          this.offsetCorrectionX,
-        y:
-          this.getters.getColRowOffset("ROW", this.zone.top, targetZone.top) +
-          this.offsetCorrectionY,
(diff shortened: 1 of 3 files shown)


src/helpers/internal_viewport.ts
```diff
@@ -203,18 +203,34 @@ export class InternalViewport {
...
+        this.getters.getColRowOffset("ROW", this.zone.top, targetZone.top) + this.offsetCorrectionY;
+
+      const width = Math.min(
+        this.getters.getColRowOffset("COL", targetZone.left, targetZone.right + 1),
+        this.width
+      );
+      const height = Math.min(
+        this.getters.getColRowOffset("ROW", targetZone.top, targetZone.bottom + 1),
+        this.height
+      );
```
hokolomopo, line 228: Any chance you know what the "+1" are for ? Maybe add a comment to explain if you can

hokolomopo (commented):
Nice :+1: 

There's a typo in the commit name (`comute`) + I don't feel like the commit message is very english ? "compute the rect of a zone only based on
without considering"

src/helpers/internal_viewport.ts
rrahir: the +1 have always been here, its because the boundary considers the start of a header and not the end. Don't think it deserves a specific comment. Same goes when you want to generate a range in python for example

rrahir:
the commit message lackes some words indeed

src/helpers/internal_viewport.ts
hokolomopo: Oh yeah it's a col/row position, not pixels. My bad

src/helpers/internal_viewport.ts
```diff
@@ -203,18 +203,34 @@ export class InternalViewport {
...
+      const x =
+        this.getters.getColRowOffset("COL", this.zone.left, targetZone.left) +
+        this.offsetCorrectionX;
+
+      const y =
+        this.getters.getColRowOffset("ROW", this.zone.top, targetZone.top) + this.offsetCorrectionY;
+
+      const width = Math.min(
+        this.getters.getColRowOffset("COL", targetZone.left, targetZone.right + 1),
+        this.width
```
LucasLefevre, line 223: Does it work with freezed panes ?
`width` wouldn't take into account freezed cols and it would be smaller than the col offset ?
I think I'm missing something

LucasLefevre: Actually, `getColRowOffset` should be named `getHeadersDistance` or `getDistanceBetweenHeaders`...

rrahir: This is a pane aka a viewport. A frozen pane will only imply the presence of more than one viewport . Each of them computes the size of a selected zone inside their own scope (hence the clipping with its own width) and each size is then assembled together in the getter getVisibleRect which is inside the plugin SheetView.

LucasLefevre (approved):
robodoo r+

PR #1841, opened 2022-11-30
[IMP] chart: make size and position required

yunchen4:
## Description:

When creating a new chart, the size and position should be required.

Odoo task ID : [2896073](https://www.odoo.com/web#id=2896073&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/tests/test_helpers/commands_helpers.ts
+++ b/tests/test_helpers/commands_helpers.ts
@@ -26,8 +26,8 @@ import { ScorecardChartDefinition } from "../../src/types/chart/scorecard_chart"
 import { SelectionDirection, SelectionStep } from "../../src/types/selection";
 import { target } from "./helpers";
 
-const size = { width: DEFAULT_FIGURE_WIDTH, height: DEFAULT_FIGURE_HEIGHT };
-const position = {
+const DEFAULT_FIGURE_SIZE = { width: DEFAULT_FIGURE_WIDTH, height: DEFAULT_FIGURE_HEIGHT };
+const DEFAULT_FIGURE_POSITION = {
   x: 0,
   y: 0,
 };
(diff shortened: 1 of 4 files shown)


tests/test_helpers/commands_helpers.ts
```diff
@@ -22,6 +26,12 @@ import { ScorecardChartDefinition } from "../../src/types/chart/scorecard_chart"
 import { SelectionDirection, SelectionStep } from "../../src/types/selection";
 import { target } from "./helpers";
 
+const size = { width: DEFAULT_FIGURE_WIDTH, height: DEFAULT_FIGURE_HEIGHT };
+const position = {
+  x: 0,
+  y: 0,
+};
```
hokolomopo, line 33: Try to use more specific variables names, something like `defaultFigureSize/position`. Otherwise its a bit confusing, even more so since this file contains lots of different stuff.

hokolomopo (commented):
Looking good :smiley: I have only something to say on a small detail.

Don't forget to link this pr + put the branch name in the pad Odoo, so it can more easily be found.

LucasLefevre (commented):
Actually, I'm not sure about the idea of task itself.
I don't know the real benefit and it's more annoying.
With this commit, we need:
- a migration
- to export `DEFAULT_FIGURE_HEIGHT` and `DEFAULT_FIGURE_WIDTH`, import it in odoo, use them to dispatch. Or hardcode a size... :thinking: 

@pro-odoo what was your idea exactly?

pro-odoo:
> @pro-odoo what was your idea exactly?

As the default value is not used in o-spreadsheet (we always set the size in commands), I think it's cleaner to ensure that the size is given. 
But it's true that we need a migration so I'm ok to cancel this task if you want :)

PR #1843, opened 2022-12-01
[spreadsheet] paste copied sheet range as table

mamu-odoo:
## Description:
In this commit, we add an 'text/html' part in the os clipboard when copying
cells. This allow to paste the cells as a table in an external application.

Also refractor a bit how we handle the clipboard : now we put the
navigator.clipboard in a ClipboardWrapper class, which hide the
implementation details and error handling of the navigator's clipboard.

Odoo task ID : [2832383](https://www.odoo.com/web#id=2832383&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/grid/grid.ts
+++ b/src/components/grid/grid.ts
@@ -21,9 +21,11 @@ import { cellMenuRegistry } from "../../registries/menus/cell_menu_registry";
 import { colMenuRegistry } from "../../registries/menus/col_menu_registry";
 import { INSERT_LINK } from "../../registries/menus/menu_items_actions";
 import { rowMenuRegistry } from "../../registries/menus/row_menu_registry";
+import { _lt } from "../../translation";
 import {
   CellValueType,
   Client,
+  ClipboardMIMEType,
   DOMCoordinates,
   DOMDimension,
   HeaderIndex,
@@ -524,6 +526,13 @@ export class Grid extends Component<Props, SpreadsheetChildEnv> {
     if (!this.gridEl.contains(document.activeElement)) {
       return;
     }
+
+    const clipboardData = ev.clipboardData;
+    if (!clipboardData) {
+      this.displayWarningCopyPasteNotSupported();
+      return;
+    }
+
     /* If we are currently editing a cell, let the default behavior */
     if (this.env.model.getters.getEditionMode() !== "inactive") {
       return;
@@ -534,7 +543,9 @@ export class Grid extends Component<Props, SpreadsheetChildEnv> {
       this.env.model.dispatch("COPY");
     }
     const content = this.env.model.getters.getClipboardContent();
-    ev.clipboardData!.setData("text/plain", content);
+    for (const type in content) {
+      clipboardData.setData(type, content[type]);
+    }
     ev.preventDefault();
   }
 
@@ -542,18 +553,28 @@ export class Grid extends Component<Props, SpreadsheetChildEnv> {
     if (!this.gridEl.contains(document.activeElement)) {
       return;
     }
-    const clipboardData = ev.clipboardData!;
-    if (clipboardData.types.indexOf("text/plain") > -1) {
-      const content = clipboardData.getData("text/plain");
+
+    const clipboardData = ev.clipboardData;
+    if (!clipboardData) {
+      this.displayWarningCopyPasteNotSupported();
+      return;
+    }
+
+    if (clipboardData.types.indexOf(ClipboardMIMEType.PlainText) > -1) {
+      const content = clipboardData.getData(ClipboardMIMEType.PlainText);
       const target = this.env.model.getters.getSelectedZones();
-      const clipBoardString = this.env.model.getters.getClipboardContent();
-      if (clipBoardString === content) {
+      const clipboardString = this.env.model.getters.getClipboardTextContent();
+      if (clipboardString === content) {
         // the paste actually comes from o-spreadsheet itself
         interactivePaste(this.env, target);
       } else {
         interactivePasteFromOS(this.env, target, content);
       }
     }
+  }
+
+  private displayWarningCopyPasteNotSupported() {
+    this.env.raiseError(_lt("Copy/Paste is not supported in this browser."));
   }
 
   closeMenu() {
--- a/src/components/helpers/css.ts
+++ b/src/components/helpers/css.ts
@@ -98,6 +98,20 @@ export function getTextDecoration({
 }
 
 /**
+ * Convert the cell style to CSS properties.
+ */
+export function cellStyleToCss(style: Style | undefined): CSSProperties {
+  const attributes = cellTextStyleToCss(style);
+  if (!style) return attributes;
+
+  if (style.fillColor) {
+    attributes["background"] = style.fillColor;
+  }
+
+  return attributes;
+}
+
+/**
  * Convert the cell text style to CSS properties.
  */
 export function cellTextStyleToCss(style: Style | undefined): CSSProperties {
@@ -122,10 +136,11 @@ export function cellTextStyleToCss(style: Style | undefined): CSSProperties {
   return attributes;
 }
 
-export function cssPropertiesToCss(attributes: CSSProperties): string {
+export function cssPropertiesToCss(attributes: CSSProperties, newLine = true): string {
+  const separator = newLine ? "\n" : "";
   const str = Object.entries(attributes)
     .map(([attName, attValue]) => `${attName}: ${attValue};`)
-    .join("\n");
+    .join(separator);
 
-  return "\n" + str + "\n";
+  return str ? "\n" + str + "\n" : "";
 }
--- a/src/helpers/clipboard/clipboard_cells_state.ts
+++ b/src/helpers/clipboard/clipboard_cells_state.ts
@@ -1,3 +1,4 @@
+import { cellStyleToCss, cssPropertiesToCss } from "../../components/helpers";
 import { SELECTION_BORDER_COLOR } from "../../constants";
 import { SelectionStreamProcessor } from "../../selection_stream/selection_stream_processor";
 import {
@@ -13,7 +14,8 @@ import {
   UID,
   Zone,
 } from "../../types";
-import { ClipboardOperation, ClipboardOptions } from "../../types/clipboard";
+import { ClipboardMIMEType, ClipboardOperation, ClipboardOptions } from "../../types/clipboard";
+import { xmlEscape } from "../../xlsx/helpers/xml_helpers";
 import { formatValue } from "../format";
 import { range } from "../misc";
 import { createAdaptedZone, mergeOverlappingZones, positions, union } from "../zones";
@@ -71,6 +73,7 @@ export class ClipboardCellsState extends ClipboardCellsAbstractState {
         const position = { col, row, sheetId };
         cellsInRow.push({
           cell: getters.getCell(position),
+          style: getters.getCellComputedStyle(position),
           evaluatedCell: getters.getEvaluatedCell(position),
           border: getters.getCellBorder(position) || undefined,
           position,
@@ -465,7 +468,14 @@ export class ClipboardCellsState extends ClipboardCellsAbstractState {
     }
   }
 
-  getClipboardContent(): string {
+  getClipboardContent(): Record<string, string> {
+    return {
+      [ClipboardMIMEType.PlainText]: this.getPlainTextContent(),
+      [ClipboardMIMEType.Html]: this.getHTMLContent(),
+    };
+  }
+
+  private getPlainTextContent(): string {
     return (
       this.cells
         .map((cells) => {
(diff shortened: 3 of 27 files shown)


src/types/clipboard.ts
```diff
@@ -19,7 +19,7 @@ export interface ClipboardState {
   isPasteAllowed(target: Zone[], clipboardOption?: ClipboardOptions): CommandResult;
 
   paste(target: Zone[], options?: ClipboardOptions | undefined): void;
-  getClipboardContent(): string;
+  getClipboardContent(): Record<string, string>;
```
hokolomopo, line 22: After you defined an enum for the contents (html or plain) of the clipbaord, you can use it to type this (and all the functios that extend it)

```suggestion
  getClipboardContent(): Record<ClipboardContentType, string>;
```

tests/__mocks__/clipboard_item.ts
```diff
@@ -0,0 +1,21 @@
+class mockClipboardItem implements ClipboardItem {
```
hokolomopo, line 1:
```suggestion
class MockClipboardItem implements ClipboardItem {
```

tests/__mocks__/clipboard_item.ts
```diff
@@ -0,0 +1,21 @@
...
+  readonly types: string[] = [];
+  readonly presentationStyle = "unspecified";
+  private data = {};
+
+  constructor(dict) {
```
hokolomopo, line 6: No clue what the dict is. Try to find a better name, and type it.

tests/__mocks__/clipboard_item.ts
```diff
@@ -0,0 +1,21 @@
...
+    for (const key in dict) {
+      this.types.push(key);
+      this.data[key] = dict[key];
+    }
+  }
+
+  getData(): string {
```
hokolomopo, line 13: Does this really exist in a real ClipboardItem ? I cannot find it in the doc, and this doesn't seems to exist in the browser. I think we're supposed to use `getType() ` to get the content of the clipboard item.

tests/components/context_menu.test.ts
```diff
@@ -1,3 +1,9 @@
+jest.mock("../../src/registries/menus/menu_items_actions", () => {
+  return {
+    ...jest.requireActual("../../src/registries/menus/menu_items_actions"),
+    ...jest.requireActual("../__mocks__/clipboard_item"),
+  };
+});
```
hokolomopo, line 6: I don't think you need to mock the file like this since you don't mock any function of the file.

You can mock the clipboard like we mocked the charts (see `mockChart`). It's easier to read.

tests/components/grid.test.ts
```diff
@@ -1443,6 +1443,27 @@ describe("Copy paste keyboard shortcut", () => {
...
+    document.activeElement!.dispatchEvent(
+      new KeyboardEvent("keydown", { key: "=", altKey: true, bubbles: true })
+    );
+    document.body.dispatchEvent(await getClipboardEvent("copy"));
+    expect(model.getters.getClipboardContent()["text/html"]).toEqual(
+      "<table><tr><td>1</td><td></td></tr><tr><td>3</td><td>2</td></tr></table>"
+    );
+  });
+
+  test("Copied single cells are not in a html table", async () => {
```
hokolomopo, line 1460: These tests should be in `clipboard.test.ts`, and in the grid tests you should check if the clipboard given to the clipboardEvent has beed updated by grid.

tests/menu_items_registry.test.ts
```diff
@@ -143,14 +149,12 @@ describe("Menu Item actions", () => {
...
     doAction(["edit", "copy"], env);
     expect(dispatch).toHaveBeenCalledWith("COPY");
-    expect(env.clipboard.writeText).toHaveBeenCalledWith(env.model.getters.getClipboardContent());
   });
 
   test("Edit -> cut", () => {
     env.clipboard.writeText = jest.fn(() => Promise.resolve());
     doAction(["edit", "cut"], env);
     expect(dispatch).toHaveBeenCalledWith("CUT");
-    expect(env.clipboard.writeText).toHaveBeenCalledWith(env.model.getters.getClipboardContent());
```
hokolomopo, line 153: Don't delete this, check that `env.clipboard.write` has been caleld with the correct value

hokolomopo (commented):
Some randoms comments on details, but this seems good :)

tests/__mocks__/clipboard_item.ts
mamu-odoo: no , I forgot to remove it

src/helpers/clipboard/clipboard_cells_state.ts
```diff
@@ -475,8 +476,22 @@ export class ClipboardCellsState extends ClipboardCellsAbstractState {
             )
             .join("\t");
         })
-        .join("\n") || "\t"
-    );
+        .join("\n") || "\t";
+    if (this.cells.length == 1 && this.cells[0].length == 1) {
```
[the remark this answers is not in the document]
hokolomopo: Note : check if it's useful to have a `getHTMLContent ` and `getPlainTextContent` for every type of clipboard.

src/helpers/clipboard/clipboard_cells_state.ts
```diff
@@ -479,6 +488,24 @@ export class ClipboardCellsState extends ClipboardCellsAbstractState {
...
+      return this.getters.getCellText(this.cells[0][0].position);
+    }
+
+    let htmlTable = '<table border="1" style="border-collapse:collapse">';
+    for (const row of this.cells) {
+      htmlTable += "<tr>";
+      for (const cell of row) {
+        const cssStyle = cssPropertiesToCss(cellTextStyleToCss(cell.style), false);
+        htmlTable += `<td style="${cssStyle}">` + this.getters.getCellText(cell.position) + "</td>";
```
LucasLefevre, line 501: The cell content needs to be escaped in case it contains some html characters (the text `"<div/>"` in a cell should also be in the table)

src/types/clipboard.ts
```diff
@@ -2,6 +2,13 @@ import { CommandResult } from "./commands";
 import { Dimension, HeaderIndex, UID, Zone } from "./misc";
 import { GridRenderingContext } from "./rendering";
 
+export enum ClipboardContentType {
+  PLAIN = "text/plain",
+  HTML = "text/html",
```
LucasLefevre, line 7: I think we are using PascalCase for most enums
```suggestion
  PlainText = "text/plain",
  Html = "text/html",
```

src/types/clipboard.ts
```diff
@@ -2,6 +2,13 @@ import { CommandResult } from "./commands";
 import { Dimension, HeaderIndex, UID, Zone } from "./misc";
 import { GridRenderingContext } from "./rendering";
 
+export enum ClipboardContentType {
```
LucasLefevre, line 5: what about MIMEType ? That's what it is after all
```suggestion
export enum MIMEType {
```

src/types/misc.ts
```diff
@@ -183,6 +183,7 @@ export type FunctionReturnValue = string | number | boolean;
 
 export interface ClipboardCell {
   cell?: Cell;
+  style?: Style;
```
LucasLefevre, line 186: Personal opinion, but I hate when I copy something from discord or vs code, paste it somewhere (e.g. a task pad) and the style is also copied.

The first thing I always do it to fight with the editor to remove the style. Maybe it's less annoying with tables

tests/plugins/clipboard.test.ts
```diff
@@ -436,12 +437,81 @@ describe("clipboard", () => {
...
+  describe("Copied cells HTML", () => {
+    test("Copied group of cells are represented as an HTML table in the clipboard", async () => {
+      const model = new Model();
+      setCellContent(model, "A1", "1");
+      setCellContent(model, "B1", "2");
+      setCellContent(model, "A2", "3");
+      copy(model, "A1:B2");
+      const htmlContent = model.getters.getClipboardContent()[ClipboardContentType.HTML]!;
+      const parser = new DOMParser();
+      const parsedHTML = parser.parseFromString(htmlContent, "text/html");
```
LucasLefevre, line 457: I would actually *not* parse it.
I think it would be more readable if you compare the html string. Maybe with `toMatchInlineSnapshot`
maybe with `import format from "xml-formatter";` to have something better formatted

src/components/helpers/css.ts
```diff
@@ -118,14 +118,18 @@ export function cellTextStyleToCss(style: Style | undefined): CSSProperties {
...
 
-export function cssPropertiesToCss(attributes: CSSProperties): string {
+export function cssPropertiesToCss(attributes: CSSProperties, newLine = true): string {
+  const separator = newLine ? "\n" : "";
   const str = Object.entries(attributes)
     .map(([attName, attValue]) => `${attName}: ${attValue};`)
-    .join("\n");
+    .join(separator);
 
   return "\n" + str + "\n";
```
LucasLefevre, line 134: maybe don't add both "\n" if str is empty. It will result in a better formatted html string

src/registries/menus/menu_items_actions.ts
```diff
@@ -78,41 +68,44 @@ export const REDO_ACTION = (env: SpreadsheetChildEnv) => env.model.dispatch("REQ
...
-  } else {
-    env.model.dispatch("PASTE", {
-      target: env.model.getters.getSelectedZones(),
-      pasteOption: "onlyValue",
-    });
+async function paste(env: SpreadsheetChildEnv, pasteOption?: ClipboardPasteOptions) {
+  const spreadsheetClipboard = env.model.getters.getClipboardTextContent();
+  const osClipboard = await env.clipboard.readText();
+
+  if (osClipboard.status === "ok") {
```
LucasLefevre, line 86: use a switch/case statement

tests/test_helpers/clipboard.ts
```diff
@@ -0,0 +1,56 @@
...
+    return { status: "ok", content: this.content || "" };
+  }
+
+  async writeText(text: string): Promise<void> {
+    this.content = text;
+  }
+
+  async writeContent(content: ClipboardContent) {
+    this.content = content[ClipboardContentType.PLAIN];
+    return Promise.resolve();
```
LucasLefevre, line 20: not required. `async` functions always return a promise

tests/plugins/clipboard.test.ts
hokolomopo: My point with parsing the clipboard content was to make sure that the HTML il actually valid. Since we generate the HTMl by hand, it would be really easy to mess something up (missing closing tag, `style:` instead of `style=`, ...)

But I can change if you feel that the readability of the tests are more important :)

src/types/misc.ts
hokolomopo: I 100% agree with you. That being said, I feel like it make more sense for tables, since we might have different styles for headers/highligted cells ... 
Might be worth to ask around.

src/components/grid/grid.ts
```diff
@@ -484,19 +493,27 @@ export class Grid extends Component<Props, SpreadsheetChildEnv> {
...
+    if (!clipboardData) {
+      this.displayWarningCopyPasteNotSupported();
+      return;
+    }
+
+    if (clipboardData.types.indexOf(ClipboardContentType.PLAIN) > -1) {
+      const content = clipboardData.getData(ClipboardContentType.PLAIN);
       const target = this.env.model.getters.getSelectedZones();
-      const clipBoardString = this.env.model.getters.getClipboardContent();
+      const clipBoardString = this.env.model.getters.getClipboardTextContent();
```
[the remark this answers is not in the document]
hokolomopo: :mortar_board:

src/types/clipboard.ts
hokolomopo: `ClipboardMIMEType` then, because I want the enum to be lilited to supported MIME types for the clipboard, not any MIME type.

tests/plugins/clipboard.test.ts
LucasLefevre: You've got a point.
We could have both :wink: 
Compare the raw string for readability and parse it and assert there's no error

hokolomopo: I kept parsing for a test, and tested directly the html string for the others. The test do a bit more than testing that we can parse as html, because the DOMParser will manage to parse about anything as html, and "fix" the errors as it goes.

hokolomopo:
@LucasLefevre fixed the tests

src/helpers/clipboard/navigator_clipboard_wrapper.ts
```diff
@@ -0,0 +1,65 @@
...
+  | { status: "permissionDenied" | "notImplemented" };
+
+export interface ClipboardWrapperInterface {
+  writeContent(clipboardContent: ClipboardContent): Promise<void>;
+  writeText(text: string): Promise<void>;
+  readText(): Promise<ClipboardReadResult>;
+  clearClipboard(): Promise<void>;
+}
+
+export class ClipboardWrapper implements ClipboardWrapperInterface {
```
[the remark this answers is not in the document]
LucasLefevre: Actually, I changed my mind. We don't care about the name since the world don't/shouldn't care about the implementation.
We could even go a step further to hide the implementation and have a factory function which would instantiate this class and return it, typed as the interface. The implementation wouldn't need to be exposed at all.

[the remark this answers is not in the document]
hokolomopo: Why not. Now `spreadsheet.ts` only call `instantiateClipboard()`, which instanciate a `WebClipboardWrapper` and returns a `ClipboardInterface`.

src/types/misc.ts
rrahir: I agree that it's a pita but in the end, it can still be useful in some cases. Mamu made the good repark that trl Shift v actually pastes the text/plain and not the text/html content, don't recall where he foundthe invormation but it looks damn true

src/helpers/clipboard/navigator_clipboard_wrapper.ts
```diff
@@ -0,0 +1,69 @@
...
+  | { status: "permissionDenied" | "notImplemented" };
+
+export interface ClipboardInterface {
+  write(clipboardContent: ClipboardContent): Promise<void>;
+  writeText(text: string): Promise<void>;
+  readText(): Promise<ClipboardReadResult>;
+  clear(): Promise<void>;
+}
+
+export function instantiateClipboard(): ClipboardInterface {
```
LucasLefevre, line 14: :rocket:

LucasLefevre (approved):
z'est pardyyy
robodoo r+

LucasLefevre:
robodoo r-

LucasLefevre:
you can add yourself as co-author of the commit :wink:

LucasLefevre (commented):
robodoo r+

PR #1844, opened 2022-12-01
[IMP] Error: Lookup module's functions now return NA Errors

anhe-odoo:
## Task Description

When using a function of the LOOKUP module, whe should get a N/A error when the value we are looking for doesn't exist in the specified range. To be able to get error that make sense anyway (as the NA error is silent nowadays), we then added the possibility to put a custom error message when throwing a NA error. If a custom message is added, the error will have the classical error level, while it will have the silent error level in the other case.

## Related task
Odoo task ID : [3067622](https://www.odoo.com/web#id=3067622&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/functions/module_lookup.ts
+++ b/src/functions/module_lookup.ts
@@ -6,6 +6,7 @@ import {
   MatrixArgValue,
   PrimitiveArgValue,
 } from "../types";
+import { NotAvailableError } from "../types/errors";
 import { args } from "./arguments";
 import {
   assert,
@@ -22,6 +23,14 @@ import {
 const DEFAULT_IS_SORTED = true;
 const DEFAULT_MATCH_MODE = 0;
 const DEFAULT_SEARCH_MODE = 1;
+
+function assertAvailable(variable, searchKey) {
+  if (variable === undefined) {
+    throw new NotAvailableError(
+      _lt("Did not find value '%s' in [[FUNCTION_NAME]] evaluation.", toString(searchKey))
+    );
+  }
+}
 
 // -----------------------------------------------------------------------------
 // COLUMN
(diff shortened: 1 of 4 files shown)


LucasLefevre:
robodoo delegate=laa-odoo

src/functions/module_lookup.ts
```diff
@@ -184,17 +187,22 @@ export const LOOKUP: AddFunctionDescription = {
...
     }
 
-    assert(
-      () => index <= nbRow - 1,
-      _lt("[[FUNCTION_NAME]] evaluates to an out of range column value %s.", (index + 1).toString())
-    );
+    if (index >= nbRow) {
+      throw new NotAvailableError(
+        _lt(
+          "[[FUNCTION_NAME]] evaluates to an out of range column value %s.",
```
laa-odoo, line 201: maybe forgot to change the message here because the previous error message (```(index >= nbCol)```) changes

src/functions/module_lookup.ts
```diff
@@ -446,10 +456,11 @@ export const XLOOKUP: AddFunctionDescription = {
     }
 
     const _defaultValue = defaultValue?.();
-    assert(
-      () => !!_defaultValue,
-      _lt("Did not find value '%s' in [[FUNCTION_NAME]] evaluation.", toString(searchKey))
-    );
+    if (!_defaultValue) {
+      throw new NotAvailableError(
```
laa-odoo, line 460: To avoid repetition and maintain consistency

maybe it would be good to create a new helper assert function,

ex ```assertAvailable ``` in which to pass only the condition

laa-odoo (changes requested):
Apart from small comments, nothing to say :)

laa-odoo:
Can be distinguished in the name of the pr/commit, that the change applies to the whole module and not just the lookup function

laa-odoo (approved):

laa-odoo:
robodo r+

laa-odoo:
robodoo r+

PR #1847, opened 2022-12-02
[IMP] Long centered aligned cell contents overlapping with end non-empty cells

yunchen4:
## Description:

Now when a cell with long contents (overflow) is centered aligned and the cell which is located at the end of the overflowing contents also has contents (not empty), both contents will overlap, which causes annoying visual results. This bug only occurs when only one end cell has contents and the other doesn't (when both end cells have contents it seems good). 

The root cause is that when the available space (rect width) is equal to or more than the text width, it is not guaranteed that both the left half and the right half of the contents are perfectly fitted into the empty cells. Hence the fix includes separate judgement for the left and right half of the available space and text width. 

Odoo task ID : [3088107](https://www.odoo.com/web#id=3088107&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_feature/renderer.ts
+++ b/src/plugins/ui_feature/renderer.ts
@@ -664,17 +664,20 @@ export class RendererPlugin extends UIPlugin {
         case "center": {
           const emptyZone = {
             ...zone,
-            right: nextColIndex,
             left: previousColIndex,
+            right: nextColIndex,
           };
-          const { x, y, width, height } = this.getters.getVisibleRect(emptyZone);
+          const { x, y, height, width } = this.getters.getVisibleRect(emptyZone);
+          const halfContentWidth = contentWidth / 2;
+          const boxMiddle = box.x + box.width / 2;
           if (
-            width < contentWidth ||
-            previousColIndex === col ||
-            nextColIndex === col ||
+            x + width < boxMiddle + halfContentWidth ||
+            x > boxMiddle - halfContentWidth ||
             fontSizePX > height
           ) {
-            box.clipRect = { x, y, width, height };
+            const clipX = x > boxMiddle - halfContentWidth ? x : boxMiddle - halfContentWidth;
+            const clipWidth = x + width - clipX;
+            box.clipRect = { x: clipX, y, width: clipWidth, height };
           }
           break;
         }
--- a/tests/plugins/renderer.test.ts
+++ b/tests/plugins/renderer.test.ts
@@ -946,55 +946,64 @@ describe("renderer", () => {
     });
   });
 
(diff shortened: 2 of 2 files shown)


src/plugins/ui_feature/renderer.ts
```diff
@@ -647,19 +647,30 @@ export class RendererPlugin extends UIPlugin {
...
+            right: nextColIndex,
+          };
+          const {
+            x,
+            y,
+            height,
+            width: leftWidth,
+          } = this.getters.getVisibleRect(emptyZoneOnTheLeft);
+          const { width: rightWidth } = this.getters.getVisibleRect(emptyZoneOnTheRight);
+          const halfTextWidth = textWidth / 2;
```
rrahir, line 665: the fix looks good. I wonder if we could rely on a single rect to achieve the same result though. Something like 
```javascript
const emptyZone = {
  ...zone,
  right: nextColIndex,
  left: previousColIndex,
};
const { x, y, width, height } = this.getters.getVisibleRect(emptyZone);
const boxMiddle = box.x + box.width / 2;
if (
  x + width < boxMiddle + contentWidth / 2 ||
  x > boxMiddle - contentWidth / 2 ||
  fontSizePX > height
) {
  box.clipRect = { x, y, width, height };
}
```
Which should to the same thing. I find it a bit confusing to work with leftWidth and rightWidth, I had to write it down to know what the code was actually doing. My suggestion (imo , super ubjective, of course) looks more readable? What doyou think ?

Also the names `emptyZoneOnTheRight/Left`  don't seem clear to me :/

src/plugins/ui_feature/renderer.ts
```diff
@@ -647,19 +647,30 @@ export class RendererPlugin extends UIPlugin {
...
           if (
-            width < contentWidth ||
+            leftWidth - box.width / 2 < halfTextWidth ||
+            rightWidth - box.width / 2 < halfTextWidth ||
             previousColIndex === col ||
             nextColIndex === col ||
```
rrahir, line 670: I think these 2 condifitions `... === col` can be removed, they will be englobed in the other conditions you added :)

rrahir (changes requested):
Apart from my comments, could you edit the commit title to match the [guidelines](https://www.odoo.com/documentation/16.0/contributing/development/coding_guidelines.html)? something like `[FIX] renderer: ...`

rrahir:
Oh also, it would require a test as well!

rrahir:
@robodoo r+

PR #1853, opened 2022-12-04
[FIX] figure: Change component structure to leverage navigator native behaviour

rrahir:
## Description:

Up until now, we'd have to run some nauseating computations to display a
figure according to their position in the spreadsheet (both pane and
scroll of the panes were used). This commit introduc a new html
architecture where a figure is enclosed in a container that hide any
overflowing part. The figuer will then be translated inside that
container, effectively being hidden when positioned outside the visible
scrope of the wrapper.

This approach allows us to get rid of some of the more tedious
computations to determine which part of a figure should be visible as
well as offering the possibility to remove most of the borders logic as
well. All in all, simpler code.

In the long run, it'd be great to have the figure component completely
agnostic of its container and to have its parent (figure container) in
charge of handling the frozen pane/scroll.

Odoo task ID : [3103586](https://www.odoo.com/web#id=3103586&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/figures/figure/figure.ts
+++ b/src/components/figures/figure/figure.ts
@@ -11,7 +11,7 @@ import { css } from "../../helpers/css";
 import { gridOverlayPosition } from "../../helpers/dom_helpers";
 import { startDnd } from "../../helpers/drag_and_drop";
 
-type Anchor =
+type ResizeAnchor =
   | "top left"
   | "top"
   | "top right"
@@ -30,19 +30,14 @@ const ACTIVE_BORDER_WIDTH = 2;
 
 css/*SCSS*/ `
   div.o-figure {
-    box-sizing: content-box;
+    box-sizing: border-box;
     position: absolute;
     width: 100%;
     height: 100%;
 
-    bottom: 0px;
-    right: 0px;
     border: solid ${FIGURE_BORDER_COLOR};
     &:focus {
       outline: none;
-    }
-    &.active {
-      border: solid ${SELECTION_BORDER_COLOR};
     }
 
     &.o-dragging {
@@ -51,17 +46,17 @@ css/*SCSS*/ `
     }
   }
 
+  div.o-active-figure-border {
+    box-sizing: border-box;
+    z-index: 1;
+    border: ${ACTIVE_BORDER_WIDTH}px solid ${SELECTION_BORDER_COLOR};
+  }
+
   .o-figure-wrapper {
     position: absolute;
     box-sizing: content-box;
 
-    .o-figure-overflow-wrapper {
-      position: absolute;
-      overflow: hidden;
-      width: 100%;
-      height: 100%;
-    }
-    .o-anchor {
+    .o-fig-resizer {
       z-index: ${ComponentsImportance.ChartAnchor};
       position: absolute;
       width: ${ANCHOR_SIZE}px;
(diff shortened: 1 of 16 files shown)


src/components/figures/figure/figure.xml
```diff
@@ -1,67 +1,75 @@
...
+            <t
+              t-component="figureRegistry.get(this.props.figure.tag).Component"
+              t-key="this.props.figure.id"
+              sidePanelIsOpen="props.sidePanelIsOpen"
+              onFigureDeleted="props.onFigureDeleted"
+              figure="displayedFigure"
+            />
+          </div>
+          <t t-if="isSelected">
+            <div class="w-100 h-100 o-active-figure-border position-absolute pe-none"/>
```
LucasLefevre, line 29: `o-active` is enough, right ? Or do you really want to talk about borders here?
```suggestion
            <div class="w-100 h-100 o-active position-absolute pe-none"/>
```

src/components/figures/figure/figure.xml
```diff
@@ -1,67 +1,75 @@
...
+              t-on-mousedown="(ev) => this.resize(1,-1, ev)"
+            />
+            <div
+              class="o-anchor o-right"
+              t-att-style="this.getAnchorPosition('right')"
+              t-on-mousedown="(ev) => this.resize(1,0, ev)"
+            />
+            <div
+              class="o-anchor o-bottomRight"
+              t-att-style="this.getAnchorPosition('bottom right', info)"
```
LucasLefevre, line 47: `info` is unused and can be removed

LucasLefevre (commented):
This is incredibly simpler than before :heart_eyes: 

I don't fully understand the anchor thing.

Also, the figure content is shifted when you select it. Probably by the border's width.
Also, the bottom and right borders of scorecards aren't displayed. 
scorecard borders

src/components/figures/figure/figure.ts
```diff
@@ -126,123 +129,76 @@ export class FigureComponent extends Component<Props, SpreadsheetChildEnv> {
     return this.env.model.getters.getSelectedFigureId() === this.props.figure.id;
   }
 
-  /** Get the current figure size, which is either the stored figure size of the DnD figure size */
-  private getFigureSize() {
-    const { width, height } = this.displayedFigure;
-    return { width, height };
+  get gontainerStyle(): string {
```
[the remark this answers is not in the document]
rrahir: sounds good doesn't it ?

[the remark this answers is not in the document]
rrahir: ptain il ya 2 containerstyle ...

src/components/figures/figure/figure.xml
```diff
@@ -1,67 +1,75 @@
...
-            sidePanelIsOpen="props.sidePanelIsOpen"
-            onFigureDeleted="props.onFigureDeleted"
-            figure="displayedFigure"
-          />
+    <div
+      class="o-figure-container-new position-absolute pe-none"
+      t-att-class="{'overflow-hidden': !dnd.isActive}"
+      t-att-style="gontainerStyle">
+      <div
+        class="o-figure-anchor w-0 h-0 overflow-visible position-absolute"
```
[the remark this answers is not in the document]
rrahir: it is not. We could probvably rename those anchors with a proper name which matches their use - like `resizer`

The idea of the 0 sized div is to have a "hook" (maybe better than anchor ?) on which we place the actual figure part. It's an invisible point of reference and when we scroll, we move that point of reference.  since it contains the figure, it cannot be invisible, so 0 size^^

src/components/figures/figure/figure.xml
```diff
@@ -1,67 +1,75 @@
...
-          t-on-keyup.stop="">
-          <t
-            t-component="figureRegistry.get(this.props.figure.tag).Component"
-            t-key="this.props.figure.id"
-            sidePanelIsOpen="props.sidePanelIsOpen"
-            onFigureDeleted="props.onFigureDeleted"
-            figure="displayedFigure"
-          />
+    <div
+      class="o-figure-container-new position-absolute pe-none"
```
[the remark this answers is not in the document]
rrahir: I'm pretty sure it's an artefact when I was generating classes all around. I will make a general pass on everything and try to clean this up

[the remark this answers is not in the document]
rrahir: the class being useless at ,i'll drop it

src/components/figures/figure/figure.ts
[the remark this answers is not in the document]
LucasLefevre: That's why "gontainer" existed :upside_down_face:

src/components/figures/figure/figure.ts
```diff
@@ -346,28 +334,31 @@ export class FigureComponent extends Component<Props, SpreadsheetChildEnv> {
     this.dnd.height = figure.height;
 
     const onMouseMove = (ev: MouseEvent) => {
+      const { x: mainViewportX, y: mainViewportY } =
+        this.env.model.getters.getMainViewportCoordinates();
+      const { offsetX: scrollOffsetX, offsetY: scrollOffsetY } =
```
rrahir, line 339: is the name scrolloffset necessary ? SheetScrollInfo also container scrollbaroffsetX, they'r a bit to close imo. but nitpick, really

src/components/figures/figure/figure.ts
```diff
@@ -193,12 +193,13 @@ export class FigureComponent extends Component<Props, SpreadsheetChildEnv> {
       left: ${left}px;
       top: ${top}px;
       width: ${width}px;
-      height: ${height}px
+      height: ${height}px;
+      border: red solid 1px;
```
rrahir, line 197: note, don't forget to remove

src/components/figures/figure/figure.ts
LucasLefevre: I actually had forgotten `SheetScrollInfo` had `offsetScrollbarX` :sweat_smile: 

What I thought was: when I see "offset", I ask myself *"offset of what ?"* If I see the word "scroll" I immediatly know what it is.

rrahir:
@LucasLefevre I updated the hook name and added some explanations in the ascii :)

src/components/figures/figure/figure.ts
```diff
@@ -97,20 +92,82 @@ css/*SCSS*/ `
...
+ * The container position and size is the same as the split pane
+ * the figure is part of.
+ *
+ * The hook allows to decorrelate the figure position from the scrollvalues.
+ * It represents the inverse viewport position: the position of the viewport
+ * considering the container as the reference point.
+ *
+ * --: container limits
+ * 🟥: hook
+ * ⭐: chart top/left position
```
LucasLefevre: chart => figure

LucasLefevre (commented):
Pushed a commit with some ideas.
What do you think?
Maybe we should ask others from the team if they understand well enough the schema/explanation without reading the code in depth

tests/setup/resize_observer.mock.ts
```diff
@@ -0,0 +1,35 @@
+class MockResizeObserver {
+  public cb: Function;
+  constructor(cb: Function) {
+    this.cb = cb;
+  }
+  observe() {
+    //@ts-ignore
+    global.resizers.add(this);
```
LucasLefevre, line 8: isn't that a memory leak ?
The reference to the callback and hence to the component is kept alive here

rrahir:
@LucasLefevre Added commit to address the memory leak. Note tthat I didn't use a weakset because I need to iterate on its members in `global.resizers`

rrahir:
i f'd up the disconnect part, for tomorrow...

tests/components/scorecard_chart.test.ts
```diff
@@ -325,60 +353,52 @@ describe("Scorecard charts", () => {
...
 
   test("Font size scale down if we put a long key value", async () => {
-    createScorecardChart(
+    await createScorecardChart(
       model,
       { keyValue: "A1", baseline: "B2", title: "This is a title" },
       chartId
     );
-    await nextTick();
+    // await nextTick();
```
LucasLefevre, line 401: coucou

PR #1860, opened 2022-12-06
[FIX] Close data filter popover in dashboard

yunchen4:
## Description:

Now in the dashboard mode, if users want to close an open filter popover, they need to click the filter icon button again and cannot do it just by clicking the grids. 

This PR makes closing popover via clicking grids possible.

Odoo task ID : [3089264](https://www.odoo.com/web#id=3089264&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/dashboard/dashboard.xml
+++ b/src/components/dashboard/dashboard.xml
@@ -1,6 +1,11 @@
 <templates>
   <t t-name="o-spreadsheet-SpreadsheetDashboard" owl="1">
-    <div class="o-grid o-two-columns" tabindex="-1" t-on-wheel="onMouseWheel" t-on-copy="copy">
+    <div
+      class="o-grid o-two-columns"
+      tabindex="-1"
+      t-on-wheel="onMouseWheel"
+      t-on-copy="copy"
+      t-on-click="onClosePopover">
       <div class="mx-auto h-100 position-relative" t-ref="grid" t-att-style="gridContainer">
         <GridOverlay
           onCellHovered.bind="onCellHovered"
--- a/src/components/filters/filter_icon/filter_icon.xml
+++ b/src/components/filters/filter_icon/filter_icon.xml
@@ -1,6 +1,6 @@
 <templates>
   <t t-name="o-spreadsheet-FilterIcon" owl="1">
-    <div class="o-filter-icon" t-att-style="style" t-on-click="props.onClick">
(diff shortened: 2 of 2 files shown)


src/components/filters/filter_icons_overlay/fitler_icons_overlay.xml
```diff
@@ -8,7 +8,10 @@
         position="getFilterHeaderPosition(header)"
         isActive="isFilterActive(header)"
         t-key="'icon_'+header.col + '_' + header.row"
-        onClick="() => this.toggleFilterMenu(header)"
+        onClick="(event) => {
+          this.toggleFilterMenu(header); 
+          event.stopPropagation();
+        }"
```
pro-odoo, line 14:
```suggestion
        onClick.stop="() => this.toggleFilterMenu(header)"
```
You can use `stop` on the handler in order to automatically call `stopPropagation` on the event. See https://github.com/odoo/owl/blob/master/doc/reference/event_handling.md#modifiers

yunchen4: @pro-odoo I tried to use `t-on-click.stop` but an error constantly shows up: 
`Uncaught Error: Invalid handler (expected a function, received: 'undefined')`
And jest tests also failed because of the same error. 
I searched on the codebase and found almost all usages of `t-on-click` is related to some original DOM components (like `div` or `button`). Here we use it in `FilterIcon` component which is defined by us. Does it probably cause issue? 
I will keep using `onClick` until finding out what goes wrong :)

rrahir: You are correct, the issue is that the onCLick in this file is actually a props pased to the component FilterIcon and not an event handling ;-)  I guess that pierre suggestion would rather address the `t-on-click`defined on the component itself.
```html
<div class="o-filter-icon" t-att-style="style" t-on-click="props.onClick">
```
You could add a `.stop` there :)

rrahir (changes requested):

rrahir:
the content is good :)  Only 2 things:
- Could I as you to redirect this PR to version 16.0 ? It shouldn't be to troublesome as we did not change those files between 16.0 and master
- More importantly, don't forget to add tests with the changes you introduce :)

PR #1861, opened 2022-12-06
[IMP] registries: always populate chart/figure registries

hokolomopo:
## Description

Before this, the chart and figure registries were populated across different files. This meant that sometimes we tried to use theses registries, but they were empty because the right files weren't loaded.

Now the registries are populated at the same time they are defined. Importing the registry index should populate all the registries.

Moved the file `menu_items_registry.ts` inside `registry/helpers` and removed it from the registry index to avoid circular dependencies.

Also moved the file `registry.ts` inside the folder `registry/helpers`.

Odoo task ID : [3092792](https://www.odoo.com/web#id=3092792&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


LucasLefevre (approved):
LGTM
Just a small question/suggestion:

do we really want a `helpers` folder here ? `./registries/helpers/registry`
Could just be `./registries/registry`.
What do you think?

`registry.ts` is the basic file. IMO it should not be hidden away in a helper folder

hokolomopo:
> LGTM Just a small question/suggestion:
> 
> do we really want a `helpers` folder here ? `./registries/helpers/registry` Could just be `./registries/registry`. What do you think?
> 
> `registry.ts` is the basic file. IMO it should not be hidden away in a helper folder

@LucasLefevre my goal was to distinguish between the files used for type/class definition of a registry (`registry.ts`/`menu_items_registry.ts`), and the implementation for the registries used in spreadsheet (`figureRegistry`/`chartRegistry`/...). 

But I do agree that `helper` is a wrong name. Not sure what the "correct" hierarchy of forlder would be. Or maybe I can just not care, and throw everything in the registry folder (current commit). Any good idea/suggestions ?

LucasLefevre (commented):
robodoo r+

PR #1864, opened 2022-12-06
[FIX] grid: Capture context menu events triggered by keyboard

rrahir:
Currently, we handle the contextmenu events by capturing the event on the canvas. This made complete sense when the canvas was the default focused element in the DOM but its no longer the case since 19a7ccf1.

The current code would still work because when clicking on the contextmenu mouse button, the DOM will first dispatch a focus event on the canvas and dispatch the context menu event afterwards.

However, a user can trigger a contextmenu event with his keyboard (on non retard proprietary keyboards, hit "Menu key" or Shift+F10). With the change introduced in 19a7ccf1, the event will not be captured by the canvas since there wasn't a click to focus it beforehand and we end up displaying the default browser context menu.

This commit introduced a dedicated callback for this situation that will display the contextmenu under the top left cell of the current selection.

Task 3093099

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/grid.ts
+++ b/src/components/grid.ts
@@ -180,7 +180,11 @@ const TEMPLATE = xml/* xml */ `
         />
     </t>
     <t else="1">
-      <input class="position-absolute" style="z-index:-1000;" t-on-input="onInput" t-ref="hiddenInput"/>
+      <input class="position-absolute"
+        style="z-index:-1000;"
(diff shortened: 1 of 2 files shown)


LucasLefevre (approved):
robodoo r+

PR #1865, opened 2022-12-06
16.0 edgescroll though hidden rar

rrahir:
[FIX] EdgeScroll: Allow edge scrolling through hidden columns

The previous implementation did not allow to edge scroll from top to
bottom (resp. left to right) through hidden rows (resp. columns). The
issue is that we decided to scroll by 1 header at each step but when the
resulting header is hidden we fallback on the next (ascending)
header. You therefore end up trying to scroll up (resp. left) and the
code would force you to scroll the current top (resp left) header.

Task 2857272


pro-odoo (approved):
robodoo r+

PR #1866, opened 2022-12-06
[IMP] components: Remove sidePanelIsOpen props from components

rrahir:
This props was forwarded all the way down to the figures so they could register themsleves as the content of the side panel. However, this was not the right way to handle it for several reasons:

1. This feature only made sense to change the side panel content when a user would alter selecting figures. The implementation here would display the figure side panel even though another type of side panel was already open,e.g. Conditional formats or Search and Replace.

2. The chart side panel already handles this itself since it updates itself when the user selects a new figure.

task 3087981

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/figures/figure/figure.ts
+++ b/src/components/figures/figure/figure.ts
@@ -118,7 +118,6 @@ css/*SCSS*/ `
 
 interface Props {
   figure: Figure;
-  sidePanelIsOpen: boolean;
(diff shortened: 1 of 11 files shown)


LucasLefevre (approved):
robodoo r+

PR #1874, opened 2022-12-09
[FIX] collaborative: don't endlessly dispatch new revision

LucasLefevre:
Steps to reproduce
------------------

- set a number in one cell
- for that cell, hit the "Increase decimal places" button in the toolbar
  (which dispatches a `SET_DECIMAL` command)

=> now every time you open the spreadsheet, a new revision is dispatched,
leading to an enormous amount of initial revisions over time and thousand
of initial RPC requests.
Another consequence is to prevent any snapshot from happening, meaning
the issue will never disappear.

Why does it happen?
-------------------

Once upon a time, `"SET_DECIMAL"` was a core command (https://github.com/odoo/o-spreadsheet/commit/92ca319c8d001fd5e19d03fc1e531b3a8b30f1df).
That means some of those commands are in the initial revisions replayed
when opening the spreadsheet.

When `"SET_DECIMAL"` is replayed, it's handled by `FormatPlugin` **which
is a UI plugin!**
The plugin dispatches a `"SET_FORMATTING"` command as part of `"SET_DECIMAL"`
handling.
Since this `"SET_FORMATTING"` command is a core command coming from a UI
plugin => it's considered as a new revision which is dispatched to the
server.

The same `"SET_DECIMAL"` command is replayed again and again every time
the spreadsheet is open. Hence more and more revisions with
`"SET_FORMATTING"` are created.

Why thousands of requests?
--------------------------
When opening the spreadsheet, you'll see potentially thousands of
requests sending the revision to the server.
It will do one request per initial revision *after* the `"SET_DECIMAL"`
(including all useless previous `"SET_FORMATTING"`! This
adds one request every time you open the spreadsheet).
That's because the `"SET_FORMATTING"` is first dispatched with `"SET_DECIMAL"`
base revisionId (which is not the last). When following initial revisions
are treated, it retries to send the revision (with the new base revision
id).
Every attempt it refused by the server because the base revision id
doesn't match the last server revision id. That is until the very last
attempt which has the correct current server revision id.
Note that all those requests are performed synchronously.

Why no snapshot?
----------------
The snapshot will never be accepted by the server because replaying
commands and the snapshot are executed synchronously.
That means they are both based on the same `serverRevisionId`.

Since the `"SET_FORMATTING"` is first, it's accepted by the server and
the snapshot's `serverRevisionId` is outdated by the time the request
arrives at the server.

opw : [2951236](https://www.odoo.com/web#id=2951236&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)
opw : [3045623](https://www.odoo.com/web#id=3045623&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/model.ts
+++ b/src/model.ts
@@ -266,7 +266,10 @@ export class Model extends owl.core.EventBus implements CommandDispatcher {
 
   private onRemoteRevisionReceived({ commands }: { commands: CoreCommand[] }) {
     for (let command of commands) {
+      const previousStatus = this.status;
+      this.status = Status.RunningCore;
       this.dispatchToHandlers(this.uiPlugins, command);
+      this.status = previousStatus;
     }
     this.finalize();
   }
@@ -404,7 +407,10 @@ export class Model extends owl.core.EventBus implements CommandDispatcher {
       case Status.Finalizing:
         throw new Error(_lt("Cannot dispatch commands in the finalize state"));
       case Status.RunningCore:
-        throw new Error("A UI plugin cannot dispatch while handling a core command");
+        if (isCoreCommand(command)) {
+          throw new Error(`A UI plugin cannot dispatch ${type} while handling a core command`);
+        }
+        this.dispatchToHandlers(this.handlers, command);
     }
     return DispatchResult.Success;
   };
--- a/src/collaborative/session.ts
+++ b/src/collaborative/session.ts
@@ -37,6 +37,7 @@ export class Session extends EventBus<CollaborativeEvent> {
   private pendingMessages: StateUpdateMessage[] = [];
 
   private waitingAck: boolean = false;
+  private isReplayingInitialRevisions = false;
 
   private processedRevisions: Set<UID> = new Set();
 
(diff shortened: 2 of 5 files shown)


src/model.ts
```diff
@@ -266,7 +266,10 @@ export class Model extends owl.core.EventBus implements CommandDispatcher {
 
   private onRemoteRevisionReceived({ commands }: { commands: CoreCommand[] }) {
     for (let command of commands) {
+      const previousStatus = this.status;
+      this.status = Status.RunningCore;
```
LucasLefevre, line 270: this and the `isReplayingInitialRevisions` check in session actually serves the same purpose.
With this, the error in session will never occur because `A UI plugin cannot dispatch while handling a core command` will be thrown first.
Should I still keep the check in session then? Just in case

LucasLefevre: Hmm it breaks some tests

LucasLefevre: Edition plugin dispatches `STOP_EDITION` when handing remove col/row removed...

pro-odoo (approved):
robodoo r+

PR #1880, opened 2022-12-09
[FIX] renderer: box clipRect includes cell borders

hokolomopo:
## Description:

There was a problem in the renderer, the cell content could be displayed
over its border. This was because the `clipRect` of the cell included
the cell's borders.

Odoo task ID : [3002154](https://www.odoo.com/web#id=3002154&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/plugins/ui_feature/renderer.ts
```diff
@@ -733,4 +747,16 @@ export class RendererPlugin extends UIPlugin {
...
+
+  /**
+   * Computes the coordinates and size to draw the zone on the canvas, but exclude the borders of the cells.
+   */
+  private getVisibleRectWithoutBorders(zone: Zone, thinLineWidth: number): Rect {
+    const visibleRect = this.getters.getVisibleRect(zone);
+    return {
+      ...visibleRect,
+      height: visibleRect.height - thinLineWidth,
+      width: visibleRect.width - thinLineWidth,
```
LucasLefevre, line 759: Why only `- thinLineWidth` and not `- 2*thinLineWidth`?
When I think about borders, I would think they are all around. It looks like something is wrong

Also, `thinLineWidth` does not make much sense here. Why does this method needs to know it's "thin" ? Something like `borderWidth` maybe ?

tests/plugins/renderer.test.ts
```diff
@@ -297,12 +362,12 @@ describe("renderer", () => {
...
         {
           id: 1,
-          cols: { 0: { size: 5 }, 2: { size: 25 } },
+          cols: { 0: { size: 6 }, 2: { size: 26 } },
           colNumber: 3,
           cells: {
             A1: { content: "123456" },
             A2: { content: "=A1" },
-            C1: { content: "123456" },
+            C1: { content: "123458" },
```
anhe-odoo, line 370: Why did you changed these values ? (same for the setCellContent below)

anhe-odoo:
Just two small comments:
1. Maybe use something related to "grid" instead of simply "border", to make a difference between the grid lines and the cell borders (as the future border PR will make us able to have other border width but it won't impact this fix)
2. I suppose that the commented code won't be kept, so don't forget to remove it :)

Else LGTM :)

tests/plugins/renderer.test.ts
hokolomopo: To have an easier time at debugging and not have 2 cells with the same value.

LucasLefevre:
Thanks for the work anyway

PR #1887, opened 2022-12-13
[IMP] Grid: hit ESC to clean clipboard selection

yunchen4:
## Description:

Now if there is a clipboard selection and we don't copy/paste in the spreadsheet, the clipboard selection will stay. 

This PR makes hitting ESC key to clean clipboard highlight selection possible. The idea of implementation is basically to add a new command to clean the selection in the clipboard, and add a keyboard event handler in `grid`, which dispatch the new command. 

To copy GSheets behaviors, if a popover or context menu is opened at the same time with a clipboard highlight selection, hitting ESC key will first close the popover/cell context menu without cleaning the clipboard selection. After that, hitting ESC again will clean the clipboard selection. 
The top bar/bottom bar menus are not handled. 

Odoo task ID : [2714347](https://www.odoo.com/web#id=2714347&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/grid/grid.ts
+++ b/src/components/grid/grid.ts
@@ -153,7 +153,9 @@ export class Grid extends Component<Props, SpreadsheetChildEnv> {
   }
 
   onClosePopover() {
-    this.closeOpenedPopover();
+    if (this.env.model.getters.hasOpenedPopover()) {
+      this.closeOpenedPopover();
+    }
     this.focus();
   }
 
@@ -185,6 +187,16 @@ export class Grid extends Component<Props, SpreadsheetChildEnv> {
         sheetId: this.env.model.getters.getActiveSheetId(),
         target: this.env.model.getters.getSelectedZones(),
       });
+    },
+    ESCAPE: () => {
+      /** TODO: Clean once we introduce proper focus on sub components. Grid should not have to handle all this logic */
+      if (this.env.model.getters.hasOpenedPopover()) {
+        this.closeOpenedPopover();
+      } else if (this.menuState.isOpen) {
+        this.closeMenu();
+      } else {
+        this.env.model.dispatch("CLEAN_CLIPBOARD_HIGHLIGHT");
+      }
     },
     "CTRL+A": () => this.env.model.selection.loopSelection(),
     "CTRL+Z": () => this.env.model.dispatch("REQUEST_UNDO"),
@@ -361,7 +373,9 @@ export class Grid extends Component<Props, SpreadsheetChildEnv> {
       this.env.model.dispatch("PREPARE_SELECTION_INPUT_EXPANSION");
     }
 
-    this.closeOpenedPopover();
+    if (this.env.model.getters.hasOpenedPopover()) {
+      this.closeOpenedPopover();
+    }
     if (this.env.model.getters.getEditionMode() === "editing") {
       this.env.model.dispatch("STOP_EDITION");
     }
@@ -413,7 +427,9 @@ export class Grid extends Component<Props, SpreadsheetChildEnv> {
   processArrows(ev: KeyboardEvent) {
     ev.preventDefault();
     ev.stopPropagation();
-    this.closeOpenedPopover();
+    if (this.env.model.getters.hasOpenedPopover()) {
+      this.closeOpenedPopover();
+    }
 
     updateSelectionWithArrowKeys(ev, this.env.model.selection);
 
@@ -495,7 +511,9 @@ export class Grid extends Component<Props, SpreadsheetChildEnv> {
   }
 
   toggleContextMenu(type: ContextMenuType, x: Pixel, y: Pixel) {
-    this.closeOpenedPopover();
+    if (this.env.model.getters.hasOpenedPopover()) {
+      this.closeOpenedPopover();
+    }
     this.menuState.isOpen = true;
     this.menuState.position = { x, y };
     this.menuState.menuItems = registries[type]
--- a/src/plugins/ui_feature/cell_popovers.ts
+++ b/src/plugins/ui_feature/cell_popovers.ts
@@ -12,7 +12,11 @@ import { UIPlugin } from "../ui_plugin";
  * Plugin managing the display of components next to cells.
  */
 export class CellPopoverPlugin extends UIPlugin {
-  static getters = ["getCellPopover", "getPersistentPopoverTypeAtPosition"] as const;
+  static getters = [
+    "getCellPopover",
+    "getPersistentPopoverTypeAtPosition",
+    "hasOpenedPopover",
+  ] as const;
 
   private persistentPopover?: CellPosition & { type: CellPopoverType };
 
@@ -82,6 +86,10 @@ export class CellPopoverPlugin extends UIPlugin {
           ...popover,
           anchorRect: this.computePopoverAnchorRect(position),
         };
+  }
+
+  hasOpenedPopover() {
+    return this.persistentPopover !== undefined;
   }
 
   getPersistentPopoverTypeAtPosition({ col, row }: Position): CellPopoverType | undefined {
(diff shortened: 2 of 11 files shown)


src/plugins/ui_feature/cell_popovers.ts
```diff
@@ -84,6 +88,10 @@ export class CellPopoverPlugin extends UIPlugin {
         };
   }
 
+  hasOpenedPopover() {
+    return this.persistentPopover != null;
```
LucasLefevre, line 92: what if `this.persistentPopover` is `undefined` ?

tests/components/context_menu.test.ts
```diff
@@ -221,6 +221,19 @@ describe("Context Menu", () => {
     expect(fixture.querySelector(".o-menu")).toBeFalsy();
   });
 
+  test("right click on a cell, then hitting esc key closes a context menu", async () => {
+    await rightClickCell(model, "C8");
+    expect(getSelectionAnchorCellXc(model)).toBe("C8");
+    await nextTick();
```
LucasLefevre, line 227: those two lines are not really relevant for the test here

tests/components/context_menu.test.ts
```diff
@@ -221,6 +221,19 @@ describe("Context Menu", () => {
...
+    expect(fixture.querySelector(".o-menu")).toBeTruthy();
+
+    document
+      .querySelector(".o-grid-overlay")!
+      .dispatchEvent(new KeyboardEvent("keydown", { key: "Escape", bubbles: true }));
+    await nextTick();
```
LucasLefevre, line 233: you can use the `keyDown` helper

tests/components/grid.test.ts
```diff
@@ -662,6 +668,19 @@ describe("Grid component", () => {
...
+
+    test("Hitting esc key correctly closes the filter menu", async () => {
+      createFilter(model, "A1:A2");
+      await nextTick();
+      await simulateClick(".o-filter-icon");
+      expect(fixture.querySelectorAll(".o-filter-menu")).toHaveLength(1);
+
+      document
+        .querySelector(".o-grid-overlay")!
+        .dispatchEvent(new KeyboardEvent("keydown", { key: "Escape", bubbles: true }));
```
LucasLefevre, line 680: you can use the `keyDown` helper (everywhere)

tests/components/grid.test.ts
```diff
@@ -662,6 +668,19 @@ describe("Grid component", () => {
       await simulateClick(".o-filter-icon");
       expect(fixture.querySelectorAll(".o-filter-menu")).toHaveLength(1);
     });
+
+    test("Hitting esc key correctly closes the filter menu", async () => {
```
LucasLefevre, line 672: can be moved to file `filter_menu.test.ts`
That's probably the first place we would go looking to find if this behavior is tested

tests/components/grid.test.ts
```diff
@@ -1442,6 +1468,75 @@ describe("Copy paste keyboard shortcut", () => {
...
+    expect(getClipboardVisibleZones(model).length).toBe(1);
+    await nextTick();
+    document
+      .querySelector(".o-grid")!
+      .dispatchEvent(new KeyboardEvent("keydown", { key: "Escape", bubbles: true }));
+    await nextTick();
+    expect(getClipboardVisibleZones(model).length).toBe(0);
+
+    selectCell(model, "A1");
+    cut(model, "A1");
```
LucasLefevre, line 1485: split the copy, and cut in two tests

tests/components/grid.test.ts
```diff
@@ -1405,6 +1424,13 @@ describe("Copy paste keyboard shortcut", () => {
     return event;
   }
 
+  function getClipboardVisibleZones(model: Model): Zone[] {
```
LucasLefevre, line 1427: That's a duplicate of an existing helper function, right ?
You can move it to the `getters_helpers.ts`

tests/plugins/clipboard.test.ts
```diff
@@ -88,6 +89,24 @@ describe("clipboard", () => {
     expect(getCell(model, "D3")).toBeUndefined();
   });
 
+  test("can clean the clipboard visible zones", () => {
+    const model = new Model();
+    setCellContent(model, "B2", "b2");
+    expect(getCell(model, "B2")).toMatchObject({
+      content: "b2",
+    });
```
LucasLefevre, line 97: not useful for this test. We really don't care about the content here, just the outlined zone

LucasLefevre (commented):
only minor comments :)

src/plugins/ui_feature/cell_popovers.ts
yunchen4: > what if `this.persistentPopover` is `undefined` ?

@LucasLefevre `undefined == null` is true so if `this.persistentPopover` is undefined this will return `false`, which should make sense? (Unless when `this.persistentPopover` is `undefined` it means having opened popover).

LucasLefevre: Oh yeah didn't pay attention you used `!=` instead of `!==`
Using `!=` and `==` are error prone and lead to unexpected results.

I think we should never use those operators and be explicit in the predicate expression

tests/components/filter_menu.test.ts
```diff
@@ -201,6 +201,18 @@ describe("Filter menu component", () => {
...
 
+    test("Hitting esc key correctly closes the filter menu", async () => {
+      createFilter(model, "A1:A2");
+      await nextTick();
+      await simulateClick(".o-filter-icon");
+      expect(fixture.querySelectorAll(".o-filter-menu")).toHaveLength(1);
+      document
+        .querySelector(".o-grid")!
+        .dispatchEvent(new KeyboardEvent("keydown", { key: "Escape", bubbles: true }));
+      await nextTick();
```
yunchen4, line 212: use `keyDown` helper here will not pass the tests, but using these lines will. Don't know why (the same thing happens for the filter menu in the `grid.test.ts`, too )

src/components/grid/grid.ts
```diff
@@ -176,6 +176,15 @@ export class Grid extends Component<Props, SpreadsheetChildEnv> {
         target: this.env.model.getters.getSelectedZones(),
       });
     },
+    ESCAPE: () => {
+      if (this.env.model.getters.hasOpenedPopover()) {
+        this.closeOpenedPopover();
```
rrahir, line 181: is the getter really necessary to reach your goal ? we have several calls to this function inside the component and none of them takes the time to check if the popover is opened. We should either check everytwhere or never do it, don't you think ?

yunchen4: > is the getter really necessary to reach your goal ? we have several calls to this function inside the component and none of them takes the time to check if the popover is opened. We should either check everytwhere or never do it, don't you think ?

@rrahir If we want to keep this order of closing/cleaning something, I would say yes. The idea is when there's a popover, hitting ESC will close it (and will not clean the clipboard selection). If there's no popover but a context menu open, hitting ESC closes the context menu.

rrahir: my point is that you could call both functions 100% of the time without checking that it's necessary beforehand, just like we do in other places in the same file. I you want to have a "check" step before closing the opened popovr then it should be used widely accrosse the `Grid` component, not just here ;-)

yunchen4: @rrahir 
Yes, it's not necessary to check beforehand if we want to close something. 
Here `if...else...` is used to make sure that if there is a clipboard selection and a open popover together it will only close the popover, and not clean the clipboard selection (GSheets behaves this way). I need to put something in the `if` statement so that's why the getter is needed. 
If hitting ESC once to close the open popover and clean clipboard selection at the same time is what we want, then the getter is not needed.

rrahir: Ok my bad, I must be partly blind.
Anyway since you have made that getter, please use it to wrap other calls to `this.closeOpenedPopover();` it's not a big optimisations but we don't need a full dispatch when it's not required.

Could you add a note at the top of the piece of code ?
/** TODO:  Clean once we introduce proper focus on sub components. Grid should not have to handle all this logic */

rrahir:
@robodoo r+

rrahir:
@robodoo rebase-ff

PR #1889, opened 2022-12-14
[FIX] Range: prevent crash with sheetname containing an exclamation mark 

yunchen4:
## Description:

This commit addresses an issue in the range plugin that would crash when
handling sheetnames containing an exclamation mark (e.g. `'sheet!1'!A1:A2`)
There were 2 main symptoms:
1. when using such references in a selection input, an error would be thrown
during the processing
2. when writing a reference in a formulu through the composer

In both cases, it would cause the app to crash entirely.

Note that this commit only prevents the crash from happening but testing
these ranges highlighted that we do not properly handle exclamation
marks inside a sheetName. This will be addressed in  task 3112299

Odoo task ID : [3100211](https://www.odoo.com/web#id=3100211&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/tests/plugins/range.test.ts
+++ b/tests/plugins/range.test.ts
@@ -89,6 +89,7 @@ describe("range plugin", () => {
       sheets: [
         { id: "s1", name: "s1", rows: 10, cols: 10 },
         { id: "s2", name: "s 2", rows: 10, cols: 10 },
+        { id: "s1!!", name: "s1!!", rows: 10, cols: 10 },
       ],
     });
     m.dispatch("USE_RANGE", { sheetId: m.getters.getActiveSheetId(), rangesXC: ["B2:D4"] });
@@ -432,6 +433,20 @@ describe("range plugin", () => {
         renameSheet(m, "s1", name);
         const range = m.getters.getRangeFromSheetXC("s1", "A1");
         expect(m.getters.getRangeString(range, "tao")).toBe(`'${name}'!A1`);
+      }
+    );
+
+    test.each([
+      ["s1!!!A1:A9", "'s1!!'!A1:A9"],
+      // TODO: the output is incorrect - should be fixed in task 3112299
+      // ["'s1!!'!A1:A9", "'!A1:A9"],
+      ["s1!!!A1:s1!!!A9", "'s1!!'!A9"],
+      ["s1!!!A1:s1!!!A9", "'s1!!'!A9"],
+    ])(
+      "xc with more than one exclamation mark does not throw error",
+      (rangeString, expectedString) => {
+        const range = m.getters.getRangeFromSheetXC("s1!!", rangeString);
+        expect(m.getters.getRangeString(range)).toBe(expectedString);
       }
     );
   });
--- a/src/plugins/core/range.ts
+++ b/src/plugins/core/range.ts
@@ -303,7 +303,7 @@ export class RangeAdapter implements CommandHandler<CoreCommand> {
     let sheetName = "";
     let prefixSheet = false;
     if (sheetXC.includes("!")) {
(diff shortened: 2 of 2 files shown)


tests/plugins/range.test.ts
```diff
@@ -434,6 +434,13 @@ describe("range plugin", () => {
         expect(m.getters.getRangeString(range, "tao")).toBe(`'${name}'!A1`);
       }
     );
+
+    test.each([["s1!A1:s1!A9"], ["s1!!!A1:s1!!!A9"]])(
```
LucasLefevre, line 438: I'm pretty sure we don't properly support everywhere ranges with the sheet name on both sides.
It require a bit more extensive testing than that.

Also `"s1!!!A1:s1!!!A9"` is not a valid string reference :wink:  If there is a `!` in the sheet name, then the sheet name needs to be enclosed with single quotes `"'s1!!'!A1:'s1!!'!A9"`

tests/plugins/range.test.ts
```diff
@@ -434,6 +434,13 @@ describe("range plugin", () => {
...
+      "xc with more than one exclamation mark does not throw error",
+      (range) => {
+        expect(() => m.getters.getRangeFromSheetXC("s1", range)).not.toThrowError();
```
LucasLefevre, line 441: Better to test the result rather than just that it doesn't throw

yunchen4: @LucasLefevre When there are too many exclamation marks in the input, the input border will become red, which indicates something wrong in the input, which I think it's clear. 
The real problem here is in the current 16.0 implementation this will come with errors and the freezing or even crashed application, and users cannot edit the input (a lot of error dialogs pop up and no confirm button below the input) which is very annoying. 
For now I only deal with the error/crash, give the users possibility to change the inputs and don't care the results after parsing. I think we can create a new task for it?

tests/plugins/range.test.ts
yunchen4: > I'm pretty sure we don't properly support everywhere ranges with the sheet name on both sides. It requires a bit more extensive testing than that (does the tokenizer handle that correctly?).

I checked the range string which is split by "!" and the result arrays are all reversed, so it will not cause error so that no freezing or crashes. 

For the correctness of parsing multiple ! marks, I think none of them can do it correctly for now. For example, `s1!A1:s1!A9` will be split as `["s1", "A1:s1", "A9"]`. But because the input border will become red when inputs are like this, I think it's clear for me to change my input. 

> Also `"s1!!!A1:s1!!!A9"` is not a valid string reference wink If there is a `!` in the sheet name, then the sheet name needs to be enclosed with single quotes `"'s1!!'!A1:'s1!!'!A9"`

Because we parse our range string first using "!", it will also cause error if the split result array is not reversed, and the input border will also be red (implies it is not a valid input).

rrahir (changes requested):
Definitely means we need to fix the parsing/interpretation issue ASAP.

rrahir:
Another thing, the fix in itself has nothing to do with the charts, so the commit title should be updated acordingly

tests/plugins/range.test.ts
yunchen4: @rrahir So the first thing I need to test is something more valid, like `s1!!!A1:A9`, even when they're not parsed correctly for now?

rrahir: yup, I created a task to address that other issue, because it's also impacting version 15.0

rrahir:
@robodoo r+

PR #1890, opened 2022-12-14
[IMP] selection_input: consistent range colors

somu-odoo:
## Description:

Current behaviour of range colors upon being focused in selection input is it shows different colors every time the sidepanel is opened while range is being focused.

To bring a consistency on these colors, simply an index parameter is passed in getNextColor method in order to remove the global variable that is maintained for color index.

Odoo task ID : [2877044](https://www.odoo.com/web#id=2877044&cids=2&menu_id=4720&action=333&active_id=2328&model=project.task&view_type=form)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/color.ts
+++ b/src/helpers/color.ts
@@ -26,12 +26,6 @@ export function colorNumberString(color: number): Color {
   return toHex(color.toString(16).padStart(6, "0"));
 }
 
-let colorIndex = 0;
-export function getNextColor() {
-  colorIndex = ++colorIndex % colors.length;
-  return colors[colorIndex];
-}
-
 /**
  * Converts any CSS color value to a standardized hex6 value.
  * Accepts: hex3, hex6, hex8 and rgb (rgba is not supported)
--- a/src/plugins/ui_feature/selection_input.ts
+++ b/src/plugins/ui_feature/selection_input.ts
@@ -1,4 +1,4 @@
-import { getComposerSheetName, getNextColor, positionToZone, zoneToXc } from "../../helpers/index";
+import { colors, getComposerSheetName, positionToZone, zoneToXc } from "../../helpers/index";
 import { StreamCallbacks } from "../../selection_stream/event_stream";
 import { SelectionEvent } from "../../types/event_stream";
 import { Command, CommandResult, Highlight, LAYERS, UID } from "../../types/index";
@@ -185,7 +185,7 @@ export class SelectionInputPlugin extends UIPlugin implements StreamCallbacks<Se
       ...values.map((xc, i) => ({
         xc,
         id: (this.ranges.length + i + 1).toString(),
-        color: getNextColor(),
+        color: colors[(this.ranges.length + i) % colors.length],
       }))
     );
   }
--- a/tests/plugins/selection_input.test.ts
+++ b/tests/plugins/selection_input.test.ts
@@ -644,4 +644,18 @@ describe("selection input plugin", () => {
     model.dispatch("CHANGE_RANGE", { id, rangeId: idOfRange(model, id, 0), value: "1:1" });
     expect(model.getters.getSelectionInput(id)[0].xc).toBe("1:1");
   });
+
+  test("consistent range colors upon refocusing multiple times", () => {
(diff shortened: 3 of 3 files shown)


src/helpers/color.ts
```diff
@@ -26,9 +26,8 @@ export function colorNumberString(color: number): Color {
   return toHex(color.toString(16).padStart(6, "0"));
 }
 
-let colorIndex = 0;
-export function getNextColor() {
-  colorIndex = ++colorIndex % colors.length;
+export function getNextColor(index: number) {
+  const colorIndex = ++index % colors.length;
```
rrahir, line 30: since you increment the index beforehand, you will skip the first element of `colors`  which should be orange.

src/plugins/ui_feature/selection_input.ts
```diff
@@ -185,7 +185,7 @@ export class SelectionInputPlugin extends UIPlugin implements StreamCallbacks<Se
       ...values.map((xc, i) => ({
         xc,
         id: (this.ranges.length + i + 1).toString(),
-        color: getNextColor(),
+        color: getNextColor(this.ranges.length + i),
```
rrahir, line 188: The helper does not make sense anymore if you have to provide an index to a function called "getNextColor", it defeats the purpose. 
```suggestion
        color: colors[this.ranges.length +1 % colors.length]
```
should do the trick and also address the other issue I  raise on the other comment

somu-odoo: Was not sure if we needed the helper function in any future case that's why I didn't removed it earlier but now I removed it and added your suggested change

src/plugins/ui_feature/selection_input.ts
```diff
@@ -185,7 +185,7 @@ export class SelectionInputPlugin extends UIPlugin implements StreamCallbacks<Se
       ...values.map((xc, i) => ({
         xc,
         id: (this.ranges.length + i + 1).toString(),
-        color: getNextColor(),
+        color: colors[this.ranges.length + i + (1 % colors.length)],
```
rrahir, line 188: There is a slight problem with this function; once this.range.length +i is bigger than the length of color, then you are accessing an index that does not exist.
try 
```suggestion
        color: colors[(this.ranges.length + i) % colors.length],
```

note that I removed +1 on purpose otherwise, you start accessing the color of index 1 and not 0

rrahir (changes requested):

src/plugins/ui_feature/selection_input.ts
somu-odoo: Yep realised that issue later on, and +1 was there just for an increment but since " i " is going to increment anyway removing it seems fair, plus it also covers skipping of first color

tests/plugins/selection_input.test.ts
```diff
@@ -644,4 +644,22 @@ describe("selection input plugin", () => {
     model.dispatch("CHANGE_RANGE", { id, rangeId: idOfRange(model, id, 0), value: "1:1" });
     expect(model.getters.getSelectionInput(id)[0].xc).toBe("1:1");
   });
+
+  test("consistent range colors upon refocusing multiple times", () => {
```
rrahir, line 648: this doesn't test what you fixed, actually. This issue was occuring when we would disable a selection input and enable a new one afterwards. I suggest that you have a look at the commands that are dispatched when you switch from one tab to the other in a chart sidepanel.

In a generic manner, try to first write the test to reproduced the issue before working on the fix ;-)

rrahir (changes requested):

tests/plugins/selection_input.test.ts
somu-odoo: Thanks for the tip, and yes you were correct I indeed missed one command which was mainly responsible for this behaviour, I have changed the test case, and now I think it tests the actual fix.

rrahir:
@robodoo r+

rrahir:
thanks :) :+1:

PR #1902, opened 2022-12-17
[REF] edition: force consistency in the plugin state

rrahir:
## Description:

[REF] edition: force consistency in the plugin state
The plugin internal state is globally inconsistent. Some values are
properly resetted when start/stopping the edition, some are not.
Furthermore, functions are designed such that we need to call fucntions
in a very specific order in order not to crash [1].

This commit forces a stronger typing in the plugin.

[1] https://github.com/odoo/o-spreadsheet/pull/1739#discussion_r1041233112

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_stateful/edition.ts
+++ b/src/plugins/ui_stateful/edition.ts
@@ -23,7 +23,6 @@ import {
   Command,
   CommandResult,
   Format,
-  HeaderIndex,
   Highlight,
   LocalCommand,
   Locale,
@@ -35,7 +34,7 @@ import {
   Zone,
 } from "../../types";
 import { SelectionEvent } from "../../types/event_stream";
-import { UIPlugin } from "../ui_plugin";
+import { UIPlugin, UIPluginConfig } from "../ui_plugin";
 
 export type EditionMode =
   | "editing"
@@ -48,6 +47,42 @@ export interface ComposerSelection {
   start: number;
   end: number;
 }
+
+interface EditingState {
+  col: number;
+  row: number;
+  mode: "editing" | "selecting";
+  sheetId: UID;
+  currentContent: string;
+  currentTokens: EnrichedToken[];
+  selectionStart: number;
+  selectionEnd: number;
+  selectionInitialStart: number;
+  initialContent: string | undefined;
+  previousRef: string;
+  previousRange: Range | undefined;
+  colorIndexByRange: { [xc: string]: number };
+}
+
+interface EmptyState {
+  col: undefined;
+  row: undefined;
+  sheetId: undefined;
+  mode: "inactive";
+  currentContent: "";
+  currentTokens: EnrichedToken[];
+  selectionStart: 0;
+  selectionEnd: 0;
+  selectionInitialStart: 0;
+  initialContent: undefined;
+  previousRef: undefined;
+  previousRange: undefined;
+  colorIndexByRange: {};
+}
+
+type State = EmptyState | EditingState;
+
+export const SelectionIndicator = "␣";
 
 export class EditionPlugin extends UIPlugin {
   static getters = [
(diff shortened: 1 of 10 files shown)


tests/plugins/edition.test.ts
```diff
@@ -157,6 +158,7 @@ describe("edition", () => {
       start: 0,
       end: 0,
     });
+    model.dispatch("START_EDITION");
```
rrahir, line 161: could simply call start_edition with the base content and selection ...

src/plugins/ui_stateful/edition.ts
```diff
@@ -164,41 +221,40 @@ export class EditionPlugin extends UIPlugin {
...
-      case "CHANGE_HIGHLIGHT":
-        const cmdRange = this.getters.getRangeFromRangeData(cmd.range);
-        const newRef = this.getRangeReference(cmdRange, this.previousRange!.parts);
-        this.selectionStart = this.selectionInitialStart;
-        this.selectionEnd = this.selectionInitialStart + this.previousRef.length;
-        this.replaceSelection(newRef);
-        this.previousRef = newRef;
-        this.selectionStart = this.currentContent.length;
-        this.selectionEnd = this.currentContent.length;
+        this.state.selectionInitialStart = previousRefToken!.start;
```
rrahir, line 244: the 5 previous lines now need to be called by this handle ('START_CHANGE_HIGHLIGHTS")  but also when handling 2 selection events ..; and also when activating a new sheet ....  It turns out that we do not really need to store this reference and range because their value and position can actually be infered during the handling of the selection event (since we have previous zone and the new zone in the payload) but this mean doing all the filter and find token everytime we handle a selection event.

Since the code needs to filter and then find in array with a rather 'heavy' condition, I feel like it will not scale that much if we were to call it everytime we change the selection ...  

@LucasLefevre do you have an opinion on the matter ?

src/components/highlight/highlight/highlight.ts
```diff
@@ -34,50 +34,24 @@ export class Highlight extends Component<Props, SpreadsheetChildEnv> {
...
-          left: Math.min(pivotCol, lastCol),
-          top: Math.min(pivotRow, lastRow),
-          right: Math.max(pivotCol, lastCol),
-          bottom: Math.max(pivotRow, lastRow),
-        };
-
-        newZone = this.env.model.getters.expandZone(activeSheetId, newZone);
-
-        if (!isEqual(newZone, currentZone)) {
-          this.env.model.dispatch("CHANGE_HIGHLIGHT", {
```
LucasLefevre, line 76: `CHANGE_HIGHLIGHT` does not seems to be used anymore and can be removed :)
(Tests need to be adapted)

src/plugins/ui_stateful/edition.ts
LucasLefevre: not sure to understand what you mean.
I think it's a hidden message from my brain telling me to stop working for now
I'll try again tomorrow

src/selection_stream/selection_stream_processor.ts
```diff
@@ -373,6 +366,26 @@ export class SelectionStreamProcessor
     });
   }
 
+  /**
+   *  ---- PRIVATE ----
+   */
+
+  private _setZone(
+    anchor: AnchorZone,
+    mode: SelectionEvent["mode"] = "overrideSelection"
```
LucasLefevre, line 375: I would force to be explicit when calling this method by not providing a default value for `mode`

LucasLefevre (commented):
looking good on the surface :)
I'm not particularly fond of writing `this.state....` everywhere in the edition plugin but I guess it's the price to pay to have a stronger type

src/components/highlight/highlight/highlight.ts
rrahir: yup, I still need to adapt the tests :'( I also want to rename start_hcange_highlight to setlect_highlight, it describes it better

src/plugins/ui_stateful/edition.ts
rrahir: We can report this discussion for another task/pr - I realized that we are supposed (?) to support unbounded zone in the resize stuff but we are not really doing it that atm because the Highlight composer don't. 

since selection processor does not handle those zones (yet) we can leave it for another improvement

src/plugins/ui_stateful/edition.ts
```diff
@@ -60,19 +95,30 @@ export class EditionPlugin extends UIPlugin {
...
-  private previousRange: Range | undefined = undefined;
-  private colorIndexByRange: { [xc: string]: number } = {};
+  private state!: State;
+
+  constructor(config: UIPluginConfig) {
+    super(config);
+    this.setDefaultState();
+  }
+
+  setDefaultState() {
```
anhe-odoo, line 105: Isn't it possible to use the "values" from the EmptyState interface here (something like `this.state = new EmptyState` ? This would avoid to write all the values when reinitialising this.state. But this is more like a personnal relfexion, as I'm totally not sure if this can be done

PR #1903, opened 2022-12-19
[IMP] Chart: use first row as headers

yunchen4:
## Description:

This PR changes the current option "Datasets include title" to "Use row/col x as headers", which is more similar to the functionality in GSheets. It will exclude the first row/col in the datasets/labels when creating the chart.

Odoo task ID : [3099995](https://www.odoo.com/web#id=3099995&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/chart/line_bar_pie_panel/config_panel.ts
+++ b/src/components/side_panel/chart/line_bar_pie_panel/config_panel.ts
@@ -1,4 +1,6 @@
 import { Component, useState } from "@odoo/owl";
+import { createRange } from "../../../../helpers";
+import { createDataSets } from "../../../../helpers/figures/charts";
 import { BarChartDefinition } from "../../../../types/chart/bar_chart";
 import { LineChartDefinition } from "../../../../types/chart/line_chart";
 import { PieChartDefinition } from "../../../../types/chart/pie_chart";
@@ -92,6 +94,27 @@ export class LineBarPieConfigPanel extends Component<Props, SpreadsheetChildEnv>
     this.props.updateChart({
       aggregated: ev.target.checked,
     });
+  }
+
+  calculateHeaderPosition(): number | undefined {
+    if (this.isDatasetInvalid || this.isLabelInvalid) {
+      return undefined;
+    }
+    const getters = this.env.model.getters;
+    const sheetId = getters.getActiveSheetId();
+    const labelRange = createRange(getters, sheetId, this.labelRange);
+    const dataSets = createDataSets(
+      getters,
+      this.dataSeriesRanges,
+      sheetId,
+      this.props.definition.dataSetsHaveTitle
+    );
+    if (dataSets.length) {
+      return dataSets[0].dataRange.zone.top + 1;
+    } else if (labelRange) {
+      return labelRange.zone.top + 1;
+    }
+    return undefined;
   }
 }
 
--- a/src/helpers/figures/charts/chart_common.ts
+++ b/src/helpers/figures/charts/chart_common.ts
@@ -27,7 +27,7 @@ import { formatValue } from "../../format";
 import { isDefined } from "../../misc";
 import { copyRangeWithNewSheetId } from "../../range";
 import { rangeReference } from "../../references";
-import { isFullRow, toUnboundedZone, zoneToDimension, zoneToXc } from "../../zones";
+import { getZoneArea, isFullRow, toUnboundedZone, zoneToDimension, zoneToXc } from "../../zones";
 
 /**
  * This file contains helpers that are common to different charts (mainly
@@ -151,8 +151,8 @@ export function createDataSets(
   const dataSets: DataSet[] = [];
   for (const sheetXC of dataSetsString) {
     const dataRange = getters.getRangeFromSheetXC(sheetId, sheetXC);
-    const { unboundedZone: zone, sheetId: dataSetSheetId, invalidSheetName } = dataRange;
-    if (invalidSheetName) {
+    const { unboundedZone: zone, sheetId: dataSetSheetId, invalidSheetName, invalidXc } = dataRange;
+    if (invalidSheetName || invalidXc) {
       continue;
     }
     // It's a rectangle. We treat all columns (arbitrary) as different data series.
@@ -184,13 +184,8 @@ export function createDataSets(
           )
         );
       }
-    } else if (zone.left === zone.right && zone.top === zone.bottom) {
-      // A single cell. If it's only the title, the dataset is not added.
-      if (!dataSetsHaveTitle) {
-        dataSets.push(createDataSet(getters, dataSetSheetId, zone, undefined));
-      }
     } else {
-      /* 1 row or 1 column */
+      /* 1 cell, 1 row or 1 column */
       dataSets.push(
         createDataSet(
           getters,
@@ -256,6 +251,22 @@ export function toExcelDataset(getters: CoreGetters, ds: DataSet): ExcelChartDat
     label: ds.labelCell ? getters.getRangeString(ds.labelCell, "forceSheetReference") : undefined,
     range: getters.getRangeString(dataRange, "forceSheetReference"),
   };
+}
+
+export function toExcelLabelRange(
+  getters: CoreGetters,
+  labelRange: Range | undefined,
+  shouldRemoveFirstLabel?: boolean
+) {
+  if (!labelRange) return undefined;
+  let zone = {
+    ...labelRange.zone,
+  };
+  if (shouldRemoveFirstLabel && labelRange.zone.bottom > labelRange.zone.top) {
+    zone.top = zone.top + 1;
+  }
+  const range = labelRange.clone({ zone });
+  return getters.getRangeString(range, "forceSheetReference");
 }
 
 /**
@@ -352,6 +363,22 @@ export function checkLabelRange(
     }
   }
   return CommandResult.Success;
+}
+
+export function shouldRemoveFirstLabel(
+  labelRange: Range | undefined,
+  dataset: DataSet | undefined,
+  dataSetsHaveTitle: boolean
+) {
+  if (!dataSetsHaveTitle) return false;
+  if (!labelRange) return false;
+  if (!dataset) return true;
+  const datasetLength = getZoneArea(dataset.dataRange.zone);
+  const labelLength = getZoneArea(labelRange.zone);
+  if (labelLength < datasetLength) {
+    return false;
+  }
+  return true;
 }
 
 // ---------------------------------------------------------------------------
--- a/src/xlsx/conversion/figure_conversion.ts
+++ b/src/xlsx/conversion/figure_conversion.ts
@@ -1,4 +1,10 @@
-import { isDefined } from "../../helpers";
+import {
+  getCanonicalSheetName,
+  isDefined,
+  splitReference,
+  toUnboundedZone,
+  zoneToXc,
+} from "../../helpers";
 import { ChartDefinition, ExcelChartDefinition, FigureData } from "../../types";
 import { XLSXFigure, XLSXWorksheet } from "../../types/xlsx";
 import { convertEMUToDotValue, getColPosition, getRowPosition } from "../helpers/content_helpers";
@@ -47,15 +53,20 @@ function convertFigure(
 }
 
 function convertChartData(chartData: ExcelChartDefinition): ChartDefinition | undefined {
-  const labelRange = chartData.dataSets[0].label?.replace(/\$/g, "");
-  let dataSets = chartData.dataSets.map((data) => data.range.replace(/\$/g, ""));
+  const dataSetsHaveTitle = chartData.dataSets[0].label !== undefined;
+  const labelRange = chartData.labelRange
+    ? convertExcelRangeToSheetXC(chartData.labelRange, dataSetsHaveTitle)
+    : undefined;
+  let dataSets = chartData.dataSets.map((data) =>
+    convertExcelRangeToSheetXC(data.range, dataSetsHaveTitle)
+  );
   // For doughnut charts, in chartJS first dataset = outer dataset, in excel first dataset = inner dataset
   if (chartData.type === "pie") {
     dataSets.reverse();
   }
   return {
     dataSets,
-    dataSetsHaveTitle: false,
+    dataSetsHaveTitle,
     labelRange,
     title: chartData.title || "",
     type: chartData.type,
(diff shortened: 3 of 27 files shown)


src/components/side_panel/chart/line_bar_pie_panel/config_panel.ts
```diff
@@ -17,6 +19,10 @@ interface Props {
 interface ChartPanelState {
   datasetDispatchResult?: DispatchResult;
   labelsDispatchResult?: DispatchResult;
+  headerPosition?: {
+    direction: "row" | "column";
```
pro-odoo, line 23: We already have a type for that: https://github.com/odoo/o-spreadsheet/blob/master/src/types/misc.ts#L250

src/components/side_panel/chart/line_bar_pie_panel/config_panel.ts
```diff
@@ -87,6 +104,49 @@ export class LineBarPieConfigPanel extends Component<Props, SpreadsheetChildEnv>
       labelRange: this.labelRange,
     });
   }
+
+  calculateHeaderPosition(): typeof this.state.headerPosition {
```
pro-odoo, line 108: why `typeof` ?

src/components/side_panel/chart/line_bar_pie_panel/config_panel.ts
```diff
@@ -87,6 +104,49 @@ export class LineBarPieConfigPanel extends Component<Props, SpreadsheetChildEnv>
...
+    if (this.isDatasetInvalid || this.isLabelInvalid) {
+      return {
+        index: 1,
+        direction: "row" as const,
```
pro-odoo, line 112: `as const` should not be necessary

src/components/side_panel/chart/line_bar_pie_panel/config_panel.ts
```diff
@@ -87,6 +104,49 @@ export class LineBarPieConfigPanel extends Component<Props, SpreadsheetChildEnv>
...
+      };
+    }
+    const sheetId = this.env.model.getters.getActiveSheetId();
+    const getters = this.env.model.getters;
```
pro-odoo, line 116:
```
const getters = this.env.model.getters;
```

tests/__snapshots__/xlsx_export.test.ts.snap
```diff
@@ -87,7 +87,7 @@ Object {
                     <c:cat>
                         <c:strRef>
                             <c:f>
-                                Sheet1!A2:A5
+                                Sheet1!A3:A5
```
pro-odoo, line 90: Is it really correct ? I'm not sure...

tests/collaborative/collaborative_sheet_manipulations.test.ts
```diff
@@ -575,7 +575,11 @@ describe("Collaborative Sheet manipulation", () => {
       });
       expect([alice, bob, charlie]).toHaveSynchronizedValue(
         (user) => user.getters.getChartDefinition(chartId),
-        { ...chartDef, dataSets: ["A1:A3", "H1:H3"], labelRange: "H3" }
+        {
+          ...chartDef,
+          dataSets: ["A1:A3", "H1:H3"],
+          labelRange: "H3",
+        }
```
pro-odoo, line 582: Seems useless diff (actually the whole file)

tests/plugins/chart/basic_chart.test.ts
```diff
@@ -76,14 +76,14 @@ describe("datasource tests", function () {
       model,
       {
         dataSets: ["Sheet1!B1:B4", "Sheet1!C1:C4"],
-        labelRange: "Sheet1!A2:A4",
+        labelRange: "Sheet1!A1:A4",
```
pro-odoo, line 79:
```suggestion
        labelRange: "Sheet1!A2:A4",
        dataSetsHaveTitle: true,
```
Considering the test data, it should be it, right ? Or I missed something ?

pro-odoo (changes requested):
Little issue with excel import/export:
* Download demo sheet as xlsx
* Import it
 => The line graph is not correct, the range has change and the row header is not ticked anymore.

tests/plugins/chart/basic_chart.test.ts
yunchen4: > Considering the test data, it should be it, right ? Or I missed something ?

@pro-odoo Previously when the datasets have titles, the category XC will be automatically decreased by one, so if the dataset range is `B1:B4` including title, the label range will be changed into `A2:B4`. 
Now I cancel this automatic behavior and make `use row x as headers` to exclude the label header.

src/components/side_panel/chart/line_bar_pie_panel/config_panel.ts
```diff
@@ -26,6 +38,10 @@ export class LineBarPieConfigPanel extends Component<Props, SpreadsheetChildEnv>
   private state: ChartPanelState = useState({
     datasetDispatchResult: undefined,
     labelsDispatchResult: undefined,
+    headerPosition: {
+      direction: "ROW",
```
LucasLefevre, line 42: I don't think we need to support "column".
It's enough for most cases.
Keep it simple :)

src/xlsx/conversion/figure_conversion.ts
```diff
@@ -47,15 +51,18 @@ function convertFigure(
 }
 
 function convertChartData(chartData: ExcelChartDefinition): ChartDefinition | undefined {
-  const labelRange = chartData.dataSets[0].label?.replace(/\$/g, "");
-  let dataSets = chartData.dataSets.map((data) => data.range.replace(/\$/g, ""));
+  const dataSetsHaveTitle = chartData.dataSets[0].label !== undefined;
+  const labelRange = convertExcelLabelToSheetXC(chartData.labelRange, dataSetsHaveTitle);
```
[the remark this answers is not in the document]
yunchen4: @LucasLefevre Because now we exclude the whole row rather than a cell when we ticked `dataset has title`, it will also influence the exporting .xlsx file. And also when we import .xlsx file we also want to automatically tick "use row x as headers" if applicable, so part of code are also about it.

src/components/side_panel/chart/line_bar_pie_panel/config_panel.ts
```diff
@@ -34,6 +38,9 @@ export class LineBarPieConfigPanel extends Component<Props, SpreadsheetChildEnv>
   setup() {
     this.dataSeriesRanges = this.props.definition.dataSets;
     this.labelRange = this.props.definition.labelRange;
+    if (!this.isDatasetInvalid && !this.isLabelInvalid) {
+      this.state.headerIndex = this.calculateHeaderPosition();
+    }
```
yunchen4, line 43: I still keep the state here because two `dispatchResult` are also states. If I don't use state for header index, when the dataset/label is actually invalid, the states of dispatch results are not updated immediately (which leads to valid judgement), so `calculateHeadPosition()` will throw error and the whole app crashes. I don't know whether there is any way better than states 🤔

LucasLefevre (commented):
partial review from last week :)

src/components/side_panel/chart/line_bar_pie_panel/config_panel.ts
```diff
@@ -30,10 +32,12 @@ export class LineBarPieConfigPanel extends Component<Props, SpreadsheetChildEnv>
 
   private dataSeriesRanges: string[] = [];
   private labelRange: string | undefined;
+  private headerPosition: number = 1;
 
   setup() {
     this.dataSeriesRanges = this.props.definition.dataSets;
     this.labelRange = this.props.definition.labelRange;
+    this.headerPosition = this.calculateHeaderPosition();
```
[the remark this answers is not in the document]
yunchen4: @LucasLefevre I remembered I tried to calculate it dynamically, but the problem is the label range and the data set need to be valid first, then we can do the calculation. However, the dispatch results which determine the validness are stored in the state which is not updated immediately (if it works like React in which states are updated in batch, or at least when calling `calculateHeaderPosition()` they are not updated, even when keeping the first lines of code in `calculateHeaderPosition()` which are used to return early if invalid), and it will throw errors to cause the whole app crash. Using `t-if` cannot avoid it either. 
Hence I store `headerPosition` and only update it in setup and after the dataset and label range are confirmed, to be sure that it will not cause error and I think it is enough to update the header position only after confirmation. 
Do you have any idea to make it calculate dynamically without throwing errors related to invalid dataset/label range?

src/components/side_panel/chart/line_bar_pie_panel/config_panel.ts
```diff
@@ -86,13 +95,36 @@ export class LineBarPieConfigPanel extends Component<Props, SpreadsheetChildEnv>
...
 
   onUpdateAggregated(ev) {
     this.props.updateChart({
       aggregated: ev.target.checked,
     });
   }
+
+  calculateHeaderPosition(): number {
+    if (this.isDatasetInvalid || this.isLabelInvalid) {
+      return 1;
```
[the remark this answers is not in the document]
rrahir: You could return `undefined` when the dataset/label is invalid, and use that specific value to decide wheter or not  to display the checkbox ?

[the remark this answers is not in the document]
yunchen4: @rrahir I can directly use `isDatasetInvalid` and `isLabelInvalid` to do the conditional rendering, but the problem is when calling `calculateHeaderPosition()` these two invalid indicators are not updated on time (see my comment in the last conversation).

rrahir (changes requested):

src/components/side_panel/chart/line_bar_pie_panel/config_panel.ts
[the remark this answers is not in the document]
yunchen4: @rrahir Maybe I can push the commit where it is (I think) dynamically compute `headerPosition` but it crashes when the dataset input is invalid? I understand the flow but there might be something I put wrongly : ( But I now know that I need to return something like `undefined` to differentiate it. 
> here is the gist of the diff that I applied it seems to work like a charm.
What gist?

src/helpers/figures/charts/chart_common.ts
```diff
@@ -258,6 +253,73 @@ export function toExcelDataset(getters: CoreGetters, ds: DataSet): ExcelChartDat
...
+  return getters.getRangeString(range, "forceSheetReference");
+}
+
+export function convertExcelDatasetToSheetXC(
+  ds: ExcelChartDataset,
+  dataSetsHaveTitle: boolean
+): string {
+  let sheet = "";
+  if (ds.range.includes("!")) {
+    sheet = ds.range.split("!")[0] + "!";
```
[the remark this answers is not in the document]
yunchen4: @rrahir I wait until #1990 is merged?

src/helpers/figures/charts/chart_common.ts
```diff
@@ -258,6 +253,73 @@ export function toExcelDataset(getters: CoreGetters, ds: DataSet): ExcelChartDat
...
+  }
+  const dataRange = ds.range.replace(/\$/g, "");
```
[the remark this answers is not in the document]
yunchen4: I saw in other similar functions they have this step so I did it

src/helpers/figures/charts/chart_common.ts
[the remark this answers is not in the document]
rrahir: it is, so you can rebase and use `splitReference`  ;-)

src/helpers/figures/charts/chart_common.ts
```diff
@@ -258,6 +253,73 @@ export function toExcelDataset(getters: CoreGetters, ds: DataSet): ExcelChartDat
...
+  let dataZone = toUnboundedZone(dataRange);
```
rrahir, line 285: you can feed the `xc` from plitReference` here :)

src/helpers/figures/charts/chart_common.ts
```diff
@@ -258,6 +253,73 @@ export function toExcelDataset(getters: CoreGetters, ds: DataSet): ExcelChartDat
...
+export function convertExcelLabelToSheetXC(
+  label: string | undefined,
+  dataSetsHaveTitle: boolean
+): string | undefined {
+  if (label === undefined) return undefined;
+  let sheet = "";
+  if (label.includes("!")) {
+    sheet = label.split("!")[0] + "!";
+  }
+  const labelRange = label.replace(/\$/g, "");
```
rrahir, line 308: same comment as above

src/helpers/figures/charts/chart_common.ts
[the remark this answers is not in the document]
yunchen4: @rrahir I checked and you're right, I'll delete the replacement of dollar sign. But what do you mean by fixed rows/cols? What's the use case? 🤔

src/helpers/figures/charts/chart_common.ts
```diff
@@ -258,6 +253,75 @@ export function toExcelDataset(getters: CoreGetters, ds: DataSet): ExcelChartDat
...
+  return getters.getRangeString(range, "forceSheetReference");
+}
+
+export function convertExcelDatasetToSheetXC(
+  ds: ExcelChartDataset,
+  dataSetsHaveTitle: boolean
+): string {
+  let { sheetName, xc } = splitReference(ds.range);
+  if (sheetName) {
+    sheetName = sheetName + "!";
```
rrahir, line 282: this causes an issue, we have to discuss this tomorrow. THe sheetName yields does not have quotes its the pure sheet name. So when you combine your sheetName and dataXC later on, you should check taht hte quotes are present if  necessary. a part of this logic is present and is inadequately called `getComposerSheetName` so you can use  this ATM but it highlights the need to set some kind of guidelines on how to process sheetNames

src/xlsx/conversion/figure_conversion.ts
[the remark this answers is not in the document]
rrahir: for the record, we are already converting excel stuff in the function (if you look at the function arguments)

we could argue that those helpers `convertExcelDatasetToSheetXC` and `convertExcelLabelToSheetXC` don't belong in chart common - they should be defined in the place where they are used -`figure_conversion.ts`. 

FUrthermore, both helpers seem to do the exact same thing (we only process the range of the ExcelChartdataset, and in the same way we do with the label) So it'd be nice to have a single function `convertExcelRangeToSHeetXC` or have both functions call a common helper.

demo/data.js
```diff
@@ -186,7 +186,7 @@ export const demoData = {
           data: {
             type: "line",
             title: "Line",
-            labelRange: "Sheet1!A27:A35",
+            labelRange: "Sheet1!A26:A35",
```
rrahir, line 189: useless diff ?

rrahir (changes requested):
Lots of changes done in the label ranges which are not necessary - those should be rolled back.

I think what we are missing (unless I didn't see the test, if so I apologize) a test that checks that exporting a chart with the label range containing the title row or not leads to the same exported xlsx.
e.g.
``` javascript
{
    labelRange: A1:A3,
    dataSets: [B1:B3],
    dataSetsHaveTitle: true
}
```
genetares the same xlsx as

``` javascript
{
    labelRange: A2:A3,
    dataSets: [B1:B3],
    dataSetsHaveTitle: true
}
```

tests/xlsx_export.test.ts
```diff
@@ -861,7 +885,7 @@ describe("Test XLSX export", () => {
       createChart(
         model,
         {
-          dataSets: ["Sheet1!B2:B4", "Sheet1!C12:C4"],
+          dataSets: ["Sheet1!B2:B4", "Sheet1!C2:C4"],
```
yunchen4, line 888: I doubted those `C12` were typos.

PR #1909, opened 2022-12-21
[FIX] Composer: select headers when composer is open

yunchen4:
## Description:

This PR fixes the usability problem that when the composer is open, the headers cannot be selected. 

Odoo task ID : [2835588](https://www.odoo.com/web#id=2835588&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/grid/grid.ts
+++ b/src/components/grid/grid.ts
@@ -471,8 +471,7 @@ export class Grid extends Component<Props, SpreadsheetChildEnv> {
     const lastZone = zones[zones.length - 1];
     let type: ContextMenuType = "CELL";
     if (!isInside(col, row, lastZone)) {
-      this.env.model.dispatch("UNFOCUS_SELECTION_INPUT");
-      this.env.model.dispatch("STOP_EDITION");
+      this.env.model.selection.getBackToDefault();
       this.env.model.selection.selectCell(col, row);
     } else {
       if (this.env.model.getters.getActiveCols().has(col)) {
--- a/src/components/headers_overlay/headers_overlay.ts
+++ b/src/components/headers_overlay/headers_overlay.ts
@@ -200,6 +200,9 @@ abstract class AbstractResizer extends Component<ResizerProps, SpreadsheetChildE
       this.startMovement(ev);
       return;
     }
+    if (this.env.model.getters.getEditionMode() === "editing") {
+      this.env.model.selection.getBackToDefault();
+    }
     this.startSelection(ev, index);
   }
 
--- a/src/plugins/ui/edition.ts
+++ b/src/plugins/ui/edition.ts
@@ -401,7 +401,9 @@ export class EditionPlugin extends UIPlugin {
       { cell: { col: this.col, row: this.row }, zone },
       {
         handleEvent: this.handleEvent.bind(this),
-        release: () => (this.mode = "inactive"),
+        release: () => {
+          this.stopEdition();
+        },
       }
     );
   }
--- a/src/selection_stream/event_stream.ts
+++ b/src/selection_stream/event_stream.ts
@@ -86,6 +86,17 @@ export class EventStream<Event> {
   }
 
   /**
+   * Release whichever subscription in charge and get back to the default subscription
+   */
+  getBackToDefault() {
+    if (this.mainSubscription === this.defaultSubscription) {
+      return;
+    }
+    this.mainSubscription?.callbacks.release?.();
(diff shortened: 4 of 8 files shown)


src/components/headers_overlay/headers_overlay.ts
```diff
@@ -200,6 +200,10 @@ abstract class AbstractResizer extends Component<ResizerProps, SpreadsheetChildE
       this.startMovement(ev);
       return;
     }
+    if (this.env.model.getters.getEditionMode() === "editing") {
+      this.env.model.dispatch("UNFOCUS_SELECTION_INPUT");
+      this.env.model.dispatch("STOP_EDITION");
+    }
```
LucasLefevre, line 206: as discussed, we should add a possibility on the selection processor to kick the active handler, whatever it is.
This part should not know any detail about who it is and how it works

src/components/grid/grid.ts
```diff
@@ -471,8 +471,8 @@ export class Grid extends Component<Props, SpreadsheetChildEnv> {
     const lastZone = zones[zones.length - 1];
     let type: ContextMenuType = "CELL";
     if (!isInside(col, row, lastZone)) {
-      this.env.model.dispatch("UNFOCUS_SELECTION_INPUT");
-      this.env.model.dispatch("STOP_EDITION");
+      this.env.model.dispatch("STOP_EDITION"); // to keep the content
```
rrahir, line 474: this should be handled by the `getBacktoDefault` function. the release callback set in the edition plugin ( https://github.com/odoo/o-spreadsheet/blob/master/src/plugins/ui_stateful/edition.ts#L418 ) should do the same thing that the `dispatch("stop_edition")` does. you can probably extract the logic in a single function.

src/selection_stream/event_stream.ts
```diff
@@ -85,6 +85,14 @@ export class EventStream<Event> {
     this.mainSubscription = this.defaultSubscription;
   }
 
+  /**
+   * Release whichever subscription in charge and get back to the default subscription
+   */
+  getBackToDefault() {
+    this.mainSubscription?.callbacks.release?.();
```
rrahir, line 92: I wonder if we should *not* call the release when the mainsubscription is the same a the default one.

tests/components/composer.test.ts
```diff
@@ -1206,6 +1207,26 @@ describe("composer", () => {
     expect(fixture.querySelectorAll(".o-grid div.o-composer")).toHaveLength(0);
   });
 
+  test("The composer should be closed before selecting headers", async () => {
+    composerEl = await typeInComposerGrid("=");
+    composerEl.dispatchEvent(new MouseEvent("click"));
```
rrahir, line 1212: why is this necessary ? Don't you have the focus once you start typing?

tests/components/composer.test.ts
```diff
@@ -1206,6 +1207,26 @@ describe("composer", () => {
...
+    expect(fixture.querySelectorAll(".o-grid-composer .o-composer")).toHaveLength(1);
+    await selectColumnByClicking(model, "C");
+    await nextTick();
```
rrahir, line 1215: is the nextTick relevant here? I don't really see what it could change

src/plugins/ui/edition.ts
```diff
@@ -401,7 +401,10 @@ export class EditionPlugin extends UIPlugin {
       { cell: { col: this.col, row: this.row }, zone },
       {
         handleEvent: this.handleEvent.bind(this),
-        release: () => (this.mode = "inactive"),
+        release: () => {
+          this.stopEdition();
+          this.mode = "inactive";
```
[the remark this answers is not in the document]
yunchen4: > this is already taken care of in stopSelection :)

I didn't find any method called `stopSelection` in codebase :( which one did you refer to?

tests/components/composer.test.ts
yunchen4: Because I type in "=" I need to click the cell again to make it into editing mode (otherwise it is selecting mode). I've changed into "Hello" and this click event is deleted now

tests/components/composer.test.ts
yunchen4: Deleting this line will make the test fail. I've added `await nextTick()` in `selectColumnByClicking` helper and this line is deleted.

rrahir:
@robodoo r+

PR #1921, opened 2022-12-29
[IMP] Edition: composer selection set to the saved activated sheet selection rather than reset to (0,0)

yunchen4:
## Description:

Previously when in the composer selecting mode, after activating another sheet, no matter whether the activated sheet has saved selection (other than (0,0)) or not, the composer selection will always be reset to (0,0).

This commit makes the composer selection set to the saved activated sheet selection after activating another sheet, if there's any. Otherwise it will reset to (0,0). 

Odoo task ID : [3077701](https://www.odoo.com/web#id=3077701&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_stateful/edition.ts
+++ b/src/plugins/ui_stateful/edition.ts
@@ -201,10 +201,11 @@ export class EditionPlugin extends UIPlugin {
           this.resetContent();
         }
         if (cmd.sheetIdFrom !== cmd.sheetIdTo) {
+          const activePosition = this.getters.getActivePosition();
           const { col, row } = this.getters.getNextVisibleCellPosition({
             sheetId: cmd.sheetIdTo,
-            col: 0,
-            row: 0,
+            col: activePosition.col,
+            row: activePosition.row,
           });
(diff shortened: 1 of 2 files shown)


src/plugins/ui_stateful/edition.ts
```diff
@@ -201,10 +201,11 @@ export class EditionPlugin extends UIPlugin {
           this.resetContent();
         }
         if (cmd.sheetIdFrom !== cmd.sheetIdTo) {
+          const activePositionInSheetTo = this.getters.getSheetPosition(cmd.sheetIdTo);
```
rrahir, line 204: the sheetIdTo is actually the active sheet, because it was already updated in GridSelectionPlugin. You could write 
```suggestion
          const activePosition = this.getters.getActivePosition();
```

rrahir:
@robodoo r+

PR #1926, opened 2023-01-03
[IMP] Autofill: alphanumeric values

yunchen4:
## Description:

This PR implements the autofill functionality of alphanumeric values (e.g. prefix1). The idea is to use regular expressions to identify the string part and the number part, and reuse some existing logic to calculate the number.

Odoo task ID : [2569019](https://www.odoo.com/web#id=2569019&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/registries/autofill_modifiers.ts
+++ b/src/registries/autofill_modifiers.ts
@@ -9,6 +9,7 @@ import {
   Getters,
   IncrementModifier,
 } from "../types/index";
+import { AlphanumericIncrementModifier } from "./../types/autofill";
 import { Registry } from "./registry";
 
 /**
@@ -19,6 +20,23 @@ import { Registry } from "./registry";
 export const autofillModifiersRegistry = new Registry<AutofillModifierImplementation>();
 
 autofillModifiersRegistry
+  .add("ALPHANUMERIC_INCREMENT_MODIFIER", {
+    apply: (rule: AlphanumericIncrementModifier, data: AutofillData) => {
+      rule.current += rule.increment;
+      const content = `${rule.prefix}${rule.current
+        .toString()
+        .padStart(rule.numberPostfixLength || 0, "0")}`;
+      return {
+        cellData: {
+          border: data.border,
+          style: data.cell && data.cell.style,
+          format: data.cell && data.cell.format,
+          content,
+        },
+        tooltip: { props: { content } },
+      };
+    },
+  })
   .add("INCREMENT_MODIFIER", {
     apply: (rule: IncrementModifier, data: AutofillData) => {
       rule.current += rule.increment;
--- a/src/registries/autofill_rules.ts
+++ b/src/registries/autofill_rules.ts
@@ -1,6 +1,7 @@
 import { DATETIME_FORMAT } from "../constants";
 import { evaluateLiteral } from "../helpers/cells";
 import { AutofillModifier, Cell, CellValueType } from "../types/index";
+import { EvaluatedCell } from "./../types/cells";
 import { Registry } from "./registry";
 
 /**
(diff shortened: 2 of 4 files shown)


src/registries/autofill_rules.ts
```diff
@@ -66,6 +92,27 @@ autofillRulesRegistry
...
     sequence: 10,
   })
+  .add("increment_alphanumeric_value", {
+    condition: (cell: Cell) =>
+      !cell.isFormula &&
+      evaluateLiteral(cell.content).type === CellValueType.text &&
+      alphaNumericValueReg.test(cell.content),
+    generateRule: (cell: Cell, cells: Cell[]) => {
+      const endNumber = parseInt(cell.content.match(endNumberReg)![0]);
+      const startText = cell.content.match(startTextReg)![0];
```
LucasLefevre, line 102: IMHO `prefix` is a better name than `startText`

LucasLefevre (commented):
I would introduce a new `ALPHANUMERIC_INCREMENT_MODIFIER` rule

src/registries/autofill_rules.ts
```diff
@@ -18,11 +18,15 @@ interface AutofillRule {
...
+const endNumberReg = /(\d+)$/;
+const startTextReg = /^[\D]*/;
+const alphaNumericValueReg = /[[\D]*[\d]+]*/;
+
 /**
  * Get the consecutive xc that are of type "number" or "date".
  * Return the one which contains the given cell
  */
-function getGroup(cell: Cell, cells: (Cell | undefined)[]): number[] {
+function getGroup(cell: Cell, cells: (Cell | undefined)[], groupStartText?: string): number[] {
```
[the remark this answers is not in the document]
yunchen4: @LucasLefevre Do you mean: we add a callback parameter as a filter function, `getGroup` finally gets a group of cells, and then we do mapping/filtering in the according rule? 
e.g. 
`
function getGroup(
  cell: Cell,
  cells: (Cell | undefined)[],
  filter: (evaluatedCell: EvaluatedCell) => boolean
) {
  let group: EvaluatedCell[] = [];
  let found: boolean = false;
  for (let x of cells) {
    ...
    const cellValue = evaluateLiteral(x?.content);
    if (filter(cellValue)) {
      group.push(cellValue);
    } else {
      ...
    }
  }
  return group;
}
`
and in the increment number rule we map the group we get: 
`
const group = getGroup(
        cell,
        cells,
        (evaluatedCell) => evaluatedCell.type === CellValueType.number
      ).map((cell) => cell.value as number);
`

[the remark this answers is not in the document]
LucasLefevre: Oh yeah, you would need to map and cast... :unamused: 
well I don't know which is simpler/more readable...
what do you think?

src/registries/autofill_rules.ts
```diff
@@ -18,20 +19,28 @@ interface AutofillRule {
 
 export const autofillRulesRegistry = new Registry<AutofillRule>();
 
+const numberPostfixRegExp = /(\d+)$/;
+const stringPrefixRegExp = /^[\D]*/;
+const alphaNumericValueRegExp = /\D*\d+/;
```
LucasLefevre, line 24: should actually be `/\D+\d+/;`, right?

tests/plugins/autofill.test.ts
```diff
@@ -331,6 +331,32 @@ describe("Autofill", () => {
       expect(getCellContent(model, "A4")).toBe("test");
     });
 
+    describe("Autofill alphanumeric values", () => {
```
LucasLefevre, line 334: add a test with numbers inside de prefix: e.g. `"pr3fix1"`

src/registries/autofill_modifiers.ts
```diff
@@ -19,6 +20,25 @@ import { Registry } from "./registry";
...
 
 autofillModifiersRegistry
+  .add("ALPHANUMERIC_INCREMENT_MODIFIER", {
+    apply: (rule: AlphanumericIncrementModifier, data: AutofillData) => {
+      rule.current += rule.increment;
+      const prefix = rule.prefix;
+      const content = `${prefix}${rule.current
+        .toString()
+        .padStart(rule.numberPostfixLength || 0, "0")}`;
+      const tooltipValue = rule.prefix ? content : formatValue(rule.current, data.cell?.format);
```
LucasLefevre, line 30: does formatting with the format make sense ? Shouldn't it be always `content`?
prefix should always be truthy, right ? The autofill rule itself doesn't make sense

pro-odoo (approved):
robodoo r+

PR #1928, opened 2023-01-03
[IMP] Function: add `hidden` tag to hide functions

yunchen4:
## Description:

This PR adds the tag `hidden` to hide some functions to users, so users will not have the chance to see their existences.

Odoo task ID : [3102309](https://www.odoo.com/web#id=3102309&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/composer/composer/composer.ts
+++ b/src/components/composer/composer/composer.ts
@@ -407,12 +407,14 @@ export class Composer extends Component<Props, SpreadsheetChildEnv> {
 
   showAutocomplete(searchTerm: string) {
     this.autoCompleteState.showProvider = true;
-    let values = Object.entries(functionRegistry.content).map(([text, { description }]) => {
-      return {
-        text,
-        description,
-      };
-    });
+    let values = Object.entries(functionRegistry.content)
+      .filter(([_, { hidden }]) => !hidden)
+      .map(([text, { description }]) => {
+        return {
+          text,
+          description,
+        };
+      });
     if (searchTerm) {
       values = fuzzyLookup(searchTerm, values, (t) => t.text);
     } else {
--- a/src/types/functions.ts
+++ b/src/types/functions.ts
@@ -36,6 +36,7 @@ export interface AddFunctionDescription {
   args: ArgDefinition[];
   returns: [ArgType];
   isExported?: boolean;
+  hidden?: boolean;
 }
 
 export interface FunctionDescription extends AddFunctionDescription {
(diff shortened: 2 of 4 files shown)


src/components/composer/composer/composer.ts
```diff
@@ -169,6 +167,17 @@ export class Composer extends Component<Props, SpreadsheetChildEnv> {
...
+    return Object.keys(functionRegistry.content)
+      .filter((key) => {
+        return !functionRegistry.content[key].hidden;
+      })
+      .reduce((prevContent, currentKey) => {
+        prevContent[currentKey] = functionRegistry.content[currentKey];
+        return prevContent;
+      }, {});
+  }
+
```
yunchen4, line 180: Should I convert it into a state and use `useEffect` to update the state? Using `useEffect` seems to have less calculation.

src/components/composer/composer/composer.ts
```diff
@@ -370,7 +379,7 @@ export class Composer extends Component<Props, SpreadsheetChildEnv> {
 
   showAutocomplete(searchTerm: string) {
     this.autoCompleteState.showProvider = true;
-    let values = Object.entries(functionRegistry.content).map(([text, { description }]) => {
+    let values = Object.entries(this.nonHiddenFunctions).map(([text, { description }]) => {
```
LucasLefevre, line 382: what about just adding `.filter(([text, { hidden }]) => !hidden)` ?
It look simpler than unpack, filter, reduce
```suggestion
    let values = Object.entries(functionRegistry.content).filter(([text, { hidden }]) => !hidden).map(([text, { description }]) => {
```

src/components/composer/composer/composer.ts
LucasLefevre: in practice I don't think it is a performance bottleneck to filter the functions every time.
There only a few hundred functions. Iterating that amount of items should be very fast

src/types/functions.ts
```diff
@@ -36,13 +36,15 @@ export interface AddFunctionDescription {
...
   isExported?: boolean;
+  hidden?: boolean;
 }
 
 export interface FunctionDescription extends AddFunctionDescription {
   minArgRequired: number;
   maxArgPossible: number;
   nbrArgRepeating: number;
   getArgToFocus: (argPosition: number) => number;
+  hidden?: boolean;
```
LucasLefevre: not useful since it extends `AddFunctionDescription` which has the same property definition

LucasLefevre (approved):
LGTM
(I applied my comment)
Side note about commit messages: the line length shouldn't be above ~80 characters :wink: 
robodoo r+

PR #1932, opened 2023-01-05
[FIX] renderer: grid lines visible when background color is set

yunchen4:
## Description:

This PR fixes the problem of almost invisible grid lines when the background color is set to be some color similar to cell border color.

The solution is to add borders with similar but darker color above the background color according to the text align and whether it is overflowing.

Odoo task ID : [3103403](https://www.odoo.com/web#id=3103403&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/plugins/ui_feature/renderer.ts
```diff
@@ -144,24 +145,72 @@ export class RendererPlugin extends UIPlugin {
...
       if (style.fillColor && style.fillColor !== "#ffffff") {
         ctx.fillStyle = style.fillColor || "#ffffff";
-        ctx.fillRect(box.x, box.y, box.width, box.height);
+        ctx.fillRect(
+          box.x - thinLineWidth / 2,
+          box.y - thinLineWidth / 2,
+          box.width + thinLineWidth,
+          box.height + thinLineWidth
+        );
+        if (areGridLinesVisible) {
```
LucasLefevre, line 173: drawing grid lines seems scattered in several places. There's one part in `drawBackground` and there some here now.
Could it be moved all together somewhere ?
Maybe a new method `drawGridLines` ?

src/plugins/ui_feature/renderer.ts
```diff
@@ -144,24 +145,72 @@ export class RendererPlugin extends UIPlugin {
...
+              y2: box.y,
+            });
+            //bottom
+            this.drawBorder(ctx, borderColor, thinLineWidth, {
+              x1: box.x - thinLineWidth / 2,
+              y1: box.y + box.height,
+              x2: box.x + box.width + thinLineWidth / 2,
+              y2: box.y + box.height,
+            });
+            if (box.content?.align === "left") {
```
LucasLefevre, line 195: I don't understand what those two if statements do. If I move their body outside, I don't see anything different.

tests/plugins/renderer.test.ts
```diff
@@ -398,9 +406,183 @@ describe("renderer", () => {
...
+      sheets: [
+        {
+          colNumber: 1,
+          rowNumber: 3,
+        },
+      ],
+    });
+    model.dispatch("SET_FORMATTING", {
+      sheetId: model.getters.getActiveSheetId(),
+      target: [toZone("A1")],
```
LucasLefevre, line 431:
```suggestion
      target: target("A1"),
```

src/plugins/ui_feature/renderer.ts
yunchen4: @LucasLefevre These two if conditions are used to only render the left/right colorful border when the cell content is overflowing and left/right aligned. If the text alignment is center, both left and right colorful borders are not rendered. 
GSheets has this kind of behavior (Excel web handles this background color pretty roughly). 
left
right
center

src/plugins/ui_feature/renderer.ts
yunchen4: > drawing grid lines seems scattered in several places. There's one part in `drawBackground` and there some here now. Could it be moved all together somewhere ? Maybe a new method `drawGridLines` ?

@LucasLefevre I am not sure what you mean "moved all together somewhere". Do you mean:
1. extract these three lines of code into a new function and call this new function in each place 
OR 
2. draw these grid lines all in once together with the drawing in `drawBackground`? 

If it is 1: sure!
If it is 2: I don't think the "gridlines" in `drawCellBackground` (gridlines 1) can be drawn together with the gridlines in `drawBackground` (gridlines 2). Gridlines1 is used to differentiate the borders between the user-defined background color and the gridlines 2 (the target of this task), so it needs to be drawn above the user-defined background color and the overflowing background (which is used to eliminate the overflowing cell separators). Gridlines2 need to be drawn under the overflowing background, otherwise the overflowing background doesn't work.

src/plugins/ui_feature/renderer.ts
```diff
@@ -54,6 +56,11 @@ import { UIPlugin } from "../ui_plugin";
 // -----------------------------------------------------------------------------
 
 export const CELL_BACKGROUND_GRIDLINE_STROKE_STYLE = "#111";
+const LEFT = "left";
+const RIGHT = "right";
+const TOP = "top";
+const BOTTOM = "bottom";
+type GridLineDirection = typeof LEFT | typeof RIGHT | typeof TOP | typeof BOTTOM;
```
hokolomopo, line 63: can probably be an Enum no?

src/plugins/ui_feature/renderer.ts
```diff
@@ -204,32 +261,113 @@ export class RendererPlugin extends UIPlugin {
...
+      x1,
+      y1,
+      x2,
+      y2,
+    }: {
+      x1: number;
+      y1: number;
+      x2: number;
+      y2: number;
+    }
```
hokolomopo, line 308: kind of a detail, but I'd be probably better if you defined the interface ``{x1 : number, y1 : ....}` somewhere else, having the arguments of the function spread on 8 lines is IMHO ulgy and unreadable.

IMO if unpacking an argument takes more than one line because of prettier, we should just unpack in the first line of the funciton instead, I think its more readable.

src/plugins/ui_feature/renderer.ts
```diff
@@ -204,32 +261,113 @@ export class RendererPlugin extends UIPlugin {
...
+      x,
+      y,
+      width,
+      height,
+    }: {
+      x: number;
+      y: number;
+      width: number;
+      height: number;
+    }
```
hokolomopo, line 333: We have the `Rect` interface for that

hokolomopo (commented):
Some small details on code style.

I'm not really sure what changed since the previous iteration that Lucas reviewed, so I'll just green the task and let him review it

rrahir:
We let the task stall for too long but it should actually target saas-161, not master

PR #1934, opened 2023-01-06
[REF] compiler: use code builder

LucasLefevre:
The goal here is to have more flexibility in the way we
assign (or not) intermediate compiled AST nodes to a variable before
using it later. This prepares the ground for automatic function argument
casting.

Currently, `compileAST` function decides to assign the result or not.

Notice how `isLazy` parameter of `compileAST` is used.
One of its effect is to assign (or not) the result to a variable
for the caller to use. (how is that linked to laziness? [🧐](https://emojipedia.org/face-with-monocle/))
It should be the caller's responsibility to do it. Only the caller
knows what it will do with the result and if it makes sense or not.

This brings noise to `compileAST` with early returns and temporary
`statement` variable.

It allows to remove `isLazy` from `compileAST`'s parameters.
That makes sense because laziness only makes sense when compiling function
arguments. It is not a property of any AST node itself.
Now laziness is located in `compileFunctionArgs` only.
## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/formulas/compiler.ts
+++ b/src/formulas/compiler.ts
@@ -4,6 +4,7 @@ import { concat, parseNumber, removeStringQuotes } from "../helpers";
 import { _lt } from "../translation";
 import { CompiledFormula } from "../types";
 import { BadExpressionError } from "../types/errors";
+import { FunctionCode, FunctionCodeBuilder, Scope } from "./code_builder";
 import { AST, ASTFuncall, parseTokens } from "./parser";
 import { rangeTokenize } from "./range_tokenizer";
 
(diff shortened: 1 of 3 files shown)


src/formulas/code_block.ts
```diff
@@ -0,0 +1,106 @@
+/**
+ * Block of code that produces a value.
+ */
+export interface CodeBlock {
```
LucasLefevre, line 4: maybe could be simply named `Code`? What do you think?

src/formulas/compiler.ts
```diff
@@ -169,7 +136,7 @@ export function compile(formula: string): CompiledFormula {
         }
       }
 
-      let listArgs: CompiledAST[] = [];
+      let listArgs: CodeBlock[] = [];
```
pro-odoo, line 139: a `listArgs` which is a list of `CodeBlock` seems weird (naming pov)

pro-odoo (commented):
👍 
Do not forget task-id in commit

src/formulas/compiler.ts
LucasLefevre: listArgs => compiledArgs ?

pro-odoo: Seems better

src/formulas/code_block.ts
LucasLefevre: I also tought about that...it works
the only thing is: it can be used for something else than the body of a function :man_shrugging:

src/formulas/compiler.ts
```diff
@@ -211,139 +178,101 @@ export function compile(formula: string): CompiledFormula {
...
       hasRange = false,
       referenceVerification: {
         functionName?: string;
         paramIndex?: number;
       } = {}
-    ): CompiledAST {
-      const codeBlocks: string[] = [];
-      let id, fnName, statement;
+    ): FunctionCode {
+      const code = new FunctionCodeBuilder(scope);
```
LucasLefevre, line 213: function ? functionCode ?
```suggestion
      const function = new FunctionCodeBuilder(scope);
```

pro-odoo:
robodoo r+

PR #1940, opened 2023-01-09
[IMP] selection: selecting row/col now add full unbounded range

anhe-odoo:
## Task Description

When selecting a range and clicking on a col/row header, the added range should be a full unbounded range (ie `A:A` or `1:1`). ATM, the resulting range is simply bounded to the maximum available cell (ie `A1:A100` or `A1:Z1`).

## Related Task
Odoo task ID : [3076514](https://www.odoo.com/web#id=3076514&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## Review Checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/core/merge.ts
+++ b/src/plugins/core/merge.ts
@@ -1,9 +1,12 @@
 import {
   clip,
+  getCanonicalSheetName,
   isDefined,
   isEqual,
   overlap,
   positions,
+  RangeImpl,
+  splitReference,
   toXC,
   toZone,
   union,
@@ -52,6 +55,7 @@ export class MergePlugin extends CorePlugin<MergeState> implements MergeState {
     "getMerge",
     "getMergesInZone",
     "isSingleCellOrMerge",
+    "getSelectionRangeString",
   ] as const;
 
   private nextId: number = 1;
@@ -153,6 +157,29 @@ export class MergePlugin extends CorePlugin<MergeState> implements MergeState {
     return Array.from(mergeIds)
       .map((mergeId) => this.getMergeById(sheetId, mergeId))
       .filter(isDefined);
+  }
+
+  /**
+   * Same as `getRangeString` but add all necessary merge to the range to make it a valid selection
+   */
+  getSelectionRangeString(range: Range, forSheetId: UID): string {
+    const rangeImpl = RangeImpl.fromRange(range, this.getters);
+    const expandedZone = this.getters.expandZone(rangeImpl.sheetId, rangeImpl.zone);
+    const expandedRange = rangeImpl.clone({
+      zone: {
+        ...expandedZone,
+        bottom: rangeImpl.isFullCol ? undefined : expandedZone.bottom,
+        right: rangeImpl.isFullRow ? undefined : expandedZone.right,
+      },
+    });
+    const rangeString = this.getters.getRangeString(expandedRange, forSheetId);
+    if (this.isSingleCellOrMerge(rangeImpl.sheetId, rangeImpl.zone)) {
+      const { sheetName, xc } = splitReference(rangeString);
+      return `${sheetName !== undefined ? getCanonicalSheetName(sheetName) + "!" : ""}${
+        xc.split(":")[0]
+      }`;
+    }
+    return rangeString;
   }
 
   /**
--- a/src/plugins/core/range.ts
+++ b/src/plugins/core/range.ts
@@ -36,7 +36,6 @@ export class RangeAdapter implements CommandHandler<CoreCommand> {
 
   static getters = [
     "getRangeString",
-    "getSelectionRangeString",
     "getRangeFromSheetXC",
     "createAdaptedRanges",
     "getRangeDataFromXc",
@@ -319,22 +318,6 @@ export class RangeAdapter implements CommandHandler<CoreCommand> {
     const rangeInterface = { prefixSheet, zone, sheetId, invalidSheetName, parts };
 
     return new RangeImpl(rangeInterface, this.getters.getSheetSize).orderZone();
-  }
-
-  /**
-   * Same as `getRangeString` but add all necessary merge to the range to make it a valid selection
-   */
-  getSelectionRangeString(range: Range, forSheetId: UID): string {
-    const rangeImpl = RangeImpl.fromRange(range, this.getters);
-    const expandedZone = this.getters.expandZone(rangeImpl.sheetId, rangeImpl.zone);
-    const expandedRange = rangeImpl.clone({
-      zone: {
-        ...expandedZone,
-        bottom: rangeImpl.isFullCol ? undefined : expandedZone.bottom,
-        right: rangeImpl.isFullRow ? undefined : expandedZone.right,
-      },
-    });
-    return this.getRangeString(expandedRange, forSheetId);
   }
 
   /**
--- a/src/plugins/core/range.ts
+++ b/src/plugins/core/range.ts
@@ -24,6 +24,7 @@ import {
   RangeData,
   RangeProvider,
   UID,
+  UnboundedZone,
   Zone,
 } from "../../types/index";
 
@@ -41,6 +42,7 @@ export class RangeAdapter implements CommandHandler<CoreCommand> {
     "getRangeDataFromXc",
     "getRangeDataFromZone",
     "getRangeFromRangeData",
+    "getRangeFromZone",
   ] as const;
 
   // ---------------------------------------------------------------------------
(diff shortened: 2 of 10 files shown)


LucasLefevre (changes requested):
Range is already managing unbounded zones, 
We should reuse it :wink:

hokolomopo (commented):
The goal of using ranges everywhere for unbonded zones was to hide all the techincal implementation from the plugins. Doing all the computations manually then using `RangeImpl` defeats a bit the purpose IMO.

Looks good otherwise 👍

src/plugins/core/range.ts
```diff
@@ -373,15 +373,27 @@ export class RangeAdapter implements CommandHandler<CoreCommand> {
...
+      sheetId: rangeImpl.sheetId,
+      col: range.zone.left,
+      row: range.zone.top,
+    });
+    const cellBotRight = this.getters.getMainCellPosition({
+      sheetId: rangeImpl.sheetId,
+      col: range.zone.right,
+      row: range.zone.bottom,
+    });
+    const sameCell = cellTopLeft.col === cellBotRight.col && cellTopLeft.row === cellBotRight.row;
```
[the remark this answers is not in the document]
anhe-odoo: In fact it has been exactly copied from zoneToXC :P But I can change both as, indeed, it's a better name

anhe-odoo:
@hokolomopo : it should be okay now :)

src/plugins/core/range.ts
```diff
@@ -399,6 +415,30 @@ export class RangeAdapter implements CommandHandler<CoreCommand> {
...
+      {
+        sheetId,
+        zone:
+          isFullCol || isFullRow
+            ? {
+                ...zone,
+                bottom: isFullCol ? undefined : zone.bottom,
+                right: isFullRow ? undefined : zone.right,
+              }
+            : zone,
```
hokolomopo, line 434: can do without the ternary `isFullCol || isFullRow ? ... : ...`

src/plugins/core/range.ts
```diff
@@ -399,6 +415,30 @@ export class RangeAdapter implements CommandHandler<CoreCommand> {
     return { _sheetId: sheetId, _zone: zone };
   }
 
+  getRangeFromZone(
+    sheetId: UID,
+    zone: Zone,
+    parts: RangePart[],
+    { isFullRow, isFullCol }
```
hokolomopo, line 422: Should be typed as boolean

src/plugins/core/range.ts
```diff
@@ -399,6 +415,30 @@ export class RangeAdapter implements CommandHandler<CoreCommand> {
     return { _sheetId: sheetId, _zone: zone };
   }
 
+  getRangeFromZone(
+    sheetId: UID,
+    zone: Zone,
+    parts: RangePart[],
```
hokolomopo, line 421: might as well make the arg optional, since we only ever use the value `[{ colFixed: false, rowFixed: false },{ colFixed: false, rowFixed: false },]`

hokolomopo (commented):
Some random details. Looks good otherwise :+1:

rrahir (changes requested):
I do'nt think adding yet another 2 arguments to the Selection events is the way to go, as we will send boh a zone and some boolean saying that this zone is actually not correct but rather unbounded.

I think we have to choose between 2 approaches:
1.  let the selectionStreamProcessor only handle zones (like right now) and the plugins would then interpret them as unbounded zones by comparing the provided zone with the size of the concerned sheet.
2. Let the selectionstreamProcessor handle unbounded zones, this would mean having a special behaviour in selection where they'd contain an unbounded zone but the getters would return a zone (not sure about this but was my first thought) 

I think I recall that the first draft involved selectionProcessor to handle unbounded zones. if so,  I might have missed the reason why it was not a good plan :/

let's discuss this :)

rrahir:
unrelated to the task directly but Unbounded zone is actually badly typed, we cannot have right and bottom be undefined at the same time

LucasLefevre:
> unrelated to the task directly but Unbounded zone is actually badly typed, we cannot have right and bottom be undefined at the same time

It was pointed out at the time and it's something that I wanted to be improved. However, it's not as straight forward as it seems.  It was decided to merge as it was because the PR had already waited way too long.
I didn't abandon the idea of improving this though :)

rrahir:
let it be known that in excel, you can actually have right and bottom undefined at the same time...

anhe-odoo:
@rrahir : It should be okay (for the moment :D )

rrahir:
> let it be known that in excel, you can actually have right and bottom undefined at the same time...

let it be known that I was probably high when I "discovered" that behaviour.

src/plugins/core/range.ts
```diff
@@ -373,15 +376,29 @@ export class RangeAdapter implements CommandHandler<CoreCommand> {
...
+      sheetId: rangeImpl.sheetId,
+      col: zone.left,
+      row: zone.top,
+    });
+    const cellBotRight = this.getters.getMainCellPosition({
+      sheetId: rangeImpl.sheetId,
+      col: zone.right,
+      row: zone.bottom,
+    });
+    const isSingleCell =
```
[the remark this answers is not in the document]
rrahir: From the conversation we had with Lucas, I think we decided to not expose the merges in plugin range, This unfortunately means to put more woirk in the selection_input and edition plugins as they'll have to detect thoses separately :/

rrahir (changes requested):

src/plugins/core/range.ts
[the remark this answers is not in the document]
anhe-odoo: I added a new getter in merge to centralize the logic :) If it suits you I'll add some testing in merge and it should then be okay :)

rrahir (commented):
I am not convinced by the latest approach :/ The issue being that we now 'officially' manipulated unbounded zones outside of ranges (which is not that good of a news  imho, we need to define a general approach) and also, the arguments of `getMergeCorrectedRangeString` are a bit fishy, activeSheet in a core plugin does not make sense but I have to say I don't really have an other idea.

I think we could try another way:   Convert the zones/unbounded zone to ranges inside the plugin and have `getMergeCorrectedRangeString` receive a target sheetId and a range, like `getRangeString` , you could then use isSingleCellOrMerge with range.zone and wouldn't have to type cast it, which seems less shady ^^

rrahir (commented):
Globally okay, for me, just requires a rebase (again sorry :'( ) and clean up the commented code :)

src/plugins/core/range.ts
```diff
@@ -384,6 +387,25 @@ export class RangeAdapter implements CommandHandler<CoreCommand> {
     return { _sheetId: sheetId, _zone: zone };
   }
 
+  getRangeFromZone(
+    sheetId: UID,
+    zone: UnboundedZone,
+    parts: RangePart[] = [
```
LucasLefevre, line 393: `parts` is unused. You can remove the parameter

tests/components/selection_input.test.ts
```diff
@@ -162,6 +168,34 @@ describe("Selection Input", () => {
...
+  });
+
+  test("can select full row as unbounded zone by clicking on header", async () => {
+    const { model } = await createSelectionInput();
+    model.selection.selectRow(2, "overrideSelection");
+    await nextTick();
+    expect(fixture.querySelector("input")!.value).toBe("3:3");
+  });
+
+  test("can correctly select a merged zone", async () => {
```
LucasLefevre, line 185: Ah ! here is the test for the first commit ^^

tests/plugins/sheets.test.ts
```diff
@@ -971,4 +971,21 @@ describe("sheets", () => {
...
+
+    // Case 1: Range without any full col/row
+    let zone = toZone("A1:E5");
+    expect(model.getters.getUnboundedZone(sheetId, zone)).toEqual(zone);
+
+    // Case 2: Range with full row
+    zone = toZone("A1:A10");
+    expect(model.getters.getUnboundedZone(sheetId, zone)).toEqual({ ...zone, bottom: undefined });
+
+    // Case 3: Range with full col
```
LucasLefevre, line 987: then split the test in 3 tests, each for one case.
It's ok if you have to copy the two "setup" lines

LucasLefevre (commented):
yo coucou
Some suggestions and comments

src/plugins/core/merge.ts
```diff
@@ -155,6 +159,29 @@ export class MergePlugin extends CorePlugin<MergeState> implements MergeState {
       .filter(isDefined);
   }
 
+  /**
+   * Same as `getRangeString` but add all necessary merge to the range to make it a valid selection
+   */
+  getSelectionRangeString(range: Range, forSheetId: UID): string {
```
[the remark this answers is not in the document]
rrahir: > As far as I can tell, the range "plugin" RangeAdapter (which is actually not a plugin) only uses its getters in its own getters.

this is not true unfortunately, we rely on the getters to adapt the ranges (though it's just in one case atm, {**hard**} maybe we could get rid of it. that being said, I don't see how it would solve the dependency cycle issue :/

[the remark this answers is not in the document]
LucasLefevre: you're right :disappointed:

src/plugins/core/range.ts
anhe-odoo: It is used in the RangeImpl constructor right behind

anhe-odoo: Oh, okay, I got what you mean

rrahir:
@robodoo r+

rrahir:
@robodoo rebase-ff

PR #1945, opened 2023-01-11
[REF] menu: simplify menu api

LucasLefevre:
The current menu item API has several problems:

- there are two concept: `MenuItem` and `FullMenuItem`. This is confusing and unnecessary. From an external point of view, "full" doesn't mean anything.

- many properties can be of different types (function or primitive value). it makes things unnecessarily complex and we had to introduce helpers to hide it (`getMenuName`, `getMenuChildren`, etc.)

- when adding a child menu item in a registry, it must be a `FullMenuItem` item, while it can be a `MenuItem` when adding a root menu item. This is inconsistent.

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/registries/menu_items_registry.ts
```diff
@@ -24,89 +25,117 @@ import { SpreadsheetChildEnv } from "../types/env";
...
     return this;
   }
 
-  /**
-   * Get a list of all elements in the registry, ordered by sequence
-   * @override
-   */
-  getAll(): FullMenuItem[] {
-    return super.getAll().sort((a, b) => a.sequence - b.sequence);
+  getMenu(): MenuItem[] {
```
pro-odoo, line 138: We could go one step further, integrate the isVisible here ?
```
getMenu(env: SpreadsheetChildEnv) {
  return createMenu(this.getAll()).filter((x) => x.isVisible(env));
}
```
We could also add this step in `createMenuItem` in order to avoid to create a menu item which is invisible

pro-odoo: Also, I'm not fan of the name `getMenu`, `getMenuItems` is not a better choice ?

src/registries/menu_items_registry.ts
```diff
@@ -24,89 +25,117 @@ import { SpreadsheetChildEnv } from "../types/env";
...
+export interface MenuItem {
+  id: string;
+  sequence: number;
+  name: (env: SpreadsheetChildEnv) => string;
+  isVisible: (env: SpreadsheetChildEnv) => boolean;
+  isEnabled: (env: SpreadsheetChildEnv) => boolean;
+  isReadonlyAllowed: boolean;
+  action?: (env: SpreadsheetChildEnv) => unknown;
+  children: (env: SpreadsheetChildEnv) => MenuItem[];
+  separator: boolean;
```
pro-odoo, line 52: Is it the time to find a better solution for the separator ?
For exemple, instead of `separator`, add a property `group`, and keep the menu items grouped and add a separator after each group ?

pro-odoo (commented):
👍 
On which branch are you based ? Push yesterday and already 8 conflicted files :D

src/registries/menu_items_registry.ts
LucasLefevre: > We could go one step further, integrate the isVisible here ?
> 
> ```
> getMenu(env: SpreadsheetChildEnv) {
>   return createMenu(this.getAll()).filter((x) => x.isVisible(env));
> }
> ```
> 
> We could also add this step in `createMenuItem` in order to avoid to create a menu item which is invisible

I thought about something similar. But actually, I don't think it's a good idea to call `isVisible` too early (or any other function property). It should be called when we really need it and when we render the menu.
Because it might be visible at one time, but not 30s later (or the other way around).
Also, that way, we have a clear separation between the static menu definition and then the "runtime" menu where the properties have to be computed 

> Also, I'm not fan of the name getMenu, getMenuItems is not a better choice ?

isn't a "group of menu items" a "menu" ? :stuck_out_tongue: 
I don't mind changing the name though

src/components/figures/figure_image/figure_image.ts
```diff
@@ -22,31 +22,30 @@ export class ImageFigure extends Component<Props, SpreadsheetChildEnv> {
   private menuButtonPosition = useAbsolutePosition(this.menuButtonRef);
   private position = useAbsolutePosition(this.imageContainerRef);
 
-  private getMenuItemRegistry(): MenuItemRegistry {
-    const registry = new MenuItemRegistry();
-    registry.add("copy", {
+  private readonly menuItems = createMenu([
```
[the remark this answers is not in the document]
LucasLefevre: If it's not extented in any way, I don't see the point of a registry. It would just be an extra useless layer.

src/registries/menu_items_registry.ts
LucasLefevre: In another PR ? :roll_eyes: 
Do you about something like adding `group: "add_row"`, `group: "add_col"` on each menu item, then group them based on the group property.
It begs some questions about the `sequence` property. How would it affect order across groups ?

pro-odoo: I don't know exactly how we could order the different groups, but it's the idea.
Anyway, it will be done in another PR.

src/components/figures/figure_image/figure_image.ts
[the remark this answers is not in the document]
pro-odoo: For now it's not extended, but someone could have a usecase. IMHO it's strange from an external point of view to have every except one "menu features" that is extensible.
But ok to keep like that in this PR

src/registries/menu_items_registry.ts
pro-odoo: > I thought about something similar. But actually, I don't think it's a good idea to call isVisible too early (or any other function property). It should be called when we really need it and when we render the menu.
Because it might be visible at one time, but not 30s later (or the other way around).
Also, that way, we have a clear separation between the static menu definition and then the "runtime" menu where properties have to be computed

Ok, it's better to have a clear separation. It's just a pity that we have to apply the filter manually each time.

> isn't a "group of menu items" a "menu" ? 😛
I don't mind changing the name though

`getMenu` does not suggest me that it's an array that we get, so at first glance I had to go and see the definition. But I agree that a menu is a group of menu items, so ok for me

tests/components/top_bar.test.ts
```diff
@@ -361,16 +360,16 @@ describe("TopBar component", () => {
...
+    const number = items.filter((item) => item.children(parent.env).length !== 0).length;
     expect(fixture.querySelectorAll(".o-topbar-menu")).toHaveLength(number);
     triggerMouseEvent(".o-topbar-menu[data-id='file']", "click");
     await nextTick();
     expect(fixture.querySelectorAll(".o-menu")).toHaveLength(1);
-    const file = topbarMenuRegistry.get("file");
-    const numberChild = getMenuChildren(file, parent.env).filter(
-      (item) => item.children.length !== 0 || item.action
-    ).length;
+    const file = topbarMenuRegistry.getMenu().find((item) => item.id === "file")!;
```
hokolomopo, line 369: Can replace all the `registry.getMenu().find(...)` in the tests whith the helper `getNode(path, registry)`

hokolomopo (commented):
This looks better, having both `FullMenuItems` and `MenuItem`exposed was kinda dumb.

Don't forget the odoo task Id in the commit (and a PR in odoo most likely)

src/registries/menu_items_registry.ts
pro-odoo: > And for the filter how about handling it in the Menu component ? We handle the isEnabled there, I don't see why we coulnd't handle the isVisible 

👍

src/registries/menu_items_registry.ts
```diff
@@ -24,89 +25,117 @@ import { Registry } from "./registry";
...
+  return menuItems.map(createMenuItem).sort((a, b) => a.sequence - b.sequence);
 }
 
-function isMenuItem(value: FullMenuItem | MenuItemsBuilder): value is FullMenuItem {
-  return typeof value !== "function";
+function createMenuItem(item: MenuItemSpec): MenuItem {
+  const name = item.name;
+  const children = item.children;
+  return {
+    id: item.id || new UuidGenerator().uuidv4(),
```
rrahir, line 69: new constructor everytime ?

rrahir (commented):
Good for me :)

yunchen4 (commented):
Good to me :D

pro-odoo:
robodoo r+

PR #1948, opened 2023-01-12
[FIX] collaborative: Ensure full payload of `UPDATE_CHART`

rrahir:
When we refactored the charts in PR #1393 , we did not take into account that the payload of `UPDATE_CHART` could contain only partial chart definitions. Since we now rely on full definition in the payload, commands with the old payload are no longer valid and will raise an error.

This commit ensures that the replayed command have the adequate payload
when possible. There are 2 situations that we do not properly support:

1. Charts created following a duplicate_sheet: In that case, we need to
command to be processed by the plugins in order to know the new charts
ids and definitions. In this case, we simply drop the update command
altogether in order to let the data load.

2. Some chart was created with an incomplete definition: in theory, this
should not happen since the menu action to create charts creates them
with a full definition. But if someone decided to craft commands
manually, then the replay of messages will crash.

Task 3131804

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/migrations/data.ts
+++ b/src/migrations/data.ts
@@ -328,6 +328,7 @@ export function repairInitialMessages(
   initialMessages = fixTranslatedSheetIds(data, initialMessages);
   initialMessages = dropCommands(initialMessages, "SORT_CELLS");
   initialMessages = dropCommands(initialMessages, "SET_DECIMAL");
+  initialMessages = fixChartDefinitions(data, initialMessages);
   return initialMessages;
 }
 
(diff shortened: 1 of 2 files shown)


src/migrations/data.ts
```diff
@@ -383,6 +384,44 @@ function dropCommands(initialMessages, commandType: string) {
...
+      const commands: CoreCommand[] = [];
+      for (const cmd of message.commands) {
+        let command = cmd;
+        switch (cmd.type) {
+          case "CREATE_CHART":
+            map[cmd.id] = cmd.definition;
+            break;
+          case "UPDATE_CHART":
+            const definition = map[cmd.id];
+            command = { ...cmd, definition: { ...definition, ...cmd.definition } };
```
pro-odoo, line 409: I think you should add `map[cmd.id] = { ...definition, ...cmd.definition }`.

Because if you have two commands `UPDATE_CHART` which update different "fields", only the last one will be update the definition.

Let's say:
`CREATE_CHART({ name: "hello", dataSets: "A1" })` 
`UPDATE_CHART({ name: "bla" })`
`UPDATE_CHART({ dataSets: "A2" })`
The chart definition will be `{ name: "hello", dataSets: "A2" }`
Right ?

pro-odoo (approved):
robodoo r+

PR #1954, opened 2023-01-16
[IMP] Grid Composer: add cell reference when scrolling

pro-odoo:
With this revision, the cell reference is displayed when scrolling in order to help the user to know where he is in the grid.

Task-id 3096000

### DIFF
--- a/src/components/composer/grid_composer/grid_composer.ts
+++ b/src/components/composer/grid_composer/grid_composer.ts
@@ -1,7 +1,7 @@
-import { Component, onMounted, useRef, useState } from "@odoo/owl";
+import { Component, onMounted, onWillUpdateProps, useRef, useState } from "@odoo/owl";
 import { ComponentsImportance, SELECTION_BORDER_COLOR } from "../../../constants";
 import { fontSizeMap } from "../../../fonts";
-import { positionToZone } from "../../../helpers";
+import { deepEquals, getComposerSheetName, positionToZone, toXC } from "../../../helpers";
 import { ComposerSelection } from "../../../plugins/ui_stateful/edition";
 import { DOMDimension, Rect, Ref, SpreadsheetChildEnv, Zone } from "../../../types/index";
 import { getTextDecoration } from "../../helpers";
@@ -10,6 +10,8 @@ import { Composer } from "../composer/composer";
 import { ZoneDimension } from "./../../../types/misc";
 
 const COMPOSER_BORDER_WIDTH = 3 * 0.4 * window.devicePixelRatio || 1;
+const GRID_CELL_REFERENCE_TOP_OFFSET = 28;
+
 css/* scss */ `
   div.o-grid-composer {
     z-index: ${ComponentsImportance.GridComposer};
@@ -19,6 +21,17 @@ css/* scss */ `
 
     display: flex;
     align-items: center;
+  }
+
+  div.o-cell-reference {
+    position: absolute;
+    z-index: ${ComponentsImportance.GridComposer};
+    background: ${SELECTION_BORDER_COLOR};
+    color: white;
+    font-size: 12px;
+    line-height: 14px;
+    padding: 6px 7px;
+    border-radius: 4px;
   }
 `;
 
@@ -46,6 +59,7 @@ export class GridComposer extends Component<Props, SpreadsheetChildEnv> {
 
   private zone!: Zone;
   private rect!: Rect;
+  private isCellReferenceVisible!: boolean;
 
   private composerState!: ComposerState;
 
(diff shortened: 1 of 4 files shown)


src/components/composer/grid_composer/grid_composer.ts
```diff
@@ -74,6 +87,39 @@ export class GridComposer extends Component<Props, SpreadsheetChildEnv> {
     });
   }
 
+  get shouldDisplayCellReference(): boolean {
+    if (this.isCellReferenceVisible) {
+      return true;
+    }
+    const { sheetId, col, row } = this.env.model.getters.getActivePosition();
+    const zone = this.env.model.getters.expandZone(sheetId, positionToZone({ col, row }));
```
rrahir, line 95: think you could use getACtiveSheetId & geSelectedZone here.

src/components/composer/grid_composer/grid_composer.ts
```diff
@@ -74,6 +87,39 @@ export class GridComposer extends Component<Props, SpreadsheetChildEnv> {
...
+    const rect = this.env.model.getters.getVisibleRect(zone);
+    if (
+      !deepEquals(rect, this.rect) ||
+      sheetId !== this.env.model.getters.getCurrentEditedCell()?.sheetId
```
rrahir, line 99:
```suggestion
      sheetId !== this.env.model.getters.getCurrentEditedCell().sheetId
```

src/components/composer/grid_composer/grid_composer.ts
```diff
@@ -74,6 +87,39 @@ export class GridComposer extends Component<Props, SpreadsheetChildEnv> {
...
+    ) {
+      this.isCellReferenceVisible = true;
+    }
+    return this.isCellReferenceVisible;
+  }
+
+  get cellReference(): string {
+    const { col, row, sheetId } = this.env.model.getters.getCurrentEditedCell();
+    const prefixSheet = sheetId !== this.env.model.getters.getActiveSheetId();
+    return `${prefixSheet ? this.env.model.getters.getSheetName(sheetId) + "!" : ""}${toXC(
```
rrahir, line 109: this whole computation pretty much already exists as an internal function in `EditionPlugin` as `getZoneReference` I don't know if it'd be worth exposing for this use, though.

src/components/composer/grid_composer/grid_composer.ts
```diff
@@ -74,6 +87,39 @@ export class GridComposer extends Component<Props, SpreadsheetChildEnv> {
...
+      col,
+      row
+    )}`;
+  }
+
+  get cellReferenceStyle(): string {
+    const { x: left, y: top } = this.rect;
+    return `
+      left: ${left - 1}px;
```
rrahir, line 118: could you cast this as contants ? why -1 ?

pro-odoo: cast this as contants ? What do you mean ?
The -1 is here because it's prettier

src/components/composer/grid_composer/grid_composer.ts
```diff
@@ -74,6 +89,39 @@ export class GridComposer extends Component<Props, SpreadsheetChildEnv> {
...
+      return true;
+    }
+    const sheetId = this.env.model.getters.getActiveSheetId();
+    const zone = this.env.model.getters.getSelectedZone();
+    const rect = this.env.model.getters.getVisibleRect(zone);
+    if (
+      !deepEquals(rect, this.rect) ||
+      sheetId !== this.env.model.getters.getCurrentEditedCell().sheetId
+    ) {
+      this.isCellReferenceVisible = true;
```
LucasLefevre, line 103: kinda strange to have a getter writing stuff. Do we need to store the result? Is it to keep it always visible as soon as it appeared once? If yes, what do you think of doing this in a `willUpdateProps` ? That way it would be more obvious it's updating the internal state (which I don't expect from a getter)

src/components/composer/grid_composer/grid_composer.ts
```diff
@@ -74,6 +89,39 @@ export class GridComposer extends Component<Props, SpreadsheetChildEnv> {
...
+    }
+    return this.isCellReferenceVisible;
+  }
+
+  get cellReference(): string {
+    const { col, row, sheetId } = this.env.model.getters.getCurrentEditedCell();
+    const prefixSheet = sheetId !== this.env.model.getters.getActiveSheetId();
+    return `${prefixSheet ? this.env.model.getters.getSheetName(sheetId) + "!" : ""}${toXC(
```
LucasLefevre, line 111: it may require quotes around the sheet name.
A bit merdique, but you can use `getComposerSheetName`
or maybe we don't care

LucasLefevre (commented):
nice :)

src/components/composer/grid_composer/grid_composer.ts
pro-odoo: I intentionaly don't put the quotes because it's not like in the composer. I think it's useless :)

rrahir: Consider a weird sheet name like `sheet is nice !!`, `sheet is nice !!!A1` vs `'sheet is nice !!'!A1`, shouldn't we have a consistent approach to always show the quotes by default and only support the absence of quotes and not the other way around ?

LucasLefevre: Whatever we decide, adding a test documenting the final decision would be nice :)
(I kinda agree with Rémi. Even If I don't really care.)

LucasLefevre (approved):
robodoo r+

PR #1961, opened 2023-01-17
[IMP] Autofill: support edge scrolling

yunchen4:
## Description:

This commit support the edge scrolling in autofill. 
The basic idea is to reuse `dragAndDropBeyondTheViewport` function. We also need to cancel the check of whether the autofill is active (which is based on the visibility of the autofill selection) to make the tooltip and handler still available during edge scrolling. 

Odoo task ID : [2870875](https://www.odoo.com/web#id=2870875&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/autofill/autofill.ts
+++ b/src/components/autofill/autofill.ts
@@ -2,9 +2,8 @@ import { Component, useState, xml } from "@odoo/owl";
 import { AUTOFILL_EDGE_LENGTH } from "../../constants";
 import { clip } from "../../helpers";
 import { HeaderIndex, SpreadsheetChildEnv } from "../../types";
-import { css } from "../helpers/css";
-import { gridOverlayPosition } from "../helpers/dom_helpers";
-import { startDnd } from "../helpers/drag_and_drop";
+import { css, cssPropertiesToCss } from "../helpers/css";
+import { dragAndDropBeyondTheViewport } from "../helpers/drag_and_drop";
 
 // -----------------------------------------------------------------------------
 // Autofill
@@ -12,35 +11,36 @@ import { startDnd } from "../helpers/drag_and_drop";
 
 css/* scss */ `
   .o-autofill {
-    height: 6px;
-    width: 6px;
-    border: 1px solid white;
     position: absolute;
+    height: ${AUTOFILL_EDGE_LENGTH}px;
+    width: ${AUTOFILL_EDGE_LENGTH}px;
+    border: 1px solid white;
+    box-sizing: border-box !important;
     background-color: #1a73e8;
+  }
 
-    .o-autofill-handler {
-      position: absolute;
-      height: ${AUTOFILL_EDGE_LENGTH}px;
-      width: ${AUTOFILL_EDGE_LENGTH}px;
-
-      &:hover {
-        cursor: crosshair;
-      }
+  .o-autofill-handler {
+    position: absolute;
+    height: ${AUTOFILL_EDGE_LENGTH}px;
+    width: ${AUTOFILL_EDGE_LENGTH}px;
+    &:hover {
+      cursor: crosshair;
     }
+  }
 
-    .o-autofill-nextvalue {
-      position: absolute;
-      background-color: #ffffff;
-      border: 1px solid black;
-      padding: 5px;
-      font-size: 12px;
-      pointer-events: none;
-      white-space: nowrap;
-    }
+  .o-autofill-nextvalue {
+    position: absolute;
+    background-color: #ffffff;
+    border: 1px solid black;
+    padding: 5px;
+    font-size: 12px;
+    pointer-events: none;
+    white-space: nowrap;
   }
 `;
 
 interface Props {
+  isVisible: boolean;
   position: Position;
 }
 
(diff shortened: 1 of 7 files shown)


src/components/autofill/autofill.ts
```diff
@@ -100,15 +99,21 @@ export class Autofill extends Component<Props, SpreadsheetChildEnv> {
...
+        this.state.position = {
+          left: ev.clientX - start.left + offsetX,
+          top: ev.clientY - start.top + offsetY,
+        };
+      } else {
+        this.state.position = {
+          left: ev.clientX,
+          // TODO: 60 is hardcoded. We need more advanced method to determine the value, considering the autofill CSS position will become negative numbers when the autofill selected region is invisible, and it has impact on tooltip & handler whose CSS position is absolute which relates to the position of autofill itself.
+          top: ev.clientY - 60,
+        };
```
yunchen4, line 115: I tried many calculations using `props.position` (the position value passed from `grid`), etc., and this is the one that works best under different scenes when the autofill selected region is invisible. 
The only point is `60` is hardcoded and I don't really know where it comes from... I got this value based on screenshot measurement and some trial-and-error 😅 Any idea is welcome

src/components/grid/grid.xml
```diff
@@ -56,7 +56,7 @@
         onMouseWheel.bind="onMouseWheel"
         onClosePopover.bind="onClosePopover"
       />
-      <t t-if="env.model.getters.getEditionMode() === 'inactive' and isAutoFillActive()">
```
yunchen4, line 59: `Autofill` component needs to be always active for tooltips and handler to work properly when edge scrolling

src/components/autofill/autofill.ts
```diff
@@ -100,15 +99,21 @@ export class Autofill extends Component<Props, SpreadsheetChildEnv> {
...
+      if (this.props.position.left > 0 && this.props.position.top > 0) {
+        // autofill selected region still visible
+        this.state.position = {
+          left: ev.clientX - start.left + offsetX,
+          top: ev.clientY - start.top + offsetY,
+        };
+      } else {
+        this.state.position = {
+          left: ev.clientX,
+          // TODO: 60 is hardcoded. We need more advanced method to determine the value, considering the autofill CSS position will become negative numbers when the autofill selected region is invisible, and it has impact on tooltip & handler whose CSS position is absolute which relates to the position of autofill itself.
```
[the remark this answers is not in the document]
yunchen4: @LucasLefevre Agreed, and I tried several ways but using `ev.clientX` and `ev.clientY` here is the simplest and most accurate solution IMHO. 
Handler and tooltip are wrapped inside the autofill div and the position of autofill div is decided by the visible rect. Once the rect is invisible both `left` and `top` become `-4`. 
We can surely change the position here but change to what? The mouse location? It seems the same as we use `clientX` and `clientY` here. And we don't care where the autofill div is (because it will be hidden visually).

src/components/grid/grid.xml
yunchen4: > when it's outside the viewport, it's actually still visible at the very top-left grid corner aa

@LucasLefevre Good eyes 😄 I'll change it to `visibility:hidden`.

src/components/autofill/autofill.ts
yunchen4: Solved, `60` is come from `top` of the grid canvas.

src/components/autofill/autofill.ts
```diff
@@ -36,11 +36,13 @@ css/* scss */ `
       font-size: 12px;
       pointer-events: none;
       white-space: nowrap;
+      visibility: visible;
```
rrahir, line 39: this is not a good approach. You are setting the vibisility of the whole component  to "invisible" and at the same time, you make the children `div`'s visible. if anything, it probably questions the current structure of the component.

src/components/autofill/autofill.ts
```diff
@@ -100,15 +106,20 @@ export class Autofill extends Component<Props, SpreadsheetChildEnv> {
...
+      if (this.props.position.left > 0 && this.props.position.top > 0) {
+        // autofill selected region still visible
+        this.state.position = {
+          left: ev.clientX - start.left + scrollX,
+          top: ev.clientY - start.top + scrollY,
+        };
+      } else {
+        this.state.position = {
+          left: ev.clientX,
+          top: ev.clientY - offsetTop,
```
rrahir, line 120: why only the top ?

rrahir (commented):
Here are some comments I had but I ended up fixing most of them myself in the commit I pushed. Still wanted to draw your attention to some points.

I did not fix the tests comments though

tests/components/autofill.test.ts
```diff
@@ -103,27 +103,18 @@ describe("Autofill component", () => {
...
-      <div
-        class="o-autofill-nextvalue"
-        style="top:-15px;left:235px;"
-      >
-        <div>
-          test
-        </div>
-      </div>
-    `);
+    expect(fixture.querySelector(".o-autofill-nextvalue")).toBeNull();
```
yunchen4, line 113: TBH I don't know why it becomes null in the test and IMO it shouldn't

rrahir: my bad, in my commit ,I changed 'mouseMove' callback to the div `o-autofill-handler` and not `o-autofill`  I don't recall changing the test to check that it was null though but I noticed that at other places, still don't know what happended, but ultimately, it is a test I should have updated and missed :'(  sorry about that. If you change the selector to `o-autofill-handler` at lign 94, it should solve your issue (keep in mind tha the top and left value fo the snapshot will have been updated

rrahir:
@robodoo r+

PR #1963, opened 2023-01-18
[FIX] top_bar: respond menu upon scaling viewport

jash-odoo:
## Description:

Earlier, if we open drop down menu from top bar with a viewport less in height than needed by the menu, menu scales out from the viewport. In order to solve this issue, applying maxHeight and overflow CSS on drop down menu does the trick.

Odoo task ID : [3127234](https://www.odoo.com/web#id=3127234&cids=2&menu_id=4720&action=4043&model=project.task&view_type=form)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/color_picker/color_picker.ts
+++ b/src/components/color_picker/color_picker.ts
@@ -7,9 +7,9 @@ import {
 } from "../../constants";
 import { hslaToRGBA, isColorValid, rgbaToHex } from "../../helpers";
 import { chartFontColor } from "../../helpers/charts";
-import { Color } from "../../types";
+import { Color, Pixel } from "../../types";
 import { SpreadsheetChildEnv } from "../../types/env";
-import { css } from "../helpers/css";
+import { css, cssPropertiesToCss } from "../helpers/css";
 
 const PICKER_PADDING = 6;
 
@@ -35,8 +35,9 @@ css/* scss */ `
     z-index: ${ComponentsImportance.ColorPicker};
     box-shadow: 1px 2px 5px 2px rgba(51, 51, 51, 0.15);
     background-color: white;
-    padding: ${PICKER_PADDING}px 0px;
     line-height: 1.2;
+    overflow-y: auto;
+    overflow-x: hidden;
     width: ${GRADIENT_WIDTH + 2 * PICKER_PADDING}px;
 
     .o-color-picker-section-name {
@@ -176,6 +177,7 @@ function computeCustomColor(ev: MouseEvent) {
 }
 
 export interface ColorPickerProps {
+  maxHeight: Pixel;
   dropdownDirection?: "left" | "right" | "center";
   onColorPicked: (color: Color) => void;
   currentColor: Color;
@@ -208,6 +210,14 @@ export class ColorPicker extends Component<ColorPickerProps, SpreadsheetChildEnv
       top: "0",
     },
   });
+
+  get dropdownStyle() {
+    const height = this.props.maxHeight;
+    return cssPropertiesToCss({
+      padding: height <= 0 ? "0px" : `${PICKER_PADDING}px 0px`,
+      "max-height": `${height}px`,
+    });
+  }
 
   onColorClick(color: Color) {
     if (color) {
--- a/src/components/popover/popover.ts
+++ b/src/components/popover/popover.ts
@@ -47,9 +47,12 @@ export class Popover extends Component<Props, SpreadsheetChildEnv> {
     // within our control and to avoid leaking into external DOM
     const horizontalPosition = `left:${this.horizontalPosition() - this.spreadsheetPosition.x}`;
     const verticalPosition = `top:${this.verticalPosition() - this.spreadsheetPosition.y}`;
-    const height = `max-height:${
+    const maxHeight = Math.max(
+      0,
       this.viewportDimension.height - BOTTOMBAR_HEIGHT - SCROLLBAR_WIDTH
-    }`;
+    );
+    const height = `max-height:${maxHeight}`;
+    const shadow = maxHeight !== 0 ? "box-shadow: 1px 2px 5px 2px rgb(51 51 51 / 15%);" : "";
     return `
       position: absolute;
       z-index: ${ComponentsImportance.Popover};
@@ -59,7 +62,7 @@ export class Popover extends Component<Props, SpreadsheetChildEnv> {
       width:${this.props.childWidth}px;
       overflow-y: auto;
       overflow-x: hidden;
-      box-shadow: 1px 2px 5px 2px rgb(51 51 51 / 15%);
+      ${shadow}
     `;
   }
 
--- a/src/components/top_bar/top_bar.ts
+++ b/src/components/top_bar/top_bar.ts
@@ -18,6 +18,7 @@ import { FullMenuItem } from "../../registries/menu_items_registry";
 import {
   Align,
   BorderCommand,
+  Pixel,
   SetDecimalStep,
   SpreadsheetChildEnv,
   Style,
@@ -84,6 +85,7 @@ interface Props {
   onClick: () => void;
   focusComposer: Omit<ComposerFocusType, "cellFocus">;
   onComposerContentFocused: (selection: ComposerSelection) => void;
+  dropdownMaxHeight: Pixel;
 }
 
 // -----------------------------------------------------------------------------
@@ -224,6 +226,8 @@ css/* scss */ `
             position: absolute;
             top: calc(100% + 5px);
             left: 0;
+            overflow-y: auto;
+            overflow-x: hidden;
             z-index: ${ComponentsImportance.Dropdown};
             box-shadow: 1px 2px 5px 2px rgba(51, 51, 51, 0.15);
             background-color: white;
@@ -291,6 +295,10 @@ export class TopBar extends Component<Props, SpreadsheetChildEnv> {
   customFormats = CUSTOM_FORMATS;
   currentFormatName = "automatic";
   fontSizes = fontSizes;
+
+  get dropdownStyle() {
+    return `max-height:${this.props.dropdownMaxHeight}px`;
+  }
 
   style: Style = {};
   state: State = useState({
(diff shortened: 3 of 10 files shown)


src/components/popover/popover.ts
```diff
@@ -47,9 +47,12 @@ export class Popover extends Component<Props, SpreadsheetChildEnv> {
     // within our control and to avoid leaking into external DOM
     const horizontalPosition = `left:${this.horizontalPosition() - this.spreadsheetPosition.x}`;
     const verticalPosition = `top:${this.verticalPosition() - this.spreadsheetPosition.y}`;
-    const height = `max-height:${
+    const max_height = Math.max(
```
hokolomopo, line 50:
```suggestion
    const maxHeight = Math.max(
```
We use camelCase for the variables

src/components/popover/popover.ts
```diff
@@ -59,7 +62,7 @@ export class Popover extends Component<Props, SpreadsheetChildEnv> {
       width:${this.props.childWidth}px;
       overflow-y: auto;
       overflow-x: hidden;
-      box-shadow: 1px 2px 5px 2px rgb(51 51 51 / 15%);
+     ${shadow};
```
hokolomopo, line 65:
```suggestion
     ${shadow}
```
The `;` is already in the `shadow`variable

src/components/popover/popover.ts
```diff
@@ -47,9 +47,12 @@ export class Popover extends Component<Props, SpreadsheetChildEnv> {
     // within our control and to avoid leaking into external DOM
     const horizontalPosition = `left:${this.horizontalPosition() - this.spreadsheetPosition.x}`;
     const verticalPosition = `top:${this.verticalPosition() - this.spreadsheetPosition.y}`;
-    const height = `max-height:${
+    const max_height = Math.max(
+      0,
       this.viewportDimension.height - BOTTOMBAR_HEIGHT - SCROLLBAR_WIDTH
-    }`;
+    );
+    const height = `max-height:${max_height}`;
```
hokolomopo, line 54: Note for forwards port: in master/16.1, the commit 9d84b33fd260132bd09799455337e843bed43213 should have fixed all these issues regarding popovers.

src/components/top_bar/top_bar.ts
```diff
@@ -292,6 +294,11 @@ export class TopBar extends Component<Props, SpreadsheetChildEnv> {
   currentFormatName = "automatic";
   fontSizes = fontSizes;
 
+  get dropdownStyle() {
+    const { height } = this.env.model.getters.getSheetViewDimension();
```
hokolomopo, line 298: I feel liks its a bit strange to have things in the topbar dedepding on the grid size but I guess it's fine for now.

src/components/color_picker/color_picker.ts
```diff
@@ -209,6 +210,11 @@ export class ColorPicker extends Component<ColorPickerProps, SpreadsheetChildEnv
     },
   });
 
+  get dropdownStyle() {
+    const { height } = this.env.model.getters.getSheetViewDimension();
```
hokolomopo, line 214: Yeah this is false, there is also color pickers in the side panels, in in this case this will be totally wrong. 
Maxheight/style should likely be given as a props.

src/components/color_picker/color_picker.ts
```diff
@@ -35,9 +35,10 @@ css/* scss */ `
     z-index: ${ComponentsImportance.ColorPicker};
     box-shadow: 1px 2px 5px 2px rgba(51, 51, 51, 0.15);
     background-color: white;
-    padding: ${PICKER_PADDING}px 0px;
```
hokolomopo, line 38: Shouldn't be deleted.

tests/components/__snapshots__/top_bar.test.ts.snap
```diff
@@ -168,6 +168,7 @@ exports[`TopBar component can set cell format 1`] = `
             </div>
             <div
               class="o-dropdown-content o-text-options  o-format-tool "
+              style="max-height:1000px"
```
hokolomopo, line 171: Add some non-snapshot test

hokolomopo (commented):
Fine work :+1: Missing some tests tho

src/components/popover/popover.ts
jash-odoo: I already check this, and same issue is also in version 16.1 and in master. This is due to popover is not updating with resizing the screen.

src/components/color_picker/color_picker.ts
jash-odoo: When we resize the screen and there isn't any place for pop-up menu, due to padding, small portion of the color picker is visible. This was also pointed out by PO during testing.

src/components/color_picker/color_picker.ts
jash-odoo: While solving this, I referred google sheet behaviour and update according to that. And about props, that's not updating with changing the screen size.

src/components/popover/popover.ts
hokolomopo: The position isn't updated, but the size should be. At least it is for me, maybe I'm missing something or its bugged. Could you show an example ?

src/components/color_picker/color_picker.ts
hokolomopo: Ok but that doesn't mean that we shoudl remove the padding at all times. We don't want to worsen the style of the ColorPicker.

What you should do is to set the padding in a `t-att-style`, and remove the padding only when the Colopicker becomes too small and the paddign becomes problematic.

src/components/popover/popover.ts
jash-odoo: [https://www.awesomescreenshot.com/video/14439047?key=915c136cc7cffaacef55c82807c18aca](url)

Check this one.

rrahir:
@robodoo delegate=hokolomopo

tests/components/__snapshots__/top_bar.test.ts.snap
jash-odoo: changes done and added test cases

src/components/popover/popover.ts
hokolomopo: In your video : 

- indeed the font size dropdown bug, it isn't a Popover but your PR should fix the problem
- the problem with the popover is that we don't update its position when we resize the window. I'm pretty sure that if we did, the update in size would be automaticall handled. The behaviour in your video is "correct" for a Popover whose position is never updated : is stay at the same position no matter the window resize, and disappear if its anchor is outside de window.

You can probably create another task to fix the issue.

src/components/color_picker/color_picker.ts
```diff
@@ -209,6 +211,17 @@ export class ColorPicker extends Component<ColorPickerProps, SpreadsheetChildEnv
     },
   });
 
+  get dropdownStyle() {
+    if (this.props.element === "topbar") {
+      const { height } = this.env.model.getters.getSheetViewDimension();
+      return `
+      padding: ${height <= 0 ? "0px 0px" : `${PICKER_PADDING}px 0px`};
```
hokolomopo, line 218:
```suggestion
      padding: ${height <= 0 ? "0px" : `${PICKER_PADDING}px 0px`};
```
nitpicking

src/components/color_picker/color_picker.ts
hokolomopo: Your props `element` is still not what I had in mind. The `ColorPicker` should have no knowledge of what is its parent. It cannot compute its maxHeight, because this would means that the `ColorPicker` knows 1) it  is in the topBar and 2) knows that the topBar is above the grid, and that a dropdown in the topbar should have maximum the size of the grid.

The `maxHeight` should be computed in the topbar, and given as props in the `ColorPicker`. That way the color picker is totally agnostic of its place in the DOM. All it knows is that it have a props `maxHeight` and should render accordingly.

jash-odoo: But the problem is in props, it won't update with changing screen size and because of that `maxHeight` will not get properly. And another point is that, Is there any way to check whether side panel is open or not?

hokolomopo: Did you look at my branch 16.0-fix-responsive-dropdown-menu-fase-test-adrm ? I gave the maxHeight as a props and it's working.

We did have problems with re-sizing the grid not triggering a re-render, but I don't think this is the problem you're having. There is 2 possible things that can happen : 
1) resizing the window cause a render. In this case the topbar is re-rendered, its t-att-style are computed, and the props given to colopicker is changer. Since its props changed, the color picker will re-render with the new props.
2) resizing the window don't cause a render. In this case, neither the t-att-style not the props will be recomputed. 

I'm 90% sure that its impossible for the change you made in the t-att-style to work, but at the same time the props to not be updated.

For your other question, I'm pretty sure it's not possible to know if the side panel is open, but opening/closing the side panel changes the grid dimension so you shouldn't need to worry about the side panel.

src/components/top_bar/top_bar.ts
hokolomopo: After discussing with the team, it did seemse more logical to give that as props, so the topbar don't have to know about the grid, but the `Spreadsheet` does. Added a commit in your branch, you can squash it with yours

tests/components/top_bar.test.ts
```diff
@@ -771,3 +771,71 @@ describe("Topbar - View", () => {
...
+    const model = new Model();
+    const { app } = await mountParent(model);
+    triggerMouseEvent('.o-tool[title="Font Size"]', "click");
+    await nextTick();
+    setTimeout(() => {
+      let height = getElComputedStyle(".o-dropdown-content.o-text-options", "maxHeight");
+      expect(parseInt(height)).toBe(
+        model.getters.getVisibleRect(model.getters.getActiveMainViewport()).height
+      );
+    }, 100);
```
hokolomopo, line 786: nonononono don't do that. The timeout will never be run, because it runs _after_ the test ends. Try to do a console.log in the timeOut to see it. Your test pass, but only because the `expect` is never run. You could throw errors and call the god of bugs in the timeout, and the test would still work.

The issue you were having is that you were missing the props `dropdownMaxHeight` in the mock topbar, at the top of this file.

jash-odoo: Update the test, please let me know if you need any changes

hokolomopo:
Fixed some last small things and quashed

@robodoo r+

hokolomopo:
@robodoo r- 

test failing

hokolomopo:
@robodoo r+

PR #1969, opened 2023-01-19
[IMP] Menu Service

hokolomopo:
## Description:

Add a menu service to handle the menus, so only one menu can be opened at a time.

Odoo task ID : [3142704](https://www.odoo.com/web#id=3142704&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/helpers/menu_service.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  getCurrentMenuId() {
+    return this.activeMenu?.id;
+  }
+
+  getCurrentMenuProps(): MenuProps | undefined {
+    const props: any = { ...this.activeMenu, depth: 1 };
+    if (!props) return undefined;
+    delete props["id"];
+    return props;
```
hokolomopo, line 38: Typing sucks

src/components/spreadsheet/spreadsheet.ts
```diff
@@ -174,6 +176,7 @@ export class Spreadsheet extends Component<SpreadsheetProps, SpreadsheetChildEnv
       toggleSidePanel: this.toggleSidePanel.bind(this),
       _t: Spreadsheet._t,
       clipboard: this.env.clipboard || instantiateClipboard(),
+      menuService: new MenuService(() => this.model.trigger("update")),
```
LucasLefevre, line 179:
```suggestion
      menuService: new MenuService(() => this.render(true)),
```
Is equivalent and makes more sense. It has nothing to do with model.
Could be optimized by taking advantage of owl reactivity

src/helpers/menu_service.ts
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  hasOpenMenu(): boolean {
+    return !!this.activeMenu;
+  }
+
+  unregisterMenu(id: string | undefined) {
+    if (id && this.activeMenu?.id === id) {
+      // call onClose after setting activeMenu to undefined to avoid possible infinite recursion
```
hokolomopo, line 60: The problem occurs if you call `mernuService.unregisterMenu()` from the callback `onClose` of the menu. I feel like it's a very easy mistake to do, and preventing it from making everything crash doesn't cost much.

src/components/grid_popover/grid_popover.ts
```diff
@@ -9,13 +10,32 @@ interface Props {
...
+      const hasPersistentPopover = this.env.model.getters.hasOpenedPopover();
+      if (isMenuOpen && hasPersistentPopover) {
+        if (!this.wasMenuOpen) {
+          this.env.model.dispatch("CLOSE_CELL_POPOVER");
+        } else {
+          this.menuService.closeActiveMenu();
+        }
+      }
+      this.wasMenuOpen = isMenuOpen;
+    });
```
hokolomopo, line 33: Maybe the persistent popover should be also registered in the menu service instead of having interaction between them and the menuService here ?

The main drawback is that the typing of the menuService would becomes more complex, and it would be harder to use it.

tests/components/figure.test.ts
```diff
@@ -423,14 +433,17 @@ describe("figures", () => {
       });
 
       test("Can open context menu on right click", async () => {
-        triggerMouseEvent(".o-figure", "contextmenu");
```
hokolomopo, line 426: The were opening the grid context menu lol

src/helpers/menu_service.ts
```diff
@@ -0,0 +1,71 @@
...
+}
+
+export class MenuService {
+  private uuidGenerator = new UuidGenerator();
+
+  private activeMenu: InternalMenu | undefined;
+
+  constructor() {}
+
+  registerMenu(menu: RegisterMenuArgs): UID {
```
LucasLefevre, line 20:
```suggestion
  openMenu(menu: RegisterMenuArgs): UID {
```

src/components/grid_popover/grid_popover.ts
```diff
@@ -9,13 +10,32 @@ interface Props {
...
+
 export class GridPopover extends Component<Props, SpreadsheetChildEnv> {
   static template = "o-spreadsheet-GridPopover";
   static components = { Popover };
 
+  menuService = useState<MenuService>(this.env.menuService);
+  private wasMenuOpen = this.menuService.hasOpenMenu();
+
+  setup() {
+    onPatched(() => {
```
LucasLefevre, line 22: `onWillUpdateProps` to avoid a useless rendering ? 
`onWillUpdateProps` is called *before* while `onPatched` is called *after*

LucasLefevre (commented):
I think a shared ("global") state is the good direction to solve this kind of problem.
It's the same problem as selection input and grid popover (currently UI plugins to share their state).

Several comments:

I wouldn't use blindly the `env` directly. For several reasons:
- it couples components to the env
- we risk conflicts with existing stuff in the env

The need to call `onWillUnmount` each time is error prone. We will definitely forget one day or another.

Now, given the two points above, a hook could solve both problems.
What do you think?

```ts

function useMenu() {
  const env = useEnv();
  const menu = env.__spreadsheet__.menuService;
  onWillUnmount(() => ...);
  return {
    open(...) {
      menu.open(...)
    },
    close() {
      ...
    },
  }
}
```

I didn't look into the details. Only the general idea, which I think I like :slightly_smiling_face:

src/helpers/menu_service.ts
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    const id = this.uuidGenerator.uuidv4();
+    const newMenu = {
+      ...menu,
+      id,
+    };
+    this.activeMenu = newMenu;
+
+    return newMenu.id;
```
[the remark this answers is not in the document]
hokolomopo: We need some kind of way to identidy the opened menu for the toggles (eg. in the bottom bar sheet).

```
if(this.contextMenuId === this.env.menuService.getCurrentMenuId()) closeMenu()
else openMenu()
```

src/components/grid_popover/grid_popover.ts
hokolomopo: Dispatching a command in `onWillUpdateProps` makes everything crash :x

PR #1970, opened 2023-01-19
[FIX] top_bar: auto selecting adjacent cells for filter

jash-odoo:
## Description:

Earlier, filter function is unable to identify adjacent available cells. To resolve this a new function created, which will help to identifying and updating the filterable zone by taking into consideration any adjacent available cells. Furthermore, if a specific zone has already been selected, the filter will only be applied to that specific zone.

Odoo task ID : [3128920](https://www.odoo.com/web#id=3128920&menu_id=4720&cids=2&action=4043&model=project.task&view_type=form)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/top_bar/top_bar.ts
+++ b/src/components/top_bar/top_bar.ts
@@ -587,6 +587,7 @@ export class TopBar extends Component<Props, SpreadsheetChildEnv> {
     if (this.cannotCreateFilter) {
       return;
     }
+    this.env.model.selection.selectTableAroundSelection();
     const sheetId = this.env.model.getters.getActiveSheetId();
     const selection = this.env.model.getters.getSelectedZones();
     interactiveAddFilter(this.env, sheetId, selection);
--- a/tests/components/top_bar.test.ts
+++ b/tests/components/top_bar.test.ts
@@ -2,6 +2,7 @@ import { Component, onMounted, onWillUnmount, useState, xml } from "@odoo/owl";
 import { ComposerFocusType } from "../../src/components/spreadsheet/spreadsheet";
 import { TopBar } from "../../src/components/top_bar/top_bar";
 import { DEFAULT_FONT_SIZE } from "../../src/constants";
+import { toZone, zoneToXc } from "../../src/helpers";
 import { Model } from "../../src/model";
 import { topbarComponentRegistry } from "../../src/registries";
 import { topbarMenuRegistry } from "../../src/registries/menus/topbar_menu_registry";
@@ -21,7 +22,7 @@ import {
   simulateClick,
   triggerMouseEvent,
 } from "../test_helpers/dom_helper";
-import { getBorder, getCell, getStyle } from "../test_helpers/getters_helpers";
+import { getBorder, getCell, getFilterTable, getStyle } from "../test_helpers/getters_helpers";
 import {
   getFigureIds,
   getNode,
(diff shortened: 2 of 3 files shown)


tests/plugins/filters.test.ts
```diff
@@ -218,6 +219,39 @@ describe("Filters plugin", () => {
       setCellContent(model, "A4", "Something");
       expect(zoneToXc(model.getters.getFilterTables(sheetId)[0].zone)).toEqual("A1:B3");
     });
+
+    test("Adjacent cells selection while applying filter on single cell", async () => {
+      const fixture = makeTestFixture();
```
pro-odoo, line 224: Do not do this. We are in a test file of a plugin. This has nothing related to UI and should not have something related to UI.
This line is use to make a test dom fixture, and is not useful in your test.
The same apply for `({ model } = await mountSpreadsheet(fixture, { model: new Model(data) }));`.
You can just instantiate your model with `const model = new Model(data);`.
If you need the UI for one of your test then your test should not belong to a plugin file test (but obviously it's not the case here).

src/helpers/ui/filter_interactive.ts
```diff
@@ -17,3 +17,58 @@ export function interactiveAddFilter(env: SpreadsheetChildEnv, sheetId: UID, tar
     env.raiseError(AddFilterInteractiveContent.nonContinuousTargets);
   }
 }
+
+export function adjecentCellsSelection(
```
pro-odoo, line 21:
```suggestion
export function adjacentCellsSelection(
```

pro-odoo: If you use VSCode, you can install this extension, it will help point up this kind of mistake: https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker

src/plugins/core/filters.ts
```diff
@@ -109,7 +110,15 @@ export class FiltersPlugin extends CorePlugin<FiltersState> implements FiltersSt
...
-        const zone = union(...cmd.target);
+        let zone = union(...cmd.target);
+        const sheetId = cmd.sheetId;
+        if (zone.top == zone.bottom && zone.left == zone.right) {
+          [zone] = adjecentCellsSelection(
+            sheetId,
+            { sheetId: sheetId, row: zone.top, col: zone.right },
+            this.getters.getCell
+          );
+        }
```
pro-odoo, line 121: Do it upstream. It's not the role of the plugin to extend the zone. It should be done in the action which trigger this command.

src/helpers/ui/filter_interactive.ts
pro-odoo, line 21: There is something that exists which does the same better: selectTableAroundSelection from selection stream processor. In addition to expand the zone, it also expand the selection to that expanded zone, which will approach the behaviour of GSheet.
Use it, directly in the function `interactiveAddFilter`.
You can check the action `CREATE_CHART` to see an exemple.

pro-odoo (changes requested):

pro-odoo:
This PR targets master, the tag in Odoo is 16.0.

tests/plugins/filters.test.ts
```diff
@@ -218,6 +218,39 @@ describe("Filters plugin", () => {
       setCellContent(model, "A4", "Something");
       expect(zoneToXc(model.getters.getFilterTables(sheetId)[0].zone)).toEqual("A1:B3");
     });
+
+    test("Adjacent cells selection while applying filter on single cell", async () => {
```
pro-odoo, line 222: This test only test that your helper (`createFilter`) is correct. If you comment your change in the file `top_bar.ts`, this test will still pass.
What you can do is to test the UI (so in another file, see my previous comment). You can check the file `top_bar.test.ts`, there are some tests of the filter tool. So you can write a test, with the same data as in this test, start with `selectCell(model, "A1")` and then click on the button, and check two things: 
- Your selection is updated
- The created filter has the correct zone.

pro-odoo (changes requested):
This solution is much better 👍  Thanks !

tests/plugins/filters.test.ts
jash-odoo: Added the new test, let me know if you need any changes

pro-odoo: So this test can be removed, it's already test in the newly added test

jash-odoo: Remove the old test

pro-odoo (approved):
robodoo r+
Thanks ! :)

PR #1983, opened 2023-01-23
[FIX] scorecard: rendering issue while resizing

hokolomopo:
## Description

Since 5514296, the rendering of the scorecard was bugging a bit during resizing, the title was blinking in and out.

This was because the size of the scorecard was updated during the `useEffect()` which changed the state of the chart, thus we had to wait for the next render to have a correct value. 

In this commit we removed the borders in the `o-figure` element. This means that we can use the size of the figure props in scorecards.

The figure borders are now set in the `o-figure-border` overlay above the figure. This mean that 1px in the chart is now covered by the border, which shouldn't be a big problem.

Odoo task ID : [3130899](https://www.odoo.com/web#id=3130899&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/tests/components/figure.test.ts
+++ b/tests/components/figure.test.ts
@@ -1,6 +1,12 @@
 import { App, Component, xml } from "@odoo/owl";
 import { Model, Spreadsheet } from "../../src";
-import { DEFAULT_CELL_HEIGHT, DEFAULT_CELL_WIDTH, MIN_FIG_SIZE } from "../../src/constants";
+import {
+  DEFAULT_CELL_HEIGHT,
+  DEFAULT_CELL_WIDTH,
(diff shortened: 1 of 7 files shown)


tests/components/figure.test.ts
```diff
@@ -410,17 +421,32 @@ describe("figures", () => {
     expect(figure.classList).not.toContain("o-dragging");
   });
 
-  test("Figure border disabled on dashboard mode", async () => {
```
hokolomopo, line 413: Note: add tests for Images in FWP

rrahir:
@robodoo r+

PR #1984, opened 2023-01-23
[FIX] menu: use capture for events

hokolomopo:
## Description

Before this, the menu were closed by a listener on the onClick of window.
This caused 2 main problems :

- on some places, we stopped the events propagation, which prevented the
menu from closing. This was the case for the topbar composer for example.

- on buttons that opened a menu we needed to add a `.stop`, because if
another menu was already open the menu would instantly close after
changing its props.

Solved this by having the externalListener of menus listen to the `capture`
phase of events. This bypass all the `.stop`, and assures that the old
menu is closed before the new one is opened for click events.

The way it is currently handled should probably be improved when we
implement a proper menu service (`querySelectorAll` on the menus and
we need to customize event payload for menu toggle to work).

Odoo task ID : [3145756](https://www.odoo.com/web#id=3145756&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/bottom_bar/bottom_bar.ts
+++ b/src/components/bottom_bar/bottom_bar.ts
@@ -3,7 +3,7 @@ import { BACKGROUND_GRAY_COLOR, BOTTOMBAR_HEIGHT, HEADER_WIDTH } from "../../con
 import { formatValue } from "../../helpers/format";
 import { interactiveRenameSheet } from "../../helpers/ui/sheet_interactive";
 import { MenuItemRegistry, sheetMenuRegistry } from "../../registries/index";
-import { Pixel, SpreadsheetChildEnv, UID } from "../../types";
+import { MenuMouseEvent, Pixel, SpreadsheetChildEnv, UID } from "../../types";
 import { css } from "../helpers/css";
 import { Menu, MenuState } from "../menu/menu";
 
@@ -106,13 +106,22 @@ interface Props {
   onClick: () => void;
 }
 
+interface BottomBarMenuState extends MenuState {
+  menuId: UID | undefined;
+}
+
 export class BottomBar extends Component<Props, SpreadsheetChildEnv> {
   static template = "o-spreadsheet-BottomBar";
   static components = { Menu };
 
   private bottomBarRef = useRef("bottomBar");
 
-  menuState: MenuState = useState({ isOpen: false, position: null, menuItems: [] });
+  menuState = useState<BottomBarMenuState>({
+    isOpen: false,
+    menuId: undefined,
+    position: null,
+    menuItems: [],
+  });
   selectedStatisticFn: string = "";
 
   setup() {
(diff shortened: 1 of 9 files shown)


src/components/bottom_bar/bottom_bar.xml
```diff
@@ -13,7 +13,7 @@
           <t t-call="o-spreadsheet-Icon.PLUS"/>
         </div>
       </Ripple>
-      <Ripple>
```
hokolomopo, line 16: Not sure if it's the "correct" way to fix the problem. I've though of some alternatives, but I'm nort sure what's the best.

1. Stop the event. This works very well, but is easy to forget to do, and what if we also want to handle the event in the bottom bar?
2. Add custom payload to the click event, so `Menu` know that it should close itself. Also easy to forget, and need a bit of typing.
3. Add the "parent element" in the props of the `Menu`. 
4. Make sure to always unmount then recreate a `Menu`. Not owly at all.

I like 2)/3) the best I think, but 1) is also very simple and have no problem for the moment.

Thoughts ?

pro-odoo: 5. Use `capture: true` in the event listener of Menu.

tests/components/top_bar.test.ts
```diff
@@ -108,6 +108,21 @@ describe("TopBar component", () => {
     app.destroy();
   });
 
+  test("Menu should be closed while clicking on composer", async () => {
+    const { app } = await mountParent();
+
+    expect(fixture.querySelectorAll(".o-menu").length).toBe(0);
+    fixture.querySelector(".o-topbar-menu[data-id='file']")!.dispatchEvent(new Event("click"));
```
pro-odoo, line 115:
```suggestion
    fixture.querySelector(".o-topbar-menu[data-id='file']")!.dispatchEvent(new Event("click", { bubbles: true ));
```
Or wait for my PR ;-)

src/components/helpers/dom_helpers.ts
```diff
@@ -14,3 +14,7 @@ export function gridOverlayPosition() {
   }
   throw new Error("Can't find spreadsheet position");
 }
+
+export function getOpenedMenus(): HTMLElement[] {
+  return Array.from(document.querySelectorAll<HTMLElement>(".o-menu"));
```
pro-odoo, line 19:
```suggestion
  return Array.from(document.querySelectorAll<HTMLElement>(".o-spreadsheet .o-menu"));
```
Just to be sure to avoid conflict with the rest of the world

src/components/composer/composer/composer.xml
```diff
@@ -12,7 +12,7 @@
         t-on-mousedown="onMousedown"
         t-on-input="onInput"
         t-on-keyup="onKeyup"
-        t-on-click.stop="onClick"
+        t-on-click="onClick"
```
pro-odoo, line 15: 🥳

src/components/bottom_bar/bottom_bar.ts
```diff
@@ -178,32 +187,34 @@ export class BottomBar extends Component<Props, SpreadsheetChildEnv> {
...
       const { top, left } = target.getBoundingClientRect();
-      this.openContextMenu(left, top, sheetMenuRegistry);
+      this.openContextMenu(left, top, sheetMenuRegistry, sheetId);
     }
   }
 
-  onContextMenu(sheet: string, ev: MouseEvent) {
-    if (this.env.model.getters.getActiveSheetId() !== sheet) {
-      this.activateSheet(sheet);
+  onContextMenu(sheetId: string, ev: MouseEvent) {
```
pro-odoo, line 211:
```suggestion
  onContextMenu(sheetId: UID, ev: MouseEvent) {
```

src/components/menu/menu.ts
```diff
@@ -162,22 +163,24 @@ export class Menu extends Component<Props, SpreadsheetChildEnv> {
     return this.menuComponentHeight(menusAbove) + this.position.y;
   }
 
-  private onClick(ev: MouseEvent) {
+  private onClick(ev: MenuMouseEvent) {
```
pro-odoo, line 166: If I do not miss something, the two functions (onClick and onContextMenu) are the same functions.

tests/components/bottom_bar.test.ts
```diff
@@ -142,6 +142,39 @@ describe("BottomBar component", () => {
...
+    expect(fixture.querySelectorAll(".o-menu")).toHaveLength(1);
+    expect(fixture.querySelector(".o-menu-item")!.textContent).toEqual("Duplicate");
+    app.destroy();
+  });
+
+  test("Can open list of sheet menu if another menu is already open", async () => {
+    const { app } = await mountTopBar();
+    expect(fixture.querySelectorAll(".o-menu")).toHaveLength(0);
+
+    triggerMouseEvent(".o-sheet-icon", "click");
```
pro-odoo, line 166: use `click` :D

pro-odoo (commented):
Seems good to me, will ask @rrahir for one more opinion.

tests/components/top_bar.test.ts
hokolomopo: change to onClick in FWP

rrahir (approved):

rrahir:
@robodoo r+

PR #1985, opened 2023-01-23
[FIX] popover: preserve scroll in popover

rrahir:
## Description:

 Currently, each new rendering will reset the popover style which will
    also reset its scrollTop(or Left) property. This is an issue that can
    already be noticed when opening a menu too big for the screen. when the
    said menu will be rerendered (for instance when opening one of the
    submenus), the scroll is resetted and the mouse is now targetting
    another submenu, making the first one unreachable.

- fix menu toggling
The 'toggling' of the menus was not really doing what it said because it
    would always end up setting up new menu items and would never close the
    menus. The functional behaviour was still good though be only because
    the menu component has an external listener that will make it close when
    we click outside the menu.

    This revision introduces a real toggling which will spare some useless
    instanciations of `Menu` and renderings.

[FIX] menu: reduce redundant calls to getter

    The getter `popover` was called independantly 3 time, hence 3
    computations but all of them take place to feed the props of the same
    component.

Odoo task ID : [3146014](https://www.odoo.com/web#id=3146014&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


pro-odoo (approved):
robodoo rebase-ff r+

PR #1986, opened 2023-01-24
[IMP] Header: auto-resize of row height

yunchen4:
## Description:

This PR implements a simple auto-resize of row height for cells with multiline text which only takes the number of lines into consideration (no wrapping mode, etc due to performance consideration and code structure). 

This PR also fixes the auto-resize of row height for empty cells. Now the default height of empty cells will be `DEFAULT_CELL_HEIGHT` instead of the height computed from the font size. 

Odoo task ID : [3095603](https://www.odoo.com/web#id=3095603&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/misc.ts
+++ b/src/helpers/misc.ts
@@ -2,6 +2,7 @@
 // Miscellaneous
 //------------------------------------------------------------------------------
 import {
+  DEFAULT_CELL_HEIGHT,
   DEFAULT_FONT,
   DEFAULT_FONT_SIZE,
   DEFAULT_FONT_WEIGHT,
@@ -11,7 +12,7 @@ import {
   PADDING_AUTORESIZE_VERTICAL,
 } from "../constants";
 import { fontSizeMap } from "../fonts";
-import { ConsecutiveIndexes, Lazy, Style, UID } from "../types";
+import { Cell, ConsecutiveIndexes, Lazy, Style, UID } from "../types";
 import { Cloneable, Pixel } from "./../types/misc";
 import { parseDateTime } from "./dates";
 /**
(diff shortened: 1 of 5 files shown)


src/helpers/misc.ts
```diff
@@ -124,10 +125,13 @@ export function computeTextLinesHeight(textLineHeight: number, numberOfLines: nu
...
-export function getDefaultCellHeight(style: Style | undefined): Pixel {
-  // TO DO: take multi text line into account to compute the real cell height in case of wrapping cell
-  const fontSize = computeTextFontSizeInPixels(style);
-  return computeTextLinesHeight(fontSize) + 2 * PADDING_AUTORESIZE_VERTICAL;
+export function getDefaultCellHeight(cell: Cell | undefined): Pixel {
+  if (!cell || !cell.content) {
+    return DEFAULT_CELL_HEIGHT;
+  }
+  const fontSize = computeTextFontSizeInPixels(cell.style);
+  const multiLineText = cell.content.split("\n");
```
hokolomopo, line 133:
```suggestion
  const multiLineText = cell.content.split(NEWLINE);
```
we have a NEWLINE constant somewhere

hokolomopo (commented):
This is nice when you type some multi line text in the composer, but we should have a follow-up task for every other case that isn't handled (wrapping/cell evaluation to empty string or to multi-line text) But those cases are way more annoying to handle

rrahir (changes requested):
Could you put a reference to the task you created to address the more global situation as adrien referred. The task number could be set in the commit message.

tests/plugins/sheetview.test.ts
```diff
@@ -868,6 +868,9 @@ describe("Viewport of Simple sheet", () => {
       top: 0,
     });
     setStyle(model, "A1:A20", { fontSize: 36 });
+    for (let i = 1; i <= 20; ++i) {
+      setCellContent(model, "A" + i, "test"); // empty cell with bigger font size will not cause viewport change
```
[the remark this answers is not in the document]
yunchen4: @rrahir Done 🙂

rrahir:
tests are failing :/

yunchen4:
> tests are failing :/

@rrahir Should be fine now 🥲

tests/collaborative/collaborative.test.ts
```diff
@@ -997,15 +997,16 @@ describe("Multi users synchronisation", () => {
...
       setStyle(bob, "A1", { fontSize: 36 });
+      setCellContent(bob, "A1", "text");
       charlie.dispatch("DUPLICATE_SHEET", {
         sheetId: firstSheetId,
         sheetIdTo: "sheet2",
       });
       charlie.dispatch("DELETE_SHEET", { sheetId: firstSheetId });
     });
-    expect([alice, bob, charlie]).toHaveSynchronizedValue(
+    expect([bob, charlie]).toHaveSynchronizedValue(
```
rrahir, line 1007: you cannot remove Alice here- the very goal of the tests is to make sure that every user has synchronized values ;-)

yunchen4: > you cannot remove Alice here- the very goal of the tests is to make sure that every user has synchronized values ;-)

@rrahir Change done. I thought it was enough to use `alice` to calculate the cell height to show `alice` also have this synchronized value

rrahir: in this very example it is, but it's better to keep a standard way to test the synchronization because makein assumptions could be problematic in the future if we were to change the assertion tested :)

rrahir:
@robodoo r+

PR #1988, opened 2023-01-24
[FIX] selection: avoid scroll when using ctrl+a loop

hokolomopo:
## Description

When the selection is updated using the ctrl+a shortcut (or when inserting a chart), the sheet shouldn't be scrolled to try to display the new selection.

Odoo task ID : [3068981](https://www.odoo.com/web#id=3068981&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/selection_stream/selection_stream_processor.ts
+++ b/src/selection_stream/selection_stream_processor.ts
@@ -53,13 +53,13 @@ interface SelectionProcessor {
   selectTableAroundSelection(): DispatchResult;
 }
 
+export type SelectionStreamProcessor = SelectionProcessor &
+  StatefulStream<SelectionEvent, AnchorZone>;
 /**
  * Processes all selection updates (usually from user inputs) and emits an event
  * with the new selected anchor
  */
-export class SelectionStreamProcessor
-  implements SelectionProcessor, StatefulStream<SelectionEvent, AnchorZone>
-{
+export class SelectionStreamProcessorImpl implements SelectionStreamProcessor {
   private stream: EventStream<SelectionEvent>;
   /**
    * "Active" anchor used as a reference to compute new anchors
--- a/src/selection_stream/selection_stream_processor.ts
+++ b/src/selection_stream/selection_stream_processor.ts
@@ -33,6 +33,7 @@ type StatefulStream<Event, State> = {
   resetAnchor: (owner: unknown, state: State) => void;
   observe: (owner: unknown, callbacks: StreamCallbacks<Event>) => void;
   release: (owner: unknown) => void;
+  getBackToDefault(): void;
 };
 
 /**
@@ -126,9 +127,14 @@ export class SelectionStreamProcessor
   /**
    * Select a new anchor
    */
-  selectZone(
+  selectZone(anchor: AnchorZone): DispatchResult {
+    return this.modifyAnchor(anchor, "overrideSelection", "ZonesSelected");
+  }
+
+  private modifyAnchor(
     anchor: AnchorZone,
-    mode: SelectionEvent["mode"] = "overrideSelection"
+    mode: SelectionEvent["mode"],
+    eventType: SelectionEvent["type"]
   ): DispatchResult {
     const sheetId = this.getters.getActiveSheetId();
     anchor = {
@@ -136,7 +142,7 @@ export class SelectionStreamProcessor
       zone: this.getters.expandZone(sheetId, anchor.zone),
     };
     return this.processEvent({
-      type: "ZonesSelected",
+      type: eventType,
       anchor,
       mode,
     });
@@ -177,7 +183,7 @@ export class SelectionStreamProcessor
     const expandedZone = this.getters.expandZone(sheetId, zone);
     const anchor = { zone: expandedZone, cell: { col: anchorCol, row: anchorRow } };
     return this.processEvent({
-      type: "AlterZoneCorner",
+      type: "AlterZone",
       mode: "updateAnchor",
       anchor: anchor,
     });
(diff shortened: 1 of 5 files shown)


src/selection_stream/selection_stream_processor.ts
```diff
@@ -338,13 +339,17 @@ export class SelectionStreamProcessor
...
+        "overrideSelection",
+        "AlterZoneCorner"
+      );
     }
 
     const tableZone = this.expandZoneToTable(anchor.zone);
 
     return !deepEquals(tableZone, anchor.zone)
-      ? this.selectZone({ ...anchor, zone: tableZone })
+      ? this.selectZone({ ...anchor, zone: tableZone }, "overrideSelection", "AlterZoneCorner")
```
hokolomopo, line 352: The only differences between the event type `AlterZoneCorner` and `ZonesSelected` is the handling in the `SheetView` plugin (`AlterZoneCorner` doesn't scroll). So I'm pretty sure that's what we want to use.

However now the name `AlterZoneCorner` is kinda bad, we change the whole selection zone here and not its corner. Maybe change it to just `AlterZone` ? Still its confusing that the difference between types `AlterZone` and `ZonesSelected` is how the `SheetView` plugin react to them...

src/selection_stream/selection_stream_processor.ts
```diff
@@ -128,15 +128,16 @@ export class SelectionStreamProcessor
    */
   selectZone(
     anchor: AnchorZone,
-    mode: SelectionEvent["mode"] = "overrideSelection"
+    mode: SelectionEvent["mode"] = "overrideSelection",
+    eventType: SelectionEvent["type"] = "ZonesSelected"
```
LucasLefevre, line 132: I don't think it's a good idea to add `eventType` to this method's API.
Especially given it's a public method but the extra argument is only used internally (from an external POV, there's no such thing as an event)

src/selection_stream/selection_stream_processor.ts
LucasLefevre: I agree with the renaming `AlterZoneCorner` => `AlterZone`

tests/plugins/selection.test.ts
```diff
@@ -1010,57 +1012,97 @@ describe("move elements(s)", () => {
...
+    let model: Model;
+    beforeEach(() => {
+      model = new Model();
+      setCellContent(model, "A1", "a");
+      setCellContent(model, "A2", "a");
+      setCellContent(model, "A3", "a");
+      setCellContent(model, "A4", "a");
+
+      selectCell(model, "A4");
+      setViewportOffset(model, 0, DEFAULT_CELL_HEIGHT);
```
LucasLefevre, line 1081: please don't. create only empty things in before each.

It makes tests unreadable because of the noise and un-debuggable again because of the noise. Tests are strongly coupled to `beforeEach` (and are no longer that independent)
It might be ok-ish right now because there's only two tests. But people will add more and more tests, people will copy this (bad) practice, etc.

LucasLefevre (changes requested):

src/selection_stream/selection_stream_processor.ts
hokolomopo: Didn't realize the method was public, you're right it doesn't make sense to expose that

tests/plugins/selection.test.ts
hokolomopo: I mean I don't want to copy 8 lines for each tests, `beforeEach` exist for that. I can create an helper function taht does the `setCellContent` and call it at the start of each test, but that's sensibly the same as `beforeEach`

src/selection_stream/selection_stream_processor.ts
```diff
@@ -338,13 +346,17 @@ export class SelectionStreamProcessor
...
+        "overrideSelection",
+        "AlterZone"
+      );
     }
 
     const tableZone = this.expandZoneToTable(anchor.zone);
 
     return !deepEquals(tableZone, anchor.zone)
-      ? this.selectZone({ ...anchor, zone: tableZone })
+      ? this.modifyAnchor({ ...anchor, zone: tableZone }, "overrideSelection", "AlterZone")
```
rrahir, line 359:
```suggestion
      ? this.modifyAnchor({ ...anchor, zone: tableZone }, "updateAnchor", "AlterZone")
```
the concept of altering a zone while overriding the zone makes no sense imho, this change does not seem to break anything

src/selection_stream/selection_stream_processor.ts
```diff
@@ -355,7 +367,7 @@ export class SelectionStreamProcessor
    */
   selectTableAroundSelection(): DispatchResult {
     const tableZone = this.expandZoneToTable(this.anchor.zone);
-    return this.selectZone({ ...this.anchor, zone: tableZone }, "updateAnchor");
+    return this.modifyAnchor({ ...this.anchor, zone: tableZone }, "overrideSelection", "AlterZone");
```
rrahir, line 370:
```suggestion
    return this.modifyAnchor({ ...this.anchor, zone: tableZone }, "updateAnchor", "AlterZone");
```

same as above

rrahir (changes requested):
Could you also  update the interface of  `SelectionProcessor` and add the function `getBackToDefault`  to it ? it is missing

hokolomopo:
> Could you also update the interface of SelectionProcessor and add the function getBackToDefault to it ? it is missing

What's the point of having an interface if it isn't exported ?

rrahir:
@robodoo r+ rebase-ff

tests/plugins/selection.test.ts
LucasLefevre: > I mean I don't want to copy 8 lines for each tests, beforeEach exist for that. I can create an helper function taht does the setCellContent and call it at the start of each test, but that's sensibly the same as beforeEach

I agree with you it's sensibly the same if you add a getter and use it everywhere. It's still slightly better since it won't "force" the `beforeEach` setup for all future tests if some are to be added here.)

However, what I may have missed to explain properly in my original comment is that test setup should be minimal for the feature under test. No data that doesn't have a real purpose. Do you need *4 cells* to show the feature is working here ? It feels like 2 cells are enough.

The reasons are still the same: no useless noise to read, to debug, to understand what the test is testing

PR #2001, opened 2023-01-26
[FIX] composer: (real) support of IME

rrahir:
The previous attempt to support IME was only tackling a small part of the problem. Typing a character in the grid (not in an open composer) was not capturing the modified input. This specific issue was solved but it didn't tackle the main issue. Writing any character will automatically close the IME. The problem occurs because we never trust the contenteditable helper value. Each rendering, we empty it and reinstert everything, the deletion causes the IME to dissapear.

This revision skips the deletion/reinsertion in the contentEditableHelper whenever the composition is active.

This leaves one issue - when inputing a character in the grid, we open a new composer and give it focus, which obviously closes the IME. This issue needs a refactoring to be addressed - is will be done in version 16.2 or after.

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/composer/composer.ts
+++ b/src/components/composer/composer.ts
@@ -74,6 +74,8 @@ const TEMPLATE = xml/* xml */ `
     t-on-click.stop="onClick"
     t-on-blur="onBlur"
     t-on-paste.stop=""
+    t-on-compositionstart="onCompositionStart"
+    t-on-compositionend="onCompositionEnd"
   />
 
   <div t-if="props.focus !== 'inactive' and (autoCompleteState.showProvider or functionDescriptionState.showDescription)"
@@ -197,6 +199,7 @@ export class Composer extends Component<Props, SpreadsheetEnv> {
     argToFocus: 0,
   });
   private isKeyStillDown: boolean = false;
+  private compositionActive: boolean = false;
 
   get assistantStyle(): string {
     if (this.props.delimitation && this.props.rect) {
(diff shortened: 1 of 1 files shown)


src/components/composer/composer.ts
```diff
@@ -333,6 +336,13 @@ export class Composer extends Component<Props, SpreadsheetEnv> {
     this.dispatch("STOP_EDITION", { cancel: true });
   }
 
+  onCompositionstart() {
+    this.compositionActive = true;
+  }
+  onCompositionend() {
```
LucasLefevre, line 342: spell checker is yelling at me: "Compositionstart" isn't a known word
```suggestion
  onCompositionStart() {
    this.compositionActive = true;
  }
  onCompositionEnd() {
```

src/components/composer/composer.ts
```diff
@@ -74,6 +74,8 @@ const TEMPLATE = xml/* xml */ `
     t-on-click.stop="onClick"
     t-on-blur="onBlur"
     t-on-paste.stop=""
+    t-on-compositionstart="onCompositionstart"
+    t-on-compositionend="onCompositionend"
```
LucasLefevre, line 78: not sure what happens but those events are never fired on my machine.
Is it browser/os dependent?

rrahir: only fired when you use an IME, if you don't have one, you won't be able to test this

rrahir: See https://developer.mozilla.org/en-US/docs/Web/API/Element/compositionend_event  and https://developer.mozilla.org/en-US/docs/Web/API/InputEvent/isComposing

LucasLefevre: Ah ok :+1: 
If it breaks something, it will break something already broken :laughing:

rrahir: taht's the idea  :p

rrahir:
@LucasLefevre It's updated

LucasLefevre (approved):
robodoo r+

PR #2013, opened 2023-01-30
[FIX] figures: copy paste through context menu not working in first try

yunchen4:
## Description:

The problem is that when users first open a spreadsheet and use context menu to copy and paste figures directly, it doesn't work. If users use keyboard shortcuts it works as normal, and after that the context menu works well, too.

The root cause seems to be the figure content hasn't been written into the clipboard when users use context menu first. Since the top bar menu works well, I use the same idea as the copy menu item in top bar menu. The reason to not reuse `ACTIONS.COPY_ACTION`/`ACTIONS.CUT_ACTION` in figure context menu is that we need to select the figure first, otherwise context menu through right-click will target at wrong selected region. 

Odoo task ID : [3159756](https://www.odoo.com/web#id=3159756&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/figures/figure_chart/figure_chart.ts
+++ b/src/components/figures/figure_chart/figure_chart.ts
@@ -50,7 +50,7 @@ export class ChartFigure extends Component<Props, SpreadsheetChildEnv> {
       action: async () => {
         this.env.model.dispatch("SELECT_FIGURE", { id: this.props.figure.id });
         this.env.model.dispatch("COPY");
-        await this.env.clipboard.clear();
+        await this.env.clipboard.write(this.env.model.getters.getClipboardContent());
       },
     });
     registry.add("cut", {
@@ -59,7 +59,7 @@ export class ChartFigure extends Component<Props, SpreadsheetChildEnv> {
       action: async () => {
         this.env.model.dispatch("SELECT_FIGURE", { id: this.props.figure.id });
         this.env.model.dispatch("CUT");
-        await this.env.clipboard.clear();
+        await this.env.clipboard.write(this.env.model.getters.getClipboardContent());
       },
     });
     registry.add("delete", {
--- a/src/components/figures/figure_image/figure_image.ts
+++ b/src/components/figures/figure_image/figure_image.ts
@@ -31,7 +31,7 @@ export class ImageFigure extends Component<Props, SpreadsheetChildEnv> {
       action: async () => {
         this.env.model.dispatch("SELECT_FIGURE", { id: this.figureId });
         this.env.model.dispatch("COPY");
-        await this.env.clipboard.clear();
+        await this.env.clipboard.write(this.env.model.getters.getClipboardContent());
       },
     });
     registry.add("cut", {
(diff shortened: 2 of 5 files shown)


src/components/figures/figure_chart/figure_chart.ts
```diff
@@ -50,7 +50,7 @@ export class ChartFigure extends Component<Props, SpreadsheetChildEnv> {
       action: async () => {
         this.env.model.dispatch("SELECT_FIGURE", { id: this.props.figure.id });
         this.env.model.dispatch("COPY");
-        await this.env.clipboard.clear();
```
yunchen4, line 53: TBH I don't know why we need `clear()`here, so I deleted it.

yunchen4:
Other menus are in `registries/menus` folder. Is it a good idea to move two figure menus into this folder?

hokolomopo:
> Other menus are in `registries/menus` folder. Is it a good idea to move two figure menus into this folder?

The difference is that the others the registries in `registries/menus` are static, and for figures they are generated on the fly because they depend on `props.figureId`. 

But if you want you can create a function that will generate them in `registries/menus`. If you do that that would be a good time to avoid having almost exactly duplicate code for image and charts

src/components/figures/figure_chart/figure_chart.ts
hokolomopo: It's supposed to do the same thig you do (` getters.getClipboardContent()` is empty when copying charts), but it seems that the implementation of `clipboard.clear()` doesn't work. 

Well it make more sense to write the content of our internal clipboard rather than clear it so it was the same value as our empty clipboard (\t) You can probaly entierly remove `.clear` from the clipboard in the env, I don't think we were using it elsewhere.

hokolomopo (commented):
Your fix works, but probably not for the reasons you think it works (and thus your commit message is a bit wrong).

How the paste works in the grid works is that we will compare the values of the getters.clipboardContent() and the navigator's clipboard (env.clipboard). If the values are different, we will paste the value of env.clipboard().

` getters.clipboardContent()` is empty (\t) when copying the charts, so it's not a problem that the charts isn't in the clipboard, it's a problem that the content of the internal clipboard and env.clipboard aren't synchronized. And this happens because : 

1) the implementation of clipbaord.clear() straight up doesn't work.
2) we don't want to clear it, but to write the empty clipboard value (\t)

So your fix is fine, just modify the comit message :) 

Maybe also add a test to check if env.clipboard has the same value as getters.clipboardContent() (env.clipboard is mocked in the tests)

rrahir:
Could you update your commit message so that it contains the full comments of Adrien on why the fix was required ?

your message is too much about the reflexion you had rather than why the code wasn't working and what the solution is. In this case, it's more relevant to mention that we were not doing what we were supposed to (clear was not the way to go) and that it caused a desynchronisation bewteen the clipboard plugin and the  env.clipoboard.

I think Lucas already metionned it but commit messages should generally have a limit of 80 characters per line :)

rrahir:
@robodoo r+

PR #1803, opened 2022-11-10
Master keyboards rar pro

pro-odoo:
## Description:

description of this task, what is implemented and why it is implemented that way.

Odoo task ID : [TASK_ID](https://www.odoo.com/web#id=TASK_ID&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] feature is organized in plugin, or UI components
- [ ] support of duplicate sheet (deep copy)
- [ ] in model/core: ranges are Range object, and can be adapted (adaptRanges)
- [ ] in model/UI: ranges are strings (to show the user)
- [ ] undo-able commands (uses this.history.update)
- [ ] multiuser-able commands (has inverse commands and transformations where needed)
- [ ] new/updated/removed commands are documented
- [ ] exportable in excel
- [ ] translations (\_lt("qmsdf %s", abc))
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes
- [ ] doc is rebuild (npm run doc)
- [ ] status is correct in Odoo

PR #2016, opened 2023-01-30
[IMP] tests: Speed up composer tests

rrahir:
## Description:

Currently, the file `composer.test.ts` takes a lot of time to execute
(43 seconds average). The reason is multi-fold:
1. There are 281 tests
2. Each test instanciate a full spreadsheet, lots of computations that
are not necessary except for ftesting the interactions of the composer
with the spreadsheet component (they currently are strongly linked).

With this revision, the tests covering the interactions with other
components are moved to a dedicated test file, which allows to test the
composer specific behaviour with a more minimalistic setup.

Note that
Benchmark
---------

`test -- composer.test.ts`
before: 43 seconds
after:  9 seconds

`test -- components`
before: 55 seconds
after: 40 seconds

Task 3163569

Odoo task ID : [3163569](https://www.odoo.com/web#id=3163569&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/tests/components/__snapshots__/top_bar.test.ts.snap
+++ b/tests/components/__snapshots__/top_bar.test.ts.snap
@@ -504,7 +504,7 @@ exports[`TopBar component can set cell format 1`] = `
         </div>
         <div
           class="o-topbar-composer bg-white"
-          style="border-color:#3266ca;"
+          style="border-color:#E0E2E4; border-right:none;"
         >
           <div
             class="o-composer-container w-100 h-100"
@@ -513,7 +513,7 @@ exports[`TopBar component can set cell format 1`] = `
               class="o-composer w-100 text-start"
               contenteditable="true"
               spellcheck="false"
-              style="padding:5px 0px 5px 8px; max-height:100px; line-height:24px; height:fit-content;"
+              style="padding:5px 0px 5px 8px; max-height:100px; line-height:24px; height:34px;"
               tabindex="1"
             />
             
@@ -913,7 +913,7 @@ exports[`TopBar component simple rendering 1`] = `
     </div>
     <div
       class="o-topbar-composer bg-white"
-      style="border-color:#3266ca;"
+      style="border-color:#E0E2E4; border-right:none;"
     >
       <div
         class="o-composer-container w-100 h-100"
@@ -922,7 +922,7 @@ exports[`TopBar component simple rendering 1`] = `
           class="o-composer w-100 text-start"
           contenteditable="true"
           spellcheck="false"
-          style="padding:5px 0px 5px 8px; max-height:100px; line-height:24px; height:fit-content;"
+          style="padding:5px 0px 5px 8px; max-height:100px; line-height:24px; height:34px;"
           tabindex="1"
         />
         
--- a/tests/components/composer.test.ts
+++ b/tests/components/composer.test.ts
@@ -1,54 +1,38 @@
-import { App } from "@odoo/owl";
+import { App, Component, onMounted, onWillUnmount, useState, xml } from "@odoo/owl";
 import {
+  Composer,
+  ComposerProps,
   selectionIndicatorClass,
   tokenColors,
 } from "../../src/components/composer/composer/composer";
-import { DEFAULT_CELL_HEIGHT, DEFAULT_CELL_WIDTH, HEADER_HEIGHT } from "../../src/constants";
-import { fontSizes } from "../../src/fonts";
-import { colors, toHex, toZone } from "../../src/helpers/index";
+import { ComposerFocusType } from "../../src/components/spreadsheet/spreadsheet";
+import { colors, toZone } from "../../src/helpers/index";
 import { Model } from "../../src/model";
-import { Highlight } from "../../src/types";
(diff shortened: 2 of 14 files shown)


tests/components/composer.test.ts
```diff
@@ -1339,154 +868,157 @@ describe("composer", () => {
...
-    await keyDown("F4");
-    expect(model.getters.getCurrentContent()).toBe("=$A$1");
-    await keyDown("F4");
-    expect(model.getters.getCurrentContent()).toBe("=A$1");
-    await keyDown("F4");
-    expect(model.getters.getCurrentContent()).toBe("=$A1");
-    await keyDown("F4");
-    expect(model.getters.getCurrentContent()).toBe("=A1");
-  });
+  describe("Loop through reference combinations (AKA F4 shortcut)", () => {
```
LucasLefevre, line 884: all those tests could be converted to plugin tests.
Except one to test the integration with the plugin. (there's really only one line in the composer)

tests/components/composer.test.ts
```diff
@@ -1594,13 +1106,9 @@ describe("composer highlights color", () => {
     expect(getHighlights(model).length).toBe(2);
     expect(getHighlights(model)[0].color).toBe(colors[0]);
     expect(getHighlights(model)[1].color).toBe(colors[1]);
+    await keyDown("Enter", { bubbles: true });
```
LucasLefevre, line 1109: it bubbles: true by default

tests/components/spreadsheet.test.ts
```diff
@@ -399,6 +434,216 @@ describe("Composer interactions", () => {
     fixture.remove();
     app.destroy();
   });
+
+  test("Selecting a range should not scroll the viewport to the current Grid selection", async () => {
```
LucasLefevre, line 438: what do you think about moving thoses composer/spreadsheet tests in their own file?

LucasLefevre (commented):
Nice job :ok_hand: 
hahaha-run

tests/components/composer.test.ts
```diff
@@ -1,89 +1,144 @@
...
 let model: Model;
 let composerEl: Element;
-let gridInputEl: Element;
-let fixture: HTMLElement;
+let fixture: HTMLDivElement;
 let app: App;
 let cehMock: ContentEditableHelper;
+let parent: Parent;
+
+jest.setTimeout(30000000);
```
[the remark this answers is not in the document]
rrahir: damn me, I'll add this shit to the pre hook I think

tests/components/__snapshots__/top_bar.test.ts.snap
```diff
@@ -513,7 +513,7 @@ exports[`TopBar component can set cell format 1`] = `
               class="o-composer w-100 text-start"
               contenteditable="true"
               spellcheck="false"
-              style="padding:5px 0px 5px 8px; max-height:100px; line-height:24px; height:fit-content;"
+              style="padding:5px 0px 5px 8px; max-height:100px; line-height:24px; height:34px;"
```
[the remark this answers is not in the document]
rrahir: Since we create the parent component of topbar without typing  the props, we were actually instanciating a top bar composer with its `focus` props as a boolean (instead of  `inactive`, `cellFocus`, `contentFocus`)  and a lot of code, specifically https://github.com/odoo/o-spreadsheet/blob/master/src/components/composer/top_bar_composer/top_bar_composer.ts#L39 ) tests against `inactive` which of course does not match a boolean. I checked that in thie specific snapshot, the composer is closed,  so the style  was wrong in the previous snapshot

tests/components/spreadsheet.test.ts
rrahir: I was thinking the same thing yes, I didn't do it because I fear that we start multiplying the test files too much but I really think it's worth in this case.

tests/components/composer.test.ts
```diff
@@ -449,195 +478,79 @@ describe("ranges and highlights", () => {
...
-    expect(composerEl.textContent).toBe("=C8");
-  });
-
-  test("type '=', select twice a cell", async () => {
-    composerEl = await typeInComposerGrid("=");
-    await clickCell(model, "C8");
-    await clickCell(model, "C8");
-    expect(composerEl.textContent).toBe("=C8");
-  });
+  // TODO: not sure the test is relevant
```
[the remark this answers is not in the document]
rrahir: Well the rationale we have with Pierre in  #1803  was that we either decide to handle a keybind (in which case, we capture and stop propagation) or  decide that we "don't care about it" .   This test is more about  " the arrowkeys will move the election inside the composer when in cntentFOcus mode" (yes the title is shit)
and the following test is basically the same thing : "when in mode selecting range for a formula", the composer handles the arrowkeys by itself.

Maybe updating the tests titles would make sense?

rrahir:
@LucasLefevre Updated the PR - still not sure about the "relevant" test

LucasLefevre (commented):
robodoo r+

PR #2023, opened 2023-02-01
[REF] parser: remove binding power

LucasLefevre:
## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/formulas/parser.ts
+++ b/src/formulas/parser.ts
@@ -1,8 +1,7 @@
 import { DEFAULT_ERROR_MESSAGE } from "../constants";
 import { parseNumber, removeStringQuotes } from "../helpers/index";
 import { _lt } from "../translation";
-import { BadExpressionError, InvalidReferenceError } from "../types/errors";
-import { UnknownFunctionError } from "./../types/errors";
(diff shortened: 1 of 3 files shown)


src/formulas/parser.ts
```diff
@@ -171,7 +171,7 @@ function parseFunctionArgs(tokens: Token[]): AST[] {
     consumeOrThrow(tokens, "COMMA", _lt("Wrong function call"));
     args.push(parseOneFunctionArg(tokens));
   }
-  consumeOrThrow(tokens, "RIGHT_PAREN", _lt("Wrong function call"));
+  consumeOrThrow(tokens, "RIGHT_PAREN", _lt("Missing closing parenthesis"));
```
pro-odoo, line 174: I think it's worth to add a test for this case and the case L144, as it does not seem to be covered

LucasLefevre: It's actually impossible to have this error message raised from here.. Because of the while condition, the next token *is* a right parenthesis. I removed the message since it's useless.
I added a test for L144

pro-odoo (approved):
robodoo rebase-ff r+

PR #2025, opened 2023-02-01
[IMP] Components: extract color picker entry component

yunchen4:
## Description:

This PR extracts the color picker and its icon (e.g. fill color, font color) into a new color picker entry component. 

Odoo task ID : [2959444](https://www.odoo.com/web#id=2959444&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/color_picker/color_picker_widget.ts
+++ b/src/components/color_picker/color_picker_widget.ts
@@ -1,0 +1,74 @@
+import { Component } from "@odoo/owl";
+import { Pixel, SpreadsheetChildEnv } from "../../types";
+import { css } from "../helpers";
+import { ColorPicker } from "./color_picker";
+
+interface Props {
+  currentColor: string | undefined;
+  toggleColorPicker: () => void;
+  showColorPicker: boolean;
+  onColorPicked: (color: string) => void;
+  icon: string;
+  dropdownDirection?: "left" | "right" | "center";
+  title?: string;
+  disabled?: boolean;
+  dropdownMaxHeight?: Pixel;
+  class?: string;
+}
+
+css/* scss */ `
+  .o-color-picker-widget {
+    display: inline-block;
+    position: relative;
+
+    .o-color-picker-button-style {
+      display: flex;
+      justify-content: center;
+      align-items: center;
+      margin: 2px;
+      padding: 3px;
+      border-radius: 2px;
+      cursor: pointer;
+      &:not([disabled]):hover {
+        background-color: rgba(0, 0, 0, 0.08);
+      }
+    }
+
+    .o-color-picker-button {
+      > span {
+        border-bottom: 4px solid;
+        height: 16px;
+        margin-top: 2px;
+      }
+
+      &[disabled] {
+        pointer-events: none;
+        opacity: 0.3;
+      }
+    }
+  }
+`;
+
+export class ColorPickerWidget extends Component<Props, SpreadsheetChildEnv> {
+  static template = "o-spreadsheet-ColorPickerWidget";
+  static components = { ColorPicker };
+
+  get iconStyle() {
+    return this.props.currentColor
+      ? `border-color: ${this.props.currentColor}`
+      : "border-bottom-style: hidden";
+  }
+}
+
+ColorPickerWidget.props = {
+  currentColor: { type: String, optional: true },
+  toggleColorPicker: Function,
+  showColorPicker: Boolean,
+  onColorPicked: Function,
+  icon: String,
+  dropdownDirection: { type: String, optional: true },
+  title: { type: String, optional: true },
+  disabled: { type: Boolean, optional: true },
+  dropdownMaxHeight: { type: Number, optional: true },
+  class: { type: String, optional: true },
+};
--- a/src/components/side_panel/chart/gauge_chart_panel/gauge_chart_design_panel.ts
+++ b/src/components/side_panel/chart/gauge_chart_panel/gauge_chart_design_panel.ts
@@ -1,10 +1,10 @@
-import { Component, useState } from "@odoo/owl";
+import { Component, useExternalListener, useState } from "@odoo/owl";
 import { deepCopy } from "../../../../helpers/index";
 import { GaugeChartDefinition, SectionRule } from "../../../../types/chart/gauge_chart";
 import { CommandResult, DispatchResult, SpreadsheetChildEnv, UID } from "../../../../types/index";
-import { ColorPicker } from "../../../color_picker/color_picker";
 import { css } from "../../../helpers/css";
 import { ChartTerms } from "../../../translations_terms";
+import { ColorPickerWidget } from "./../../../color_picker/color_picker_widget";
 
 css/* scss */ `
   .o-gauge-color-set {
@@ -67,12 +67,16 @@ interface PanelState {
 
 export class GaugeChartDesignPanel extends Component<Props, SpreadsheetChildEnv> {
   static template = "o-spreadsheet-GaugeChartDesignPanel";
-  static components = { ColorPicker };
+  static components = { ColorPickerWidget };
 
   private state: PanelState = useState({
     openedMenu: undefined,
     sectionRuleDispatchResult: undefined,
   });
+
+  setup() {
+    useExternalListener(window, "click", this.closeMenus);
+  }
 
   get designErrorMessages(): string[] {
     const cancelledReasons = [...(this.state.sectionRuleDispatchResult?.reasons || [])];
--- a/src/components/side_panel/chart/line_chart/line_chart_design_panel.xml
+++ b/src/components/side_panel/chart/line_chart/line_chart_design_panel.xml
@@ -1,22 +1,17 @@
 <templates>
   <t t-name="o-spreadsheet-LineChartDesignPanel" owl="1">
     <div>
-      <div class="o-section">
+      <div class="o-section o-chart-background-color">
         <div class="o-section-title">Background color</div>
-        <div class="o-with-color-picker">
-          Select a color...
-          <span
-            t-att-style="props.definition.background ? `border-color: ${props.definition.background}` : 'border-bottom-style: hidden'"
-            t-on-click.stop="toggleColorPicker">
-            <t t-call="o-spreadsheet-Icon.FILL_COLOR"/>
-          </span>
-          <ColorPicker
-            t-if="state.fillColorTool"
-            onColorPicked="(color) => this.updateBackgroundColor(color)"
-            currentColor="props.definition.background"
-            t-key="backgroundColor"
-          />
-        </div>
+        Select a color...
+        <ColorPickerWidget
+          currentColor="props.definition.background"
+          toggleColorPicker="() => this.toggleColorPicker()"
+          showColorPicker="state.fillColorTool"
+          onColorPicked="(color) => this.updateBackgroundColor(color)"
+          title="env._t('Background Color')"
+          icon="'o-spreadsheet-Icon.FILL_COLOR'"
+        />
       </div>
       <div class="o-section o-chart-title">
         <div class="o-section-title">Title</div>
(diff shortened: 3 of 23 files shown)


src/components/side_panel/chart/line_chart/line_chart_design_panel.xml
```diff
@@ -5,17 +5,12 @@
...
-            onColorPicked="(color) => this.updateBackgroundColor(color)"
-            currentColor="props.definition.background"
-            t-key="backgroundColor"
-          />
+          <t t-call="o-spreadsheet-FillColorPickerEntry">
+            <t t-set="currentColor" t-value="props.definition.background"/>
+            <t t-set="onClick" t-value="toggleColorPicker"/>
+            <t t-set="showPicker" t-value="state.fillColorTool"/>
+            <t t-set="onColorPicked" t-value="(color) => this.updateBackgroundColor(color)"/>
+          </t>
```
hokolomopo, line 13: While using a simple template works, using a dedicated component is almost always a better option IMHO. Some advantages I see right of the bat:

- readability (`<FillColorEntry >` vs `<t t-call="o-..-FillColorEntry>`, `onClick="toggleColorPicker"` vs `<t t-set="onClick" t-value="toggle"`)
- props validation, we garantee in the tests that we aren't just sending some junk variables, and that all the needed props are given
- can define css inside the component

src/components/side_panel/chart/main_chart_panel/fill_color_picker_entry.xml
```diff
@@ -0,0 +1,20 @@
+<templates>
+  <t t-name="o-spreadsheet-FillColorPickerEntry" owl="1">
+    <div class="o-tools" style="display: inline-block;">
+      <div class="o-tool o-dropdown o-with-color">
```
hokolomopo, line 4: We should definetly use a dedicated component everywhere we use a colorPicker and not only in the chart side panel. 
If there is the need for different style, just add a props `style` or `class` and use it to have different behaviour in the side panels/topbar. 

Side note but I think we want a `cursor: pointer` on hover in the side panel too.

hokolomopo (commented):
Pre-review. I think we should create a component and use it everywhere, rather than a template only used in the chart side panel

src/components/color_picker/color_picker_entry.ts
```diff
@@ -0,0 +1,49 @@
...
+      : "border-bottom-style: hidden";
+  }
+
+  get dropdownDirection() {
+    return this.props.dropdownDirection || "right";
+  }
+}
+
+ColorPickerEntry.props = {
+  currentColor: { type: String, optional: true },
```
yunchen4, line 42: This should not be `optional: true` from a semantic POV, but I don't know how to make it receive `undefined` value here.

hokolomopo: I'm pretty sure `optional: true` is the correct way to write it, `| udnefined` and `optional` are the same fro a JS POV, even if I do agree that sometimes `| undefined` is better to understand the code

src/components/side_panel/conditional_formatting/cell_is_rule_editor.xml
```diff
@@ -73,37 +73,25 @@
...
-            currentColor="rule.style.textColor"
-            t-key="textColor"
-          />
-        </div>
+        <ColorPickerEntry
+          currentColor="rule.style.textColor || '#000000'"
+          onClick="(ev) => this.toggleMenu('cellIsRule-textColor', ev)"
+          showColorPicker="state.openedMenu === 'cellIsRule-textColor'"
+          onColorPicked="(color) => this.setColor('textColor', color)"
+          tooltipText="'Text Color'"
```
hokolomopo, line 81: `tooltipText` should be translated I think. Pretty sure we didn't need to do it explicitely before because the `title` attribute is translated automatically.

src/components/color_picker/color_picker_entry.xml
```diff
@@ -0,0 +1,20 @@
+<templates>
+  <t t-name="o-spreadsheet-ColorPickerEntry" owl="1">
+    <div class="o-dropdown o-color-picker-entry">
+      <div
+        class="o-tool o-with-color"
```
hokolomopo, line 5: No guarantees that these class are defined where we use the colorPicker. Move the class definition inside the component if its shared between all the colorPickers, or make them props

src/components/color_picker/color_picker_entry.xml
```diff
@@ -0,0 +1,20 @@
...
+        t-on-click.stop="props.onClick"
+        t-att-title="props.tooltipText">
+        <span t-att-style="iconStyle">
+          <t t-call="{{props.icon}}"/>
+        </span>
+      </div>
+      <ColorPicker
+        t-if="props.showColorPicker"
```
hokolomopo, line 13: (didn't check the code too much so it may not be possible)

Can we put all the logic to show/hide the ColorPicker inside the `ColorPickerEntry` ?

src/components/color_picker/color_picker_entry.ts
```diff
@@ -0,0 +1,45 @@
...
+  .o-color-picker-entry span {
+    cursor: pointer;
+  }
+
+  .o-cf-cell-is-rule .o-color-picker-entry span {
+    cursor: default;
+  }
+`;
+
+export class ColorPickerEntry extends Component<Props, SpreadsheetChildEnv> {
```
hokolomopo, line 26: Not a fan of the name `ColorPickerEntry`, that being said I don't have any bright idea. `ColorPickerContainer` ? `ColorPickerButton` ?

hokolomopo (commented):
Some more comments, nothing blocking so I'll put the task in validation green. Have a nice Weekend !

src/components/color_picker/color_picker_entry.ts
yunchen4: Let's wait for others' ideas

src/components/color_picker/color_picker_entry.xml
```diff
@@ -0,0 +1,20 @@
+<templates>
+  <t t-name="o-spreadsheet-ColorPickerEntry" owl="1">
+    <div class="o-dropdown o-color-picker-entry" t-att-disabled="props.disabled">
```
rrahir, line 3: I think that the o-dropdown, and o-tools stuff should  not be part of this component. because the rules for o-dropdown and o-tools are defined anyways in topBar and sidePanel .

This component should only contain a div to set the title (by the way it should be translated befiore getting passed as props) and the div would contain the span etc..; 

```xml
<div clas="o-color-picker-stuff">
    <span ... />
</div>
<ColorPicker ... />
```

src/components/color_picker/color_picker_entry.ts
```diff
@@ -0,0 +1,71 @@
...
+      margin-top: 2px;
+      cursor: pointer;
+    }
+  }
+
+  .o-cf-cell-is-rule .o-color-picker-entry span {
+    cursor: default;
+  }
+`;
+
```
rrahir, line 45: there is one css rule that seems to be used for every color picker but that is not here: 
```css
      .o-with-color {
        .o-line-item:hover {
          outline: 1px solid gray;
        }
      }
```

yunchen4: @rrahir On master branch I only found class `o-line-item` is used to set borders in top bar, not with color picker. Did I miss something?

src/components/color_picker/color_picker_entry.xml
yunchen4: @rrahir For the title translation, I use a getter in `ColorPickerEntry` component to translate it. Is there any difference between translating before passing as props and translating when getting it?

rrahir: Yes :) The translation flow is such that we cannot reliably translate any variable, because we don't know yet what the term will be beforehand- the props might be anything ^^^

So how to translations work in Odoo:  We have a module that will parse all files in the repository and apply different rules : 

- for xml files: it will capture all text in standard HTML tags and attributes: so  attributes like ``title="text"`,  `<span> Text2</span>` -> text and Text2  will be captured. but not parametric stuff `t-att-title="text4"` will not be captured
- for python files: it will capture the arguments of function `_(...)`
- for js files, it will capture the arguments of function `_t(...)` and `_lt(...)`

Each captured term is then exported in the POT files, which in turn will be translated in different languages. the translations are stored in PO files.
When calling the actual translation inside the code, we will try to match the string inside the translation function (_t) against the terms that are translated in the POT/PO file and see if we find a match.

This approach means that the captured terms need to be actual string! If we try to translate a variable, then we **might** sometimes find a match, but since we  never explicitely wrote in the code that we wanted our specific term to be translated, it's more likely that we won't.

note: It supports strings with placeholders, like `I am %s years old`, which can become `j'ai %s ans`, `Ik ben %s jaar oud` , etc

src/components/side_panel/conditional_formatting/conditional_formatting.ts
```diff
@@ -330,7 +333,11 @@ interface State {
 
 export class ConditionalFormattingPanel extends Component<Props, SpreadsheetChildEnv> {
   static template = "o-spreadsheet-ConditionalFormattingPanel";
-  static components = { SelectionInput, IconPicker, ColorPicker };
+  static components = {
```
rrahir, line 336: useless diff

src/components/color_picker/color_picker_widget.ts
```diff
@@ -0,0 +1,74 @@
...
+  currentColor: string | undefined;
+  toggleColorPicker: () => void;
+  showColorPicker: boolean;
+  onColorPicked: (color: string) => void;
+  icon: string;
+  dropdownDirection?: "left" | "right" | "center";
+  tooltipText?: string;
+  disabled?: boolean;
+  dropdownMaxHeight?: Pixel;
+  styleClass?: string;
```
rrahir, line 16:
```suggestion
  class?: string;
```

src/components/color_picker/color_picker_widget.ts
```diff
@@ -0,0 +1,74 @@
...
+import { ColorPicker } from "./color_picker";
+
+interface Props {
+  currentColor: string | undefined;
+  toggleColorPicker: () => void;
+  showColorPicker: boolean;
+  onColorPicked: (color: string) => void;
+  icon: string;
+  dropdownDirection?: "left" | "right" | "center";
+  tooltipText?: string;
```
rrahir, line 13:
```suggestion
  title?: string;
```

src/components/side_panel/chart/scorecard_chart_panel/scorecard_chart_design_panel.ts
```diff
@@ -17,12 +17,16 @@ interface PanelState {
...
   static template = "o-spreadsheet-ScorecardChartDesignPanel";
-  static components = { ColorPicker };
+  static components = { ColorPickerWidget };
 
   private state: PanelState = useState({
     openedColorPicker: undefined,
   });
 
+  setup() {
+    useExternalListener(window as any, "click", this.closeMenus);
```
rrahir, line 27:
```suggestion
    useExternalListener(window, "click", this.closeMenus);
```
should be tested

src/components/side_panel/chart/gauge_chart_panel/gauge_chart_design_panel.ts
```diff
@@ -67,13 +67,17 @@ interface PanelState {
...
-  static components = { ColorPicker };
+  static components = { ColorPickerWidget };
 
   private state: PanelState = useState({
     openedMenu: undefined,
     sectionRuleDispatchResult: undefined,
   });
 
+  setup() {
+    useExternalListener(window as any, "click", this.closeMenus);
```
rrahir, line 78:
```suggestion
    useExternalListener(window, "click", this.closeMenus);
```
requires a test as well: :)

rrahir (commented):
Still some nitpicking but it looks really nice !!

rrahir:
@yunchen4 I changed a bit of css to make it more clear when the widget is disabled :)

rrahir:
@robodoo r+

PR #2033, opened 2023-02-02
[FIX] chart: add dimensions to canvas

yunchen4:
## Description:

This PR adds dimensions to chart canvas. It fixes the problem when changing chart background color, the dimensions of chart will temporarily change and be out of the figure. 

Odoo task ID : [3167230](https://www.odoo.com/web#id=3167230&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/figures/chart.ts
+++ b/src/components/figures/chart.ts
@@ -16,7 +16,7 @@ const { useRef } = hooks;
 const TEMPLATE = xml/* xml */ `
 <div class="o-chart-container">
   <div class="o-chart-menu" t-on-click="showMenu">${LIST}</div>
-  <canvas t-att-style="canvasStyle" t-ref="graphContainer"/>
+  <canvas class="w-100 h-100" t-att-style="canvasStyle" t-ref="graphContainer"/>
   <Menu t-if="menuState.isOpen"
     position="menuState.position"
     menuItems="menuState.menuItems"
--- a/tests/components/charts.test.ts
+++ b/tests/components/charts.test.ts
@@ -255,6 +255,31 @@ describe("figures", () => {
     });
   });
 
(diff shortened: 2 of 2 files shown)


src/components/figures/chart.ts
```diff
@@ -70,7 +70,11 @@ export class ChartFigure extends Component<Props, SpreadsheetEnv> {
   private position = useAbsolutePosition();
 
   get canvasStyle() {
-    return `background-color: ${this.state.background}`;
+    /**
+     * Add sizes to avoid chart out of figure boundaries
+     * when changing background color
+     */
+    return `background-color: ${this.state.background};width: 100%;height: 100%;`;
```
pro-odoo, line 77: IMHO I think it's simplier to add classes `w-100 h-100` in the template. 
`<canvas class="w-100 h-100" t-att-style="canvasStyle" t-ref="graphContainer"/>`

pro-odoo (commented):
Is the issue reproductible in a test ?

src/components/figures/chart.ts
yunchen4: @pro-odoo I didn't find a way to test it, especially when we use classes to set height and weight. We cannot get width and height via `getComputedStyle()`.

rrahir: but you can definitely test taht the class w-100 and h-100 are set in the classList :)

```javascript
querySelector("canvas").classList.contains("w-100")
querySelector("canvas").classList.contains("h-100")
``` something like that

pro-odoo (commented):
robodoo r+

PR #2045, opened 2023-02-03
[IMP/REF] tests: mount owl App in test mode & mount helper

hokolomopo:
Change the parameters of the owl `App` in the test. The `App` will now
be mounted in test mode. The main purpose is to enable props validation,
which ensure that the test is close to the reality, and avoid tests
failing for unclear reasons because a props was missing (https://github.com/odoo/o-spreadsheet/commit/d4663158d7f82af870409e2508be6fa161d1921c).

The Owl test mode is equivalent to dev mode, but without warning that
the app is unsuited for production. Documentation here :
https://github.com/odoo/owl/blob/master/doc/reference/app.md#dev-mode

Also added an helper `mountComponent` and modified the existing helper
`mountSpreadsheet`. The helper now creates and returns the fixture in which
the component is mounted. They also register cleanup functions to remove
the fixture and destroy the app, so we don't have to do it manually.
## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/tests/components/bottom_bar.test.ts
+++ b/tests/components/bottom_bar.test.ts
@@ -1,8 +1,7 @@
-import { App, Component, onMounted, onWillUnmount, xml } from "@odoo/owl";
+import { Component, onMounted, onWillUnmount, xml } from "@odoo/owl";
 import { BottomBar } from "../../src/components/bottom_bar/bottom_bar";
 import { Model } from "../../src/model";
 import { SpreadsheetChildEnv } from "../../src/types";
-import { OWL_TEMPLATES } from "../setup/jest.setup";
 import {
   activateSheet,
   createSheet,
@@ -11,7 +10,7 @@ import {
   setCellContent,
 } from "../test_helpers/commands_helpers";
 import { click, keyDown, simulateClick, triggerMouseEvent } from "../test_helpers/dom_helper";
-import { makeTestEnv, makeTestFixture, mockUuidV4To, nextTick } from "../test_helpers/helpers";
+import { mockUuidV4To, mountComponent, nextTick } from "../test_helpers/helpers";
 jest.mock("../../src/helpers/uuid", () => require("../__mocks__/uuid"));
 
 let fixture: HTMLElement;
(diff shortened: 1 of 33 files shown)


tests/components/bottom_bar.test.ts
```diff
@@ -38,7 +38,7 @@ async function mountBottomBar(
   env: Partial<SpreadsheetChildEnv> = {}
 ): Promise<{ parent: Parent; app: App; model: Model }> {
   const mockEnv = makeTestEnv({ ...env, model });
-  const app = new App(Parent, { props: { model }, env: mockEnv });
+  const app = new App(Parent, { props: { model }, env: mockEnv, test: true });
```
pro-odoo, line 41: What do you think about creating a helper to mount a component ?
```js
async function mount(component, target, props={}, env={}) {
    const app = new App(component, { props, env, test: true });
    app.addTemplates(OWL_TEMPLATES);
    const parent = await app.mount(target);
    return { app, parent }
}
```
With that, we could avoid thinking of adding `test: true`, OWL_TEMPLATES, ...

tests/plugins/merges.test.ts
```diff
@@ -332,7 +332,7 @@ describe("merges", () => {
       }
     }
     const fixture = makeTestFixture();
-    const app = new App(Parent);
```
rrahir, line 335: This tests the interactiveMerge, which has nothing to do with the plugins. I'm pretty sure it's already tested in the file helpers/ui.test.ts ...
If its the case, can you remove the test completely ?

tests/components/bottom_bar.test.ts
pro-odoo: To go one step further (perhaps not in this PR, but...), we could create something like in [Odoo](https://github.com/odoo/odoo/blob/master/addons/web/static/tests/helpers/cleanup.js#L7), to register a cleanup. So, we will not be forced to take care of destroying the app. See the [helper](https://github.com/odoo/odoo/blob/master/addons/web/static/tests/helpers/utils.js#L608) of web

pro-odoo (commented):
We can merge as it, my comment is only a proposition

tests/components/bottom_bar.test.ts
hokolomopo: Good idea !

rrahir: don't forget fixture ^^

pro-odoo (approved):
robodoo r+

PR #2057, opened 2023-02-06
[FIX] composer: fix wrong style attribute

rrahir:
The improvement in 3710b7df let a small typo through. unfortunately, the affected code was not properly tested.

Task 3172129

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/composer/composer/composer.ts
+++ b/src/components/composer/composer/composer.ts
@@ -169,7 +169,8 @@ export class Composer extends Component<ComposerProps, SpreadsheetChildEnv> {
       let assistantStyle: CSSProperties = {};
       if (cellY > remainingHeight) {
         // render top
-        assistantStyle.top = `$-3px`;
+        // We compensate 2 px of margin on the assistant style + 1px for design reasons
+        assistantStyle.top = `-3px`;
         assistantStyle.transform = `translate(0, -100%)`;
       }
       if (cellX + ASSISTANT_WIDTH > this.props.delimitation.width) {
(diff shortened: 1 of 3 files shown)


src/components/composer/composer/composer.ts
```diff
@@ -169,7 +169,7 @@ export class Composer extends Component<ComposerProps, SpreadsheetChildEnv> {
       let assistantStyle: CSSProperties = {};
       if (cellY > remainingHeight) {
         // render top
-        assistantStyle.top = `$-3px`;
+        assistantStyle.top = `-3px`;
```
hokolomopo, line 172: I invistigated a bit the `-3px` because why -3. I think the -3 is because : -2px to compensate the vertical margin of the `o-composer-assisstant`, and -1px to be displayed a px above the border of the composer.

If we replace the margin of the `o-composer-assistant` with padding, we don't need the -2.
The -1 for the border is unavoidable I think, but it's a bit strange since we don't do that when the assisstant is below the composer, the composer border will be hidden by the assisstant. 

We should at the very least add a comment on the why this value.

tests/components/composer_integration.test.ts
```diff
@@ -617,3 +617,25 @@ describe("Grid composer", () => {
...
+  });
+
+  test("render above the cell when not enough place below", async () => {
+    const { left, bottom } = model.getters.getActiveMainViewport();
+    selectCell(model, toXC(left, bottom));
+    await typeInComposerGrid("=s");
+    const assistantEL = fixture.querySelector(".o-composer-assistant")! as HTMLElement;
+    expect(assistantEL).toMatchSnapshot();
+    expect(assistantEL.style.width).toBe("300px");
+    expect(assistantEL.style.top).toBe("-3px");
```
hokolomopo, line 639: also test the translate property, the -3px by itself does't do much

LucasLefevre (approved):
robodoo r+

PR #2060, opened 2023-02-08
[IMP] functions: translate arguments

pro-odoo:
## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/functions/arguments.ts
+++ b/src/functions/arguments.ts
@@ -20,22 +20,8 @@ const ARG_TYPES: ArgType[] = [
   "META",
 ];
 
-/**
- * This function is meant to be used as a tag for a template strings.
- *
- * Its job is to convert a textual description of the list of arguments into an
- * actual array of Arg, suitable for consumption.
- */
-export function args(strings: string): ArgDefinition[] {
-  let lines = strings.split("\n");
-  const result: ArgDefinition[] = [];
-  for (let l of lines) {
-    l = l.trim();
-    if (l) {
-      result.push(makeArg(l));
-    }
-  }
-  return result;
+export function arg(definition: string, description: string = ""): ArgDefinition {
+  return makeArg(`${definition} ${description}`);
 }
 
 function makeArg(str: string): ArgDefinition {
--- a/src/functions/module_custom.ts
+++ b/src/functions/module_custom.ts
@@ -1,7 +1,7 @@
 import { createLargeNumberFormat } from "../helpers";
 import { _lt } from "../translation";
 import { AddFunctionDescription, PrimitiveArg, PrimitiveArgValue } from "../types";
-import { args } from "./arguments";
+import { arg } from "./arguments";
 import { toNumber } from "./helpers";
 
 // -----------------------------------------------------------------------------
@@ -9,10 +9,13 @@ import { toNumber } from "./helpers";
 // -----------------------------------------------------------------------------
 export const FORMAT_LARGE_NUMBER: AddFunctionDescription = {
   description: _lt(`Apply a large number format`),
-  args: args(`
-      value (number) ${_lt("The number.")}
-      unit (string, optional) ${_lt("The formatting unit. Use 'k', 'm', or 'b' to force the unit")}
-    `),
+  args: [
+    arg("value (number)", _lt("The number.")),
+    arg(
+      "unit (string, optional)",
+      _lt("The formatting unit. Use 'k', 'm', or 'b' to force the unit")
+    ),
+  ],
   returns: ["NUMBER"],
   computeFormat: (arg: PrimitiveArg, unit: PrimitiveArg | undefined) => {
     const value = Math.abs(toNumber(arg.value));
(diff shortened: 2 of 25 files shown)


src/functions/module_engineering.ts
```diff
@@ -10,10 +10,10 @@ const DEFAULT_DELTA_ARG = 0;
 // -----------------------------------------------------------------------------
 export const DELTA: AddFunctionDescription = {
   description: _lt("Compare two numeric values, returning 1 if they're equal."),
-  args: args(`
-  number1  (number) ${_lt("The first number to compare.")}
-  number2  (number, default=${DEFAULT_DELTA_ARG}) ${_lt("The second number to compare.")}
-  `),
+  args: args(
+    [" (number)", _lt("The first number to compare.")],
+    [` (number, default=${DEFAULT_DELTA_ARG})`, _lt("The second number to compare.")]
```
LucasLefevre, line 15: missing arg names

src/functions/module_custom.ts
```diff
@@ -9,10 +9,10 @@ import { toNumber } from "./helpers";
...
 export const FORMAT_LARGE_NUMBER: AddFunctionDescription = {
   description: _lt(`Apply a large number format`),
-  args: args(`
-      value (number) ${_lt("The number.")}
-      unit (string, optional) ${_lt("The formatting unit. Use 'k', 'm', or 'b' to force the unit")}
-    `),
+  args: args(
+    ["value (number)", _lt("The number.")],
+    ["unit (string, optional)", _lt("The formatting unit. Use 'k', 'm', or 'b' to force the unit")]
+  ),
```
LucasLefevre, line 15: Alternative way of writing the same thing.
Not sure it's better. Maybe it feels less weird than an arbitrary array with 2 elements 
WDYT?
```suggestion
  args: [
    arg("value (number)", _lt("The number.")),
    arg("unit (string, optional)", _lt("The formatting unit. Use 'k', 'm', or 'b' to force the unit"))
  ],
```

LucasLefevre (commented):
don't forget to update the doc once we are sure of API :)

src/functions/module_custom.ts
pro-odoo: As it's a code that is not often read, I think it will be easier to go and see the definition of `arg` instead of the function `args` to see what we do with the array.
Let's go for this solution

src/functions/module_lookup.ts
```diff
@@ -262,11 +237,11 @@ export const MATCH: AddFunctionDescription = {
...
   description: _lt("Row number of a specified cell."),
-  args: args(
-    `cell_reference (meta, default=${_lt(
-      "The cell in which the formula is entered by default"
-    )}) ${_lt("The cell whose row number will be returned.")}`
-  ),
+  args: [arg(
+    `cell_reference (meta)`,
+    _lt("The cell whose row number will be returned."),
+    { default: _lt("The cell in which the formula is entered by default")}
```
pro-odoo, line 243: @LucasLefevre What do you think of this API ? I do not see another good solution to translate default value

LucasLefevre: Ah yes, but we saw this particular default last week (I think you were working from home). This default value is actually not used like it should. It should be a real value, not a description for the value.
I think I would just remove it or move it to the arg description

pro-odoo: What about `true` and `false` ? It's a value, but should be translated IMO. But it can be always moved in the arg description, it's fine for me

LucasLefevre: @laa-odoo had a similar thought this morning about argument names that are not translated (`cell_reference`).
Since we decided to not translate function names, I think it's acceptable to do the same for arg names and boolean values.
Description are another story since they are real English sentences.

src/functions/module_lookup.ts
```diff
@@ -262,11 +280,14 @@ export const MATCH: AddFunctionDescription = {
...
-  args: args(
-    `cell_reference (meta, default=${_lt(
-      "The cell in which the formula is entered by default"
-    )}) ${_lt("The cell whose row number will be returned.")}`
-  ),
+  args: [
+    arg(
+      "cell_reference (meta, default='this cell')",
+      _lt(
+        "The cell whose row number will be returned. By default, this function uses the cell in which the formula is entered by default"
```
LucasLefevre, line 287: "by default" is present twice in the sentence.
Either keep the first one or second one

LucasLefevre (approved):
minor things

Translators will be busy for a while :smile:

src/functions/arguments.ts
```diff
@@ -20,22 +20,8 @@ const ARG_TYPES: ArgType[] = [
...
-  let lines = strings.split("\n");
-  const result: ArgDefinition[] = [];
-  for (let l of lines) {
-    l = l.trim();
-    if (l) {
-      result.push(makeArg(l));
-    }
-  }
-  return result;
+export function arg(nameType: string, description: string): ArgDefinition {
```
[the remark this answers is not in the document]
pro-odoo: it means name + types of the argument. `definition` seems ok to me

pro-odoo:
> minor things
> 
> Translators will be busy for a while 😄

Updated :)

LucasLefevre (approved):
robodoo r+

PR #2066, opened 2023-02-09
[FIX] selection input: dynamic highlights

yunchen4:
## Description:

This PR fixes the problem of flashing highlight when confirming a selection input. 
The solution is to make the selection input range update dynamically so the highlight can show instantly.

This PR also fixes a bug of possible duplicate range ids in selection input plugin. 
This PR also prevents the leading commas to create a new range input. The trailing comma will still lead to a new range, and the focus will also be transferred into this new range. 

Odoo task ID : [2884221](https://www.odoo.com/web#id=2884221&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/selection_input/selection_input.ts
+++ b/src/components/selection_input/selection_input.ts
@@ -1,4 +1,12 @@
-import { Component, onMounted, onPatched, onWillUnmount, useState } from "@odoo/owl";
+import {
+  Component,
+  onMounted,
+  onPatched,
+  onWillUnmount,
+  useEffect,
+  useRef,
+  useState,
+} from "@odoo/owl";
 import { SELECTION_BORDER_COLOR } from "../../constants";
 import { UuidGenerator } from "../../helpers/index";
 import { RangeInputValue } from "../../plugins/ui_feature/selection_input";
@@ -93,6 +101,7 @@ export class SelectionInput extends Component<Props, SpreadsheetChildEnv> {
     isMissing: false,
     mode: "select-range",
   });
+  private focusedInput = useRef("focusedInput");
 
   get ranges(): SelectionRange[] {
     const existingSelectionRange = this.env.model.getters.getSelectionInput(this.id);
@@ -124,6 +133,10 @@ export class SelectionInput extends Component<Props, SpreadsheetChildEnv> {
   }
 
   setup() {
+    useEffect(
+      () => this.focusedInput.el?.focus(),
+      () => [this.focusedInput.el]
+    );
     onMounted(() => this.enableNewSelectionInput());
     onWillUnmount(async () => this.disableNewSelectionInput());
     onPatched(() => this.checkChange());
@@ -170,6 +183,9 @@ export class SelectionInput extends Component<Props, SpreadsheetChildEnv> {
         ev.preventDefault();
         updateSelectionWithArrowKeys(ev, this.env.model.selection);
       }
+    } else if (ev.key === "Enter") {
+      const target = ev.target as HTMLInputElement;
+      target.blur();
     }
   }
 
@@ -199,7 +215,6 @@ export class SelectionInput extends Component<Props, SpreadsheetChildEnv> {
       rangeId,
       value: target.value,
     });
-    target.blur();
     this.triggerChange();
   }
 
--- a/src/components/selection_input/selection_input.xml
+++ b/src/components/selection_input/selection_input.xml
@@ -10,7 +10,7 @@
         <input
           type="text"
           spellcheck="false"
-          t-on-change="(ev) => this.onInputChanged(range.id, ev)"
+          t-on-input="(ev) => this.onInputChanged(range.id, ev)"
           t-on-focus="() => this.focus(range.id)"
           t-on-keydown="onKeydown"
           t-att-value="range.xc"
@@ -22,6 +22,7 @@
             'o-invalid': isInvalid || !range.isValidRange,
             'text-decoration-underline': range.isFocused and state.mode === 'select-range'
           }"
+          t-ref="{{range.isFocused ? 'focused' : 'unfocused'}}Input"
         />
         <button
           class="o-btn o-remove-selection"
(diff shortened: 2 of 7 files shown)


src/plugins/ui_feature/selection_input.ts
```diff
@@ -181,13 +182,16 @@ export class SelectionInputPlugin extends UIPlugin implements StreamCallbacks<Se
    * Insert new inputs after the given index.
    */
   private insertNewRange(index: number, values: string[]) {
+    const currentMaxId = this.ranges.length
+      ? Math.max(...this.ranges.map((range) => Number(range.id)))
+      : 0;
```
hokolomopo, line 187:
```suggestion
    const currentMaxId = Math.max(0, ...this.ranges.map((range) => Number(range.id)));
```

tests/components/conditional_formatting.test.ts
```diff
@@ -611,7 +611,7 @@ describe("UI of conditional formats", () => {
     await click(fixture, selectors.buttonSave);
 
     const sheetId = model.getters.getActiveSheetId();
-    expect(dispatch).toHaveBeenCalledWith("ADD_CONDITIONAL_FORMAT", {
+    expect(dispatch).toHaveBeenNthCalledWith(1, "ADD_CONDITIONAL_FORMAT", {
```
hokolomopo, line 614: Why this change ? The test seems to work with just `toHaveBeenCalledWith ` , and we don't really care about the exact sequence order of dispatch

tests/components/selection_input.test.ts
```diff
@@ -221,24 +220,29 @@ describe("Selection Input", () => {
...
-    expect(model.getters.getSelectionInput(id)[0].xc).toBe("C2");
-    expect(fixture.querySelectorAll("input")[1].value).toBe("A1");
-    expect(model.getters.getSelectionInput(id)[1].xc).toBe("A1");
-  });
-
-  test("manually add another cell", async () => {
+  test("manually add another range via trailing comma", async () => {
     const { model, id } = await createSelectionInput({ initialRanges: ["C2"] });
-    await writeInput(0, "C2,A1");
+    await writeInput(0, "C2,");
```
hokolomopo, line 225: We should keep the test `C2,A1`, this is still possible for it to happen if the user were to paste the value "C2,A1" directly in the input

tests/components/selection_input.test.ts
```diff
@@ -290,6 +294,15 @@ describe("Selection Input", () => {
     expect(fixture.querySelector(".input-2 .o-focused")).toBeTruthy();
   });
 
+  test("focus is transferred into the newly added input automatically when typing comma at the end", async () => {
```
hokolomopo, line 297: also test the `document.activeElement` value

hokolomopo (commented):
:+1: 
some small details

LucasLefevre (approved):
robodoo r+ :)

src/plugins/ui_feature/selection_input.ts
```diff
@@ -181,13 +182,14 @@ export class SelectionInputPlugin extends UIPlugin implements StreamCallbacks<Se
...
   private insertNewRange(index: number, values: string[]) {
+    const currentMaxId = Math.max(0, ...this.ranges.map((range) => Number(range.id)));
     this.ranges.splice(
       index,
       0,
       ...values.map((xc, i) => ({
         xc,
-        id: (this.ranges.length + i + 1).toString(),
-        color: colors[(this.ranges.length + i) % colors.length],
+        id: (currentMaxId + i + 1).toString(),
```
rrahir, line 191: you start at index 1, which means you skip `colors[0]`, why was this change required ? I don't see any test related ?

rrahir: the typing of id is UID, which means it could pretty much be any kind of string - I think it'd be way better to drop the UID and work with numbers don't you think ?

rrahir (commented):
SOrry for this, I had started a review and did not move the task, then lucas reviewed and merged. That being said, I think the task is not 100% correct and would like my comments here to be considered in a fast followup

src/components/selection_input/selection_input.ts
```diff
@@ -170,6 +183,9 @@ export class SelectionInput extends Component<Props, SpreadsheetChildEnv> {
         ev.preventDefault();
         updateSelectionWithArrowKeys(ev, this.env.model.selection);
       }
+    } else if (ev.key === "Enter") {
+      const target = ev.target as HTMLInputElement;
+      target.blur();
```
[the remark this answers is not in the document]
yunchen4: @rrahir It's indeed useless, and it's added for user experience, otherwise when users hit enter the input cursor it still there (which I feel weird). But it can be deleted anyway 🙂

src/plugins/ui_feature/selection_input.ts
yunchen4: @rrahir I changed it because in the old implementation `id` can be duplicate in some cases, which causes the some problems when focusing and updating input. My bad! I didn't notice the color will also start at index 1. I can change the calculation of `currentMaxId` to compare to `-1`. 
For the type it seems to be a legacy problem (?) But I agreed with you. Number type is better. I can change it in another PR.

src/components/selection_input/selection_input.xml
```diff
@@ -22,6 +22,7 @@
             'o-invalid': isInvalid || !range.isValidRange,
             'text-decoration-underline': range.isFocused and state.mode === 'select-range'
           }"
+          t-ref="{{range.isFocused ? 'focused' : 'unfocused'}}Input"
```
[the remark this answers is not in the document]
yunchen4: @rrahir Do you mean `t-ref` cannot be updated like `t-att-class` which also uses `range.isFocused` if clicking on another range? 
I test on demo it works well when clicking another range (the focus will change correctly), but in the jest `document.activeElement` is not updated (`o-focused` class is changed correctly).

src/components/selection_input/selection_input.ts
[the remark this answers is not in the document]
LucasLefevre: > it's added for user experience

It's not useless then :sweat_smile: 
That being said, the behavior of the Enter key could still be improved IMO. I think it should have the same effect as hitting the Confirm button

PR #2079, opened 2023-02-16
[IMP] figure: reposition figure upon row/col deletion

jash-odoo:
## Description:

This commit improves the positioning of the figures regarding the
structure of the sheet. Upon deletion of rows/columns[1], the figure will
be repositions to fit (as much as possible) inside the sheet grid, that
is having the figure not outside of the grid.

This commit also ensures this behaviour when copy/pasting a figure.

[1] A follow up work will address the visibility of the headers but this
requires a more in-depth refactoring which is far enough form the task
original goal to be left for later.

Task: 3171141

Odoo task ID : [3171141](https://www.odoo.com/web#id=3171141&cids=2&menu_id=4720&action=4043&model=project.task&view_type=form)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/clipboard/clipboard_figure_state.ts
+++ b/src/helpers/clipboard/clipboard_figure_state.ts
@@ -84,14 +84,19 @@ export class ClipboardFigureState implements ClipboardState {
    */
   paste(target: Zone[]) {
     const sheetId = this.getters.getActiveSheetId();
+    const { width, height } = this.copiedFigure;
+    const numCols = this.getters.getNumberCols(sheetId);
+    const numRows = this.getters.getNumberRows(sheetId);
+    const targetX = this.getters.getColDimensions(sheetId, target[0].left).start;
+    const targetY = this.getters.getRowDimensions(sheetId, target[0].top).start;
+    const maxX = this.getters.getColDimensions(sheetId, numCols - 1).end;
+    const maxY = this.getters.getRowDimensions(sheetId, numRows - 1).end;
     const position = {
-      x: this.getters.getColDimensions(sheetId, target[0].left).start,
-      y: this.getters.getRowDimensions(sheetId, target[0].top).start,
+      x: maxX < width ? 0 : Math.min(targetX, maxX - width),
+      y: maxY < height ? 0 : Math.min(targetY, maxY - height),
     };
-    const size = { height: this.copiedFigure.height, width: this.copiedFigure.width };
-
     const newId = new UuidGenerator().uuidv4();
-    this.copiedFigureContent.paste(sheetId, newId, position, size);
+    this.copiedFigureContent.paste(sheetId, newId, position, { height, width });
 
     if (this.operation === "CUT") {
       this.dispatch("DELETE_FIGURE", {
--- a/src/plugins/core/figures.ts
+++ b/src/plugins/core/figures.ts
@@ -53,6 +53,42 @@ export class FigurePlugin extends CorePlugin<FigureState> implements FigureState
       case "DELETE_FIGURE":
         this.removeFigure(cmd.id, cmd.sheetId);
         break;
+      case "REMOVE_COLUMNS_ROWS":
+        this.onRowColDelete(cmd.sheetId, cmd.dimension);
+    }
+  }
+
+  private onRowColDelete(sheetId: string, dimension: string) {
+    dimension === "ROW" ? this.onRowDeletion(sheetId) : this.onColDeletion(sheetId);
+  }
+
+  private onRowDeletion(sheetId: string) {
+    const numHeader = this.getters.getNumberRows(sheetId);
+    let gridHeight = 0;
+    for (let i = 0; i < numHeader; i++) {
+      gridHeight += this.getters.getRowSize(sheetId, i);
+    }
+    const figures = this.getters.getFigures(sheetId);
+    for (const figure of figures) {
+      const newY = Math.min(figure.y, gridHeight - figure.height);
+      if (newY !== figure.y) {
+        this.dispatch("UPDATE_FIGURE", { sheetId, id: figure.id, y: newY });
+      }
+    }
+  }
+
+  private onColDeletion(sheetId: string) {
+    const numHeader = this.getters.getNumberCols(sheetId);
+    let gridWidth = 0;
+    for (let i = 0; i < numHeader; i++) {
+      gridWidth += this.getters.getColSize(sheetId, i);
+    }
+    const figures = this.getters.getFigures(sheetId);
+    for (const figure of figures) {
+      const newX = Math.min(figure.x, gridWidth - figure.width);
+      if (newX !== figure.x) {
+        this.dispatch("UPDATE_FIGURE", { sheetId, id: figure.id, x: newX });
+      }
     }
   }
 
--- a/tests/plugins/clipboard_figure.test.ts
+++ b/tests/plugins/clipboard_figure.test.ts
@@ -128,6 +128,29 @@ describe.each(["chart", "image"])("Clipboard for %s figures", (type: string) => 
     expect(model.getters.getFigures(sheetId)).toHaveLength(0);
   });
 
+  test("Figure is copied to edge of the sheet", () => {
+    model.dispatch("UPDATE_FIGURE", {
+      sheetId,
+      id: figureId,
+      height: 256,
+      width: 257,
+    });
+    model.dispatch("SELECT_FIGURE", { id: figureId });
+    copy(model);
+    paste(model, "Z100");
+    const copiedFigure = model.getters.getFigure(sheetId, getCopiedFigureId())!;
+    const maxX = model.getters.getColDimensions(
+      sheetId,
+      model.getters.getNumberCols(sheetId) - 1
+    ).end;
+    const maxY = model.getters.getRowDimensions(
+      sheetId,
+      model.getters.getNumberRows(sheetId) - 1
+    ).end;
+    expect(copiedFigure.x).toBe(maxX - copiedFigure.width);
+    expect(copiedFigure.y).toBe(maxY - copiedFigure.height);
+  });
+
   describe("Paste command result", () => {
     test("Cannot paste with empty target", () => {
       model.dispatch("SELECT_FIGURE", { id: figureId });
--- a/src/components/figures/figure_container/figure_container.ts
+++ b/src/components/figures/figure_container/figure_container.ts
@@ -262,6 +262,17 @@ export class FiguresContainer extends Component<Props, SpreadsheetChildEnv> {
 
     const initialMousePosition = { x: ev.clientX, y: ev.clientY };
 
+    const maxDimensions = {
+      maxX: this.env.model.getters.getColDimensions(
+        sheetId,
+        this.env.model.getters.getNumberCols(sheetId) - 1
+      ).end,
+      maxY: this.env.model.getters.getRowDimensions(
+        sheetId,
+        this.env.model.getters.getNumberRows(sheetId) - 1
+      ).end,
+    };
+
     const { x, y } = internalFigureToScreen(this.env.model.getters, figure);
 
     const initialFig = { ...figure, x, y };
@@ -274,6 +285,7 @@ export class FiguresContainer extends Component<Props, SpreadsheetChildEnv> {
         initialMousePosition,
         initialFig,
         this.env.model.getters.getMainViewportCoordinates(),
+        maxDimensions,
         getters.getActiveSheetScrollInfo()
       );
 
--- a/src/components/helpers/figure_drag_helper.ts
+++ b/src/components/helpers/figure_drag_helper.ts
@@ -5,17 +5,20 @@ export function dragFigureForMove(
   { x: mouseInitialX, y: mouseInitialY }: PixelPosition,
   initialFigure: Figure,
   { x: viewportX, y: viewportY }: PixelPosition,
+  { maxX, maxY }: { maxX: number; maxY: number },
   { scrollX, scrollY }: SheetScrollInfo
 ): Figure {
   const minX = viewportX ? 0 : -scrollX;
   const minY = viewportY ? 0 : -scrollY;
-
-  let deltaX = mouseX - mouseInitialX;
-  const newX = Math.max(initialFigure.x + deltaX, minX);
-  let deltaY = mouseY - mouseInitialY;
-  const newY = Math.max(initialFigure.y + deltaY, minY);
-
+  const deltaX = mouseX - mouseInitialX;
+  const newX = clamp(initialFigure.x + deltaX, minX, maxX - initialFigure.width - scrollX);
+  const deltaY = mouseY - mouseInitialY;
+  const newY = clamp(initialFigure.y + deltaY, minY, maxY - initialFigure.height - scrollY);
   return { ...initialFigure, x: newX, y: newY };
+}
+
+function clamp(value: number, min: number, max: number): number {
+  return Math.min(Math.max(value, min), max);
 }
 
 export function dragFigureForResize(
--- a/tests/components/figure.test.ts
+++ b/tests/components/figure.test.ts
@@ -953,8 +953,14 @@ describe("figures", () => {
 
       test("No Y snap with bottom border below the viewport", async () => {
         const { height: viewportHeight } = model.getters.getMainViewportRect();
-        createFigure(model, { id: "f1", x: 0, y: 100, width: 100, height: viewportHeight });
-        createFigure(model, { id: "f2", x: 0, y: 0, width: 100, height: viewportHeight + 100 });
+        createFigure(model, { id: "f1", x: 0, y: 100, width: 100, height: 0.85 * viewportHeight });
+        createFigure(model, {
+          id: "f2",
+          x: 0,
+          y: 0,
+          width: 100,
+          height: 0.85 * viewportHeight + 100,
+        });
         await nextTick();
         await dragElement(".o-figure[data-id=f1]", { x: 0, y: 1 }, undefined, true);
         expect(model.getters.getFigure(sheetId, "f1")).toMatchObject({ x: 0, y: 101 });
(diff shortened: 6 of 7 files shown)


src/helpers/clipboard/clipboard_figure_state.ts
```diff
@@ -84,12 +84,20 @@ export class ClipboardFigureState implements ClipboardState {
...
-      x: this.getters.getColDimensions(sheetId, target[0].left).start,
-      y: this.getters.getRowDimensions(sheetId, target[0].top).start,
+      x:
+        viewportRect.width - colDimensions.start < size.width
+          ? viewportRect.width - size.width - 96
+          : colDimensions.start,
+      y:
+        viewportRect.height - rowDimensions.start < size.height
+          ? viewportRect.height - size.height - 28
+          : rowDimensions.start,
```
hokolomopo, line 99: Doesn't work wiht frozen pane. If there are frozen pane getMainViewportRect !== grid dimension. Ise the position of the last column/row instead.

tests/plugins/clipboard_figure.test.ts
```diff
@@ -128,6 +128,24 @@ describe.each(["chart", "image"])("Clipboard for %s figures", (type: string) =>
...
+    });
+    model.dispatch("SELECT_FIGURE", { id: figureId });
+    copy(model);
+    paste(model, "Z100");
+    const copiedFigure = model.getters.getFigure(sheetId, getCopiedFigureId());
+    const { height, width } = model.getters.getMainViewportRect();
+    const bottomRight = copiedFigure!.x + copiedFigure!.width;
+    expect(bottomRight).toBeLessThan(width);
+    const bottomHigh = copiedFigure!.y + copiedFigure!.height;
+    expect(bottomHigh).toBeLessThan(height);
```
hokolomopo, line 146: These are topRight and bottomLeft

hokolomopo (commented):
:+1:

src/helpers/clipboard/clipboard_figure_state.ts
hokolomopo: Doing somethink like `mainViewport.x + mainViewport.width` would probably work yeah, but you really shouldn't care here about viewports and whatnot, all you want is the end of the sheet. `getColDimension(lastCol)` is cleaner IMO.

src/helpers/clipboard/clipboard_figure_state.ts
```diff
@@ -84,12 +84,20 @@ export class ClipboardFigureState implements ClipboardState {
...
+    const size = { height: this.copiedFigure.height, width: this.copiedFigure.width };
+    const viewportRect = this.getters.getMainViewportRect();
+    const colDimensions = this.getters.getColDimensions(sheetId, target[0].left);
+    const rowDimensions = this.getters.getRowDimensions(sheetId, target[0].top);
     const position = {
-      x: this.getters.getColDimensions(sheetId, target[0].left).start,
-      y: this.getters.getRowDimensions(sheetId, target[0].top).start,
+      x:
+        viewportRect.width - colDimensions.start < size.width
+          ? viewportRect.width + viewportRect.x - size.width - 96
```
hokolomopo, line 94: also don't use random numbers in the operations. What are 96 and 28?

src/helpers/clipboard/clipboard_figure_state.ts
```diff
@@ -84,12 +84,20 @@ export class ClipboardFigureState implements ClipboardState {
    */
   paste(target: Zone[]) {
     const sheetId = this.getters.getActiveSheetId();
+    const size = { height: this.copiedFigure.height, width: this.copiedFigure.width };
```
rrahir, line 87: const {width ,height} = this.copiedFigure
```suggestion
    const {width, height} = this.copiedFigure
```

src/helpers/clipboard/clipboard_figure_state.ts
```diff
@@ -84,12 +84,20 @@ export class ClipboardFigureState implements ClipboardState {
...
+    const viewportRect = this.getters.getMainViewportRect();
```
rrahir, line 88: I don't think you are looking for this value but rather the size of the grid that can be seen by the client? because the "mainViewport" will have another size that the aforementioned when we have frozen panes. Specifically,  the 'Main viewport' corresponds to the bottom right viewport.  It just so happens that it is equal to the grid size when there are not frozen panes. But you should take those into account.

src/helpers/clipboard/clipboard_figure_state.ts
```diff
@@ -84,12 +84,20 @@ export class ClipboardFigureState implements ClipboardState {
...
+    const colDimensions = this.getters.getColDimensions(sheetId, target[0].left);
```
rrahir, line 89: if you only plan on using one key of an object you can destructure it directly.
```suggestion
    const {start: colStart} = this.getters.getColDimensions(sheetId, target[0].left);
```
will do create the variable `colStart` for instance. It will also make the rest of the code more readable

src/helpers/clipboard/clipboard_figure_state.ts
```diff
@@ -84,12 +84,20 @@ export class ClipboardFigureState implements ClipboardState {
...
+    const rowDimensions = this.getters.getRowDimensions(sheetId, target[0].top);
     const position = {
-      x: this.getters.getColDimensions(sheetId, target[0].left).start,
-      y: this.getters.getRowDimensions(sheetId, target[0].top).start,
+      x:
+        viewportRect.width - colDimensions.start < size.width
```
rrahir, line 93: pretty sure you lack a condition to prevent x and y to be negative.  Why not just computing the value `viewportRect.width - (size.width + colDimensions.start)` and if it's negative, subtract if from colDimensions.start ? (resp the sale thing for `y`)

tests/plugins/clipboard_figure.test.ts
```diff
@@ -128,6 +128,24 @@ describe.each(["chart", "image"])("Clipboard for %s figures", (type: string) =>
...
+    model.dispatch("UPDATE_FIGURE", {
+      sheetId,
+      id: figureId,
+      height: 256,
+      width: 257,
+    });
+    model.dispatch("SELECT_FIGURE", { id: figureId });
+    copy(model);
+    paste(model, "Z100");
+    const copiedFigure = model.getters.getFigure(sheetId, getCopiedFigureId());
```
rrahir, line 141: you can force the non-null assertion here rather than calling it on every subsequent call
```suggestion
    const copiedFigure = model.getters.getFigure(sheetId, getCopiedFigureId())!;
```

tests/plugins/clipboard_figure.test.ts
```diff
@@ -128,6 +128,24 @@ describe.each(["chart", "image"])("Clipboard for %s figures", (type: string) =>
...
+    const { height, width } = model.getters.getMainViewportRect();
+    const rightCoordinate = copiedFigure!.x + copiedFigure!.width;
+    expect(rightCoordinate).toBeLessThan(width);
```
rrahir, line 144: you could go a bit further and recompute the new size ;-)

rrahir (changes requested):

src/helpers/clipboard/clipboard_figure_state.ts
rrahir: what about this ?

src/helpers/clipboard/clipboard_figure_state.ts
rrahir: @jash-odoo   see here...

src/helpers/clipboard/clipboard_figure_state.ts
jash-odoo: Update the code and use the dimension of last column and row for identifying the end coordinate of the sheet

src/helpers/clipboard/clipboard_figure_state.ts
```diff
@@ -84,14 +84,19 @@ export class ClipboardFigureState implements ClipboardState {
...
+    const numCols = this.getters.getNumberCols(sheetId);
+    const numRows = this.getters.getNumberRows(sheetId);
+    const { start: colStart } = this.getters.getColDimensions(sheetId, target[0].left);
+    const { start: rowStart } = this.getters.getRowDimensions(sheetId, target[0].top);
+    const { end: lastColEnd } = this.getters.getColDimensions(sheetId, numCols - 1);
+    const { end: lastRowEnd } = this.getters.getRowDimensions(sheetId, numRows - 1);
     const position = {
-      x: this.getters.getColDimensions(sheetId, target[0].left).start,
-      y: this.getters.getRowDimensions(sheetId, target[0].top).start,
+      x: lastColEnd - (colStart + width) < 0 ? lastColEnd - width : colStart,
```
rrahir, line 95: What is the expected behavior when width is bigger than lasColEnd ?

You can imagine the following scenario:;
I have 2 sheets
in the second sheet, I delete all but 1 column (only have col A)
in the first sheet I create a chart, and make it large enough
copy the chart
got to sheet2, paste it

the current code will dispatch negative positions for x (you can do the same reasoning with the rows)

I think we want x to be non-negative. Could you check this with our PO ?

tests/plugins/clipboard_figure.test.ts
```diff
@@ -128,6 +128,22 @@ describe.each(["chart", "image"])("Clipboard for %s figures", (type: string) =>
...
+      height: 256,
+      width: 257,
+    });
+    model.dispatch("SELECT_FIGURE", { id: figureId });
+    copy(model);
+    paste(model, "Z100");
+    const copiedFigure = model.getters.getFigure(sheetId, getCopiedFigureId())!;
+    const { height, width } = model.getters.getMainViewportRect();
+    expect(copiedFigure.x).toBe(width - copiedFigure.width - 96);
+    expect(copiedFigure.y).toBe(height - copiedFigure.height - 28);
```
hokolomopo, line 144: Adapt the test to reflect your new computations - no viewport size + no  96/28 constants

src/helpers/clipboard/clipboard_figure_state.ts
```diff
@@ -84,14 +84,19 @@ export class ClipboardFigureState implements ClipboardState {
    */
   paste(target: Zone[]) {
     const sheetId = this.getters.getActiveSheetId();
+    const { width, height } = this.copiedFigure;
+    const numCols = this.getters.getNumberCols(sheetId);
+    const numRows = this.getters.getNumberRows(sheetId);
+    const { start: colStart } = this.getters.getColDimensions(sheetId, target[0].left);
```
hokolomopo, line 90:
```suggestion
    const targetX = this.getters.getColDimensions(sheetId, target[0].left).start;
```

Kinda nitpicking, `colStart` isn't a great variable name since we don't know which col we're talking about +  unpacking is useless (and harder to read) when you only need az single variable.

Same comment for the 4 lines below

hokolomopo (commented):
Almost there :+1: 

Don't forget to change the task state in Odoo every time

rrahir (commented):
I asked our PO about the desired behaviour but I think we need to  have a different behaviour if we try to copy in a frozen pane or not.

src/helpers/clipboard/clipboard_figure_state.ts
```diff
@@ -84,14 +84,23 @@ export class ClipboardFigureState implements ClipboardState {
...
+    const numCols = this.getters.getNumberCols(sheetId);
+    const numRows = this.getters.getNumberRows(sheetId);
+    const targetX = this.getters.getColDimensions(sheetId, target[0].left).start;
+    const targetY = this.getters.getRowDimensions(sheetId, target[0].top).start;
+    const maxX = this.getters.getColDimensions(sheetId, numCols - 1).end;
+    const maxY = this.getters.getRowDimensions(sheetId, numRows - 1).end;
+    const mainViewportRect = this.getters.getMainViewportRect();
+    const position =
+      mainViewportRect.width < width || mainViewportRect.height < height
+        ? { x: targetX, y: targetY }
```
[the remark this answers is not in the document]
jash-odoo: After thinking about this problem, I came up with few observations, 
If we try to paste a chart on the very last cell then according to the fix it should print the chart on opposite direction if there is proper viewport available. Since the available width & height of viewport is not enough for chart, current code prints it as normal case i.e. starting from same position (which results in same exact situation you found).

I propose a solution to this, but I'm not sure about how it will look, if there is no ample space in x-axis then code should reset the start point as 0th cordinate, same goes for y axis.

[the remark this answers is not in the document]
jash-odoo: When we attempt to copy in a frozen pane, we need to determine the new origin of the spreadsheet. This means that the sheet will not start from (0,0) but from some other point. We can obtain the updated (x,y) coordinates using the mainViewportRect() function if there is a frozen pane. By knowing this new coordinate, we can adjust our work accordingly and avoid creating a different behavior.

rrahir (changes requested):
Sorry for the delay in the review, I have been quite busy + took some days off.

You are still not entirely addressing the spec that François updated as you move the figures in the action. While it seems to work when you test it live (by the way, there are no tests for that feature), there are cases that you do not address, which is the multi user  use-case and the UNDO/REDO.

Since you are trying to update the figures when a (potentially unrelated) command is dispatched, the logic should be added inside the plugin, that's why we designed it this way:
1. the command is dispatched to every plugin
2. each plugin, knowing its own structure and how it should behave, decided to handle the command or not and updates its state.

It's a sort of 'event-driven' programming where the event is a command.

So how do we differentiate the logic set in the actions and in the plugins?
It's a bit "case by case" but  you can follow the rule of thumb:
 If you want to react to a generic command (here REMOVE_COLUMNS_ROWS) which seems unrelated to the change you will make (like updating the position of the figure), then the change should take place in the plugin (here figure plugin)

the rest of the time, you should weight if a behaviour is based on a user will or not. Example: The insertion of graphs

Graphs are inserted  as a figure, to which we link some graph data (in opposition of images)
Some time a go, it was decided that we would insert a graph at the middle of the viewport of the user when they manually decide to insert a chart.

We could argue that the decision to insert the chart at the center of the viewport is not the user will (they did not choose that specific behaviour) but consider the following:
1. The viewport is a UI concept, as such ,it is _**user dependent**_, the user has in fact decided 
    to scroll to a specific place before inserting their figure
2. Let's assume we'd decide to insert a chart in the center of the viewport: 
    a. We _**have to be**_ able to insert graphs in other parts of the sheet 
    b. ( We cannot do it because we would intertwine UI and CORE logic ) as a reminder

Both points (1 and 2a) lead to the following conclusion:
- Inserting a graph should be handled in the plugin as it is the lowest level command
- inserting the chart to a place depending on the user position in the viewport (be it the scroll value or their current selection)  depends on them alone, it **_HAS_** to be handled in an action.

Long story short:  core changes that are user dependent should be handled in actions; core changes that depend on other core changes (like the sutation we have in this PR) should be handled in plugins.


 @somu-odoo  @dhrp-odoo @aath-odoo @khpa-odoo   SInce we still don't have set a proper  platform to exchange knowledge, I thought it could interest you

rrahir (changes requested):
As I mentioned during the session on Thursday, we should ignore the concept of hidden columns/rows for this. 

The reason being that once you consider it, you have to handle every hide/show  feature.
This includes the dynamic visibility of rows introduced by the data filters which is _**dynamic**_ and _**user related**_.  Since the change we are making is in **core**, we cannot support **user related** behaviour.

Now, you tried to only support the hide/show which is in **core**. that looks like a good idea in theory but if you do that, then you need to be consistent. If you consider modifying the place of a figure while considering the hidden rows/columns, then you also need to handle the commands that allow to hide/show them. 

If you do not handle the hide/show commands, you create a weird behaviour where you can hide rows ,it won't move the figure, then if you remove a single row ,the figure will magically move. So it's not a good option.

if you do handle the commands, then you will create confusion for the end user because the figure will be moved when the user manually hides rows but not when the rows are hidden by a filter. and since you cannot handle rows hidden by a filter, then this option is not viable either.

All in all, it's better to just ignore the hide/show at the moment.   In a next task, we will go a step further but as I explained on Thursday, it requires a much more in depth reshaping of the definition of the figures (they should not have an absolute position but rather should be 'hooked' to a cell). this cannot be done here for several reasons, the main one being that we still don't have a clear spec and we should take time to think about it.

So please, remove the references to the hidden columns. Based on that, you can drastically reduce the number of lines you introduced. Instead of reintroducing `getColDimensions` and `getColRowOffset` I think you can simply do a for loop that sums the values of `getColSize` as I showed you. your loop can look like 
```typescript
const numHeader = this.getters.getNumberRows(sheetId)
let gridSize = 0;

for (let i=0;i<numHeader;i++) {
    gridSize += this.getters.getRowSize(sheetId, i)
}
```

In the end, I'd expect something like this code:
```typescript
private onRowDeletion(sheetId: string, dimension: string) {
    const numHeader = this.getters.getNumberRows(sheetId)
     let gridhHeight = 0;
 
     for (let i=0;i<numHeader;i++) {
         gridhHeight += this.getters.getRowSize(sheetId, i)
     }
 
     for (const figure of figures) {
         const newY = Math.min(figure.y, gridhHeight - figure.height);
         if (newY !== figure.y) {
             this.dispatch("UPDATE_FIGURE", {sheetId, id: figure.id, figure: {y:newY}});
         }
     }
 }

private onColDeletion(sheetId: string, dimension: string) {
   const numHeader = this.getters.getNumberCols(sheetId)
    let gridWidth = 0;

    for (let i=0;i<numHeader;i++) {
        gridWidth += this.getters.getColSize(sheetId, i)
    }

    for (const figure of figures) {
        const newX = Math.min(figure.x, gridWidth - figure.width);
        if (newX !== figure.x) {
            this.dispatch("UPDATE_FIGURE", {sheetId, id: figure.id, figure: {x:newX}});
        }
    }
}
```
(I didn't test this tough)

rrahir:
Also the commit message (and therefore the PR description) need to be updated accordingly to the changes

src/plugins/core/figures.ts
```diff
@@ -53,6 +53,44 @@ export class FigurePlugin extends CorePlugin<FigureState> implements FigureState
...
+
+  private onRowDeletion(sheetId: string) {
+    const numHeader = this.getters.getNumberRows(sheetId);
+    let gridHeight = 0;
+    for (let i = 0; i < numHeader; i++) {
+      gridHeight += this.getters.getRowSize(sheetId, i);
+    }
+    const figures = this.getters.getFigures(sheetId);
+    for (const figure of figures) {
+      const newFigure = { ...figure };
```
rrahir, line 73: since you are only computing `newY` and dispatch it as a partial definition of a figure (line 76), you do not have to destructure it. Destructuring would make sense if you wanted to modify some parts of the object then use that modified object as a whole; but since we can simply dispatch isolated attributed in the [command](https://github.com/odoo/o-spreadsheet/blob/saas-16.2/src/types/commands.ts#L408), it is not necessary. Same goes for the other function.

rrahir (changes requested):
still a small change and we're there

src/plugins/core/figures.ts
jash-odoo: If you are referring to destructuring of figure > newFigure, then yes that indeed seems additional to me now, I have removed it and modified the code slightly. Let me know if this is what you meant. Thanks

PR #2081, opened 2023-02-16
[FIX] Evaluation: break laziness

rrahir:
The lazy evaluation is posing problems when trying to evaluate asynchronous
functions. More specifically, asynchronous formula are handled by first
starting the async computaiton and throwing an error. This error will be caught
by the evaluation engine and the cell will be marked as "ERROR" until the async
computation requests a new evaluation.

If a formula depends on the result of several asynchronous formula, the
evaluation engine will try to evaluate the dependencies one by one, each time
catching an error and stopping the evaluation.

This will force the asynchronous formula to be evaluated one by one, preventing
them from being evaluated in batch.

FIX
===
 We propose here to drop the "early fail" when fetching the dependencies
of a formula, which stops at the first caught error. This way, the first asynchronous
function will not prevent the following from being evaluated as well,
which will allow to batch them together at the next execution tick.

As soon as we catch an error, we continue the evaluation but still raise
it afterwards to conserve the current functional behavior.

Task 3165458

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui/evaluation.ts
+++ b/src/plugins/ui/evaluation.ts
@@ -135,6 +135,7 @@ export class EvaluationPlugin extends UIPlugin {
     for (const cell of this.getAllCells()) {
       computeCell(cell);
     }
+
     this.isUpToDate = true;
 
     function handleError(e: Error | any): EvaluationError {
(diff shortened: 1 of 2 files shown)


demo/main.js
```diff
@@ -170,9 +170,9 @@ class Demo extends Component {
       this.transportService = undefined;
       this.stateUpdateMessages = [];
     }
-    this.createModel(data || demoData);
+    // this.createModel(data || demoData);
```
pro-odoo, line 173: Nope

pro-odoo:
And runbot is 
image

src/plugins/ui/evaluation.ts
```diff
@@ -254,14 +255,26 @@ export class EvaluationPlugin extends UIPlugin {
...
         const rowValues: ({ value: CellValue; format?: Format } | undefined)[] = [];
         for (let row = zone.top; row <= zone.bottom; row++) {
-          const cell = evalContext.getters.getCell(range.sheetId, col, row);
-          rowValues.push(cell ? getEvaluatedCell(cell) : undefined);
+          const cell = evalContext.getters.getCell(range.sheetId, col, row) as Cell | undefined;
+          let evalCell: { value: CellValue; format?: Format } | undefined = undefined;
+          try {
+            evalCell = cell ? getEvaluatedCell(cell) : undefined;
+          } catch (e) {
+            evalError = e;
```
pro-odoo, line 268: evalError will be the last error. Now it's the first error which is thrown, but not sure it's a big deal

pro-odoo:
Not needed with https://github.com/odoo/odoo/pull/113642

PR #2082, opened 2023-02-17
[IMP] top_bar: added and updated tooltips

dhrp-odoo:
## Description:

This PR aims towards, the title on hover over toolbar icon was not added especially in drop down menus. Updated title on hover over toolbar icon.

Odoo task ID : [3171341](https://www.odoo.com/web#id=3171341&menu_id=4720&cids=2&action=4043&model=project.task&view_type=form)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/top_bar/top_bar.xml
+++ b/src/components/top_bar/top_bar.xml
@@ -41,7 +41,7 @@
           <div t-else="" class="o-toolbar-tools">
             <div
               class="o-tool"
-              title="Undo"
+              title="Undo (Ctrl+Z)"
               t-att-class="{'o-disabled': !undoTool}"
               t-on-click="undo">
               <t t-call="o-spreadsheet-Icon.UNDO"/>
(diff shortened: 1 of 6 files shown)


src/components/top_bar/top_bar.xml
```diff
@@ -41,15 +41,15 @@
           <div t-else="" class="o-toolbar-tools">
             <div
               class="o-tool"
-              title="Undo"
+              title="Undo (Ctrl+Z)"
```
rrahir, line 44: do not add the keybinds in the title

rrahir (changes requested):

src/components/top_bar/top_bar.xml
dhrp-odoo: I thought that the task was about, Hard copying the toolips from Gsheet. Which is why I added keybinds to the title. Should I remove from all places or just this one?

rrahir: It's not what I had in mind but you are right, it's better this way

anhe-odoo:
It may be a stupid question, but shouldn't we use something like `title="env._t('Title')"` to make the titles translatable ?
And maybe add also title for the common Format :)

rrahir:
@robodoo r+

PR #2093, opened 2023-02-22
[REF] port: run server on 9090 instead of 9000

SimonGenin:
The port 9000 is the default php-fpm port. So a person working on php may have trouble trying to launch the spreadsheets. We move to the port 9090 to avoid the problem.


SimonGenin:
Backported to 15.0 => https://github.com/odoo/o-spreadsheet/pull/2094

PR #2094, opened 2023-02-22
[REF] port: run server on 9090 instead of 9000

SimonGenin:
The port 9000 is the default php-fpm port. So a person working on php may have trouble trying to launch the spreadsheets. We move to the port 9090 to avoid the possible conflict.


SimonGenin:
@robodoo up to 16.0

Need to add the file_store.js starting from 16.1

LucasLefevre (approved):
robodoo r+

PR #2095, opened 2023-02-22
[IMP] config: update release script

rrahir:
- Restore package.json before the publication (it was updated when
  installing the gh action packages)
- chain package insstall instead of calling `npm i` twice
- more control on the artifacts that we release


LucasLefevre (approved):
robodoo r+

PR #2102, opened 2023-02-23
[FIX] charts: Fix chart duplicated ids

rrahir:
## [FIX] chart: Prevent destructed charts on duplicated sheet

The fix introduced in commit https://github.com/odoo/o-spreadsheet/commit/bb2d957c2e027dacd160e39f16ecc142c6486412 was incorrect when setting the
value of `ChartPlugin.nextId` when importing the data.
It only counted the occurences of existing charts without testing against
the values of the existing id. This could still lead to duplicates.

E.g.
- Create 2 charts with id "aa" and "bb"
- Duplicate sheet -> creates a figure & chart with ids "1" and "2"
- Delete the charts "aa", "bb" and "1"
- snapshot & reload data

When reloading the data after the import,  `ChartPlugin.nextId` will
ahve a value of "2". This means that duplicating the first sheet would
create another chart with id "2".

This commit replaces the `nextId` strategy and replaces it by prefixing
the ids of the charts of the previous sheet by the id of the new sheet.

e.g.
- create a sheet `myFirstSheet`
- create chart with id `myChartId`
- duplicate the sheet to a sheet with the id `myNewSheet`
the chart id in the new sheet will be `myNewSheet??myChartId`[1]

- duplicate `myNewSheet` to `myThirdSheet`
The chart id in the new sheet will be `myThirdSheet??myChartId`

[1] as the question mark is a forbidden character in the sheet names.

## [FIX] charts: Fix chart duplicated ids

The fix proposed in https://github.com/odoo/o-spreadsheet/commit/bb2d957c2e027dacd160e39f16ecc142c6486412 unfortunately didn't solve the situation
for users that already had duplicated chart ids. The main problem taking
its roots in the duplicated chart id accross different sheets, this
commit changes the chart plugin data structure to match the one of igure
plugin. I.e. a mapping of chart ids per sheet. This solution requires
no data modification and will prevent collisions

## [FIX] figure,chart: Prevent destructive creation

The fix introduced in commit https://github.com/odoo/o-spreadsheet/commit/bb2d957c2e027dacd160e39f16ecc142c6486412 was incorect as The allowDispatch
set on figures is utterly useless as we never dispatch the 'CREATE_FIGURE'
command from outside the plugin, it's always dispatched by `ChartPlugin`ATM.

This commit fixes the allowdispatch method in `FigurePlugin` for the form
but also introduces its equivalent in `ChartPlugin`.

Odoo task ID : [3141532](https://www.odoo.com/web#id=3141532&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes
- [ ]

### DIFF
--- a/src/plugins/core/chart.ts
+++ b/src/plugins/core/chart.ts
@@ -1,5 +1,11 @@
 import { FIGURE_ID_SPLITTER, INCORRECT_RANGE_STRING } from "../../constants";
-import { deepCopy, rangeReference, zoneToDimension, zoneToXc } from "../../helpers/index";
+import {
+  deepCopy,
+  isDefined,
+  rangeReference,
+  zoneToDimension,
+  zoneToXc,
+} from "../../helpers/index";
 import {
   ApplyRangeChange,
   ChartDefinition,
@@ -29,23 +35,30 @@ import { CorePlugin } from "../core_plugin";
  * */
 
 interface ChartState {
-  readonly chartFigures: Record<UID, ChartDefinition | undefined>;
+  readonly chartFigures: { [sheetId: UID]: Record<UID, ChartDefinition | undefined> | undefined };
 }
 
 export class ChartPlugin extends CorePlugin<ChartState> implements ChartState {
-  static getters = ["getChartDefinition", "getChartDefinitionUI", "getChartsIdBySheet"];
-  readonly chartFigures: Record<UID, ChartDefinition> = {};
+  static getters = ["getChartDefinition", "getChartDefinitionUI", "getChartDefinitionsBySheet"];
+  readonly chartFigures: ChartState["chartFigures"] = {};
 
   adaptRanges(applyChange: ApplyRangeChange) {
-    for (let [chartId, chart] of Object.entries(this.chartFigures)) {
-      if (chart) {
-        this.adaptDataSetRanges(chart, chartId, applyChange);
-        this.adaptLabelRanges(chart, chartId, applyChange);
+    for (const sheetId of Object.keys(this.chartFigures)) {
+      for (const [chartId, chart] of Object.entries(this.chartFigures[sheetId] || {})) {
+        if (chart) {
+          this.adaptDataSetRanges(sheetId, chart, chartId, applyChange);
+          this.adaptLabelRanges(sheetId, chart, chartId, applyChange);
+        }
       }
     }
   }
 
-  private adaptDataSetRanges(chart: ChartDefinition, chartId: UID, applyChange: ApplyRangeChange) {
+  private adaptDataSetRanges(
+    sheetId: UID,
+    chart: ChartDefinition,
+    chartId: UID,
+    applyChange: ApplyRangeChange
+  ) {
     for (let ds of chart.dataSets) {
       if (ds.labelCell) {
         const labelCellChange = applyChange(ds.labelCell);
@@ -53,6 +66,7 @@ export class ChartPlugin extends CorePlugin<ChartState> implements ChartState {
           case "REMOVE":
             this.history.update(
               "chartFigures",
+              sheetId,
               chartId,
               "dataSets",
               chart.dataSets.indexOf(ds),
@@ -65,6 +79,7 @@ export class ChartPlugin extends CorePlugin<ChartState> implements ChartState {
           case "CHANGE":
             this.history.update(
               "chartFigures",
+              sheetId,
               chartId,
               "dataSets",
               chart.dataSets.indexOf(ds),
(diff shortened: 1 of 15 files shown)


src/plugins/core/chart.ts
```diff
@@ -162,9 +160,13 @@ export class ChartPlugin extends CorePlugin<ChartState> implements ChartState {
         const sheetFiguresFrom = this.getters.getFigures(cmd.sheetId);
         for (const fig of sheetFiguresFrom) {
           if (fig.tag === "chart") {
-            const id = this.nextId.toString();
-            this.history.update("nextId", this.nextId + 1);
-            const chartDefinition = { ...deepCopy(this.chartFigures[fig.id]), id };
+            const figureIdBase = fig.id.split(FIGURE_ID_SPLITTER).pop();
+            const duplicateFigureId = [cmd.sheetIdTo, figureIdBase].join(FIGURE_ID_SPLITTER);
```
pro-odoo, line 200: What is more efficient ? Creating an array then join on this array, or 
```js
const duplicatedFigureId = `${cmd.sheetIdTo}${FIGURE_ID_SPLITTER}${figureIdBase}`
```
I do not have the answer, and I think we don't care, but if someone has the answer...

src/plugins/core/chart.ts
```diff
@@ -210,30 +233,32 @@ export class ChartPlugin extends CorePlugin<ChartState> implements ChartState {
...
   }
 
-  getChartsIdBySheet(sheetId: UID) {
-    return Object.entries(this.chartFigures)
-      .filter((chart) => {
-        return chart[1].sheetId === sheetId;
-      })
-      .map((chart) => chart[0]);
+  getChartDefinitionsBySheet(sheetId: UID) {
+    // TODORAR check where the helper is used
```
pro-odoo, line 250: TODORAR

src/plugins/core/chart.ts
```diff
@@ -210,30 +233,32 @@ export class ChartPlugin extends CorePlugin<ChartState> implements ChartState {
   // Getters
   // ---------------------------------------------------------------------------
 
-  getChartDefinition(figureId: UID): ChartDefinition | undefined {
-    return this.chartFigures[figureId];
+  getChartDefinition(sheetId: UID, figureId: UID): ChartDefinition | undefined {
```
pro-odoo, line 245: It's an API change in a getter in stable, but honestly I do not see another solution :/

tests/__snapshots__/xlsx.test.ts.snap
```diff
@@ -1220,16 +1220,17 @@ Object {
         <c:plotArea>
             <!-- how the chart element is placed on the chart -->
             <c:layout/>
-            <c:barChart>
-                <c:barDir val=\\"col\\"/>
-                <c:grouping val=\\"clustered\\"/>
-                <c:overlap val=\\"-20\\"/>
-                <c:gapWidth val=\\"70\\"/>
+            <c:lineChart>
```
pro-odoo, line 1223: seems louche, right ?

rrahir: it's actually logic given the fix. 
If you look at the related tests `multiple charts in the same sheet` for example, we were creating the 2 charts but with the same id. This was actually destroying the first one and replacing it with the second, but it would also do this for the figure so in the end we would only have 1 exported figure and chart.

For the other test (`charts in different sheets`) we wwere creating 2 charts on different sheets but with the same id - given the structure of `ChartPlugin`  we would again have only one chart existing BUT in 2 different figures. So you'(d have fig1 in sheet 1 and fig2 in sheet2 that pointed towards the same chart. Since we fix it here,  we now correctly have 2 different charts: 1 bar and 1 line

src/plugins/core/chart.ts
rrahir: the other solution consists in forcing the unicity of the figure (and not the combination of sheetid/figureid)  and store the sheetId in the figure definition ( go from  `{[sheetId]: Record<UID,figure>}` to  `Record<UID, {...figure, sheetId}>`but then we would need to migrate the data of the client, somehow convert the doubled ids, both in the data AND the revisions.  So unless we adapt the 'migrate' function to spit out the converted data as long as a table with the converted ids, I don't know how we could manage it

src/plugins/core/chart.ts
rrahir: the interpolation is somehow better I really don't think we will feel the difference though.

On 1000000 iteration:
join : 237 ms
interpolate: 178 ms

pro-odoo: Ok so we do not care :D

pro-odoo:
robodoo r+ rebase-ff

tests/plugins/figures.test.ts
```diff
@@ -328,7 +328,7 @@ describe("figure plugin", () => {
     expect(model.getters.getActiveCell()?.evaluated.value).toBeUndefined();
   });
 
-  test("cannot duplicate figure ids", () => {
+  test("cannot duplicate figure ids on the same sheet", () => {
```
hokolomopo, line 331: Maybe also a test wicth CREATE_CHART

hokolomopo (commented):
:+1:

PR #2106, opened 2023-02-24
[FIX] viewport: viewport maxSize should consider the client size

rrahir:
Task 3204720

## Description:

Odoo task ID : [3204720](https://www.odoo.com/web#id=3204720&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/internal_viewport.ts
+++ b/src/helpers/internal_viewport.ts
@@ -1,15 +1,5 @@
 import { DEFAULT_CELL_HEIGHT, DEFAULT_CELL_WIDTH } from "../constants";
-import {
-  Dimension,
-  Getters,
-  HeaderIndex,
-  Pixel,
-  Position,
-  Rect,
-  UID,
-  Zone,
-  ZoneDimension,
-} from "../types";
+import { Dimension, Getters, HeaderIndex, Pixel, Position, Rect, UID, Zone } from "../types";
 import { intersection, isInside } from "./zones";
 
 export class InternalViewport {
@@ -23,8 +13,8 @@ export class InternalViewport {
   offsetScrollbarY: Pixel;
   canScrollVertically: boolean;
   canScrollHorizontally: boolean;
-  width: Pixel;
-  height: Pixel;
+  viewportWidth: Pixel;
+  viewportHeight: Pixel;
   offsetCorrectionX: Pixel;
   offsetCorrectionY: Pixel;
 
@@ -36,8 +26,8 @@ export class InternalViewport {
     options: { canScrollVertically: boolean; canScrollHorizontally: boolean },
     offsets: { x: Pixel; y: Pixel }
   ) {
-    this.width = sizeInGrid.width;
-    this.height = sizeInGrid.height;
+    this.viewportWidth = sizeInGrid.width;
+    this.viewportHeight = sizeInGrid.height;
     this.offsetScrollbarX = offsets.x;
     this.offsetScrollbarY = offsets.y;
     this.canScrollVertically = options.canScrollVertically;
@@ -55,7 +45,11 @@ export class InternalViewport {
 
   // PUBLIC
 
-  getMaxSize(): ZoneDimension {
+  /** Returns the maximum size (in Pixels) of the viewport relative to its allocated client size
+   * When the viewport grid size is smaller than its client width (resp. height), it will return
+   * the client width (resp. height).
+   */
+  getMaxSize(): { width: Pixel; height: Pixel } {
     const lastCol = this.getters.findLastVisibleColRowIndex(this.sheetId, "COL", {
       first: this.boundaries.left,
       last: this.boundaries.right,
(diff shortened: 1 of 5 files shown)


tests/plugins/header_visibility.test.ts
```diff
@@ -95,10 +96,12 @@ describe("Hide Columns", () => {
         },
       ],
     });
+    model.dispatch("RESIZE_SHEETVIEW", { width: DEFAULT_CELL_WIDTH, height: 1000 });
     const sheet = model.getters.getActiveSheet();
     const dimensions = model.getters.getMainViewportRect();
+    let dimensions2: Rect;
     hideColumns(model, ["B", "C", "D"], sheet.id);
-    let dimensions2 = model.getters.getMainViewportRect();
+    dimensions2 = model.getters.getMainViewportRect();
```
hokolomopo, line 104: Why this change?

src/helpers/internal_viewport.ts
```diff
@@ -72,23 +72,29 @@ export class InternalViewport {
...
-        ? Math.max(DEFAULT_CELL_HEIGHT + 5, Math.min(topRowSize, this.height - lastRowSize))
-        : 0);
+    let width = lastColEnd - this.offsetCorrectionX;
+    if (this.canScrollHorizontally) {
+      width += Math.max(
+        DEFAULT_CELL_WIDTH,
+        Math.min(leftColSize, this.width - lastColSize),
+        this.width - width
+      );
+    }
```
hokolomopo, line 88: I'd add a comment to explain what all those computation mean. W/o prior knowledge of the viewport, it's really hard to get what's going on.

src/helpers/internal_viewport.ts
```diff
@@ -72,23 +72,29 @@ export class InternalViewport {
...
+
+    let height = lastRowEnd - this.offsetCorrectionY;
+    if (this.canScrollVertically) {
+      height += Math.max(
+        DEFAULT_CELL_HEIGHT + 5,
```
hokolomopo, line 93: +5 ?

tests/plugins/grid_manipulation.test.ts
```diff
@@ -1021,9 +1021,15 @@ describe("Rows", () => {
...
-      expect(dimensions).toMatchObject({ width: 192, height: 124 });
       expect(model.getters.getNumberRows(sheetId)).toBe(6);
+      const dimensions = model.getters.getMainViewportRect();
+      expect(dimensions).toMatchObject({ width: 1000, height: 1000 });
+      model.dispatch("RESIZE_SHEETVIEW", {
+        width: DEFAULT_CELL_WIDTH,
+        height: DEFAULT_CELL_HEIGHT,
+      });
+      const newDimensions = model.getters.getMainViewportRect();
+      expect(newDimensions).toMatchObject({ width: 192, height: 124 });
```
hokolomopo, line 1032: I'm gonna trust you on that, but why resizing the sheet to `{DEFAULT_CELL_WIDTH, DEFAULT_CELL_HEIGT}` leads to a size of `{192, 124}` ?

hokolomopo (commented):
Viewport computations are really obscure and confusing. We should at least add some comments to explain what's going on IMO

src/helpers/internal_viewport.ts
```diff
@@ -72,23 +66,29 @@ export class InternalViewport {
...
-      this.offsetCorrectionY +
-      (this.canScrollVertically
-        ? Math.max(DEFAULT_CELL_HEIGHT + 5, Math.min(topRowSize, this.height - lastRowSize))
-        : 0);
+    let width = lastColEnd - this.offsetCorrectionX;
+    if (this.canScrollHorizontally) {
+      width += Math.max(
+        DEFAULT_CELL_WIDTH, // leave some minimal space to let the user know they scrolled all the way
+        Math.min(leftColSize, this.clientWidth - lastColSize), // Add pixels that allows the snapping at maximum horizontal scroll
+        this.clientWidth - width // if the viewsport grid size is smaller than its client width, return client width
```
LucasLefevre, line 80: In this case, it just results in `this.clientWidth`, right ? (+width - width cancel out)
I think I'd just to `Math.max(width, this.clientWidth)` at the end to try simplify this already not so easy max(min).
The new formatting with comment is definitely better.

rrahir: agreed, I was a bit too eager to save some lines ^^

src/helpers/internal_viewport.ts
```diff
@@ -23,8 +13,8 @@ export class InternalViewport {
   offsetScrollbarY: Pixel;
   canScrollVertically: boolean;
   canScrollHorizontally: boolean;
-  width: Pixel;
-  height: Pixel;
+  clientWidth: Pixel;
```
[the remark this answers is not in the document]
rrahir: When I discussed this with Adrien, it turns out he did not really understand the meaning of that prop. the constructor argument `sizeInGrid` is a bit more explicit but yet fails to really describe what we want to; 
It is the visible size (width/height) of the viewport inside the grid, hence the idea to reuse the term clientWidth,  maybe visibleWidth/visibleHeight would be better ?

rrahir:
@LucasLefevre bump

src/helpers/internal_viewport.ts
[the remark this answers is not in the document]
rrahir: we are definitely mixing the two concepts. the reason we did this when introducing the frozen pane is that the concept of pane size seemed to only make sense for some viewports and not the others (it's not 100%  but in a real life situation, it is)
And since we only display the content of the viewport, we decided to gowith the vieport naming.

note:: while we named the class viewport, is HAS to be aware of it maximum size because of the scrolling stuff

A more accurate way to describ things would be to stick with the "Pane" as we define them in the core data and have the pane expose a viewport (for instance to use the getVisibleRect getters). but the sutation would be reversed. the Pane would HAVE to be aware of its visible part in the client (DOM) viewport.   Do you think it's more appropriate ? (I kinda do tbh)

rrahir:
@LucasLefevre updated with viewportWidth,Height :)

LucasLefevre (commented):
robodoo r+

PR #2111, opened 2023-02-27
[FIX] Top_bar: composer helper menu close

aath-odoo:
## Description:

Fixed composer helper menu not being closed on clicking topbar menu, resulting to refocus the composer.

Odoo task ID : [3076218](https://www.odoo.com/web#id=3076218&cids=2&menu_id=4720&action=4043&model=project.task&view_type=form)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/top_bar/top_bar.ts
+++ b/src/components/top_bar/top_bar.ts
@@ -379,6 +379,7 @@ export class TopBar extends Component<Props, SpreadsheetChildEnv> {
     this.state.menuState.parentMenu = menu;
     this.isSelectingMenu = true;
     this.openedEl = ev.target as HTMLElement;
+    this.env.model.dispatch("STOP_EDITION");
   }
 
   closeMenus() {
--- a/tests/components/spreadsheet.test.ts
+++ b/tests/components/spreadsheet.test.ts
@@ -339,6 +339,15 @@ describe("Composer interactions", () => {
     fixture.remove();
     app.destroy();
   });
+
+  test("The composer helper should be closed on toggle topbar context menu", async () => {
+    await typeInComposerGrid("=sum(");
+    expect(parent.model.getters.getEditionMode()).not.toBe("inactive");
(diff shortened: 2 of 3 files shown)


tests/components/composer.test.ts
```diff
@@ -1214,6 +1214,13 @@ describe("composer", () => {
     expect(fixture.querySelectorAll(".o-grid div.o-composer")).toHaveLength(0);
   });
 
+  test("The composer helper should be closed on toggle topbar context menu", async () => {
```
rrahir, line 1217: this test would be better placed in the file `top_bar.test.ts` as it tests the code of top_bar, not just the composer ;-)

rrahir (changes requested):
nice :) just a small change

tests/components/composer.test.ts
aath-odoo: @rrahir Updated requested changes please check, though i guess the scope of the issue is in both grid and topbar so if you suggest we can keep the test case for both in top_bar.test.ts and grid.test.ts.

rrahir: How is this affecting the grid?

aath-odoo: formula-assitant to be closed if opened in grid-composer as well.

rrahir: that seems to be a bit out of the scope of the task, isn't it?  Are you sure it's not tested yet? if not then yes it would make sense to test that the formula assistant closes in the grid as well. If you add this, be sure to update the commit message accordingly

aath-odoo: Okay i've added the test and updated commit message as well.

rrahir:
@aath-odoo I took the liberty to adapt the commit a bit. I don't know why but you put back the test in the composer file, I figure it was a mistake. Anyway I put the test in its rightful place [😊](https://emojipedia.org/smiling-face-with-smiling-eyes/)

rrahir:
@robodoo r+

PR #2120, opened 2023-02-28
[FIX] model: prevent UI plugins to refuse core commands

hokolomopo:
## Description

Currently it's possible for UI plugins to handle and refuse core commands in the allowDispatch. This is a problem because this could lead to de-synchronized states in multi-user.

This commit fixes this by making sure that the core commands don't go through the allowDispatch of the UI plugins.

Odoo task ID : [3209463](https://www.odoo.com/web#id=3209463&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_core_views/filter_evaluation.ts
+++ b/src/plugins/ui_core_views/filter_evaluation.ts
@@ -14,7 +14,6 @@ import {
   Border,
   CellPosition,
   CellValueType,
-  Command,
   CommandResult,
   ExcelFilterData,
   ExcelWorkbookData,
@@ -24,7 +23,7 @@ import {
   Zone,
 } from "../../types";
 import { UIPlugin } from "../ui_plugin";
-import { UpdateFilterCommand } from "./../../types/commands";
+import { Command, LocalCommand, UpdateFilterCommand } from "./../../types/commands";
 
 export class FilterEvaluationPlugin extends UIPlugin {
   static getters = [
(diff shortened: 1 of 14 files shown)


src/plugins/ui_core_views/filter_evaluation.ts
```diff
@@ -41,7 +40,7 @@ export class FilterEvaluationPlugin extends UIPlugin {
   hiddenRows: Set<number> = new Set();
   isEvaluationDirty = false;
 
-  allowDispatch(cmd: Command) {
```
hokolomopo, line 44: Note : I changed the allowDispatch interface of the current local plugins, but I couldn't find a way to type UI plugins correctly to prevent them from handling core commands.

We can do shit like that in ts, w/o typing error: 
image

LucasLefevre: You can't prevent that.

Nothing prevents a subclass to have a more general API, as long as it implements the parent's interface.

```ts
class A {
  coucou(a: number) {
    return true
  }
}
class B extends A {
  coucou(n: number | string) {
    return true
  }
}
```
Let's say I have `a: A;`  but `a` is actually an instance of `B` under the hood, but I don't know it. It's totally fine, I don't care that `a` can actually handle/do more stuff, I just use it as an `A`

However, if it doesn't have the same API than the parent, then it does work.
```ts
class B2 extends A {
  // error: Type 'number' is not assignable to type 'string'
  coucou(n: string) {
    return true
  }
}
```

Here since `LocalCommand` (parent API) is assignable to `LocalCommand | CoreCommand` (child class), everything is fine.

cc @anhe-odoo

PR #2121, opened 2023-02-28
[IMP] functions: DATEDIF and ADDRESS

yunchen4:
## Description:

This PR adds two functions: `ADDRESS` and `DATEDIF`. 
`ADDRESS` is used to get the string representation of a cell reference based on column number and row number.
`DATEDIF` is used to calculate the time difference between two dates.

Odoo task ID : [3201259](https://www.odoo.com/web#id=3201259&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/functions/helpers.ts
```diff
@@ -48,6 +72,17 @@ export function strictToNumber(value: string | number | boolean | null | undefin
   return toNumber(value);
 }
 
+export function strictToPositiveNumber(value: string | number | boolean | null | undefined) {
+  if (value === "") {
+    throw new Error(expectNumberValueError(value));
+  }
+  const number = toNumber(value);
+  if (number <= 0) {
+    throw new Error(expectPositiveNumberValueError(number));
```
laa-odoo, line 81: Here, if you create a helper function, your function can be used elsewhere later. So your error message must take into account more cases like decimal numbers between 0 and 1

I propse you the folowing error message:
```"The function [[FUNCTION_NAME]] expects a number value to be strictly positive, but receives %s.",```

File: src/functions/module_lookup.ts [hunk 2]
```diff
@@ -32,6 +36,58 @@ function assertAvailable(variable, searchKey) {
...
+  compute: function (
+    row: PrimitiveArgValue,
+    column: PrimitiveArgValue,
+    absoluteRelativeMode: ABSOLUTE_RELATIVE_MODE = ABSOLUTE_RELATIVE_MODE.ROW_ABSOLUTE_COL_ABSOLUTE,
+    useA1Notation: PrimitiveArgValue = true,
+    sheet: string | undefined
+  ): string {
+    const _useA1Notation = toBoolean(useA1Notation);
+    const rowNumber = strictToPositiveNumber(row);
+    const colNumber = strictToPositiveNumber(column);
```
laa-odoo, line 78: To follow up the previous review messages, you can write here that:
```
const rowNumber = toNumber(row)
const colNumber = toNumber(column)
assertStrictlyPositive(rowNumber)
assertStrictlyPositive(colNumber)
```

src/functions/module_lookup.ts
```diff
@@ -32,6 +36,58 @@ function assertAvailable(variable, searchKey) {
...
+    let cellReference = toA1Notation(rowNumber, colNumber, absoluteRelativeMode);
```
laa-odoo, line 79: with ```absoluteRelativeMode``` as PrimitiveArgValue you will need to transform it into a number (use ```toNumber```) and assert it is included between 1 and 4

tests/functions/module_lookup.test.ts
```diff
@@ -2,6 +2,75 @@ import { Model } from "../../src/model";
...
+  test("functional tests on using R1C1 notation and different absolute/relative modes", () => {
+    expect(evaluateCell("A1", { A1: "=ADDRESS(27,53,1,false)" })).toBe("R27C53");
+    expect(evaluateCell("A1", { A1: "=ADDRESS(27,53,2,false)" })).toBe("R27C[53]");
+    expect(evaluateCell("A1", { A1: "=ADDRESS(27,53,3,false)" })).toBe("R[27]C53");
+    expect(evaluateCell("A1", { A1: "=ADDRESS(27,53,4,false)" })).toBe("R[27]C[53]");
+  });
+
+  test("functional tests on sheet name", () => {
+    expect(evaluateCell("A1", { A1: '=ADDRESS(27,53,,,"sheet")' })).toBe("sheet!$BA$27");
+    expect(evaluateCell("A1", { A1: '=ADDRESS(27,53,,,"sheet!")' })).toBe("'sheet!'!$BA$27");
```
laa-odoo, line 69: greate :)

src/functions/module_date.ts
```diff
@@ -59,6 +81,103 @@ export const DATE: AddFunctionDescription = {
...
+      _lt(
+        `A text abbreviation for unit of time. Accepted values are "Y" (the number of whole years between start_date and end_date), "M" (the number of whole months between start_date and end_date), "D" (the number of days between start_date and end_date), "MD" (the number of days between start_date and end_date after subtracting whole months), "YM" (the number of whole months between start_date and end_date after subtracting whole years), "YD" (the number of days between start_date and end_date, assuming start_date and end_date were no more than one year apart).`
+      )
+    ),
+  ],
+  returns: ["NUMBER"],
+  compute: function (
+    startDate: PrimitiveArgValue,
+    endDate: PrimitiveArgValue,
+    unit: string
```
laa-odoo, line 113: ```unit: PrimitiveArgValue```

src/functions/helpers.ts
```diff
@@ -126,6 +161,59 @@ export function toJsDate(value: string | number | boolean | null | undefined): D
...
+  ROW_ABSOLUTE_COL_ABSOLUTE = 1,
+  ROW_ABSOLUTE_COL_RELATIVE = 2,
+  ROW_RELATIVE_COL_ABSOLUTE = 3,
+  ROW_RELATIVE_COL_RELATIVE = 4,
+}
+
+export function toA1Notation(
+  rowNumber: number,
+  colNumber: number,
+  absoluteRelativeMode: number = ABSOLUTE_RELATIVE_MODE.ROW_ABSOLUTE_COL_ABSOLUTE
```
anhe-odoo, line 174: Is there any reason not to use the `toXC` function ?
Then, if you really need it, you should maybe keep it "simple" with an object like `{rowFixed: boolean, colFixed: boolean}` as it's done in other part of the code (e.g. in range)

src/functions/helpers.ts
```diff
@@ -126,6 +161,59 @@ export function toJsDate(value: string | number | boolean | null | undefined): D
...
+    case ABSOLUTE_RELATIVE_MODE.ROW_ABSOLUTE_COL_ABSOLUTE:
+      cellReference = `$${colLetter}$${rowNumber}`;
+      break;
+    default:
+      throw new Error(expectNumberRangeError(1, 4, absoluteRelativeMode));
+  }
+  return cellReference;
+}
+
+export function fromA1NotationToR1C1Notation(A1Notation: string): string {
```
laa-odoo, line 197: Maybe a regexp is overkill here, let's avoid regexp. Maybe it would be better to compute R1C1 from rowNumber,  colNumber and absoluteRelativeMode too

anhe-odoo (commented):
Should be discussed with the team, but now that we have the new "RC" notation and the "A1" notation, maybe we should change the "XC" notation name in the code (with function as `toXC` which, in fact, let us convert from a zone to an "A1 notation") to avoid confusion ?

src/functions/helpers.ts
yunchen4: @laa-odoo Thanks! I'll change it to another type of error (to be greater than or equal to 1) because it's more clear to users 😄

src/functions/helpers.ts
yunchen4: @anhe-odoo Thanks! No special reason - simply because I don't remember we have this helper 😅 I'll use `toXC()` instead.
I'm not sure whether we should add `A1Notation` in all related existing method names. For now we only use R1C1 in `ADDRESS` so IMHO it's not necessary.

hokolomopo:
You should also add your functions to the sheet in the demo data :)

src/functions/module_lookup.ts
```diff
@@ -32,6 +38,67 @@ function assertAvailable(variable, searchKey) {
...
+      () => Object.values(ABSOLUTE_RELATIVE_MODE).includes(_absoluteRelativeMode),
+      expectNumberRangeError(1, 4, _absoluteRelativeMode)
+    );
+    const _useA1Notation = toBoolean(useA1Notation);
+    let cellReference: string;
+    if (_useA1Notation) {
+      cellReference = toA1Notation(rowNumber, colNumber, _absoluteRelativeMode);
+    } else {
+      cellReference = toR1C1Notation(rowNumber, colNumber, _absoluteRelativeMode);
+    }
```
laa-odoo, line 93: I wonder if the helpers aren't overkill here: I agree it helps to understand here, it's good to decompose, but I'm not sure, these helpers will ever be reused.

If we have the opportunity to concentrate the business logic of each formula, let's do it because there are many formulas.

Perhaps inserting the logic directly into the "if" with a shorter code will avoid the creation of new helpers. 
It would give something like this:

```javascript
if (_useA1Notation) {
  rangePart = { 
    rowFixed: [1,2].includes(absoluteRelativeMode) ? true : false,
    colFixed: [1,3].includes(absoluteRelativeMode) ? true : false,
  }
  cellReference = toXC(colNumber-1, rowNumber-1, rangePart)
} else {
  rowPart = [1,2].includes(absoluteRelativeMode) ? `R${rowNumber}` : `R[${rowNumber}]`;
  colPart = [1,3].includes(absoluteRelativeMode) ? `C${colNumber}` : `C[${colNumber}]`;
  cellReference = rowPart + colPart;
}
```

wait to see what others think before considering this change as relevant

LucasLefevre: I agree :+1: 
Even more given those modes that are unlikely to be used elsewhere

LucasLefevre:
robodoo delegate=laa-odoo

laa-odoo:
in this case, you can remove the enum ABSOLUTE_RELATIVE_MODE too
--> you can replace ```ABSOLUTE_RELATIVE_MODE.ROW_ABSOLUTE_COL_ABSOLUTE``` in the arg definition by a default constant defined at the start of the file ```DEFAULT_RELATIVE_MODE = 1;```

laa-odoo:
@robodoo r+

PR #2126, opened 2023-03-01
[IMP] history: add the possibility to repeat commands at redo

hokolomopo:
## Description:

Now if we do a redo and the redo stack is empty, we will try the redo
the last played command if possible. The list of repeatable commands
is in defined in the repeat_commands_registry.ts file.

Repeated commands are transformed to match the current selection/sheet.
To be able to repeat local commands, the "root" command of a revision
was added to the history.

Odoo task ID : [2508693](https://www.odoo.com/web#id=2508693&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/collaborative/revisions.ts
+++ b/src/collaborative/revisions.ts
@@ -1,4 +1,4 @@
-import { ClientId, CoreCommand, HistoryChange, RevisionData, UID } from "../types";
+import { ClientId, Command, CoreCommand, HistoryChange, RevisionData, UID } from "../types";
 
 export class Revision implements RevisionData {
   public readonly id: UID;
@@ -19,6 +19,7 @@ export class Revision implements RevisionData {
     id: UID,
     clientId: ClientId,
     commands: readonly CoreCommand[],
+    readonly rootCommand?: Command | "SNAPSHOT" | "REMOTE" | undefined,
     changes?: readonly HistoryChange[]
   ) {
     this.id = id;
--- a/src/collaborative/session.ts
+++ b/src/collaborative/session.ts
@@ -19,6 +19,7 @@ import {
   StateUpdateMessage,
   TransportService,
 } from "../types/collaborative/transport_service";
+import { Command, CommandTypes } from "./../types/commands";
 import { transformAll } from "./ot/ot";
 import { Revision } from "./revisions";
 
@@ -79,9 +80,15 @@ export class Session extends EventBus<CollaborativeEvent> {
    * Add a new revision to the collaborative session.
    * It will be transmitted to all other connected clients.
    */
-  save(commands: CoreCommand[], changes: HistoryChange[]) {
+  save(rootCommand: Command, commands: CoreCommand[], changes: HistoryChange[]) {
     if (!commands.length || !changes.length || !this.canApplyOptimisticUpdate()) return;
-    const revision = new Revision(this.uuidGenerator.uuidv4(), this.clientId, commands, changes);
+    const revision = new Revision(
+      this.uuidGenerator.uuidv4(),
+      this.clientId,
+      commands,
+      rootCommand,
+      changes
+    );
     this.revisions.append(revision.id, revision);
     this.trigger("new-local-state-update", { id: revision.id });
     this.sendUpdateMessage({
@@ -193,6 +200,19 @@ export class Session extends EventBus<CollaborativeEvent> {
     return this.pendingMessages.length === 0;
   }
 
+  /**
+   * Get the last local revision whose root command isn't in the given list of ignored commands
+   * */
+  getLastLocalNonEmptyRevision(ignoredRootCommands: CommandTypes[]): Revision | undefined {
+    const revisions = this.revisions.getRevertedExecution();
+    for (const rev of revisions) {
+      if (rev.rootCommand === "SNAPSHOT") return undefined;
+      if (!rev.rootCommand || rev.rootCommand === "REMOTE") continue;
+      if (!ignoredRootCommands.includes(rev.rootCommand?.type) && rev.commands.length) return rev;
+    }
+    return undefined;
+  }
+
   private _move(position: ClientPosition) {
     // this method is debounced and might be called after the client
     // left the session.
@@ -260,7 +280,7 @@ export class Session extends EventBus<CollaborativeEvent> {
           return;
         }
         const { clientId, commands } = message;
-        const revision = new Revision(message.nextRevisionId, clientId, commands);
+        const revision = new Revision(message.nextRevisionId, clientId, commands, "REMOTE");
         if (revision.clientId !== this.clientId) {
           this.revisions.insert(revision.id, revision, message.serverRevisionId);
           const pendingCommands = this.pendingMessages
@@ -273,7 +293,7 @@ export class Session extends EventBus<CollaborativeEvent> {
         }
         break;
       case "SNAPSHOT_CREATED": {
-        const revision = new Revision(message.nextRevisionId, "server", []);
+        const revision = new Revision(message.nextRevisionId, "server", [], "SNAPSHOT");
         this.revisions.insert(revision.id, revision, message.serverRevisionId);
         this.dropPendingHistoryMessages();
         this.trigger("snapshot");
--- a/src/history/factory.ts
+++ b/src/history/factory.ts
@@ -23,20 +23,22 @@ export function buildRevisionLog(
       revision.setChanges(changes);
     },
     (revision: Revision) => revertChanges([revision]),
-    (id: UID) => new Revision(id, "empty", [], []),
+    (id: UID) => new Revision(id, "empty", []),
     {
       with: (revision: Revision) => (toTransform: Revision) => {
         return new Revision(
           toTransform.id,
           toTransform.clientId,
-          transformAll(toTransform.commands, revision.commands)
+          transformAll(toTransform.commands, revision.commands),
+          toTransform.rootCommand
         );
       },
       without: (revision: Revision) => (toTransform: Revision) => {
         return new Revision(
           toTransform.id,
           toTransform.clientId,
-          transformAll(toTransform.commands, revision.commands.map(inverseCommand).flat())
+          transformAll(toTransform.commands, revision.commands.map(inverseCommand).flat()),
+          toTransform.rootCommand
         );
       },
     }
--- a/src/history/repeat_commands/repeat_commands_generic.ts
+++ b/src/history/repeat_commands/repeat_commands_generic.ts
@@ -1,0 +1,30 @@
+import { deepCopy } from "../../helpers";
+import { Command, Getters } from "../../types";
+
+export const genericRepeatsTransforms = [
+  repeatSheetDependantCommand,
+  repeatTargetDependantCommand,
+  repeatPositionDependantCommand,
+];
+
+export function repeatSheetDependantCommand<T extends Command>(getters: Getters, command: T): T {
+  if (!("sheetId" in command)) return command;
+
+  return { ...deepCopy(command), sheetId: getters.getActiveSheetId() };
+}
+
+export function repeatTargetDependantCommand<T extends Command>(getters: Getters, command: T): T {
+  if (!("target" in command) || !Array.isArray(command.target)) return command;
+
+  return {
+    ...deepCopy(command),
+    target: getters.getSelectedZones(),
+  };
+}
+
+export function repeatPositionDependantCommand<T extends Command>(getters: Getters, command: T): T {
+  if (!("row" in command) || !("col" in command)) return command;
+
+  const { col, row } = getters.getActivePosition();
+  return { ...deepCopy(command), col, row };
+}
--- a/src/history/repeat_commands/repeat_commands_specific.ts
+++ b/src/history/repeat_commands/repeat_commands_specific.ts
@@ -1,0 +1,147 @@
+import { deepCopy, range, UuidGenerator } from "../../helpers";
+import { Getters } from "../../types";
+import {
+  AddColumnsRowsCommand,
+  AutoresizeColumnsCommand,
+  AutoresizeRowsCommand,
+  CreateChartCommand,
+  CreateFigureCommand,
+  CreateImageOverCommand,
+  CreateSheetCommand,
+  DeleteCellCommand,
+  HideColumnsRowsCommand,
+  InsertCellCommand,
+  PasteCommand,
+  RemoveColumnsRowsCommand,
+  RepeatPasteCommand,
+  ResizeColumnsRowsCommand,
+  SortCommand,
+} from "./../../types/commands";
+import { repeatSheetDependantCommand } from "./repeat_commands_generic";
+
+const uuidGenerator = new UuidGenerator();
+
+export function repeatCreateChartCommand(
+  getters: Getters,
+  cmd: CreateChartCommand
+): CreateChartCommand {
+  return {
+    ...repeatSheetDependantCommand(getters, cmd),
+    id: uuidGenerator.uuidv4(),
+  };
+}
+
+export function repeatCreateImageCommand(
+  getters: Getters,
+  cmd: CreateImageOverCommand
+): CreateImageOverCommand {
+  return {
+    ...repeatSheetDependantCommand(getters, cmd),
+    figureId: uuidGenerator.uuidv4(),
+  };
+}
+
+export function repeatCreateFigureCommand(
+  getters: Getters,
+  cmd: CreateFigureCommand
+): CreateFigureCommand {
+  const newCmd = repeatSheetDependantCommand(getters, cmd);
+  newCmd.figure.id = uuidGenerator.uuidv4();
+  return newCmd;
+}
+
+export function repeatCreateSheetCommand(
+  getters: Getters,
+  cmd: CreateSheetCommand
+): CreateSheetCommand {
+  const newCmd = deepCopy(cmd);
+  newCmd.sheetId = uuidGenerator.uuidv4();
+
+  const sheetName = cmd.name || getters.getSheet(getters.getActiveSheetId()).name;
+  // Extract the prefix of the sheet name (everything before the number at the end of the name)
+  const namePrefix = sheetName.match(/(.+?)\d*$/)?.[1] || sheetName;
+
+  newCmd.name = getters.getNextSheetName(namePrefix);
+
+  return newCmd;
+}
+
+export function repeatAddColumnsRowsCommand(
+  getters: Getters,
+  cmd: AddColumnsRowsCommand
+): AddColumnsRowsCommand {
+  const currentPosition = getters.getActivePosition();
+  return {
+    ...repeatSheetDependantCommand(getters, cmd),
+    base: cmd.dimension === "COL" ? currentPosition.col : currentPosition.row,
+  };
+}
+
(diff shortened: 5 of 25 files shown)


tests/plugins/repeat_commands.test.ts
```diff
@@ -0,0 +1,504 @@
...
+    TEST_COMMANDS.REMOVE_FILTER_TABLE,
+    TEST_COMMANDS.HIDE_SHEET,
+  ])("Sheet dependant command are adapted to current sheet %s", (command: CoreCommand) => {
+    createSheet(model, { sheetId: "42" });
+    activateSheet(model, "42");
+    const transformed = getRepeatedCoreCommands(model.getters, [command])[0];
+    expect(transformed?.sheetId).toEqual("42");
+  });
+
+  test.each([TEST_COMMANDS.UPDATE_CELL!, TEST_COMMANDS.CLEAR_CELL!, TEST_COMMANDS.SET_BORDER])(
```
anhe-odoo, line 149: Why have you kept these "!" for UPDATE_CELL and CLEAR_CELL ? Is there something special with these commands ?

src/collaborative/session.ts
```diff
@@ -171,6 +178,8 @@ export class Session extends EventBus<CollaborativeEvent> {
       data: { ...data, revisionId: snapshotId },
       version: MESSAGE_VERSION,
     });
+    const revs = this.revisions.getRevertedExecution();
+    console.log(revs);
```
anhe-odoo, line 182: console.log ?

src/history/repeat_commands/repeat_commands_generic.ts
```diff
@@ -0,0 +1,39 @@
...
+  if (!("sheetId" in command)) return command;
+
+  return { ...deepCopy(command), sheetId: getters.getActiveSheetId() };
+}
+
+export function transformTargetDependantCommand<T extends Command>(
+  getters: Getters,
+  command: T
+): T {
+  if (!("target" in command) || !Array.isArray(command.target)) return command;
```
anhe-odoo, line 20: Why is it needed to be an Array ? To ensure it's a range ? 
And shouldn't we check that the target isn't a fixed col/row zone ?

src/history/repeat_commands/repeat_commands_specific.ts
```diff
@@ -0,0 +1,156 @@
...
+  RemoveColumnsRowsCommand,
+  RepeatPasteCommand,
+  ResizeColumnsRowsCommand,
+  SortCommand,
+} from "./../../types/commands";
+import { transformSheetDependantCommand } from "./repeat_commands_generic";
+
+const uuidGenerator = new UuidGenerator();
+
+export function repeatCreateChartCommand(
```
anhe-odoo, line 26: Probably another stupid question, but can't a chart also be a "target" dependant command ? If we select a zone and make a CREATE_CHART, then select another zone and CLTR+Y ?

src/plugins/ui_stateful/clipboard.ts
```diff
@@ -152,10 +155,19 @@ export class ClipboardPlugin extends UIPlugin {
...
-        const state = new ClipboardOsState(cmd.text, this.getters, this.dispatch, this.selection);
-        state.paste(cmd.target);
+        this.state = new ClipboardOsState(cmd.text, this.getters, this.dispatch, this.selection);
+        this.state.paste(cmd.target);
+        this.lastPasteState = this.state;
         this.status = "invisible";
         break;
+      case "REPEAT_PASTE": {
+        this.lastPasteState?.paste(cmd.target, {
+          pasteOption: cmd.pasteOption,
```
anhe-odoo, line 165: As pasteOption can be undefined, shouldn't we keep the option of the "last" past ? Maybe it's the case and I simply missed it

anhe-odoo (commented):
Good work. Some small questions but else look good to me :)

src/history/repeat_commands/repeat_commands_generic.ts
hokolomopo: I need to check that it's an array because of the `PasteCFCommand`, which have a target that is a cell... I should probably just change this command, it's not coherent to have command whose target isn't like all the others.

Target are an array of `Zone`, and there is no concept of fixed col/row in `Zone`.

hokolomopo: Added a commit to rename the payload of `PasteCFCommand`. No miragtion needed since it's a UI command.

src/history/repeat_commands/repeat_commands_specific.ts
```diff
@@ -0,0 +1,156 @@
...
+  return {
+    ...transformSheetDependantCommand(getters, cmd),
+    elements: [cmd.dimension === "COL" ? currentSelection[0].left : currentSelection[0].top],
+  };
+}
+
+export function repeatInsertOrDeleteCellCommand<T extends InsertCellCommand | DeleteCellCommand>(
+  getters: Getters,
+  cmd: T,
+  childCommands: readonly CoreCommand[]
```
LucasLefevre, line 106:
```suggestion
```
Remove when it's unused

src/history/repeat_commands/repeat_commands_specific.ts
```diff
@@ -0,0 +1,156 @@
...
+): T {
+  const currentSelection = getters.getSelectedZones();
+  return {
+    ...deepCopy(cmd),
+    zone: currentSelection[0],
+  };
+}
+
+export function repeatAutoresizeCommand<T extends AutoresizeColumnsCommand | AutoresizeRowsCommand>(
```
LucasLefevre, line 115:
```suggestion
export function repeatAutoResizeCommand<T extends AutoresizeColumnsCommand | AutoresizeRowsCommand>(
```
Let's not repeat mistakes from the past

src/registries/repeat_commands_registry.ts
```diff
@@ -0,0 +1,149 @@
...
+export function getRepeatedLocalCommandGeneric<T extends LocalCommand>(
+  getters: Getters,
+  cmd: T,
+  childCommands: readonly CoreCommand[]
+): T | undefined {
+  return genericTransform(getters, cmd);
+}
+
+/** Apply generic transforms (sheetID, target, ...) to the core command children of a local command */
+export function getRepeatedLocalCommandChildrenGeneric<T extends LocalCommand>(
```
LucasLefevre, line 107: this function is only used once for STOP_EDITION. maybe we could just move (and rename) this to a specific repeat function `repeatStopEdition`. It would be easier to understand than this kinda barbaric function name

src/types/commands.ts
```diff
@@ -501,6 +501,12 @@ export interface PasteCommand {
   pasteOption?: ClipboardPasteOptions;
 }
 
+export interface RepeatPasteCommand {
+  type: "REPEAT_PASTE";
```
LucasLefevre, line 505: there are two different strategies for the two repeatable commands from state-full plugins (STOP_EDITION and PASTE)
PASTE => new command + new state in the plugin
STOP_EDITION => check core command

Couldn't paste work similarly to STOP_EDITION ?

src/history/repeat_commands/repeat_revision.ts
```diff
@@ -0,0 +1,38 @@
...
+  getters: Getters
+): CoreCommand[] | LocalCommand | undefined {
+  const commands = revision.commands;
+  if (!revision.rootCommand || typeof revision.rootCommand !== "object") {
+    return undefined;
+  }
+
+  if (isCoreCommand(revision.rootCommand)) {
+    return getRepeatedCoreCommands(getters, commands);
+  }
```
LucasLefevre, line 35: if `rootCommand` is a core command, than `commands` is `[revision.rootCommand]`.
Maybe it's more straight forward to understand/explicit like this? What do you think ?
```suggestion
  if (isCoreCommand(revision.rootCommand)) {
    return getRepeatedCoreCommands(getters, [revision.rootCommand]);
  }
```

src/model.ts
```diff
@@ -211,13 +210,20 @@ export class Model extends EventBus<any> implements CommandDispatcher {
...
-      canUndo: this.history.canUndo.bind(this.history),
-      canRedo: this.history.canRedo.bind(this.history),
       getClient: this.session.getClient.bind(this.session),
       getConnectedClients: this.session.getConnectedClients.bind(this.session),
       isFullySynchronized: this.session.isFullySynchronized.bind(this.session),
     } as Getters;
 
+    this.history = new LocalHistory(
+      this.dispatchFromCorePlugin,
+      this.dispatch,
```
LucasLefevre, line 220: I really feel dis-confortable with this. It drives us even more and more away from the plugin architecture. Going back will be harder and harder if we continue in this direction with funky/non-standard stuff.

If `this.session.on("revision-undone" ...` inside `LocalHistory` is moved to another (core) plugin (with session made available in the plugins config), we wouldn't need this double dispatch function and `LocalHistory` could be a "real" UI plugin ?

LucasLefevre (commented):
coucou :)

src/history/factory.ts
```diff
@@ -23,20 +23,22 @@ export function buildRevisionLog(
...
-    (id: UID) => new Revision(id, "empty", [], []),
+    (id: UID) => new Revision(id, "empty", [], undefined, []),
     {
       with: (revision: Revision) => (toTransform: Revision) => {
         return new Revision(
           toTransform.id,
           toTransform.clientId,
-          transformAll(toTransform.commands, revision.commands)
+          transformAll(toTransform.commands, revision.commands),
+          toTransform.rootCommand
```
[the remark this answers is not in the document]
hokolomopo: The root command itself yeah, but we can still repeat it since we adapt the sheetId/target to the current sheet/current selection.

Except for `DUPLICATE_SHEET`, but maybe it should duplicate the active sheet, and not re-duplicate the sheet previously duplicated.

Do you see a problem with that ?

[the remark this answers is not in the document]
LucasLefevre: > Except for DUPLICATE_SHEET, but maybe it should duplicate the active sheet, and not re-duplicate the sheet previously duplicated

Or simply not repeatable ? It looks like it's not repeatable on GSheet

src/history/repeat_commands/repeat_commands_specific.ts
```diff
@@ -0,0 +1,156 @@
...
+  getters: Getters,
+  cmd: CreateImageOverCommand
+): CreateImageOverCommand {
+  return {
+    ...transformSheetDependantCommand(getters, cmd),
+    figureId: uuidGenerator.uuidv4(),
+  };
+}
+
+export function repeatCreateFigureCommand(
```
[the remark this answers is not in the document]
hokolomopo: Yeah. Still added it in case we do standalone `CREATE_FIGURE` in the future, but I can remove it if you want.

src/registries/repeat_commands_registry.ts
```diff
@@ -0,0 +1,149 @@
...
+ *  Registry containing all the command that can be repeated on redo, and function to transform them
+ *  to the current state of the model.
+ *
+ * If the transform function is undefined, the command will be transformed using generic transformations.
+ * (change the sheetId, the row, the col, the target, the ranges, to the current active sheet & selection)
+ *
+ */
+export const repeatCommandTransformRegistry = new Registry<RepeatTransform | undefined>();
+
+repeatCommandTransformRegistry.add("UPDATE_CELL", undefined);
```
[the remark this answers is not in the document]
hokolomopo: Yeah probably better.

`repeatableLocalCommandsRegistry ` was inconstent becuase it have 2 generic trnaform : either transorm the root command, or transform all the child commands.

src/history/repeat_commands/repeat_commands_generic.ts
```diff
@@ -0,0 +1,39 @@
+import { deepCopy } from "../../helpers";
+import { Command, Getters } from "../../types";
+
+export const genericTransforms = [
+  transformSheetDependantCommand,
+  transformTargetDependantCommand,
+  transformPositionDependantCommand,
+];
+
+export function transformSheetDependantCommand<T extends Command>(getters: Getters, command: T): T {
```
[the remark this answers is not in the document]
hokolomopo: Maybe `repeatTransform` ? I'm not too sure about `repeat` alone...

tests/plugins/history.test.ts
```diff
@@ -204,11 +205,28 @@ describe("Model history", () => {
     expect(getCell(model, "A2")).toBeUndefined();
 
     expect(model.getters.canUndo()).toBe(false);
-    expect(model.getters.canRedo()).toBe(true);
+    expect(model["history"]["redoStack"].length).toBe(1);
```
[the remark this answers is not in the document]
hokolomopo: Yeah we are definetly testing an internal behaviour now. I can keep it like that or remove the test if you want, I don't really care.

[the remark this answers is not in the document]
LucasLefevre: you can remove it ;)

src/registries/repeat_commands_registry.ts
hokolomopo: If I don't need to use it as a generic for odoo commands, I'll do that yeah.

src/types/commands.ts
hokolomopo: No. I tried to have a behaviour like that at first, but I couldn't do it correctly. It works for stop edition, because the behaviour is simple, but the core commands dispatched by paste depends on where we paste/the state of the spreadsheet. Some exmples : 

- paste at the edge of the sheet. We will add new col/rows to not paste something out of the sheet. But now if we repeat the command, we don't want to add new col rows.
    - the inserse is also true. If we repeat at the edge of the sheet, we want to add new col/rows.
- similar problem with filters (maybe with merge too?)
-  when pasting CF, I have no idea looking at the `ADD_CONDITIONAL_FORMAT` what ranges were added by paste, and what ranges were there before

src/history/factory.ts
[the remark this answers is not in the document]
hokolomopo: Probably easier yeah

src/collaborative/revisions.ts
```diff
@@ -19,6 +19,7 @@ export class Revision implements RevisionData {
     id: UID,
     clientId: ClientId,
     commands: readonly CoreCommand[],
+    readonly rootCommand: Command | "SNAPSHOT" | "REMOTE" | undefined,
```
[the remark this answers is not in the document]
hokolomopo: For remote revisions yes, but then I do'nt know if the revision is a snapshot (in which case I should stop and we cannot repeat a command)

src/model.ts
hokolomopo: Chanegd history into a real UI plugin, and handled the `resvision-redone/undone` inside the model.

src/types/commands.ts
LucasLefevre: ok thanks for explaining. Could you add the explanation in the commit message so that others reading the commit will understand our choices. People won't read the 50+ comments on the PR

tests/test_helpers/constants.ts
```diff
@@ -43,3 +45,140 @@ export const TEST_CHART_DATA = {
...
+  ADD_COLUMNS_ROWS: {
+    type: "ADD_COLUMNS_ROWS",
+    position: "after",
+    dimension: "ROW",
+    base: 0,
+    quantity: 1,
+    sheetId: "sheetId",
+  },
+};
+export const TEST_COMMANDS = TEST_COMMANDS_PARTIAL as CommandMapping;
```
LucasLefevre, line 184: this could use the `satisfies` operator introduced in Typescript 4.9
... for later

src/model.ts
```diff
@@ -196,8 +194,6 @@ export class Model extends EventBus<any> implements CommandDispatcher {
 
     this.session = this.setupSession(workbookData.revisionId);
 
-    this.history = new LocalHistory(this.dispatchFromCorePlugin, this.session);
```
LucasLefevre, line 199: yeaaaah, that's nice :partying_face:

LucasLefevre (commented):
bloup bloup :)
Almost ready :)

src/plugins/ui_plugin.ts
```diff
@@ -40,11 +42,13 @@ export class UIPlugin<State = any, C = Command> extends BasePlugin<State, C> {
...
   protected ui: UIActions;
   protected selection: SelectionStreamProcessor;
-  constructor({ getters, stateObserver, dispatch, uiActions, selection }: UIPluginConfig) {
+  protected session: Session;
+  constructor({ getters, stateObserver, dispatch, uiActions, selection, session }: UIPluginConfig) {
     super(stateObserver, dispatch);
     this.getters = getters;
     this.ui = uiActions;
     this.selection = selection;
+    this.session = session;
```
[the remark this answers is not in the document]
hokolomopo: makes sense

src/registries/repeat_commands_registry.ts
```diff
@@ -0,0 +1,128 @@
...
+): CoreCommand[] {
+  const repeatedCommands: CoreCommand[] = [];
+  for (const cmd of commands) {
+    const isRepeatable = repeatCommandTransformRegistry.contains(cmd.type);
+    if (!isRepeatable) continue;
+
+    const specificTransform = repeatCommandTransformRegistry.get(cmd.type);
+    if (!specificTransform) {
+      const result = genericRepeat(getters, cmd);
+      repeatedCommands.push(result);
```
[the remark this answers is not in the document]
hokolomopo: Yup :)

src/registries/repeat_commands_registry.ts
hokolomopo: Looks like we don't need that in odoo, changed it to `repeatStopEdition`

tests/test_helpers/constants.ts
hokolomopo: When we have upgraded our version of TS to 4.9 I might have a look at what it does 😅

rrahir:
forgot to mention the circular dependency at build :/

src/registries/repeat_commands_registry.ts
hokolomopo: Edit: cannot make it into a specific transform because of cyclic imports. `repeat_commands_registry.ts` need `repeatStopEdition` to put in the registry => `repeatStopEdition` need `repeatCoreCommand` => `repeatCoreCommand` need the registry. 

I put it back in this file, but changed the name to `repeatLocalCommandChildren` which is elss of a mouthful

src/collaborative/session.ts
```diff
@@ -193,6 +200,17 @@ export class Session extends EventBus<CollaborativeEvent> {
     return this.pendingMessages.length === 0;
   }
 
+  /** Get the last local revision that doesn't come from a REQUEST_REDO command */
+  getLastLocalNonRedoRevision(): Revision | undefined {
+    const revisions = this.revisions.getRevertedExecution();
```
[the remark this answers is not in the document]
hokolomopo: After looking a bit into that, if I understand correctly revisions from standard undo/redo are created inside the selective_history, and are empty. So they are not retuened in this method. But this isn't very intuitive, so I changed things a bit. 

LMK if the code looks clearer now

LucasLefevre:
robodoo rebase-ff bordel :grimacing:

PR #2140, opened 2023-03-01
[FIX] sheet: prevent deleting all visible columns/rows

hokolomopo:
## [FIX] sheet: prevent deleting all visible columns/rows

We currently prevent deleting all columns, but we can still delete all the visible columns, which leads to a broken sheet with no column.

This commit prevent it in the allowDisatch of sheet.ts, and also disable the button in the header context menu.

## [FIX] header_overlay: select col when there is a single row

The getter `findFirstVisibleColRowIndex` was bugged and skipped the last
col/row of the sheet. This meant that we couldn't select a row (or col)
if there was a single col (or row) in the sheet.

## [FIX] menu: dynamically update visible menu items

Before this commit, the menu visible menu items were given as props of the
`Menu` component. This meant that contrarily to the property `isEnabled` of
the menu items, the `isVisible` property was computed once when opening
the menu, and never updated.

This commit fixes this issue by making the `Menu` component compute the
`isVisible` property at each render, dynamically updating which menu items
are visible based on the state of the model.

Odoo task ID : [3204730](https://www.odoo.com/web#id=3204730&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/bottom_bar/bottom_bar.ts
+++ b/src/components/bottom_bar/bottom_bar.ts
@@ -189,7 +189,7 @@ export class BottomBar extends Component<Props, SpreadsheetChildEnv> {
 
   openContextMenu(x: Pixel, y: Pixel, registry: MenuItemRegistry, menuId?: UID) {
     this.menuState.isOpen = true;
-    this.menuState.menuItems = registry.getAll().filter((x) => x.isVisible(this.env));
+    this.menuState.menuItems = registry.getAll();
     this.menuState.position = { x, y };
     this.menuState.menuId = menuId;
   }
--- a/src/components/figures/figure_chart/figure_chart.ts
+++ b/src/components/figures/figure_chart/figure_chart.ts
@@ -124,7 +124,7 @@ export class ChartFigure extends Component<Props, SpreadsheetChildEnv> {
   private openContextMenu(position: DOMCoordinates) {
     const registry = this.getMenuItemRegistry();
     this.menuState.isOpen = true;
-    this.menuState.menuItems = registry.getAll().filter((x) => x.isVisible(this.env));
+    this.menuState.menuItems = registry.getAll();
     this.menuState.position = position;
   }
 
--- a/src/components/grid/grid.ts
+++ b/src/components/grid/grid.ts
@@ -491,9 +491,7 @@ export class Grid extends Component<Props, SpreadsheetChildEnv> {
     this.closeOpenedPopover();
     this.menuState.isOpen = true;
     this.menuState.position = { x, y };
-    this.menuState.menuItems = registries[type]
-      .getAll()
-      .filter((item) => !item.isVisible || item.isVisible(this.env));
+    this.menuState.menuItems = registries[type].getAll();
   }
 
   copy(cut: boolean, ev: ClipboardEvent) {
--- a/src/components/menu/menu.ts
+++ b/src/components/menu/menu.ts
@@ -111,6 +111,10 @@ export class Menu extends Component<Props, SpreadsheetChildEnv> {
     });
   }
 
+  get visibleMenuItems(): MenuItem[] {
+    return this.props.menuItems.filter((x) => x.isVisible(this.env));
+  }
+
   get subMenuPosition(): DOMCoordinates {
     const position = Object.assign({}, this.subMenu.position);
     position.y -= this.subMenu.scrollOffset || 0;
@@ -118,7 +122,7 @@ export class Menu extends Component<Props, SpreadsheetChildEnv> {
   }
 
   get menuHeight(): Pixel {
-    return this.menuComponentHeight(this.props.menuItems);
+    return this.menuComponentHeight(this.visibleMenuItems);
   }
 
   get subMenuHeight(): Pixel {
@@ -159,7 +163,7 @@ export class Menu extends Component<Props, SpreadsheetChildEnv> {
    * and the menu item at a given index.
    */
   private subMenuVerticalPosition(position: Pixel): Pixel {
-    const menusAbove = this.props.menuItems.slice(0, position);
+    const menusAbove = this.visibleMenuItems.slice(0, position);
     return this.menuComponentHeight(menusAbove) + this.position.y;
   }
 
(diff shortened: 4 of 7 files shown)


src/plugins/core/header_visibility.ts
```diff
@@ -91,6 +92,16 @@ export class HeaderVisibilityPlugin extends CorePlugin {
     return this.hiddenHeaders[sheetId].COL[index];
   }
 
+  getAllVisibleHeaders(sheetId: UID, dimension: Dimension): HeaderIndex[] {
+    const visibleHeaders: HeaderIndex[] = [];
+    for (let i = 0; i < this.hiddenHeaders[sheetId][dimension].length; i++) {
```
rrahir, line 97: I think the allow dispatch should occur in this plugin rather than in core/sheet

src/registries/menus/row_menu_registry.ts
```diff
@@ -67,11 +72,8 @@ rowMenuRegistry
     action: ACTIONS.HIDE_ROWS_ACTION,
     isVisible: (env: SpreadsheetChildEnv) => {
       const sheetId = env.model.getters.getActiveSheetId();
-      const hiddenRows = env.model.getters.getHiddenRowsGroups(sheetId).flat();
-      return (
-        env.model.getters.getNumberRows(sheetId) >
-        hiddenRows.length + env.model.getters.getElementsFromSelection("ROW").length
-      );
+      const visibleRows = env.model.getters.getAllVisibleHeaders(sheetId, "ROW").length;
+      return visibleRows > env.model.getters.getElementsFromSelection("ROW").length;
```
anhe-odoo, line 76: Same question about the case were we have more hidden columns than visible columns (e.g. B to C hidden) and we select A to D. Will `getElementsFromSelection` only contains `[A, D]` or ´[A, B, C, D]` ?

anhe-odoo (commented):
First review, will continue tomorrow but you can already have a look at my questions :)

src/plugins/core/header_visibility.ts
```diff
@@ -31,6 +32,15 @@ export class HeaderVisibilityPlugin extends CorePlugin {
...
           : CommandResult.TooManyHiddenElements;
       }
+      case "REMOVE_COLUMNS_ROWS":
+        if (!this.hiddenHeaders[cmd.sheetId]) {
+          return CommandResult.InvalidSheetId;
+        }
+        const length = this.getters.getAllVisibleHeaders(cmd.sheetId, cmd.dimension).length;
+        if (length <= cmd.elements.length) {
+          return CommandResult.NotEnoughElements;
+        }
```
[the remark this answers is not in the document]
hokolomopo: You're 100% right actually, all my conditions are false 🙃

src/plugins/core/header_visibility.ts
```diff
@@ -31,6 +32,15 @@ export class HeaderVisibilityPlugin extends CorePlugin {
           ? CommandResult.Success
           : CommandResult.TooManyHiddenElements;
       }
+      case "REMOVE_COLUMNS_ROWS":
+        if (!this.hiddenHeaders[cmd.sheetId]) {
```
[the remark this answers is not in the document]
hokolomopo: It's written that way (copy pasted from above) because if the sheet is invalid, we will try to access `this.hiddenHeaders[invalidSheetId][something]`, which will crash. I changed it to `if (!this.getters.tryGetSheet(cmd.sheetId))`, but now it isn't that clear _why_ this check is needed. 

I don't mind either way TBH.

src/plugins/core/header_visibility.ts
```diff
@@ -131,6 +145,16 @@ export class HeaderVisibilityPlugin extends CorePlugin {
     return consecutiveIndexes;
   }
 
+  private getAllVisibleHeaders(sheetId: UID, dimension: Dimension): HeaderIndex[] {
+    const visibleHeaders: HeaderIndex[] = [];
+    for (const i of range(0, this.hiddenHeaders[sheetId][dimension].length)) {
+      if (!this.hiddenHeaders[sheetId][dimension][i]) {
+        visibleHeaders.push(i);
+      }
+    }
```
LucasLefevre, line 154:
```ts
    return range(0, this.hiddenHeaders[sheetId][dimension].length).filter(
      (i) => !this.hiddenHeaders[sheetId][dimension][i]
    );
```

LucasLefevre (commented):
only a small question and suggestion :)

src/plugins/core/header_visibility.ts
```diff
@@ -31,6 +32,14 @@ export class HeaderVisibilityPlugin extends CorePlugin {
           ? CommandResult.Success
           : CommandResult.TooManyHiddenElements;
       }
+      case "REMOVE_COLUMNS_ROWS":
+        if (!this.getters.tryGetSheet(cmd.sheetId)) {
+          return CommandResult.InvalidSheetId;
+        }
```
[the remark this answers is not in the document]
hokolomopo: Yeah but in the model we try to `allowDispatch` on _all_ the plugins, and we don't stop if some plugin `allowDispatch` returns an error. And here we will crash if we try to call `getHiddenColsGroups()` with an invalid sheet id.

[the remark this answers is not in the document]
LucasLefevre: Ah yes, true

tests/components/grid_manipulation.test.ts
```diff
@@ -95,6 +102,24 @@ describe("Context Menu add/remove row/col", () => {
     });
   });
 
+  test("cannot delete nor hide all cols with contextmenu", async () => {
+    setSelection(model, [zoneToXc(model.getters.getSheetZone(model.getters.getActiveSheetId()))]);
+    simulateContextMenu(".o-col-resizer", COLUMN_D);
+    await nextTick();
+    expect(fixture.querySelector(".o-menu div[data-name='delete_column']")).toBeFalsy();
```
rrahir, line 109: it's more correct to check `toBeNull'

hokolomopo: Changed it :) We do it at like 876 places tho, but probably not worth changing it

LucasLefevre:
robodoo rebase-ff

PR #2148, opened 2023-03-02
[FIX] filters: filter can hide all the visible rows of a sheet

hokolomopo:
## Description

It's currently possible to use data filters to hide all the rows in a sheet,
leaving a buggy sheet with no visible rows.

Fixing this is a bit tricky, since the filtered rows are an UI concept, and
thus we cannot rely on the allowDispatch of "HIDE_COLUMNS_ROWS" to prevent
the user from hiding all the rows. Using allowDispatch isn't possible
because the filtered values (and the values of the cells) are different
from an user to another. This means that the allowDispatch could return
a success for an user, and a failure for the other, leading to a
dis-synchronized model state for the 2 user.

Fortunately we can fix this by disabling data filters whose header row is
hidden (by the user or by another filter). By construction of the filters,
it becomes impossible to hide all the rows of a sheet.

This may still becomes an issue in the future if we implement something
like filters being applied to a column header. In that case, we would need
to handle sheets with all their rows hidden.

Odoo task ID : [3205608](https://www.odoo.com/web#id=3205608&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui/filter_evaluation.ts
+++ b/src/plugins/ui/filter_evaluation.ts
@@ -72,6 +72,9 @@ export class FilterEvaluationPlugin extends UIPlugin {
       case "CREATE_SHEET":
         this.filterValues[cmd.sheetId] = {};
         break;
+      case "HIDE_COLUMNS_ROWS":
+        this.updateHiddenRows();
+        break;
       case "UPDATE_FILTER":
         this.updateFilter(cmd);
         this.updateHiddenRows();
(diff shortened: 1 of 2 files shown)


rrahir (approved):
it's even better when you realize that Gsheet does not handle this at all

rrahir:
@hokolomopo actually, it could be nice to be more precise in the commit on why we cannot rely on the allowdispatch of a UI concept (because the values might be different from a user to another)

hokolomopo:
@rrahir added some more lines in the commit messages. And yeah, GSheet doesn't bother with all that and just crashes 🤷

rrahir:
@robodoo r+

PR #2157, opened 2023-03-03
[FIX] figure,chart: Enforce unicity of figure ids

rrahir:
## Description:

The issue of duplicated chart ids was first addressed in PR https://github.com/odoo/o-spreadsheet/pull/2102 by
trying to defined chart ids per sheet (just like figures)
Unfortunately, the fix was not appropriate for several reasons:
1. Some commands in Odoo were not dispatching the sheetId along with the
   chartId, making the mapping sheetId, chartId hazardous
2. There was absolutely 0 verification that the commands targeting a
   chartId were also providing a sheet Id that matched. So the said
   commands cannot be trusted either

This commit is exploring the other solution that is forcing the unicity
of a figure id. The data are adapted so that figures with a duplicated
id well have the latter updated to ensure unicity.

This commit also tries to solve the wrong `sheetId` parameter in
`UPDATE_CHART` by simply ignoring it in the commands. It's not necesarry
since we now have the unicity of figure ids.
 [15.0-fix-charts-rar](https://github.com/odoo/o-spreadsheet/compare/15.0-fix-charts-rar) ([#2157](https://github.com/odoo/o-spreadsheet/pull/2157))

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/chart_panel.ts
+++ b/src/components/side_panel/chart_panel.ts
@@ -7,6 +7,7 @@ import {
   DispatchResult,
   Figure,
   SpreadsheetEnv,
+  UID,
 } from "../../types/index";
 import { ColorPicker } from "../color_picker";
 import * as icons from "../icons";
@@ -166,6 +167,8 @@ export class ChartPanel extends Component<Props, SpreadsheetEnv> {
   static components = { SelectionInput, ColorPicker };
   private getters = this.env.getters;
 
+  private chartSheetId: UID = this.findSheetId(this.props.figure.id);
+
   private state: ChartPanelState = useState(this.initialState(this.props.figure));
 
   async willUpdateProps(nextProps: Props) {
@@ -174,14 +177,15 @@ export class ChartPanel extends Component<Props, SpreadsheetEnv> {
       return;
     }
     if (nextProps.figure.id !== this.props.figure.id) {
+      this.chartSheetId = this.findSheetId(nextProps.figure.id);
       this.state.panel = "configuration";
       this.state.fillColorTool = false;
       this.state.datasetDispatchResult = undefined;
       this.state.labelsDispatchResult = undefined;
       this.state.chart = this.env.getters.getChartDefinitionUI(
-        this.env.getters.getActiveSheetId(),
+        this.chartSheetId,
         nextProps.figure.id
-      );
+      )!;
     }
   }
 
@@ -239,7 +243,7 @@ export class ChartPanel extends Component<Props, SpreadsheetEnv> {
   private updateChart(definition: ChartUIDefinitionUpdate): DispatchResult {
     return this.env.dispatch("UPDATE_CHART", {
       id: this.props.figure.id,
-      sheetId: this.getters.getActiveSheetId(),
+      sheetId: this.chartSheetId,
       definition,
     });
   }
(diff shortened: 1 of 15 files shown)


src/plugins/core/chart.ts
```diff
@@ -245,13 +257,16 @@ export class ChartPlugin extends CorePlugin<ChartState> implements ChartState {
...
     const data: ChartDefinition = this.chartFigures[figureId];
+    if (!data) {
+      return undefined;
+    }
     const dataSets: ExcelChartDataset[] = data.dataSets
       .map((ds: DataSet) => this.toExcelDataset(ds))
       .filter((ds) => ds.range !== ""); // && range !== INCORRECT_RANGE_STRING ? show incorrect #ref ?
     return {
-      ...this.getChartDefinitionUI("forceSheetReference", figureId),
+      ...this.getChartDefinitionUI("forceSheetReference", figureId)!,
```
pro-odoo, line 269: If it's possible to remove the `!`, it's better

src/components/side_panel/chart_panel.ts
```diff
@@ -239,7 +242,7 @@ export class ChartPanel extends Component<Props, SpreadsheetEnv> {
   private updateChart(definition: ChartUIDefinitionUpdate): DispatchResult {
     return this.env.dispatch("UPDATE_CHART", {
       id: this.props.figure.id,
-      sheetId: this.getters.getActiveSheetId(),
+      sheetId: this.state.chartSheetId,
```
LucasLefevre, line 245: sheetId in this command is still unused, right ?

src/components/side_panel/chart_panel.ts
```diff
@@ -178,10 +180,11 @@ export class ChartPanel extends Component<Props, SpreadsheetEnv> {
       this.state.fillColorTool = false;
       this.state.datasetDispatchResult = undefined;
       this.state.labelsDispatchResult = undefined;
+      this.state.chartSheetId = this.findSheetId(nextProps.figure.id);
       this.state.chart = this.env.getters.getChartDefinitionUI(
-        this.env.getters.getActiveSheetId(),
+        this.state.chartSheetId,
```
LucasLefevre, line 185: why is it in the state ?

src/migrations/data.ts
```diff
@@ -343,6 +385,55 @@ function dropCommands(initialMessages, commandType: string) {
   return messages;
 }
 
+function fixUpdateChart(data: Partial<WorkbookData>, initialMessages: StateUpdateMessage[]) {
```
LucasLefevre, line 388: I don't understand why it's needed since `cmd.sheetId` is unused anyway
(and should be eventually removed)

src/components/side_panel/chart_panel.ts
rrahir: no good reason , no need for reactivity as it changes with the figure - you're right

src/components/side_panel/chart_panel.ts
rrahir: not here ,but it's actually used starting v16.0

src/migrations/data.ts
rrahir: again, used in 16.0+ atm

pro-odoo (commented):
robodoo r+

PR #2162, opened 2023-03-03
[FIX] range: show ref error after removing col/row in formulas

yunchen4:
## Description:

Previouly after we remove cols/rows that are used in formulas, the formula will not show error reminding invalid references, but keep the previous range and give the wrong results.

This PR fixes this problem. After deleting rows/cols the formulas refer to removed ones will show errors. It's done by change `newRange` when removing into an invalid range. 

The tests are co-authored with Adrien Minne (@hokolomopo).

Odoo task ID : [2719611](https://www.odoo.com/web#id=2719611&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/zones.ts
+++ b/src/helpers/zones.ts
@@ -65,6 +65,9 @@ export function isZoneValid(zone: Zone): boolean {
   if (isNaN(zone.bottom) || isNaN(zone.top) || isNaN(zone.left) || isNaN(zone.right)) {
     return false;
   }
+  if (zone.top < 0 || zone.left < 0 || zone.right < 0 || zone.bottom < 0) {
+    return false;
+  }
   return zone.bottom >= zone.top && zone.right >= zone.left;
 }
 
--- a/tests/formulas/formulas.test.ts
+++ b/tests/formulas/formulas.test.ts
@@ -1,6 +1,12 @@
 import { Model, normalize } from "../../src";
 import { INCORRECT_RANGE_STRING } from "../../src/constants";
-import { createSheetWithName } from "../test_helpers/commands_helpers";
+import {
+  createSheetWithName,
+  deleteColumns,
+  deleteRows,
+  setCellContent,
+} from "../test_helpers/commands_helpers";
+import { getCellContent, getCellText } from "../test_helpers/getters_helpers";
 
 function moveFormula(model: Model, formula: string, offsetX: number, offsetY: number): string {
   const sheetId = model.getters.getActiveSheetId();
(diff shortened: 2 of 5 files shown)


tests/formulas/formulas.test.ts
```diff
@@ -100,3 +106,29 @@ describe("createAdaptedRanges", () => {
...
+  let model: Model;
+  beforeEach(() => {
+    model = new Model({
+      sheets: [
+        {
+          colNumber: 10,
+          rowNumber: 10,
+        },
+      ],
+    });
```
hokolomopo, line 120:
```suggestion
    model = new Model();
```
we do'nt really care about the number of cols at initialization in those tests

tests/plugins/range.test.ts
```diff
@@ -124,6 +124,53 @@ describe("range plugin", () => {
       });
     });
 
+    describe("create a range and remove multiple columns", () => {
+      beforeEach(() => {
+        m = new Model({
```
hokolomopo, line 129: model is already defined in the parent describe.
I know you copy/pasted, but the syntax in wrong, you can remove the `rows: 10, cols: 10`. They are 1) useless and 2) wrong, that's not how we define the number of cols/rows (see your other tests in `formula.test.ts`)

tests/plugins/range.test.ts
```diff
@@ -124,6 +124,53 @@ describe("range plugin", () => {
...
+
+      test("including one column before the start and the first column", () => {
+        deleteColumns(m, ["C", "B"]);
+        expect(m.getters.getUsedRanges()).toEqual(["B2:D5"]);
+      });
+
+      test("including one column after the end and the last column", () => {
+        deleteColumns(m, ["G", "F"]);
+        expect(m.getters.getUsedRanges()).toEqual(["C2:E5"]);
+      });
```
hokolomopo, line 171: Might be worth to add a test if every col/row were deleted

tests/formulas/formulas.test.ts
```diff
@@ -100,3 +106,29 @@ describe("createAdaptedRanges", () => {
...
+  });
+
+  test("delete multiple columns, including ones in formula", () => {
+    setCellContent(model, "A1", "=C1");
+    deleteColumns(model, ["B", "C"]);
```
hokolomopo, line 125: We should have both tests `delete B,C` and `delete C,D`, because one was working the other not

src/helpers/zones.ts
```diff
@@ -65,6 +65,9 @@ export function isZoneValid(zone: Zone): boolean {
   if (isNaN(zone.bottom) || isNaN(zone.top) || isNaN(zone.left) || isNaN(zone.right)) {
     return false;
   }
+  if (zone.top < 0 || zone.left < 0 || zone.right < 0 || zone.bottom < 0) {
+    return false;
+  }
```
hokolomopo, line 70: You're 100% right, but out of curiosity what made you change that in this particular PR ? I do'nt think it breaks any test ATM. (Come to think of it, it really _should_ break a test, might be worth adding one)

hokolomopo (commented):
Some comments on the tests, otherwise it seems to be working well now :)

src/helpers/zones.ts
yunchen4: @hokolomopo Changing it can have the correct error text (from something like `undefined for rows` to `invalid reference`) in error tooltip. 
I think the reason why it didn't break the tests is because the error is still correctly caught by us, so it will still show in the error tooltip, and the app won't crash.

tests/formulas/formulas.test.ts
yunchen4: @hokolomopo Deleting C and D will also not work (show #REF error in Google Sheets). Should I test it anyway?

hokolomopo: Deleting `B,C` and deleting `C,D` should have the same result (=#REF)

The problam was that before your PR delting `C,D` worked, the formula became `=#REF`, but deleting `B,C` didn't work, the formula didn't change. That's why I'm saying that we should test both.

tests/plugins/range.test.ts
```diff
@@ -124,6 +124,62 @@ describe("range plugin", () => {
...
+
+      test("all columns in the sheet", () => {
+        const numberOfCols = m.getters.getNumberCols(m.getters.getActiveSheetId());
+        deleteColumns(
```
rrahir, line 175: you cannot delete all the columns of the sheet, this command is rejected so it will never actually be processed by `RangeAdapter` nor `PluginTestRange`, I might be missing the point of that test ?

tests/formulas/formulas.test.ts
```diff
@@ -100,3 +106,34 @@ describe("createAdaptedRanges", () => {
...
+describe("Remove columns/rows that are references of formula", () => {
+  let model: Model;
+  beforeEach(() => {
+    model = new Model();
+  });
+
+  test("delete multiple columns, including the one in formula and the one before it", () => {
+    setCellContent(model, "A1", "=C1");
+    deleteColumns(model, ["B", "C"]);
+    expect(getCellContent(model, "A1")).toEqual("#ERROR");
```
rrahir, line 119: you could go further on an check which dependencies are now invalid references

tests/plugins/range.test.ts
```diff
@@ -124,6 +124,62 @@ describe("range plugin", () => {
       });
     });
 
+    describe("create a range and remove multiple columns", () => {
+      beforeEach(() => {
```
rrahir, line 128: I seems that all those test were already passing before your fix, it would be nice to have a test or two that tests the fix. For this you'll need to adapt `PluginTestRange`

rrahir (changes requested):
The tests don't seem to make much sense

tests/plugins/range.test.ts
yunchen4: @rrahir I knew this behavior so it is here to make the test range more complete. Should I just add a comment here or delete this (and similar) tests?

rrahir (approved):

rrahir:
@robodoo r+

PR #2182, opened 2023-03-06
[IMP] spreadsheet: separate ZoneDimension and PixelDimension

hokolomopo:
## Description

We currently have getters/helpers that return a size/dimension in `{width; height}`, but sometime the width is in pixel, and sometimes it's in columns. It is sometime unclear which is which without looking at the getter.

The goal here is to clarify this situation by defining 2 interfaces to avoid confusion:

- ZoneDimension : `{ nOfCols, nOfRows }`
- PixelDimension: `{width, height}`

Odoo task ID : [3217977](https://www.odoo.com/web#id=3217977&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/types/misc.ts
+++ b/src/types/misc.ts
@@ -85,8 +85,8 @@ export interface UnboundedZone {
 }
 
 export interface ZoneDimension {
-  height: HeaderIndex;
-  width: HeaderIndex;
+  numberOfRows: HeaderIndex;
+  numberOfCols: HeaderIndex;
 }
 
 export type Align = "left" | "right" | "center" | undefined;
--- a/src/types/rendering.ts
+++ b/src/types/rendering.ts
@@ -1,12 +1,5 @@
 import { Alias, Align, Border, Pixel, Style, VerticalAlign, Zone } from "./misc";
 
-export type Rect = {
-  x: Pixel;
-  y: Pixel;
-  width: Pixel;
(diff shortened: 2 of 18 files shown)


src/types/rendering.ts
```diff
@@ -15,6 +8,13 @@ export interface DOMCoordinates {
   y: Pixel;
 }
 
+export interface PixelDimension {
```
rrahir, line 11: Just noticed we already have `DOMDimension`, guess we can stick with it

hokolomopo: Indeed, it is done :)

src/types/misc.ts
```diff
@@ -85,8 +85,8 @@ export interface UnboundedZone {
 }
 
 export interface ZoneDimension {
-  height: HeaderIndex;
-  width: HeaderIndex;
+  nOfRows: HeaderIndex;
```
rrahir, line 88: not really fond of that name, guess we could see with the others what they think.. @odoo/rd-spreadsheet

hokolomopo: changed to numberOfCols/numberOfRows

rrahir:
@robodoo r+

PR #2188, opened 2023-03-08
[FIX] correctly auto-size o-spreadsheet

LucasLefevre:
## Description:

Sizing `o-spreadsheet` has always been difficult because it auto-size itself.

We've had cases where it would start with a sensible size but then shrink to 0 height (ever seen that when debugging QUnit tests ?).

We rely on `o-grid-overlay`'s natural size in the DOM to actually sync the size in the model.
However, its natural size was influenced by the initial state of the model.
Because the default size in the model was (1000, 1000) the canvas is sized to (1000,1000) and since the canvas is a child
of  `o-grid-overlay`, its natural size is ~ (1000, 1000).
You had to apply very strict css rules to `o-spreadsheet` or `o-grid-overlay` to force another size.
In particular, it didn't work well if you want css such that "`o-spreadsheet` fills the remaining place".
If the available space was smaller than 1000, it couldn't properly "fill the remaining space" because it was already larger!
A standard monitor height is 1080, taking into account, a header, the topbar and the bottom bar, you probably never have more than 1000 pixels available.

With this change, it's much easier to size `o-spreadsheet` the way you want because it's not influenced by it's initial default size. 

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/demo/main.js
+++ b/demo/main.js
@@ -66,7 +66,7 @@ let start;
 
 class Demo extends Component {
   setup() {
-    this.state = useState({ key: 0 });
+    this.state = useState({ key: 0, displayHeader: false });
     this.stateUpdateMessages = [];
     this.client = {
       id: uuidGenerator.uuidv4(),
@@ -92,6 +92,11 @@ class Demo extends Component {
       sequence: 13,
       isReadonlyAllowed: true,
       action: () => this.model.updateMode("normal"),
+    });
+    topbarMenuRegistry.addChild("read_write", ["file"], {
+      name: () => (this.state.displayHeader ? "Hide header" : "Show header"),
+      isReadonlyAllowed: true,
+      action: () => (this.state.displayHeader = !this.state.displayHeader),
     });
 
     topbarMenuRegistry.add("notify", {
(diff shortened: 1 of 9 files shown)


src/constants.ts
```diff
@@ -207,8 +207,15 @@ export enum ComponentsImportance {
...
   ChartAnchor = 1000,
 }
+let DEFAULT_SHEETVIEW_SIZE = 0;
 
-export const DEFAULT_SHEETVIEW_SIZE = 1000;
+export function getDefaultSheetViewSize() {
+  return DEFAULT_SHEETVIEW_SIZE;
+}
+
+export function setDefaultSheetViewSize(size: number) {
```
pro-odoo, line 216: Should be exported in `index.ts` in order to use it in Odoo

demo/main.js
```diff
@@ -66,7 +66,7 @@ let start;
 
 class Demo extends Component {
   setup() {
-    this.state = useState({ key: 0 });
+    this.state = useState({ key: 0, displayHeader: false });
```
LucasLefevre, line 69: funnily enough, the issue this commit is fixing is illustrated when the header appears.
o-spreadsheet is bigger and it auto-size in a not-so-good manner

rrahir:
@robodoo r+

PR #2222, opened 2023-03-14
[FIX] xlsx export: empty string in filtered range

hokolomopo:
## Description

When we try to export a sheet that have a data filter, and that in the filtered zone there is a formula evaluating to an empty string, Excel will say that it found an issue when opening the xlsx.

That's because in the xlsx we have to put all the values that are not filtered, except for empty values. This was correctly handled for simple empty cells, but not for cells evaluating to an empty string.

Odoo task ID : [3231699](https://www.odoo.com/web#id=3231699&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/demo/main.js
+++ b/demo/main.js
@@ -170,9 +170,9 @@ class Demo extends Component {
       this.transportService = undefined;
       this.stateUpdateMessages = [];
     }
-    this.createModel(data || demoData);
+    // this.createModel(data || demoData);
     // this.createModel(makeLargeDataset(26, 10_000, ["numbers"]));
-    // this.createModel({});
+    this.createModel({});
   }
 
   createModel(data) {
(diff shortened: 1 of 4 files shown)


demo/main.js
```diff
@@ -170,9 +170,9 @@ class Demo extends Component {
       this.transportService = undefined;
       this.stateUpdateMessages = [];
     }
-    this.createModel(data || demoData);
+    // this.createModel(data || demoData);
     // this.createModel(makeLargeDataset(26, 10_000, ["numbers"]));
-    // this.createModel({});
+    this.createModel({});
```
pro-odoo, line 175: Revert

hokolomopo: oops

rrahir:
@robodoo r+

PR #2244, opened 2023-03-20
[FIX] side panel: onSidePanelClose prop never call on X symbol click

SimonGenin:
The prop onSidePanelClose was correctly passed to the component inside the side panel, but this prop needed also to be executed by the side panel component itself when it is in charge of closing itself, by example when clicking the X button.

## Description:

Odoo task ID : [3239748] (https://www.odoo.com/web#cids=1&menu_id=4720&action=333&active_id=2328&model=project.task&view_type=form&id=3239748)


src/components/spreadsheet/spreadsheet.ts
```diff
@@ -260,6 +260,7 @@ export class Spreadsheet extends Component<SpreadsheetProps, SpreadsheetChildEnv
   closeSidePanel() {
     this.sidePanel.isOpen = false;
     this.focusGrid();
+    this.sidePanel.panelProps?.onCloseSidePanel();
```
SimonGenin, line 263: I've a doubt, should it be `this.sidePanel.panelProps?.onCloseSidePanel?();` instead ?

LucasLefevre (approved):
robodoo r+

PR #2255, opened 2023-03-21
[FIX] tokenizer: detect xc with sheet name starting with a number

rrahir:
The tokenizer would interpret the sheet name starting by a number as a number token, preventing it to be detected as a proper symbol.

Task: 3215464

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


tests/formulas/tokenizer.test.ts
```diff
@@ -157,6 +157,14 @@ describe("tokenizer", () => {
       { type: "OPERATOR", value: "=" },
       { type: "SYMBOL", value: "'a '' b'!A1" },
     ]);
+    expect(tokenize("=1name!A1")).toEqual([
```
LucasLefevre, line 160: you could add a test where the sheet name is only numbers `123!A1`

rrahir: done, and it works !!  That being said, I really don't like the way I fix this, but I think it's better to ensure it's not detected as a number rather than let is pass and try to recompose it with `mergeSymbolsIntoRanges`  wdyt ?

LucasLefevre: >  I really don't like the way I fix this

Neither do I, our tokenizer is too smart by trying to do too many things at once

> it's better to ensure it's not detected as a number rather than let is pass and try to recompose it with mergeSymbolsIntoRanges wdyt

definitely, especially in 15.0

LucasLefevre (approved):
robodoo r+

PR #2269, opened 2023-03-24
[FIX] misc: prevent row resize caused by multiline formulas

dhrp-odoo:
## Description:

Previously, typing a formula with multiple lines (e.g. = A1 + [NEWLINE] 2)
caused the rows to resize, as the formula was displayed on two lines.
To prevent this unnecessary row resize, a hard-coded value of `1` is now
returned from `getDefaultCellHeight` in the case of formula syntax.

Odoo task ID : [3221078](https://www.odoo.com/web#id=3221078&cids=2&menu_id=4720&action=333&active_id=2328&model=project.task&view_type=form)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/misc.ts
+++ b/src/helpers/misc.ts
@@ -129,8 +129,10 @@ export function getDefaultCellHeight(cell: Cell | undefined): Pixel {
     return DEFAULT_CELL_HEIGHT;
   }
   const fontSize = computeTextFontSizeInPixels(cell.style);
-  const multiLineText = cell.content.split(NEWLINE);
-  return computeTextLinesHeight(fontSize, multiLineText.length) + 2 * PADDING_AUTORESIZE_VERTICAL;
+
+  // the number of lines should be computed from the formula result, but it's not evaluated at this point
+  const numberOfLines = cell.isFormula ? 1 : cell.content.split(NEWLINE).length;
+  return computeTextLinesHeight(fontSize, numberOfLines) + 2 * PADDING_AUTORESIZE_VERTICAL;
 }
 
 export function computeTextWidth(context: CanvasRenderingContext2D, text: string, style: Style) {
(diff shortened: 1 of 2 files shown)


tests/plugins/formatting.test.ts
```diff
@@ -561,4 +561,12 @@ describe("Autoresize", () => {
     model.dispatch("AUTORESIZE_COLUMNS", { sheetId, cols: [0] });
     expect(model.getters.getColSize(sheetId, 0)).toBe(initialSize);
   });
+
+  test("Test to verify row height calculation for cells with formula syntax", async () => {
+    const initialSize = model.getters.getRowSize(sheetId, 0);
+    setCellContent(model, "A2", "10");
+    setCellContent(model, "A1", "=A2\n\n+5\n\n-3\n\n-9");
```
LucasLefevre, line 568: you don't any reference to show this behavior. Let's keep it simple and not add the reference
```suggestion
    setCellContent(model, "A1", "=5\n\n-3\n\n-9");
```

src/helpers/misc.ts
```diff
@@ -129,7 +129,7 @@ export function getDefaultCellHeight(cell: Cell | undefined): Pixel {
     return DEFAULT_CELL_HEIGHT;
   }
   const fontSize = computeTextFontSizeInPixels(cell.style);
-  const multiLineText = cell.content.split(NEWLINE);
+  const multiLineText = cell.isFormula ? [cell.content] : cell.content.split(NEWLINE);
```
LucasLefevre, line 132: then the array contains a single element which is a multiline text. That's weird because it's supposed to contain only single lines. It works because it's only used for `.length`, but still, it's weird.
I think I would just hard-code `1`.
Maybe with a comment
```suggestion
  // the number of lines should be computed from the formula result, but it's not evaluated at this point
  const numberOfLines = cell.isFormula ? 1 : cell.content.split(NEWLINE).length;
```

LucasLefevre (commented):
two small comments :)

dhrp-odoo:
> two small comments :)

I have made the changes you have suggested and also updated the Commit and PR message accordingly.

LucasLefevre (approved):
Nice, Thanks :)
robodoo r+

PR #2276, opened 2023-03-27
[IMP] color_picker : added accessibility to custom color picker

khpa-odoo:
Enabled custom color picker to open by clicking on Custom part no matter where we click (not only on the +), beside the already added colors.

Task-3223388

## Description:

Odoo task ID : [3223388](https://www.odoo.com/web#id=3223388&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/color_picker/color_picker.xml
+++ b/src/components/color_picker/color_picker.xml
@@ -21,13 +21,13 @@
           </div>
         </div>
         <div class="o-separator"/>
-        <div class="o-color-picker-section-name">
+        <div
+          class="o-color-picker-section-name o-color-picker-toggler"
+          t-on-click="toggleColorPicker">
           <span>Custom</span>
         </div>
-        <div class="colors-grid">
-          <div
-            class="o-color-picker-toggler o-color-picker-line-item"
-            t-on-click.stop="toggleColorPicker">
+        <div class="colors-grid o-color-picker-toggler" t-on-click.stop="toggleColorPicker">
+          <div class="o-color-picker-line-item o-color-picker-toggler">
             <div align="center">+</div>
           </div>
           <div
--- a/tests/components/color_picker.test.ts
+++ b/tests/components/color_picker.test.ts
@@ -51,12 +51,15 @@ describe("Color Picker buttons", () => {
     expect(onColorPicked).toHaveBeenCalledWith("");
   });
 
-  test("Click on '+' button toggle the custom color part", async () => {
+  test("Clicking on '+', custom section except custom added colors must toggle the custom color picker", async () => {
     await mountColorPicker();
-    await simulateClick(".o-color-picker-toggler");
-    expect(fixture.querySelector(".o-custom-selector")).toBeDefined();
-    await simulateClick(".o-color-picker-toggler");
-    expect(fixture.querySelector(".o-custom-selector")).toBeNull();
+    const elements = fixture.querySelectorAll(".o-color-picker-toggler");
+    for (const element of elements) {
+      await simulateClick(element);
+      expect(fixture.querySelector(".o-custom-selector")).toBeDefined();
+      await simulateClick(element);
+      expect(fixture.querySelector(".o-custom-selector")).toBeNull();
+    }
   });
 
   test("Can pick a standard color", async () => {


hokolomopo (commented):
Not bad, just some details and a random test that probably shouldn't be there

src/components/color_picker/color_picker.xml
```diff
@@ -21,13 +21,11 @@
...
         <div class="o-separator"/>
-        <div class="o-color-picker-section-name">
+        <div class="o-color-picker-section-name" t-on-click.stop="toggleColorPicker">
           <span>Custom</span>
         </div>
-        <div class="colors-grid">
-          <div
-            class="o-color-picker-toggler o-color-picker-line-item"
-            t-on-click.stop="toggleColorPicker">
+        <div class="colors-grid o-color-picker-toggler" t-on-click.stop="toggleColorPicker">
```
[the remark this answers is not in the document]
khpa-odoo: I've added the class in section name as well. Also, I've made necessary changes required in the test case to test the click on custom section name as well as color grid.

tests/components/color_picker.test.ts
```diff
@@ -51,12 +51,15 @@ describe("Color Picker buttons", () => {
...
 
-  test("Click on '+' button toggle the custom color part", async () => {
+  test("Clicking on '+', custom section except custom added colors must toggle the custom color picker", async () => {
     await mountColorPicker();
-    await simulateClick(".o-color-picker-toggler");
-    expect(fixture.querySelector(".o-custom-selector")).toBeDefined();
-    await simulateClick(".o-color-picker-toggler");
-    expect(fixture.querySelector(".o-custom-selector")).toBeNull();
+    const elements = fixture.querySelectorAll(".o-color-picker-toggler");
+    elements.forEach(async (element) => {
```
hokolomopo, line 57:
```suggestion
    for(const element of elements) {...}{
```
We try to avoid `foreach` in this repo, because some people find them ugly.

Side note, but when you're done with your changes, you should change the state of the task in Odoo, so we know we can do something for you. You should usually put the tak back in Rd Dev and green it (or in RD Test green if it was already tested, and you didn't change much)

khpa-odoo: Ok. Applied suggested changes and changed it's state.

src/components/color_picker/color_picker.xml
```diff
@@ -21,13 +21,13 @@
...
+        <div
+          class="o-color-picker-section-name o-color-picker-toggler"
+          t-on-click="toggleColorPicker">
           <span>Custom</span>
         </div>
-        <div class="colors-grid">
-          <div
-            class="o-color-picker-toggler o-color-picker-line-item"
-            t-on-click.stop="toggleColorPicker">
+        <div class="colors-grid o-color-picker-toggler" t-on-click.stop="toggleColorPicker">
```
LucasLefevre, line 29: Actually, I'd remove the classes `.o-color-picker-toggler` completely.
In the tests, I would target directly elements with their existing classes `.o-color-picker-section-name` and `.colors-grid`. That would be more explicit that we want *those* elements to toggle the color picker.

What's no longer explicitly tested is that clicking on the "+" opens the color picker.
What we can do to test it more easily: add a custom class on the "+" (like `.o-color-picker-plus`) and target it

tests/components/color_picker.test.ts
```diff
@@ -51,12 +51,15 @@ describe("Color Picker buttons", () => {
...
-  test("Click on '+' button toggle the custom color part", async () => {
+  test("Clicking on '+', custom section except custom added colors must toggle the custom color picker", async () => {
     await mountColorPicker();
-    await simulateClick(".o-color-picker-toggler");
-    expect(fixture.querySelector(".o-custom-selector")).toBeDefined();
-    await simulateClick(".o-color-picker-toggler");
-    expect(fixture.querySelector(".o-custom-selector")).toBeNull();
+    const elements = fixture.querySelectorAll(".o-color-picker-toggler");
+    for (const element of elements) {
+      await simulateClick(element);
```
LucasLefevre, line 58: See my previous comment :) 
```suggestion
    const selectors = [".o-color-picker-section-name", ".colors-grid", ".o-color-picker-plus"]
    for (const selector of selectors) {
      const element = fixture.querySelector(selector)
      await simulateClick(element);
```

khpa-odoo: Yes I thought so. But the classes .o-color-picker-section-name and .colors-grid are applied to Standard as well as Custom section. Then to explicitly access custom section name and grid, how about we add classes like o-custom-section-name & o-custom-colors-grid?

LucasLefevre: Oh yes, you're right...
let's keep it like that then and add `.o-color-picker-toggler` on the "+" as well so it's tested. That's good enough :)

src/components/color_picker/color_picker.xml
```diff
@@ -21,13 +21,13 @@
...
+          <div class="o-color-picker-line-item">
```
LucasLefevre, line 30:
```suggestion
          <div class="o-color-picker-toggler o-color-picker-line-item">
```

rrahir:
the style is unfortunately broken since the change you made, there should be some css lines to update :) 
image

khpa-odoo:
Updated the PR 👍

src/components/color_picker/color_picker.xml
rrahir: the comment is still relevant :)

src/components/color_picker/color_picker.ts
```diff
@@ -56,6 +56,9 @@ css/* scss */ `
       border-radius: 50px;
       border: ${ITEM_BORDER_WIDTH}px solid #666666;
       padding: 0px;
+      display: flex;
```
rrahir, line 59: I still see the broken style on the button even with this change, but I realized that the issue is already present in version 16.0 so we should address it in 16.0 directly. I think we can get rid of these lines, then?

khpa-odoo: Okay sure.

src/components/color_picker/color_picker.xml
rrahir: ok does make sense, I didn't see the comments below Sorry about that

rrahir:
@robodoo r+

PR #2291, opened 2023-03-30
[FIX] XLSX: open exported .xlsx file containing images in Excel

yunchen4:
## Description:

This PR fixes the problem of openning exported .xlsx file containing images in Excel.

The fix is exporting the mime type of images together with other XML files. The mime type is stored when uploading new images.

Note that for images that are already in the database, the mime type in o-spreadsheet is undefined. When exporting files including these images, the mime type will be `images/jpeg` by default. 

Odoo task ID : [3239093](https://www.odoo.com/web#id=3239093&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/figures/images/image_provider.ts
+++ b/src/helpers/figures/images/image_provider.ts
@@ -13,7 +13,7 @@ export class ImageProvider implements ImageProviderInterface {
     const file = await this.getImageFromUser();
     const path = await this.fileStore.upload(file);
     const size = await this.getImageSize(path);
-    return { path, size };
+    return { path, size, mimetype: file.type };
   }
 
   private getImageFromUser(): Promise<File> {
--- a/src/types/xlsx.ts
+++ b/src/types/xlsx.ts
@@ -157,7 +157,7 @@ export interface XLSXExportXMLFile {
 
 export interface XLSXExportImageFile {
   path: string;
-  imagePath: string;
+  imageSrc: string;
 }
 
 export interface XLSXExport {
--- a/src/xlsx/helpers/xml_helpers.ts
+++ b/src/xlsx/helpers/xml_helpers.ts
@@ -85,6 +85,12 @@ export function createOverride(partName: string, contentType: string): XMLString
   `;
 }
 
+export function createDefaultXMLElement(extension: string, contentType: string): XMLString {
+  return escapeXml/*xml*/ `
+    <Default Extension="${extension}" ContentType="${contentType}" />
+  `;
+}
+
 export function joinXmlNodes(xmlNodes: XMLString[]): XMLString {
   return new XMLString(xmlNodes.join("\n"));
 }
--- a/src/xlsx/xlsx_writer.ts
+++ b/src/xlsx/xlsx_writer.ts
@@ -10,6 +10,7 @@ import {
 } from "../types/xlsx";
 import { XLSXExportXMLFile } from "./../types/xlsx";
 import { CONTENT_TYPES, NAMESPACE, RELATIONSHIP_NSR, XLSX_RELATION_TYPE } from "./constants";
+import { IMAGE_MIMETYPE_EXTENSION_MAPPING } from "./conversion";
 import { createChart } from "./functions/charts";
 import { addConditionalFormatting } from "./functions/conditional_formatting";
 import { createDrawing } from "./functions/drawings";
@@ -37,6 +38,7 @@ import {
   convertWidthToExcel,
 } from "./helpers/content_helpers";
 import {
+  createDefaultXMLElement,
   createOverride,
   createXMLFile,
   escapeXml,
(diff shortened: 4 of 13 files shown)


src/xlsx/xlsx_writer.ts
```diff
@@ -142,19 +147,22 @@ function createWorksheets(data: ExcelWorkbookData, construct: XLSXStructure): XL
 
     const images = sheet.images;
     for (const image of images) {
+      const contentType = image.data.contentType ?? (await getImageContentType(image.data.path));
```
hokolomopo, line 150: I would probably pre-fetch all the `img.contentType` in `model`, so we don't have to worry about async functions and whatnot in the `xlsx_writer`. Just my opinon tho.

src/types/xlsx.ts
```diff
@@ -158,6 +158,7 @@ export interface XLSXExportXMLFile {
 export interface XLSXExportImageFile {
   path: string;
   imagePath: string;
+  contentType?: string;
```
hokolomopo, line 161: Was already like that before this PR, but it's unclear what are `path` and `imagePath`. Woudl deserve better name and/or a bit of docstring.

src/xlsx/helpers/xml_helpers.ts
```diff
@@ -85,6 +85,12 @@ export function createOverride(partName: string, contentType: string): XMLString
   `;
 }
 
+export function createDefault(extension: string, contentType: string): XMLString {
```
hokolomopo, line 88: `createDefaultXMLElement` ? Or another name, but createDefault alone seems a bit strange

hokolomopo (commented):
Soem naming issues, but sounds good otherwise 👍

src/model.ts
```diff
@@ -585,7 +585,7 @@ export class Model extends EventBus<any> implements CommandDispatcher {
    * This prove to be necessary if the client did not trigger that evaluation in the first place
    * (e.g. open a document with several sheet and click on download before visiting each sheet)
    */
-  exportXLSX(): XLSXExport {
+  async exportXLSX(): Promise<XLSXExport> {
```
LucasLefevre, line 588: as discussed: don't support old databases. There's only a handful of them (if there are some)

src/types/xlsx.ts
LucasLefevre: It really does. I don't know the difference

src/xlsx/xlsx_writer.ts
```diff
@@ -142,19 +143,26 @@ function createWorksheets(data: ExcelWorkbookData, construct: XLSXStructure): XL
 
     const images = sheet.images;
     for (const image of images) {
+      const contentType = image.data.contentType;
```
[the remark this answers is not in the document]
yunchen4: > if we don't have a contentType, then we should just drop the addition of that image in excel altogether because there is a non negligeable chance that it will still corrupt the generated file.

For the images that are already uploaded before this PR, when calling `createWorksheets()`, `contentType` (now `mimetype` according to your first comment) in `Image` type doesn't exist. The mime type only exist in the database. 

If we don't export images without `mimetype`, old images will not be exported. Is it better to set the mimetype to be `image/jpeg` by default for old images? Only when the mimetype is not supported we won't export them.

src/xlsx/xlsx_writer.ts
```diff
@@ -142,19 +143,26 @@ function createWorksheets(data: ExcelWorkbookData, construct: XLSXStructure): XL
...
       const xlsxImageId = convertImageId(image.id);
+      let imageFileName = `image${xlsxImageId}`;
+      if (contentType) {
+        imageFileName = `image${xlsxImageId}.${contentType.split("/")[1]}`;
```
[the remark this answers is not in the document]
rrahir: I meant when you build the excelworkbookdata with `exportXLSX` :)

src/helpers/figures/images/image_provider.ts
```diff
@@ -13,7 +13,7 @@ export class ImageProvider implements ImageProviderInterface {
     const file = await this.getImageFromUser();
     const path = await this.fileStore.upload(file);
     const size = await this.getImageSize(path);
-    return { path, size };
+    return { path, size, mimetype: file.type };
```
rrahir, line 16: let's throw when the mimetype is not supported in the provided list

src/xlsx/xlsx_writer.ts
```diff
@@ -140,25 +141,32 @@ function createWorksheets(data: ExcelWorkbookData, construct: XLSXStructure): XL
...
           type: XLSX_RELATION_TYPE.image,
         }
       );
       drawingRelIds.push(imageRelId);
       files.push({
-        path: `xl/media/image${xlsxImageId}`,
-        imagePath: image.data.path,
+        path: `xl/media/${imageFileName}`,
+        imageSrc: image.data.path,
+        mimetype: mimeType,
```
rrahir, line 165: think we can drop this

rrahir:
@robodoo r+

PR #2298, opened 2023-03-31
[FIX] viewport: infinite loop with hidden headers

hokolomopo:
## Description

When we move the selection we scroll the viewport so that new selection is visible. Previously, the strategy to do this was (when moving up):

1) set the viewport offset to the row above the current topmost row 2) adapt the viewport top/bottom to the new offset 3) repeat while the selection is not visible

This doesn't work if the row above the current topmost row is hidden, since it have the same position has the current row. This led to an infinite loop with the structure of the code.

With this commit, the hidden rows are skipped in step 1). The loop scrolling the viewport was also rewritten, so it's impossible to have an infinite loop even if there is a bug somewhere.

Odoo task ID : [3244431](https://www.odoo.com/web#id=3244431&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


rrahir:
ptit rebase et puis on merge ^^

src/helpers/internal_viewport.ts
```diff
@@ -143,53 +143,55 @@ export class InternalViewport {
...
-      for (topRow = this.top; topRow >= 0; topRow--) {
-        if (!this.getters.isRowHidden(sheetId, topRow)) {
-          break;
-        }
+    const { end } = this.getters.getRowDimensions(sheetId, targetRow);
+    const maxRow = this.getters.getNumberRows(sheetId);
+
+    if (this.offsetY + this.height + this.offsetCorrectionY < end) {
+      for (let row = this.top; this.offsetY + this.height + this.offsetCorrectionY < end; row++) {
+        if (row > maxRow) break;
```
rrahir, line 179: maybe write this with some brackets, actually. I know it seems like wasted space/lines but I'd argue that the brackets are
- more readable in general
- less error prone (see https://www.imperialviolet.org/2014/02/22/applebug.html as a nice example)
- less likely to break upon future changes

I feel it's worth having a guideline over the whole code base for this

rrahir (commented):
a small change could be welcome as well - I know we have those single lines at several places in the code but we should go in a common direction rather than spread that practice

src/helpers/internal_viewport.ts
hokolomopo: Personnaly, I think that no bracket makes the code a bit easier to read, and not error prone _as long as you can write everything on a single line_. If you have a line break with prettier, it becomes error prone I agree. 

But we could use a guideline for the whole code base yeah. Idk how @LucasLefevre feels about it.

hokolomopo: I do like my one liners 😢 But if you both perfer brackets, I can take the habit of puttin them.

hokolomopo:
@rrahir rebased

rrahir:
@robodoo r+

rrahir:
thanks :)

rrahir:
@robodoo retry

PR #2319, opened 2023-04-05
[FIX] ContentEditableHelper: fix getCurrentSelection for Firefox

rrahir:
How to reproduce:
In firefox
- from the grid, input `=sum(`
- hit backspace -> the selection is set between `s` and `u` while it should be after `m`.

Starting from a cell content `=SUM(`  we have the composer content
```html
<span>=</span>
<span>sum</span>
<span>(</span>
```
When we delete the parenthesis, here is what google does
```html
<span>=</span>
<span>sum{document.selection}</span>
```
and firefox
```html
<span>=</span>
<span>sum</span>{document.selection}
```

The function `ContentEditableHelper.getCurrentSelection` is biased as it would assume that the selection would always end up in the deepest node (here a node `#text`).

This commit modifies the function to handle all selection nodes.

Task: 3071026

## Description:

Odoo task ID : [3071026](https://www.odoo.com/web#id=3071026&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/composer/content_editable_helper.ts
+++ b/src/components/composer/content_editable_helper.ts
@@ -105,16 +105,34 @@ export class ContentEditableHelper {
   getCurrentSelection() {
     let { startElement, endElement, startSelectionOffset, endSelectionOffset } =
       this.getStartAndEndSelection();
-    let startSizeBefore = this.findSizeBeforeElement(startElement!);
-    let endSizeBefore = this.findSizeBeforeElement(endElement!);
+
+    let startSizeBefore = this.findSelectionIndex(startElement!, startSelectionOffset);
+    let endSizeBefore = this.findSelectionIndex(endElement!, endSelectionOffset);
 
     return {
-      start: startSizeBefore + startSelectionOffset,
-      end: endSizeBefore + endSelectionOffset,
+      start: startSizeBefore,
+      end: endSizeBefore,
     };
   }
 
-  private findSizeBeforeElement(nodeToFind: Node): number {
+  /**
+   * Computes the text 'index' inside this.el based on the currently selected node and its offset.
+   * The selected node is either a Text node or an Element node.
+   *
+   * case 1 -Text node:
+   * the offset is the number of characters from the start of the node. We have to add this offset to the
+   * content length of all previous nodes.
+   *
+   * case 2 - Element node:
+   * the offset is the number of child nodes before the selected node. We have to add the content length of
+   * all the bnodes prior to the selected node as well as the content of the child node before the offset.
+   *
+   * See the MDN documentation for more details.
+   * https://developer.mozilla.org/en-US/docs/Web/API/Range/startOffset
+   * https://developer.mozilla.org/en-US/docs/Web/API/Range/endOffset
+   *
+   */
+  private findSelectionIndex(nodeToFind: Node, nodeOffset: number) {
     let it = this.iterateChildren(this.el);
     let usedCharacters = 0;
     let current = it.next();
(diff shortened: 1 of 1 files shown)


rrahir:
@fw-bot up to saas-16.1

saas-16.2 requires a more in depth fix, but since we cannot really test it, makes sense to have another dedicated PR

src/components/composer/content_editable_helper.ts
```diff
@@ -128,6 +129,21 @@ export class ContentEditableHelper {
       current = it.next();
     }
 
+    if (current.value === nodeToFind) {
+      if (!current.value.hasChildNodes()) {
+        usedCharacters += nodeOffset;
+      } else {
+        const children = [...current.value.childNodes].slice(0, nodeOffset);
```
hokolomopo, line 136: Add a comment or a link of the [documentation](https://developer.mozilla.org/en-US/docs/Web/API/Range/endOffset), because the offset being either a number of chars, or a number of children is not intuitive _at all_

src/components/composer/content_editable_helper.ts
```diff
@@ -128,6 +129,21 @@ export class ContentEditableHelper {
...
+        usedCharacters += children.reduce((acc: number, child: Node) => {
+          if (child.textContent) {
+            return acc + child.textContent.length;
+          } else {
+            return acc;
+          }
+        }, 0);
```
hokolomopo, line 143:
```suggestion
        for(const child of children){
            if (!child.textContent) continue;
            usedCharacters += child.textContent.length;
        };
```
Cleaner than a reduce, no ?

hokolomopo (commented):
Commit message is a bit messed up, with some french

src/components/composer/content_editable_helper.ts
rrahir: Agreed, it sems to makes sense in 16.2 but it seems clearer

rrahir:
@hokolomopo could you have a look at let me know what you think ?

src/components/composer/content_editable_helper.ts
```diff
@@ -105,16 +105,25 @@ export class ContentEditableHelper {
...
-      end: endSizeBefore + endSelectionOffset,
+      start: startSizeBefore,
+      end: endSizeBefore,
     };
   }
 
-  private findSizeBeforeElement(nodeToFind: Node): number {
+  /**
+   * FComputes the text 'index' inside this.el based on the currently selected node and its offset.
+   * the offset might be inside a text node or inside a child node. In the latter case, we have to
```
hokolomopo, line 120:
```suggestion
   * Computes the text 'index' inside this.el based on the currently selected node and its offset.
   * The offset might be inside a text node or inside a child node. In the latter case, we have to
```
typo

src/components/composer/content_editable_helper.ts
```diff
@@ -105,16 +105,25 @@ export class ContentEditableHelper {
...
+   * take account of all previous nodes contents. See the MDN documentation for more details.
```
hokolomopo, line 121: If we only read the docstring, it's a bit unclear that if(typeof node === TextNode), offset unit = number of characters, else offset unit = number of child nodes

hokolomopo (commented):
Nitpicking the docstring 🙄

src/components/composer/content_editable_helper.ts
rrahir: well that's what the hyperlinks are for, i don't feel like it's relevant to explain things in full details here

LucasLefevre (commented):
robodoo r+

PR #2321, opened 2023-04-05
[FIX] ContentEditableHelper: fix getCurrentSelection for Firefox

rrahir:
How to reproduce:
In firefox
- from the grid, input `=sum(`
- hit backspace -> the selection is set between `s` and `u` while it should be after `m`.

Starting from a cell content `=SUM(`  we have the composer content
```html
<span>=</span>
<span>sum</span>
<span>(</span>
```

When we delete the parenthesis, here is what google does
```html
<span>=</span>
<span>sum{document.selection}</span>
```
and firefox
```html
<span>=</span>
<span>sum</span>{document.selection}
```

A similar test can be done by writing text on multiple lines then try to select it all with 'Ctrl+A' shortcut. Once you release the keys, the selection will change to something incorrect rather than be set on the whole text.

The function `ContentEditableHelper.getCurrentSelection` is biased as it would assume that the selection would always end up in the deepest node (here a node `#text`).

This commit modifies the function to handle all selection nodes.

Task: 3071026

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/composer/content_editable_helper.ts
+++ b/src/components/composer/content_editable_helper.ts
@@ -147,16 +147,33 @@ export class ContentEditableHelper {
   getCurrentSelection() {
     let { startElement, endElement, startSelectionOffset, endSelectionOffset } =
       this.getStartAndEndSelection();
-    let startSizeBefore = this.findSizeBeforeElement(startElement!);
-    let endSizeBefore = this.findSizeBeforeElement(endElement!);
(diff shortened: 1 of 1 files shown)


LucasLefevre (commented):
robodoo r+

PR #2326, opened 2023-04-07
[FIX] filters: export xlsx of empty cells

hokolomopo:
## Description

441e9f8 imrpoved the handling of empty cells inside filters in the xlsx export, but didn't fix it fully.

When empty cells are present in the filtered zone, we should add an attribute `blank="1"` to the `filters` tag in the XML.

Odoo task ID : [3231170](https://www.odoo.com/web#id=3231170&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/xlsx/functions/table.ts
+++ b/src/xlsx/functions/table.ts
@@ -42,15 +42,10 @@ function addAutoFilter(table: ExcelFilterTableData): XMLString {
 }
 
 function addFilterColumns(table: ExcelFilterTableData): XMLString[] {
-  const tableZone = toZone(table.range);
   const columns: XMLString[] = [];
-  for (const i of range(0, zoneToDimension(tableZone).width)) {
-    const filter = table.filters[i];
-    if (!filter || !filter.filteredValues.length) {
-      continue;
-    }
+  for (const filter of table.filters) {
     const colXml = escapeXml/*xml*/ `
-      <filterColumn ${formatAttributes([["colId", i]])}>
+      <filterColumn ${formatAttributes([["colId", filter.colId]])}>
         ${addFilter(filter)}
       </filterColumn>
       `;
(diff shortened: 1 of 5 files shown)


src/xlsx/functions/table.ts
```diff
@@ -42,15 +42,13 @@ function addAutoFilter(table: ExcelFilterTableData): XMLString {
 }
 
 function addFilterColumns(table: ExcelFilterTableData): XMLString[] {
-  const tableZone = toZone(table.range);
   const columns: XMLString[] = [];
-  for (const i of range(0, zoneToDimension(tableZone).width)) {
-    const filter = table.filters[i];
-    if (!filter || !filter.filteredValues.length) {
+  for (const filter of table.filters) {
+    if (!filter || (!filter.filteredValues.length && !filter.displayBlanks)) {
```
LucasLefevre, line 47: seems already broken in 16.0, but it should be the other way around I think.
If you uncheck every value in o-spreadsheet and export, all values are checked.

hokolomopo: The condition was indeed wrong, but it turns out that the main problem is that we're not supposed to filter every value in the table column. It's impossible to do that in Excel and Calc. And it's a bit buggy if we import a filter where everything is filtered.

We could :
1) Disable filters where everything is filtered at export. That's really annoying 'cause we'd need to recompute teh hidden rows taking those disabled filters into account.
2) Don't allow a filter to filter everything. That's actually really annoying too, because our filters are dynamic (filtered row change when cell values change), which is not the case in Excel. We'd have to handle the case where a cell value change make every value filtered.
3) Accept that the export is a bit buggy in this case.

LucasLefevre: as discussed:
there's no easy fix, and the issue is minor: if you change anything in the filter in excel: everything is back to normal
so 3. seems reasonable

LucasLefevre (approved):
robodoo r+

PR #2327, opened 2023-04-10
[IMP] filter_menu: sort date values in chronological order

dhrp-odoo:
## Description:

Earlier, the filter menu values in the codebase sorting string values based on the standard alphabetical order, including dates, which was not the desired behavior

To resolve this issue, the `sortValues` function was added, which takes an array of values as input and correctly sorts them based on their data type.

Numeric values are sorted in ascending order, while dates are sorted in chronological order. Strings are sorted based on the standard alphabetical order, with numeric and case sensitivity taken into account.

Odoo task ID : [3234694](https://www.odoo.com/web#id=3234694&menu_id=4720&cids=2&action=333&active_id=2328&model=project.task&view_type=form)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/components/filters/filter_menu/filter_menu.ts
```diff
@@ -183,6 +181,40 @@ export class FilterMenu extends Component<Props, SpreadsheetChildEnv> {
...
+
+      switch (true) {
+        case aIsNumeric && bIsNumeric:
+          return +a - +b;
+        case aIsNumeric:
+          return -1;
+        case bIsNumeric:
+          return 1;
+        case aIsDate && bIsDate:
+          return +new Date(a) - +new Date(b);
```
LucasLefevre, line 206: It most probably won't work properly with other languages than English

dhrp-odoo: > It most probably won't work properly with other languages than English. I'm not sure there is a proper solution to the issue

Yes, that's correct. The `Date.parse()` method uses the English language conventions for parsing dates, and therefore it may not work correctly with input strings that use different languages. 

Maybe we can handle non-English date strings by converting them to English format before parsing them but I haven’t checked for its possibility yet.

On a second note, I guess we can ask PO how he thinks about this situation? What’s your opinion?

LucasLefevre: > Maybe we can handle non-English date strings by converting them to English format before parsing them

Well "untranslate" the date strings would be hacky, unreliable, fragile, and can work only for some very specific cases we choose to address.

>  I guess we can ask PO how he thinks about this situation?

François agreed to keep the current behavior. (since he agreed to cancel the task last week)

dhrp-odoo: Just to be sure, by current behavior you're referring to `not sorting` the dates or is it this one with English only sorting? However, personally I will say letting it work only for English (Words & Numbers) are still better than nothing. Let me know what should we aim for?

dhrp-odoo:
@LucasLefevre, I have made changes as per the conversation.

tests/components/filter_menu.test.ts
```diff
@@ -96,6 +96,40 @@ describe("Filter menu component", () => {
...
+    test("[Intentional] Dates in other languages (except 'english') are not sorted in chronological order", async () => {
+      createFilter(model, "A1:A5");
+      setCellContent(model, "A2", "Şubat 2023");
+      setCellContent(model, "A3", "Temmuz 2022");
+      setCellContent(model, "A4", "Nisan 2023");
+      setCellContent(model, "A5", "Ekim 2022");
+
+      await openFilterMenu();
+      const menuValues = getFilterMenuValues();
+      expect(menuValues.map((val) => val.value)).toEqual([
```
LucasLefevre, line 125: This test was useful :smile: 
I was wondering why the values were not ordered chronologically.
It turns out `Date.parse("sqfqsfdqsfqsdf 2020")` is happily parsed to the 01/01/2020 :man_facepalming: 
Hence it's considered a date instead of a regular string.

It's illustrated by this test I just wrote. It's passing in master but not with this branch
```ts
    test("sort according to letters before numbers", async () => {
      createFilter(model, "A1:A5");
      setCellContent(model, "A2", "AAA 2022");
      setCellContent(model, "A3", "BBB 2020");
      setCellContent(model, "A4", "CCC 2020");
      setCellContent(model, "A5", "DDD 2022");

      await openFilterMenu();
      const menuValues = getFilterMenuValues();
      expect(menuValues.map((val) => val.value)).toEqual([
        "AAA 2022",
        "BBB 2020",
        "CCC 2020",
        "DDD 2022",
      ]);
    });
```

LucasLefevre (commented):
this is very much harder than I thought. It doesn't even work properly for English without other side effects.
If you can find an alternative strategy, we can explore it. Otherwise, we'll cancel the task

tests/components/filter_menu.test.ts
dhrp-odoo: > this is very much harder than I thought. It doesn't even work properly for English without other side effects. If you can find an alternative strategy, we can explore it. Otherwise, we'll cancel the task

Yeah, I'm just wondering about that How “AAA 2022” will be converted to 1 January 2022.

Well, When I try to make it work for English dates only it also affects non-english dates.

I have made several attempts to address this issue, but the methods I have tried so far have proven to be both inflexible and expensive. I also attempted to utilize external libraries, such as Luxon, but found the implementation process to be overly complex. Additionally, Luxon requires a locale to function properly. Due to these challenges, I have not yet pushed the code. May I ask for your suggestion ?

LucasLefevre: > Yeah, I'm just wondering about that How “AAA 2022” will be converted to 1 January 2022.

The magic of javascipt `Date` :man_shrugging: 

> I also attempted to utilize external libraries, such as Luxon

We will have to introduce Luxon in the near future for another task (support local date formats). Maybe we will be able to come up with something when we will do it.

> May I ask for your suggestion ?

I'd cancel the task

dhrp-odoo: Alright, make sense, Thanks for your inputs

PR #2330, opened 2023-04-12
[IMP] find_and_replace: sheet granularity in Find & Replace

jash-odoo:
## Description:

Previously, the find and replace functionality only worked on the active sheet,
which made it difficult to determine if the content was available in any other
sheet or to replace content in all sheets. To address this issue, the code has
been updated to search for content in all available sheets and replace it in
all sheets.

Implemented range highlighting for user-selected search scopes to assist in
identifying the target area.

Maintained search index and match positions when switching sheets, preventing
unnecessary jumps to the first match.

Optimized row and column hiding behavior to manage values more efficiently during
search and replace operations.

Improvements:

Implemented a feature to find in 'allSheet', 'thisSheet', and
'specificRange' search scopes

Highlight the selected range in specificRange search scope.
Managed sheet changes in 'allSheet' mode. This is specifically for
scenarios where the user is not intentionally changing the sheet, but
it's instead due to matching queries

Use 'selectionInput' for obtaining specific ranges

Manage the state of specific ranges whenever we modify the 'searchScope'
Organized 'match' and 'matchIndex' first by 'sheetIndex', followed by
'row', and then by 'column'

Handled scenarios where the sheet changes due to a matching, which
previously caused a shift of focus from the side panel to the grid

Odoo task ID : [3250881](https://www.odoo.com/web#id=3250881&cids=2&menu_id=4720&action=333&active_id=2328&model=project.task&view_type=form)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/find_and_replace/find_and_replace.ts
+++ b/src/components/side_panel/find_and_replace/find_and_replace.ts
@@ -1,7 +1,10 @@
 import { Component, onMounted, onWillUnmount, useEffect, useRef } from "@odoo/owl";
-import { SearchOptions } from "../../../plugins/ui_feature";
+import { zoneToXc } from "../../../helpers";
+import { _t } from "../../../translation";
+import { SearchOptions } from "../../../types/find_and_replace";
 import { SpreadsheetChildEnv } from "../../../types/index";
 import { css } from "../../helpers/css";
+import { SelectionInput } from "../../selection_input/selection_input";
 
 css/* scss */ `
   .o-find-and-replace {
@@ -18,7 +21,7 @@ css/* scss */ `
       }
       .o-input-count {
         width: fit-content;
-        padding: 4 0 4 4;
+        padding: 4px 0 4px 4px;
       }
     }
   }
@@ -30,10 +33,12 @@ interface Props {
 
 export class FindAndReplacePanel extends Component<Props, SpreadsheetChildEnv> {
   static template = "o-spreadsheet-FindAndReplacePanel";
+  static components = { SelectionInput };
 
   private debounceTimeoutId;
-  private showFormulaState: boolean = false;
+  private initialShowFormulaState: boolean = false;
 
+  private dataRange: string = "";
   private searchInput = useRef("searchInput");
   private replaceInput = useRef("replaceInput");
 
@@ -57,15 +62,38 @@ export class FindAndReplacePanel extends Component<Props, SpreadsheetChildEnv> {
     return (this.replaceInput.el as HTMLInputElement)?.value || "";
   }
 
-  setup() {
-    this.showFormulaState = this.env.model.getters.shouldShowFormulas();
+  get allSheetsMatchesCount() {
+    return _t("%s in all sheets", this.env.model.getters.getAllSheetMatchesCount());
+  }
 
+  get currentSheetMatchesCount() {
+    return _t("%(matches)s in sheet %(sheetName)s", {
+      matches: this.env.model.getters.getActiveSheetMatchesCount(),
+      sheetName: this.env.model.getters.getSheetName(this.env.model.getters.getActiveSheetId()),
+    });
+  }
+
+  get specificRangeMatchesCount() {
+    const range = this.searchOptions.specificRange;
+    if (!range) {
+      return "";
+    }
+    const { _sheetId, _zone } = range;
+    return _t("%(matches)s in range %(range)s of sheet %(sheetName)s", {
+      matches: this.env.model.getters.getSpecificRangeMatchesCount().toString(),
+      range: zoneToXc(_zone),
+      sheetName: this.env.model.getters.getSheetName(_sheetId),
+    });
+  }
+
+  setup() {
+    this.initialShowFormulaState = this.env.model.getters.shouldShowFormulas();
     onMounted(() => this.searchInput.el?.focus());
 
     onWillUnmount(() => {
       clearTimeout(this.debounceTimeoutId);
       this.env.model.dispatch("CLEAR_SEARCH");
-      this.env.model.dispatch("SET_FORMULA_VISIBILITY", { show: this.showFormulaState });
+      this.env.model.dispatch("SET_FORMULA_VISIBILITY", { show: this.initialShowFormulaState });
     });
 
     useEffect(
@@ -77,6 +105,9 @@ export class FindAndReplacePanel extends Component<Props, SpreadsheetChildEnv> {
     );
   }
 
+  onFocusSearch() {
+    this.updateDataRange();
+  }
   onInput() {
     this.debouncedUpdateSearch();
   }
@@ -115,9 +146,32 @@ export class FindAndReplacePanel extends Component<Props, SpreadsheetChildEnv> {
     this.updateSearch({ matchCase });
   }
 
+  changeSearchScope(ev) {
+    const searchScope = ev.target.value;
+    this.updateSearch({ searchScope });
+  }
+
+  onSearchRangeChanged(ranges: string[]) {
+    this.dataRange = ranges[0];
+  }
+
+  updateDataRange() {
+    if (!this.dataRange) {
+      return;
+    }
+    if (this.searchOptions.searchScope === "specificRange") {
+      const specificRange = this.env.model.getters.getRangeFromSheetXC(
+        this.env.model.getters.getActiveSheetId(),
+        this.dataRange
+      ).rangeData;
+      this.updateSearch({ specificRange });
+    }
+  }
+
   onSelectPreviousCell() {
     this.env.model.dispatch("SELECT_SEARCH_PREVIOUS_MATCH");
   }
+
   onSelectNextCell() {
     this.env.model.dispatch("SELECT_SEARCH_NEXT_MATCH");
   }
--- a/src/components/side_panel/find_and_replace/find_and_replace.ts
+++ b/src/components/side_panel/find_and_replace/find_and_replace.ts
@@ -1,4 +1,5 @@
-import { Component, onMounted, onWillUnmount, useEffect, useRef, useState } from "@odoo/owl";
+import { Component, onMounted, onWillUnmount, useEffect, useRef } from "@odoo/owl";
+import { SearchOptions } from "../../../plugins/ui_feature";
 import { SpreadsheetChildEnv } from "../../../types/index";
 import { css } from "../../helpers/css";
 
@@ -27,23 +28,14 @@ interface Props {
   onCloseSidePanel: () => void;
 }
 
-interface FindAndReplaceState {
-  toSearch: string;
-  replaceWith: string;
-  searchOptions: {
-    matchCase: boolean;
-    exactMatch: boolean;
-    searchFormulas: boolean;
-  };
-}
-
 export class FindAndReplacePanel extends Component<Props, SpreadsheetChildEnv> {
   static template = "o-spreadsheet-FindAndReplacePanel";
-  private state: FindAndReplaceState = useState(this.initialState());
+
   private debounceTimeoutId;
   private showFormulaState: boolean = false;
 
   private searchInput = useRef("searchInput");
+  private replaceInput = useRef("replaceInput");
 
   get hasSearchResult() {
     return this.env.model.getters.getCurrentSelectedMatchIndex() !== null;
@@ -51,6 +43,18 @@ export class FindAndReplacePanel extends Component<Props, SpreadsheetChildEnv> {
 
   get pendingSearch() {
     return this.debounceTimeoutId !== undefined;
+  }
+
+  get searchOptions() {
+    return this.env.model.getters.getSearchOptions();
+  }
+
+  get toSearch() {
+    return (this.searchInput.el as HTMLInputElement)?.value || "";
+  }
+
+  get toReplace() {
+    return (this.replaceInput.el as HTMLInputElement)?.value || "";
   }
 
   setup() {
@@ -66,15 +70,14 @@ export class FindAndReplacePanel extends Component<Props, SpreadsheetChildEnv> {
 
     useEffect(
       () => {
-        this.state.searchOptions.searchFormulas = this.env.model.getters.shouldShowFormulas();
-        this.searchFormulas();
+        const showFormula = this.env.model.getters.shouldShowFormulas();
+        this.updateSearch({ searchFormulas: showFormula });
       },
       () => [this.env.model.getters.shouldShowFormulas()]
     );
   }
 
-  onInput(ev) {
-    this.state.toSearch = ev.target.value;
+  onInput() {
     this.debouncedUpdateSearch();
   }
 
@@ -94,11 +97,22 @@ export class FindAndReplacePanel extends Component<Props, SpreadsheetChildEnv> {
     }
   }
 
-  searchFormulas() {
+  searchFormulas(ev) {
+    const showFormula = ev.target.checked;
     this.env.model.dispatch("SET_FORMULA_VISIBILITY", {
-      show: this.state.searchOptions.searchFormulas,
+      show: showFormula,
     });
-    this.updateSearch();
+    this.updateSearch({ searchFormulas: showFormula });
+  }
+
+  searchExactMatch(ev) {
+    const exactMatch = ev.target.checked;
+    this.updateSearch({ exactMatch });
+  }
+
+  searchMatchCase(ev) {
+    const matchCase = ev.target.checked;
+    this.updateSearch({ matchCase });
   }
 
   onSelectPreviousCell() {
@@ -107,10 +121,15 @@ export class FindAndReplacePanel extends Component<Props, SpreadsheetChildEnv> {
   onSelectNextCell() {
     this.env.model.dispatch("SELECT_SEARCH_NEXT_MATCH");
   }
-  updateSearch() {
+
+  updateSearch(updateSearchOptions?: Partial<SearchOptions>) {
+    const searchOptions = {
+      ...this.env.model.getters.getSearchOptions(),
+      ...updateSearchOptions,
+    };
     this.env.model.dispatch("UPDATE_SEARCH", {
-      toSearch: this.state.toSearch,
-      searchOptions: this.state.searchOptions,
+      toSearch: this.toSearch,
+      searchOptions,
     });
   }
 
@@ -124,30 +143,14 @@ export class FindAndReplacePanel extends Component<Props, SpreadsheetChildEnv> {
 
   replace() {
     this.env.model.dispatch("REPLACE_SEARCH", {
-      replaceWith: this.state.replaceWith,
+      replaceWith: this.toReplace,
     });
   }
 
   replaceAll() {
     this.env.model.dispatch("REPLACE_ALL_SEARCH", {
-      replaceWith: this.state.replaceWith,
+      replaceWith: this.toReplace,
     });
-  }
-
-  // ---------------------------------------------------------------------------
-  // Private
-  // ---------------------------------------------------------------------------
-
-  private initialState(): FindAndReplaceState {
-    return {
-      toSearch: "",
-      replaceWith: "",
-      searchOptions: {
-        matchCase: false,
-        exactMatch: false,
-        searchFormulas: false,
-      },
-    };
   }
 }
 
--- a/src/components/side_panel/find_and_replace/find_and_replace.xml
+++ b/src/components/side_panel/find_and_replace/find_and_replace.xml
@@ -16,29 +16,29 @@
             /
             <t t-esc="env.model.getters.getSearchMatches().length"/>
           </div>
-          <div t-elif="!pendingSearch and state.toSearch !== ''" class="o-input-count">0 / 0</div>
+          <div t-elif="!pendingSearch and toSearch !== ''" class="o-input-count">0 / 0</div>
         </div>
         <div>
           <!-- TODO: go through this css, the group misses a padding and could profit from BootStrap -->
           <label class="o-checkbox mt-1">
             <input
-              t-model="state.searchOptions.matchCase"
-              t-on-change="updateSearch"
+              t-att-checked="searchOptions.matchCase"
+              t-on-change="searchMatchCase"
               type="checkbox"
             />
             <span>Match case</span>
           </label>
           <label class="o-checkbox">
             <input
-              t-model="state.searchOptions.exactMatch"
-              t-on-change="updateSearch"
+              t-att-checked="searchOptions.exactMatch"
+              t-on-change="searchExactMatch"
               type="checkbox"
             />
             <span>Match entire cell content</span>
           </label>
           <label class="o-checkbox">
             <input
-              t-model="state.searchOptions.searchFormulas"
+              t-att-checked="searchOptions.searchFormulas"
               t-on-change="searchFormulas"
               type="checkbox"
             />
@@ -63,7 +63,7 @@
           <input
             type="text"
             class="o-input o-input-without-count"
-            t-model="state.replaceWith"
+            t-ref="replaceInput"
             t-on-keydown="onKeydownReplace"
           />
         </div>
--- a/src/components/side_panel/find_and_replace/find_and_replace.xml
+++ b/src/components/side_panel/find_and_replace/find_and_replace.xml
@@ -9,6 +9,7 @@
             t-ref="searchInput"
             class="o-input o-input-with-count"
             t-on-input="onInput"
+            t-on-focus="onFocusSearch"
             t-on-keydown="onKeydownSearch"
           />
           <div class="o-input-count" t-if="hasSearchResult">
@@ -17,6 +18,22 @@
             <t t-esc="env.model.getters.getSearchMatches().length"/>
           </div>
           <div t-elif="!pendingSearch and toSearch !== ''" class="o-input-count">0 / 0</div>
+        </div>
+        <select
+          class="o-input o-type-selector o-type-range-selector mt-3 mb-3"
+          t-on-change="changeSearchScope">
+          <option value="allSheets">All sheets</option>
+          <option value="activeSheet">Current sheet</option>
+          <option value="specificRange">Specific range</option>
+        </select>
+        <div t-if="searchOptions.searchScope === 'specificRange'">
+          <SelectionInput
+            ranges="() => [this.dataRange]"
+            onSelectionChanged="(ranges) => this.onSearchRangeChanged(ranges)"
+            onSelectionConfirmed.bind="updateDataRange"
+            hasSingleRange="true"
+            required="true"
+          />
         </div>
         <div>
           <!-- TODO: go through this css, the group misses a padding and could profit from BootStrap -->
@@ -44,6 +61,17 @@
             />
             <span>Search in formulas</span>
           </label>
+        </div>
+        <div class="o-matches-count mt-4" t-if="toSearch !== ''">
+          <div t-if="searchOptions.searchScope === 'specificRange' and dataRange != ''">
+            <t t-esc="specificRangeMatchesCount"/>
+          </div>
+          <div>
+            <t t-esc="currentSheetMatchesCount"/>
+          </div>
+          <div>
+            <t t-esc="allSheetsMatchesCount"/>
+          </div>
         </div>
       </div>
       <div class="o-sidePanelButtons">
--- a/src/components/spreadsheet/spreadsheet.ts
+++ b/src/components/spreadsheet/spreadsheet.ts
@@ -394,7 +394,7 @@ export class Spreadsheet extends Component<SpreadsheetProps, SpreadsheetChildEnv
   }
   focusGrid() {
     if (!this._focusGrid) {
-      throw new Error("_focusGrid should be exposed by the grid component");
+      return;
     }
     this._focusGrid();
   }
--- a/src/components/spreadsheet/spreadsheet.ts
+++ b/src/components/spreadsheet/spreadsheet.ts
@@ -4,7 +4,9 @@ import {
   onPatched,
   onWillUnmount,
   onWillUpdateProps,
+  useEffect,
   useExternalListener,
+  useRef,
   useState,
   useSubEnv,
 } from "@odoo/owl";
@@ -255,6 +257,7 @@ export class Spreadsheet extends Component<SpreadsheetProps, SpreadsheetChildEnv
 
   sidePanel!: SidePanelState;
   composer!: ComposerState;
+  spreadsheetRef = useRef("spreadsheet");
 
   private _focusGrid?: () => void;
 
@@ -295,6 +298,24 @@ export class Spreadsheet extends Component<SpreadsheetProps, SpreadsheetChildEnv
       clipboard: this.env.clipboard || instantiateClipboard(),
       startCellEdition: (content?: string) => this.onGridComposerCellFocused(content),
     });
+
+    useEffect(
+      () => {
+        /**
+         * Only refocus the grid if the active element is not a child of the spreadsheet
+         * (i.e. activeElement is outside of the spreadsheetRef component)
+         * and spreadsheet is a child of that element. Anything else means that the focus
+         * is on an element that needs to keep it.
+         */
+        if (
+          !this.spreadsheetRef.el!.contains(document.activeElement) &&
+          document.activeElement?.contains(this.spreadsheetRef.el!)
+        ) {
+          this.focusGrid();
+        }
+      },
+      () => [this.env.model.getters.getActiveSheetId()]
+    );
 
     useExternalListener(window as any, "resize", () => this.render(true));
     useExternalListener(window, "beforeunload", this.unbindModelEvents.bind(this));
--- a/src/plugins/ui_feature/find_and_replace.ts
+++ b/src/plugins/ui_feature/find_and_replace.ts
@@ -44,7 +44,11 @@ interface SearchMatch {
 
 export class FindAndReplacePlugin extends UIPlugin {
   static layers = [LAYERS.Search];
-  static getters = ["getSearchMatches", "getCurrentSelectedMatchIndex"] as const;
+  static getters = [
+    "getSearchMatches",
+    "getCurrentSelectedMatchIndex",
+    "getSearchOptions",
+  ] as const;
   private searchMatches: SearchMatch[] = [];
   private selectedMatchIndex: number | null = null;
   private currentSearchRegex: RegExp | null = null;
@@ -114,6 +118,9 @@ export class FindAndReplacePlugin extends UIPlugin {
   }
   getCurrentSelectedMatchIndex(): number | null {
     return this.selectedMatchIndex;
+  }
+  getSearchOptions(): SearchOptions {
+    return this.searchOptions;
   }
   // ---------------------------------------------------------------------------
   // Search
--- a/src/plugins/ui_feature/find_and_replace.ts
+++ b/src/plugins/ui_feature/find_and_replace.ts
@@ -1,24 +1,12 @@
 import { escapeRegExp } from "../../helpers";
 import { canonicalizeNumberContent } from "../../helpers/locale";
-import {
-  CellPosition,
-  Color,
-  Command,
-  GridRenderingContext,
-  HeaderIndex,
-  LAYERS,
-  Position,
-} from "../../types/index";
+import { isInside } from "../../helpers/zones";
+import { SearchOptions, SearchOptionsInternal } from "../../types/find_and_replace";
+import { CellPosition, Color, Command, GridRenderingContext, LAYERS } from "../../types/index";
 import { UIPlugin } from "../ui_plugin";
 
 const BORDER_COLOR: Color = "#8B008B";
 const BACKGROUND_COLOR: Color = "#8B008B33";
-
-export interface SearchOptions {
-  matchCase: boolean;
-  exactMatch: boolean;
-  searchFormulas: boolean;
-}
 
 enum Direction {
   previous = -1,
@@ -26,10 +14,8 @@ enum Direction {
   next = 1,
 }
 
-interface SearchMatch {
+interface SearchMatch extends CellPosition {
   selected: boolean;
-  col: HeaderIndex;
-  row: HeaderIndex;
 }
 
 /**
@@ -48,17 +34,39 @@ export class FindAndReplacePlugin extends UIPlugin {
     "getSearchMatches",
     "getCurrentSelectedMatchIndex",
     "getSearchOptions",
+    "getAllSheetMatchesCount",
+    "getActiveSheetMatchesCount",
+    "getSpecificRangeMatchesCount",
   ] as const;
-  private searchMatches: SearchMatch[] = [];
+
+  private allSheetsMatches: SearchMatch[] = [];
+  private activeSheetMatches: SearchMatch[] = [];
+  private specificRangeMatches: SearchMatch[] = [];
+
+  // fixme: why do we make selectedMatchIndex on top of a selected
+  // property in the matches?
   private selectedMatchIndex: number | null = null;
   private currentSearchRegex: RegExp | null = null;
-  private searchOptions: SearchOptions = {
+  private searchOptions: SearchOptionsInternal = {
     matchCase: false,
     exactMatch: false,
     searchFormulas: false,
+    searchScope: "allSheets",
+    specificRange: undefined,
   };
   private toSearch: string = "";
   private isSearchDirty = false;
+
+  get searchMatches(): SearchMatch[] {
+    switch (this.searchOptions.searchScope) {
+      case "allSheets":
+        return this.allSheetsMatches;
+      case "activeSheet":
+        return this.activeSheetMatches;
+      case "specificRange":
+        return this.specificRangeMatches;
+    }
+  }
 
   // ---------------------------------------------------------------------------
   // Command Handling
@@ -67,7 +75,9 @@ export class FindAndReplacePlugin extends UIPlugin {
   handle(cmd: Command) {
     switch (cmd.type) {
       case "UPDATE_SEARCH":
-        this.updateSearch(cmd.toSearch, cmd.searchOptions);
+        const rangeData = cmd.searchOptions.specificRange;
+        const specificRange = rangeData && this.getters.getRangeFromRangeData(rangeData);
+        this.updateSearch(cmd.toSearch, { ...cmd.searchOptions, specificRange });
         break;
       case "CLEAR_SEARCH":
         this.clearSearch();
@@ -97,7 +107,9 @@ export class FindAndReplacePlugin extends UIPlugin {
         this.isSearchDirty = true;
         break;
       case "ACTIVATE_SHEET":
-        this.refreshSearch();
+        if (this.searchOptions.searchScope === "activeSheet") {
+          this.isSearchDirty = true;
+        }
         break;
     }
   }
@@ -116,12 +128,26 @@ export class FindAndReplacePlugin extends UIPlugin {
   getSearchMatches(): SearchMatch[] {
     return this.searchMatches;
   }
+
   getCurrentSelectedMatchIndex(): number | null {
     return this.selectedMatchIndex;
   }
   getSearchOptions(): SearchOptions {
-    return this.searchOptions;
+    return { ...this.searchOptions, specificRange: this.searchOptions.specificRange?.rangeData };
   }
+
+  getAllSheetMatchesCount(): number {
+    return this.allSheetsMatches.length;
+  }
+
+  getActiveSheetMatchesCount(): number {
+    return this.activeSheetMatches.length;
+  }
+
+  getSpecificRangeMatchesCount(): number {
+    return this.specificRangeMatches.length;
+  }
+
   // ---------------------------------------------------------------------------
   // Search
   // ---------------------------------------------------------------------------
@@ -130,7 +156,7 @@ export class FindAndReplacePlugin extends UIPlugin {
    * Will update the current searchOptions and accordingly update the regex.
    * It will then search for matches using the regex and store them.
    */
-  private updateSearch(toSearch: string, searchOptions: SearchOptions) {
+  private updateSearch(toSearch: string, searchOptions: SearchOptionsInternal) {
     this.searchOptions = searchOptions;
     if (toSearch !== this.toSearch) {
       this.selectedMatchIndex = null;
@@ -139,14 +165,16 @@ export class FindAndReplacePlugin extends UIPlugin {
     this.updateRegex();
     this.refreshSearch();
   }
+
   /**
    * refresh the matches according to the current search options
    */
   private refreshSearch() {
-    const matches: SearchMatch[] = this.findMatches();
-    this.searchMatches = matches;
+    this.selectedMatchIndex = null;
+    this.findMatches();
     this.selectNextCell(Direction.current);
   }
+
   /**
    * Updates the regex based on the current searchOptions and
    * the value toSearch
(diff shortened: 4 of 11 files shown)


src/plugins/ui_feature/find_and_replace.ts
```diff
@@ -138,37 +154,119 @@ export class FindAndReplacePlugin extends UIPlugin {
...
-  private sortByRowThenColumn(a: Position, b: Position) {
-    if (a.row === b.row) {
-      return a.col - b.col;
+  private findMatchesInSheet(
+    sheetId: string,
+    sheetIndex: number,
+    matches: SearchMatch[],
+    specificRange?: string
+  ) {
+    let top, left, bottom, right;
```
hokolomopo, line 217: Don't ever, unless you have no other choice, type variable as `any`. Your code can do random stuff, because `rangeData._zone.bottom` can be undefined, and you didn't notice it because you didn't type teh variable. The whole point of typescript is to avoid such errors.

src/plugins/ui_feature/find_and_replace.ts
```diff
@@ -138,37 +154,119 @@ export class FindAndReplacePlugin extends UIPlugin {
...
-      }
+        break;
+      case "thisSheet":
+        const sheetId = this.getters.getActiveSheetId();
+        const sheetIndex = this.getters.getSheetIds().indexOf(sheetId);
+        this.findMatchesInSheet(sheetId, sheetIndex, matches);
+        break;
+      case "specificRange":
+        const specificRange = this.searchOptions.specificRange;
+        if (specificRange) {
```
hokolomopo, line 180:
```suggestion
        if (!specificRange) {
        return []
        }
```

src/plugins/ui_feature/find_and_replace.ts
```diff
@@ -138,37 +154,119 @@ export class FindAndReplacePlugin extends UIPlugin {
...
+          let rangeSheet, rangeCells;
+
+          if (specificRange.includes("!")) {
+            [rangeSheet, rangeCells] = specificRange.split("!");
+          } else {
+            rangeSheet = this.getters.getSheetName(this.getters.getActiveSheetId());
+            rangeCells = specificRange;
```
hokolomopo, line 187: Use `getters.getRangeFromSheetXC`, don't to it by hand.

src/plugins/ui_feature/find_and_replace.ts
```diff
@@ -138,37 +154,119 @@ export class FindAndReplacePlugin extends UIPlugin {
...
+    sheetId: string,
+    sheetIndex: number,
+    matches: SearchMatch[],
+    specificRange?: string
+  ) {
+    let top, left, bottom, right;
+
+    if (specificRange) {
+      const rangeData = this.getters.getRangeDataFromXc(sheetId, specificRange);
+      top = rangeData._zone.top;
```
rrahir, line 221: _{fieldName} is supposed not  to be accessed from the outside. I think you are looking for `getRangeFromSheetXC`

src/plugins/ui_feature/find_and_replace.ts
```diff
@@ -138,37 +154,119 @@ export class FindAndReplacePlugin extends UIPlugin {
...
+            [rangeSheet, rangeCells] = specificRange.split("!");
+          } else {
+            rangeSheet = this.getters.getSheetName(this.getters.getActiveSheetId());
+            rangeCells = specificRange;
+          }
+
+          const [startCell, endCell] = rangeCells.split(":");
+
+          const finalEndCell = endCell || startCell;
+          const specificRangeWithEndCell = `${startCell}:${finalEndCell}`;
```
hokolomopo, line 193: We obviously have helpers to do this, there's no way we'd want to parse a string into a zone by hand... You even used one (even if it was the wrong one) in `findMatchesInSheet`, why would you do it by hand here ??

src/plugins/ui_feature/find_and_replace.ts
hokolomopo, line 221: That's a bit my bad, I should have written some doc for ranges, but the fact that you have to acess private fields (prefixed by underscore) should tell you that there's something wrong.

The correct getter to use is `getRangeFromSheetXc`

src/plugins/ui_feature/find_and_replace.ts
hokolomopo: And if you're gonna use `top, left, right, bottom`, just directly use a `Zone`...

src/plugins/ui_feature/find_and_replace.ts
```diff
@@ -138,37 +154,119 @@ export class FindAndReplacePlugin extends UIPlugin {
...
+      const rangeData = this.getters.getRangeDataFromXc(sheetId, specificRange);
+      top = rangeData._zone.top;
+      left = rangeData._zone.left;
+      bottom = rangeData._zone.bottom;
+      right = rangeData._zone.right;
+    } else {
+      top = 0;
+      left = 0;
+      bottom = this.getters.getNumberCols(sheetId);
+      right = this.getters.getNumberRows(sheetId);
```
hokolomopo, line 229: `zone = this.getters.getSheetZone(sheetId)`

src/plugins/ui_feature/find_and_replace.ts
```diff
@@ -138,37 +154,119 @@ export class FindAndReplacePlugin extends UIPlugin {
...
-    return a.row > b.row ? 1 : -1;
+
+    const cells = this.getters.getCells(sheetId);
+
+    for (const cell of Object.values(cells)) {
+      const { col, row } = this.getters.getCellPosition(cell.id);
+
+      if (col < left || col > right || row < top || row > bottom) {
+        continue;
+      }
```
hokolomopo, line 239: Do the same thing that in the getter `getEvaluatedCellsInZone` instead of fetching all the cells and skipping tose not in the zone

hokolomopo: You probably don't even need the cell itself if the only thing you do is th get its position for `getSearchableString`

src/plugins/ui_feature/find_and_replace.ts
```diff
@@ -216,45 +329,57 @@ export class FindAndReplacePlugin extends UIPlugin {
   /**
    * Replace the value of the currently selected match
    */
-  private replace(replaceWith: string) {
-    if (this.selectedMatchIndex === null || !this.currentSearchRegex) {
+  private replace(replaceWith: string, updateSearch: boolean = true) {
+    const selectedMatch =
+      this.selectedMatchIndex !== null ? this.searchMatches[this.selectedMatchIndex] : null;
```
hokolomopo, line 334: Opinion : early return if this.selectedMatchIndex is null is more redable than a ternary

src/plugins/ui_feature/find_and_replace.ts
```diff
@@ -216,45 +329,57 @@ export class FindAndReplacePlugin extends UIPlugin {
...
-    const selectedMatch = matches[this.selectedMatchIndex];
-    const sheetId = this.getters.getActiveSheetId();
-    const cell = this.getters.getCell({ sheetId, ...selectedMatch });
+    const cell = this.getters.getCell(selectedMatch);
     if (cell?.isFormula && !this.searchOptions.searchFormulas) {
-      this.selectNextCell(Direction.next);
+      if (updateSearch) {
+        this.selectNextCell(Direction.next);
+      }
     } else {
```
hokolomopo, line 343: do eary return rather than if/else.  And add empty lines sometimes. The whole function is kinda hard to read

src/plugins/ui_feature/find_and_replace.ts
```diff
@@ -216,45 +329,57 @@ export class FindAndReplacePlugin extends UIPlugin {
...
     }
-    const matches = this.searchMatches;
-    const selectedMatch = matches[this.selectedMatchIndex];
-    const sheetId = this.getters.getActiveSheetId();
-    const cell = this.getters.getCell({ sheetId, ...selectedMatch });
+    const cell = this.getters.getCell(selectedMatch);
     if (cell?.isFormula && !this.searchOptions.searchFormulas) {
-      this.selectNextCell(Direction.next);
+      if (updateSearch) {
+        this.selectNextCell(Direction.next);
```
hokolomopo, line 341: We probably should do the `selectNextCell` outside of `this.replace`, it's probably better than having an argument on whether we should do the update here or not

hokolomopo (commented):
The UI part is mostly fine, except for a wrong props that makes the `SelectionInput` not work.

There is a lot of problems in the plugin part. 

The tests are missng, but that was announced.

src/plugins/ui_feature/find_and_replace.ts
```diff
@@ -138,37 +154,119 @@ export class FindAndReplacePlugin extends UIPlugin {
...
+      right = rangeData._zone.right;
+    } else {
+      top = 0;
+      left = 0;
+      bottom = this.getters.getNumberCols(sheetId);
+      right = this.getters.getNumberRows(sheetId);
     }
-    return a.row > b.row ? 1 : -1;
+
+    const cells = this.getters.getCells(sheetId);
```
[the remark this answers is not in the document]
rrahir: Actually, it's useless to call the cells here as  we always end up calling `getSearchableString` which will fetch the cell - So you really only need to iterate through the positions  inside the zone, which can be easily generated with  `function positions(zone: Zone)`.

src/plugins/ui_feature/find_and_replace.ts
```diff
@@ -138,37 +154,119 @@ export class FindAndReplacePlugin extends UIPlugin {
...
+              this.findMatchesInSheet(sheet, sheetIndex, matches, specificRangeWithEndCell);
+            }
+          }
+        }
+        break;
     }
-    return matches.sort(this.sortByRowThenColumn);
+
+    // Sort the matches by sheet index, row, and column
+    return matches.sort(this.sortBySheetIndexThenRowThenColumn);
```
rrahir, line 208: the sheetIndex reference here can completely be removed because you decide on your own what the sheetIndex is and use it to sort things out. It would be way simpler to fetch the matches, sort them and then concatenate them  sheet by sheet afterwards, you'd achieve the exact same result.

src/plugins/ui_feature/find_and_replace.ts
```diff
@@ -138,37 +154,119 @@ export class FindAndReplacePlugin extends UIPlugin {
...
   }
 
-  private sortByRowThenColumn(a: Position, b: Position) {
-    if (a.row === b.row) {
-      return a.col - b.col;
+  private findMatchesInSheet(
+    sheetId: string,
+    sheetIndex: number,
+    matches: SearchMatch[],
```
rrahir, line 214: Remove the aggregator and just generate and return the list you go through in this function.
If you have to call the function several times, you can concatenate the results

src/plugins/ui_feature/find_and_replace.ts
```diff
@@ -138,37 +154,119 @@ export class FindAndReplacePlugin extends UIPlugin {
...
+          row,
+          selected: false,
+        };
+        matches.push(match);
+      }
+    }
+  }
+
+  private sortBySheetIndexThenRowThenColumn(a: SearchMatch, b: SearchMatch) {
+    return a.sheetIndex !== b.sheetIndex
```
rrahir, line 259: sheetIndex is useless, you can rollback the change to the previous implementation of `sortByRowThenColumn`

rrahir (changes requested):
First pass, Still have to go through some of the component logic and Adrien is also reviewing this

src/components/side_panel/find_and_replace/find_and_replace.ts
```diff
@@ -35,6 +36,36 @@ css/* scss */ `
...
+    display: flex;
+    align-items: center;
+    margin-bottom: 8px;
+  }
+
+  .o-dropdown-group .o-dropdown {
+    width: 100%;
+    margin-top: 10px;
+    padding: 4px 8px;
+    font-size: 14px;
```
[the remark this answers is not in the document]
rrahir: this is something that we do in other places as well :/ I talked to François about this

src/plugins/ui_feature/find_and_replace.ts
```diff
@@ -274,6 +399,10 @@ export class FindAndReplacePlugin extends UIPlugin {
     const sheetId = this.getters.getActiveSheetId();
 
     for (const match of this.searchMatches) {
+      if (match.sheetId !== sheetId) {
```
rrahir, line 402: 👍

src/plugins/ui_feature/find_and_replace.ts
```diff
@@ -85,9 +90,20 @@ export class FindAndReplacePlugin extends UIPlugin {
         this.clearSearch();
         break;
       case "ACTIVATE_SHEET":
+        this.userInitiatedSheetChange = true;
```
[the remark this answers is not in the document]
rrahir: won't work as intended either.
e.g.
1- on sheet 1, write ="test"
2- switch to sheet 2
3- open find & replace side panel
4- start typing 't' -> it says it found one match: ok
5- now continue typing 'e' -> the active sheet now switched to Sheet1

the issue with the boolean approach here is that it will work for 1 handling of the command (step 4° but not for the following steps (step 5)

src/types/commands.ts
```diff
@@ -778,6 +780,15 @@ export interface RefreshSearchCommand {
...
 
+export interface UpdateSearchScopeCommand {
+  type: "UPDATE_SEARCH_SCOPE";
+  searchScope: SearchOptions["searchScope"];
+}
+
+export interface UpdateSpecificRangeCommand {
+  type: "UPDATE_SPECIFIC_RANGE";
+  specificRange: SearchOptions["specificRange"];
+}
```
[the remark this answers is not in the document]
rrahir: agreed, update_search seems sufficient

rrahir (changes requested):
Found several functional issue that I will write down in the pad once the migrtion od odoo.com is done

src/plugins/ui_feature/find_and_replace.ts
```diff
@@ -85,9 +90,20 @@ export class FindAndReplacePlugin extends UIPlugin {
         this.clearSearch();
```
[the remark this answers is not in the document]
jash-odoo: It's possible that the reason for not clearing the search during UPDATE_CELL is because updating a cell's content does not affect the positions of the search matches. On the other hand, when adding or removing columns/rows, the search may become invalid because the column and row indexes of the cells have changed.

jash-odoo:
@rrahir and @hokolomopo can you please check this again?

src/plugins/ui_feature/find_and_replace.ts
```diff
@@ -222,47 +328,71 @@ export class FindAndReplacePlugin extends UIPlugin {
...
-        content: newContent,
-      });
-      this.searchMatches.splice(this.selectedMatchIndex, 1);
-      this.selectNextCell(Direction.current);
+      return this.selectNextCell(Direction.next);
+    }
+    const replaceRegex = new RegExp(
+      this.currentSearchRegex.source,
+      this.currentSearchRegex.flags + "g"
+    );
```
hokolomopo, line 338: We recently bumped our version of JS, we can drop this part and use `replaceAll`

src/plugins/ui_feature/find_and_replace.ts
```diff
@@ -289,5 +423,16 @@ export class FindAndReplacePlugin extends UIPlugin {
...
       }
     }
+    if (this.confirmedSelectedZone && this.confirmedSelectedZone.sheetId === sheetId) {
+      const selectedZone = this.confirmedSelectedZone.zone;
+      if (this.searchOptions.searchScope === "specificRange") {
+        if (!selectedZone) {
+          return;
+        }
+        const { x, y, width, height } = this.getters.getVisibleRect(selectedZone);
+        ctx.strokeStyle = "#FA9C1B";
```
hokolomopo, line 433: the color should be a constant in `constant.ts`

tests/plugins/find_and_replace.test.ts
```diff
@@ -15,85 +15,252 @@ let searchOptions: SearchOptions;
...
+      const matchIndex = model.getters.getCurrentSelectedMatchIndex();
+      if (searchScope === "thisSheet") {
+        expect(matches).toHaveLength(3);
+        expect(matchIndex).toStrictEqual(0);
+        expect(matches[0]).toStrictEqual({
+          sheetId: "Sheet1",
+          col: 0,
+          row: 1,
+          selected: true,
+        });
```
hokolomopo, line 56: These expect taking 6 lines each make the tests huge and frankly unreadable.

tests/plugins/find_and_replace.test.ts
```diff
@@ -15,85 +15,252 @@ let searchOptions: SearchOptions;
...
-    const matches = model.getters.getSearchMatches();
-    const matchIndex = model.getters.getCurrentSelectedMatchIndex();
-    expect(matches).toHaveLength(4);
-    expect(matchIndex).toStrictEqual(0);
-    expect(matches[0]).toStrictEqual({ col: 0, row: 1, selected: true });
-    expect(matches[1]).toStrictEqual({ col: 0, row: 2, selected: false });
-    expect(matches[2]).toStrictEqual({ col: 0, row: 3, selected: false });
-    expect(matches[3]).toStrictEqual({ col: 0, row: 4, selected: false });
-  });
+  test.each(["allSheets", "thisSheet", "specificRange"])(
```
hokolomopo, line 38: why would you ever do a test.each if there is a big `if/else` in the test wich next to nothing in common for all the cases ?

tests/plugins/find_and_replace.test.ts
```diff
@@ -102,194 +269,789 @@ describe("basic search", () => {
...
+          selected: false,
+        });
+        expect(matches[1]).toStrictEqual({
+          sheetId: "s1",
+          col: 0,
+          row: 1,
+          selected: true,
+        });
+        expect(matches[2]).toStrictEqual({
+          sheetId: "Sheet2",
```
hokolomopo, line 642: sheetID should be `s2`, it's confusion that the format of the ids of sheet1 and 2 are different, it looks like a sheet Id and a sheet name

tests/plugins/find_and_replace.test.ts
```diff
@@ -102,194 +269,789 @@ describe("basic search", () => {
...
+          col: 0,
+          row: 2,
+          selected: true,
+        });
+      }
+    }
+  );
+  //   model.dispatch("UPDATE_SEARCH", { toSearch: "1", searchOptions });
+  //   let matches = model.getters.getSearchMatches();
+  //   let matchIndex = model.getters.getCurrentSelectedMatchIndex();
```
hokolomopo, line 932: commented code

hokolomopo (commented):
Didn't read the tests in much details, these are very hard to read with your prettier splitting everything into 4894 lines.

And the next time you ping us _twice_ to get a review make sure to fix _**ALL**_ the comments of the previous reviews .....

src/plugins/ui_feature/find_and_replace.ts
```diff
@@ -222,47 +328,71 @@ export class FindAndReplacePlugin extends UIPlugin {
...
   }
+
   /**
    * Apply the replace function to all the matches one time.
    */
   private replaceAll(replaceWith: string) {
-    const matchCount = this.searchMatches.length;
-    for (let i = 0; i < matchCount; i++) {
+    this.selectedMatchIndex = 0;
+    while (this.searchMatches.length > 0) {
```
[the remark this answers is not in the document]
jash-odoo: I have reverted to the old code, and the issue I was facing earlier seems to be working fine. Now my question is, are you suggesting me to change the current behavior of replace OR just use the old code for `replaceAll` as it is? Please correct me if I have misinterpreted your message

src/plugins/ui_feature/find_and_replace.ts
```diff
@@ -138,35 +152,116 @@ export class FindAndReplacePlugin extends UIPlugin {
...
+    const rangeData = specificRange
+      ? this.getters.getRangeDataFromXc(sheetId, specificRange)._zone
+      : this.getters.getSheetZone(sheetId);
+
+    const boundedZone: Zone = {
+      left: rangeData.left,
+      top: rangeData.top,
+      right: rangeData.right ?? this.getters.getNumberCols(sheetId),
+      bottom: rangeData.bottom ?? this.getters.getNumberRows(sheetId),
+    };
```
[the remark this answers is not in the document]
jash-odoo: Previously, the cells were obtained using the "getCells" method, which iterated over the entire sheet. However, in cases where a specific range is required, this method is unnecessary. In order to address this issue, the range data is determined and the zone is retrieved if a specific range is present. However, it should be noted that the "getRangeDataFromXc" method may return an unbounded zone, thus requiring the zone to be bounded.

If the "getEvaluatedCellsInZone" method is used instead of "getCells", the cell ID is not received automatically. Therefore, it is necessary to manually add the columns and rows to determine the range.

Please let me know if there is any other possible way to deal with this situation.

src/plugins/ui_feature/find_and_replace.ts
```diff
@@ -138,35 +152,116 @@ export class FindAndReplacePlugin extends UIPlugin {
...
+        break;
     }
-    return matches.sort(this.sortByRowThenColumn);
+
+    for (const sheetId of sheets) {
+      matches.push(...this.findMatchesInSheet(sheetId, specificRange));
+    }
+    const sortedMatches = matches.sort((a, b) =>
+      this.sortBySheetinitialStateIndexThenRowThenColumn(a, b)
+    );
```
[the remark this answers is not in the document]
jash-odoo: Consider the following scenario:

1. The text "here" is written in Sheet 1, Row 2.
2. The text "here" is written in Sheet 2, Row 1.
3. The text "here" is written in Sheet 3, Row 3.
4. The active sheet is set to Sheet 1.
In this case, if we directly sort (current behavior), the first match will be found in Sheet 2, followed by Sheet 1, and finally Sheet 3. (AllSheet mode)

It should be noted that if the active sheet is Sheet 3, the first match should be found in Sheet 3.

That's why we need the sheetIndex to first give priority to activeSheet find results followed by indexing all the remaining searches according to the sheets. It also prevents from suddenly jumping from activeSheet to the "first found" result sheet.

src/plugins/ui_feature/find_and_replace.ts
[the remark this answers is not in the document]
hokolomopo: https://github.com/odoo/o-spreadsheet/pull/2330#discussion_r1169737983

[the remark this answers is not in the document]
rrahir: Hence the comment from @hokolomopo that you should NOT use  `getRangeDataFromXc` but rather use `getRangeFromSheetXc` .This will yield a `Range` `which has a public attribute `range.zone: Zone` which, as you can see in the definition of its class, will be a `Zone`and not an ÙnboundedZone`.  So the general comment in the last review of Adrien stands, please address every remark beforehand - we're not making progress here ...

src/components/side_panel/find_and_replace/find_and_replace.ts
```diff
@@ -72,12 +81,58 @@ export class FindAndReplacePanel extends Component<Props, SpreadsheetChildEnv> {
...
+  private focusButton(buttonName: string) {
+    setTimeout(() => {
+      const el = this.selectionInputref.el!;
+      if (el) {
+        const buttonElement = this.findAndReplaceRef.el!.querySelector(
+          `button[name=${buttonName}]`
+        ) as HTMLButtonElement;
+        buttonElement.focus();
+      }
+    }, 0);
```
hokolomopo, line 125: That's a hack alright. I won't comment on the `setTimeout`, there's probably a way to do it less hack but without proper focus managment it's not easy.

But did you even test it ? My comment was `when we search something on all sheets, when the search switch sheet the side panel lose focus and the grid takes it`. This is not fixed at all ? Why are we doing stuff on the selecitonInput where we want to search input to be focused ? And why focus the button ??

Start by writign a fixing test when fixing thing. And the fix should aim to make the test pass.

src/components/side_panel/find_and_replace/find_and_replace.ts
```diff
@@ -72,12 +81,58 @@ export class FindAndReplacePanel extends Component<Props, SpreadsheetChildEnv> {
...
+  }
+
+  onFocusSearch() {
+    const el = this.selectionInputref.el!;
+    if (el) {
+      const confirmButton = el.querySelector(".o-btn-action.o-selection-ok") as HTMLButtonElement;
+      confirmButton.click();
+    }
```
hokolomopo, line 133: If you need to do `querySelector` to select something on a child component, your code is most likely wrong. Can't you mimick the behaviour of the confirm button of the input, ie. dispatch  a `UNFOCUS_SELECTION_INPUT`. I think it will cancel the selection rather than conforming it, but it's not that bad. And functionnaly its better to call this on a click anywhere on the side pannel rather than on focus on the search input.

 Or probably better, and I think I told you to do that, SelectionInput could call it's confirm on `blur`. Probably with a new props `confirmOnBlur :boolean`.

hokolomopo (commented):
Write tests when you're fixing stuff. I still have the same bugs as before.

src/components/side_panel/find_and_replace/find_and_replace.ts
rrahir: I have investigated the issue partially and well it seems that we are trying to fight back against the 'autofocus' of the grid when changing sheet. Upon investigation, it seems that the effect was too radical as it would steal the focus of any focusable element inside spreadsheet (be it the composer, or in this very case, a selection input). I have added a commit that aims to solve this issue, I don't think you will need this hack anymore.

src/components/side_panel/find_and_replace/find_and_replace.ts
```diff
@@ -72,12 +81,58 @@ export class FindAndReplacePanel extends Component<Props, SpreadsheetChildEnv> {
...
+        ) {
+          this.state.searchOptions.specificRange = this.lastSpecificRange;
+        } else if (
+          this.state.searchOptions.searchScope !== "specificRange" &&
+          this.state.searchOptions.specificRange
+        ) {
+          this.lastSpecificRange = this.state.searchOptions.specificRange;
+          this.state.searchOptions.specificRange = undefined;
+        }
+        this.updateSearch();
```
rrahir, line 105: Couldn't you just conserve the value of specificRange at all times and not send it in the command payload when `searchScopre !== "specificRange"`?

rrahir (commented):
The commit title contains a typo ;-)

Don't forget the different comments left by Adrien that are not addressed yet.

Also I wonder if we really need all those states in the component as it seems (I might be wrong) that every change of the different parameters, (scope,match, rangedata) are followed by a dispatch which will imply a full rerendering

src/plugins/ui_feature/find_and_replace.ts
jash-odoo: when using replaceAll, it's important to consider the case sensitivity. Here's the scenario:

1. We've got a cell with the content "FORMAT."
2. Now, let's say we want to find the word "for" and replace it with "dor."

Due to its case sensitivity, replaceAll won't allow us to make that change.

jash-odoo: when using replaceAll, it's important to consider the case sensitivity. Here's the scenario:

1. We've got a cell with the content "FORMAT."
2. Now, let's say we want to find the word "for" and replace it with "dor."

Due to its case sensitivity, replaceAll won't allow us to make that change.

hokolomopo: I don't see why the case sensitivity change anything. The flag for case sensitivity will be in `this.currentSearchRegex`, calling replace or replaceAll don't change that ?

rrahir: [ReplaceAll](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll) take regex as paremeter, you don't have to  rebuild it with the global `g` anymore

jash-odoo: If I don't rebuild regex with global `g` I'm getting [error](https://deepscan.io/docs/rules/bad-replace-all-arg). This is also written in above mentioned documentation. I may be wrong but I'm a bit confused since both things are contradicting.

src/components/side_panel/find_and_replace/find_and_replace.xml
```diff
@@ -88,7 +108,8 @@
         <button
           t-att-disabled="env.model.getters.getCurrentSelectedMatchIndex() === null"
           t-on-click="replace"
-          class="o-sidePanelButton">
+          class="o-sidePanelButton"
+          name="replace">
```
rrahir, line 112: is this useful?

rrahir (changes requested):
getting close :)

src/plugins/ui_feature/find_and_replace.ts
```diff
@@ -289,5 +386,16 @@ export class FindAndReplacePlugin extends UIPlugin {
         }
       }
     }
+    if (this.searchOptions.specificRange && this.searchOptions.specificRange.sheetId === sheetId) {
```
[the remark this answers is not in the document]
rrahir: retrospectively, I understand what you have tried to achieve, having the zone still visible even when the highlight is not focues. if so, I think it's worth checking the functional behaviour with @Francois-frgi   but at the very least, the zone should be the same shade of color as the match *, not orange with purple match

* maybe some darker purple, or even the same purple but without transpalrency might do the trick

src/plugins/ui_feature/find_and_replace.ts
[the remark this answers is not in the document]
jash-odoo: Already resolved this issue.

src/plugins/ui_feature/find_and_replace.ts
```diff
@@ -138,35 +150,93 @@ export class FindAndReplacePlugin extends UIPlugin {
...
+    return matches;
   }
 
-  private sortByRowThenColumn(a: Position, b: Position) {
-    if (a.row === b.row) {
-      return a.col - b.col;
-    }
-    return a.row > b.row ? 1 : -1;
+  private sortBySheetinitialStateIndexThenRowThenColumn(a: SearchMatch, b: SearchMatch) {
+    return a.sheetId !== b.sheetId
```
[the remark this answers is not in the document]
jash-odoo: Yes, you are absolutely right. In our case, the search results are already ordered. However, we also need to know the index of each search result. To achieve this, if we have the search sequence from Sheet 1, we can determine the index of each search result accurately.

Let's consider a scenario where the search scope is set to all sheets, and we are currently on Sheet 2. If we find something that already exists on Sheet 1, our functionality will start searching from Sheet 2. In this situation, if we don't have the search sequence from Sheet 1, we won't be able to understand the search index accurately. Therefore, we require this function to obtain the search sequence from Sheet 1.

src/components/side_panel/find_and_replace/find_and_replace.ts
```diff
@@ -111,6 +120,28 @@ export class FindAndReplacePanel extends Component<Props, SpreadsheetChildEnv> {
     this.updateSearch();
   }
 
+  onSearchRangeChanged(ranges: string[]) {
+    this.dataRange = ranges;
```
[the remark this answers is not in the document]
jash-odoo: Am just extract the first element of ranges, updating that accordingly.

src/plugins/ui_feature/find_and_replace.ts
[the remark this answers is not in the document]
Francois-frgi: @jash-odoo I like the fact that we can see the range while looking for reference but, for consistency, it might indeed look better to align colors (either putting all in orange or all in purple).

cc: @rrahir

src/components/side_panel/find_and_replace/find_and_replace.ts
```diff
@@ -101,7 +111,6 @@ export class FindAndReplacePanel extends Component<Props, SpreadsheetChildEnv> {
 
   onFocusSidePanel() {
     this.state.searchOptions.searchFormulas = this.env.model.getters.shouldShowFormulas();
-    this.env.model.dispatch("REFRESH_SEARCH");
```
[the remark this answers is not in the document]
jash-odoo: Previously, there was an issue where, in "All sheets" mode, clicking on the side panel after manually changing the active sheet would take you back to the previous sheet.

However consider the case where we are on the 3rd search of Sheet 1 and click on the side panel. This will refresh the search and move us back to the first search.

src/plugins/ui_feature/find_and_replace.ts
```diff
@@ -138,35 +150,93 @@ export class FindAndReplacePlugin extends UIPlugin {
...
+
+  private findMatchesInSheet(sheetId: string, specificRange?: Range) {
+    const matches: SearchMatch[] = [];
+    const rangeData =
+      this.searchOptions.searchScope === "specificRange" && specificRange
+        ? specificRange.zone
+        : this.getters.getSheetZone(sheetId);
+
+    if (!this.currentSearchRegex) {
+      return [];
```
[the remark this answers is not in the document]
jash-odoo: We assign currentSearchRegex as null in the initial state and also when clearing the search. Therefore, we receive a TypeScript error indicating that it can be null. To resolve this, we can initialize it with `new RegExp("")`, and when clearing the search, we can set the same value to currentSearchRegex. By doing so, we can eliminate the need for this condition

src/plugins/ui_feature/find_and_replace.ts
```diff
@@ -138,35 +150,93 @@ export class FindAndReplacePlugin extends UIPlugin {
...
-        if (
-          cell &&
-          this.currentSearchRegex &&
-          this.currentSearchRegex.test(this.getSearchableString({ sheetId, col, row }))
-        ) {
-          const match: SearchMatch = { col, row, selected: false };
+    for (let row = rangeData.top; row <= rangeData.bottom; row++) {
+      for (let col = rangeData.left; col <= rangeData.right; col++) {
+        const position: CellPosition = { sheetId, col, row };
+        if (this.currentSearchRegex.test(this.getSearchableString(position))) {
```
rrahir, line 221:
```suggestion
        if (this.currentSearchRegex?.test(this.getSearchableString(position))) {
```
Optional chaining

src/plugins/ui_feature/find_and_replace.ts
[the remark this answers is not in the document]
rrahir: you ddon't understand. this is what I meant https://github.com/odoo/o-spreadsheet/pull/2330/files#r1229381286

src/plugins/ui_feature/find_and_replace.ts
jash-odoo: I tried doing that, but it's still effecting the test cases. Whenever we search for =1 in test case, it won't work but in UI it seems to be working fine. Unless I'm doing something wrong. One more thing, these failing test cases are getting error on search regex but our fix is implemented for replace. Do you have any other approach in mind?

src/components/side_panel/find_and_replace/find_and_replace.ts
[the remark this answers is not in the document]
rrahir: Well then we introduce another issue - 
Open find & replace, start a search
go to top bar > view > show formula
refocus the side panel
the 'search in formula' boolean is ticked but the search is not updated so I don't think it's the proper fix.

the problem is that we'd refresh the search at every focus whereas we should have only done it when the value of "shouldShowFormulas" was updated. That being said, I alseo think that reacting on 'focus' was a bad idea fro the start and we should rather react when the 'showFormula' is updated for the whole sheet.
this can be achieved with a useEffect that spies on `this.env.model.getters.shouldShowFormulas`

src/plugins/ui_feature/find_and_replace.ts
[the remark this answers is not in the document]
rrahir: Ok, I think I understand the issue. If I'm not mistaken, the search index is computed in the plugin such that it would be correctly displayed in the sidepanel. I don't think it's good or bad in itself but since it's important for the sidepanel*, there should be an explicitely related test. I don't think there is because when I removed the sorting function, the side panel tests did not fail.

So, I think that the function findMatches just does too many things. it should not change the state of the plugin just yield the results, regardless of the current sheet etc. 
the logic of the offset should be handled on the fly in `selectNextCell` when `selectedMatchIndex=== null`, instead of relying on a variable computed elsewhere. Otherwise, you have 2 functions but one of them needs to be called before the second one for the code to make any sense.

What I would do in `refreshSearch`:
- Before calling findMatches, reset selectedMatchIndex
- this.searchMatches = this.findMatches()
At this point we have properly reset the state according to the new search regex
- `this.findMatches` only finds the matches by looping over all sheets starting with the first one
- then call SelectNextCell (which probably should be called `selectNextMatch` btw) and in that function IIF `selectedMatchIndex === null` then you compute the offset

rrahir (changes requested):

src/plugins/ui_feature/find_and_replace.ts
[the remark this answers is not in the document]
jash-odoo: Hello @rrahir, could you please review the changes? I have added the test for this and made the necessary adjustments accordingly.

src/components/side_panel/find_and_replace/find_and_replace.ts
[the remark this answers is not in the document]
rrahir: I think the method onfocussidepanel has become obsolete  now

rrahir:
there are STILL functional issues and errors due to the bad use of component props. the same errors that were mentioned by  @hokolomopo  in his first review https://github.com/odoo/o-spreadsheet/pull/2330#discussion_r1169648933  the test is still not present. I want you to go through this pull request, through every message and make sure you addressed them.

Now for some actual content review:

1/  I have committed a change in your branch; I expect you not to erase them by push forcing once again without looking at the remote.

2/ IMHO there is a functional issue with the "this sheet' behaviour. If I select "this sheet" with a search term, and then change sheet, I don't have any visual clue that my search does not concern the new sheet. This is something we should discuss with @Francois-frgi

src/plugins/ui_feature/find_and_replace.ts
```diff
@@ -258,11 +324,23 @@ export class FindAndReplacePlugin extends UIPlugin {
...
+
+    const isColHidden = this.getters.isColHidden(sheetId, col);
+    const isRowHidden = this.getters.isRowHidden(sheetId, row);
+
+    if (isColHidden || isRowHidden) {
+      return "";
+    }
+
+    if (!this.searchOptions.searchFormulas && cell.isFormula) {
+      return this.getters.getEvaluatedCell(position).formattedValue;
```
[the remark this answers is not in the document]
rrahir: It is indeed an known issue with the evaluation. you can take the same approach as FilterEvaluationPlugin and FilterEvaluationPlugin which monitor the commands dispatched and when deemed necessary, update their internal state in `finalize`. That function is guaranteed to be executed after the evaluation of the cells

src/components/spreadsheet/spreadsheet.ts
```diff
@@ -226,6 +229,23 @@ export class Spreadsheet extends Component<SpreadsheetProps, SpreadsheetChildEnv
       startCellEdition: (content: string) => this.onGridComposerCellFocused(content),
     });
 
+    useEffect(
+      () => {
```
rrahir, line 233: this should be part of another dedicated commit

src/components/spreadsheet/spreadsheet.ts
```diff
@@ -222,6 +225,23 @@ export class Spreadsheet extends Component<SpreadsheetProps, SpreadsheetChildEnv
...
+        /**
+         * Only refocus the grid if the active element is not a child of the spreadsheet
+         * and spreadsheet is a child of that element. Anything else means that the focus
+         * is on an element that needs to keep it.
+         */
+        if (
+          !this.spreadsheetRef.el!.contains(document.activeElement) &&
+          document.activeElement?.contains(this.spreadsheetRef.el!)
```
VincentSchippefilt, line 237: for future reference, this means "activeElement is outside the spreadsheetRef component"

rrahir:
there are still several issues with the translations, the values displayed, the third commit is introducing some weird behavior that could mislead the user. I will take over the task to address these.

tests/plugins/find_and_replace.test.ts
```diff
@@ -145,133 +218,200 @@ describe("basic search", () => {
...
+    const matchIndex = model.getters.getCurrentSelectedMatchIndex();
+    expect(matches).toHaveLength(4);
+    expect(matchIndex).toStrictEqual(3);
+    expect(matches[0]).toStrictEqual({ sheetId: "s1", col: 0, row: 1, selected: false });
+    expect(matches[1]).toStrictEqual({ sheetId: "s1", col: 0, row: 2, selected: false });
+    expect(matches[2]).toStrictEqual({ sheetId: "s1", col: 0, row: 3, selected: false });
+    expect(matches[3]).toStrictEqual({ sheetId: "s2", col: 0, row: 1, selected: true });
+  });
+
+  test("modifying cells update the search", () => {
```
hokolomopo, line 240: what the hell is the difference between this and "refresh search when cell is updated" ? I do'nt know and I do'nt care to read it if I have no idea what its about reading the test title

tests/plugins/find_and_replace.test.ts
```diff
@@ -145,133 +218,200 @@ describe("basic search", () => {
...
     matches = model.getters.getSearchMatches();
     matchIndex = model.getters.getCurrentSelectedMatchIndex();
-    expect(matches).toHaveLength(6);
+    expect(getActivePosition(model)).toBe("A2");
+    expect(matches).toHaveLength(4);
+    expect(matchIndex).toStrictEqual(3);
+    expect(matches[3]).toStrictEqual({ sheetId: "s2", col: 0, row: 1, selected: true });
+  });
+
+  test("update cell will also update the search", () => {
```
hokolomopo, line 294: "also"? Is there also no difference with "refresh search when cell is updated" and "modifying cells update the search" ?

tests/plugins/find_and_replace.test.ts
```diff
@@ -145,133 +218,200 @@ describe("basic search", () => {
...
-    activateSheet(model, sheet2);
+    setCellContent(model, "B2", '="hey"');
     matches = model.getters.getSearchMatches();
     matchIndex = model.getters.getCurrentSelectedMatchIndex();
     expect(matches).toHaveLength(3);
     expect(matchIndex).toStrictEqual(0);
-    activateSheet(model, sheet1);
+  });
+
+  test.skip("Will search a modified cell", () => {
```
hokolomopo, line 322: Skip ? And how is it the forth freaking test with a **_slightly different name that means exactly the same thing_**

tests/plugins/find_and_replace.test.ts
```diff
@@ -32,62 +32,135 @@ let searchOptions: SearchOptions;
...
     expect(matches).toHaveLength(4);
     expect(matchIndex).toStrictEqual(0);
-    expect(matches[0]).toStrictEqual({ col: 0, row: 1, selected: true });
-    expect(matches[1]).toStrictEqual({ col: 0, row: 2, selected: false });
-    expect(matches[2]).toStrictEqual({ col: 0, row: 3, selected: false });
-    expect(matches[3]).toStrictEqual({ col: 0, row: 4, selected: false });
+    expect(matches[0]).toStrictEqual({ sheetId: "s1", col: 0, row: 1, selected: true });
+    expect(matches[1]).toStrictEqual({ sheetId: "s1", col: 0, row: 2, selected: false });
+    expect(matches[2]).toStrictEqual({ sheetId: "s1", col: 0, row: 3, selected: false });
+    expect(matches[3]).toStrictEqual({ sheetId: "s2", col: 0, row: 1, selected: false });
```
hokolomopo, line 148: Yeah there's way too much stuff that we don't care about in the beforeEach, we dont need to test 3-4 matches in each and every test...

tests/plugins/find_and_replace.test.ts
```diff
@@ -145,133 +218,200 @@ describe("basic search", () => {
...
     model.dispatch("UPDATE_SEARCH", { toSearch: "1", searchOptions });
-    expect(model.getters.getSearchMatches()).toHaveLength(4);
+    expect(model.getters.getSearchMatches()).toHaveLength(3);
     model.dispatch("UPDATE_SEARCH", { toSearch: "", searchOptions });
     expect(model.getters.getSearchMatches()).toHaveLength(0);
   });
 
-  test.skip("Will search a modified cell", () => {
-    // not implemented
+  test("start searching from current sheet and update index", () => {
```
hokolomopo, line 226: test name tells us what we do, but not what is suppoed to happen

tests/plugins/find_and_replace.test.ts
```diff
@@ -145,133 +218,200 @@ describe("basic search", () => {
...
+    matches = model.getters.getSearchMatches();
+    matchIndex = model.getters.getCurrentSelectedMatchIndex();
+    expect(matches.length).toBe(3);
+    expect(matchIndex).toStrictEqual(0);
+    expect(matches[0]).toStrictEqual({ sheetId: "s1", col: 1, row: 0, selected: true });
+    expect(matches[1]).toStrictEqual({ sheetId: "s1", col: 0, row: 2, selected: false });
+    expect(matches[2]).toStrictEqual({ sheetId: "s1", col: 0, row: 3, selected: false });
+  });
+
+  test("search begins from current sheet with index starting at the first sheet in allSheets search scope", () => {
```
hokolomopo, line 260: that's the exact same test as "start searching from current sheet and update index"?? with one more line ?

tests/plugins/find_and_replace.test.ts
```diff
@@ -145,133 +218,200 @@ describe("basic search", () => {
...
+    expect(getActivePosition(model)).toBe("A2");
     expect(matches).toHaveLength(4);
+    expect(matchIndex).toStrictEqual(3);
+    expect(matches[0]).toStrictEqual({ sheetId: "s1", col: 0, row: 1, selected: false });
+    expect(matches[1]).toStrictEqual({ sheetId: "s1", col: 0, row: 2, selected: false });
+    expect(matches[2]).toStrictEqual({ sheetId: "s1", col: 0, row: 3, selected: false });
+    expect(matches[3]).toStrictEqual({ sheetId: "s2", col: 0, row: 1, selected: true });
+  });
+
+  test("search in 2nd Sheet with Scope activeSheet and Change to allSheets Starting searching from Current Sheet", () => {
```
hokolomopo, line 275: that's a lot of words with random Capitalization to not Say what the Test is supposed to be Doing, because I Sure Don't know what Is supposed to happend when Searching in the second Sheet with a scope of ActiveSheet but then I change to AllSheet and Stuff happens

tests/plugins/find_and_replace.test.ts
```diff
@@ -290,163 +430,343 @@ describe("basic search", () => {
...
+    const sheet2 = "s2";
+    createSheet(model, { sheetId: sheet2 });
+    setCellContent(model, "A3", "1", sheet2);
+    searchOptions = {
+      matchCase: false,
+      exactMatch: false,
+      searchFormulas: false,
+      searchScope: "activeSheet",
+      specificRange: { _sheetId: sheet2, _zone: toZone("A1:D13") },
+    };
```
hokolomopo, line 472: create a DEFAULT_SEARCH_OPTION constant if you're about to repeat the same thing for the 4th beforeEach

tests/plugins/find_and_replace.test.ts
```diff
@@ -290,163 +430,343 @@ describe("basic search", () => {
...
-    expect(matches[1]).toStrictEqual({ col: 0, row: 3, selected: false });
-    expect(matches[2]).toStrictEqual({ col: 0, row: 4, selected: false });
-    expect(matches[3]).toStrictEqual({ col: 0, row: 5, selected: false });
+    expect(matches[0]).toStrictEqual({ sheetId: "s1", col: 0, row: 1, selected: true });
+    expect(matches[1]).toStrictEqual({ sheetId: "s1", col: 0, row: 3, selected: false });
+    expect(matches[2]).toStrictEqual({ sheetId: "s1", col: 0, row: 4, selected: false });
   });
 });
+
 describe("next/previous cycle", () => {
```
hokolomopo, line 457: Yeah I didn't read all that because I kinda gave up, but I'm sure there no need for three whole hundreds and twenty lines of code to test "previous/next". And even less so after we tested that all the matches are correct for 450 lines in the previous tests

hokolomopo (commented):
ok

src/plugins/ui_feature/find_and_replace.ts
```diff
@@ -44,26 +30,52 @@ interface SearchMatch {
...
     matchCase: false,
     exactMatch: false,
     searchFormulas: false,
+    searchScope: "allSheets",
+    specificRange: undefined,
   };
   private toSearch: string = "";
   private isSearchDirty = false;
 
+  get searchMatches(): SearchMatch[] {
```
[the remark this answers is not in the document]
hokolomopo: Every other plugin starts with allowDispatch/handle (or range management), which I think is nice because that's the most important part of the plugin. But it's a detail

tests/components/find_replace_side_panel.test.ts
```diff
@@ -117,10 +152,61 @@ describe("find and replace sidePanel component", () => {
...
+      setSelection(model, ["A1:B2"]);
+      await nextTick();
+      await simulateClick(selectors.confirmSearchRange);
+      expect((document.querySelector(selectors.searchRange) as HTMLInputElement).value).toBe(
+        "A1:B2"
+      );
+    });
+
+    test.each(["allSheets", "activeSheet"] as const)(
+      "%s after changing selection will re-select the match",
```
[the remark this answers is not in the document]
rrahir: `Changing scope to %s after changing selection will re-select the match`?
If it's the idea, then I think the test doesn't even belong  in the side panel tests

LucasLefevre (commented):
ploup :)

src/plugins/ui_feature/find_and_replace.ts
```diff
@@ -44,26 +30,52 @@ interface SearchMatch {
 
 export class FindAndReplacePlugin extends UIPlugin {
   static layers = [LAYERS.Search];
-  static getters = ["getSearchMatches", "getCurrentSelectedMatchIndex"] as const;
-  private searchMatches: SearchMatch[] = [];
+  static getters = [
+    "getSearchMatches",
+    "getCurrentSelectedMatchIndex",
+    "getSearchOptions",
+    "getTotalMatchesAllSheets",
```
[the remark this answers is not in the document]
rrahir: `get{Scope}MatchesCount` ?

src/types/find_and_replace.ts
```diff
@@ -0,0 +1,13 @@
+import { Range, RangeData } from "./range";
+
+export interface SearchOptions {
+  matchCase: boolean;
+  exactMatch: boolean;
+  searchFormulas: boolean;
+  searchScope: "allSheets" | "activeSheet" | "specificRange";
+  specificRange?: RangeData;
```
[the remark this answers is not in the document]
rrahir: it is typed for both the getter AND the command, which is supposed to be serializable (at least for now). It felt a bit weird to have a different type between the get & set, meaning that the component would have to do the conversion, whereas we only do it inside the plugin now.

tests/spreadsheet/side_panel_component.test.ts
```diff
@@ -151,20 +153,21 @@ describe("Side Panel", () => {
     expect(onCloseSidePanel).toHaveBeenCalled();
   });
 
-  test("Switching from one sidepanel to another executes the onCloseSidePanel callback", async () => {
```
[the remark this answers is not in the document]
rrahir: rebase issue :/

src/types/find_and_replace.ts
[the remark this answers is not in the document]
LucasLefevre: ok.
It's been a while I've been thinking we should rework those unbounded ranges. The first version never evolved and can be improved.

LucasLefevre (approved):
robodoo r+ rebase-ff

PR #2333, opened 2023-04-13
[FIX] core: Prevent dispatch of invalid command payloads

rrahir:
One can dispatch commands that try to delete or hide  sheet headers. Unfortunately, those commands are not rejecetd at the allowDispatch but more importanttly, the handlers assume that the payload is correct and do not verify that the elements passed actually make sense.

E.g.

from a standard sheet with 26 columns:

- dispatch a command to delete column with index 26 and 27
```typescript
model.dispatch('REMOVE_COLUMNS_ROWS',
    { sheetId: model.getters.getActiveSheetId(), dimension: "COL", elements: [26,27]})  
```
-> visually, you only see the first 24 columns even though you haven't really deleted the last ('Y' and 'Z')

Task: 3272589

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/plugins/core/header_visibility.ts
```diff
@@ -27,9 +27,14 @@ export class HeaderVisibilityPlugin extends CorePlugin {
           cmd.dimension === "COL"
             ? this.getters.getNumberCols(cmd.sheetId)
             : this.getters.getNumberRows(cmd.sheetId);
-        return (hiddenGroup || []).flat().concat(cmd.elements).length < elements
-          ? CommandResult.Success
-          : CommandResult.TooManyHiddenElements;
+        const hiddenElements = (hiddenGroup || []).flat();
+        if (hiddenElements.length >= elements) {
+          return CommandResult.TooManyHiddenElements;
```
hokolomopo, line 32: Idk if we really want to support it, but techincally we could do dispatch("HIDE_COL", indexes : [ 0,0,0,0,0,0,0,0,0,0,0,0,0....] and then this condition would be wrong. That's a bit of an edge casez and a malformed command tho.

hokolomopo (commented):
looks good 👍

src/plugins/core/header_visibility.ts
hokolomopo: Wait no, you meant to compare `cmd.elements.length  + hiddenElements.length>= elements` right ? You compare the elements currently hidden in the sheet, with the number of elements in the sheet there.

rrahir: yup I messed up the adaptation -  I fixed the condition and cast everything in sets to purge multiple entries as well  -  I feel like we should have the command purge at the start of dispatch inside model but we have to discuss this. model does not know what goes on in the plugins but it sure does know some of the commands - we could argue about their definition ... @LucasLefevre

LucasLefevre: I'm not sure to understand what you mean

rrahir:
@hokolomopo  updated ^^

tests/plugins/header_visibility.test.ts
```diff
@@ -313,4 +314,24 @@ describe("Hide Rows", () => {
     hideRows(model, [0]);
     expect(model.getters.getSelectedZones()).toEqual([toZone("A1:D4")]);
   });
+
+  test("Cannot hide unexisting columns", () => {
+    const model = new Model();
```
hokolomopo, line 319: nitpick: already a model created in `beforeEach`

hokolomopo (commented):
Test nitpicking for lul

PR #2340, opened 2023-04-13
[16.0] Various easy performance boosts

hokolomopo:
## [IMP] header_visibility: improve perfs of `findVisibleHeader`

The way the getter `findVisibleHeader` was implemented as that it received
an argument with all the indexes it need to check to find a visible header.

This means that if we want to get the first visible col of a sheet,
we have to first generate an array with all the indexes of the cols of the
sheet, then looping on this array to find a visible col. This was very
wasteful as :

1) we created an array that needed to be garbage collected after
2) the array contains all the indexes, but most of the time we only
loop on the first few

Changed the arguments of `findVisibleHeader` to receive `from` and `to`
arguments.

Benchmark :
on a sheet 26 cols x 10.000 rows filled with numbers, try to delete the last
column

Before : ~2050ms. Most time taken by :
- Garbage Collection : ~737ms
- range() : ~729ms

After: ~542ms. Most time taken by :
- Garbage Collection : ~124ms
- getRowTallestCellSize() : ~105ms

## [IMP] sheet: move cells methods performance

The methods `moveCellOn*()` in `sheet.ts` did an useless `parseInt()` to
loop through the indexes of an array, slightly decreasing the performance.

For when we really have to parse a string to an int, Number() is also
slightly faster than parseInt().

This saves 20-40ms on a 10.000 rows sheet with 26 columns, when deleting the
first column.

## [IMP] formats: performance of `splitNumber()`

To compute the formatted value of a number, we need to split it into
its integer digits and decimal digits. This was done by the function
`splitNumber()`, which was using Intl.NumberFormat, which handles
everything for us, but is quite slow.

Now convert the number to string using Number.toString(), and handle
the max number of decimal and the rounding by hand. We will still use
Intl.NumberFormat for numbers too big/small, for which Number.toString()
returns an exponential notation.

Rough benchmark :
Sheet 26 cols x 10.000 rows, using a CF on the whole sheet to force the
evaluation of the lazy values.

w/ sheet filled with integers:
old: splitNumber ~299ms
new: splitNumber ~9.5ms

w/ sheet filled with floats:
old: splitNumber ~555ms
new: splitNumber ~30ms

## [IMP] model: store handler arrays

Currently the model create a new array every time `this.handlers` or
`this.coreHandlers` are accessed. That means that
we create new arrays every time we dispatch a command, leading to
quite a bit of garbage collection.

When deleting the first column of a 26 x 1000 sheet filled with numbers,
initializing the arrays saves ~100ms in garbage collection.

Odoo task ID : [3272878](https://www.odoo.com/web#id=3272878&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/format.ts
+++ b/src/helpers/format.ts
@@ -188,12 +188,98 @@ const numberRepresentation: Intl.NumberFormat[] = [];
  * - all digit stored in the decimal part of the number
  *
  * The 'maxDecimal' parameter allows to indicate the number of digits to not
- * exceed in the decimal part, in which case digits are rounded
+ * exceed in the decimal part, in which case digits are rounded.
  *
- * Intl.Numberformat is used to properly handle all the roundings.
- * e.g. 1234.7  with format ### (<> maxDecimals=0) should become 1235, not 1234
  **/
 function splitNumber(
+  value: number,
+  maxDecimals: number = MAX_DECIMAL_PLACES
+): { integerDigits: string; decimalDigits: string | undefined } {
+  const asString = value.toString();
+  if (asString.includes("e")) return splitNumberIntl(value, maxDecimals);
+
+  if (Number.isInteger(value)) {
+    return { integerDigits: asString, decimalDigits: undefined };
+  }
+
+  const indexOfDot = asString.indexOf(".");
+  let integerDigits = asString.substring(0, indexOfDot);
+  let decimalDigits: string | undefined = asString.substring(indexOfDot + 1);
+
+  if (maxDecimals === 0) {
+    if (Number(decimalDigits[0]) >= 5) {
+      integerDigits = (Number(integerDigits) + 1).toString();
+    }
+    return { integerDigits, decimalDigits: undefined };
+  }
+
+  if (decimalDigits.length > maxDecimals) {
+    const { integerDigits: roundedIntegerDigits, decimalDigits: roundedDecimalDigits } =
+      limitDecimalDigits(decimalDigits, maxDecimals);
+
+    decimalDigits = roundedDecimalDigits;
+    if (roundedIntegerDigits !== "0") {
+      integerDigits = (Number(integerDigits) + Number(roundedIntegerDigits)).toString();
+    }
+  }
+
+  return { integerDigits, decimalDigits: removeTrailingZeroes(decimalDigits || "") };
+}
+
+/**
+ *  Return the given string minus the trailing "0" characters.
+ *
+ * @param numberString : a string of integers
+ * @returns the numberString, minus the eventual zeroes at the end
+ */
+function removeTrailingZeroes(numberString: string): string | undefined {
+  let i = numberString.length - 1;
+  while (i >= 0 && numberString[i] === "0") {
+    i--;
+  }
+  return numberString.slice(0, i + 1) || undefined;
+}
+
+/**
+ * Limit the size of the decimal part of a number to the given number of digits.
+ */
+function limitDecimalDigits(
+  decimalDigits: string,
+  maxDecimals: number
+): {
+  integerDigits: string;
+  decimalDigits: string | undefined;
+} {
+  let integerDigits = "0";
+  let resultDecimalDigits: string | undefined = decimalDigits;
+
+  // Note : we'd want to simply use number.toFixed() to handle the max digits & rounding,
+  // but it has very strange behaviour. Ex: 12.345.toFixed(2) => "12.35", but 1.345.toFixed(2) => "1.34"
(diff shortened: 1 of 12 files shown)


src/helpers/misc.ts
```diff
@@ -212,6 +212,22 @@ export function range(start: number, end: number, step = 1) {
   return array;
 }
 
+/**
+ * Iterate trough integers from start (included) to end (included).
+ */
+export function* iterateIntegers(start: number, end: number): Generator<number> {
```
LucasLefevre, line 218: :+1: 
I would have the same API and behavior as `range` (start, end, step) (which itself replicates the python `range` function).
Maybe even name it `rangeIterator` or something like ?

LucasLefevre, line 218: return type can be iterable
`: Iterable<number>`

src/plugins/core/sheet.ts
```diff
@@ -834,10 +834,10 @@ export class SheetPlugin extends CorePlugin<SheetState> implements SheetState {
   }
 
   private moveCellOnColumnsDeletion(sheet: Sheet, deletedColumn: number) {
-    for (let [index, row] of Object.entries(sheet.rows)) {
-      const rowIndex = parseInt(index, 10);
```
LucasLefevre, line 838: > This saves 20-40ms on a 10.000 rows sheet with 26 columns, when deleting the
first column.

to you have an rough idea of the total time before/after?

LucasLefevre (commented):
nice :+1:

src/plugins/core/sheet.ts
hokolomopo: I don't have the performance snapshot anymore, but it doesn't really save that much time (20-60ms over 500-700ms I think). But hey it's free.

rrahir (commented):
Nice imp!

Just a note, you don't have to change it right now but for future commits, it would be nice to add the trail `Task: {tasknumber}` instead of `Odoo task`to help us out when we merge everything in Odoo ;-)

src/helpers/format.ts
```diff
@@ -188,14 +188,62 @@ const numberRepresentation: Intl.NumberFormat[] = [];
...
+        slicedDecimalDigits.slice(0, i - 1) + (Number(slicedDecimalDigits[i - 1]) + 1);
+    }
+
+    decimalDigits = slicedDecimalDigits;
+  }
+
+  return { integerDigits, decimalDigits: removeTrailingZeroes(decimalDigits) };
+}
+
+function removeTrailingZeroes(decimalDigits: string): string {
```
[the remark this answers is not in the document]
hokolomopo: Changed the arg name to "numberString". I can Put it in `splitNumber ` directly if you want, but I feel like the function is already long/complex enough as is :/

hokolomopo:
Fixed a bug in `splitNumber()`, I didn't handle correctly the case where we round up and the last digit is a "9". Added a test. Unfortunately, it makes the function more complex :/ I'd really like to use simply Number.toFixed(), but it has some strange illogical behavour...

TODO for tomorrow: also check if it's worth changing `range()` to an iterator, to avoid allocating arrays where we don't have to.

src/helpers/misc.ts
LucasLefevre: btw, the `range` implementation could then just become `return [...rangeIterator(start, end, step)]`

hokolomopo: I think I'd rather make the function `range()` return an iterator, and have the dev make the spread manually (`[...range()]` if needed.

My point being that 80% of the time we only use `range()` for a `for` loop, and in this case allocating a whole array is wastefull. And there's no way peoples (me included) will think about using `rangeIterator` instead of `range` in their loops to save memory. But the otehr way around, TS will complain if we try to use an iterator as an array.

src/helpers/format.ts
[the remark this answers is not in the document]
rrahir: could you maybe just add some docstring saying that the input is a string of integers, that would be nice ^^ it looks like nitpick, agreed, but the function expects a really specific input based on the way it's used, which is not a good design so if we keep it outside, let's make sure its payload is clear

src/migrations/data.ts
```diff
@@ -46,7 +46,7 @@ export function load(data?: any, verboseImport?: boolean): WorkbookData {
       }
     }
   }
-  data = JSON.parse(JSON.stringify(data));
+  data = deepCopy(data);
```
rrahir, line 49: not imported in the file :o

rrahir:
@robodoo r+

PR #2369, opened 2023-04-17
[FIX] clipboard: Do not update viewport on `paste`

yunchen4:
## Description:

This PR fixes the problem that when pasting, the viewport will move if the end of pasted area is not in the current viewport. This problem is especially obvious when pasting a full col/row into a cell. After fixing, the viewport will not move, so it will stay at the location before pasting.

Odoo task ID : [3271366](https://www.odoo.com/web#id=3271366&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/tests/plugins/clipboard.test.ts
+++ b/tests/plugins/clipboard.test.ts
@@ -26,6 +26,7 @@ import {
   setCellFormat,
   setSelection,
   setStyle,
+  setViewportOffset,
   setZoneBorders,
   undo,
 } from "../test_helpers/commands_helpers";
@@ -556,6 +557,13 @@ describe("clipboard", () => {
     expect(model.getters.getMerges(sheetId).map(zoneToXc)).toEqual(["B2:C3"]);
   });
 
+  test("pasting from OS will not change the viewport", () => {
+    const model = new Model();
+    const viewport = model.getters.getActiveMainViewport();
+    pasteFromOSClipboard(model, "C60", "a\t1\nb\t2");
+    expect(model.getters.getActiveMainViewport()).toEqual(viewport);
+  });
+
   test("pasting numbers from windows clipboard => interpreted as number", () => {
     const model = new Model();
     pasteFromOSClipboard(model, "C1", "1\r\n2\r\n3");
@@ -598,6 +606,18 @@ describe("clipboard", () => {
     expect(getCellContent(model, "E2")).toBe("a2");
     expect(getCellContent(model, "F1")).toBe("c1");
     expect(getCellContent(model, "F2")).toBe("c2");
+  });
+
+  test("Viewport won't move after pasting", () => {
+    const model = new Model();
+    copy(model, "A1:B2");
+
+    setSelection(model, ["C60:D70"]);
+    setViewportOffset(model, 0, 0);
+    const viewport = model.getters.getActiveMainViewport();
+
+    paste(model, "C60:D70");
+    expect(model.getters.getActiveMainViewport()).toEqual(viewport);
   });
 
   describe("copy/paste a zone in a larger selection will duplicate the zone on the selection as long as it does not exceed it", () => {
--- a/src/helpers/clipboard/clipboard_cells_state.ts
+++ b/src/helpers/clipboard/clipboard_cells_state.ts
@@ -281,7 +281,7 @@ export class ClipboardCellsState extends ClipboardCellsAbstractState {
     if (height > 1 || width > 1 || isCutOperation) {
       const zones = this.pastedZones(target, width, height);
       const newZone = isCutOperation ? zones[0] : union(...zones);
-      this.selection.selectZone({ cell: { col, row }, zone: newZone });
+      this.selection.selectZone({ cell: { col, row }, zone: newZone }, { scrollIntoView: false });
     }
   }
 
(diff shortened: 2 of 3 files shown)


src/selection_stream/selection_stream_processor.ts
```diff
@@ -40,7 +40,7 @@ type StatefulStream<Event, State> = {
  * Allows to select cells in the grid and update the selection
  */
 interface SelectionProcessor {
-  selectZone(anchor: AnchorZone): DispatchResult;
+  selectZone(anchor: AnchorZone, keepViewport?: boolean): DispatchResult;
```
hokolomopo, line 43: We try to avoid random booleans in function interface, because it makes the code harder to understand at first glance. Here I don't even think you have to do this, you can probably use `modifyAnchro` instead of `selectZone`. Look at what I did in #1988

hokolomopo (commented):
Didn't read the tests yet

src/selection_stream/selection_stream_processor.ts
yunchen4: @hokolomopo I am not sure I get what you mean. There are also other places using `selectZone` which don't need to keep viewport (e.g. grid selection using arrow keys), and that's why I add `keepViewport` argument in `selectZone`. `modifyAnchor` is private so I cannot use it externally. Do you mean make `modifyAnchor` public?

hokolomopo: From outside the `SelectionProcessor` probably make `modifyAnchor` public ? I'm not even sure the SelectionProcessor should know about the concept of viewports. I don't think peoples here like API with random booleans @LucasLefevre 

Internally to the `SelectionProcessor`, my bad that's not `modifyAnchor`, but the event.type `AlterZone` which you can use that don't move the viewport. I don't think you need the `keepViewport` boolean in the event.

LucasLefevre: > I don't think peoples here like API with random booleans @LucasLefevre

That's true. Boolean params are a code smell.
- It makes things much harder to read without knowing by heart what every argument is.
- It's also a sign the function is doing two things (violating the single responsibility principle)
- it increases the cyclomatic complexity because there must be an `if` statement somewhere. Which makes the code again less maintainable/readable

A solution is often to just have two different functions, instead of one with a param

LucasLefevre (commented):
I didn't check what would be the best way to fix the issue.
Small comment however:
Managing the viewport changes with the selection is already a mess and several things (modes) are there just for that. The original idea was indeed that the selection processor didn't need to know about viewports (it is not completely true given the previous statement)

Feel free to break everything and refactor if you want

tests/plugins/clipboard.test.ts
```diff
@@ -703,6 +724,19 @@ describe("clipboard", () => {
...
+      cut(model, "A1:B2");
+
+      selectCell(model, "C60");
+      setAnchorCorner(model, "D70");
+      setViewportOffset(model, 0, 0);
+      const viewport = model.getters.getActiveMainViewport();
+
+      paste(model, "C60:D70");
+      expect(model.getters.getActiveMainViewport()).toEqual(viewport);
+    });
```
hokolomopo, line 739: I'm not sure it's useful to repeat the same test with copy and cut and paste singel zone and paste multple zone, it shouldn't really matter. Even copy/paste entier col/row is the same, from the PoV of the clipboard that's just a big zone, there's nothing special here right ?

If you really think it's important, do a test.each(), let's not copy/paste the same code 8 times.

hokolomopo (commented):
Seems good 👍 

A bit strange to have the selectionProcessor have a method name directly referencing the viewport, we'll see if the reviewer have a better name.

tests/plugins/clipboard.test.ts
yunchen4: I've changed the position of this test. You're right, copy or cut doesn't matter for this task 🙂

src/selection_stream/selection_stream_processor.ts
```diff
@@ -131,6 +132,10 @@ export class SelectionStreamProcessorImpl implements SelectionStreamProcessor {
     return this.modifyAnchor(anchor, "overrideSelection", "ZonesSelected");
   }
 
+  selectZoneAndKeepViewport(anchor: AnchorZone): DispatchResult {
```
LucasLefevre, line 135: @rrahir what do you think ?
I think we must admit the selection processor cannot completely be decoupled from viewport (the various modes introduced just for the sake of handling them differently in the viewport already hinted it)

rrahir: Tought I posted an answer, apparently never did, sorry. 
It was kind of admitted and they're not that decoupled seeing as viewport needs to listen to everything already.

I agree that we cannot work around it anymore even though the split made sense from a functional pov at the start. But I don't know how to solve that issue :/

rrahir: the processor could expose its dependency and the methods have a 'scrollIntoView' argument ? that might alleviate the number of mode and types but also mean that the components would take control of what's displayed. That makes sense for the components, but maybe less for the plugins?

rrahir:
@robodoo r+

PR #2371, opened 2023-04-18
[FIX] SelectionInputPlugin: prevent multiple range in 'singleRange' i…

rrahir:
…nput

The SelectionInputPlugin only partially prevent the addition of multiple ranges while in `inputHasSingleRange` mode. It would handle the command `ADD_EMPTY_RANGE` but unfortunately, we can also add multiple ranges by providing a range string composed of several xc joined by a comma.

This commit ensures that we reject such ranges through the command `CHANGE_RANGE` and also at the selectionInput creation.

Task: 3237798

## Description:

Odoo task ID : [3237798](https://www.odoo.com/web#id=3237798&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui/selection_input.ts
+++ b/src/plugins/ui/selection_input.ts
@@ -56,6 +56,11 @@ export class SelectionInputPlugin extends UIPlugin implements StreamCallbacks<Se
     initialRanges: string[],
     private readonly inputHasSingleRange: boolean
   ) {
+    if (inputHasSingleRange && initialRanges.length > 1) {
+      throw new Error(
+        "Input with a single range cannot be instantiated with several range references."
+      );
+    }
     super(getters, state, dispatch, config, selection);
     this.insertNewRange(0, initialRanges);
     this.activeSheet = this.getters.getActiveSheetId();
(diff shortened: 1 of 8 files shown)


src/plugins/ui/selection_input.ts
```diff
@@ -76,6 +81,11 @@ export class SelectionInputPlugin extends UIPlugin implements StreamCallbacks<Se
           return CommandResult.MaximumRangesReached;
         }
         break;
+      case "CHANGE_RANGE":
```
hokolomopo, line 84: From a functional POV, what happens in the side panel looks strange when you paste something like `A1,B2`. You press confirm, and there no error message, but the input is red, and its back at the previous value

hokolomopo (commented):
A functional problem (I think) and the CSS is broken, otherwise small details :)

src/plugins/ui/selection_input.ts
rrahir: ok  I changed the behavior - if you input a set or ranges 'A1,A2,...' it will take the first element and drop the rest of the string

LucasLefevre (approved):
robodoo r+

LucasLefevre:
robodoo rebase-ff

PR #2380, opened 2023-04-20
Add some (a lot) of array functions

hokolomopo:
Add functions:
- UNIQUE
- EXPAND
- FILTER
- TRANSPOSE
- MUNIT
- RANDARRAY
- FLATTEN
- FREQUENCY
- ARRAY.CONSTRAIN
- CHOOSECOLS
- CHOOSEROWS
- SUMPRODUCT
- MINVERSE
- MDETERMs
- MMULT
- SUMX2MY2
- SUMX2PY2
- SUMXMY2
- TOCOL
- TOROWs
- SPLIT
- HSTACK
- VSTACK
- WRAPCOLS
- WRAPROWS

Odoo task ID : [3184778](https://www.odoo.com/web#id=3184778&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


hokolomopo:
TODO : demo data

tests/functions/module_array.test.ts
```diff
@@ -136,6 +137,108 @@ describe("FLATTEN function", () => {
...
+      A3: "3", B3: "3", C3: "5",
+      A4: "4", B4: "4",
+      A5: "5", B5: "6",
+     };
+    const model = getModelFromGrid(grid);
+    setCellContent(model, "D1", "=FREQUENCY(A1:B6, C1:C3)");
+    expect(getRangeValues(model, "D1:D4")).toEqual([4, 2, 3, 1]);
+  });
+
+  test("Non-number values are ignored", () => {
```
LucasLefevre, line 434: see my other comment. What about `="1"` ?

tests/functions/module_array.test.ts
```diff
@@ -8,6 +8,60 @@ import {
...
+
+  test("Constraint array", () => {
+    // prettier-ignore
+    const grid = {
+      A1: "A1", B1: "B1", C1: "C1",
+      A2: "A2", B2: "B2", C2: "C2",
+      A3: "A3", B3: "B2", C3: "C3",
+    };
+    const model = getModelFromGrid(grid);
+    setCellContent(model, "D1", "=ARRAY.CONSTRAIN(A1:C3, 2, 2)");
```
LucasLefevre, line 28: you can add a test with:
- `rows` and `columns` being zero
- `rows` and `columns` being negative
- `rows` and `columns` being non-numbers (strings or boolean)

tests/functions/module_array.test.ts
```diff
@@ -448,6 +448,55 @@ describe("FREQUENCY function", () => {
...
+      "#ERROR"
+    ); // @compatibility: on google sheets, return #VALUE!
+  });
+
+  test("Range values must have the same dimensions", () => {
+    expect(evaluateCell("D1", { D1: "=SUMPRODUCT(A1:A2, A1:B1)" })).toBe("#ERROR"); // @compatibility: on google sheets, return #VALUE!
+    expect(evaluateCell("D1", { D1: "=SUMPRODUCT(A1:A2, A1:A3)" })).toBe("#ERROR"); // @compatibility: on google sheets, return #VALUE!
+  });
+
+  test("SUMPRODUCT with numbers arguments", () => {
```
LucasLefevre, line 707: what happens if number arguments are mixed with ranges?

src/functions/module_array.ts
```diff
@@ -259,6 +262,70 @@ export const FREQUENCY: AddFunctionDescription = {
...
+    arg(
+      "square_matrix (number, range<number>)",
+      _lt(
+        "An range with an equal number of rows and columns representing a matrix whose determinant will be calculated."
+      )
+    ),
+  ],
+  returns: ["NUMBER"],
+  compute: function (matrix: ArgValue): number {
+    if (!Array.isArray(matrix)) {
```
LucasLefevre, line 314: you can use `isMatrix`

src/functions/helper_matrices.ts
```diff
@@ -104,3 +104,32 @@ function swapMatrixRows(matrix: number[][], row1: number, row2: number) {
...
+
+/**
+ * Matrix multiplication of 2 matrices.
+ * ex: matrix1 : n x l, matrix2 : m x n => result : m x l
+ *
+ * Note: we use indexing [col][row] instead of the standard mathematical notation [row][col]
+ */
+export function multiplyMatrices(matrix1: Matrix<number>, matrix2: Matrix<number>): Matrix<number> {
+  if (matrix1.length !== matrix2[0].length) {
+    throw new Error("Cannot multiply matrices : incompatible matrices size.");
```
LucasLefevre, line 116: missing translation

LucasLefevre (commented):
Already submitting my comments for the first functions.
I'll let @anhe-odoo and @laa-odoo review as well :)

src/functions/module_array.ts
```diff
@@ -83,6 +90,59 @@ export const FLATTEN: AddFunctionDescription = {
...
+// -----------------------------------------------------------------------------
+export const FREQUENCY: AddFunctionDescription = {
+  description: _lt("Calculates the frequency distribution of a range."),
+  args: [
+    arg("data (range<any>)", _lt("The array of ranges containing the values to be counted.")),
+    arg("classes (number, range<any>)", _lt("The range containing the set of classes.")),
+  ],
+  returns: ["RANGE<NUMBER>"],
+  compute: function (data: MatrixArgValue, classes: MatrixArgValue): CellValue[][] {
+    const _data = flattenRowFirst([data], (val) => val).filter(isCellValueANumber);
```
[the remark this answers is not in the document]
hokolomopo: I intentionally left out ="12" strings to have the same behaviour as GSheet IIRC, but didn't take into account that we already have a different behaviour as GSheet for theses strings. 

I can use `isNumber` instead if you think it's better.

src/functions/module_math.ts
```diff
@@ -992,6 +992,35 @@ export const RAND: AddFunctionDescription = {
   isExported: true,
 };
 
+// -----------------------------------------------------------------------------
+// RANDARRAY
+// -----------------------------------------------------------------------------
+export const RANDARRAY: AddFunctionDescription = {
```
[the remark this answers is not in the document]
hokolomopo: I kinda copy/paster the one that Laexis implmeneted for the tests multi-cells withotu asking questions, I'll add the rest of the arguments :)

tests/functions/module_filter.test.ts
```diff
@@ -0,0 +1,71 @@
...
+    expect(getRangeValuesAsMatrix(model, "D1:E2")).toEqual([
+      ["hey", ""],
+      ["olà", ""],
+    ]);
+  });
+
+  test("UNIQUE function with only_once argument to true", () => {
+    const grid = { A1: "hey", A2: "hey", A3: "hey", B1: "olà", B2: "olà", B3: "bjr" };
+    const model = getModelFromGrid(grid);
+    setCellContent(model, "D1", "=UNIQUE(A1:B3, false, true)");
```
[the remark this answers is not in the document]
hokolomopo: Works too. A bit shorter, but maybe slightly harder to read ?

src/functions/helpers.ts
```diff
@@ -710,3 +710,25 @@ function toCellValue(value: any): CellValue {
...
   return Array.isArray(x) && Array.isArray(x[0]);
 }
+
+/**
+ * Flatten an array of items, where each item can be a single value or a 2D array, and apply the
+ * callback to each element.
+ *
+ * The 2D array are flattened row first.
+ */
+export function flattenRowFirst<T, M>(items: Array<T | Matrix<T>>, callback: (val: T) => M): M[] {
```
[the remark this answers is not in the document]
hokolomopo: Nope this doesn't work, since the callback is typed `(val: T) => M`. I could do `(val) => val as M` I guess but I'd rather not :p 
This is never used without a callback ATM anyway.

tests/functions/module_filter.test.ts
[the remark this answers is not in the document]
LucasLefevre: @laa-odoo what do you think?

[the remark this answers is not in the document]
hokolomopo: image
VS
image

tests/functions/module_array.test.ts
hokolomopo: Handled it and added a test

tests/functions/module_array.test.ts
hokolomopo: Was supposed to fail like in the test `Range values must have the same dimensions` , but the helper `assertSameDimensions` had an error. Fixed it and added a test.

tests/functions/module_array.test.ts
hokolomopo: Yup I was missing some asserts in the functions

src/functions/module_math.ts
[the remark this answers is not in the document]
hokolomopo: Or maybe I copied the `RANDARRAY` from GSheet ? It takes only 2 arguments. I'll implement the rest anyway :)

tests/functions/module_filter.test.ts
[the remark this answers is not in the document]
LucasLefevre: the current version might be slightly easier to read "step by step" indeed.
And having the `model` makes it mode flexible as well
I'm ok either way :)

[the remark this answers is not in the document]
laa-odoo: ok with all the previous comment,

however I wonder if we should not have an "evaluateRange" function which in addition to checking the value, checks if the matrix spreads well in the desired area and does not exceed ?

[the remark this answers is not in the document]
hokolomopo: Yeah I was wondering about that too when I was coding, and decided not to do it. I can add something like 

```js
expect(noValuesOutsideOfRange("D1:F3")).toBe(true);
```

at the end of every test if you think it's worth it.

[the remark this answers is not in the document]
laa-odoo: seems like a good idea :)

src/functions/module_array.ts
```diff
@@ -0,0 +1,826 @@
...
+  ],
+  returns: ["NUMBER"],
+  compute: function (arrayX: ArgValue, arrayY: ArgValue): number {
+    assertSameDimensions(
+      "The arguments array_x and array_y must have the same dimensions.",
+      arrayX,
+      arrayY
+    );
+
+    const _array = flattenRowFirst([arrayX, arrayY], toNumber);
```
anhe-odoo, line 530: It could be easier to read if we keep the same way to write it as in SUMX2MY2 and SUMXMY2 (flatten each separately and then a for loop), just to keep the same pattern ? Or it may be less effective ?

anhe-odoo (commented):
Very good job, it's really nice to have all these arrays formulas already implemented :)

Some comments (and some questions), I'll post some more this evening or tomorrow :)

src/functions/module_array.ts
```diff
@@ -0,0 +1,826 @@
...
+    arg(
+      "square_matrix (number, range<number>)",
+      _lt(
+        "An range with an equal number of rows and columns representing a matrix whose multiplicative inverse will be calculated."
+      )
+    ),
+  ],
+  returns: ["RANGE<NUMBER>"],
+  compute: function (matrix: ArgValue): CellValue[][] {
+    const _matrix: Matrix<number> = matrixMap(toMatrixArgValue(matrix), toNumber);
```
anhe-odoo, line 367: When do you need the `toMatrixArgValue` ? Is it only when the user enter a number as argument directly, or also when targeting a range with only a cell (I think in this last case we still have a matrix but not sure) ?

src/functions/helper_assert.ts
```diff
@@ -0,0 +1,35 @@
...
+}
+
+export function assertArraySameDimensions(errorStr: string, ...args: any[][]) {
+  const dims = args[0].length;
+  for (const arg of args) {
+    assert(() => arg.length === dims, errorStr);
+  }
+}
+
+export function assertPositive(errorStr: string, arg: number) {
```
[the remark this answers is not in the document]
hokolomopo: https://github.com/odoo/o-spreadsheet/pull/2380#discussion_r1219190012

tests/functions/module_array.test.ts
```diff
@@ -0,0 +1,1130 @@
...
+    expect(evaluateCell("A1", { A1: "=MDETERM(1, 0)" })).toBe("#BAD_EXPR"); // @compatibility: on google sheets, return #N/A
+  });
+
+  test("Argument must be a square matrix", () => {
+    const grid = { A1: "1", B1: "0", A2: "0", B2: "1" };
+    expect(evaluateCell("D1", { D1: "=MDETERM(A1:B1)", ...grid })).toBe("#ERROR"); // @compatibility: on google sheets, return #VALUE!
+    expect(evaluateCell("D1", { D1: "=MDETERM(A1:A2)", ...grid })).toBe("#ERROR"); // @compatibility: on google sheets, return #VALUE!
+  });
+
+  test("Argument must only contain number, or values convertibale to number", () => {
```
anhe-odoo, line 498: Convertibale ? :D

tests/functions/module_array.test.ts
```diff
@@ -0,0 +1,1130 @@
...
+    expect(evaluateCell("D1", { D1: "=MINVERSE(A1:A2)", ...grid })).toBe("#ERROR"); // @compatibility: on google sheets, return #VALUE!
+  });
+
+  test("Argument must be an invertible matrix", () => {
+    const grid = { A1: "1", B1: "1", A2: "1", B2: "1" };
+    expect(evaluateCell("D1", { D1: "=MINVERSE(0)" })).toBe("#ERROR"); // @compatibility: on google sheets, return #NUM!
+    expect(evaluateCell("D1", { D1: "=MINVERSE(A1:B2)", ...grid })).toBe("#ERROR"); // @compatibility: on google sheets, return #NUM!
+  });
+
+  test("Argument must only contain number, or values convertibale to number", () => {
```
anhe-odoo, line 555: Convertibale :D

tests/test_helpers/helpers.ts
```diff
@@ -331,6 +332,62 @@ export function evaluateCell(xc: string, grid: GridDescr): any {
   return gridResult[xc];
 }
 
+/**
+ * Evaluate the final state of a all the cells in the given range xc according th the different values
```
anhe-odoo, line 336: "of a all"

tests/functions/module_array.test.ts
hokolomopo: :eyes:

src/functions/module_array.ts
```diff
@@ -0,0 +1,826 @@
...
+      scans the array by row."
+    )
+  ),
+];
+
+export const TOCOL: AddFunctionDescription = {
+  description: _lt("Transforms a range of cells into a single column."),
+  args: TO_COL_ROW_ARGS,
+  returns: ["RANGE<ANY>"],
+  //TODO compute format
```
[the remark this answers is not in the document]
hokolomopo: Yup

src/functions/module_array.ts
hokolomopo: The code would be a bit longer, but performances should ba about the same. But it's probably a good idea, `reduce` are always a bit hard to read.

src/types/cells.ts
```diff
@@ -40,6 +40,7 @@ interface EvaluatedCellProperties {
 }
 
 export type CellValue = string | number | boolean;
+export type OptionalCellValue = CellValue | undefined;
```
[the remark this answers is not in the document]
laa-odoo: More generally, the type is used only very few times. I think we can do without it.

And cases ```OptionalCellValue[][]``` correspond to ```MatrixArgValue```

src/functions/module_array.ts
```diff
@@ -0,0 +1,826 @@
...
+  arg(
+    `scan_by_column (number, default=${TO_COL_ROW_DEFAULT_SCAN})`,
+    _lt(
+      "Whether the array should be scanned by column. True scans the array by column and false (default) \
+      scans the array by row."
+    )
+  ),
+];
+
+export const TOCOL: AddFunctionDescription = {
```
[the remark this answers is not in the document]
hokolomopo: Not a bad ida, but the typing becomes very anoying, we'd need conditional types and whatnot, so I don't think it's worth?

tests/functions/module_filter.test.ts
```diff
@@ -0,0 +1,71 @@
+import { setCellContent } from "../test_helpers/commands_helpers";
+import { getCellContent, getCellError } from "../test_helpers/getters_helpers";
+import { evaluateCell, getModelFromGrid, getRangeValuesAsMatrix } from "../test_helpers/helpers";
+
```
laa-odoo, line 73: will be good to test if UNIQUE make not difference between a raw value and a value inferred from formula
```
const grid = { A1: "3", A2: "=1+2", B1:"kikou", B2:"=CONCAT("ki","kou")}
setCellContent(model, "A1", "=UNIQUE(A1:B2)");
    expect(getRangeValuesAsMatrix(model, "D1:E3")).toEqual([
      ["3", "kikou"],
      ["", ""],
    ]);

expect(getRangeValuesAsMatrix(model, "A1:A2")).toEqual([["hey"], [""]]);
```

src/functions/helper_matrices.ts
```diff
@@ -0,0 +1,136 @@
...
+  return matrix;
+}
+
+/**
+ * Invert a matrix and compute its determinant using Gaussian Elimination.
+ *
+ * The Matrix should be a square matrix, and should be indexed [col][row] instead of the
+ * standard mathematical indexing [row][col].
+ */
+export function invertMatrix(M: Matrix<number>): {
```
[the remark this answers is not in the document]
hokolomopo: Probably not worth to duplicate a lot of code + comments just to save some computations

src/types/functions.ts
```diff
@@ -11,6 +11,7 @@ export type ArgType =
   | "RANGE<NUMBER>"
   | "RANGE<DATE>"
   | "RANGE<STRING>"
+  | "RANGE<ANY>"
```
laa-odoo, line 14: "RANGE" do the same thing

LucasLefevre:
robodoo delegate=laa-odoo

src/types/functions.ts
LucasLefevre: doesn't hurt to have an explicit alternative IMHO

src/functions/module_array.ts
```diff
@@ -0,0 +1,60 @@
...
+    arg(
+      "columns (boolean, optional)",
+      _lt("The number of columns in the expanded array. If missing, columns will not be expanded.")
+    ),
+    arg("pad_with (any, default=0)", _lt("The value with which to pad.")), // @compatibility: on Excel, pad with #N/A
+  ],
+  returns: ["RANGE<ANY>"],
+  //TODO computeFormat
+  compute: function (
+    array: MatrixArgValue,
```
laa-odoo, line 178:
```
    array: MatrixArgValue,
    rows?: PrimitiveArgValue ,
    columns?: PrimitiveArgValue,
    padWith: PrimitiveArgValue = 0,
    ```
    
    Allow to ignore```_padwith```  ?

File: src/functions/module_array.ts [hunk 23]
```diff
@@ -0,0 +1,60 @@
...
+    rows: PrimitiveArgValue | undefined,
+    columns: PrimitiveArgValue | undefined,
+    padWith: PrimitiveArgValue | undefined
+  ): CellValue[][] {
```
laa-odoo, line 182: Matrix<CellValue>

src/functions/module_array.ts
```diff
@@ -0,0 +1,60 @@
...
+    );
+
+    const result: MatrixArgValue = [];
+    for (let col = 0; col < _columns; col++) {
+      result[col] = [];
+      for (let row = 0; row < _rows; row++) {
+        result[col][row] = array[col]?.[row] || _padWith;
+      }
+    }
+    return toCellValueMatrix(result);
```
laa-odoo, line 209: I think that's overkill.

If we reconstruct an array, in the previous steps by reassigning all the values, then we can handle the undefined case there.

Here we go through the matrix a second time for nothing

laa-odoo: (note that here ```àrray``` contains only values having the type CellValue or undefined, not null)

src/functions/module_array.ts
```diff
@@ -0,0 +1,60 @@
...
+
+// -----------------------------------------------------------------------------
+// EXPAND
+// -----------------------------------------------------------------------------
+export const EXPAND: AddFunctionDescription = {
+  description: _lt("Expands or pads an array to specified row and column dimensions."),
+  args: [
+    arg("array (range<any>)", _lt("The array to expand.")),
+    arg(
+      "rows (number, optional)",
```
laa-odoo, line 166: non optional

src/functions/module_filter.ts
```diff
@@ -1,8 +1,69 @@
...
+      _lt("The arguments conditions must have the same dimensions."),
+      ...conditions
+    );
+    const _conditions = conditions.map((c) => c.flat());
+
+    const mode = conditions[0].length === 1 ? "row" : "col";
+    range = mode === "row" ? transpose2dArray(range) : range;
+
+    assertArraySameDimensions(
+      _lt("The range and conditions must have the same dimensions."),
```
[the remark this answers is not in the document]
laa-odoo: And therefore the assert does not have the correct name 

maybe it could be interesting, instead of using ```assertSameDimensions``` and ```assertArrayDimension```

--> use something like ```assertSameColDimensions``` and ```assertSameRowDimensions```  ?
we could pass in the arguments of these functions the range and the conditions at the same time

src/functions/module_math.ts
```diff
@@ -996,6 +997,56 @@ export const RAND: AddFunctionDescription = {
...
+        _min.toString()
+      )
+    );
+
+    const result: number[][] = Array(_cols);
+    for (let col = 0; col < _cols; col++) {
+      result[col] = Array(_rows);
+      for (let row = 0; row < _rows; row++) {
+        const random = Math.random() * (_max - _min) + _min;
+        result[col][row] = _whole_number ? Math.round(random) : random;
```
laa-odoo, line 1042: the distribution is no longer uniform when applying the round directly in this way :)

src/functions/module_array.ts
```diff
@@ -66,6 +66,22 @@ export const EXPAND: AddFunctionDescription = {
...
 };
 
+// -----------------------------------------------------------------------------
+// FLATTEN
+// -----------------------------------------------------------------------------
+export const FLATTEN: AddFunctionDescription = {
+  description: _lt("Flattens all the values from one or more ranges into a single column."),
+  args: [
+    arg("range (range<any>)", _lt("The first range to flatten.")),
+    arg("range2 (range<any>, repeating)", _lt("Additional ranges to flatten.")),
```
laa-odoo, line 221: range and range2 could be simple value: ```any, range<any>```

maybe add test with simple values
the code will probably need to be modified to consider this case

src/functions/module_filter.ts
```diff
@@ -1,8 +1,69 @@
...
+
+// -----------------------------------------------------------------------------
+// FILTER
+// -----------------------------------------------------------------------------
+export const FILTER: AddFunctionDescription = {
+  description: _lt(
+    "Returns a filtered version of the source range, returning only rows or columns that meet the specified conditions."
+  ),
+  args: [
+    arg("range (range<any>)", _lt("The data to be filtered.")),
```
laa-odoo, line 21: I think could be any on Google

src/functions/module_array.ts
```diff
@@ -58,3 +65,22 @@ export const EXPAND: AddFunctionDescription = {
   },
   isExported: true,
 };
+
+// -----------------------------------------------------------------------------
+// TRANSPOSE
+// -----------------------------------------------------------------------------
+export const TRANSPOSE: AddFunctionDescription = {
+  description: _lt("Transposes the rows and columns of a range."),
+  args: [arg("range (range<any>)", _lt("The range to be transposed."))],
```
laa-odoo, line 680: I think could be any on Google

src/functions/module_array.ts
```diff
@@ -82,6 +82,62 @@ export const FLATTEN: AddFunctionDescription = {
...
+      .map(toNumber);
+    const _classes = flattenRowFirst([classes], (val) => val)
+      .filter((val) => typeof val === "number" || (typeof val === "string" && isNumber(val)))
+      .map(toNumber);
+    /**
+     * Returns the frequency distribution of the data in the classes, ie. the number of elements in the range
+     * between each classes.
+     *
+     * For example:
+     * - data = [1, 2, 3, 4, 5]
```
laa-odoo, line 252: In the example, you can use unsorted value for data
This will be more intuitive to understand that FREQUENCY don't need data in a sorted order to work correctly

Maybe a test for that could be a good idea :)

src/functions/module_array.ts
```diff
@@ -82,6 +82,62 @@ export const FLATTEN: AddFunctionDescription = {
   isExported: false,
 };
 
+// -----------------------------------------------------------------------------
+// FREQUENCY
+// -----------------------------------------------------------------------------
+export const FREQUENCY: AddFunctionDescription = {
```
laa-odoo, line 233: I wonder about this formula: Excel and google have different behavior.

On google, the results are always given in ascending order of classes.

while excel respects the order of the given classes

We could believe that it is practical at excel because you can give your classes according to a descending order.

But I find it weird, because in this case the last value of the result should also be placed at the beginning, which is not the case...

So, I wonder if we should not do like google for this formula --> always give the result in ascending order of classes

hokolomopo: In cases like this where the behaviour isn't the same on GSheet and Exel, I like to keep the excel bahavour sicne we support direct conversions o_spreadhseet <=> excel. 

Functionnaly I don't really care about what we do, but note that we can get Google behaviour by sorting our classes, but if we implement Google behaviour, we cannot have Excel behaviour.

laa-odoo: okay, so maybe add a comment in test to talk about compatibility with Google

src/functions/module_array.ts
```diff
@@ -97,6 +97,46 @@ export const CHOOSECOLS: AddFunctionDescription = {
...
+    arg("array (range<any>)", _lt("The array that contains the rows to be returned.")),
+    arg("row_num (number, range<number>)", _lt("The first row index of the rows to be returned.")),
+    arg(
+      "row_num2 (number, range<number>, repeating)",
+      _lt("The rows indexes of the rows to be returned.")
+    ),
+  ],
+  returns: ["RANGE<ANY>"],
+  //TODO computeFormat
+  compute: function (array: MatrixArgValue, ...rows: ArgValue[]): CellValue[][] {
```
laa-odoo, line 133: array: ArgValue

src/functions/module_array.ts
```diff
@@ -97,6 +97,46 @@ export const CHOOSECOLS: AddFunctionDescription = {
...
+        (_rows.find((row) => row <= 0 || row > array[0].length) || 0).toString()
+      )
+    );
+
+    const result: CellValue[][] = Array(array.length);
+    for (let col = 0; col < array.length; col++) {
+      result[col] = Array(_rows.length);
+      for (let row = 0; row < _rows.length; row++) {
+        const rowIndex = _rows[row] - 1; // -1 because rows arguments are 1-indexed
+        result[col][row] = toCellValue(array[col][rowIndex]);
```
laa-odoo, line 149: toCellValue(array[col][rowIndex]) --> array[col][rowIndex] || 0

laa-odoo (changes requested):

laa-odoo:
I'm not done the review, I'm doing it little by little. Here especially comments on the type

src/functions/helpers.ts
```diff
@@ -715,6 +715,17 @@ function compareCellValues(left: CellValue | undefined, right: CellValue | undef
   return typeOrder;
 }
 
+export function matrixMap<T, M>(matrix: Matrix<T>, fn: (value: T) => M): Matrix<M> {
```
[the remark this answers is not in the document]
laa-odoo: my bad, Same structure but you need to create new array inside.

Maybe prefer create a mapArgs function that could take into account, simple value too. 
As we are in helper function, we could use directly ArgValue/PrimitiveArgValue/MatrixArgValue to type the inputs.
Like the visitArgs and the reduceArgs helper, this helper will use two callback to work

src/functions/module_array.ts
```diff
@@ -269,6 +277,48 @@ export const FREQUENCY: AddFunctionDescription = {
...
+  returns: ["NUMBER"],
+  compute: function (...args: ArgValue[]): number {
+    assertSameDimensions(_lt("All the ranges must have the same dimensions."), ...args);
+    if (args.every(isMatrix)) {
+      let result = 0;
+      for (let i = 0; i < args[0].length; i++) {
+        for (let j = 0; j < args[0][i].length; j++) {
+          let product = 1;
+          for (const range of args) {
+            product *= toNumber(range[i][j]);
```
laa-odoo, line 451: Let's keep functional consistency:
numbers in string form should not be taken into account when provided by reference

maybe use visitNumbers (may need to change it) to keep consistency

laa-odoo (changes requested):

tests/functions/module_text.test.ts
```diff
@@ -490,6 +492,62 @@ describe("SEARCH formula", () => {
...
+  });
+
+  test("delimiter argument should not be empty", () => {
+    expect(evaluateCell("A1", { A1: '=SPLIT("Hello", "", 1, 1)' })).toBe("#ERROR"); // @compatibility: on google sheets, return #VALUE!
+    expect(evaluateCell("A1", { A1: '=SPLIT("Hello", B1, 1, 1)', B1: '=""' })).toBe("#ERROR"); // @compatibility: on google sheets, return #VALUE!
+  });
+
+  test("Simple split", () => {
+    const model = new Model();
+    setCellContent(model, "A1", "Hello there, General Kenobi");
```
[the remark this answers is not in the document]
hokolomopo: image

laa-odoo (approved):
Well done ! It will be a good functional contribution :)

laa-odoo:
@robodoo rebase-ff

laa-odoo:
@robodoo r+

LucasLefevre:
robodoo retry

PR #2401, opened 2023-04-25
[FIX] lookup: horizontal search in LOOKUP function

dhrp-odoo:
## Description:

Previously, the `=LOOKUP` function did not work properly when performing a
horizontal search, as it would always pass `searchArray[0][index]` to the
`assertAvailable` function parameter.

To resolve this issue, we added a check to determine whether it is a horizontal
or vertical search, and passed the appropriate parameter to the `assertAvailable`
function accordingly.

Odoo task ID : [3277145](https://www.odoo.com/web#id=3277145&menu_id=4720&cids=2&action=333&active_id=2328&model=project.task&view_type=form)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/functions/module_lookup.ts
+++ b/src/functions/module_lookup.ts
@@ -170,7 +170,12 @@ export const LOOKUP: AddFunctionDescription = {
       rangeLength,
       getElement
     );
-    assertAvailable(searchArray[0][index], searchKey);
+
+    if (index === -1) assertAvailable(undefined, searchKey);
+
+    verticalSearch
+      ? assertAvailable(searchArray[0][index], searchKey)
+      : assertAvailable(searchArray[index][nbRow - 1], searchKey);
 
     if (resultRange === undefined) {
       return (
--- a/tests/functions/module_lookup.test.ts
+++ b/tests/functions/module_lookup.test.ts
@@ -102,6 +102,40 @@ describe("LOOKUP formula", () => {
   test("Accents and uppercase are ignored", () => {
     expect(evaluateCell("A1", { A1: '=LOOKUP("epee", B1)', B1: "Épée" })).toBe("Épée");
   });
+
+  test("Horizontal search in LOOKUP function with and without result range", () => {
+    const grid = {
+      A1: "1",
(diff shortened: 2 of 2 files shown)


src/functions/module_lookup.ts
```diff
@@ -170,7 +170,12 @@ export const LOOKUP: AddFunctionDescription = {
       rangeLength,
       getElement
     );
-    assertAvailable(searchArray[0][index], searchKey);
+
+    if (resultRange === undefined && !verticalSearch && index !== -1) {
```
LucasLefevre, line 174: mixin the `index === -1` case with this case makes it not particularly easy to read because it's technically not linked.
I would handle this case explicitly before. Something like this
```ts
if (index === -1) {
    assertAvailable(undefined, searchKey);
}
```

LucasLefevre, line 174: Why `resultRange === undefined` ?
No matter if `resultRange` is defined or not, the search key must be found in the search array.
And when removing it all tests pass.

All in all, I think only `if (!verticalSearch) {` remains. Then both cases can be switched to avoid the negation in the if predicate. Negations are harder to read ;)

tests/functions/module_lookup.test.ts
```diff
@@ -102,6 +102,18 @@ describe("LOOKUP formula", () => {
   test("Accents and uppercase are ignored", () => {
     expect(evaluateCell("A1", { A1: '=LOOKUP("epee", B1)', B1: "Épée" })).toBe("Épée");
   });
+
+  test("Horizontal search in LOOKUP function with single-cell range", () => {
```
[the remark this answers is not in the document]
LucasLefevre: could you add a test with a similar test with a `result_range` ? (If there's not already)

dhrp-odoo:
Hi @LucasLefevre, I have made the suggested changes and also added test cases. Thanks :)

LucasLefevre (commented):
Thanks :)
robodoo r+

PR #2411, opened 2023-04-26
[FIX] collaborative: ignore bad snapshot revision

LucasLefevre:
## Description:

In odoo:
First steps:
- open an icognito window (let's call if browser A)
- login to odoo
- open a new blank tab
- close odoo tab (you should have no odoo tab open in that browser)

In a normal browser (not incognito)
- open a spreadsheet
- force a snapshot (starting with v16, you can use File > Clear History)
- copy the spreadsheet url and open it in the browser A => crash

What happens is: when loading the spreadsheet in browser A, the bus recieves the SNAPSHOT_CREATED revision. But its "serverRevisionId" doesn't match the current spreadsheet revision id! The current revision id was updated to the next id when snapshotting!

With this commit, the session is now more robust to invalid "serverRevisionId". Up until now, it was only robust when receiving "REMOTE_REVISION" messages. Now it's extended to snapshot and undo/redo messages.

Note: I removed the `revisionId` from the triggered event payload because it's actually never used and it was in my way since `message.serverRevisionId` is not defined in every message type, TS would yell at me.

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


rrahir:
@robodoo r+

PR #2441, opened 2023-05-03
[IMP] topbar: add shortcuts for opening menus

yunchen4:
## Description:

This PR implements the shortcuts to open topbar menus. It also implements the nagivation via keyboard in the menus.

The basic idea is to use an external listener to monitor keydown events in the topbar, and use props to do the bi-directional communication between topbar and menu component.

Odoo task ID : [3281058](https://www.odoo.com/web#id=3281058&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/grid/grid.ts
+++ b/src/components/grid/grid.ts
@@ -203,8 +203,6 @@ export class Grid extends Component<Props, SpreadsheetChildEnv> {
       /** TODO: Clean once we introduce proper focus on sub components. Grid should not have to handle all this logic */
       if (this.env.model.getters.hasOpenedPopover()) {
         this.closeOpenedPopover();
-      } else if (this.menuState.isOpen) {
-        this.closeMenu();
       } else if (this.env.model.getters.isPaintingFormat()) {
         this.env.model.dispatch("CANCEL_PAINT_FORMAT");
       } else {
--- a/src/components/menu/menu.ts
+++ b/src/components/menu/menu.ts
@@ -1,4 +1,13 @@
-import { Component, onWillUpdateProps, useExternalListener, useRef, useState } from "@odoo/owl";
+import {
+  Component,
+  onMounted,
+  onPatched,
+  onWillUpdateProps,
+  useEffect,
+  useExternalListener,
+  useRef,
+  useState,
+} from "@odoo/owl";
 import { Action } from "../../actions/action";
 import {
   BG_HOVER_COLOR,
@@ -10,7 +19,7 @@ import {
   MENU_VERTICAL_PADDING,
   MENU_WIDTH,
 } from "../../constants";
-import { DOMCoordinates, MenuMouseEvent, Pixel, SpreadsheetChildEnv, UID } from "../../types";
+import { DOMCoordinates, MenuMouseEvent, Pixel, Ref, SpreadsheetChildEnv, UID } from "../../types";
 import { css } from "../helpers/css";
 import { getOpenedMenus, isChildEvent } from "../helpers/dom_helpers";
 import { useAbsoluteBoundingRect } from "../helpers/position_hook";
@@ -26,6 +35,10 @@ css/* scss */ `
     padding: ${MENU_VERTICAL_PADDING}px 0px;
     width: ${MENU_WIDTH}px;
     box-sizing: border-box !important;
+
+    &:focus {
+      outline: none;
+    }
 
     .o-menu-item {
       display: flex;
@@ -53,7 +66,6 @@ css/* scss */ `
       }
 
       &:not(.disabled) {
-        &:hover,
         &.o-menu-item-active {
           background-color: ${BG_HOVER_COLOR};
         }
@@ -84,6 +96,8 @@ interface Props {
   onClose: () => void;
   onMenuClicked?: (ev: CustomEvent) => void;
   menuId?: UID;
+  navigateBackToParent?: (direction: "left" | "right") => void;
+  shouldSelectFirstItemWhenFirstOpened?: boolean;
 }
 
 export interface MenuState {
@@ -92,6 +106,7 @@ export interface MenuState {
   position: null | DOMCoordinates;
   scrollOffset?: Pixel;
   menuItems: Action[];
+  shouldSelectFirstItemWhenFirstOpened?: boolean;
 }
 export class Menu extends Component<Props, SpreadsheetChildEnv> {
   static template = "o-spreadsheet-Menu";
@@ -100,23 +115,69 @@ export class Menu extends Component<Props, SpreadsheetChildEnv> {
   static defaultProps = {
     depth: 1,
   };
+  private clickableMenuItems: Action[] = [];
   private subMenu: MenuState = useState({
     isOpen: false,
     position: null,
     scrollOffset: 0,
     menuItems: [],
+    shouldSelectFirstItemWhenFirstOpened: false,
+  });
+  private menuItemSelection = useState({
+    selectedIndex: this.props.shouldSelectFirstItemWhenFirstOpened ? 0 : undefined,
   });
   private menuRef = useRef("menu");
   private position: DOMCoordinates = useAbsoluteBoundingRect(this.menuRef);
+  private menuItemRefs: { [id: string]: { action: Action; ref: Ref<HTMLElement> } } = {};
+  childrenHaveIcon: boolean = false;
 
   setup() {
+    for (const menuItem of this.props.menuItems) {
+      this.menuItemRefs[menuItem.id] = {
+        action: menuItem,
+        ref: useRef(menuItem.id),
+      };
+    }
+    this.childrenHaveIcon = this.props.menuItems.some(
+      (menuItem) => !!menuItem.icon || !!menuItem.isActive
+    );
+    if (
+      this.menuItemSelection.selectedIndex === undefined &&
+      this.props.shouldSelectFirstItemWhenFirstOpened
+    ) {
+      this.menuItemSelection.selectedIndex = 0;
+    }
     useExternalListener(window, "click", this.onExternalClick, { capture: true });
     useExternalListener(window, "contextmenu", this.onExternalClick, { capture: true });
+    onMounted(() => this.focus());
     onWillUpdateProps((nextProps: Props) => {
       if (nextProps.menuItems !== this.props.menuItems) {
         this.closeSubMenu();
       }
     });
+    onPatched(() => this.focus());
+    useEffect(
+      () => {
+        this.clickableMenuItems = this.props.menuItems.filter(
+          (menuItem) => menuItem.isEnabled(this.env) && menuItem.isVisible(this.env)
+        );
+        this.closeSubMenu();
+        this.menuItemSelection.selectedIndex = undefined;
+        if (
+          this.menuItemSelection.selectedIndex === undefined &&
+          this.clickableMenuItems.length > 0 &&
+          this.props.shouldSelectFirstItemWhenFirstOpened
+        ) {
+          this.menuItemSelection.selectedIndex = 0;
+        }
+      },
+      () => [this.menuItemsAndSeparators.length]
+    );
+  }
+
+  focus() {
+    if (this.subMenu.isOpen) return;
+    this.menuRef.el?.focus();
   }
 
   get menuItemsAndSeparators(): MenuItemOrSeparator[] {
@@ -165,8 +226,11 @@ export class Menu extends Component<Props, SpreadsheetChildEnv> {
     };
   }
 
-  get childrenHaveIcon(): boolean {
-    return this.props.menuItems.some((menuItem) => !!menuItem.icon || !!menuItem.isActive);
+  isMenuItemActive(menuItem: Action): boolean {
+    if (this.menuItemSelection.selectedIndex === undefined) {
+      return this.isParentMenu(this.subMenu, menuItem);
+    }
+    return menuItem.id === this.clickableMenuItems[this.menuItemSelection.selectedIndex]?.id;
   }
 
   getIconName(menu: Action) {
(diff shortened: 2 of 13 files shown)


src/components/menu/menu.xml
```diff
@@ -7,19 +7,22 @@
...
             <t t-set="isMenuRoot" t-value="isRoot(menuItem)"/>
             <t t-set="isMenuEnabled" t-value="isEnabled(menuItem)"/>
             <div
+              t-ref="{{menuItem.id}}"
               t-att-title="getName(menuItem)"
               t-att-data-name="menuItem.id"
-              t-on-click="(ev) => this.onClickMenu(menuItem, menuItem_index, ev)"
-              t-on-mouseover="(ev) => this.onMouseOver(menuItem, menuItem_index, ev)"
+              t-on-click="(ev) => this.onClickMenu(menuItem, ev)"
+              t-on-mousemove="(ev) => this.onMouseMove(menuItem, ev)"
```
yunchen4, line 23: The reason to use `mousemove` here is using `mouseover` will cause bug when I select something (e.g. insert all formulas) using mouse first, and then use ArrowUp and ArrowDown to navigate the menu. I'm not sure whether I need to debounce `mousemove`.

src/components/menu/menu.ts
```diff
@@ -233,18 +277,43 @@ export class Menu extends Component<Props, SpreadsheetChildEnv> {
    * If the given menu is not disabled, open it's submenu at the
    * correct position according to available surrounding space.
    */
-  openSubMenu(menu: Action, menuIndex: number, ev: MouseEvent) {
-    const parentMenuEl = ev.currentTarget as HTMLElement;
-    if (!parentMenuEl) return;
+  openSubMenu(parentMenuEl: HTMLElement, parentMenu: Action) {
+    if (!parentMenuEl || !parentMenu) return;
```
rrahir, line 281: this conflicts with the typing

src/components/menu/menu.ts
```diff
@@ -233,18 +277,43 @@ export class Menu extends Component<Props, SpreadsheetChildEnv> {
...
+      this.backToParentRight();
+      return;
+    }
+    const parentMenu = this.clickableMenuItems[this.subMenu.menuIndex];
+    if (parentMenu.children(this.env).length === 0) {
+      this.backToParentRight();
+      return;
+    }
+    const { ref } = this.menuItemRefs[this.subMenu.menuIndex];
+    const parentMenuEl = ref.el as HTMLElement;
```
rrahir, line 314: the `el` is actually undefined once the component is unmounted, it would make sense to early return in that case

hokolomopo (commented):
Didn't read much, 2-3 simple comments. Added some bugs in the pad in odoo.

The feature is really really nice tho :heart_eyes:

src/components/menu/menu.ts
```diff
@@ -256,25 +325,112 @@ export class Menu extends Component<Props, SpreadsheetChildEnv> {
...
+    } else if (this.subMenu.menuIndex === this.clickableMenuItems.length - 1) {
+      this.subMenu.menuIndex = 0;
+    } else {
+      this.subMenu.menuIndex++;
+    }
+    const el = this.menuItemRefs[this.subMenu.menuIndex]?.ref.el;
+    el?.scrollIntoView(false);
+  }
+
+  private selectLastMenuItem() {
```
rrahir, line 418:
```suggestion
  private selectPreviousMenuItem() {
```

src/components/menu/menu.ts
```diff
@@ -256,25 +325,112 @@ export class Menu extends Component<Props, SpreadsheetChildEnv> {
...
+  forSubMenuToComeBack(direction: string, propagation: boolean) {
+    this.closeSubMenu();
+    if (this.props.backToParent === undefined) return;
+    if (propagation) {
+      this.props.backToParent(direction);
+    }
+  }
+
+  private selectNextMenuItem() {
+    if (this.subMenu.menuIndex === undefined) {
```
rrahir, line 407: if you want to loop, you could maybe write it like this
```javascript
const len = this.clickableMenuItems.length
this.subMenu.menuIndex = ( ( this.subMenu.menuIndex | 0 ) - 1 + len ) % len
```
just a suggestion though, we could argue taht your version is more readable

src/components/menu/menu.ts
```diff
@@ -256,25 +325,112 @@ export class Menu extends Component<Props, SpreadsheetChildEnv> {
...
+    } else {
+      this.subMenu.menuIndex--;
+    }
+    const el = this.menuItemRefs[this.subMenu.menuIndex].ref?.el;
+    el?.scrollIntoView(false);
+  }
+
+  private backToParentRight() {
+    if (!this.props.backToParent) return;
+    this.props.backToParent("right", true);
```
rrahir, line 432: you don't have to early return
```suggestion
    this.props?.backToParent("right", true);
```

rrahir (commented):
Really nice feature !!  Ihave using my mouse :p

src/components/menu/menu.xml
yunchen4: Yes, I tried `mouseEnter`. It has the same problem as `mouseOver` when the menu item list is too long (with scroll bar) and the menu is scrolled, because `mouseEnter` will be triggered when it enters a new menu item (div) even though the mouse doesn't move (only the scroll bar moves), whereas `mouseMove` won't.

src/components/menu/menu.ts
```diff
@@ -105,23 +120,48 @@ export class Menu extends Component<Props, SpreadsheetChildEnv> {
...
   private position: DOMCoordinates = useAbsoluteBoundingRect(this.menuRef);
+  private menuItemRefs: { action: Action; ref: Ref<HTMLElement> }[] = [];
 
   setup() {
+    this.childrenHaveIcon = this.props.menuItems.some(
+      (menuItem) => !!menuItem.icon || !!menuItem.isActive
+    );
+    this.clickableMenuItems = this.props.menuItems.filter(
+      (menuItem) => menuItem.isEnabled(this.env) && menuItem.isVisible(this.env)
+    );
```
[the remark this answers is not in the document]
yunchen4: @hokolomopo You're right, so I add a hook to calculate `clickableMenuItems` and `childrenHaveIcon`. 
For refs, since I need to use `useRef`, I cannot update them in the render hook, so I make it change with `menuItems` in props. I also add key to each menu to force rendering when switching different menus.

src/components/menu/menu.ts
yunchen4: I'll keep it because it's more readable 😊

src/components/menu/menu.ts
yunchen4: `backToParent` is called here so I cannot use `?` to avoid early return :(

src/components/composer/composer/composer.ts
```diff
@@ -437,6 +437,7 @@ export class Composer extends Component<ComposerProps, SpreadsheetChildEnv> {
     if (this.env.model.getters.isReadonly()) {
       return;
     }
+    this.composerRef.el?.focus();
```
yunchen4, line 440: This line is added to fix the bug: when the context menu/topbar menu is opened, clicking the topbar composer will throw error. But it's not perfect: the first click on the topbar composer won't set the cursor position accurately. I tried to fix it completely but failed; however I don't think it's a big problem so I keep it.

src/components/menu/menu.ts
rrahir: Sorry the suggestion was bad, I was more thinkng of something like:
you don't have to early return
```suggestion
    this.props.backToParent?.("right", true);
```

tests/components/context_menu.test.ts
```diff
@@ -905,3 +908,231 @@ describe("Context menu separator", () => {
     expect(fixture.querySelectorAll(".o-menu .o-separator").length).toBe(0);
   });
 });
+
+describe("Context menu navigation via keyboard", () => {
+  beforeAll(() => {
+    window.HTMLElement.prototype.scrollIntoView = jest.fn();
```
hokolomopo, line 914: We can probably define that in jest.setup once and for all

tests/components/top_bar.test.ts
```diff
@@ -873,4 +879,216 @@ test("The menu items are orderer by their sequence", async () => {
...
+    expect(menuButton!.classList).not.toContain("active");
+    expect(fixture.querySelectorAll(".o-menu")).toHaveLength(0);
+    window.dispatchEvent(new KeyboardEvent("keydown", { key: "F", altKey: true, bubbles: true }));
+    await nextTick();
+    expect(menuButton!.classList).toContain("active");
+    expect(fixture.querySelectorAll(".o-menu")).toHaveLength(1);
+    topbarMenuRegistry.content = menuDefinitions;
+  });
+
+  test("ALT+E to open Edit menu", async () => {
```
hokolomopo, line 904: can probably do a test.each for all of these

tests/components/top_bar.test.ts
```diff
@@ -873,4 +879,216 @@ test("The menu items are orderer by their sequence", async () => {
...
+    expect(fixture.querySelector(".o-topbar-menu[data-id='file']")!.classList).toContain("active");
+    topbarMenuRegistry.content = menuDefinitions;
+  });
+
+  test("Menu with all disabled menu items won't break the navigation", async () => {
+    const menuDefinitions = Object.assign({}, topbarMenuRegistry.content);
+    topbarMenuRegistry.add("disabled", {
+      name: "disabled",
+    });
+    topbarMenuRegistry.addChild("disabledChild", ["disabled"], {
```
hokolomopo, line 1044: image
disabled child

tests/components/top_bar.test.ts
```diff
@@ -873,4 +879,216 @@ test("The menu items are orderer by their sequence", async () => {
...
+    expect(fixture.querySelectorAll(".o-menu")).toHaveLength(1);
+    topbarMenuRegistry.content = menuDefinitions;
+  });
+
+  test("ALT+E to open Edit menu", async () => {
+    await mountParent();
+    const menuButton = fixture.querySelector(".o-topbar-menu[data-id='edit']");
+    expect(menuButton!.classList).not.toContain("active");
+    expect(fixture.querySelectorAll(".o-menu")).toHaveLength(0);
+    window.dispatchEvent(new KeyboardEvent("keydown", { key: "E", altKey: true, bubbles: true }));
```
hokolomopo, line 909: use helper `keyDown`

hokolomopo (commented):
Great work :D

src/components/menu/menu.ts
```diff
@@ -100,23 +116,65 @@ export class Menu extends Component<Props, SpreadsheetChildEnv> {
...
     position: null,
     scrollOffset: 0,
     menuItems: [],
+    menuIndex: this.props.shouldSelectFirstItem ? 0 : undefined,
+    shouldSelectFirstItem: false,
   });
   private menuRef = useRef("menu");
   private position: DOMCoordinates = useAbsoluteBoundingRect(this.menuRef);
+  private menuItemRefs: { [id: string]: { action: Action; ref: Ref<HTMLElement> } } = {};
+  // @ts-ignore: childrenHaveIcon will be used in XML
```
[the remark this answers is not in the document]
yunchen4: @hokolomopo It's for performance. If I keep `childrenHaveIcon` as getters, using keyboard to navigate a lot of menu items (e.g. insert->functions->all) would be very slow.

src/components/menu/menu.ts
```diff
@@ -251,25 +338,109 @@ export class Menu extends Component<Props, SpreadsheetChildEnv> {
...
+
+  onKeydown(ev: KeyboardEvent) {
+    ev.preventDefault();
+    ev.stopPropagation();
+    if (ev.key.startsWith("Arrow")) {
+      this.processArrows(ev);
+    } else if (ev.key === "Enter") {
+      if (this.subMenu.menuIndex === undefined) return;
+      const menu = this.clickableMenuItems[this.subMenu.menuIndex];
+      if (this.isEnabled(menu)) {
```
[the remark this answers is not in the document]
yunchen4: @hokolomopo `this.isEnabled` also check whether the menu is read-only allowed, so it's not only `menu.isEnabled()`. Deleting it would make one test in `context_menu` fail.

src/components/menu/menu.ts
```diff
@@ -251,25 +338,109 @@ export class Menu extends Component<Props, SpreadsheetChildEnv> {
...
+    }
+  }
+
+  private selectNextMenuItem() {
+    if (this.clickableMenuItems.length === 0) return;
+    if (this.subMenu.menuIndex === undefined) {
+      this.subMenu.menuIndex = 0;
+    } else if (this.subMenu.menuIndex === this.clickableMenuItems.length - 1) {
+      this.subMenu.menuIndex = 0;
+    } else {
```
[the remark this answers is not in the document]
yunchen4: After discussion I'll keep it now.

src/components/grid/grid.ts
```diff
@@ -353,7 +351,8 @@ export class Grid extends Component<Props, SpreadsheetChildEnv> {
   focus() {
     if (
       !this.env.model.getters.getSelectedFigureId() &&
-      this.env.model.getters.getEditionMode() === "inactive"
+      this.env.model.getters.getEditionMode() === "inactive" &&
+      !this.env.model.getters.hasOpenedPopover()
```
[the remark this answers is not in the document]
yunchen4: @hokolomopo It fixes the bug when the filter is opening, search input cannot be focused by clicking.

src/components/menu/menu.ts
```diff
@@ -100,23 +116,65 @@ export class Menu extends Component<Props, SpreadsheetChildEnv> {
...
+  private menuItemRefs: { [id: string]: { action: Action; ref: Ref<HTMLElement> } } = {};
+  // @ts-ignore: childrenHaveIcon will be used in XML
+  childrenHaveIcon: boolean = false;
 
   setup() {
+    this.props.menuItems.forEach((menuItem) => {
+      this.menuItemRefs[menuItem.id] = {
+        action: menuItem,
+        ref: useRef(menuItem.id),
+      };
```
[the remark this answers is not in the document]
yunchen4: @hokolomopo I've added `t-key` on the ref divs. From my manual testing I didn't find anything wrong even when corporating (menu items suddenly becomes invisible).

src/components/menu/menu.ts
```diff
@@ -100,23 +116,65 @@ export class Menu extends Component<Props, SpreadsheetChildEnv> {
...
+    });
     useExternalListener(window, "click", this.onExternalClick, { capture: true });
     useExternalListener(window, "contextmenu", this.onExternalClick, { capture: true });
+    onWillRender(() => {
```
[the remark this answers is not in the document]
yunchen4: @hokolomopo Items in `this.clickableMenuItems` can change as you and Remi mentioned, and for now I didn't find a way to "monitor" it, because the state of menu items depends on env as well. 
I also tried to make `clickableMenuItems` as getter, but it has performance issue when there are too many menu items (e.g. insert->function->all). 
We discussed about `onWillRender` this afternoon and it's indeed not good. After looking into the menus on runbot master I didn't find any menu item that would become disabled (but still visible) when corporating (something like undo & redo is local). I only found freeze & unfreeze panel menu item can become invisible and visible from time to time when corporating. Hence I use `useEffect` together with the length of `menuItemsAndSeparators` to kind of "monitor" changes of `clickableMenuItems`. This is called much less times than `onWillRender`. And also in `useEffect` I used a simpler fallback (I simply reset the item selection 😅). 
Feel free to think about it twice when you're reviewing 😊

PR #2442, opened 2023-05-03
[FIX] chart: no dataset with single cell

LucasLefevre:
When creating a chart where the dataset is a single cell and the first row is used as data set title, the dataset actually doesn't contain any data.

Before commit e96eef5, this case was detected and the dataset was ignored. Now, it's not ignored but the dataset range is invalid (since it contains no cell at all when the title is exluced).

It breaks down the line when exporting the chart to excel. This commit restores the previous behavior.

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/figures/charts/chart_common.ts
+++ b/src/helpers/figures/charts/chart_common.ts
@@ -184,8 +184,13 @@ export function createDataSets(
           )
         );
       }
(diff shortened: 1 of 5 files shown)


rrahir:
@robodoo r+

PR #2451, opened 2023-05-04
Version history required changes

SimonGenin:
Fix a bug in side panel.

Add timestamp to revisions.

Fix sub menus in top bar order to respect their sequence.

### DIFF
--- a/src/collaborative/revisions.ts
+++ b/src/collaborative/revisions.ts
@@ -20,7 +20,8 @@ export class Revision implements RevisionData {
     clientId: ClientId,
     commands: readonly CoreCommand[],
     readonly rootCommand?: Command | "SNAPSHOT" | "REMOTE" | undefined,
-    changes?: readonly HistoryChange[]
+    changes?: readonly HistoryChange[],
+    readonly timestamp?: number
   ) {
     this.id = id;
     this.clientId = clientId;
--- a/src/types/collaborative/transport_service.ts
+++ b/src/types/collaborative/transport_service.ts
@@ -13,6 +13,7 @@ export interface RemoteRevisionMessage extends AbstractMessage {
   commands: readonly CoreCommand[];
   nextRevisionId: UID;
   serverRevisionId: UID;
+  timestamp?: number;
 }
 
 export interface RevisionUndoneMessage extends AbstractMessage {
--- a/src/collaborative/session.ts
+++ b/src/collaborative/session.ts
@@ -87,7 +87,8 @@ export class Session extends EventBus<CollaborativeEvent> {
       this.clientId,
       commands,
       rootCommand,
-      changes
+      changes,
+      Date.now()
     );
     this.revisions.append(revision.id, revision);
     this.trigger("new-local-state-update", { id: revision.id });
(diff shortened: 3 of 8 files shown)


src/components/side_panel/side_panel/side_panel.ts
```diff
@@ -204,10 +205,17 @@ export class SidePanel extends Component<Props, SpreadsheetChildEnv> {
...
+      lastSidePanelComponentSeen: this.props.component,
+    });
+    onWillUpdateProps((nextProps) => {
+      this.state.panel = sidePanelRegistry.get(nextProps.component);
+      // changing from one side panel to another should call the onCloseSidePanel as
+      // it may contain important side effects
+      if (this.state.lastSidePanelComponentSeen !== nextProps.component) {
+        this.state.lastSidePanelComponentSeen = nextProps.component;
+        this.props.panelProps?.onCloseSidePanel?.();
+      }
```
SimonGenin, line 217: This feel underwhelmingly complex, but a more simple approach would create an infinite cycle of props updating.

SimonGenin: the simpler approach being:
if (this.props.component && this.props.component !== nextProps.component) {
	// ... 
}

LucasLefevre: What about handling this in the Spreadsheet component?
It's not rocket science but I think it works and it's simple
```diff
diff --git a/src/components/spreadsheet/spreadsheet.ts b/src/components/spreadsheet/spreadsheet.ts
index a2b65a4e..6445fd5f 100644
--- a/src/components/spreadsheet/spreadsheet.ts
+++ b/src/components/spreadsheet/spreadsheet.ts
@@ -258,6 +258,9 @@ export class Spreadsheet extends Component<SpreadsheetProps, SpreadsheetChildEnv
   }
 
   openSidePanel(panel: string, panelProps: any) {
+    if (this.sidePanel.isOpen && panel !== this.sidePanel.component) {
+      this.sidePanel.panelProps?.onCloseSidePanel?.();
+    }
     this.sidePanel.component = panel;
     this.sidePanel.panelProps = panelProps;
     this.sidePanel.isOpen = true;
```

src/collaborative/revisions.ts
```diff
@@ -20,7 +20,8 @@ export class Revision implements RevisionData {
     clientId: ClientId,
     commands: readonly CoreCommand[],
     readonly rootCommand?: Command | "SNAPSHOT" | "REMOTE" | undefined,
-    changes?: readonly HistoryChange[]
+    changes?: readonly HistoryChange[],
+    readonly timestamp?
```
LucasLefevre, line 24: with the type :wink: 
`readonly timestamp?: number,`

src/types/collaborative/transport_service.ts
```diff
@@ -13,6 +13,7 @@ export interface RemoteRevisionMessage extends AbstractMessage {
   commands: readonly CoreCommand[];
   nextRevisionId: UID;
   serverRevisionId: UID;
+  timestamp?: number;
```
LucasLefevre, line 16: I'm actually not sure what it should be.
A real timestamp (number) ? An ISO 8601 string `"2023-05-05T06:56"` ? A `Date` object ?
A string is certainly more readable when debugging, but it's not type safe since it will happily accept the string "prout".
Odoo gives the `create_date` as an ISO string, which can also be taken into consideration

SimonGenin: I'd say there is no "real better" option. However, the most surprising type for a timestamp amongst string, number or date would be string imo. 
So I'm kinda inclined to keep it as a number.

src/components/side_panel/side_panel/side_panel.ts
SimonGenin: Good call it works, so I can remove this ugly stuff

LucasLefevre (approved):
robodoo r+

LucasLefevre:
robodoo rebase-ff

PR #2455, opened 2023-05-05
[FIX] repeat: faster can redo check

LucasLefevre:
With more and more pending revision, `canRedo` was taking more and more time. Even if it's not that slow (~10ms for 3700 revisions), it's done at every render! Scrolling feels laggy.

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


LucasLefevre:
@hokolomopo what do you think?

src/collaborative/session.ts
```diff
@@ -90,6 +91,9 @@ export class Session extends EventBus<CollaborativeEvent> {
       changes
     );
     this.revisions.append(revision.id, revision);
+    if (rootCommand.type !== "REQUEST_REDO") {
+      this.lastLocalOperation = revision;
+    }
```
hokolomopo, line 96: I think this is useless ATM, request redo won't have any sub-commands and will always be bounced by the `if` above. 

Also we don't handle `REQUEST_REDO` at all. We rely on the fact that we won't repeat a command as long as the revision that was undone is not redone (and thus this.lastLocalOperation becomes valid again). But this can be wrong if the revision in the undo stack is dropped because of a remote command. We probably want to update `this.lastLocalOperation ` to a valid value when we catch a `REQUEST_UNDO`

hokolomopo (commented):
I think we need to handle `REQUEST_REDO`, but otherwise it looks better and simpler to what I did :)

src/collaborative/session.ts
rrahir: > I think this is useless ATM, request redo won't have any sub-commands and will always be bounced by the `if` above.

Didn't you introduce the repetition of commands recently? :-p  


> Also we don't handle `REQUEST_REDO` at all. We rely on the fact that we won't repeat a command as long as the revision that was undone is not redone (and thus this.lastLocalOperation becomes valid again). But this can be wrong if the revision in the undo stack is dropped because of a remote command. We probably want to update `this.lastLocalOperation ` to a valid value when we catch a `REQUEST_UNDO`
I also think we end up with a dirty `lastLocalOperation` on undo.

LucasLefevre: > But this can be wrong if the revision in the undo stack is dropped because of a remote command. We probably want to update this.lastLocalOperation to a valid value when we catch a REQUEST_UNDO

@hokolomopo 
The transformed operation might be dropped, the local operation still bears the user's intention (at the time he performed the operation). And the repeat process will adapts the operation to the current state.
Can we really end up in a situation where something wrong happens?

hokolomopo: You might be right. I found a somewhat of a bug but I don't think it's related to your PR: 

- user A updates a cell in sheet1
- user A undo
- user B delete sheet 1
- user A try "redo"
    - nothing visible happens
- user A redo again
    - the first update cell is repeated at the current selection

LucasLefevre: not related to this PR indeed. It's also buggy in master.

rrahir:
@robodoo r+

PR #2458, opened 2023-05-09
[FIX] composer: autocomplete in topbar composer

jash-odoo:
## Description:

Previously, when we write a formula and tried to autocomplete it by pressing tab or enter, it wouldn't update topbar composer. To fix this, we update a onKeydown to manage the enter and tab buttons.

Odoo task ID : [3291781](https://www.odoo.com/web#id=3291781&cids=2&menu_id=4720&action=333&active_id=2328&model=project.task&view_type=form)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/composer/composer/composer.ts
+++ b/src/components/composer/composer/composer.ts
@@ -247,6 +247,11 @@ export class Composer extends Component<Props, SpreadsheetChildEnv> {
         this.autocompleteAPI.moveDown();
       }
     }
+    if (!this.env.model.getters.isSelectingForComposer()) {
+      const { start, end } = this.contentHelper.getCurrentSelection();
+      this.env.model.dispatch("CHANGE_COMPOSER_CURSOR_SELECTION", { start, end });
+      this.isKeyStillDown = true;
+    }
   }
 
   private processTabKey(ev: KeyboardEvent) {
@@ -308,11 +313,6 @@ export class Composer extends Component<Props, SpreadsheetChildEnv> {
       handler.call(this, ev);
     } else {
       ev.stopPropagation();
-    }
-    const { start, end } = this.contentHelper.getCurrentSelection();
-    if (!this.env.model.getters.isSelectingForComposer()) {
-      this.env.model.dispatch("CHANGE_COMPOSER_CURSOR_SELECTION", { start, end });
-      this.isKeyStillDown = true;
     }
   }
 
--- a/tests/components/autocomplete_dropdown.test.ts
+++ b/tests/components/autocomplete_dropdown.test.ts
@@ -186,6 +186,25 @@ describe("Functions autocomplete", () => {
       expect(document.activeElement).toBe(composerEl);
       expect(fixture.querySelectorAll(".o-autocomplete-value")).toHaveLength(0);
     });
+
+    test.each(["Enter", "Tab"])(
+      "=S(A1:A5) +ENTER complete the function --> =SUM(A1:A5)",
+      async (buttonkey) => {
+        await typeInComposerGrid("=S(A1:A5)");
+        model.dispatch("STOP_COMPOSER_RANGE_SELECTION");
+        model.dispatch("CHANGE_COMPOSER_CURSOR_SELECTION", { start: 2, end: 2 });
+        await nextTick();
+        await typeInComposerGrid("U", false);
+        expect(model.getters.getCurrentContent()).toBe("=SU(A1:A5)");
+        expect(model.getters.getComposerSelection()).toEqual({ start: 3, end: 3 });
+        expect(document.activeElement).toBe(composerEl);
+        expect(fixture.querySelectorAll(".o-autocomplete-value")).toHaveLength(1);
+        await nextTick();
+        composerEl.dispatchEvent(new KeyboardEvent("keydown", { key: buttonkey }));
+        await nextTick();
+        expect(composerEl.textContent).toBe("=SUM(A1:A5)");
+      }
+    );
   });
 
   describe("autocomplete functions SUM IF", () => {
@@ -250,6 +269,12 @@ describe("Autocomplete parenthesis", () => {
       compute: () => 1,
       returns: ["ANY"],
     });
+    functionRegistry.add("SUMIF", {
+      description: "do sumif",
+      args: args(``),
+      compute: () => 1,
+      returns: ["ANY"],
+    });
     functionRegistry.add("SZZ", {
       description: "do something",
       args: args(``),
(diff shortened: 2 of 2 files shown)


tests/components/autocomplete_dropdown.test.ts
```diff
@@ -186,6 +186,25 @@ describe("Functions autocomplete", () => {
       expect(document.activeElement).toBe(composerEl);
       expect(fixture.querySelectorAll(".o-autocomplete-value")).toHaveLength(0);
     });
+
+    test.each(["Enter", "Tab"])(
+      "=S(A1:A5) +ENTER complete the function --> =SUM(A1:A5)",
+      async (buttonkey) => {
```
hokolomopo, line 192:
```suggestion
      "=S(A1:A5) + %s complete the function --> =SUM(A1:A5)",
      async (buttonKey) => {
```

hokolomopo (commented):
Looks good :+1:

src/components/composer/composer/composer.ts
```diff
@@ -309,8 +309,8 @@ export class Composer extends Component<Props, SpreadsheetChildEnv> {
     } else {
       ev.stopPropagation();
     }
-    const { start, end } = this.contentHelper.getCurrentSelection();
-    if (!this.env.model.getters.isSelectingForComposer()) {
+    if (!this.env.model.getters.isSelectingForComposer() && !["Enter", "Tab"].includes(ev.key)) {
```
rrahir, line 312: this will break the multi line inputs starting saas-16.1. If I'm not mistaken,  the specificity of Enter and Tab is that they stop the edition inside their handlers, so we might just have to check that we are not  in selection mode and we are still editing, Since there are only two modes of edition, we can just check that 
```typescript
this.env.model.getters.editionMode() === "editing"
```

LucasLefevre: >  the specificity of Enter and Tab is that they stop the edition inside their handlers

not when their effect is to autocomplete.

LucasLefevre, line 312: What I don't like about the proposed solution here is that handlers business are leaking in the generic `onKeydown`.
I'm wondering if the selection changed shouldn't be moved inside the `else`.
And managed manually by the handlers if they need to (probably only arrow keys, and they are already, to a certain extend)

rrahir: might be a better solution, yup

jash-odoo: Are you referring to moving the external condition into the if condition? If so, then it seems to work on UI but test cases are not liking it. Second approach I tried was to remove the external condition completely, and that also did solve the issue but it breaks the test cases as well.

So I just want to get a more clear idea why do we have the external condition? Handler can take care of the code with desired behavior quiet well, yet we have an dispatch outside to set the cursor place. (Yes I have referred the commit which brought the change but got confused a bit)

jash-odoo: I guess now I got the idea behind it, I have made some changes to my code according to your suggestion. Hopefully this will serve as required. Let me know if any changes are required / if I misunderstand something.

src/components/composer/composer/composer.ts
```diff
@@ -302,17 +302,27 @@ export class Composer extends Component<Props, SpreadsheetChildEnv> {
...
+    this.isKeyStillDown = true;
+  }
+
   onKeydown(ev: KeyboardEvent) {
     let handler = this.keyMapping[ev.key];
     if (handler) {
       handler.call(this, ev);
+      if (
+        this.env.model.getters.getEditionMode() === "editing" &&
+        !(["Enter", "Tab"].includes(ev.key) && !(ev.altKey || ev.ctrlKey))
```
rrahir, line 317: while the code changed, the situation raised by Lucas is still present, you even managed to make it worse...  I am confused because I don't know where the issue lies in our explanation, I will try to be more clear.

Lucas said: 
> What I don't like about the proposed solution here is that handlers business are leaking in the generic onKeydown.

this means that what we would like to get rid of is: " the mention of `Enter` or `Tab`  because these notions of which key was pressed are handled in the [**handlers**](https://github.com/odoo/o-spreadsheet/commit/d71874993cb17d71ff972e208a536c195b7b4b31#diff-ea6fa27863d7bc4c77db3f9c449298a78c2e689d06ed2d92f0e3207fc4025251R312),  not in the general `onKeydown`  function.  your current proposition does not change that at all.

We want  `updateSelectionRequired`  to be only called by specific handlers that require it - so in `keymapping`, some entries would handle it, others would not.

Aside from this,  there is no point in forseeing the upcoming multine behaviour with 
```javascript
!(ev.altKey || ev.ctrlKey)
```
1. multi line is not there yet, so this accomplishes nothing
2. this only applies to `Enter`, not `Tab`
3.  so again, this hints that it should be handled separately in the handlers, and not in the generic caller `onKeyDown`

tests/components/autocomplete_dropdown.test.ts
```diff
@@ -312,6 +337,30 @@ describe("Autocomplete parenthesis", () => {
...
+  test.each(["Enter", "Tab"])(
+    "=SUM( + edit SUM with autocomplete does not add left parenthesis",
+    async (buttonkey) => {
+      await typeInComposerGrid("=SUM(");
+      // go behind the letter "M"
+      model.dispatch("STOP_COMPOSER_RANGE_SELECTION");
+      model.dispatch("CHANGE_COMPOSER_CURSOR_SELECTION", { start: 4, end: 4 });
+      await nextTick();
+      // show autocomplete
+      await typeInComposerGrid("I", false);
```
rrahir, line 349: actually, the 'sumif' function is not part of the functionregistry in this file,  see `beforeAll` at line 238, this means that you should **not** have any autocomplete suggestion once you add the `I` in the formula name. you should either use the compone t behaviour to do everything, instead of dispatching commands in the model directly, or maybe consider that the test is not at the right place in the file?

tests/components/autocomplete_dropdown.test.ts
```diff
@@ -312,6 +337,30 @@ describe("Autocomplete parenthesis", () => {
     expect(model.getters.getComposerSelection()).toEqual({ start: 5, end: 5 });
   });
 
+  test.each(["Enter", "Tab"])(
+    "=SUM( + edit SUM with autocomplete does not add left parenthesis",
```
rrahir, line 341: is it really what you were testing against? Was the bug allowing one to duplicate the left parenthesis?

rrahir (changes requested):
As Adrien said, he did not test the behaviour in older version. turns out this bug has been there for a long time. Could you target this fix in version 15.0?

tests/components/autocomplete_dropdown.test.ts
jash-odoo: Not primarily, I was going through the test cases and I thought it's a good thing to add test cases for these two cases as well. We can remove it ofcourse if it's an extra.

tests/components/autocomplete_dropdown.test.ts
jash-odoo: L272, I have already added `SUMIF` in functionRegisty for `Autocomplete Parenthesis` set of test cases.

PR #2462, opened 2023-05-10
[FIX] ContentEditableHelper: do not throw on missing node

rrahir:
Currently,the method `findSizeBeforeElement` will throw if called when the selection is not targetting a node inside the ContentEditableHelper instance.

Unfortunately, this can happen since commit d4663158. If a menu is opened and we decide to click on the composer, the click event if first caught by the menu global listener which will close the menu and invoke a callback that focuses the grid; this focus will alter the selection of `document`. The event is then handled by the composer component but at that point, the selection is not targetting it anymore.

With this revision, we change the behavior of `findSizeBeforeElement` to return a default value of 0 in those situations as we simply don't know how to determine the selection outside of the scope of the ContentEditableHelper.

There is a functional drawback to this solution:
When clicking on a specific character of the composer when a menu is opened, the selection will be set to `{ start: 0, end: 0 }`.

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/composer/content_editable_helper.ts
+++ b/src/components/composer/content_editable_helper.ts
@@ -199,7 +199,14 @@ export class ContentEditableHelper {
       current = it.next();
     }
     if (current.value !== nodeToFind) {
-      throw new Error("Cannot find the node in the children of the element");
+      /** This situation can happen if the code is called while the selection is not currently on the ContentEditableHelper.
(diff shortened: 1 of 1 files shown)


LucasLefevre (approved):
robodoo r+

PR #2473, opened 2023-05-11
[FIX] clipboard: fix behaviour when selection is updated before the paste

khpa-odoo:
## Description:

Cut/Copy:
After a cut/copy operation, if the selection is updated before paste,  the content was pasted w/o any style. Also, the cut zone was not destroyed in case of cut operation.

This fix solves the problem by using the cells stored inside the Clipboard Plugin instead of the current state of the cells in the grid.

Also, this fix restricts pasting the content multiple times after a cut operation and clears the clipboard after a cut and single paste as well.

Only Format:
Copy external content and do 'Paste' or 'Paste-Only Value' on a cell. Now,  performing a 'Paste-Only Format' on a different cell would paste the content as well.

This fix solves the problem by handling content in case the pasteOption is 'onlyFormat'.

Odoo task ID : [3135623](https://www.odoo.com/web#id=3135623&cids=2&menu_id=4720&action=333&active_id=2328&model=project.task&view_type=form)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/clipboard/clipboard_cells_state.ts
+++ b/src/helpers/clipboard/clipboard_cells_state.ts
@@ -30,7 +30,7 @@ interface CopiedTable {
 
 /** State of the clipboard when copying/cutting cells */
 export class ClipboardCellsState extends ClipboardCellsAbstractState {
-  private readonly cells: ClipboardCell[][];
+  private cells: ClipboardCell[][];
   private readonly copiedTables: CopiedTable[];
   private readonly zones: Zone[];
 
@@ -210,6 +210,13 @@ export class ClipboardCellsState extends ClipboardCellsAbstractState {
       });
     }
     this.pasteCopiedTables(target);
+    this.cells.forEach((row) => {
+      row.forEach((c) => {
+        if (c.cell) {
+          c.cell = undefined;
+        }
+      });
+    });
   }
 
   /**
@@ -479,7 +486,9 @@ export class ClipboardCellsState extends ClipboardCellsAbstractState {
         .map((cells) => {
           return cells
             .map((c) =>
-              c.cell ? this.getters.getCellText(c.position, this.getters.shouldShowFormulas()) : ""
+              this.getters.shouldShowFormulas() && c.cell?.isFormula
+                ? c.cell?.content || ""
+                : c.evaluatedCell?.formattedValue || ""
             )
             .join("\t");
         })
--- a/src/helpers/clipboard/clipboard_os_state.ts
+++ b/src/helpers/clipboard/clipboard_os_state.ts
@@ -36,7 +36,10 @@ export class ClipboardOsState extends ClipboardCellsAbstractState {
     return CommandResult.Success;
   }
 
-  paste(target: Zone[]) {
+  paste(target: Zone[], clipboardOption?: ClipboardOptions) {
+    if (clipboardOption?.pasteOption === "onlyFormat") {
+      return;
+    }
     const values = this.values;
     const pasteZone = this.getPasteZone(target);
     const { left: activeCol, top: activeRow } = pasteZone;
--- a/src/plugins/ui_stateful/clipboard.ts
+++ b/src/plugins/ui_stateful/clipboard.ts
@@ -16,6 +16,7 @@ import {
   isCoreCommand,
   LAYERS,
   LocalCommand,
+  UID,
   Zone,
 } from "../../types/index";
 import { UIPlugin } from "../ui_plugin";
@@ -43,8 +44,8 @@ export class ClipboardPlugin extends UIPlugin {
   private status: "visible" | "invisible" = "invisible";
   private state?: ClipboardState;
   private lastPasteState?: ClipboardState;
-
   private _isPaintingFormat: boolean = false;
+  private originSheetId?: UID;
 
   // ---------------------------------------------------------------------------
   // Command Handling
@@ -87,6 +88,7 @@ export class ClipboardPlugin extends UIPlugin {
         const zones = ("cutTarget" in cmd && cmd.cutTarget) || this.getters.getSelectedZones();
         this.state = this.getClipboardState(zones, cmd.type);
         this.status = "visible";
+        this.originSheetId = this.getters.getActiveSheetId();
         break;
       case "PASTE":
         if (!this.state) {
@@ -95,9 +97,6 @@ export class ClipboardPlugin extends UIPlugin {
         const pasteOption = cmd.pasteOption || (this._isPaintingFormat ? "onlyFormat" : undefined);
         this._isPaintingFormat = false;
         this.state.paste(cmd.target, { pasteOption, shouldPasteCF: true, selectTarget: true });
-        if (this.state.operation === "CUT") {
-          this.state = undefined;
-        }
         this.lastPasteState = this.state;
         this.status = "invisible";
         break;
@@ -176,6 +175,15 @@ export class ClipboardPlugin extends UIPlugin {
         this.status = "visible";
         break;
       }
+      case "DELETE_SHEET":
+        if (this.state?.operation !== "CUT") {
+          return;
+        }
+        if (this.originSheetId === cmd.sheetId) {
+          this.state = undefined;
+          this.status = "invisible";
+        }
+        break;
       default:
         if (isCoreCommand(cmd)) {
           this.status = "invisible";
--- a/src/plugins/ui_stateful/clipboard.ts
+++ b/src/plugins/ui_stateful/clipboard.ts
@@ -65,7 +65,7 @@ export class ClipboardPlugin extends UIPlugin {
         return this.state.isPasteAllowed(cmd.target, { pasteOption });
       case "PASTE_FROM_OS_CLIPBOARD": {
         const state = new ClipboardOsState(cmd.text, this.getters, this.dispatch, this.selection);
-        return state.isPasteAllowed(cmd.target);
+        return state.isPasteAllowed(cmd.target, { pasteOption: cmd.pasteOption });
       }
       case "INSERT_CELL": {
         const { cut, paste } = this.getInsertCellsTargets(cmd.zone, cmd.shiftDimension);
@@ -156,7 +156,7 @@ export class ClipboardPlugin extends UIPlugin {
       }
       case "PASTE_FROM_OS_CLIPBOARD":
         this.state = new ClipboardOsState(cmd.text, this.getters, this.dispatch, this.selection);
-        this.state.paste(cmd.target);
+        this.state.paste(cmd.target, { pasteOption: cmd.pasteOption });
         this.lastPasteState = this.state;
         this.status = "invisible";
         break;
--- a/src/types/commands.ts
+++ b/src/types/commands.ts
@@ -541,6 +541,7 @@ export interface PasteFromOSClipboardCommand {
   type: "PASTE_FROM_OS_CLIPBOARD";
   target: Zone[];
   text: string;
+  pasteOption?: ClipboardPasteOptions;
 }
 
 export interface AutoresizeColumnsCommand {
--- a/tests/components/grid.test.ts
+++ b/tests/components/grid.test.ts
@@ -13,7 +13,7 @@ import {
 import { buildSheetLink, toCartesian, toHex, toZone, zoneToXc } from "../../src/helpers";
 import { createEmptyWorkbookData } from "../../src/migrations/data";
 import { Model } from "../../src/model";
-import { Align, HeaderDimensions, UID } from "../../src/types";
+import { Align, ClipboardMIMEType, HeaderDimensions, UID } from "../../src/types";
 import { getClipboardEvent, MockClipboardData } from "../test_helpers/clipboard";
 import {
   copy,
@@ -31,6 +31,7 @@ import {
   selectColumn,
   selectRow,
   setCellContent,
+  setCellFormat,
   setSelection,
   setStyle,
   updateFilter,
@@ -54,6 +55,7 @@ import {
   getCellContent,
   getCellText,
   getClipboardVisibleZones,
+  getEvaluatedCell,
   getSelectionAnchorCellXc,
   getStyle,
 } from "../test_helpers/getters_helpers";
@@ -1425,6 +1427,56 @@ describe("Copy paste keyboard shortcut", () => {
     document.body.dispatchEvent(getClipboardEvent("paste", clipboardData));
     expect(getCellContent(model, "A1")).toEqual("");
     expect(getCellContent(model, "A2")).toEqual("things");
+  });
+
+  test("cut zone gets cleared on paste if content/style is altered after cut", async () => {
+    setCellContent(model, "A1", "things");
+    setStyle(model, "A1", { bold: true });
+    selectCell(model, "A1");
+    document.body.dispatchEvent(getClipboardEvent("cut", clipboardData));
+    setCellContent(model, "A1", "new content");
+    setStyle(model, "A1", { bold: false });
+    selectCell(model, "A2");
+    document.body.dispatchEvent(getClipboardEvent("paste", clipboardData));
+    expect(getCellContent(model, "A2")).toEqual("things");
+    expect(getStyle(model, "A2")).toEqual({ bold: true });
+    expect(getCell(model, "A1")).toBe(undefined);
+  });
+
+  test("Cut of a formula cell, and enabling showFormulas should return content", async () => {
+    model.dispatch("SET_FORMULA_VISIBILITY", { show: true });
+    setCellContent(model, "A1", "1");
+    setCellFormat(model, "A1", "m/d/yyyy");
+    document.body.dispatchEvent(getClipboardEvent("cut", clipboardData));
+    expect(clipboardData.getData(ClipboardMIMEType.PlainText)).toEqual(getCellContent(model, "A1"));
+    model.dispatch("SET_FORMULA_VISIBILITY", { show: false });
+    selectCell(model, "A2");
+    document.body.dispatchEvent(getClipboardEvent("paste", clipboardData));
+    expect(getCellContent(model, "A2")).toEqual("12/31/1899");
+  });
+
+  test("Cut of a formula cell, or non-formula cell with showFormulas should return its formattedValue", async () => {
+    setCellContent(model, "A1", "1");
+    setCellFormat(model, "A1", "m/d/yyyy");
+    document.body.dispatchEvent(getClipboardEvent("cut", clipboardData));
+    expect(clipboardData.getData(ClipboardMIMEType.PlainText)).toEqual(
+      getEvaluatedCell(model, "A1").formattedValue
+    );
+    selectCell(model, "A2");
+    document.body.dispatchEvent(getClipboardEvent("paste", clipboardData));
+    expect(getCellContent(model, "A2")).toEqual("12/31/1899");
+
+    model.dispatch("SET_FORMULA_VISIBILITY", { show: true });
+    setCellContent(model, "B1", "1");
+    selectCell(model, "B1");
+    document.body.dispatchEvent(getClipboardEvent("cut", clipboardData));
+    expect(clipboardData.getData(ClipboardMIMEType.PlainText)).toEqual(
+      getEvaluatedCell(model, "B1").formattedValue
+    );
+    model.dispatch("SET_FORMULA_VISIBILITY", { show: false });
+    selectCell(model, "B2");
+    document.body.dispatchEvent(getClipboardEvent("paste", clipboardData));
+    expect(getCellContent(model, "B2")).toEqual("1");
   });
 
   test("can paste value only with CTRL+SHIFT+V", async () => {
--- a/tests/menu_items_registry.test.ts
+++ b/tests/menu_items_registry.test.ts
@@ -17,6 +17,7 @@ import {
   setAnchorCorner,
   setCellContent,
   setSelection,
+  setStyle,
 } from "./test_helpers/commands_helpers";
 import { getCell, getCellContent, getEvaluatedCell } from "./test_helpers/getters_helpers";
 import {
@@ -156,6 +157,34 @@ describe("Menu Item actions", () => {
     expect(getCellContent(model, "C3")).toEqual("a1");
   });
 
+  test("Paste only-format from OS clipboard should paste nothing", async () => {
+    await env.clipboard.writeText("Copy in OS clipboard");
+    selectCell(model, "A1");
+    await doAction(["edit", "paste_special", "paste_special_format"], env);
+    expect(dispatch).toHaveBeenCalledWith("PASTE_FROM_OS_CLIPBOARD", {
+      text: "Copy in OS clipboard",
+      target: target("A1"),
+      pasteOption: "onlyFormat",
+    });
+    expect(getCellContent(model, "A1")).toEqual("");
+  });
+
+  test("Internal copy followed by OS copy should not bring paste format from internal copy", async () => {
+    setCellContent(model, "C1", "c1");
+    setStyle(model, "C1", { fillColor: "#FA0000" });
+    selectCell(model, "C1");
+    doAction(["edit", "copy"], env); // first copy from grid
+    await env.clipboard.writeText("Then copy in OS clipboard");
+    selectCell(model, "A1");
+    await doAction(["edit", "paste_special", "paste_special_format"], env);
+    expect(dispatch).toHaveBeenCalledWith("PASTE_FROM_OS_CLIPBOARD", {
+      text: "Then copy in OS clipboard",
+      target: target("A1"),
+      pasteOption: "onlyFormat",
+    });
+    expect(getCellContent(model, "A1")).toEqual("");
+  });
+
   test("Edit -> paste_special should be hidden after a CUT ", () => {
     model.dispatch("CUT", { cutTarget: env.model.getters.getSelectedZones() });
     expect(getNode(["edit", "paste_special"]).isVisible(env)).toBeFalsy();
@@ -182,13 +211,26 @@ describe("Menu Item actions", () => {
     expect(dispatch).toHaveBeenCalledWith("PASTE_FROM_OS_CLIPBOARD", {
       target: target("A1"),
       text,
+      pasteOption: "onlyValue",
     });
   });
 
-  test("Edit -> paste_special -> paste_special_format", () => {
-    doAction(["edit", "paste_special", "paste_special_format"], env);
+  test("Edit -> paste_special -> paste_special_format", async () => {
+    doAction(["edit", "copy"], env);
+    await doAction(["edit", "paste_special", "paste_special_format"], env);
     expect(dispatch).toHaveBeenCalledWith("PASTE", {
       target: env.model.getters.getSelectedZones(),
+      pasteOption: "onlyFormat",
+    });
+  });
+
+  test("Edit -> paste_special -> paste_special_format from OS clipboard", async () => {
+    const text = "in OS clipboard";
+    await env.clipboard.writeText(text);
+    await doAction(["edit", "paste_special", "paste_special_format"], env);
+    expect(dispatch).toHaveBeenCalledWith("PASTE_FROM_OS_CLIPBOARD", {
+      target: target("A1"),
+      text,
       pasteOption: "onlyFormat",
     });
   });
--- a/tests/menu_items_registry.test.ts
+++ b/tests/menu_items_registry.test.ts
@@ -143,6 +143,19 @@ describe("Menu Item actions", () => {
     expect(getCellContent(model, "A1")).toEqual("");
   });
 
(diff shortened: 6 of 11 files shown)


tests/menu_items_registry.test.ts
```diff
@@ -173,13 +173,18 @@ describe("Menu Item actions", () => {
     expect(dispatch).toHaveBeenCalledWith("PASTE_FROM_OS_CLIPBOARD", {
       target: target("A1"),
       text,
+      pasteOption: "onlyValue",
     });
   });
 
-  test("Edit -> paste_special -> paste_special_format", () => {
```
hokolomopo, line 179: Why change this test ? Only "paste_special_value from OS clipboard" should have been changed right ?

src/plugins/ui_stateful/clipboard.ts
```diff
@@ -157,9 +156,10 @@ export class ClipboardPlugin extends UIPlugin {
       }
       case "PASTE_FROM_OS_CLIPBOARD":
         this.state = new ClipboardOsState(cmd.text, this.getters, this.dispatch, this.selection);
-        this.state.paste(cmd.target);
+        this.state.paste(cmd.target, { pasteOption: cmd.pasteOption });
         this.lastPasteState = this.state;
         this.status = "invisible";
+        this.allowPaste = true;
```
hokolomopo, line 162: Is that useful ?

src/plugins/ui_stateful/clipboard.ts
```diff
@@ -87,17 +87,16 @@ export class ClipboardPlugin extends UIPlugin {
         const zones = ("target" in cmd && cmd.target) || this.getters.getSelectedZones();
         this.state = this.getClipboardState(zones, cmd.type);
         this.status = "visible";
+        this.allowPaste = true;
         break;
       case "PASTE":
-        if (!this.state) {
+        if (!this.state || !this.allowPaste) {
           break;
         }
```
hokolomopo, line 95: this should be in the allowDispatch

src/plugins/ui_stateful/clipboard.ts
```diff
@@ -87,17 +87,16 @@ export class ClipboardPlugin extends UIPlugin {
...
+        if (this.state?.operation === "CUT") this.allowPaste = false;
```
hokolomopo, line 96:
```suggestion
        if (this.state?.operation === "CUT") {
            this.allowPaste = false;
        }
```
some peoples in the teeam don't like one-line condition, so try to avoid them.

tests/plugins/clipboard.test.ts
```diff
@@ -190,16 +192,140 @@ describe("clipboard", () => {
...
+    pasteFromOSClipboard(model, "C3", content, "onlyFormat");
+    expect(getCellContent(model, "C3")).toBe("");
+  });
+
+  test("Internal copy followed by OS copy should not bring paste format from internal copy", async () => {
+    const model = new Model();
+    setCellContent(model, "A1", "a1");
+    setStyle(model, "A1", { fillColor: "#FA0000" });
+    copy(model, "A1");
+    expect(getStyle(model, "A1")).toEqual({ fillColor: "#FA0000" });
```
hokolomopo, line 273: This is useless no?

src/helpers/clipboard/clipboard_os_state.ts
```diff
@@ -45,10 +45,16 @@ export class ClipboardOsState extends ClipboardCellsAbstractState {
     this.addMissingDimensions(numberOfCols, numberOfRows, activeCol, activeRow);
     for (let i = 0; i < values.length; i++) {
       for (let j = 0; j < values[i].length; j++) {
+        let content: string;
+        if (clipboardOption?.pasteOption === "onlyFormat") {
+          content = "";
```
hokolomopo, line 50: This should be undefined not an empty string, we don't want to clear the cell when pasting only the format. 
Or better we can just do an early return for `onlyFormat`, we don't need to loop on the cells and dispatch UPDATE_CELL  to do nothing on them.

hokolomopo (commented):
You should make sure taht the tests actually test what you want, here some of the tests you created don't really test the issue. 

A good way to do that is to launch the test without your fix; they should fail here, and only work after your fix.

tests/menu_items_registry.test.ts
khpa-odoo: Since we've changed the case to be called for `onlyFormat`, this test case would expect paste_special_format action to dispatch "PASTE_FROM_OS_CLIPBOARD" instead of "PASTE". Hence the change :)

src/plugins/ui_stateful/clipboard.ts
khpa-odoo: If we cut something, 'this.allowPaste' will be set to false. When we paste external text now, the first paste works perfectly. When you try to paste that text again, "PASTE" is called, and 'this.allowPaste = false' prevents the content from being pasted. As a result, we set 'this.allowPaste = true' to allow repeated pastes when using external content.

tests/plugins/clipboard.test.ts
khpa-odoo: Main objective behind this test case was to check an edge case with "onlyFormat". Once we copy something internally, & followed we make an external copy, it was observed that upon special pasting format, it was bringing the format of first internal copy, which is obviously wrong. Hence style and content is set beforehand. Although your point does stay that pasteFromOSClipboard method will disregard the above anyway, But then I don't really can figure out how to recreate this issue in test case.

src/helpers/clipboard/clipboard_os_state.ts
```diff
@@ -45,12 +45,16 @@ export class ClipboardOsState extends ClipboardCellsAbstractState {
...
       for (let j = 0; j < values[i].length; j++) {
-        this.dispatch("UPDATE_CELL", {
-          row: activeRow + i,
-          col: activeCol + j,
-          content: values[i][j],
-          sheetId,
-        });
+        if (clipboardOption?.pasteOption === "onlyFormat") {
+          return;
+        } else {
```
hokolomopo, line 50: place the `if() return` before the loop.

src/plugins/ui_stateful/clipboard.ts
```diff
@@ -52,19 +52,31 @@ export class ClipboardPlugin extends UIPlugin {
 
   allowDispatch(cmd: LocalCommand): CommandResult {
     switch (cmd.type) {
+      case "COPY":
+        this.allowPaste = true;
+        break;
```
hokolomopo, line 57: We should't change the state of the plugin in the `allowDispatch`. It doesn't change much ATM, but we could imagine another plugin that have an allowDispatch on `COPY`. We could then have the situation:

- ClipboardPlugin allowDispatch COPY, set `allowPaste` to true
- Random plugin block the COPY in its allowDispatch.
- In this case the state of the ClipboardPlugin is wrong, because allowPatch = true, but we didn't actually copy anything.

You did well adding the `CommandResult.RepeatitivePasteAfterCut` in the allowDispatch. But the changes on the allowPaste should stay in the `handle`.

src/types/commands.ts
```diff
@@ -1169,6 +1170,7 @@ export const enum CommandResult {
   SplitWillOverwriteContent,
   NoSplitSeparatorInSelection,
   NoActiveSheet,
+  RepeatitivePasteAfterCut,
```
hokolomopo, line 1173:
```suggestion
  RepeatitedPasteAfterCut,
```

tests/menu_items_registry.test.ts
hokolomopo: We should have a test testing `paste_special_format` calling `PASTE_FROM_OS_CLIPBOARD` and another one calling `PASTE` then :slightly_smiling_face:

tests/menu_items_registry.test.ts
```diff
@@ -174,17 +174,42 @@ describe("Menu Item actions", () => {
...
+        target: target("A1"),
+        text,
+        pasteOption: undefined,
+      });
+    } else {
+      expect(dispatch).toHaveBeenCalledWith("PASTE", {
+        target: target("A1"),
+        pasteOption: undefined,
+      });
+    }
```
hokolomopo, line 210: The `if/else` have nothing to do here, you are writing a test so you are controlling whether test === content or not. As stated in my previous comment, you should have 2 separate tests : one for PASTE_FROm_OS and one for PASTE.

tests/plugins/clipboard.test.ts
hokolomopo: I get your point. The test should be in grid.ts and use the copy/paste function of the grid then. That way you mimick real life behaviour. 

Doing it here is useless, as using `pasteFromOSClipboard` will totally ignore the clipboard state. Doing the test in the grid actually test wheter you call `paste` or `pasteFromOS`, and thus whetehr you paste the internal style or not.

tests/plugins/clipboard.test.ts
```diff
@@ -190,16 +192,77 @@ describe("clipboard", () => {
...
+    expect(getCellContent(model, "C5")).toBe("");
+    expect(getStyle(model, "C5")).toEqual({});
+  });
+
+  test("cannot paste multiple times after cut", () => {
+    const model = new Model();
+    setCellContent(model, "B2", "b2");
+    expect(getCellContent(model, "B2")).toBe("b2");
+    setStyle(model, "B2", { bold: true });
+    expect(getCell(model, "B2")!.style).toEqual({ bold: true });
```
hokolomopo, line 228:
```suggestion
    setCellContent(model, "B2", "b2");
    setStyle(model, "B2", { bold: true });
```
we can trust `setCellContent/Style` to do their jobs; we don't need to test every single thing. Same in the other tests.

tests/plugins/clipboard.test.ts
```diff
@@ -190,16 +192,77 @@ describe("clipboard", () => {
...
+    cut(model, "A1");
+    setCellContent(model, "A1", "Orange", sheet1);
+
+    createSheet(model, { sheetId: sheet2 });
+    activateSheet(model, sheet2);
+    deleteSheet(model, sheet1);
+
+    paste(model, "sheet2!A2");
+    expect(getCellContent(model, "A2", sheet2)).toBe("Apple");
+    expect(getCell(model, "A2")!.style).toEqual({ bold: true });
```
hokolomopo, line 257: The clipboard still wasn't invalidated  here since the `paste` worked...

hokolomopo (commented):
Almost there ! :blush: 

FYI I'll be in vacation until the 30/05.

src/helpers/clipboard/clipboard_cells_state.ts
```diff
@@ -481,7 +481,9 @@ export class ClipboardCellsState extends ClipboardCellsAbstractState {
         .map((cells) => {
           return cells
             .map((c) =>
-              c.cell ? this.getters.getCellText(c.position, this.getters.shouldShowFormulas()) : ""
+              this.getters.shouldShowFormulas()
+                ? c.cell?.content
+                : c.evaluatedCell?.formattedValue || ""
```
hokolomopo, line 486:
```suggestion
              this.getters.shouldShowFormulas()
                ? c.cell?.content || ""
                : c.evaluatedCell?.formattedValue || ""
```

src/plugins/ui_stateful/clipboard.ts
```diff
@@ -60,11 +62,16 @@ export class ClipboardPlugin extends UIPlugin {
...
+          this.status = "invisible";
+          return CommandResult.RepeatedPasteAfterCut;
+        }
         const pasteOption = cmd.pasteOption || (this._isPaintingFormat ? "onlyFormat" : undefined);
         return this.state.isPasteAllowed(cmd.target, { pasteOption });
       case "PASTE_FROM_OS_CLIPBOARD": {
         const state = new ClipboardOsState(cmd.text, this.getters, this.dispatch, this.selection);
-        return state.isPasteAllowed(cmd.target);
+        const result = state.isPasteAllowed(cmd.target, { pasteOption: cmd.pasteOption });
+        return result;
```
hokolomopo, line 74:
```suggestion
        return state.isPasteAllowed(cmd.target, { pasteOption: cmd.pasteOption });
```
nitpick

src/plugins/ui_stateful/clipboard.ts
```diff
@@ -60,11 +62,16 @@ export class ClipboardPlugin extends UIPlugin {
         if (!this.state) {
           return CommandResult.EmptyClipboard;
         }
+        if (!this.allowPaste) {
+          this.status = "invisible";
```
hokolomopo, line 66: Is changing status here useful ?

src/plugins/ui_stateful/clipboard.ts
```diff
@@ -176,6 +188,14 @@ export class ClipboardPlugin extends UIPlugin {
         this.status = "visible";
         break;
       }
+      case "DELETE_SHEET":
+        if (this.state?.operation !== "CUT") {
+          return;
+        }
+        if (this.originSheetId === cmd.sheetId) {
+          this.allowPaste = false;
```
hokolomopo, line 196: should also make the this.status invisible I think

tests/components/grid.test.ts
```diff
@@ -1427,6 +1427,20 @@ describe("Copy paste keyboard shortcut", () => {
...
+  test("cut zone gets cleared on paste if content/style is altered after cut", async () => {
+    setCellContent(model, "A1", "things");
+    setStyle(model, "A1", { bold: true });
+    selectCell(model, "A1");
+    document.body.dispatchEvent(getClipboardEvent("cut", clipboardData));
+    setCellContent(model, "A1", "new content");
+    setStyle(model, "A1", { bold: false });
+    selectCell(model, "A2");
+    document.body.dispatchEvent(getClipboardEvent("paste", clipboardData));
+    expect(getCellContent(model, "A2")).toEqual("things");
```
hokolomopo, line 1439: A2 should contain "new content" no (for CUT) ? Here the changes we made to A1 are totally lost

tests/menu_items_registry.test.ts
```diff
@@ -146,6 +147,37 @@ describe("Menu Item actions", () => {
     expect(getCellContent(model, "A1")).toEqual("");
   });
 
+  test("'Edit -> paste' if copied from grid and content altered before paste", async () => {
+    setCellContent(model, "A1", "a1");
+    doAction(["edit", "copy"], env); // first copy from grid
+    setCellContent(model, "A1", "os clipboard");
+    selectCell(model, "C3");
+    doAction(["edit", "paste"], env);
+    await nextTick();
```
hokolomopo, line 156: nextTick() useless I think

tests/menu_items_registry.test.ts
```diff
@@ -146,6 +147,37 @@ describe("Menu Item actions", () => {
...
+
+  test("Internal copy followed by OS copy should not bring paste format from internal copy", async () => {
+    setCellContent(model, "C1", "c1");
+    setStyle(model, "C1", { fillColor: "#FA0000" });
+    selectCell(model, "C1");
+    doAction(["edit", "copy"], env); // first copy from grid
+    await env.clipboard.writeText("Then copy in OS clipboard");
+    selectCell(model, "A1");
+    doAction(["edit", "paste_special", "paste_special_format"], env);
+    await nextTick();
```
hokolomopo, line 172: nextTick() useless I think

tests/menu_items_registry.test.ts
```diff
@@ -174,17 +206,32 @@ describe("Menu Item actions", () => {
...
       text,
+      pasteOption: "onlyValue",
     });
   });
 
-  test("Edit -> paste_special -> paste_special_format", () => {
+  test("Edit -> paste_special -> paste_special_format", async () => {
+    doAction(["edit", "copy"], env);
     doAction(["edit", "paste_special", "paste_special_format"], env);
+    await nextTick();
```
hokolomopo, line 216: nextTick() useless I think

tests/menu_items_registry.test.ts
```diff
@@ -174,17 +206,32 @@ describe("Menu Item actions", () => {
...
       target: env.model.getters.getSelectedZones(),
       pasteOption: "onlyFormat",
     });
   });
 
+  test("Edit -> paste_special -> paste_special_format from OS clipboard", async () => {
+    const text = "in OS clipboard";
+    await env.clipboard.writeText(text);
+    doAction(["edit", "paste_special", "paste_special_format"], env);
+    await nextTick();
```
hokolomopo, line 227: nextTick() useless I think

tests/plugins/clipboard.test.ts
```diff
@@ -190,16 +192,54 @@ describe("clipboard", () => {
...
+    expect(getCell(model, "C2")!.style).toEqual({ bold: true });
+
+    paste(model, "E5");
+    expect(getCell(model, "E5")).toBe(undefined);
+  });
+
+  test("Clipboard should be invalidated when sheet is deleted", async () => {
+    const model = new Model();
+    const sheet1 = model.getters.getActiveSheetId();
+    const sheet2 = "sheet2";
```
hokolomopo, line 224:
```suggestion
    const sheet1Id = model.getters.getActiveSheetId();
    const sheet2Id = "sheet2";
```
nitpick

tests/plugins/clipboard.test.ts
```diff
@@ -190,16 +192,54 @@ describe("clipboard", () => {
...
+    createSheet(model, { sheetId: sheet2 });
+
+    setCellContent(model, "A1", "Apple", sheet1);
+    setStyle(model, "A1", { bold: true });
+    cut(model, "A1");
+
+    activateSheet(model, sheet2);
+    deleteSheet(model, sheet1);
+    pasteFromOSClipboard(model, "A2", "Apple");
```
hokolomopo, line 233:
```suggestion
    paste(model, "A2");
```
Should be a normal paste.

src/plugins/ui_stateful/clipboard.ts
```diff
@@ -156,10 +165,13 @@ export class ClipboardPlugin extends UIPlugin {
         break;
       }
       case "PASTE_FROM_OS_CLIPBOARD":
-        this.state = new ClipboardOsState(cmd.text, this.getters, this.dispatch, this.selection);
-        this.state.paste(cmd.target);
-        this.lastPasteState = this.state;
-        this.status = "invisible";
+        if (this.allowPaste) {
```
hokolomopo, line 168: `PASTE_FROM_OS_CLIPBOARD` should ignore `this.allowPaste`, since it won't have the correct value. Imagine a situation when we cup/pasted cometing, `this.allowPaste` is false; but then we copy something in another application and try to paste it in the spreadsheet. We don't want to block the paste.

In practice, `PASTE` will be used by grid when the clipboard content was copied from the spreadhseet, `PASTE_FROM_OS_CLIPBOARD` will be used when the clipboard content comes from somewhere else.

hokolomopo (commented):
I'm back ! Some comments :)

tests/components/grid.test.ts
khpa-odoo: If we go according to the specs of this task, which says PASTE the old values & styles made before CUT/COPY, then I guess this test does solve the purpose right?. Also, this behaviour is same as it was before this fix/task. So do you intend to keep the new values after cut instead? Or keep the old ones as it was?

tests/menu_items_registry.test.ts
khpa-odoo: It seems (referring to other test cases) after every single doAction performed, we have to add an `await nextTick()` before expecting a command dispatch for that action. I tried removing them as you suggested, but all 4 test cases failed...

hokolomopo: Yeah the problem seems that the function `paste` called by the action is async, so we indeed need the `await nextTick();`. Maybe we should change the design and alway sawait the actions, but don't worry about it then.

tests/components/grid.test.ts
hokolomopo: When we copy, we should put things in the clipboard, and then paste it as it was copied, without worrying about the changes in the sheet.

Cut/paste is a bit trickier since we delete some sheet content, and we don't want to lose information. Cut can be seen as moving cells from one place to another, so we should use the current value of the sheet.

That's the theory anyway, but looking at the current code we never really did that. I'd say you can leave it like that, and we'll do another task to handle cut correctly.

tests/plugins/clipboard.test.ts
```diff
@@ -190,16 +192,54 @@ describe("clipboard", () => {
...
+    cut(model, "B2");
+    paste(model, "C2");
+    expect(getCellContent(model, "C2")).toBe("b2");
+    expect(getCell(model, "C2")!.style).toEqual({ bold: true });
+
+    paste(model, "E5");
+    expect(getCell(model, "E5")).toBe(undefined);
+  });
+
+  test("Clipboard should be invalidated when sheet is deleted", async () => {
```
hokolomopo, line 221:
```suggestion
  test("Cut clipboard should be invalidated when sheet is deleted",  () => {
```

tests/plugins/clipboard.test.ts
```diff
@@ -190,16 +192,54 @@ describe("clipboard", () => {
     expect(getCell(model, "C2")!.style).toEqual({ bold: true });
   });
 
+  test("copying external content & paste-format on a cell will not paste content", async () => {
```
hokolomopo, line 195:
```suggestion
  test("copying external content & paste-format on a cell will not paste content",  () => {
```

hokolomopo (commented):
Some last nitpick, I'll green the task for the next stage :)

tests/menu_items_registry.test.ts
hokolomopo: Fixed it in https://github.com/odoo/o-spreadsheet/pull/2522, you should remove the nextTick and await the `doAction` when it's merged and you rebase your branch :)

khpa-odoo: Cool :)

tests/components/grid.test.ts
khpa-odoo: Okay, I'll leave it as it is then.

Suggestion: We can try to replicate the cut/paste behaviour as in gsheet in the new task.

src/helpers/clipboard/clipboard_cells_state.ts
```diff
@@ -481,7 +481,9 @@ export class ClipboardCellsState extends ClipboardCellsAbstractState {
         .map((cells) => {
           return cells
             .map((c) =>
-              c.cell ? this.getters.getCellText(c.position, this.getters.shouldShowFormulas()) : ""
+              this.getters.shouldShowFormulas()
```
rrahir, line 484:
```suggestion
              this.getters.shouldShowFormulas() && cell?.isFormula
``` 
you should preserve the same logic as `getCellText`. With the current suggestion, if I were to **CUT** a cell with `{content: 1, format: "mm/dd/yyyy"}`  while `this.getters.shouldShowFormulas() === true` I would end up with  `1` in my clipboard, whereas other spreadsheets would yield `12/31/1899`.

I think a test would be welcome here to make sure we don't break tahat behaviour in the future :)

src/plugins/ui_stateful/clipboard.ts
```diff
@@ -43,8 +44,9 @@ export class ClipboardPlugin extends UIPlugin {
   private status: "visible" | "invisible" = "invisible";
   private state?: ClipboardState;
   private lastPasteState?: ClipboardState;
-
+  private allowPaste: boolean = true;
```
rrahir, line 47: After discussing this issue of infinite with the team and PO, we think it would be best (and overall simpler) to empty the OS clipboard as long as the standard clipboard after the **paste** of a **cut** command.  While it is not what google sheet does, i does not make much sense to keep pasting the content of a cut - an action that only makes sense inside the spreadsheet itself.

rrahir (changes requested):
Overall, nice job :+1: 
See my two comments for some modifications I'd like you to add.

On another topic, when you have  several bugs to fix like you do here, please make a dedicated commit for each fix. 2 perks to this:
- it's way easier to review - here I had to guess which change did what in your PR
- there is no loss of information - the commit title here says nothing about the "onlyFormat" fix which has nothing to do with the other fix you are working on

It would be nice if you could split the two changes, (with their respective tests) in separated commits.

src/plugins/ui_stateful/clipboard.ts
khpa-odoo: Hey, I did actually try emptying the clipboard earlier by making its state undefined if operation is 'cut'. But in that case, call would go to paste from os clipboard and there it would be difficult to distinguish whether the content in clipboard is from outside of the sheet or it is from spreadsheet itself.
I've tried multiple approaches to solve this issue but nothing seems to be effective so far. Do you have any suggestions on how this can be achieved?

hokolomopo: The idea would be to also clear the **OS** Clipboard after a paste from cut.

You cannot do that from the plugin tho, you have to do that from the grid/menu item. Use `env.clipboard.write` to write stuff from the OS clipboard.

khpa-odoo:
Hey @rrahir! I've pushed the changes. Do check out once :)

src/plugins/ui_stateful/clipboard.ts
rrahir: @khpa-odoo I don't think you have addressed this point. the idea would be to completely get rid of the 'allowPaste' boolean that you added and just assume that we empty the clipboard (both the one of the plugin and of the os once we successfully paste after a cut.  In order to do that,you should edit the `src/actions/menu_items_actions.ts  function paste()` to clear the OS clipboard one the paste action was successful in the plugin, but only if we come from a CUT operation, for this you have the getter `isCutOperation`

src/helpers/clipboard/clipboard_cells_state.ts
rrahir: I don't think I see a test for this

tests/components/grid.test.ts
```diff
@@ -1427,6 +1428,34 @@ describe("Copy paste keyboard shortcut", () => {
...
+    expect(getStyle(model, "A2")).toEqual({ bold: true });
+    expect(getCell(model, "A1")).toBe(undefined);
+  });
+
+  test("Cut and paste of a date format cell, by toggling formula visibility should return date", async () => {
+    model.dispatch("SET_FORMULA_VISIBILITY", { show: true });
+    setCellContent(model, "A1", "1");
+    setCellFormat(model, "A1", "m/d/yyyy");
+    document.body.dispatchEvent(getClipboardEvent("cut", clipboardData));
+    model.dispatch("SET_FORMULA_VISIBILITY", { show: false });
```
rrahir, line 1450: you can definitely check the content of clipboardDatza, which is what we really are interested in

tests/components/grid.test.ts
```diff
@@ -1427,6 +1428,34 @@ describe("Copy paste keyboard shortcut", () => {
...
+    setCellContent(model, "A1", "new content");
+    setStyle(model, "A1", { bold: false });
+    selectCell(model, "A2");
+    document.body.dispatchEvent(getClipboardEvent("paste", clipboardData));
+    expect(getCellContent(model, "A2")).toEqual("things");
+    expect(getStyle(model, "A2")).toEqual({ bold: true });
+    expect(getCell(model, "A1")).toBe(undefined);
+  });
+
+  test("Cut and paste of a date format cell, by toggling formula visibility should return date", async () => {
```
rrahir, line 1445: it is a broader behaviour than this, I just provided one example. if you look at the code it returns the formattedvalue as long as 1) the cell is not a formula, 2) we don't set 'formulaVisibility' to true. The test should be about 
1) 'cut a non formula cell returns its formatted value'
2) and another test 'cut a formula cell returns the content if `showFormulas === true` where you cut paste with showformulas with true and false values. 
The second test (or equivalent) might already exist though, I have not read the whole file.

khpa-odoo: Pushed the changes :+1: 
Do checkout the previous comment as well.

rrahir (commented):
last small comment, would be nice if you could add a test as well :)

khpa-odoo:
> last small comment, would be nice if you could add a test as well :)

We do have a test case for 'onlyFormat' in case of `PASTE_FROM_OS_CLIPBOARD` in `menu_item_registry.test.ts`. Do you want me to edit it? Or write a new one for something else? Let me know :)

rrahir:
> > last small comment, would be nice if you could add a test as well :)
> 
> We do have a test case for 'onlyFormat' in case of `PASTE_FROM_OS_CLIPBOARD` in `menu_item_registry.test.ts`. Do you want me to edit it? Or write a new one for something else? Let me know :)

I think another test would be nice:)

khpa-odoo:
> I think another test would be nice:)
Added a new one :+1: 
Check it out. Thanks :)

rrahir (approved):
cheers :+1: 
@robodoo r+

rrahir:
@robodoo rebase-ff

PR #1771, opened 2022-10-28
[FIX] composer: Pasting/editing fixes

rrahir:
When copying in the composer, the paste event was still propagated to the parent (i.e. `Spreadsheet`) which would handle it and dispath either `PASTE` or `PASTE_FROM_OS_CLIPBOARD` basically writing the wrong content on the cell on top of endin the edition and closing the composer.

Note: couldn't manage to properly mock the paste event on the contenteditable div. The test at least ensure taht we stop the propagation of the paste event.

Part of 2863362

## Description:

description of this task, what is implemented and why it is implemented that way.

Odoo task ID : [2863362](https://www.odoo.com/web#id=2863362&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] feature is organized in plugin, or UI components
- [ ] support of duplicate sheet (deep copy)
- [ ] in model/core: ranges are Range object, and can be adapted (adaptRanges)
- [ ] in model/UI: ranges are strings (to show the user)
- [ ] undo-able commands (uses this.history.update)
- [ ] multiuser-able commands (has inverse commands and transformations where needed)
- [ ] new/updated/removed commands are documented
- [ ] exportable in excel
- [ ] translations (\_lt("qmsdf %s", abc))
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes
- [ ] doc is rebuild (npm run doc)
- [ ] status is correct in Odoo

PR #2475, opened 2023-05-12
[IMP] composer: double click on range will select the whole range

yunchen4:
## Description:

Previously, when double clicking on a range in a formula, for example `A1:A30`, only `A1` or `A30` will be selected because of the colon.

This PR improves the selection so `A1:A30` the whole range will be selected when double clicking.

Normal text including colon will not be impacted. Double clicking `A1:A30` won't select the whole string.

Odoo task ID : [3266412](https://www.odoo.com/web#id=3266412&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/composer/composer/composer.ts
+++ b/src/components/composer/composer/composer.ts
@@ -416,6 +416,37 @@ export class Composer extends Component<ComposerProps, SpreadsheetChildEnv> {
     this.processTokenAtCursor();
   }
 
+  onDblClick() {
+    if (this.env.model.getters.isReadonly()) {
+      return;
+    }
+    const composerContent = this.env.model.getters.getCurrentContent();
+    const isValidFormula = composerContent.startsWith("=");
+
+    if (isValidFormula) {
+      const tokens = this.env.model.getters.getCurrentTokens();
+      const currentSelection = this.contentHelper.getCurrentSelection();
+      if (currentSelection.start === currentSelection.end) return;
+
+      const currentSelectedText = composerContent.substring(
+        currentSelection.start,
+        currentSelection.end
+      );
+      const token = tokens.filter(
+        (token) =>
+          token.value.includes(currentSelectedText) &&
+          token.start <= currentSelection.start &&
+          token.end >= currentSelection.end
+      )[0];
+      if (token.type === "REFERENCE") {
+        this.env.model.dispatch("CHANGE_COMPOSER_CURSOR_SELECTION", {
+          start: token.start,
+          end: token.end,
+        });
+      }
+    }
+  }
+
   // ---------------------------------------------------------------------------
   // Private
   // ---------------------------------------------------------------------------
@@ -448,8 +479,7 @@ export class Composer extends Component<ComposerProps, SpreadsheetChildEnv> {
    */
   private getContentLines(): HtmlContent[][] {
     let value = this.env.model.getters.getCurrentContent();
-    const isValidFormula =
-      value.startsWith("=") && this.env.model.getters.getCurrentTokens().length > 0;
+    const isValidFormula = value.startsWith("=");
 
     if (value === "") {
       return [];
--- a/src/components/composer/composer/composer.xml
+++ b/src/components/composer/composer/composer.xml
@@ -18,6 +18,7 @@
         t-on-paste.stop=""
         t-on-compositionstart="onCompositionStart"
         t-on-compositionend="onCompositionEnd"
+        t-on-dblclick="onDblClick"
       />
 
       <div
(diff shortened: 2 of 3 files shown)


src/components/composer/composer/composer.ts
```diff
@@ -447,6 +447,29 @@ export class Composer extends Component<ComposerProps, SpreadsheetChildEnv> {
     this.processTokenAtCursor();
   }
 
+  onDblClick() {
+    const composerContent = this.env.model.getters.getCurrentContent();
```
anhe-odoo, line 451: Dont we need the same thing as in the onClick() ?

```
if (this.env.model.getters.isReadonly()) {
  return;
}
```

anhe-odoo:
Do you think there will be a way to avoid passing in `onClick` when calling the `onDblClick` event ? Currently, when we double click, we first select only the clicked part and then the selection is updated to get the full range (I mean these two steps are visible). Do you think there could be a way to directly select the full range instead ?

yunchen4:
@anhe-odoo For now I don't have a better way to do it. I tried to bypass `onClick` using `setTimeout`, or recording the last timestamp when clicking is triggered and comparing the difference, but they all have side-effects (e.g. cannot have correct focus on composer). I notice Google sheets also has the same problem, so IMO it's not an easy fix.

src/components/composer/composer/composer.ts
```diff
@@ -447,6 +447,32 @@ export class Composer extends Component<ComposerProps, SpreadsheetChildEnv> {
...
+  onDblClick() {
+    if (this.env.model.getters.isReadonly()) {
+      return;
+    }
+    const composerContent = this.env.model.getters.getCurrentContent();
+    const isValidFormula =
+      composerContent.startsWith("=") && this.env.model.getters.getCurrentTokens().length > 0;
+
+    if (isValidFormula) {
+      const tokens = this.getColoredTokens();
```
LucasLefevre, line 459: can't you use `getters.getCurrentTokens()` instead of `getColoredTokens` here ?
It looks weird since this it has nothing to do with colors. (even though `getColoredTokens` is wrongly named in the first place since it doesn't return tokens)

src/components/composer/composer/composer.ts
```diff
@@ -400,6 +400,35 @@ export class Composer extends Component<ComposerProps, SpreadsheetChildEnv> {
...
+    if (isValidFormula) {
+      const tokens = this.env.model.getters.getCurrentTokens();
+      const currentSelection = this.contentHelper.getCurrentSelection();
+      if (currentSelection.start === currentSelection.end) return;
+
+      const currentSelectedText = composerContent.substring(
+        currentSelection.start,
+        currentSelection.end
+      );
+      const token = tokens.filter((line) => line.value.includes(currentSelectedText))[0];
```
LucasLefevre, line 420: Looking for the first token which includes the selected text doesn't work properly when several tokens contains the same text.
E.g. `=A1+A1` and double click the second "A1" will actually select the first one.
You can probably play around with `token.start` and `token.end`

src/components/composer/composer/composer.ts
```diff
@@ -400,6 +400,38 @@ export class Composer extends Component<ComposerProps, SpreadsheetChildEnv> {
     this.processTokenAtCursor();
   }
 
+  onDblClick() {
+    if (this.env.model.getters.isReadonly()) {
+      return;
+    }
+    const composerContent = this.env.model.getters.getCurrentContent();
+    const isValidFormula =
+      composerContent.startsWith("=") && this.env.model.getters.getCurrentTokens().length > 0;
```
rrahir, line 409: there is something I don't understand here: if the content starts with "=" then getCurrentTokens.length will always be >0 since we have the '=' token. the only other way to have tokens empty would be (to my knowledge) to have the tokenize crash but if it crashed then this bit of code should not work either. So I don't realy understand what it does... @hokolomopo  I think you introduced this condition with the multiline, do you have a clue ?

hokolomopo: I don't remember at all this line, and the git blame points on you on some composer fix https://github.com/odoo/o-spreadsheet/pull/1771/files 😛 

But I'd say that you're right and that the second condition is useless ? I don't see how we could have a content and no tokens...
image

rrahir: damn it, I haven't checked properly, sorry :/
Anyway, I think we can get rid of that condition...

yunchen4: I deleted them 🙂

rrahir:
@robodoo r+

PR #2491, opened 2023-05-18
[FIX] viewport: warning message on resized viewport size

dhrp-odoo:
## Description:

Fix warning not closing issue caused by changing the value of `DEFAULT_SHEETVIEW_SIZE` to 0 from 1000 in a previous 
PR #2188. The problem arose due to `xRatio` and `yRatio` becoming `NaN` and `Infinity` prior to mounting.

To resolve this issue, Added an if condition that returns the function if either xRatio or yRatio is Infinity.

Odoo task ID : [3318806](https://www.odoo.com/web#id=3318806&cids=2&menu_id=4720&action=333&active_id=2328&model=project.task&view_type=form)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/spreadsheet/spreadsheet.ts
+++ b/src/components/spreadsheet/spreadsheet.ts
@@ -264,6 +264,12 @@ export class Spreadsheet extends Component<SpreadsheetProps, SpreadsheetChildEnv
     const { xRatio, yRatio } = this.env.model.getters.getFrozenSheetViewRatio(
       this.env.model.getters.getActiveSheetId()
     );
+
+    if (!isFinite(xRatio) || !isFinite(yRatio)) {
+      // before mounting, the ratios can be NaN or Infinity if the viewport size is 0
+      return;
+    }
+
     if (yRatio > MAXIMAL_FREEZABLE_RATIO || xRatio > MAXIMAL_FREEZABLE_RATIO) {
       if (this.isViewportTooSmall) {
         return;
--- a/tests/components/spreadsheet.test.ts
+++ b/tests/components/spreadsheet.test.ts
@@ -1,6 +1,7 @@
-import { Model } from "../../src";
+import { Model, setDefaultSheetViewSize } from "../../src";
 import { OPEN_CF_SIDEPANEL_ACTION } from "../../src/actions/menu_items_actions";
 import { Spreadsheet } from "../../src/components";
+import { getDefaultSheetViewSize } from "../../src/constants";
 import { functionRegistry } from "../../src/functions";
 import { toZone } from "../../src/helpers";
 import { SpreadsheetChildEnv } from "../../src/types";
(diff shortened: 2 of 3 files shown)


tests/components/spreadsheet.test.ts
```diff
@@ -256,6 +256,16 @@ test("Spreadsheet detects frozen panes that exceed the limit size at start", asy
   expect(notifyUser).toHaveBeenCalled();
 });
 
+test("Warns user when viewport is too small for frozen panes but stops warning after resizing/Unmounted", async () => {
+  setDefaultSheetViewSize(0);
```
LucasLefevre, line 260: I would add a comment here stating this is supposed to represent the "real life" where the default size is 0

tests/components/spreadsheet.test.ts
```diff
@@ -256,6 +256,16 @@ test("Spreadsheet detects frozen panes that exceed the limit size at start", asy
...
+  const notifyUser = jest.fn();
+  const model = new Model({ sheets: [{ panes: { xSplit: 0, ySplit: 20 } }] });
+  ({ parent, fixture } = await mountSpreadsheet({ model }, { notifyUser }));
+  expect(notifyUser).toHaveBeenCalledTimes(0);
+
+  setDefaultSheetViewSize(1000); // reset the default view size
```
LucasLefevre, line 266: It would be more robust to save the value before changing it to 0, instead of hard-coding 1000. That way if the value one day changes, this test won't risk setting back a wrong value.
You can use `getDefaultSheetViewSize()`

src/components/spreadsheet/spreadsheet.ts
```diff
@@ -264,6 +264,9 @@ export class Spreadsheet extends Component<SpreadsheetProps, SpreadsheetChildEnv
     const { xRatio, yRatio } = this.env.model.getters.getFrozenSheetViewRatio(
       this.env.model.getters.getActiveSheetId()
     );
+
+    if (!isFinite(xRatio) || !isFinite(yRatio)) return; // Returning NaN and Infinity before mounted causing warning does not disappear
```
LucasLefevre, line 268: minor style detail: we tend to avoid one-line ifs. It's sometimes harder to read and spot the `return`.
I would also slightly rephrase the comment
```suggestion
    if (!isFinite(xRatio) || !isFinite(yRatio)) {
      // before mounting, the ratios can be NaN or Infinity if the viewport size is 0
      return;
    }
```

LucasLefevre (commented):
only details, thanks for the changes

dhrp-odoo:
Hi @LucasLefevre, I have made the requested changes.

Thanks :)

LucasLefevre (approved):
thanks
robodoo r+

LucasLefevre (commented):
robodoo r+ (changed the commit title prefix from "[FIX] Spreadsheet:" to "[FIX] viewport:")

tests/setup/resize_observer.mock.ts
```diff
@@ -6,7 +6,7 @@ class MockResizeObserver {
   observe() {
     //@ts-ignore
     global.resizers.add(this);
-    this.cb();
+    Promise.resolve().then(() => this.cb());
```
rrahir, line 9: @LucasLefevre  @dhrp-odoo  the tests seem to pass without it but become problematic in future situations. Do you recall why you had to introducs this?

dhrp-odoo: Sorry for the delay; it's been a public holiday week for us.

As specified in the task specs, The issue stemmed from `getFrozenSheetViewRatio` returning NaN and Infinity before mounting, leading to a warning message persisting even after a reload.

To test this behavior, The change was aimed at ensuring a more synchronous behavior in the tests to align with real function behavior.

I got your point; presently, the test passes even without this alteration.
Have there been any changes to related APIs that might have influenced this observed shift in test outcomes?

PR #2498, opened 2023-05-22
[IMP] ColorPicker: Extend the custom color selection

rrahir:
## Description:

The current color picker does not allow to visually choose a custom
color accross the whole web-compliant color spectrum. More specifically,
we could not change the saturation of the color, its level of grays [1].

This commit addresses this issue by changing the helper to use the HSL
color representation. This was based on the visual of Google sheet
color picker. As a side effet, it also addresses the magnifying glass
issue reported.

Small caveat of this approach is that there is no direct equivalence
between the HSL and RBG/HEX representations. As we store the color hex
value, the conversion from hex to hsl and vice versa might sometimes
differ slightly. However, the end user should not be able to notice it
and still has the possibility to introduce a specific hex color at will.

[1] Users could already do it by specifying a hex code directly but it
was not feasible with the visual helper.

Task: 3272474

Odoo task ID : [3272474](https://www.odoo.com/web#id=3272474&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/color_picker/color_picker.ts
+++ b/src/components/color_picker/color_picker.ts
@@ -1,53 +1,64 @@
 import { Component, useState } from "@odoo/owl";
 import {
   COLOR_PICKER_DEFAULTS,
+  ICON_EDGE_LENGTH,
   MENU_SEPARATOR_BORDER_WIDTH,
   MENU_SEPARATOR_PADDING,
   SEPARATOR_COLOR,
 } from "../../constants";
-import { hslaToRGBA, isColorValid, isSameColor, rgbaToHex, toHex } from "../../helpers";
+import {
+  clip,
+  hexToHSLA,
+  hslaToHex,
+  isColorValid,
+  isHSLAValid,
+  isSameColor,
+  toHex,
+} from "../../helpers";
 import { chartFontColor } from "../../helpers/figures/charts";
-import { Color, Pixel, Rect } from "../../types";
+import { Color, HSLA, Pixel, PixelPosition, Rect } from "../../types";
 import { SpreadsheetChildEnv } from "../../types/env";
 import { css, cssPropertiesToCss } from "../helpers/css";
+import { startDnd } from "../helpers/drag_and_drop";
 import { Popover, PopoverProps } from "../popover/popover";
 
-const PICKER_PADDING = 6;
-
 const LINE_VERTICAL_PADDING = 1;
-const LINE_HORIZONTAL_PADDING = 6;
 
-const ITEM_HORIZONTAL_MARGIN = 1;
-const ITEM_EDGE_LENGTH = 18;
+const PICKER_PADDING = 8;
 const ITEM_BORDER_WIDTH = 1;
-
+const ITEM_EDGE_LENGTH = 18;
 const ITEMS_PER_LINE = 10;
-const PICKER_WIDTH =
-  ITEMS_PER_LINE * (ITEM_EDGE_LENGTH + ITEM_HORIZONTAL_MARGIN * 2 + 2 * ITEM_BORDER_WIDTH) +
-  2 * LINE_HORIZONTAL_PADDING;
+const MAGNIFIER_EDGE = 16;
+const ITEM_GAP = 2;
 
-const GRADIENT_WIDTH = PICKER_WIDTH - 2 * LINE_HORIZONTAL_PADDING - 2 * ITEM_BORDER_WIDTH;
-const GRADIENT_HEIGHT = PICKER_WIDTH - 50;
+const CONTENT_WIDTH =
+  ITEMS_PER_LINE * (ITEM_EDGE_LENGTH + 2 * ITEM_BORDER_WIDTH) + (ITEMS_PER_LINE - 1) * ITEM_GAP;
+
+const INNER_GRADIENT_WIDTH = CONTENT_WIDTH - 2 * ITEM_BORDER_WIDTH;
+const INNER_GRADIENT_HEIGHT = CONTENT_WIDTH - 30 - 2 * ITEM_BORDER_WIDTH;
+
+const CONTAINER_WIDTH = CONTENT_WIDTH + 2 * PICKER_PADDING;
 
 css/* scss */ `
   .o-color-picker {
-    padding: ${PICKER_PADDING}px 0px;
+    padding: ${PICKER_PADDING}px 0;
+    /** FIXME: this is useless, overiden by the popover container */
     box-shadow: 1px 2px 5px 2px rgba(51, 51, 51, 0.15);
     background-color: white;
     line-height: 1.2;
     overflow-y: auto;
     overflow-x: hidden;
-    width: ${GRADIENT_WIDTH + 2 * PICKER_PADDING}px;
+    width: ${CONTAINER_WIDTH}px;
 
     .o-color-picker-section-name {
-      margin: 0px ${ITEM_HORIZONTAL_MARGIN}px;
-      padding: 4px ${LINE_HORIZONTAL_PADDING}px;
+      margin: 0px ${ITEM_BORDER_WIDTH}px;
+      padding: 4px ${PICKER_PADDING}px;
     }
     .colors-grid {
       display: grid;
-      padding: ${LINE_VERTICAL_PADDING}px ${LINE_HORIZONTAL_PADDING}px;
+      padding: ${LINE_VERTICAL_PADDING}px ${PICKER_PADDING}px;
       grid-template-columns: repeat(${ITEMS_PER_LINE}, 1fr);
-      grid-gap: ${ITEM_HORIZONTAL_MARGIN * 2}px;
+      grid-gap: ${ITEM_GAP}px;
     }
     .o-color-picker-toggler-button {
       display: flex;
@@ -77,16 +88,16 @@ css/* scss */ `
       }
     }
     .o-buttons {
-      padding: 6px;
+      padding: ${PICKER_PADDING}px;
       display: flex;
       .o-cancel {
-        margin: 0px ${ITEM_HORIZONTAL_MARGIN}px;
         border: ${ITEM_BORDER_WIDTH}px solid #c0c0c0;
         width: 100%;
         padding: 5px;
         font-size: 14px;
         background: white;
         border-radius: 4px;
+        box-sizing: border-box;
         &:hover:enabled {
           background-color: rgba(0, 0, 0, 0.08);
         }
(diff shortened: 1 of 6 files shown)


src/components/color_picker/color_picker.ts
```diff
@@ -106,55 +118,92 @@ css/* scss */ `
...
         border: ${ITEM_BORDER_WIDTH}px solid #c0c0c0;
-        width: ${GRADIENT_WIDTH}px;
+        box-sizing: border-box;
+        width: 100%;
         height: ${GRADIENT_HEIGHT}px;
-        &:hover {
-          cursor: crosshair;
+        position: relative;
+      }
+      /** TODORAR: add constants for this shit */
```
anhe-odoo, line 132: Should this TODO be adressed in this PR ?

src/components/color_picker/color_picker.ts
```diff
@@ -246,50 +319,99 @@ export class ColorPicker extends Component<ColorPickerProps, SpreadsheetChildEnv
     this.props.onColorPicked("");
   }
 
-  setCustomColor(ev: Event) {
-    if (!isColorValid(this.state.currentColor)) {
+  AddCustomColor(ev: Event) {
```
anhe-odoo, line 322: `addCustomColor` ? (lowercase "a")

src/components/color_picker/color_picker.ts
```diff
@@ -246,50 +319,99 @@ export class ColorPicker extends Component<ColorPickerProps, SpreadsheetChildEnv
...
+    if (!isHSLAValid(this.state.currentHslaColor) || !isColorValid(this.state.customHexColor)) {
       ev.stopPropagation();
       this.state.isCurrentColorInvalid = true;
       return;
     }
-    const color = toHex(this.state.currentColor);
+    ev.stopPropagation();
```
anhe-odoo, line 328: I'm not sure it's really needed, but the `ev.stopPropagation` can be put before the if to avoid writing it twice in both cases

src/components/color_picker/color_picker.ts
```diff
@@ -246,50 +319,99 @@ export class ColorPicker extends Component<ColorPickerProps, SpreadsheetChildEnv
...
+    // TODORAR: does not work ,we probably should use the magnifier logic instead
+    // while hiding the pointer
+    // copy what goes on in figure drag & drop
+    const onMouseMove = (ev: MouseEvent) => {
+      const currentMousePosition = { x: ev.clientX, y: ev.clientY };
+      const deltaX = currentMousePosition.x - initialMousePosition.x;
+      const deltaY = currentMousePosition.y - initialMousePosition.y;
+
+      // TODORAR too complicated
+      const currentGradientCoordinates = {
```
anhe-odoo, line 375: Maybe defining something like 
```
const fullBorderSize = ITEM_BORDER_WIDTH * 2 + 2;
const maximalCoordinates = {
    x: GRADIENT_WIDTH - fullBorderSize,
    y: GRADIENT_HEIGHT - fullBorderSize
};
```
defining outside of the onMouseMove, to be able to do something like 
`x : clip(initialGradientCoordinates.x + deltaX, 0, maximalCoordinates.x)` ?

But once again I'm not sure it will help

src/helpers/color.ts
```diff
@@ -274,6 +283,26 @@ export function rgbaToHSLA(rgba: RGBA): HSLA {
...
-  return isColorValid(color1) && isColorValid(color2) && toHex(color1) === toHex(color2);
+export function hslaToHex(hsla: HSLA): Color {
+  return rgbaToHex(hslaToRGBA(hsla));
+}
+
+export function hexToHSLA(hex: Color): HSLA {
+  return rgbaToHSLA(colorToRGBA(hex));
+}
+
+export function isSameColor(color1: Color, color2: Color, tolerance: number = 0): boolean {
```
anhe-odoo, line 294: It may be overkill here, but there are some "scientifically validated" way to compute the distance between two colors as well as the tolerance to consider two colors to be identical : 

https://en.wikipedia.org/wiki/Color_difference#Tolerance

tests/components/color_picker.test.ts
```diff
@@ -120,7 +122,7 @@ describe("Color Picker buttons", () => {
     await simulateClick(".o-color-picker-toggler");
     setInputValueAndTrigger(".o-custom-input-preview input", "this is not a color", "input");
     await nextTick();
-    expect(fixture.querySelector(".o-wrong-color")).toBeNull();
+    expect(fixture.querySelector(".o-wrong-color")).not.toBeNull();
     await simulateClick(".o-add-button");
     expect(fixture.querySelector(".o-wrong-color")).not.toBeNull();
```
anhe-odoo, line 127: I'm not sure the click and the "add button" and the testing part after is still needed now :)

src/components/color_picker/color_picker.ts
```diff
@@ -232,64 +255,139 @@ export class ColorPicker extends Component<ColorPickerProps, SpreadsheetChildEnv
...
-    this.state.style.background = computeCustomColor(ev);
-    const shiftFromCursor = 10;
-    this.state.style.left = `${ev.offsetX + shiftFromCursor}px`;
-    this.state.style.top = `${ev.offsetY + shiftFromCursor}px`;
+  setHexColor() {
+    const val = (this.hexInputRef.el as HTMLInputElement)?.value;
+    this.state.customHexColor = val;
+    this.state.isCurrentColorInvalid = false;
+    if (!isColorValid(val)) {
+      this.state.isCurrentColorInvalid = true;
```
LucasLefevre, line 376: does `isCurrentColorInvalid` needs to be in a state ? Can't it be computed ?
Also, I'd rename it to `isHexColorInvalid`

src/components/color_picker/color_picker.ts
```diff
@@ -232,64 +255,139 @@ export class ColorPicker extends Component<ColorPickerProps, SpreadsheetChildEnv
...
-    const { display, background, left, top } = this.state.style;
-    return cssPropertiesToCss({
-      display,
-      "background-color": display === "block" ? background : undefined,
-      left,
-      top,
-    });
+  addCustomColor(ev: Event) {
+    if (!isHSLAValid(this.state.currentHslaColor) || !isColorValid(this.state.customHexColor)) {
+      ev.stopPropagation();
```
LucasLefevre, line 384: then we can have `.stop` in the template if we stop the event in all cases

src/components/color_picker/color_picker.xml
```diff
@@ -41,38 +41,47 @@
             t-attf-style="background-color:{{color}};"
             t-on-click="() => this.onColorClick(color)">
             <div
-              t-if="props.currentColor === color"
+              t-if="isSameColor(props.currentColor, color)"
```
LucasLefevre, line 44: :eyes: I thought this was already fixed

src/components/color_picker/color_picker.xml
```diff
@@ -41,38 +41,47 @@
...
+              t-ref="hexInput"
+              t-att-value="state.customHexColor"
+              t-on-input="setHexColor"
             />
-            <div class="o-color-preview" t-attf-style="background-color:{{state.currentColor}}"/>
+            <div class="o-color-preview" t-att-style="colorPreviewStyle"/>
           </div>
           <div class="o-custom-input-buttons">
-            <button class="o-add-button" t-on-click="setCustomColor">Add</button>
+            <button class="o-add-button" t-on-click.stop="addCustomColor">Add</button>
```
LucasLefevre, line 79: Oh ! it in already in the template ! :smile: 
`ev.stopPropagation();` in the handler can be removed then :)

src/helpers/color.ts
LucasLefevre: Overkill indeed I think.
We do not build a drawing software ^^

LucasLefevre (commented):
Nice :)

src/helpers/color.ts
rrahir: yup, we discussed this with Anthony afterwards and since the idea here is to have a 'conversion' tolerance rather than a 'human perception' tolerance, the  distance seems ok.

src/components/color_picker/color_picker.ts
rrahir: :facepalm:

src/components/color_picker/color_picker.ts
rrahir: I changed the variable definitions and it's now a bit better :)

src/components/color_picker/color_picker.xml
rrahir: same here, I recall the forwardport of the "+" sign was a bit messy, it might be a relicate

src/components/color_picker/color_picker.ts
rrahir: > does isCurrentColorInvalid needs to be in a state ? Can't it be computed ?
the state is not necessary indeed, i'll remove it


> Also, I'd rename it to `isHexColorInvalid`

The user can technically write `rgb(1,1,1)` and it will be converted via `toHex` helper, so it's kind of a type `Color`  in a large sense. 
I think the helper does this for historical reasons where we did not "force" hex representation for the colors - which we still don't really do. But I think we could explore that possibility in the future, it could simplify some of the color helpers that we have.

LucasLefevre:
Thanks :)

> note that I had to adapt one behaviour

looks reasonable to me :) especially given a user can't really get into that situation himself (I think)

Looks ready to r+ when the commits are squashed :)

LucasLefevre:
Oh ! I just remember a strange behavior when testing yesterday.
I you manually add the alpha channel to `0` in the hex input (e.g. `#523C3C00`), then the color will always be white, no matter what you do with the color gradient/saturation.
What do you think of this behavior ? I think it's confusing and one might wonder why the f*ck the color stays white. Should we strip the alpha part?

LucasLefevre:
> note that I had to adapt one behaviour

We could also let `customHexColor` be an empty string if the initial color is invalid, and say an empty string is valid when computing if it's valid or nor. Then maybe it should be renamed `isHexColorInputValid` since it wouldn't be the color per-se which is valid (and an empty string isn't a valid a color), but the input value

rrahir:
> > note that I had to adapt one behaviour
> 
> We could also let `customHexColor` be an empty string if the initial color is invalid, and say an empty string is valid when computing if it's valid or nor. Then maybe it should be renamed `isHexColorInputValid` since it wouldn't be the color per-se which is valid, but the input value (an empty string isn't a valid a color, but would be a valid input value)

okay yes, i'll just disable the add button when the input is empty then, otherwise, the user could empty the input then just spam "add" and nothing would happen

rrahir:
@LucasLefevre  new fixround commit - I addressed the different points we discussed

LucasLefevre (approved):

LucasLefevre (approved):
robodoo r+

PR #2523, opened 2023-06-01
[FIX] composer: do not autocomplete boolean values

somu-odoo:
## Description:

Currently fuzzy lookup is showing random results for true or false values, which is not correct. This commit fixes that issue by hard coding the lookup values for true and false for now.

Future scope for same:
Create True and False functions, which upon autocomplete will return true and false respectively without any arguments required. May need to change a bit of tokenizer

Odoo task ID : [3213850](https://www.odoo.com/web#id=3213850&cids=2&menu_id=4720&action=333&active_id=2328&model=project.task&view_type=form)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/composer/composer/composer.ts
+++ b/src/components/composer/composer/composer.ts
@@ -418,7 +418,11 @@ export class Composer extends Component<ComposerProps, SpreadsheetChildEnv> {
         };
       });
     if (searchTerm) {
-      values = fuzzyLookup(searchTerm, values, (t) => t.text);
+      if (searchTerm.toLowerCase() === "true" || searchTerm.toLowerCase() === "false") {
+        values = [];
+      } else {
+        values = fuzzyLookup(searchTerm, values, (t) => t.text);
+      }
     } else {
       // alphabetical order
       values = values.sort((a, b) => a.text.localeCompare(b.text));
--- a/tests/components/autocomplete_dropdown.test.ts
+++ b/tests/components/autocomplete_dropdown.test.ts
@@ -309,6 +309,16 @@ describe("Functions autocomplete", () => {
       ).toBe("SUM");
     });
   });
+
+  describe("Edge cases with autocomplete", () => {
+    test("autocomplete dropdown should not trigger on =true or =false", async () => {
+      restoreDefaultFunctions();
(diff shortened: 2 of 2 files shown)


LucasLefevre:
Even though I suggested playing with the fuzzy score as a first idea, it looks like it's not going to work by looking at the code here.
I agree with @hokolomopo 's suggestion(s)

Suggestion 1 is definitely something you can do.
If you feel like giving a try at suggestion 2, you're welcome to try. But it's probably not so easy
If you don't feel like it, no problem at all, we'll do it later :)

somu-odoo:
I was aware of the fuzzy lookup concept, but I was moving with the approach to deal with score which after both the comments I'm sure is a wrong approach. I will give suggestion 2 a try, if it doesn't work then will implement suggestion 1.

LucasLefevre:
ok, don't spend too much time on it. If you see it's gonna result in something too hairy, it'll be for later.
I would do Suggestion 1, in any case. As Adrien pointed out, we can do both

somu-odoo:
Hey @LucasLefevre I tried implementing second method, but modifying tokenizer seems to be bit tricky and may take some time, for now I have removed my previous code, reverted the changes on test cases, and added a hard coded line for true & false along with one test case. Let me know if this works or if it needs any changes.

hokolomopo:
It make no sense to have the `if(serachedTerm === "true")` in the helper ... The helper isn't specific to the composer, does it makes any sense that fuzzy lookup don't work for "true" in data filters?

The condition should be before the call to the fuzzy lookup helper...

LucasLefevre (commented):
Hi,
Could you update the commit title ?
something `[FIX] composer: do not autocomplete boolean values`

You can also add a reference to the task adding TRUE and FALSE functions which is task 3360232

somu-odoo:
Hey, @LucasLefevre updated as you suggested. I had a look into your PR, just a curious question, once that PR is merged this code won't be needed right? Considering all the changes made, autocomplete will auto close once =TRUE is set.

LucasLefevre:
fw-bot up to saas-16.3

PR #2553, opened 2023-06-05
[REF] Selection: Let caller control selection visibilty

rrahir:
The first design of the processor was too intertwined with the viewport,
and we ended up designing some commands mode/types  so that they would
simultaneously:
- fit our need to distinguish use cases that required the viewport(1)
  to move or not.
- still made sense from a functional point of view (the types & modes
  reflect the intention)

Unfortunately, we already have 9 different combinations of types, only
two third of them really used in order to distinguish intensions) and
a bug was reported (see https://github.com/odoo/o-spreadsheet/pull/2369) taht ipplies that this does not cover
all the expected behaviours.

We might have mistaken the role of the processor by thinking the API
so that the viewport would determine itself how to react to an update
of the selection when we actually implemented are methods that suit the
caller's will to reflect the change of the selection in the viewport or
not.

The processor is just a channel that should:

- dispatch the selection changes to the right listener
- in some cases notify the viewport that it should react to that change

This commit gets rid of the `SelectionEvent.type` in favor of specific
options to explicitely request some side effect behaviours like:
- scroll the viewport to display the newly selected cells
- whether the zone should be treated as an unbounded zone or not

Task: 3348042

## Description:

Odoo task ID : [3348042](https://www.odoo.com/web#id=3348042&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_core_views/sheetview.ts
+++ b/src/plugins/ui_core_views/sheetview.ts
@@ -133,29 +133,24 @@ export class SheetViewPlugin extends UIPlugin {
   }
 
   private handleEvent(event: SelectionEvent) {
-    switch (event.type) {
-      case "HeadersSelected":
-      case "AlterZone":
-        break;
-      case "ZonesSelected":
-        let { col, row } = findCellInNewZone(event.previousAnchor.zone, event.anchor.zone);
-        if (event.mode === "updateAnchor") {
-          const oldZone = event.previousAnchor.zone;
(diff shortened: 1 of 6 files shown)


LucasLefevre (approved):
robodoo r+

PR #2556, opened 2023-06-06
[FIX] chart panel: title state in chart design panel 

jash-odoo:
## Description:

Implemented t-model for state management of the chart title within
the sidepanel. By using t-model, the input value remains persistent
throughout rerendering, even when triggered by dispatches like multi-user
commands. This implementation ensures consistent display of the correct
value entered in the title input

Odoo task ID : [2991474](https://www.odoo.com/web#id=2991474&cids=2&menu_id=4720&action=333&active_id=2328&model=project.task&view_type=form)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/chart/line_bar_pie_panel/design_panel.ts
+++ b/src/components/side_panel/chart/line_bar_pie_panel/design_panel.ts
@@ -1,4 +1,5 @@
 import { Component, useExternalListener, useState } from "@odoo/owl";
+import { _t } from "../../../../translation";
 import { BarChartDefinition } from "../../../../types/chart/bar_chart";
 import { LineChartDefinition } from "../../../../types/chart/line_chart";
 import { PieChartDefinition } from "../../../../types/chart/pie_chart";
@@ -15,6 +16,7 @@ interface Props {
 }
 
 interface State {
+  title: string;
   fillColorTool: boolean;
 }
 
@@ -23,6 +25,7 @@ export class LineBarPieDesignPanel extends Component<Props, SpreadsheetChildEnv>
   static components = { ColorPickerWidget };
 
   private state: State = useState({
+    title: "",
     fillColorTool: false,
   });
 
@@ -31,6 +34,7 @@ export class LineBarPieDesignPanel extends Component<Props, SpreadsheetChildEnv>
   }
 
   setup() {
+    this.state.title = _t(this.props.definition.title);
     useExternalListener(window as any, "click", this.onClick);
   }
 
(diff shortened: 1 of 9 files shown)


src/components/side_panel/chart/line_bar_pie_panel/design_panel.ts
```diff
@@ -44,9 +48,9 @@ export class LineBarPieDesignPanel extends Component<Props, SpreadsheetChildEnv>
     });
   }
 
-  updateTitle(ev) {
+  updateTitle() {
     this.props.updateChart(this.props.figureId, {
-      title: ev.target.value,
+      title: _t(this.state.title),
```
hokolomopo, line 53: The translation should be at the setup, not here. 

The title of the chart might be translated (we extract chart titles for translation in dashboard), so translatting it at startup make sense. But the new title that the user just inputted will not be in the translation and even if it was a translatable string, we want to store the non-translated one in DB.

hokolomopo (commented):
Seems good :+1:

rrahir (changes requested):
could you add a test for this fix please ?

jash-odoo:
Hello @rrahir I added the test for same, can you please review this again?

tests/components/charts.test.ts
```diff
@@ -328,12 +328,31 @@ describe("charts", () => {
...
 
+  test.each(["basicChart", "scorecard", "gauge"])(
+    "edit chart title followed by clicking on grid will update the title of %s",
+    async (chartType: string) => {
+      createTestChart(chartType);
+      await nextTick();
+      await simulateClick(".o-figure");
+      await simulateClick(".o-figure-menu-item");
+      await simulateClick(".o-menu div[data-name='edit']");
+      await nextTick();
```
rrahir, line 346: you should check the current value first

tests/components/charts.test.ts
```diff
@@ -328,12 +328,31 @@ describe("charts", () => {
...
+      await simulateClick(".o-panel .inactive");
+      setInputValueAndTrigger(".o-chart-title input", "new_title", "input");
+      setInputValueAndTrigger(".o-chart-title input", "new_title", "change");
```
rrahir, line 349: In master, this specific line triggers the `updateTitle` method of the sidepanel which in turn updates the chart. This  means the test passes in master. you should remove it (which would make the test fail in master) and the see if your fix works

rrahir (changes requested):

rrahir (approved):
Nice :)

@robodoo r+

PR #2600, opened 2023-06-13
[FIX] LineBarPieConfigPanel: charts data series invalid input

dhrp-odoo:
## Description:

Previously, pressing the enter key with an invalid dataset in the charts side panel would result in an error.

This issue has been resolved by implementing a check using the `isConfirmable`. If an invalid range is detected, the function now returns immediately.

Odoo task ID : [3355063](https://www.odoo.com/web#id=3355063&cids=2&menu_id=4720&action=333&active_id=2328&model=project.task&view_type=form)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/components/selection_input/selection_input.ts
```diff
@@ -198,6 +198,10 @@ export class SelectionInput extends Component<Props, SpreadsheetChildEnv> {
         updateSelectionWithArrowKeys(ev, this.env.model.selection);
       }
     } else if (ev.key === "Enter") {
+      if (!this.isConfirmable) {
+        return; // Return if the dataset is invalid
+      }
```
hokolomopo, line 203: Not enough to fix the problem, it can still occur if we call `props.onSelectionConfirmed` from the `removeInput` method.

We should either:
1) ensure that we never fire a `props.onSelectionConfirmed` in the selection input when a range is invalid
2) handle wrong ranges in the `onDataSeriesConfirmed` of the side panel

Not sure which one is best, I'd probably go with 2) because the configPanel is already aware of the invalid ranges via `onDataSeriesRangesChanged` so it make sense to handle them correctly on confirm.

tests/components/charts.test.ts
```diff
@@ -731,6 +732,38 @@ describe("charts", () => {
...
+      createChart(
+        model,
+        {
+          dataSets: ["Sheet1!B2:B4"],
+          labelRange: "Sheet1!A2:A4",
+          type: "line",
+          title: "line_chart",
+        },
+        chartId
+      );
```
hokolomopo, line 745: use `createTestChart` for shorter test

hokolomopo (commented):
We can still get a traceback when deleting a range when another range is invalid :x

src/components/selection_input/selection_input.ts
dhrp-odoo: I have made the required changes and removed it from the `keydown` because it will now be handled in the `onDataSeriesConfirmed` function. Please review the changes and let me know if any potential issues.

LucasLefevre (commented):
could you split the diff in two different commits (one with the style fix, a second one with the range fix.)

dhrp-odoo:
Hi @LucasLefevre, I have implemented the suggested changes. Could you please review again and confirm if this is what you are expecting?

LucasLefevre (approved):
robodoo r+ rebase-ff
Nice, thanks :)

PR #2610, opened 2023-06-14
[FIX] sheet: duplicate sheet name problem when renaming

yunchen4:
## Description:

This PR fixes the problem that when editing sheet name and the new sheet name is the same as the old one, but the case is different, an error will occur. The sheet name should be case insensitive.

If after converting names into lower cases, two names are the same, then the renaming process will stop.

Odoo task ID : [3370632](https://www.odoo.com/web#id=3370632&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/core/sheet.ts
+++ b/src/plugins/core/sheet.ts
@@ -639,10 +639,16 @@ export class SheetPlugin extends CorePlugin<SheetState> implements SheetState {
   }
 
   private checkSheetName(cmd: RenameSheetCommand | CreateSheetCommand): CommandResult {
+    const originalSheetName = this.getters.tryGetSheetName(cmd.sheetId);
+    if (originalSheetName !== undefined && cmd.name === originalSheetName) {
+      return CommandResult.UnchangedSheetName;
+    }
+
     const { orderedSheetIds, sheets } = this;
     const name = cmd.name && cmd.name.trim().toLowerCase();
-
-    if (orderedSheetIds.find((id) => sheets[id]?.name.toLowerCase() === name)) {
+    if (
+      orderedSheetIds.find((id) => sheets[id]?.name.toLowerCase() === name && id !== cmd.sheetId)
+    ) {
       return CommandResult.DuplicatedSheetName;
     }
     if (FORBIDDEN_IN_EXCEL_REGEX.test(name!)) {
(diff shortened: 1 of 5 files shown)


src/plugins/core/sheet.ts
```diff
@@ -703,8 +705,8 @@ export class SheetPlugin extends CorePlugin<SheetState> implements SheetState {
     const oldName = sheet.name;
     this.history.update("sheets", sheet.id, "name", name.trim());
     const sheetIdsMapName = Object.assign({}, this.sheetIdsMapName);
-    sheetIdsMapName[name] = sheet.id;
```
LucasLefevre, line 706: Does it make a difference to move this line?

LucasLefevre (commented):
LGTM :+1: :)
could you add a plugin test in `sheets.test.ts` as well ? Thanks

LucasLefevre:
> add a commandResult.NameUnchanged

good idea :+1: 

>  I think we should extend that check to be case insensitive

People should be allowed to rename a sheet from "something" to "Something"

rrahir:
> People should be allowed to rename a sheet from "something" to "Something"

you're right I missed  that possibility

src/plugins/core/sheet.ts
```diff
@@ -639,10 +639,16 @@ export class SheetPlugin extends CorePlugin<SheetState> implements SheetState {
   }
 
   private checkSheetName(cmd: RenameSheetCommand | CreateSheetCommand): CommandResult {
+    const originalSheetName = this.getters.tryGetSheetName(cmd.sheetId);
+    if (originalSheetName !== undefined && cmd.name === originalSheetName) {
```
LucasLefevre, line 643: same as this , right ?
```suggestion
    if (cmd.name === originalSheetName) {
```

tests/plugins/sheets.test.ts
```diff
@@ -154,6 +154,21 @@ describe("sheets", () => {
...
   });
 
+  test("Rename command won't be dispatched if the name is unchanged (case sensitive)", () => {
+    const model = new Model({ sheets: [{ id: "11", name: "Sheet1" }] });
+    expect(renameSheet(model, "11", "Sheet1")).toBeCancelledBecause(
+      CommandResult.UnchangedSheetName
+    );
+  });
+
+  test("Sheet name is case insensitive", () => {
```
LucasLefevre, line 164:
```suggestion
  test("can change sheet name case", () => {
```

LucasLefevre (commented):
LGTM
only some suggestions :)

src/plugins/core/sheet.ts
yunchen4: @LucasLefevre Since `checkSheetName` is used elsewhere, `originalSheetName` is possibly `undefined` if the sheet doesn't exist.

LucasLefevre (approved):
Thanks :)
robodoo r+

PR #2625, opened 2023-06-21
[IMP] topbar: allow double clicking on paint format

yunchen4:
## Description:

In Microsoft office desktop, when double clicking paint format button, the paint format mode will be kept until it is manually disabled via clicking again or hitting ESC key.

This PR introduces the same functionality. A test helper function `doubleClick` is also added.

Odoo task ID : [3358256](https://www.odoo.com/web#id=3358256&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/components/grid_overlay/grid_overlay.ts
```diff
@@ -208,3 +222,17 @@ GridOverlay.props = {
   onGridMoved: Function,
   gridOverlayDimensions: String,
 };
+
+const CURSOR_SVG = /*xml*/ `
```
hokolomopo, line 226: This should probably be in icon.ts. And maybe the icon design coule be improved

hokolomopo: And w ehsould probably put it if a Css variable (if we support them), or in a css class, because having this chunk inlinde fucks up the document inspection

src/components/paint_format_button/paint_format_button.ts
```diff
@@ -0,0 +1,46 @@
...
+
+css/* scss */ `
+  .o-menu-item-button {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    margin: 2px;
+    border-radius: 2px;
+    min-width: 20px;
+  }
```
hokolomopo, line 13: Don't redefine a class that is already defined elsewhere. The code is already messy ATM, we use this class in the topbar when it's defined in acton button... We should probably move the class in spreadsheet.ts

src/types/commands.ts
```diff
@@ -537,6 +537,14 @@ export interface ActivatePaintFormatCommand {
   type: "ACTIVATE_PAINT_FORMAT";
 }
 
+export interface CancelPaintFormatCommand {
+  type: "CANCEL_PAINT_FORMAT";
+}
+
+export interface KeepPaintFormatCommand {
+  type: "KEEP_PAINT_FORMAT";
+}
```
hokolomopo, line 546: Just add a boolean `persistent: true` in `ACTIVATE_PAINT_FORMAT`, dispatching 2 commands isn't needed and probably a bad idea

src/components/paint_format_button/paint_format_button.ts
hokolomopo: Or just pass the class as props

src/components/paint_format_button/paint_format_button.xml
```diff
@@ -0,0 +1,15 @@
+<templates>
+  <t t-name="o-spreadsheet-PaintFormatButton" owl="1">
+    <span
+      class="o-menu-item-button"
+      t-att-title="env._t('Paint Format')"
```
hokolomopo, line 5:
```suggestion
      title="Paint Format"
```
titles are translated by default

src/plugins/ui_stateful/clipboard.ts
```diff
@@ -95,10 +96,12 @@ export class ClipboardPlugin extends UIPlugin {
           break;
         }
         const pasteOption = cmd.pasteOption || (this._isPaintingFormat ? "onlyFormat" : undefined);
-        this._isPaintingFormat = false;
         this.state.paste(cmd.target, { pasteOption, shouldPasteCF: true, selectTarget: true });
         this.lastPasteState = this.state;
-        this.status = "invisible";
+        if (!this.keepPaintingFormat) {
+          this._isPaintingFormat = false;
```
hokolomopo, line 102:
```suggestion
          this.isPaintingFormat = false;
```
while you're touching this, remove the leading underscore, we don't use this for private attributes

src/plugins/ui_stateful/clipboard.ts
```diff
@@ -46,6 +46,7 @@ export class ClipboardPlugin extends UIPlugin {
   private lastPasteState?: ClipboardState;
   private _isPaintingFormat: boolean = false;
   private originSheetId?: UID;
+  private keepPaintingFormat: boolean = false;
```
hokolomopo, line 49:
```suggestion
  private isPaintFormatPersistent: boolean = false;
``` 
What do you think about this name ?

src/plugins/ui_stateful/clipboard.ts
```diff
@@ -184,6 +187,16 @@ export class ClipboardPlugin extends UIPlugin {
           this.status = "invisible";
         }
         break;
+      case "KEEP_PAINT_FORMAT": {
+        this.keepPaintingFormat = true;
+        break;
+      }
+      case "CANCEL_PAINT_FORMAT": {
+        this._isPaintingFormat = false;
```
hokolomopo, line 195: I'd add a condition in the allowDispatch to check if we are currently in a paint format. Not a real problem in practice, but it doesn't hurt.

tests/components/bottom_bar.test.ts
```diff
@@ -250,8 +251,7 @@ describe("BottomBar component", () => {
     test("Double click on the sheet name make it editable and give it the focus", async () => {
       const sheetName = fixture.querySelector<HTMLElement>(".o-sheet-name")!;
       expect(sheetName.getAttribute("contenteditable")).toEqual("false");
-      triggerMouseEvent(sheetName, "dblclick");
-      await nextTick();
+      await doubleClick(sheetName);
```
hokolomopo, line 254: Nice :smiley:

tests/components/top_bar.test.ts
```diff
@@ -195,15 +196,67 @@ describe("TopBar component", () => {
...
+
+      expect(paintFormatTool.classList.contains("active")).toBeFalsy();
+    });
+
+    test("Double click to activate and keep it", async () => {
+      const { model } = await mountParent();
+      const paintFormatTool = fixture.querySelector('.o-menu-item-button[title="Paint Format"]')!;
+
+      expect(paintFormatTool.classList.contains("active")).toBeFalsy();
+
```
hokolomopo, line 223: nitpick: I think a new line after _every_ line of code is a bit ugly

hokolomopo (commented):
I don't think we should have a `KEEP_FORMAT` command. Otherwise mostly details :+1:

src/plugins/ui_stateful/clipboard.ts
yunchen4: @hokolomopo There's a getter to get this attribute with the same name but without underscore.

src/components/grid_overlay/grid_overlay.ts
hokolomopo: You could put the `cursor : ${CURSOR_SVG}` in a `paint-cursor` class, and enable/disable the class with some code. I think that would make the code more readable. At least the svg would be in the CSS and not in the HTML.

src/actions/format_actions.ts
```diff
@@ -246,10 +246,7 @@ export const formatCF: ActionSpec = {
 
 export const paintFormat: ActionSpec = {
```
hokolomopo, line 247: The action isn't used anymore, we can delete it

src/components/paint_format_button/paint_format_button.ts
```diff
@@ -0,0 +1,29 @@
...
+  get isActive() {
+    return this.env.model.getters.isPaintingFormat();
+  }
+
+  onDblClick() {
+    this.env.model.dispatch("ACTIVATE_PAINT_FORMAT", { persistent: true });
+  }
+
+  togglePaintFormat() {
+    if (this.env.model.getters.isPaintingFormat()) {
```
hokolomopo, line 20:
```suggestion
    if (this.isActive) {
```

tests/components/top_bar.test.ts
```diff
@@ -195,15 +196,64 @@ describe("TopBar component", () => {
...
+      await click(paintFormatTool);
+
+      expect(paintFormatTool.classList.contains("active")).toBeTruthy();
+
+      await click(paintFormatTool);
+
+      expect(paintFormatTool.classList.contains("active")).toBeFalsy();
+    });
+
+    test("When paint format (keep) is activated, single click will exit paint format mode", async () => {
```
hokolomopo, line 245:
```suggestion
    test("When paint format (persistent) is activated, single click will exit paint format mode", async () => {
```

hokolomopo (commented):
Some last details :)

src/components/grid_overlay/grid_overlay.ts
LucasLefevre: I agree with Adrien's proposition. Let's use a class

src/plugins/ui_stateful/clipboard.ts
```diff
@@ -46,6 +46,7 @@ export class ClipboardPlugin extends UIPlugin {
   private lastPasteState?: ClipboardState;
   private _isPaintingFormat: boolean = false;
   private originSheetId?: UID;
+  private isPaintFormatPersistent: boolean = false;
```
LucasLefevre, line 49: what do you think about merging both `_isPaintingFormat` and `isPaintFormatPersistent` into a single string property that would only take 3 values ?
Currently, the state `_isPaintingFormat = false` and `isPaintFormatPersistent = true` is technically possible even if it makes no sense and should never happen. With a single property and only 3 possible values, the inconsistent state is impossible by design (make impossible state impossible ;))

We need to come up with nice name though. The first thing coming to my mind is: `"inactive" | "active" | "persistent"` but I don't think those names are really good

tests/components/top_bar.test.ts
```diff
@@ -195,15 +196,55 @@ describe("TopBar component", () => {
...
+    test("Single click to activate paint format (once)", async () => {
+      const { model } = await mountParent();
+      const paintFormatTool = fixture.querySelector('.o-menu-item-button[title="Paint Format"]')!;
+      expect(paintFormatTool.classList.contains("active")).toBeFalsy();
+
+      await click(paintFormatTool);
+      expect(paintFormatTool.classList.contains("active")).toBeTruthy();
+
+      model.dispatch("PASTE", {
+        target: [{ top: 1, bottom: 1, left: 1, right: 1 }],
```
LucasLefevre, line 209:
```suggestion
        target: target("B2"),
```
same below

tests/components/grid.test.ts
```diff
@@ -873,19 +874,44 @@ describe("Grid component", () => {
...
+      gridMouseEvent(model, "mouseup", "D8");
+      expect(getCell(model, "D8")).toBeUndefined();
+    });
+
+    test("can keep the paint format mode", async () => {
+      setCellContent(model, "B2", "b2");
+      selectCell(model, "B2");
+      model.dispatch("SET_FORMATTING", {
+        sheetId: model.getters.getActiveSheetId(),
+        target: [{ left: 1, right: 1, top: 1, bottom: 1 }],
```
LucasLefevre, line 902:
```suggestion
        target: target("B2"),
```

tests/components/grid.test.ts
LucasLefevre, line 1: you can also add a test:
set some style on a cell, double click the paint brush to copy the style, then after update the cell style: painting other cell should use the original style (at least that's the current behavior. Did not tested what excel does.

LucasLefevre (commented):
nice and simple :)

LucasLefevre (approved):
robodoo r+
I really like the feature :)

PR #2626, opened 2023-06-22
[FIX] scorecard: measure text with bold/italic style

hokolomopo:
The style of a scorecard's element wasn't taken into account when measuring the text to get the best font size of the chart's elements. This lead to baseline/keyValue overflowing when their cell's style was bold, because we computed a font size for the non-bold text to fill the scorecard, but then we rendered the text with a bold style.

Odoo task ID : [3380613](https://www.odoo.com/web#id=3380613&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/tests/components/scorecard_chart.test.ts
+++ b/tests/components/scorecard_chart.test.ts
@@ -1,9 +1,11 @@
 import { Model } from "../../src";
 import { UID } from "../../src/types";
 import { ScorecardChartDefinition } from "../../src/types/chart/scorecard_chart";
+import { MockCanvasRenderingContext2D } from "../setup/canvas.mock";
 import {
   createScorecardChart as createScorecardChartHelper,
   setCellContent,
+  setStyle,
   updateChart,
 } from "../test_helpers/commands_helpers";
 import { dragElement, getElStyle, simulateClick } from "../test_helpers/dom_helper";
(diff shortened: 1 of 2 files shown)


tests/components/scorecard_chart.test.ts
```diff
@@ -281,6 +283,28 @@ describe("Scorecard charts", () => {
...
   });
 
+  test.each([{ keyValue: "A3" }, { baseline: "A3" }])(
+    "Scorecard elements take the bold/italic style into account when finding the best font size",
+    async (scorecardArg) => {
+      const string = "This is a long string that will be the keyvalue";
+      let ctxFontOFMeasuredValue = "";
+      jest
+        .spyOn(MockCanvasRenderingContext2D.prototype, "measureText")
+        .mockImplementation(function (this: MockCanvasRenderingContext2D, text) {
```
rrahir, line 293: you have to restore the mock for the following tests that use it. I think you either the mock is incomplete and fails to fallback on the original behaviour or the wait you implement it is correct and it should be cleared after the test with [this](https://jestjs.io/docs/mock-function-api#mockfnmockrestore)

rrahir (changes requested):

tests/components/scorecard_chart.test.ts
hokolomopo: Added a line to clean the mock :+1: 

I think it's better for the mock to return a random value rather than fallback to the original behaviour, so we can catch when we fail to clean the mock (like in this task)

LucasLefevre (approved):
robodoo r+

PR #2628, opened 2023-06-22
[IMP] conditionnal_formatting: use formula as value

anhe-odoo:
## Description:

In the "CellIsRule" part of the conditionnal formatting, we only allow to use a pure number (or string) as value, while GSheet and Excel allow to use formula.
This task add the possibility to use a formula instead of a value, using the same behaviour as Google and Excel (ie the formula is adapted to each cell taking into account the position of the cell, just as if we have some kind of "autofilled" condition for each cell.

Introducing a new getter getTranslatedCellFormula, some other parts of the code has been adapted to use it (paste, autofill and sort) to avoid to duplicate the same process. Moreover, the getter buildFormulaContent has been removed to only use the getFormulaCellContent getter, as the previous was only calling the first, but the getForm... was already used in Odoo so, in order to avoid changing stuff in Odoo, we keep the getFormulaCellContent.

## Relate Task
-Odoo task ID : [3222537](https://www.odoo.com/web#id=3222537&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## Review Checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


tests/plugins/conditional_formatting.test.ts
```diff
@@ -935,6 +935,60 @@ describe("conditional formats types", () => {
       });
     });
 
+    test("Operator GreaterThan with formula", () => {
+      model.dispatch("ADD_CONDITIONAL_FORMAT", {
+        cf: {
+          rule: {
+            type: "CellIsRule",
+            operator: "GreaterThan",
+            values: ["=SUM(A1:B2)/4"],
```
hokolomopo, line 944: Maybe also a test with a simple reference "=A1"

hokolomopo (commented):
I still think that the functional behaviour is stupid as fuck and barely usable by a standard user (I know it's ghseet behaviour, but still). 

We should at the very least have a comment/doc to explain what happens.

src/plugins/core/cell.ts
```diff
@@ -298,8 +298,28 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
...
+    formulaCell: {
+      compiledFormula: CompiledFormula;
+      dependencies?: Range[];
+    }
+  ) {
+    const dependencies = this.getters.createAdaptedRanges(
+      formulaCell.dependencies ||
+        formulaCell.compiledFormula.dependencies.map((d) =>
+          this.getters.getRangeFromSheetXC(sheetId, d)
+        ),
```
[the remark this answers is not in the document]
anhe-odoo: No, it's not exactly the same object. Sometimes we have -1 for the "right" value of the zone in `formulaCell.dependencies` while we have 0 in `compiledFormula.dependencies` :/

src/components/side_panel/conditional_formatting/cell_is_rule_editor.xml
```diff
@@ -24,15 +24,16 @@
...
           t-model="rule.values[0]"
           t-att-class="{ 'o-invalid': isValue1Invalid }"
           class="o-input o-cell-is-value o-required"
+          t-on-keydown="(ev) => this.onKeydown(ev)"
         />
         <t t-if="rule.operator === 'Between' || rule.operator === 'NotBetween'">
           <input
             type="text"
-            placeholder="and value"
+            placeholder="and value or formula"
```
hokolomopo, line 36: No F4 handling on this input ?

src/plugins/ui_stateful/edition.ts
```diff
@@ -297,37 +298,52 @@ export class EditionPlugin extends UIPlugin {
     }
   }
 
-  // ---------------------------------------------------------------------------
-  // Misc
-  // ---------------------------------------------------------------------------
+  getUpdatedReference(selection, content) {
```
hokolomopo, line 301: typing

tests/components/conditional_formatting.test.ts
```diff
@@ -225,6 +225,24 @@ describe("UI of conditional formats", () => {
       });
     });
 
+    test("can edit an existing CellIsRule", async () => {
```
hokolomopo, line 228: Test should have a more explicit title with F4

src/plugins/ui_stateful/edition.ts
hokolomopo, line 301:
```suggestion
  getCycledReference(selection, content) {
```
maybe ?

hokolomopo (commented):
:+1:

rrahir (commented):
overall looks good :) I'm not sure about the F4 shortcut though, it should either be removed or go further (I think it's useless personally)

anhe-odoo:
@rrahir : It should be okay now :)

tests/plugins/clipboard.test.ts
```diff
@@ -1766,7 +1766,7 @@ describe("clipboard", () => {
 
   test.each([
     ["=A1", "=#REF"],
-    ["=SUM(A1:B1)", "=SUM(#REF)"],
+    //["=SUM(A1:B1)", "=SUM(#REF)"],
```
rrahir, line 1769: ?

tests/plugins/conditional_formatting.test.ts
```diff
@@ -935,6 +935,108 @@ describe("conditional formats types", () => {
...
+      expect(getStyle(model, "B2")).toEqual({ fillColor: "#ff0f0f" });
+    });
+
+    test("Operator GreaterThan with fixed simple reference", () => {
+      model.dispatch("ADD_CONDITIONAL_FORMAT", {
+        cf: {
+          rule: {
+            type: "CellIsRule",
+            operator: "GreaterThan",
+            values: ["=$A$2"],
```
rrahir, line 968: I would go as far as testing the different combinations of fixed col/ row  (a$2, $A2)

rrahir:
@robodoo r+

PR #2636, opened 2023-06-26
[FIX] clipboard: invalidate clipboards on alteration in cut zone

khpa-odoo:
## Description:

_Before this PR:_
If content or style of cut zone was altered before paste, cut zone was cleared and old/copied content and style were pasted.

_After this PR:_
Any modification in cut zone was before performing paste operation would restrict the paste and invalidate internal and OS clipboard both.

Odoo task ID : [3358431](https://www.odoo.com/web#id=3358431&menu_id=4720&cids=2&action=333&active_id=2328&model=project.task&view_type=form)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/helpers/clipboard/clipboard_cells_state.ts
```diff
@@ -473,6 +481,17 @@ export class ClipboardCellsState extends ClipboardCellsAbstractState {
     }
   }
 
+  private getCellData(cellsInRow, position): ClipboardCell[] {
```
hokolomopo, line 484: Giving an array as argument of a function to push things inside should be avoided. Either the function should do the whole looping on the row and return an array, or it should return a single cell.

And don't forget to type the arguments.

src/helpers/clipboard/clipboard_cells_state.ts
```diff
@@ -505,7 +524,9 @@ export class ClipboardCellsState extends ClipboardCellsAbstractState {
     for (const row of this.cells) {
       htmlTable += "<tr>";
       for (const cell of row) {
-        const cssStyle = cssPropertiesToCss(cellStyleToCss(cell.style));
+        const cssStyle = cssPropertiesToCss(
+          cellStyleToCss(this.getters.getCellComputedStyle(cell.position))
+        );
```
hokolomopo, line 529: This is correct, but once again 
1) try to make different commits for different fixes. Or at least mention it in the commit message/PR message
2) Create a test when you fix something

tests/components/grid.test.ts
```diff
@@ -1429,18 +1432,42 @@ describe("Copy paste keyboard shortcut", () => {
     expect(getCellContent(model, "A2")).toEqual("things");
   });
 
-  test("cut zone gets cleared on paste if content/style is altered after cut", async () => {
+  test("Cannot paste if content is altered after cut", async () => {
```
hokolomopo, line 1435: The 2 tests should probably be in `clipboard.test.ts` since we don't test anything dependant to grid

src/helpers/clipboard/clipboard_cells_state.ts
```diff
@@ -108,6 +102,20 @@ export class ClipboardCellsState extends ClipboardCellsAbstractState {
...
 
   isPasteAllowed(target: Zone[], clipboardOption?: ClipboardOptions): CommandResult {
+    let updatedCells: ClipboardCell[][] = [];
+    this.cells.forEach((cell) => {
+      let rowCells: ClipboardCell[] = [];
+      cell.forEach((rowCell) => {
+        rowCells = this.getCellData(rowCells, rowCell.position);
+      });
+      updatedCells.push(rowCells);
+    });
```
hokolomopo, line 112: Use `for(const .. of ...)` rather than forEach, and use "const" instead of "let".

hokolomopo (commented):
The fix seems good, some small implmentation problem tho

src/helpers/clipboard/clipboard_cells_state.ts
khpa-odoo: It must have slipped my mind...sorry about that.
I've split the commit and added a test case now :+1: 
Let me know if it needs any amendments :)

src/helpers/clipboard/clipboard_cells_state.ts
```diff
@@ -108,6 +102,20 @@ export class ClipboardCellsState extends ClipboardCellsAbstractState {
...
+      const rowCells: ClipboardCell[] = [];
+      for (const rowCell of cell) {
+        rowCells.push(this.getCellData(rowCell.position));
+      }
+      updatedCells.push(rowCells);
+    }
+
+    if (!deepEquals(this.cells, updatedCells) && this.operation === "CUT") {
+      // cannot paste if content or style in cut zone is altered before paste
+      this.cells = [];
```
LucasLefevre, line 116: the state should only be modified by commands (see command query separation pattern [1])
Here the state is modified even though the command is being rejected.
Besides, rejecting the command is not what we want. We don't want to prevent updating a cell in the clipped zones.

The clipboard state should be invalidated in the clipboard plugin, by handling `UPDATE_CELL` commands.
You can add an abstract method `isInvalidatedBy(cmd: Command): boolean`,  handle `UPDATE_CELL` commands in there and return `true`/`false` depending on the clipped zones.
In a following commit, you can remove `isColRowDirtyingClipboard` and move the implementation in the new `isInvalidatedBy` by handling `ADD_COLUMNS_ROWS` and `REMOVE_COLUMNS_ROWS`

[1] https://github.com/odoo/o-spreadsheet/blob/saas-16.3/doc/extending/architecture.md#model-commands-and-getters

LucasLefevre (changes requested):

src/helpers/clipboard/clipboard_cells_state.ts
khpa-odoo: Hi! Sorry for the delayed response.
I understood what you wanted to imply and changed my approach accordingly. It took a while since I faced some edge cases while solving the issue.
Although I've implemented the necessary changes now. Thank you!

LucasLefevre (commented):
submitting this partial review now, waiting for the PO's input on task

src/components/grid/grid.ts
```diff
@@ -353,6 +353,10 @@ export class Grid extends Component<Props, SpreadsheetChildEnv> {
   };
 
   focus() {
+    if (this.env.model.getters.isCutZoneAltered()) {
+      this.emptyClipboard();
```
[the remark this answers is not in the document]
khpa-odoo: I definitely agree to this point! Any approach would make the code seem ugly if we want to go out of the way to achieve this. 
In that case, I'll make the related changes later according to PO's opinion and push the remaining changes for now.

khpa-odoo:
Dropped invalidation of OS clipboard and pushed the changes :+1:

rrahir (changes requested):

khpa-odoo:
Applied all changes :+1:

PR #2642, opened 2023-06-27
[FIX] ActionButton: update SVG icon to display current style in top bar

aath-odoo:
## Description:

Upadated the SVG icon in the top bar to reflect the current style. The icon now correctly displays the appropriate style based on the selected style.

Odoo task ID : [3341969](https://www.odoo.com/web#id=3341969&cids=2&menu_id=3940&action=4043&model=project.task&view_type=form)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/tests/components/top_bar.test.ts
+++ b/tests/components/top_bar.test.ts
@@ -915,3 +915,26 @@ test("The menu items are orderer by their sequence", async () => {
   expect(menuItems[1].dataset.name).toBe("second");
   expect(menuItems[2].dataset.name).toBe("third");
 });
+
+describe("Topbar svg icon", () => {
+  test.each([
+    [{ align: "left" }, "Horizontal align", "align-left"],
+    [{ align: "center" }, "Horizontal align", "align-center"],
+    [{ align: "right" }, "Horizontal align", "align-right"],
+    [{ verticalAlign: "top" }, "Vertical align", "align-top"],
+    [{ verticalAlign: "middle" }, "Vertical align", "align-middle"],
+    [{ verticalAlign: "bottom" }, "Vertical align", "align-bottom"],
+    [{ wrapping: "clip" }, "Wrapping", "wrapping-clip"],
+    [{ wrapping: "wrap" }, "Wrapping", "wrapping-wrap"],
+    [{ wrapping: "overflow" }, "Wrapping", "wrapping-overflow"],
+  ])("Icon in top bar matches the selected cell style", async (style, buttonTitle, iconClass) => {
+    const model = new Model();
+    setStyle(model, "A1", style as Style);
+
+    ({ fixture } = await mountSpreadsheet({ model }));
+    await nextTick();
+
+    const icon = fixture.querySelector(`.o-menu-item-button[title="${buttonTitle}"] svg`);
+    expect(icon?.classList.contains(iconClass)).toBeTruthy();
+  });
+});
--- a/src/actions/action.ts
+++ b/src/actions/action.ts
@@ -36,7 +36,7 @@ export interface ActionSpec {
   /**
    * Can be defined to display an icon
    */
-  icon?: string;
+  icon?: string | ((env: SpreadsheetChildEnv) => string);
   /**
    * is the action allowed when running spreadsheet in readonly mode
    */
@@ -67,7 +67,7 @@ export interface Action {
   isVisible: (env: SpreadsheetChildEnv) => boolean;
   isEnabled: (env: SpreadsheetChildEnv) => boolean;
   isActive?: (env: SpreadsheetChildEnv) => boolean;
-  icon: string;
+  icon: (env: SpreadsheetChildEnv) => string;
   isReadonlyAllowed: boolean;
   execute?: (env: SpreadsheetChildEnv) => unknown;
   children: (env: SpreadsheetChildEnv) => Action[];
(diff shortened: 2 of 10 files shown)


tests/components/top_bar.test.ts
```diff
@@ -342,6 +356,21 @@ describe("TopBar component", () => {
       await click(fixture, `.o-menu-item-button[title="${iconTitle}"]`);
       expect(model.getters.getCurrentStyle()).toEqual(expectedStyle);
     });
+    test.each([
+      ["Top", { verticalAlign: "top" }],
+      ["Middle", { verticalAlign: "middle" }],
+      ["Bottom", { verticalAlign: "bottom" }],
+    ])("can set vertical alignmen '%s't with the toolbar", async (iconTitle, expectedStyle) => {
```
rrahir, line 363:
```suggestion
    ])("can set vertical alignment '%s' with the toolbar", async (iconTitle, expectedStyle) => {
```

tests/components/top_bar.test.ts
```diff
@@ -309,6 +309,20 @@ describe("TopBar component", () => {
       await click(fixture, `.o-menu-item-button[title="${iconTitle}"]`);
       expect(model.getters.getCurrentStyle()).toEqual(expectedStyle);
     });
+    test.each([["Left (Ctrl+Shift+L)"], ["Center (Ctrl+Shift+E)"], ["Right (Ctrl+Shift+R)"]])(
+      "Update icon on change",
+      async (iconTitle) => {
```
rrahir, line 314: You should directly test against the actions and not look at how they are rendered. I suggest you look at how Lucas did it here : https://github.com/odoo/o-spreadsheet/pull/2651/files#diff-e729fc703acf88b1fee2adb7d8a03b2081192d64b7ae73e52dc7cee34fc7d306R879  He's testing the dynamic change of the key `description` so it's really close to what you are doing :)

tests/components/top_bar.test.ts
rrahir: same comment as above

tests/components/top_bar.test.ts
```diff
@@ -375,6 +404,20 @@ describe("TopBar component", () => {
       await click(fixture, `.o-menu-item-button[title="${iconTitle}"]`);
       expect(model.getters.getCurrentStyle()).toEqual(expectedStyle);
     });
+    test.each([["Overflow"], ["Wrap"], ["Clip"]])(
+      "can set the wrapping state '%s' with the toolbar",
```
rrahir, line 408: same comment as above

rrahir (changes requested):

tests/components/top_bar.test.ts
aath-odoo: Hello @rrahir ,I attempted the approach you recommended. However, due to the absence of a registry for action buttons on the toolbar, I couldn't utilize the getNode() as it uses topbarmenuregistry function as shown in Lucas's pull request. Nevertheless, I devised an alternative method to associate the icon with the current cell style. Please take a look.

tests/components/top_bar.test.ts
```diff
@@ -873,3 +873,25 @@ test("The menu items are orderer by their sequence", async () => {
...
+  [{ align: "right" }, "align-right", "Horizontal align"],
+  [{ wrapping: "clip" }, "wrapping-clip", "Wrapping"],
+  [{ wrapping: "wrap" }, "wrapping-wrap", "Wrapping"],
+  [{ wrapping: "overflow" }, "wrapping-overflow", "Wrapping"],
+])(
+  "Icon in top bar matches the selected cell style",
+  async (style, expectedTitleActive, buttonTitle) => {
+    const model = new Model();
+    setStyle(model, "A1", style as Style);
+    await mountParent(model);
```
rrahir, line 892: wrap this test in a describe where you instantiate a single parent then in the test.each, set the style, await a next tick

rrahir (changes requested):
The implementation became way to generic and tries to handle apples an oranges all the same. You should keep the distinction between the different kind of options (wrapping vs horizontal vs vertical).
 I think the implementation of 2-3 weeks ago was better, except for the helpers that you could add to improve.


I would like @somu-odoo  to go through the pr as well onece you do the changes :)

src/helpers/ui/icon_alignment.ts
```diff
@@ -0,0 +1,64 @@
+import {
+  DEFAULT_HORIZONTAL_ALIGN,
+  DEFAULT_VERTICAL_ALIGN,
+  DEFAULT_WRAPPING_MODE,
+} from "../../constants";
+import { SpreadsheetChildEnv } from "../../types";
+
+export function getIconAlignment(env: SpreadsheetChildEnv, alignmentType: string) {
```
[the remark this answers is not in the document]
aath-odoo: Okay i've done the changes and got the previous implementation of the fix if it's what you were referring to. It seems you're right we don't need to make it too generic, helpers functions did the job :), although do you think we should update the icons in topbar menu as well or we can keep them as is?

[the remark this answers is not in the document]
rrahir: I think it's good as it is :)

rrahir:
@robodoo r+

PR #2643, opened 2023-06-27
[IMP] Composer: Select autocomplete results on mouseover

rrahir:
the current behaviour of the composer formula autocompletion is somehow limited as the function description can only be seen on the currently selected entry, which can only be modified via keyboard navigation. This revision allows the selection to be updated by hovering on one of the entries of the dropdown.

Task: 3387413

## Description:

Odoo task ID : [3387413](https://www.odoo.com/web#id=3387413&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/composer/autocomplete_dropdown/autocomplete_dropdown.ts
+++ b/src/components/composer/autocomplete_dropdown/autocomplete_dropdown.ts
@@ -5,13 +5,12 @@ import { AutocompleteValue } from "../composer/composer";
 css/* scss */ `
   .o-autocomplete-dropdown {
     pointer-events: auto;
+    cursor: pointer;
     background-color: #fff;
     max-width: 400px;
-    & > div:hover {
-      background-color: #f2f2f2;
-    }
+
     .o-autocomplete-value-focus {
-      background-color: rgba(0, 0, 0, 0.08);
+      background-color: #f2f2f2;
     }
 
     & > div {
@@ -28,6 +27,7 @@ interface Props {
   values: AutocompleteValue[];
   selectedIndex: number | undefined;
   onValueSelected: (value: string) => void;
+  onValueHovered: (index: string) => void;
 }
 
 export class TextValueProvider extends Component<Props> {
@@ -38,4 +38,5 @@ TextValueProvider.props = {
   values: Array,
   selectedIndex: { type: Number, optional: true },
   onValueSelected: Function,
+  onValueHovered: Function,
 };
--- a/src/components/composer/autocomplete_dropdown/autocomplete_dropdown.ts
+++ b/src/components/composer/autocomplete_dropdown/autocomplete_dropdown.ts
@@ -26,6 +26,7 @@ css/* scss */ `
 interface Props {
   values: AutocompleteValue[];
   selectedIndex: number | undefined;
+  getHtmlContent: (value: string) => string;
   onValueSelected: (value: string) => void;
   onValueHovered: (index: string) => void;
 }
@@ -37,6 +38,7 @@ export class TextValueProvider extends Component<Props> {
 TextValueProvider.props = {
   values: Array,
   selectedIndex: { type: Number, optional: true },
+  getHtmlContent: Function,
   onValueSelected: Function,
   onValueHovered: Function,
 };
--- a/src/components/composer/composer/composer.ts
+++ b/src/components/composer/composer/composer.ts
@@ -420,12 +420,12 @@ export class Composer extends Component<ComposerProps, SpreadsheetChildEnv> {
           text,
           description,
         };
+      })
+      .sort((a, b) => {
+        return a.text.length - b.text.length || a.text.localeCompare(b.text);
       });
     if (searchTerm) {
       values = fuzzyLookup(searchTerm, values, (t) => t.text);
-    } else {
-      // alphabetical order
-      values = values.sort((a, b) => a.text.localeCompare(b.text));
     }
     this.autoCompleteState.values = values.slice(0, 10);
     this.autoCompleteState.selectedIndex = 0;
--- a/src/components/composer/composer/composer.ts
+++ b/src/components/composer/composer/composer.ts
@@ -1,5 +1,5 @@
 import { Component, onMounted, onWillUnmount, useEffect, useRef, useState } from "@odoo/owl";
-import { DEFAULT_FONT, NEWLINE } from "../../../constants";
+import { COMPOSER_ASSISTANT_COLOR, DEFAULT_FONT, NEWLINE } from "../../../constants";
 import { EnrichedToken } from "../../../formulas/index";
 import { functionRegistry } from "../../../functions/index";
 import { clip, fuzzyLookup, getZoneArea, isEqual, splitReference } from "../../../helpers/index";
@@ -15,6 +15,7 @@ import {
   SpreadsheetChildEnv,
 } from "../../../types/index";
 import { css, cssPropertiesToCss } from "../../helpers/css";
+import { getHtmlContentFromPattern } from "../../helpers/html_content_helpers";
 import { updateSelectionWithArrowKeys } from "../../helpers/selection_helpers";
 import { ComposerFocusType } from "../../spreadsheet/spreadsheet";
 import { TextValueProvider } from "../autocomplete_dropdown/autocomplete_dropdown";
@@ -24,6 +25,8 @@ import { FunctionDescriptionProvider } from "../formula_assistant/formula_assist
 const functions = functionRegistry.content;
 
 const ASSISTANT_WIDTH = 300;
+
+const AUTOCOMPLETE_ENTRIES = 10;
 
 export const selectionIndicatorClass = "selector-flag";
 const selectionIndicatorColor = "#a9a9a9";
@@ -83,6 +86,13 @@ css/* scss */ `
       position: absolute;
       margin: 1px 4px;
       pointer-events: none;
+
+      .o-semi-bold {
+        /** FIXME: to remove in favor of Bootstrap
+        * 'fw-semibold' when we upgrade to Bootstrap 5.2
+        */
+        font-weight: 600 !important;
+      }
     }
   }
 `;
@@ -111,6 +121,7 @@ interface AutoCompleteState {
   selectedIndex: number | undefined;
   values: AutocompleteValue[];
   type: "function" | "dataValidation";
+  getHtmlContent: (text: string) => HtmlContent[];
 }
 
 interface FunctionDescriptionState {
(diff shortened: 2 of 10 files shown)


src/components/composer/composer/composer.ts
```diff
@@ -397,24 +401,56 @@ export class Composer extends Component<ComposerProps, SpreadsheetChildEnv> {
...
-    this.autoCompleteState.values = values.slice(0, 10);
+
+    this.autoCompleteState.values = values.slice(0, AUTOCOMPLETE_ENTRIES);
+    this.autoCompleteState.getHtmlContent = (value) =>
+      this.getHtmlContentFromPattern(searchTerm, value);
     this.autoCompleteState.selectedIndex = 0;
   }
 
+  private getHtmlContentFromPattern(pattern: string, value: string): HtmlContent[] {
+    const pendingHtmlContent: HtmlContent[] = [];
```
hokolomopo, line 420: This probably should be in the `TextValueProvider` to not pollute the Composer component I'd say

src/components/composer/composer/composer.ts
```diff
@@ -397,24 +401,56 @@ export class Composer extends Component<ComposerProps, SpreadsheetChildEnv> {
...
+        pendingHtmlContent.push(
+          { value: textSlice, color: "" },
+          { value: value[i], color: COMPOSER_ASSISTANT_COLOR }
+        );
+        patternIndex++;
+        previousIndex = i + 1;
+      }
+    }
+    pendingHtmlContent.push({ value: value.slice(previousIndex) });
+    const HtmlContent = pendingHtmlContent.filter((content) => content.value);
```
hokolomopo, line 438:
```suggestion
    const htmlContent = pendingHtmlContent.filter((content) => content.value);
```

tests/components/autocomplete_dropdown.test.ts
```diff
@@ -412,3 +445,20 @@ describe("autocomplete boolean functions", () => {
     }
   );
 });
+describe("composer entries", () => {
+  test("Autocomplente entries are sorted by length and then alphanumerically", async () => {
```
hokolomopo, line 449: We should ideally mock the function registry for all the tests, or else the test will break when we introduce a new function. We shoudl create a helper so we can do it easily in one line tho, something like


```js
mockFunctionRegistry(["SEC", "SUM", "SLN", ...]);
```

hokolomopo (commented):
Nice feature ! 👍

src/components/composer/composer/composer.ts
```diff
@@ -397,24 +401,56 @@ export class Composer extends Component<ComposerProps, SpreadsheetChildEnv> {
...
+    this.autoCompleteState.getHtmlContent = (value) =>
+      this.getHtmlContentFromPattern(searchTerm, value);
     this.autoCompleteState.selectedIndex = 0;
   }
 
+  private getHtmlContentFromPattern(pattern: string, value: string): HtmlContent[] {
+    const pendingHtmlContent: HtmlContent[] = [];
+    pattern = pattern.toLowerCase();
+    let previousIndex = 0;
+    let len = value.length;
```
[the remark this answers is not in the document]
rrahir: yup, I copied the `fuzzyMatch` function :/

src/components/composer/composer/composer.ts
rrahir: I'm not sure, if we do that, we'd have to pass the pattern to `TextValueProvider` so we'd move the whole logic to the child component (to stay consistent) but I don't feel it should be the case, the 'pattern' is no concern of the `TextValueProvider`

src/constants.ts
```diff
@@ -24,6 +24,7 @@ export const BACKGROUND_HEADER_FILTER_COLOR = "#E6F4EA";
 export const BACKGROUND_HEADER_SELECTED_FILTER_COLOR = "#CEEAD6";
 export const SEPARATOR_COLOR = "#E0E2E4";
 export const ICONS_COLOR = "#4A4F59";
+export const COMPOSER_ASSISTANT_COLOR = "#9B359B";
```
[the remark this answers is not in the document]
rrahir: @Francois-frgi  waht do you think of this? I had the same feeling depending on the luminosity of the screen I was using :/  FYI, it's the color we use in the formula assistant to highlight the  current argument that the user is editing.

tests/components/autocomplete_dropdown.test.ts
```diff
@@ -343,6 +368,14 @@ describe("Autocomplete parenthesis", () => {
 });
 
 describe("composer Assistant", () => {
+  beforeEach(async () => {
+    genericBeforeEach();
+  });
+
+  afterAll(() => {
+    restoreDefaultFunctions();
+  });
```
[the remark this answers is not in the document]
rrahir: the only thing that counts is that the registry is back to normal when we get out of this test file. And the issue is that some tests require a different registry than  the one defined in the genericBeforeEach, like in `autocomplete boolean functions`  but I guess we can just add them to the list of existing functions (since we actually already added them in the real functionRegistry >< )

tests/components/autocomplete_dropdown.test.ts
rrahir: I agree on the idea, but you'd have 2 ways to do thing then. you either mock it with existing functions or you define new functions for the sake of the test (what we currently do).

src/components/composer/composer/composer.ts
hokolomopo: I'd rather have a `highlighterPattern` props than more things in the composer. Just my opnion tho.

tests/components/autocomplete_dropdown.test.ts
hokolomopo: Do we ever need to mock it with existing functions ? :thinking:

rrahir: maybe not :o

src/components/composer/autocomplete_dropdown/autocomplete_dropdown.ts
```diff
@@ -6,11 +6,9 @@ css/* scss */ `
   .o-autocomplete-dropdown {
     pointer-events: auto;
     background-color: #fff;
-    & > div:hover {
-      background-color: #f2f2f2;
-    }
+
```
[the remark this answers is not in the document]
rrahir: addressed :)

src/components/composer/composer/composer.ts
rrahir: that means intrinsically linking the behaviour of `TextValueProvider` with their parent , the highlighterPattern does not make much sense by itself, does it?
That being said, the `getHtmlContentFromPattern` could be a simple helper as it has nothing to do with the composer

rrahir: @LucasLefevre  do you have any thoughts for this?

rrahir: I changed getHtmlContentFromPattern  to be a helper

src/components/composer/composer/composer.ts
```diff
@@ -431,6 +431,10 @@ export class Composer extends Component<ComposerProps, SpreadsheetChildEnv> {
     this.autoCompleteState.selectedIndex = 0;
   }
 
+  updateAutoCompleteIndex(index: number) {
+    this.autoCompleteState.selectedIndex = clip(0, index, 10);
```
LucasLefevre, line 449: probably good to extract 10 to a constant at the top of the file

tests/components/autocomplete_dropdown.test.ts
```diff
@@ -39,33 +39,42 @@ async function typeInComposer(text: string, fromScratch: boolean = true) {
   return composerEl;
 }
 
-beforeEach(() => {
+const genericBeforeEach = () => {
```
LucasLefevre, line 42: I'm not sure what the purpose of this function is if it's called only one, in a `beforeEach` :thinking:

src/components/helpers/html_content_helpers.ts
```diff
@@ -0,0 +1,25 @@
+import { HtmlContent } from "../composer/composer/composer";
+
+export function getHtmlContentFromPattern(
```
LucasLefevre, line 3: I think that's worth an docstring with an example.
And maybe even a few unit tests demonstrating what it does/how it works

src/components/composer/composer/composer.ts
```diff
@@ -424,10 +429,19 @@ export class Composer extends Component<ComposerProps, SpreadsheetChildEnv> {
...
+
     if (searchTerm) {
-      values = fuzzyLookup(searchTerm, values, (t) => t.text);
+      values = fuzzyLookup(searchTerm, values, (t) => t.text).slice(0, AUTOCOMPLETE_ENTRIES);
     }
-    this.autoCompleteState.values = values.slice(0, 10);
+
+    this.autoCompleteState.values = values.slice(0, AUTOCOMPLETE_ENTRIES);
+    this.autoCompleteState.getHtmlContent = (value) =>
+      getHtmlContentFromPattern(
```
LucasLefevre, line 439: why is it a function ? Can't it be set to the result of `getHtmlContentFromPattern` ?

src/components/helpers/html_content_helpers.ts
```diff
@@ -0,0 +1,25 @@
...
+  pattern: string,
+  value: string,
+  highlightColor: string,
+  className: string
+): HtmlContent[] {
+  const pendingHtmlContent: HtmlContent[] = [];
+  pattern = pattern.toLowerCase();
+
+  for (const patternChar of pattern) {
+    const index = value.toLocaleLowerCase().indexOf(patternChar);
```
LucasLefevre, line 13: what if the char is not found ?

tests/components/__snapshots__/autocomplete_dropdown.test.ts.snap
```diff
@@ -10,7 +10,18 @@ exports[`Functions autocomplete autocomplete simple snapshot with =S 1`] = `
...
     >
-      SUM
+      <span
+        class="text-decoration-underline"
+        style="color: #9B359B;"
+      >
+        S
+      </span>
+      <span
+        style="color: undefined;"
```
LucasLefevre, line 20: `"color: undefined;"` looks louche. `style` attribute should probably be omitted if there's no color

src/components/composer/composer/composer.ts
rrahir: See https://github.com/odoo/o-spreadsheet/pull/2643#discussion_r1253082096

LucasLefevre (approved):
robodoo r+ rebase-ff

PR #2645, opened 2023-06-27
[IMP] TopBarComposer: allow selection in readonly mode

dhrp-odoo:
## Description:

Previously, users were unable to select specific portions within cell content in readonly mode.

To achieve this, Added the 'user-select-text' class to enable selection. Additionally, encountered an issue where the focus was shifted to the hiddenInput when users made a selection, resulting in the removal of the selection. To resolve this issue, Added `t-on-click.stop=""`  in TopBarComposer.

Odoo task ID : [3378521](https://www.odoo.com/web#id=3378521&menu_id=4720&cids=2&action=333&active_id=2328&model=project.task&view_type=form)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/composer/top_bar_composer/top_bar_composer.xml
+++ b/src/components/composer/top_bar_composer/top_bar_composer.xml
@@ -1,6 +1,9 @@
 <templates>
   <t t-name="o-spreadsheet-TopBarComposer" owl="1">
-    <div class="o-topbar-composer bg-white" t-att-style="containerStyle">
+    <div
+      class="o-topbar-composer bg-white user-select-text"
+      t-on-click.stop=""
+      t-att-style="containerStyle">
       <Composer
         focus="props.focus"
         inputStyle="composerStyle"
--- a/src/components/composer/composer/composer.ts
+++ b/src/components/composer/composer/composer.ts
@@ -71,10 +71,6 @@ css/* scss */ `
       padding-right: 3px;
       outline: none;
 
-      &.unfocusable {
-        pointer-events: none;
-      }
-
       p {
         margin-bottom: 0px;
 
--- a/src/components/composer/composer/composer.xml
+++ b/src/components/composer/composer/composer.xml
@@ -3,7 +3,7 @@
     <div class="o-composer-container w-100 h-100">
       <div
         class="o-composer w-100 text-start"
-        t-att-class="{ 'text-muted': env.model.getters.isReadonly(), 'unfocusable': env.model.getters.isReadonly(), 'active': props.focus !== 'inactive' }"
+        t-att-class="{ 'text-muted': env.model.getters.isReadonly(), 'active': props.focus !== 'inactive' }"
         t-att-style="props.inputStyle"
         t-ref="o_composer"
         tabindex="1"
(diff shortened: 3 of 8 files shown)


src/components/top_bar/top_bar.ts
```diff
@@ -188,6 +188,14 @@ export class TopBar extends Component<Props, SpreadsheetChildEnv> {
     this.closeMenus();
   }
 
+  onTopbarClick() {
+    if (this.env.model.getters.isReadonly()) {
```
LucasLefevre, line 192: There's an unintended effect with this.
If you click anywhere on the top bar, let's say open a menu, the focus is no longer on the grid and you can move the selection with the keyboard anymore.
I think another way of solving this is to stop the click event on the top bar composant only with a `t-on-click.stop=""`

I think you can also try to write a test testing the focus is kept on the composer when clicking on it

src/components/composer/top_bar_composer/top_bar_composer.xml
```diff
@@ -1,6 +1,6 @@
 <templates>
   <t t-name="o-spreadsheet-TopBarComposer" owl="1">
-    <div class="o-topbar-composer bg-white" t-att-style="containerStyle">
+    <div class="o-topbar-composer bg-white user-select-text" t-att-style="containerStyle">
```
LucasLefevre, line 3:
```suggestion
    <div class="o-topbar-composer bg-white user-select-text" t-on-click.stop="" t-att-style="containerStyle">
```

LucasLefevre (commented):
nice :)
only a suggestion with an unintended effect

src/components/top_bar/top_bar.ts
dhrp-odoo: I have made the necessary changes, but there is a discrepancy in the test results. When clicking on the composer in the readonly mode, the developer tool shows it as the `activeElement`, but during test execution, the entire HTML body is obtained instead. Do you have any suggestions to explain this inconsistency?

image

image

LucasLefevre: Hmm I see. This doesn't surprise me.
In tests, it's not a full browser like Chrome or Firefox. It's a basic/partial implementation of web and DOM APIs for node js. It's https://github.com/jsdom/jsdom

Since it's only a simplified/partial implementation, some things does not properly work and doesn't reflect reality. For example, text selection is not implemented at all in jsdom. Focus is usually working, but maybe it's not working here.
Can you push the failing test and I'll have a look ?

dhrp-odoo: Okay, I have pushed the test, Thanks for your time!!!

LucasLefevre: I checked but couldn't really understand what's going on. I don't want to spend too much time on this.
@rrahir wants to take a look.
If we can't find anything, we'll just remove the test (and pray it doesn't break in the future :crossed_fingers: )

LucasLefevre (approved):
robodoo r+

@dhrp-odoo The issue in the test was indeed linked to jsdom behavior. Tricky stuff. Rémi used the `simulateClick` helper in the test and changed its implementation to better work with jsdom behavior.

Thanks @dhrp-odoo and @rrahir for your work

PR #2673, opened 2023-07-11
[FIX] SheetPlugin: Prevent deletion of all non-frozen headers

dhrp-odoo:
## Description:

This PR contains two commits that tackle the following issues:

***Commit 1:***

Issue: Deleting rows/columns, except the frozen ones, would break the spreadsheet. Hiding some rows and then deleting all rows, except the frozen ones, or using a data filter to hide some rows and then deleting all rows, would also cause issues.

Solution: Implemented a validation mechanism in the `SheetPlugin.` The `allowDispatch` function now checks for attempts to delete all rows or columns when frozen panes are present. If such actions are detected, the corresponding command will be rejected.

***Commit 2:***

Issue: The `searchHeaderIndex` function returned `-1` when trying to hide all rows/cols, leading to a broken sheet. This occurred because the `adjustViewportZoneY` function used this value as top.

Solution: Modified the `searchHeaderIndex` function to handle `top` and `left` `-1` cases correctly. Now, when trying to hide all rows or columns, the function reassigns the value of `boundaries.top` and `boundaries.left` respectively, resolving the issue.

Task: [3414127](https://www.odoo.com/web#id=3414127&cids=2&menu_id=4720&action=333&active_id=2328&model=project.task&view_type=form)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/internal_viewport.ts
+++ b/src/helpers/internal_viewport.ts
@@ -365,6 +365,9 @@ export class InternalViewport {
       this.boundaries.right,
       this.searchHeaderIndex("COL", this.viewportWidth, this.left)
     );
+    if (this.left === -1) {
+      this.left = this.boundaries.left;
+    }
     if (this.right === -1) {
       this.right = this.getters.getNumberCols(sheetId) - 1;
     }
@@ -382,6 +385,9 @@ export class InternalViewport {
       this.boundaries.bottom,
       this.searchHeaderIndex("ROW", this.viewportHeight, this.top)
     );
+    if (this.top === -1) {
+      this.top = this.boundaries.top;
+    }
     if (this.bottom === -1) {
       this.bottom = this.getters.getNumberRows(sheetId) - 1;
     }
--- a/src/plugins/core/header_visibility.ts
+++ b/src/plugins/core/header_visibility.ts
@@ -5,7 +5,7 @@ import { CorePlugin } from "../core_plugin";
 
 export class HeaderVisibilityPlugin extends CorePlugin {
   static getters = [
-    "canRemoveHeaders",
+    "checkElementsIncludeAllVisibleHeaders",
     "getHiddenColsGroups",
     "getHiddenRowsGroups",
     "isRowHiddenByUser",
@@ -41,7 +41,7 @@ export class HeaderVisibilityPlugin extends CorePlugin {
         if (!this.getters.tryGetSheet(cmd.sheetId)) {
           return CommandResult.InvalidSheetId;
         }
-        if (!this.canRemoveHeaders(cmd.sheetId, cmd.dimension, cmd.elements)) {
+        if (this.checkElementsIncludeAllVisibleHeaders(cmd.sheetId, cmd.dimension, cmd.elements)) {
           return CommandResult.NotEnoughElements;
         }
         return CommandResult.Success;
@@ -97,9 +97,13 @@ export class HeaderVisibilityPlugin extends CorePlugin {
     return;
   }
 
-  canRemoveHeaders(sheetId: UID, dimension: Dimension, elements: HeaderIndex[]): boolean {
+  checkElementsIncludeAllVisibleHeaders(
+    sheetId: UID,
+    dimension: Dimension,
+    elements: HeaderIndex[]
+  ): boolean {
     const visibleHeaders = this.getAllVisibleHeaders(sheetId, dimension);
-    return !includesAll(elements, visibleHeaders);
+    return includesAll(elements, visibleHeaders);
   }
 
   isRowHiddenByUser(sheetId: UID, index: HeaderIndex): boolean {
--- a/src/plugins/core/sheet.ts
+++ b/src/plugins/core/sheet.ts
@@ -4,10 +4,12 @@ import {
   deepCopy,
   getUnquotedSheetName,
   groupConsecutive,
+  includesAll,
   isDefined,
   isZoneInside,
   isZoneValid,
   positions,
+  range,
   toCartesian,
 } from "../../helpers/index";
 import { _lt, _t } from "../../translation";
@@ -71,6 +73,7 @@ export class SheetPlugin extends CorePlugin<SheetState> implements SheetState {
     "getSheetSize",
     "getSheetZone",
     "getPaneDivisions",
+    "checkElementsIncludeAllNonFrozenHeaders",
   ] as const;
 
   readonly sheetIdsMapName: Record<string, UID | undefined> = {};
@@ -138,6 +141,10 @@ export class SheetPlugin extends CorePlugin<SheetState> implements SheetState {
             : this.getNumberRows(cmd.sheetId);
         if (Math.min(...cmd.elements) < 0 || Math.max(...cmd.elements) > elements) {
           return CommandResult.InvalidHeaderIndex;
+        } else if (
+          this.checkElementsIncludeAllNonFrozenHeaders(cmd.sheetId, cmd.dimension, cmd.elements)
+        ) {
+          return CommandResult.NotEnoughElements;
         } else {
           return CommandResult.Success;
         }
@@ -481,6 +488,27 @@ export class SheetPlugin extends CorePlugin<SheetState> implements SheetState {
       panes.ySplit = base;
     }
     this.history.update("sheets", sheetId, "panes", panes);
+  }
+
+  /**
+   * Checks if all non-frozen header indices are present in the provided elements of selected rows/columns.
+   * This validation ensures that all rows or columns cannot be deleted when frozen panes exist.
+   */
+  checkElementsIncludeAllNonFrozenHeaders(
+    sheetId: UID,
+    dimension: Dimension,
+    elements: HeaderIndex[]
+  ): boolean {
+    const paneDivisions = this.getters.getPaneDivisions(sheetId);
+    const startIndex = dimension === "ROW" ? paneDivisions.ySplit : paneDivisions.xSplit;
+    const endIndex = this.getters.getNumberHeaders(sheetId, dimension);
+
+    if (!startIndex) {
+      return false;
+    }
+
+    const indicesToCheck = range(startIndex, endIndex);
+    return includesAll(elements, indicesToCheck);
   }
 
   // ---------------------------------------------------------------------------
--- a/src/registries/menus/topbar_menu_registry.ts
+++ b/src/registries/menus/topbar_menu_registry.ts
@@ -103,11 +103,13 @@ topbarMenuRegistry
     name: ACTIONS.REMOVE_ROWS_NAME,
     sequence: 80,
     action: ACTIONS.REMOVE_ROWS_ACTION,
+    isVisible: (env: SpreadsheetChildEnv) => ACTIONS.CAN_REMOVE_COLUMNS_ROWS("ROW", env),
   })
   .addChild("edit_delete_column", ["edit"], {
     name: ACTIONS.REMOVE_COLUMNS_NAME,
     sequence: 90,
     action: ACTIONS.REMOVE_COLUMNS_ACTION,
+    isVisible: (env: SpreadsheetChildEnv) => ACTIONS.CAN_REMOVE_COLUMNS_ROWS("COL", env),
   })
   .addChild("edit_delete_cell_shift_up", ["edit"], {
     name: _lt("Delete cell and shift up"),
--- a/tests/menu_items_registry.test.ts
+++ b/tests/menu_items_registry.test.ts
@@ -301,6 +301,17 @@ describe("Menu Item actions", () => {
         elements: [3, 4],
       });
     });
+
+    test("selecting all non-frozen rows should hide the option for deletion", async () => {
+      const sheetId = model.getters.getActiveSheetId();
+      const lastRow = model.getters.getNumberRows(sheetId) - 1;
+
+      freezeRows(model, 4, sheetId);
+      selectRow(model, 4, "newAnchor");
+      selectRow(model, lastRow, "updateAnchor");
+
+      expect(getNode(path).isVisible(env)).toBeFalsy();
+    });
   });
 
   describe("Edit -> edit_delete_column", () => {
@@ -362,6 +373,17 @@ describe("Menu Item actions", () => {
         dimension: "COL",
         elements: [3, 4],
       });
+    });
+
+    test("selecting all non-frozen columns should hide the option for deletion", async () => {
+      const sheetId = model.getters.getActiveSheetId();
+      const lastColumn = model.getters.getNumberCols(sheetId) - 1;
+
+      freezeColumns(model, 3, sheetId);
+      selectColumn(model, 3, "newAnchor");
+      selectColumn(model, lastColumn, "updateAnchor");
+
+      expect(getNode(path).isVisible(env)).toBeFalsy();
     });
   });
 
--- a/src/registries/menus/col_menu_registry.ts
+++ b/src/registries/menus/col_menu_registry.ts
@@ -72,11 +72,7 @@ colMenuRegistry
     name: ACTIONS.REMOVE_COLUMNS_NAME,
     sequence: 90,
     action: ACTIONS.REMOVE_COLUMNS_ACTION,
-    isVisible: (env: SpreadsheetChildEnv) => {
-      const sheetId = env.model.getters.getActiveSheetId();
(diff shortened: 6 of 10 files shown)


src/plugins/core/sheet.ts
```diff
@@ -233,6 +256,39 @@ export class SheetPlugin extends CorePlugin<SheetState> implements SheetState {
...
+    const paneDivisions = this.getters.getPaneDivisions(sheetId);
+    const startIndex = dimension === "ROW" ? paneDivisions.ySplit : paneDivisions.xSplit;
+    const endIndex =
+      dimension === "ROW"
+        ? this.getters.getNumberRows(sheetId) - 1
+        : this.getters.getNumberCols(sheetId) - 1;
+
+    const hiddenColsRowsGroups: Set<HeaderIndex> = new Set(
+      dimension === "ROW"
+        ? this.getters.getHiddenRowsGroups(sheetId).flat()
```
hokolomopo, line 273: Just scrolled through the pr really fast, and this won't work because, as I noted in the pad, the headers can be hidden by UI features (data filters). I'm not 100% sur on the best way to proceed, but this definitely shouldn'tbe handled in core (or maybe as a temporary stopgag measure).

dhrp-odoo: Hi, I'm aware of this issue. Currently this code fixes the issue with delete and hide of rows/cols in normal cases. I'm trying to figure out a way to fix data filters, but it seems rather tricky.

Second point you mentioned that I should not include it in core, well freeze is originally controlled in sheet.ts, so it made more sense to modify the allowDispatch there. I'll try to look for more alternatives

rrahir (changes requested):
Several comments:
1.this requires at least a rebase - we have updated some APIs
2. Looking at the video in the task I feeel like you go beyond the scope that we wanted to cover at first which  is - you cannot delete all the rows/columns of the unfrozen parts of a spreadsheet. I don't understand how this is related to their visibility,

src/plugins/ui/header_visibility_ui.ts
```diff
@@ -12,6 +21,66 @@ export class HeaderVisibilityUIPlugin extends UIPlugin {
...
 
+  allowDispatch(cmd: Command): CommandResult {
+    switch (cmd.type) {
+      case "REMOVE_COLUMNS_ROWS":
+        if (!this.getters.tryGetSheet(cmd.sheetId)) {
+          return CommandResult.InvalidSheetId;
+        }
+
+        if (this.isAllHeaderIndices(cmd.sheetId, cmd.dimension, cmd.elements)) {
+          this.ui.notifyUI({
```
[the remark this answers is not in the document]
dhrp-odoo: Hi, I am aware that we have updated the API to catch user errors. However, since it was in the master branch, I thought I would update my code there.

Nevertheless, I completely agree with you and will handle this using the example you provided.

src/helpers/internal_viewport.ts
```diff
@@ -365,6 +365,9 @@ export class InternalViewport {
       this.boundaries.right,
       this.searchHeaderIndex("COL", this.viewportWidth, this.left)
     );
+    if (this.left === -1) {
+      this.left = this.getters.getPaneDivisions(sheetId).xSplit;
```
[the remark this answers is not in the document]
dhrp-odoo: image

If we attempt to filter all rows using a data filter, in this case, the `searchHeaderIndex` function returns -1, resulting in a broken sheet.

src/plugins/ui/header_visibility_ui.ts
```diff
@@ -12,6 +21,66 @@ export class HeaderVisibilityUIPlugin extends UIPlugin {
     "isHeaderHidden",
   ] as const;
 
+  allowDispatch(cmd: Command): CommandResult {
+    switch (cmd.type) {
+      case "REMOVE_COLUMNS_ROWS":
```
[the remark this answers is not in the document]
dhrp-odoo: I tried to implement the fix in core earlier, but as suggested by ADRM
Here: https://github.com/odoo/o-spreadsheet/pull/2673#discussion_r1262465671
Discord: https://discord.com/channels/678381219515465750/687337979743830044/1131556525123571732

This approach makes sense since we won't be able to access the rows hidden by the UI data filter, which necessitates keeping our code within the UI.

If we don’t go beyond the issues written in specs of this task, then I believe we can skip the data filter validation part, and can ONLY handle the removal of columns/rows (via the core).

Since we have a similar file in core and (this) as an extended part in UI, I thought it would be the best place to drop CORE commands through allowDispatch. Although, in future versions (16.3 & above) this won’t work either. So, going to core would be the only choice but then again data filter has to be handled separately in UI.

hokolomopo (commented):
Hey ! 

I think your commit on the internal viewport is mostly good, just take into account @rrahir comment (replace `this.getters.getPaneDivisions(sheetId).xSplit` with `this.boundaries.left` maybe? I'm not 100% sure, I'm not too familiar with the viewport's code).

The commit on HeaderVisibilityUIPlugin is just plain wrong tho. You cannot and should not user allowDispatch on a _core_ command in an _UI_ plugin. This either break the collaborative if a command was accepted by an user and not the other, or won't work at all if the remote command don't go through the allowDispatch.

The good news is, you don't really need to do this anymore ! Your other commit make it so the viewport won't break if all the rows that are not in the frozen pane are hidden/deleted. So you could leave it alone.

But I think you're right and it's nice to prevent the user from deleting/hiding all the non-frozen rows. But this is impossible to include the data filters in all this. We have to keep everything in core, and ignore the data filters (which, again, do'nt break the sheet anymore with your other commit).

What i'd do is to update the getter `canRemoveHeaders` with what you've done, so we cannot hide/remove all the rows below the frozen pane. The getter is used in the `allowDispatch`, and is used to disable the menu items. So you do'nt have to notify the user, since the menu items to make the wrong action aren't there in the first place. 

If you discussed the error message with Francois, and he'd rather have an error message than the disabled menu items, you should write an `interactiveDeleteRows` method instead of using a notifyUI in a  plugin (look at `interactiveAddMerge` for an example).

Let me know if something is unclear or you have any questions.

dhrp-odoo:
Hello @hokolomopo,

Indeed, RAR's observation was correct. The solution for the Internal Viewport was overly broad, I've implemented the changes as per suggestion. 

Regarding the second commit, I've introduced an `interactiveRemoveColumnsRows` method. This method rejects the command, generating a result that can be managed through an interactive call. The corresponding command is handled within the core's `SheetPlugins allowDispatch`.

Additionally, I believe that we no longer need to be concerned about hidden rows. Please let me know if you have any potential edge cases in mind.

Please review the changes again at your earliest convenience.

Thank you

src/plugins/core/sheet.ts
```diff
@@ -233,6 +235,36 @@ export class SheetPlugin extends CorePlugin<SheetState> implements SheetState {
     }
   }
 
+  /**
+   * Checks if all non-frozen header indices are present in the provided elements of selected rows/columns.
+   * This validation ensures that all rows or columns cannot be deleted when frozen panes exist.
+   */
+  private checkHeaderVisibility(
```
hokolomopo, line 242: function name is wrong, this has nothing to do with visibility

src/plugins/core/sheet.ts
```diff
@@ -233,6 +235,36 @@ export class SheetPlugin extends CorePlugin<SheetState> implements SheetState {
...
+        ? this.getters.getNumberRows(sheetId) - 1
+        : this.getters.getNumberCols(sheetId) - 1;
+
+    if (!startIndex) {
+      return false;
+    }
+
+    for (let i = startIndex; i <= endIndex; i++) {
+      if (!elements.includes(i)) {
+        // If the index is not present in the elements of selected rows/columns, then return false.
```
hokolomopo, line 260: This comment is kinda usless, you juste explain 2 easy to understand lines of code.

src/helpers/ui/sheet_interactive.ts
```diff
@@ -38,3 +38,24 @@ export function interactiveRenameSheet(env: SpreadsheetChildEnv, sheetId: UID, e
...
+) {
+  const result = env.model.dispatch("REMOVE_COLUMNS_ROWS", {
+    sheetId: env.model.getters.getActiveSheetId(),
+    dimension: dimension,
+    elements: elements,
+  });
+  const errorMessage = _lt(
+    `Sorry, it is not possible to delete all non-frozen ${
+      dimension === "ROW" ? "rows" : "columns"
+    }.`
```
hokolomopo, line 55: translations don't work with backticks. Use standard string with placeholders `%s`

hokolomopo (commented):
Some details and we'll be good I think

hokolomopo:
Also remove the commit to "prevent deletion of all non-frozen headers". Your current title make it looks like we can delete _all_ the headers when there's a frozen pane

src/plugins/core/sheet.ts
```diff
@@ -233,6 +235,35 @@ export class SheetPlugin extends CorePlugin<SheetState> implements SheetState {
...
+    sheetId: UID,
+    dimension: Dimension,
+    elements: HeaderIndex[]
+  ): boolean {
+    const paneDivisions = this.getters.getPaneDivisions(sheetId);
+    const startIndex = dimension === "ROW" ? paneDivisions.ySplit : paneDivisions.xSplit;
+    const endIndex =
+      dimension === "ROW"
+        ? this.getters.getNumberRows(sheetId) - 1
+        : this.getters.getNumberCols(sheetId) - 1;
```
hokolomopo, line 252: use `getNumberHeaders`

hokolomopo (commented):
Thinking about it some more, what you propose isn't wrong but it give us a consistency issue: removing all the cols is not possible because the menu item is hidden, but removing all the non-frozen header do'nt hide the menu item but gives an error message. We really should do the same thing every time. 

I'd keep the menu item hidden for everything, since it is what we had before. You should make your function a getter, and call it in the isVisible of the menu items. You can also probably simplify the writing of it by using the helper  `includesAll` (see in `canRemoveHeaders`).  Also rename `canRemoveHeaders` to `checkElementsIncludeAllVisibleHeaders` to be consistent.

dhrp-odoo:
> Thinking about it some more, what you propose isn't wrong but it give us a consistency issue: removing all the cols is not possible because the menu item is hidden, but removing all the non-frozen header do'nt hide the menu item but gives an error message. We really should do the same thing every time.
> 
> I'd keep the menu item hidden for everything, since it is what we had before. You should make your function a getter, and call it in the isVisible of the menu items. You can also probably simplify the writing of it by using the helper `includesAll` (see in `canRemoveHeaders`). Also rename `canRemoveHeaders` to `checkElementsIncludeAllVisibleHeaders` to be consistent.

I've made the suggested changes, including adding the test case.

src/plugins/core/header_visibility.ts
```diff
@@ -97,7 +97,11 @@ export class HeaderVisibilityPlugin extends CorePlugin {
...
   }
 
-  canRemoveHeaders(sheetId: UID, dimension: Dimension, elements: HeaderIndex[]): boolean {
+  checkElementsIncludeAllVisibleHeaders(
+    sheetId: UID,
+    dimension: Dimension,
+    elements: HeaderIndex[]
+  ): boolean {
     const visibleHeaders = this.getAllVisibleHeaders(sheetId, dimension);
     return !includesAll(elements, visibleHeaders);
```
hokolomopo, line 106:
```suggestion
    return includesAll(elements, visibleHeaders);
```
remove the `!` and update the calls of this method. With the "!" it does the opposite of wha its name says

src/helpers/ui/sheet_interactive.ts
```diff
@@ -38,3 +38,23 @@ export function interactiveRenameSheet(env: SpreadsheetChildEnv, sheetId: UID, e
     error: errorText,
   });
 }
+
+export function interactiveRemoveColumnsRows(
```
hokolomopo, line 42: Remove the interactive method, sicne we decided to go with hidden menu items instead.

tests/components/grid_manipulation.test.ts
```diff
@@ -150,6 +151,20 @@ describe("Context Menu add/remove row/col", () => {
     expect(fixture.querySelector(".o-menu div[data-name='hide_rows']")).toBeNull();
   });
 
+  test("cannot delete all non-frozen rows with contextmenu when frozen panes exist", async () => {
```
hokolomopo, line 154: The test should be in `menu_items_registry.test.ts`

hokolomopo (commented):
Hopefully last changes before r+

tests/components/grid_manipulation.test.ts
dhrp-odoo: Since the `cannot delete nor hide all rows with contextmenu` test case is written here, that's why I have included my test case here. [TESTCASE](https://github.com/odoo/o-spreadsheet/blob/34011c3de11b883dbd9ca0494d3ff8e09e5b58c8/tests/components/grid_manipulation.test.ts#L123-L151)

Should we consider adding this to the menu_items_registry.test.ts file?

src/helpers/ui/sheet_interactive.ts
hokolomopo: The topbar also uses the menu item, so the option won't show either :)

dhrp-odoo: Well, In my case the option still appears from the top bar edit.
Maybe my fix was not in the right place, let me check

hokolomopo: Mh yeah we didn't use the exact same menu item in 16.0. Extract the isVisible in a helper `CAN_REMOVE_COLUMNS` then, and put it also in the topbar edit menu item.

dhrp-odoo: I have implemented the suggested changes; It should look better now :)

src/registries/menus/topbar_menu_registry.ts
```diff
@@ -103,11 +104,13 @@ topbarMenuRegistry
     name: ACTIONS.REMOVE_ROWS_NAME,
     sequence: 80,
     action: ACTIONS.REMOVE_ROWS_ACTION,
+    isVisible: (env: SpreadsheetChildEnv) => CAN_REMOVE_COLUMNS_ROWS("ROW", env),
```
hokolomopo, line 107: the helper should be at the same place as the other menu item helpers (`lenu_item_action.ts`), and be imported/used like the others  `ACTION.CAN_REMOVE_COLUMNS_ROW`

tests/menu_items_registry.test.ts
```diff
@@ -301,6 +303,20 @@ describe("Menu Item actions", () => {
...
+      const sheetId = model.getters.getActiveSheetId();
+      const lastRow = model.getters.getNumberRows(sheetId) - 1;
+
+      freezeRows(model, 4, sheetId);
+      selectRow(model, 4, "newAnchor");
+      selectRow(model, lastRow, "updateAnchor");
+
+      simulateContextMenu(".o-row-resizer", ROW_5);
+      await nextTick();
+      expect(fixture.querySelector(".o-menu div[data-name='delete_row']")).toBeNull();
```
hokolomopo, line 318: Change it to use the `getNode(path).isVisible` like the most other tests of this file. The way you did it won't work at the forward port, we optimized this file by not mounting the spreadsheet, disallowing the use of contextClicks to open the menu

tests/menu_items_registry.test.ts
```diff
@@ -363,6 +379,20 @@ describe("Menu Item actions", () => {
...
+      const sheetId = model.getters.getActiveSheetId();
+      const lastColumn = model.getters.getNumberCols(sheetId) - 1;
+
+      freezeColumns(model, 3, sheetId);
+      selectColumn(model, 3, "newAnchor");
+      selectColumn(model, lastColumn, "updateAnchor");
+
+      simulateContextMenu(".o-col-resizer", COLUMN_D);
+      await nextTick();
+      expect(fixture.querySelector(".o-menu div[data-name='delete_column']")).toBeNull();
```
hokolomopo, line 394: same here

tests/menu_items_registry.test.ts
dhrp-odoo: Changes made

hokolomopo (commented):
@robodoo r+

thanks for the work 👍

hokolomopo:
@robodoo rebase-ff

PR #2680, opened 2023-07-12
[FIX] find_and_replace: Update cell don't update the search

jash-odoo:
## Description:

Previously, when we updated a cell that content was already searched in the Find and Replace, it wouldn't be considered a match. However, this PR aims to solve that issue. With this update, whenever a cell is updated, the search will also update and include the cell if it matches

Task: : [3422481](https://www.odoo.com/web#id=3422481&cids=2&menu_id=4720&action=333&active_id=2328&model=project.task&view_type=form)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui/find_and_replace.ts
+++ b/src/plugins/ui/find_and_replace.ts
@@ -53,6 +53,7 @@ export class FindAndReplacePlugin extends UIPlugin {
     searchFormulas: false,
   };
   private toSearch: string = "";
+  private isSearchDirty = false;
 
   // ---------------------------------------------------------------------------
   // Command Handling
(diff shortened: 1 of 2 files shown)


src/plugins/ui/find_and_replace.ts
```diff
@@ -84,13 +85,23 @@ export class FindAndReplacePlugin extends UIPlugin {
       case "ADD_COLUMNS_ROWS":
         this.clearSearch();
         break;
+      case "UPDATE_CELL":
```
rrahir, line 88: Unfortunately, this is not sufficient.  the evaluation process can be somehow asynchronous. Some functions can require a remote call to get the data fro the server. In such cases (we do this for odoo functions for instance),  we dispatch the command `EVALUATE_CELLS` that will restart the evaluation, the search should therefore also be updated on `EVALUATE_CELLS`.

this is something worth noting for everyone  @somu-odoo

rrahir (changes requested):

rrahir:
@robodoo r+

PR #2682, opened 2023-07-13
[FIX] find_and_replace: show formula checkbox not updating properly 

jash-odoo:
## Description:

Previously, when opening the side panel of the "Find and Replace" feature and then navigating to the "View" option in the top bar menu to click on "Show Formula," the checkbox wouldn't update. The checkbox would only update when clicking on the side panel.

This commit addresses this issue by managing the state of the checkbox, allowing it to update when using the "Show Formula" option from the top bar menu.

Deprecate onFocusSidePanel, because PR #2680 already manage updating search with updating or evaluating cell. So this will not require any more.

With the deprecation of onFocusSidePanel, the dispatching of REFRESH_SEARCH is no longer necessary and has thus been removed.

Task: [3422489](https://www.odoo.com/web#id=3422489&cids=2&menu_id=4720&action=333&active_id=2328&model=project.task&view_type=form)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/components/side_panel/find_and_replace/find_and_replace.ts
```diff
@@ -92,7 +100,6 @@ export class FindAndReplacePanel extends Component<Props, SpreadsheetChildEnv> {
   }
 
   onFocusSidePanel() {
-    this.state.searchOptions.searchFormulas = this.env.model.getters.shouldShowFormulas();
```
rrahir, line 95: since we just approved #2680, you should be able to get rid of `onFocusSidePanel`altogether

rrahir:
@robodoo r+

@jash-odoo  in general , don't bother too much with the update of tsdoc when fixing stuff in old versions ;-)

PR #2685, opened 2023-07-17
[IMP] model: don't dispatch local commands to core plugins

hokolomopo:
## Description

Before this commit, local commands were dispatched to core plugins. This made no sense from an architecture standpoint, as core plugins shouldn't be aware of UI commands.

Task:

## Description:

Task: : [3184514](https://www.odoo.com/web#id=3184514&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/tests/model.test.ts
+++ b/tests/model.test.ts
@@ -126,6 +126,19 @@ describe("Model", () => {
     expect(receivedCommands).not.toContain("COPY");
   });
 
+  test("Core plugins handle don't receive UI commands", () => {
+    const receivedCommands: CommandTypes[] = [];
+    class MyCorePlugin extends CorePlugin {
+      handle(cmd: CoreCommand) {
+        receivedCommands.push(cmd.type);
+      }
+    }
+    addTestPlugin(corePluginRegistry, MyCorePlugin);
+    const model = new Model();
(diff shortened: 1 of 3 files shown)


tests/model.test.ts
```diff
@@ -130,6 +130,19 @@ describe("Model", () => {
...
   });
 
+  test("Core plugins handle don't receive UI commands", () => {
+    const receivedCommands: CommandTypes[] = [];
+    class MyCorePlugin extends CorePlugin {
+      handle(cmd: CoreCommand) {
+        receivedCommands.push(cmd.type);
+      }
+    }
+    corePluginRegistry.add("myCorePlugin", MyCorePlugin);
```
LucasLefevre, line 140: you should clean the plugin registry because it leads to a memory leak. Every model instantiated after this test will have this plugin and `receivedCommands` will grow and grow with every dispatched command
(I'm sure there are other places where we don't properly clean test plugins)

LucasLefevre (commented):
LGTM

PR #2686, opened 2023-07-17
[REF] composer, grid: homogenize event handling

hokolomopo:
## Description

In the composer, we had different strategies to handle events. To handle `NumpadDecimal`, we intercepted the event and did the modification to the composer content by hand.

For new line event, we intercepted Ctrl+Enter, then dispatched an InputEvent `insertParagraph`, then we still had to handle the event by hand. This commit remove the step of dispatching a new InputEvent, which was useless.

Note on the why: the `insertParagraph` should have been handled by the contentEditable, but isn't since it's dispatched by the code, and is not trusted. Thus it needed to still be handled by hand.

This commit also made use the same helper in both `grid.ts` and `composer.ts` to transform the keyboard event into strings. We now keep the native JS CamelCase for events keys instead of upper-casing everything.

Task:

## Description:

Task: : [3380401](https://www.odoo.com/web#id=3380401&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/composer/composer/composer.ts
+++ b/src/components/composer/composer/composer.ts
@@ -9,12 +9,14 @@ import { ComposerSelection } from "../../../plugins/ui_stateful/edition";
 import {
   Color,
   CSSProperties,
+  Direction,
   DOMDimension,
   FunctionDescription,
   Rect,
   SpreadsheetChildEnv,
 } from "../../../types/index";
 import { css, cssPropertiesToCss } from "../../helpers/css";
+import { keyboardEventToShortcutString } from "../../helpers/dom_helpers";
 import { getHtmlContentFromPattern } from "../../helpers/html_content_helpers";
 import { updateSelectionWithArrowKeys } from "../../helpers/selection_helpers";
 import { ComposerFocusType } from "../../spreadsheet/spreadsheet";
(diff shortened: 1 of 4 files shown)


src/components/composer/composer/composer.ts
```diff
@@ -338,7 +346,13 @@ export class Composer extends Component<ComposerProps, SpreadsheetChildEnv> {
   }
 
   onKeydown(ev: KeyboardEvent) {
-    let handler = this.keyMapping[ev.key] || this.keyCodeMapping[ev.code];
+    if (ev.key.startsWith("Arrow")) {
```
anhe-odoo, line 349: It may be a better way, but I personnaly think that it was more readable to have all the handler together instead of some processors defined in the handlers variables above and the arrow-key processor directly called here :/

anhe-odoo (commented):
Nices changes :) I'm just not sure about the idea to take the Arrow keys handling out of the handlers

src/components/composer/composer/composer.ts
anhe-odoo: I personally prefer to have the four in the handlers, but maybe I'm the only one to find more-readable :D Should be discussed with others maybe ?

LucasLefevre (approved):
sorry about the time this PR has been hanging.
I rebased and resolved the conflicts.

Just like Anthony, I'm not sure about the idea to take the Arrow keys handling out of the handlers, but let's merge it. It's been waiting for far too long

robodoo r+

src/components/helpers/dom_helpers.ts
```diff
@@ -41,3 +41,29 @@ export function* iterateChildren(el: Node): Generator<Node> {
 export function getOpenedMenus(): HTMLElement[] {
   return Array.from(document.querySelectorAll<HTMLElement>(".o-spreadsheet .o-menu"));
 }
+
+const letterRegex = /^[a-zA-Z]$/;
+
+/**
+ * Transform a keyboard event into a shortcut string that represent this event. The letters keys will be uppercased.
```
rrahir, line 48: what's the rationale of uppercasing the letter keys but not the rest?

hokolomopo: Readability. Stuff like "PageDown" or "ArrowUp" is more readable in CamelCase than in all uppercase IMO.

PR #2690, opened 2023-07-18
[FIX] SortPlugin: fix crash whith empty-ish cells

Dj0ulo:
**Current behavior before PR:**
When sorting the cells (in `sortZone` method), it is editing the cells directly inside the loop. When it dispatches the "UPDATE_CELL" command, the content of the cell can be considered as empty and the cell will be removed as well as the corresponding `cellId` from the `SheetPlugin.cellPosition` array.
Therefore, it won't be able to retrieve the position of the cell in a future iteration of the loop to retrieve its value (with `this.getters.getCellPosition`)

**Description of the fix:**
Store the cell updates in an array to apply them all at once at the end.

OPW: : [3422772](https://www.odoo.com/web#id=3422772&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui/sort.ts
+++ b/src/plugins/ui/sort.ts
@@ -12,6 +12,7 @@ import {
   SortDirection,
   SortOptions,
   UID,
+  UpdateCellCommand,
   Zone,
 } from "../../types/index";
 import { UIPlugin } from "../ui_plugin";
@@ -280,18 +281,17 @@ export class SortPlugin extends UIPlugin {
     const sortedIndex: number[] = sortedIndexOfSortTypeCells.map((x) => x.index);
 
     const [width, height]: [number, number] = [cells.length, cells[0].length];
-
+    const cellUpdates: Omit<UpdateCellCommand, "type">[] = [];
     for (let c: HeaderIndex = 0; c < width; c++) {
       for (let r: HeaderIndex = 0; r < height; r++) {
         let cell = cells[c][sortedIndex[r]];
         let newCol: HeaderIndex = sortZone.left + c * stepX;
         let newRow: HeaderIndex = sortZone.top + r * stepY;
-        let newCellValues: any = {
+        let newCellValues: Omit<UpdateCellCommand, "type"> = {
           sheetId: sheetId,
           col: newCol,
           row: newRow,
           content: "",
-          value: "",
         };
         if (cell) {
           let content: string = cell.content;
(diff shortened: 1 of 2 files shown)


tests/plugins/sort.test.ts
```diff
@@ -337,6 +337,41 @@ describe("Basic Sorting", () => {
...
+        left: 0,
+        bottom: 2,
+        right: 0,
+      },
+      sortDirection: "ascending",
+      sortOptions: {
+        emptyCellAsZero: true,
+        sortHeaders: true,
+      },
+    });
```
LucasLefevre, line 370: You can use the test helper :wink:  It makes the test much easier to read with the range references as strings
```ts
    sort(model, {
      zone: "A2:A3",
      anchor: "A2",
      direction: "ascending",
      sortOptions: { emptyCellAsZero: true },
    });
```
Also, `sortHeaders` isn't relevent for this test, it can be omitted.

src/plugins/ui/sort.ts
```diff
@@ -307,9 +307,10 @@ export class SortPlugin extends UIPlugin {
           newCellValues.format = cell.format;
           newCellValues.value = cell.evaluated.value;
         }
-        this.dispatch("UPDATE_CELL", newCellValues);
+        cellUpdates.push(newCellValues);
       }
     }
+    cellUpdates.forEach((value) => this.dispatch("UPDATE_CELL", value));
```
LucasLefevre, line 313:
```suggestion
    cellUpdates.forEach((cmdPayload) => this.dispatch("UPDATE_CELL", cmdPayload));
```

src/plugins/ui/sort.ts
```diff
@@ -280,7 +280,7 @@ export class SortPlugin extends UIPlugin {
     const sortedIndex: number[] = sortedIndexOfSortTypeCells.map((x) => x.index);
 
     const [width, height]: [number, number] = [cells.length, cells[0].length];
-
+    const cellUpdates: any[] = [];
```
LucasLefevre, line 283: `any` should be use very very rarely in Typescript :wink: It removes all type safety the language provides.
And most of the time, in the places where you don't (need to) know the type, `unknown` is preferred over `any`

```suggestion
    const cellUpdates: Omit<UpdateCellCommand, "type">[]= [];
```

LucasLefevre (commented):
Thanks!
Some suggestions :wink:

tests/plugins/sort.test.ts
Dj0ulo: Sure, thank you ! I was actually hoping you would correct me here. I was not able to find the right combination of parameters with the test helper 😅

src/plugins/ui/sort.ts
Dj0ulo: No problem, it's just that the variable `newCellValues` has `any` type and contains the `value` attribute which is not in `UpdateCellCommand`

Dj0ulo: Done !

LucasLefevre (approved):
robodoo r+
Thanks!

PR #2755, opened 2023-08-03
[IMP] Translation: remove _lt function

anhe-odoo:
## Task Description

Following https://github.com/odoo/odoo/pull/130179, this PR aims to remove the use of _lt function, using now _t instead. To be able to take translations loading into account, a new function _loaded has been added to the setTranslationFunction method, indicating wether the translation has been loaded or not.

By default, we then have _t returning only lazy string, and once a translation has been loaded, it will return translated string directly.

## Related Task

- Task: : [3446734](https://www.odoo.com/web#id=3446734&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/translation.ts
+++ b/src/translation.ts
@@ -1,6 +1,5 @@
 /*
- * usage: every string should be translated either with _lt if they are registered with a registry at
- *  the load of the app or with Spreadsheet._t in the templates. Spreadsheet._t is exposed in the
+ * usage: every string should be translated with Spreadsheet._t in the templates. Spreadsheet._t is exposed in the
  *  sub-env of Spreadsheet components as _t
  * */
 
@@ -11,6 +10,7 @@ export type TranslationFunction = (
 
 // define a mock translation function, when o-spreadsheet runs in standalone it doesn't translate any string
 let _translate: TranslationFunction = (s) => s;
+let _loaded: () => boolean = () => false;
 
 function sprintf(s: string, ...values: string[] | [{ [key: string]: string }]): string {
   if (values.length === 1 && typeof values[0] === "object") {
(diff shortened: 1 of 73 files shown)


src/translation.ts
```diff
@@ -26,34 +26,33 @@ function sprintf(s: string, ...values: string[] | [{ [key: string]: string }]):
...
     super(str);
   }
 
   valueOf() {
     const str = super.valueOf();
-    return sprintf(_translate(str), ...this.values);
+    if (_loaded()) {
+      return sprintf(_translate(str), ...this.values);
+    } else {
+      throw new Error(`translation error`);
```
LucasLefevre, line 54: actually, I don't think it's a good thing to throw an error here. It prevents using the library without setting a translation function.
The setup should be minimal to use the lib and a translation function is definitely an advanced setup

src/translation.ts
```diff
@@ -26,34 +26,33 @@ function sprintf(s: string, ...values: string[] | [{ [key: string]: string }]):
  * Allow to inject a translation function from outside o-spreadsheet.
  * @param tfn the function that will do the translation
  */
-export function setTranslationMethod(tfn: TranslationFunction) {
+export function setTranslationMethod(tfn: TranslationFunction, loaded: () => boolean = () => true) {
```
LucasLefevre, line 29: docstring needs to be updated

anhe-odoo:
@LucasLefevre : I missed the comment about the docstring friday, but it should be okay now :)

LucasLefevre (approved):
robodoo r+

LucasLefevre (commented):
robodoo r-
wait

LucasLefevre (commented):
robodoo r+

PR #2757, opened 2023-08-04
[FIX] HeaderSizePlugin: Prevent state desynchronization

rrahir:
HOW TO REPRODUCE
----------------
- Go to the last row of a sheet
- Make sure that at least one cell contains a style (fillColor for
  instance)
- Insert a row after the last row
- duplicate the sheet

-> crash

The issue is a combination of 2 commits https://github.com/odoo/o-spreadsheet/commit/61f1c527fa549f38896be586514facb175e64021 and https://github.com/odoo/o-spreadsheet/commit/9086ac9b79532ef74ec14588850575f44e5bedf6.
Because of the order of the plugins, the headerSizePlugin would handle
the command `ADD_COLUMNS_ROWS` weirdly. This could prove specifically
prove problematic when inserting a **ROW** after a header containing
cells with some style.

The command is processed as such:
- CellPlugin handles the original command. Because the header after which we
  insert contains a cell with some style, it dispatches a subcommand
  `UPDATE_CELL` to propagate the style in the newly created row/column
- HeaderSizePlugin handles `UPDATE_CELL` and adds an entry to its state
  for the new row/column index of the updated cell, therefore adding 1
  more row to the internal mapping.
- HeaderSizePlugin handles the original command but it mistakenly
  assumes that its internal state contains the old quantity of rows
  but it was already updated when handling the `UPDATE_CELL` subcommand.

Task: 3450174

## Description:

Task: : [3450174](https://www.odoo.com/web#id=3450174&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/core/header_size.ts
+++ b/src/plugins/core/header_size.ts
@@ -105,7 +105,8 @@ export class HeaderSizePlugin extends CorePlugin<HeaderSizeState> implements Hea
         }
         break;
       case "UPDATE_CELL":
-        if (!this.sizes[cmd.sheetId]?.["ROW"]?.[cmd.row]?.manualSize) {
+        const row = this.sizes[cmd.sheetId]?.["ROW"]?.[cmd.row];
+        if (row && !row.manualSize) {
           const { sheetId, row } = cmd;
           this.history.update(
             "sizes",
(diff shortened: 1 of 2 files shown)


LucasLefevre (approved):
robodoo r+

PR #2786, opened 2023-08-09
[FIX] tests: avoid memory leak

LucasLefevre:
We are facing an important memory leak while running tests with recent versions of node.js (>=16.11)

`node --expose-gc ./node_modules/.bin/jest --runInBand --logHeapUsage`

Run this command with node 16.10, you'll see a ~ constant memory usage. Upgrade to any more recent node version, you'll see a steady increase of memory usage which can go as a high as several gigabytes.

This is a known issue which doesn't seem be properly adressed yet. https://github.com/jestjs/jest/issues/11956

Several alternatives:

Backport to node 16.10
----------------------

it works but
- everybody would need to use this exact version, including the CI platform (runbot).
- This version doesn't support lock file version 3, which we are currently using in recent o-spreadsheet versions
- we won't benefit from recent bug fixes, including security patches

Use jest `workerIdleMemoryLimit` flag
-------------------------------------

Jest introduced a flag as a work-around for the issue. It specifies a memory limit for workers before they are recycled.
It was only added in jest 29, that's the reason why the update to jest 29 was backported to v16 (see previous commit)
The current test suite takes around 450-650MB, it seems reasonable to set the limit to 800MB.

Use --no-compilation-cache
--------------------------

Run tests with `node --no-compilation-cache ./node_modules/jest/bin/jest.js` This was suggested on the jest issue. However, we prefer the most basic approach of the jest `workerIdleMemoryLimit` flag as it is much more naive and doesn't mess with node internals (--no-complation-cache doesn't seem to be officially documented).

Note:
we accept to live with the memory leak on ealier versions, we don't have a lot of bug fixes.

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -125,6 +125,7 @@
       "json",
       "node"
     ],
+    "workerIdleMemoryLimit": "800MB",
     "setupFilesAfterEnv": [
       "<rootDir>/tests/setup/jest.setup.ts"
     ]


rrahir:
@robodoo r+

PR #2813, opened 2023-08-22
[FIX] demo: console error messages

hokolomopo:
Error in the Demo component are caught and print in the console. Most of the errors are caught by Owl and wrapped, in which case we want to print the `error.cause` property. But some errors aren't caught by Owl and thus `error.cause` is undefined (eg. non-defined component in template).

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


demo/main.js
```diff
@@ -302,8 +302,12 @@ class Demo extends Component {
   }
 
   raiseError(content, callback) {
-    window.alert(content);
-    callback?.();
+    this.notifyUser({
```
rrahir, line 305: I'd rather have a specific behaviour to handle `unhandledrejection` rather than completely changing the behaviour of raiseError altogether.

rrahir (changes requested):

demo/main.js
hokolomopo: Good point, the raiseError is used in business code. I'll just change the `unhandledrejection` and the `onError`.

rrahir:
@robodoo r+

PR #2814, opened 2023-08-22
[IMP] chart: use locale in chart labels/tooltips

hokolomopo:
## Description

Since the introduction of locales, we can use them to correctly format the chart's labels/tooltips rather than relying on `toLocaleString`.

## Description:

Task: : [3473566](https://www.odoo.com/web#id=3473566&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/helpers/figures/charts/bar_chart.ts
```diff
@@ -230,9 +230,8 @@ function getBarConfiguration(
       ticks: {
         color: fontColor,
         callback: (value) => {
-          return localeFormat.format
-            ? formatValue(value, localeFormat)
-            : value?.toLocaleString() || value;
+          const { locale, format } = localeFormat;
+          return formatValue(value, { locale, format: !format && value > 1000 ? "#,##" : format });
```
rrahir, line 234: this will unfortunately not pass (starting typescript 5.1.6 at least)

rrahir (changes requested):

src/helpers/figures/charts/bar_chart.ts
hokolomopo: I'm surprised it worked before. Should be good now

PR #2819, opened 2023-08-24
[IMP] charts: add an 'extraData' key in definition

anhe-odoo:
## Task Description:

This PR aims to add a new extraData key in the chart definition to be able to add extraData when trying to extend the lib. For example, this key is now used in odoo to keep the link to an Odoo Menu when copying chart.

## Related Task/PR:

- Task: : [3380568](https://www.odoo.com/web#id=3380568&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## Review Checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/figures/charts/abstract_chart.ts
+++ b/src/helpers/figures/charts/abstract_chart.ts
@@ -23,11 +23,13 @@ export abstract class AbstractChart {
   readonly title: string;
   abstract readonly type: ChartType;
   protected readonly getters: CoreGetters;
+  readonly extraData?: any;
 
   constructor(definition: ChartDefinition, sheetId: UID, getters: CoreGetters) {
     this.title = definition.title;
     this.sheetId = sheetId;
     this.getters = getters;
+    this.extraData = definition.extraData;
   }
 
   /**
--- a/src/types/chart/scorecard_chart.ts
+++ b/src/types/chart/scorecard_chart.ts
@@ -10,6 +10,7 @@ export interface ScorecardChartDefinition {
   readonly background?: Color;
   readonly baselineColorUp: Color;
   readonly baselineColorDown: Color;
+  readonly extraData?: any;
 }
 
 export type BaselineMode = "text" | "difference" | "percentage";
--- a/src/helpers/figures/charts/bar_chart.ts
+++ b/src/helpers/figures/charts/bar_chart.ts
@@ -103,6 +103,7 @@ export class BarChart extends AbstractChart {
       type: "bar",
       verticalAxisPosition: "left",
       labelRange: context.auxiliaryRange || undefined,
+      extraData: context.extraData,
     };
   }
 
@@ -116,6 +117,7 @@ export class BarChart extends AbstractChart {
       auxiliaryRange: this.labelRange
         ? this.getters.getRangeString(this.labelRange, this.sheetId)
         : undefined,
+      extraData: this.extraData,
     };
   }
 
@@ -159,6 +161,7 @@ export class BarChart extends AbstractChart {
       title: this.title,
       stacked: this.stacked,
       aggregated: this.aggregated,
+      extraData: this.extraData,
     };
   }
 
--- a/src/helpers/figures/charts/gauge_chart.ts
+++ b/src/helpers/figures/charts/gauge_chart.ts
@@ -198,6 +198,7 @@ export class GaugeChart extends AbstractChart {
           value: "40",
         },
       },
+      extraData: context.extraData,
     };
   }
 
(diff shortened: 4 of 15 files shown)


LucasLefevre (commented):
THIS REQUIRES AN UPGRADE SCRIPT !!!! :grimacing: :grimacing: :grimacing: :grimacing: :grimacing: :grimacing: :grimacing: :grimacing: :sob:

src/types/chart/scorecard_chart.ts
```diff
@@ -10,6 +10,7 @@ export interface ScorecardChartDefinition {
   readonly background?: Color;
   readonly baselineColorUp: Color;
   readonly baselineColorDown: Color;
+  readonly extraData?: any;
```
hokolomopo, line 13: I'd add some tests to make sure the extradata is kept when duplicating the chart.  Otherwise looks fine :ok_hand:

src/helpers/figures/charts/abstract_chart.ts
```diff
@@ -23,11 +23,13 @@ export abstract class AbstractChart {
   readonly title: string;
   abstract readonly type: ChartType;
   protected readonly getters: CoreGetters;
+  readonly extraData?: any;
```
rrahir, line 26:
```suggestion
  readonly extraData?: {};
```
I think we could force it to be an object at least. We should also document this in the doc but I won't push for it right now, being the worst example to follow for this...

rrahir (changes requested):

LucasLefevre:
I'm feeling a bit uneasy with this. It feels like we had an adhoc solution to fight a deeper root cause (non deterministic ids, data synchronization done wrong).
It also defies the purpose of plugins where each plugin is (supposed to be) responsible for its own data.

Can't we think of an alternative solution ? I'll keep thinking about it in the back of my head.
The issue comes with "duplicate sheet" and copy/paste, right?

rrahir:
the issue I notice in the case of duplicate_sheet is that we lack the information to properly treat figures. We only give sheetFrom > sheetTo and that's it.However, figures are separate entities that need to process an internal state (in comparison with a cell or a col , row, w/e). So for cells, we have a 1 level of depth but for figures we have several as we rely on them to generate charts, images and attach menus to it. 

For this issue, I could suggest that we , finally, decide to mimic plugin structure of sheet   in the figure plugin (i.e. Record(sheetId, Object }) but we decided not to go in that direction because we thought it was a hassle to maintain that structure for all subjacent plugins (chart, image, menus). That being said, it does not address the copy/paste inside a same sheet.

Copy is basically the same because when you paste, at least the way it works now, we don't expose to the other plugins the origin of the paste. We dispatch CREATE_CHART with an identical payload and not something like DUPLICATE_CHART  with a from,to payload. I don't know if's that the solution but this means more commands, transformations to cover charts and images!  We definitely need to expose more context to the plugins that need to react to those commands based on their own internal state.

Another idea, we delegate the copy/cut/paste role to the different plugins and they store by themselves the information they require to duplicate themselves. This is a good solution for the data filters for example. But we need some kind of 'supreme ruler' that would decide the core information that every plugin would rely on (like the new chart id). this needs to be decided before we invoke each of these plugins callbacks so the clipboard plugin would still require to know what kind action we are doing and what to dispatch at the end.


Another point, I guess we are bound to go through commands (atm at least) to have the changes taken into account?



the main points to think about could be:
- state synchronization while they are historized independently from one another
- Odoo menu might be the only plugin which does not depend directly on the sheet but on an item inside the sheet
- self duplication with commands that lack execution context (we can only create, which works for the main plugin but not those which have a parallel state)

rrahir:
Forgot to add my opinion on the matter : if we are not ok with externalData (which seems like a band-aid,  I agree) then I would go with dispatching the responsibility to the plugins and admit that on paste, we'd have to predefine a certain number of arguments, like the Object id.

But I also think that we could address the first point I made about the internal plugin states of figure-related objects to simplify the duplication process (for DUPLICATE_SHEET)

hokolomopo:
Some thoughts:

- delegate the copy/cut/paste role to the different plugins sounds kinda nice. Two issues : 
    - pasting cells is not always a trivial "copy everything from zone A1:A2 to zone B1:B2. If we copy a single cell for exmaple, and the paste target is a zone, we will repeat the cell on each position of the zone. I'm not sure how this would look like in a "each plugin manage its own thing" architecture. 
    - that doesn't solve the issue of the ChartMenuPlugin. The plugin still have no clue what ids the newly duplicated figures have.
- I'm not even sure having commands `DUPLICATE_CHART, {fromChartId, toChartId}` works. Because you can copy a chart, then delete it, then paste it. In that case the `fromChartId`doesn't make sense anymore. Unless we never clean up deleted data.

I agree with having a record sheetId => chartId => chartData for the plugins tho. It would trivialise the sheet duplication.

anhe-odoo:
Closing as these changes are not needed anymore with the clipboard refactoring

PR #2821, opened 2023-08-25
[FIX] clipboard: copy/paste of CF in another sheet

hokolomopo:
## Description

There was an issue when copying/pasting a cell with a CF in another sheet. In the other sheet, a new CF was created with the same id as the original CF. This works fine when copy/pasting a CF for the first time. But after that, if the user were to modify the CF in the original sheet, a second copy/paste would overwrite the CF in the other sheet with the new changes.

This commit changes a bit the strategy for copy/pasting CFs in another sheet. Instead of always creating a new CF with the same id, the plugin tries to find a CF with the same rule in another sheet to put the copied cells into. If no CF with the same rule is found, then a new one with a new Id is created.

Task:

## Description:

Task: : [3479451](https://www.odoo.com/web#id=3479451&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


hokolomopo:
will be put in 15.0

PR #2827, opened 2023-08-28
[FIX] locale: string dates in CF aren't (un)localized

hokolomopo:
## Description

Using CF with date string works (eg. value is greater than "10/20/2012"), but the date string is not canonicalized when sending the CF rule in the model. This cause the CF to stop working when the locale is changed, because "10/20/2012" is not a valid date in french for example.

This commit fixes the issue by canonicalizing the date string before sending it to the model, and localizing it back to display it to the user.

The CF evaluation also used the current locale to evaluate the CF values. This was wrong and it should use the default locale, because CF values are canonicalized before sending them to the model.

Task:

## Description:

Task: : [3481670](https://www.odoo.com/web#id=3481670&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/link/link_editor/link_editor.ts
+++ b/src/components/link/link_editor/link_editor.ts
@@ -1,7 +1,7 @@
 import { Component, onMounted, useRef, useState } from "@odoo/owl";
 import { markdownLink } from "../../../helpers";
 import { detectLink, urlRepresentation } from "../../../helpers/links";
-import { canonicalizeContent } from "../../../helpers/locale";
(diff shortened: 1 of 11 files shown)


LucasLefevre (approved):
just a small comment... to add a comment ^^
LGTM otherwise :)

LucasLefevre (approved):
robodoo r+

PR #2830, opened 2023-08-28
[FIX] clipboard: copy/paste of CF in another sheet

hokolomopo:
## Description

There was an issue when copying/pasting a cell with a CF in another sheet. In the other sheet, a new CF was created with the same id as the original CF. This works fine when copy/pasting a CF for the first time. But after that, if the user were to modify the CF in the original sheet, a second copy/paste would overwrite the CF in the other sheet with the new changes.

This commit changes a bit the strategy for copy/pasting CFs in another sheet. Instead of always creating a new CF with the same id, the plugin tries to find a CF with the same rule in another sheet to put the copied cells into. If no CF with the same rule is found, then a new one with a new Id is created.

Task:

## Description:

Task: : [3479451](https://www.odoo.com/web#id=3479451&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/misc.ts
+++ b/src/helpers/misc.ts
@@ -248,3 +248,27 @@ export function isDefined<T>(argument: T | undefined): argument is T {
 }
 
 export const DEBUG: { [key: string]: any } = {};
+
+/**
+ * Compares two objects.
+ */
+export function deepEquals<T extends Object>(o1: T, o2: T): boolean {
+  if (o1 === o2) return true;
+  if ((o1 && !o2) || (o2 && !o1)) return false;
+
+  // Objects can have different keys if the values are undefined
+  const keys = new Set<string>();
+  Object.keys(o1).forEach((key) => keys.add(key));
+  Object.keys(o2).forEach((key) => keys.add(key));
+
+  for (let key of keys) {
+    if (typeof o1[key] !== typeof o1[key]) return false;
+    if (typeof o1[key] === "object") {
+      if (!deepEquals(o1[key], o2[key])) return false;
+    } else {
+      if (o1[key] !== o2[key]) return false;
+    }
+  }
+
+  return true;
+}
--- a/src/helpers/misc.ts
+++ b/src/helpers/misc.ts
@@ -252,9 +252,11 @@ export const DEBUG: { [key: string]: any } = {};
 /**
  * Compares two objects.
  */
-export function deepEquals<T extends Object>(o1: T, o2: T): boolean {
+export function deepEquals(o1: any, o2: any): boolean {
   if (o1 === o2) return true;
   if ((o1 && !o2) || (o2 && !o1)) return false;
+  if (typeof o1 !== typeof o2) return false;
+  if (typeof o1 !== "object") return o1 === o2;
 
   // Objects can have different keys if the values are undefined
   const keys = new Set<string>();
--- a/src/plugins/ui/evaluation_conditional_format.ts
+++ b/src/plugins/ui/evaluation_conditional_format.ts
@@ -1,6 +1,13 @@
 import { parsePrimitiveContent } from "../../helpers/cells";
-import { colorNumberString, isInside, recomputeZones, toXC, toZone } from "../../helpers/index";
-import { clip, isDefined } from "../../helpers/misc";
+import {
+  colorNumberString,
+  isInside,
+  recomputeZones,
+  toXC,
+  toZone,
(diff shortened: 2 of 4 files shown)


anhe-odoo (commented):
The fix loog good to me, nice idea :) I juste have a small comment for the deepEquals method and, as it can be a really useful method in the future, I think it can be a good idea to add some "general" tests for this methods :)

src/helpers/misc.ts
```diff
@@ -248,3 +248,31 @@ export function isDefined<T>(argument: T | undefined): argument is T {
 }
 
 export const DEBUG: { [key: string]: any } = {};
+
+/**
+ * Compares two objects.
+ */
+export function deepEquals<T extends Object>(o1: T, o2: T): boolean {
+  if (o1 === o2) return true;
+  if ((o1 && !o2) || (o2 && !o1)) return false;
```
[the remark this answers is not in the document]
hokolomopo: 1) Do XOR even exist in JS ?
2) It would probably be harder to read for anyone that didn't use XOR regularely (like me)

src/helpers/misc.ts
```diff
@@ -248,3 +248,31 @@ export function isDefined<T>(argument: T | undefined): argument is T {
...
+  for (let key of keys) {
+    if (typeof o1[key] !== typeof o1[key]) return false;
+    if (typeof o1[key] === "object") {
+      if (!deepEquals(o1[key], o2[key])) return false;
+    } else {
+      if (o1[key] !== o2[key]) return false;
+    }
+  }
+
+  return true;
```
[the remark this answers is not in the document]
hokolomopo: Yeah the method wasn't inteded to compare simple types at first, but `<T extends Object>` doen't prevent to use the method for simple types. Will add that + some tests

src/plugins/ui/evaluation_conditional_format.ts
```diff
@@ -484,11 +492,26 @@ export class EvaluationConditionalFormatPlugin extends UIPlugin {
...
+    const cfInTarget = this.getters
+      .getConditionalFormats(targetSheetId)
+      .find((cf) => areConditionalFormatEquivalents(originCF, cf));
+
+    if (cfInTarget) {
+      return cfInTarget;
+    }
+
+    const cf = { ...originCF, id: this.uuidGenerator.uuidv4(), ranges: [] };
+    this.dispatch("ADD_CONDITIONAL_FORMAT", { cf, target: [], sheetId: targetSheetId });
```
LucasLefevre, line 514: this method is called `getBrol` but actually contains a side-effect (dispatch).
I'd move the dispatch outside.

src/helpers/misc.ts
```diff
@@ -248,3 +248,33 @@ export function isDefined<T>(argument: T | undefined): argument is T {
...
+}
+
+export function areConditionalFormatEquivalents(cf1: ConditionalFormat, cf2: ConditionalFormat) {
```
LucasLefevre, line 278: I'd keep it inside the CF file if it's used only there (and it's quite specific to CF)

tests/plugins/clipboard.test.ts
```diff
@@ -1292,6 +1292,41 @@ describe("clipboard", () => {
...
+    const sheet2Id = model.getters.getSheets()[1].id;
+
+    const cf = createEqualCF("2", { fillColor: "#00FF00" }, "cfId");
+    model.dispatch("ADD_CONDITIONAL_FORMAT", {
+      cf,
+      target: target("A1"),
+      sheetId: sheet1Id,
+    });
+
+    model.dispatch("COPY", { target: [toZone("A1")] });
```
LucasLefevre, line 1308:
```suggestion
    model.dispatch("COPY", { target: target("A1") });
```
To be consistent with 3 lines above ^^

LucasLefevre (commented):
Nice :)
Only minor details

src/plugins/ui/evaluation_conditional_format.ts
hokolomopo: The best part is that this dispatch is 100% useless, it is not dispatched because the allowDispatch prevent dispatching a CF with empty ranges :upside_down_face:

LucasLefevre (approved):
robodoo r+ rebase-ff :)

PR #2853, opened 2023-09-05
[FIX] GridOverlay: Hovering over a figure should not open popovers

dhrp-odoo:
## Description:

Previously, when hovering over the chart, the cell popover would erroneously open. This issue occurred because `updateMousePosition` assigned the `clientX` and `clientY` values, and `getPosition` found the cell's position, creating a bug when hovering over the chart.

To address this problem, a conditional check has been added to ensure that `gridRef` is equal to the event target.

Task: : [3475001](https://www.odoo.com/web#id=3475001&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/grid_overlay/grid_overlay.ts
+++ b/src/components/grid_overlay/grid_overlay.ts
@@ -47,9 +47,11 @@ function useCellHovered(
     }
   }
   function updateMousePosition(e: MouseEvent) {
-    x = e.offsetX;
-    y = e.offsetY;
-    lastMoved = Date.now();
+    if (gridRef.el === e.target) {
+      x = e.offsetX;
+      y = e.offsetY;
+      lastMoved = Date.now();
+    }
   }
 
   function recompute() {
--- a/tests/components/figure.test.ts
+++ b/tests/components/figure.test.ts
@@ -31,7 +31,7 @@ let fixture: HTMLElement;
 let model: Model;
 let sheetId: UID;
 
-function createFigure(
+export function createFigure(
   model: Model,
   figureParameters: Partial<CreateFigureCommand["figure"]> = {},
   sheetId: UID = model.getters.getActiveSheetId()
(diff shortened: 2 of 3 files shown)


src/components/grid_overlay/grid_overlay.ts
```diff
@@ -59,17 +59,23 @@ function useCellHovered(
     }
   }
 
+  function shouldUpdateMousePosition(e: MouseEvent) {
+    if (gridRef.el === e.target) {
+      return updateMousePosition(e);
+    }
```
hokolomopo, line 65: Put the condition inside `updateMousePosition`. 
1) it's 2 lines
2) we expect a function named "should..." to return a boolean, not to update the position if needed

tests/components/figure.test.ts
```diff
@@ -40,8 +40,8 @@ function createFigure(
     id: "someuuid",
     x: 1,
     y: 1,
-    height: 100,
-    width: 100,
+    height: 200,
+    width: 200,
```
hokolomopo, line 44: use the `figureParameters` argument instead of modifying the default value and having to change a random test that has nothing to do with what you're fixing...

tests/components/grid.test.ts
```diff
@@ -868,6 +869,18 @@ describe("error tooltip", () => {
     expect(document.querySelector(".o-error-tooltip")).not.toBeNull();
   });
 
+  test("Hovering over a figure should not open popovers", async () => {
+    createFigure(model);
+    await nextTick();
+    setCellContent(model, "C3", "[label](url.com)");
+
+    triggerMouseEvent(".o-figure", "mousemove", 192, 46); // C3 cell is at 192, 46
```
hokolomopo, line 877: If we ever change `DEFAULT_CELL_WIDTH/HEIGHT` this will break. Always use constants when possible.

dhrp-odoo: @hokolomopo, I've put into effect the recommended suggestions.

hokolomopo (commented):
@robodoo r+

PR #2854, opened 2023-09-06
[FIX] grid: hover cell wrongly used event.offset

hokolomopo:
## Description

The `useCellHovered` hook used the `event.offsetX` and `event.offsetY` to determine the hovered cell. This is wrong because the offset is relative to the target element, and thus was wrong when the user hovered something like a chart.

Fix this by using the `event.clientX` and `event.clientY` and the grid position instead.

Task:

## Description:

Task: : [3495630](https://www.odoo.com/web#id=3495630&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/grid_overlay/grid_overlay.ts
+++ b/src/components/grid_overlay/grid_overlay.ts
@@ -9,6 +9,7 @@ import {
   SpreadsheetChildEnv,
 } from "../../types";
 import { FiguresContainer } from "../figures/figure_container/figure_container";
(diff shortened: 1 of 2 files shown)


hokolomopo:
duplicate of #2853

PR #2857, opened 2023-09-07
[REF] Clipboard: refactor copy-paste handling

anhe-odoo:
## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/plugins/core/conditional_format.ts
```diff
@@ -57,6 +70,203 @@ export class ConditionalFormatPlugin
 
   readonly cfRules: { [sheet: string]: ConditionalFormatInternal[] } = {};
 
+  constructor(config: CorePluginConfig) {
+    super(config);
+    clipboardHandlersRegistries.cellHandlers.add("CONDITIONAL_FORMAT", {
+      copy: (
```
LucasLefevre, line 76: I'm actually wondering it this needs to/should be defined here in the (core) plugins themselves. It could also be defined elsewhere (as long as it's extensible for each plugin).
Something that looks weird is they are given the UI getters and the UI dispatch

src/plugins/core/conditional_format.ts
```diff
@@ -57,6 +70,203 @@ export class ConditionalFormatPlugin
...
+        getters: Getters,
+        dispatch: CommandDispatcher["dispatch"],
+        isCutOperation: boolean,
+        data: ClipboardData
+      ): any => {
```
LucasLefevre, line 81: :eyes: 
Let typescript infer it or use `unknown`

src/plugins/core/conditional_format.ts
```diff
@@ -57,6 +70,203 @@ export class ConditionalFormatPlugin
...
+        );
+
+        return {
+          cellPositions,
+          isCutOperation,
+        };
+      },
+      paste: (
+        target: Zone[],
+        clippedContent: any,
```
LucasLefevre, line 118: Could be a generic. Something like
```ts

interface ClipBrol<T> {
  copy(): T
  paste(data: T): void
}
```

LucasLefevre (commented):
Some comments/ideas/questions
we can also discuss it live :)

src/plugins/core/cell.ts
```diff
@@ -58,6 +77,484 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
...
+        }
+      });
+    });
+  }
+
+  /**
+   * Add columns and/or rows to ensure that col + width and row + height are still
+   * in the sheet
+   */
+  protected addMissingDimensions(
```
LucasLefevre, line 341: each clipboard handler could return its own target zone.
The clipboard can then add the missing cols/rows and select the union of zone at the end

src/plugins/core/cell.ts
```diff
@@ -58,6 +77,484 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
   ] as const;
   readonly nextId = 1;
   public readonly cells: { [sheetId: string]: { [id: string]: Cell } } = {};
+  private selection: SelectionStreamProcessor | undefined;
+
+  constructor(config: CorePluginConfig) {
+    super(config);
+    clipboardHandlersRegistries.cellHandlers.add("CELL", {
```
LucasLefevre, line 84: move out of core.

src/helpers/clipboard/clipboard_os_state.ts
```diff
@@ -18,18 +19,49 @@ export class ClipboardOsState extends ClipboardCellsAbstractState {
   constructor(
     content: string,
```
LucasLefevre, line 20: `paste` of `ClipboardOsState` shouldn't be the same as paste of the cell handler ? They both are pasting cells after all.
I believe the `content` (raw content of the os clipboard) should be converted to the same data structure as what the cell handler `copy` would give.

LucasLefevre (commented):
I believe we can go even further :wink: We should try removing all `Clipboard[...]State` and "helper" classes (such as `ClipboardFigureImage`). Even if it means giving back more responsibility to the clipboard plugin.

IMO, there should be only two "things": the clipboard plugin and the handlers, no other intermediary abstractions.

src/clipboard_handlers/figures.ts
```diff
@@ -0,0 +1,90 @@
...
+    switch (figure.tag) {
+      case "chart":
+        copiedFigureContent = new ClipboardFigureChart(
+          this.dispatch,
+          this.getters,
+          sheetId,
+          data.figureId
+        );
+        break;
+      case "image":
```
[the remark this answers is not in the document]
anhe-odoo: It should be available in today's update, with a new function called on the handlers ´getPasteTarget` that should return the new figure id, and this is will the be given when calling the `paste` method, so every handler will be able to know if there is a new figure

src/clipboard_handlers/conditional_format.ts
```diff
@@ -0,0 +1,127 @@
...
+  copy(isCutOperation: boolean, data: ClipboardData): any {
+    if (!("zones" in data) || !data.zones.length) {
+      return {};
+    }
+
+    const { rowsIndex, columnsIndex } = getCellPositions(data.zones);
+    const sheetId = this.getters.getActiveSheetId();
+    const cellPositions = rowsIndex.map((row) =>
+      columnsIndex.map((col) => ({ col, row, sheetId }))
+    );
```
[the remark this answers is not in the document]
anhe-odoo: In fact the helper positions is already used to generate rowIndex and columnsIndex in the getCellPositions helper, but (I think) we have to use this helper to handle some special case with multiple zones selection, and then cannot only rely on the zones only

src/clipboard_handlers/borders.ts
```diff
@@ -0,0 +1,105 @@
...
+      }
+    } else {
+      for (const zone of target) {
+        for (let col = zone.left; col <= zone.right; col++) {
+          for (let row = zone.top; row <= zone.bottom; row++) {
+            this.pasteZone(sheetId, col, row, content);
+          }
+        }
+      }
+    }
```
hokolomopo, line 82: Do you think this logic can be factorized and handled directly in the clipboard ? My idea would be to make the handlers.paste only paste once, but eh Clipboard would call multiple time handler.paste, once for every repeated zone.

Or if it's not possible/ugly to put that in the clipboard, at least have a helper `pasteFromCopy(pasteZone: Function, target, width, height)`, I feel like every handler have the exact same loop.

Actually I don't think the Filters have this repeating behaviour ATM, but I'm pretty sure it's more of an oversiht thatn anything else

src/clipboard_handlers/cell.ts
```diff
@@ -0,0 +1,289 @@
...
+    if (!("zones" in data) || !data.zones.length) {
+      return;
+    }
+    const sheetId = this.getters.getActiveSheetId();
+    const zones = data.zones;
+    if (!zones.length) {
+      return {
+        cells: [[]],
+        zones: [],
+        isCutOperation,
```
hokolomopo, line 41: `isCutOperation` should probably be handled by the clipboard and given as argument to the paste, rather than each handler having to store it in their state

src/clipboard_handlers/cell.ts
```diff
@@ -0,0 +1,289 @@
...
+    target: ClipboardPasteTarget,
+    content: ClipboardContent,
+    options?: ClipboardOptions | undefined
+  ) {
+    if (!("cells" in content)) {
+      return;
+    }
+    if (!("zones" in target) || !target.zones.length) {
+      return;
+    }
```
hokolomopo, line 124: single if ?

src/clipboard_handlers/chart.ts
```diff
@@ -0,0 +1,111 @@
...
+    }
+    if (option?.pasteOption !== undefined) {
+      return CommandResult.WrongFigurePasteOption;
+    }
+    return CommandResult.Success;
+  }
+
+  isCutAllowed(data: ClipboardData) {
+    return CommandResult.Success;
+  }
```
hokolomopo, line 110: useless

src/clipboard_handlers/cell.ts
hokolomopo: same in almost every `handler.paste`

src/clipboard_handlers/cell.ts
```diff
@@ -0,0 +1,289 @@
...
+    content: ClipboardContent,
+    clipboardOption?: ClipboardOptions
+  ): CommandResult {
+    if (!("cells" in content)) {
+      return CommandResult.Success;
+    }
+    if (content.isCutOperation && clipboardOption?.pasteOption !== undefined) {
+      // cannot paste only format or only value if the previous operation is a CUT
+      return CommandResult.WrongPasteOption;
+    }
```
hokolomopo, line 100: So we can paste format only from a cut operation if there is no cells in the copy area ? Something seems strange here ...

src/plugins/ui_stateful/clipboard.ts
```diff
@@ -55,18 +60,32 @@ export class ClipboardPlugin extends UIPlugin {
     switch (cmd.type) {
       case "CUT":
         const zones = this.getters.getSelectedZones();
-        const state = this.getClipboardState(zones, cmd.type);
-        return state.isCutAllowed(zones);
+        return this.isCutAllowedOnAnyPlugin(zones);
+      //@ts-ignore
```
hokolomopo, line 64: 🙃

src/plugins/ui_stateful/clipboard.ts
```diff
@@ -55,18 +60,32 @@ export class ClipboardPlugin extends UIPlugin {
...
+        if (!this.copiedData) {
           return CommandResult.EmptyClipboard;
         }
         const pasteOption =
           cmd.pasteOption || (this.paintFormatStatus !== "inactive" ? "onlyFormat" : undefined);
-        return this.state.isPasteAllowed(cmd.target, { pasteOption });
-      case "PASTE_FROM_OS_CLIPBOARD": {
-        const state = new ClipboardOsState(cmd.text, this.getters, this.dispatch, this.selection);
-        return state.isPasteAllowed(cmd.target, { pasteOption: cmd.pasteOption });
+        return this.isPasteAllowedOnAnyPlugin(cmd.target, this.copiedData, { pasteOption });
```
hokolomopo, line 73: handler, not plugin :) Actually I think you can juste change your method names to `this.isPasteAllowed`, `this.copy`, `this.paste` etc...

src/plugins/ui_stateful/clipboard.ts
```diff
@@ -82,21 +101,6 @@ export class ClipboardPlugin extends UIPlugin {
         }
         break;
       }
-      case "INSERT_CELL": {
```
hokolomopo, line 85: nitpick: changing the order of switch cases beaks the git history

src/plugins/ui_stateful/clipboard.ts
```diff
@@ -171,57 +189,51 @@ export class ClipboardPlugin extends UIPlugin {
...
+          this.copiedData = undefined;
         }
         break;
       }
       case "REMOVE_COLUMNS_ROWS": {
         this.status = "invisible";
 
-        // If we remove a col/row inside or before the cut area, we invalidate the clipboard
-        if (this.state?.operation !== "CUT") {
+        // If we add a col/row inside or before the cut area, we invalidate the clipboard
```
hokolomopo, line 221: we do be removing col/rows here, not adding them

src/plugins/ui_stateful/clipboard.ts
```diff
@@ -260,6 +272,183 @@ export class ClipboardPlugin extends UIPlugin {
     }
   }
 
+  private convertOSClipboardData(text: string): ClipboardData {
```
hokolomopo, line 275: Maybe this should be in the handlers. It seems kinda strange that the clipboard does have to know the internal state of the CellHandler. And in the future we will add more things handled in the OS clipoard, not just simple text cells

src/plugins/ui_stateful/selection.ts
```diff
@@ -685,7 +682,8 @@ export class GridSelectionPlugin extends UIPlugin {
         bottom: !isCol ? base + thickness - 1 : this.getters.getNumberRows(cmd.sheetId) - 1,
       },
     ];
-    state.paste(pasteTarget, { selectTarget: true });
+    // @ts-ignore
```
hokolomopo, line 685: 🙃

hokolomopo (commented):
Looks good 🥳 I didn't read into much details the implemntation  of copy/paste of each handler, I trust in the tests and your copy/paste skills 😉

src/clipboard_handlers/cell.ts
anhe-odoo: I'm not sure about what you mean. The "if !cells..." is there to check if we have copied cells or figure. Where there is no cell in the content, we won't call the paste of cell.ts, so we won't copy any format.

src/helpers/clipboard/clipboard_helpers.ts
```diff
@@ -0,0 +1,57 @@
...
+  const columnsIndex = [...new Set(cellsPosition.map((p) => p.col))].sort((a, b) => a - b);
+  const rowsIndex = [...new Set(cellsPosition.map((p) => p.row))].sort((a, b) => a - b);
+  return { zones, clippedZones, columnsIndex, rowsIndex };
+}
+
+/**
+ * The clipped zone is copied as many times as it fits in the target.
+ * This returns the list of zones where the clipped zone is copy-pasted.
+ */
+export function splitZone(target: Zone[], splitWidth: number, splitHeight: number): Zone[] {
```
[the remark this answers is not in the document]
anhe-odoo: You're right :) It's done

src/plugins/ui_stateful/clipboard.ts
```diff
@@ -135,8 +155,9 @@ export class ClipboardPlugin extends UIPlugin {
             bottom: multipleRowsInSelection ? zone.top : zone.top - 1,
             top: multipleRowsInSelection ? zone.top : zone.top - 1,
           };
-          const state = this.getClipboardStateForCopyCells([copyTarget], "COPY");
-          state.paste([zone], {
+          this.originSheetId = this.getters.getActiveSheetId();
+          this.copiedData = this.copy("COPY", [copyTarget]);
+          this.paste([zone], {
             pasteOption: undefined,
             shouldPasteCF: true,
```
LucasLefevre, line 162: `shouldPasteCF` is still "une entorce" to a generic clipboard (which is not supposed to know exactly what is copy/pasted). If it's not trivial to fix, let's keep it that way for now and merge.

src/plugins/ui_stateful/clipboard.ts
```diff
@@ -42,10 +47,10 @@ export class ClipboardPlugin extends UIPlugin {
   ] as const;
 
   private status: "visible" | "invisible" = "invisible";
-  private state?: ClipboardState;
-  private lastPasteState?: ClipboardState;
   private paintFormatStatus: "inactive" | "oneOff" | "persistent" = "inactive";
   private originSheetId?: UID;
+  private copiedData?: any;
```
LucasLefevre, line 52: Let's avoid `any`. It also always backfires

LucasLefevre (commented):
quick review while waiting at the garage :)
It's looking good :+1:

src/plugins/ui_stateful/clipboard.ts
```diff
@@ -42,10 +47,10 @@ export class ClipboardPlugin extends UIPlugin {
...
+  private _isCutOperation?: boolean;
```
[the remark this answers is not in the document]
anhe-odoo: The idea of this was to keep the `isCutOperation` getters, which is used in some other parts of the code (in the menu actions for example). Should I use remove the getter and use another way to get the information elsewhere ?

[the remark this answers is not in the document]
LucasLefevre: Ah oui, j'avais pas vu le conflit

LucasLefevre (approved):
robodoo r+

PR #2858, opened 2023-09-11
[FIX] SelectionInput: Unable to select range after confirm on empty selection input

aath-odoo:
## Description:

Previously, when a user cleared all values from the selection input and then clicked 'Confirm,' they were unable to select a range again, and the 'Confirm' and 'Reset' buttons would also disappear. This issue has now been fixed.

Task: [3493185](https://www.odoo.com/web#id=3493185&cids=2&menu_id=4720&action=333&active_id=2328&model=project.task&view_type=form)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


LucasLefevre (commented):
robodoo r+

src/plugins/ui_feature/selection_input.ts
```diff
@@ -65,6 +65,11 @@ export class SelectionInputPlugin extends UIPlugin implements StreamCallbacks<Se
           return CommandResult.MaximumRangesReached;
         }
         break;
+      case "REMOVE_RANGE":
+        if (this.ranges.length === 1) {
+          return CommandResult.MinimumRangesReached;
+        }
+        break;
```
aath-odoo, line 72: Hello @LucasLefevre,  It seems that we need to remove all the ranges during the reset process because it is currently causing a bug. Please take a look at this video for a demonstration of the issue (https://www.awesomescreenshot.com/video/20884856?key=a1a0e236cbff5d2bca764059673df95b) :)

aath-odoo: Hello @LucasLefevre , I attempted to address the issue, but it appears that for the reset operation, we must remove all ranges. However, the current implementation with REMOVE_RANGE in allowDispatch() is causing a bug as it won't let us remove range if there's a single range which eventually will always be true at some point in case of reset. 
Therefore, I'm wondering if it would be feasible to prevent the command from dispatching instead of handling it in allowDispatch to enable it to work for reset. Alternatively, we can consider introducing an optional flag in the REMOVE_RANGE command to distinguish between reset and confirm operations. or else if you have any alternate suggestion please let me know, Thanks :)

rrahir:
@robodoo r+

PR #2868, opened 2023-09-12
[FIX] clipboard: cut/paste whole CF across sheets

hokolomopo:
## Description

When cutting a whole conditional format and pasting it in another sheet, the CF wasn't removed from the first sheet.

Task: : [3502777](https://www.odoo.com/web#id=3502777&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui/evaluation_conditional_format.ts
+++ b/src/plugins/ui/evaluation_conditional_format.ts
@@ -459,6 +459,10 @@ export class EvaluationConditionalFormatPlugin extends UIPlugin {
 
     currentRanges = currentRanges.concat(toAdd);
     const newRange: string[] = recomputeZones(currentRanges, toRemove);
+    if (newRange.length === 0) {
+      this.dispatch("REMOVE_CONDITIONAL_FORMAT", { id: cf.id, sheetId });
+      return;
+    }
     this.dispatch("ADD_CONDITIONAL_FORMAT", {
       cf: {
         id: cf.id,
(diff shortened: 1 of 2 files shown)


src/plugins/ui/evaluation_conditional_format.ts
```diff
@@ -459,6 +459,10 @@ export class EvaluationConditionalFormatPlugin extends UIPlugin {
 
     currentRanges = currentRanges.concat(toAdd);
     const newRange: string[] = recomputeZones(currentRanges, toRemove);
+    if (newRange.length === 0) {
+      this.dispatch("REMOVE_CONDITIONAL_FORMAT", { id: cf.id, sheetId });
+      return;
+    }
```
anhe-odoo, line 465: Instead of adding the "REMOVE" command before the "ADD" command, can't we simply make the ADD command correctly remove the previous range ?

hokolomopo: The point here is to remove the CF if it has no ranges (we don't dispatch ADD command afterward). 

The previous implementation dispatched a ADD_CF with empty ranges, which was rejected by the allowDispatch. Which is fine, because we don't wan't to be allowed to create a new CF with no ranges.

anhe-odoo: My bad, I missed the "return" after the remove. Ok for me then :)

rrahir:
@robodoo r+

PR #2873, opened 2023-09-13
[FIX] headers_overlay: reselect full column or row as dataseries range 

khpa-odoo:
## Description:

Steps to reproduce:
- Click on header "B" and insert a chart.
- Click on selection input to edit the chart data series and then click on
  header "A". The data is now "A:A".
- Click on header "B" again. The data series should be updated with "B:B",
  but nothing happens instead.

This issue is created since the boolean `waitingForMove` is true when a
column is currently selected, and a user tries to reselect that column
while editing chart dataseries. In this case, `startMovement` method is called
followed by an early return and hence `startSelection` is not called.

The root cause of this bug is that the `startMovement` feature should
only be activated by the active grid selection and not while using
`SelectionInput` or selecting a range while editing a cell. This commit
introduces a new getter `isGridSelectionActive` in order to identify if
some features should be accessible or not to the end user.

An ulterior pull request will address the other features that should be
deactivated while not using the grid selection, (like autofill, merge,
...)

Task: : [3481617](https://www.odoo.com/web#id=3481617&menu_id=4720&cids=2&action=333&active_id=2328&model=project.task&view_type=form)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/headers_overlay/headers_overlay.ts
+++ b/src/components/headers_overlay/headers_overlay.ts
@@ -197,7 +197,13 @@ abstract class AbstractResizer extends Component<ResizerProps, SpreadsheetChildE
       return;
     }
     if (this.state.waitingForMove === true) {
-      this.startMovement(ev);
+      if (!this.env.model.getters.isGridSelectionActive()) {
+        this._selectElement(index, false);
+      } else {
+        // FIXME: Consider reintroducing this feature for all type of selection if we find
+        // a way to have the grid selection follow the other selections evolution
+        this.startMovement(ev);
+      }
       return;
     }
     if (this.env.model.getters.getEditionMode() === "editing") {
--- a/src/plugins/ui/selection.ts
+++ b/src/plugins/ui/selection.ts
@@ -107,6 +107,7 @@ export class GridSelectionPlugin extends UIPlugin {
     "getSheetPosition",
     "isSelected",
     "getElementsFromSelection",
+    "isGridSelectionActive",
   ] as const;
 
   private gridSelection: {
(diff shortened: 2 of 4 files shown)


rrahir:
The issue is not about the selection input, it's rather about some features being specifically dedicated to the "grid selection" and not the other (composer or selection input as it turns out there are only 2 other selection streams). Such features should probably be disabled when not dealing with the grid selection. This means going further in the fix and I think it's important to discuss this with the team.

2 generic commands relative to the PR form:
The commit title is wrong - the fix clearly does not take place in selection input...
The commit description does not represent the actual issue. The manipulation you describe here works perfectly...
Please pay attention to this. Documenting a change properly is more important than the change itself.

khpa-odoo:
@rrahir Regarding the commit message, I've changed it and I'm sorry for the oversight... I'll be cautious henceforth!

About disabling some features when not dealing with 'grid selection', I'll be awaiting the team's consensus to know the desired results before proceeding further with the fix :-)

src/components/headers_overlay/headers_overlay.ts
```diff
@@ -197,6 +197,9 @@ abstract class AbstractResizer extends Component<ResizerProps, SpreadsheetChildE
       return;
     }
     if (this.state.waitingForMove === true) {
+      if (!this.env.model.getters.isGridSelectionActive()) {
+        this._selectElement(index, false);
+      }
       this.startMovement(ev);
```
khpa-odoo, line 203: @rrahir Are we really disabling `startMovement` as mentioned in commit description?

Although IMO, `startMovement` should not be disabled to allow drag and drop of cols/rows while using 'SelectionInput'. (Gsheet allows it too)

LucasLefevre: Yes because it really is moving the columns/rows, not just the selection.
I agree it could be nice if it would move the selection (in the selection input), but this is a stable release, we can't reinvent the feature. It can be an improvement for master (but it's not a priority)

VincentSchippefilt:
robodoo r+

rrahir:
@robodoo r+

PR #2914, opened 2023-09-22
[FIX] Find&Replace: purge duplicated handler

rrahir:
The forward port commit e03958bf introduced a duplication in the switch statement of the command handler of F&R plugin.

Task: /


LucasLefevre (approved):
robodoo r+

PR #2929, opened 2023-09-25
[FIX] Charts: fix incoherent panel state

anhe-odoo:
## Description:

This task aims to address some issues in the side panel of a chart:
1. When having only one point in a dataseries and selecting "First
row as header", the point disappear from the chart (legit), but
we don't have the possibility to uncheck this checkbox.
2. When having a normal dataset and another dataset containing one
point, checking the "... as header" checkbox works has expected, but
unchecking it don't make the 'one-point' dataset come back

Related Task:

Task: 3380568
## Related Task(s):
- Task: [3380568](https://www.odoo.com/web#id=3380568&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## Review Checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/charts/chart_common.ts
+++ b/src/helpers/charts/chart_common.ts
@@ -182,13 +182,8 @@ export function createDataSets(
           )
         );
       }
-    } else if (zone.left === zone.right && zone.top === zone.bottom) {
-      // A single cell. If it's only the title, the dataset is not added.
-      if (!dataSetsHaveTitle) {
-        dataSets.push(createDataSet(getters, dataSetSheetId, zone, undefined));
-      }
     } else {
-      /* 1 row or 1 column */
+      /* 1 cell, 1 row or 1 column */
       dataSets.push(
         createDataSet(
           getters,
--- a/tests/components/charts.test.ts
+++ b/tests/components/charts.test.ts
@@ -1259,6 +1259,53 @@ describe("figures", () => {
       await simulateClick(".o-menu div[data-name='edit']");
       expect(document.querySelector("input[name='labelsAsText']")).toBeFalsy();
     });
+
+    test("Side panel correctly reacts to has_header checkbox check/uncheck (with only one point)", async () => {
+      createTestChart("basicChart");
+      updateChart(model, chartId, { type: "line", labelRange: "C2", dataSets: ["A1"] });
+      await nextTick();
+      await simulateClick(".o-figure");
+      await simulateClick(".o-chart-menu-item");
+      await simulateClick(".o-menu div[data-name='edit']");
+
+      const checkbox = document.querySelector("input[name='labelsAsText']") as HTMLInputElement;
+      expect(checkbox.checked).toBe(false);
+
+      await simulateClick(checkbox);
+      expect(checkbox.checked).toBe(true);
+    });
+
+    test("Side panel correctly reacts to has_header checkbox check/uncheck (with two datasets)", async () => {
+      createTestChart("basicChart");
+      updateChart(model, chartId, { type: "line", labelRange: "C2", dataSets: ["A1:A2", "A1"] });
+      await nextTick();
+      await simulateClick(".o-figure");
+      await simulateClick(".o-chart-menu-item");
+      await simulateClick(".o-menu div[data-name='edit']");
+
+      const checkbox = document.querySelector("input[name='labelsAsText']") as HTMLInputElement;
+      expect(checkbox.checked).toBe(false);
+
+      expect(checkbox.checked).toBe(false);
+      expect((model.getters.getChartDefinition(chartId) as LineChartDefinition).dataSets).toEqual([
+        "A1:A2",
+        "A1",
+      ]);
+
+      await simulateClick(checkbox);
+      expect(checkbox.checked).toBe(true);
+      expect((model.getters.getChartDefinition(chartId) as LineChartDefinition).dataSets).toEqual([
+        "A1:A2",
+        "A1",
+      ]);
+
+      await simulateClick(checkbox);
+      expect(checkbox.checked).toBe(false);
+      expect((model.getters.getChartDefinition(chartId) as LineChartDefinition).dataSets).toEqual([
+        "A1:A2",
+        "A1",
+      ]);
+    });
   });
 });
 
--- a/tests/plugins/chart/__snapshots__/basic_chart.test.ts.snap
+++ b/tests/plugins/chart/__snapshots__/basic_chart.test.ts.snap
@@ -741,7 +741,21 @@ Object {
   "background": "#FFFFFF",
   "chartJsConfig": Object {
     "data": Object {
-      "datasets": Array [],
+      "datasets": Array [
+        Object {
+          "backgroundColor": "#1F77B4",
+          "borderColor": "rgb(31,119,180)",
+          "data": Array [
+            undefined,
+            undefined,
+            undefined,
+          ],
+          "fill": false,
+          "label": "30",
+          "lineTension": 0,
+          "pointBackgroundColor": "rgb(31,119,180)",
+        },
+      ],
       "labels": Array [
         "P4",
         "P5",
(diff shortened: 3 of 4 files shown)


src/helpers/figures/charts/bar_chart.ts
```diff
@@ -110,47 +116,47 @@ export class BarChart extends AbstractChart {
...
+    const definition = this.getDefinitionWithSpecificDataSets(dataSets, this.labelRange, sheetId);
     return new BarChart(definition, sheetId, this.getters);
   }
 
   getDefinition(): BarChartDefinition {
-    return this.getDefinitionWithSpecificDataSets(this.dataSets, this.labelRange);
+    const dataSets = this.getDataSets(false).map((ds: DataSet) =>
+      this.getters.getRangeString(ds.dataRange, this.sheetId)
+    );
+    return this.getDefinitionWithSpecificDataSets(dataSets, this.labelRange);
```
hokolomopo, line 147: Isn't that the same as just returning this.definition that was given to the constructor ? The goal was to return the raw definiton without truncating the datasets right ? If yes we can remove the ugly boolean argument in this.getDataSets() and make it a getter (or init at constructor)

tests/components/charts.test.ts
```diff
@@ -597,6 +599,216 @@ describe("charts", () => {
...
+
+      const figures = fixture.querySelectorAll(".o-figure");
+      await simulateClick(figures[0] as HTMLElement);
+      await simulateClick(".o-figure-menu-item");
+      await simulateClick(".o-menu div[data-name='edit']");
+      await nextTick();
+
+      const dataSeries = fixture.querySelectorAll(
+        ".o-sidePanel .o-sidePanelBody .o-chart .o-data-series"
+      )[0];
```
hokolomopo, line 620: you could create directly dataSeriesInput to avoid querySelector("input") every time

same in the other tests

hokolomopo (commented):
Seems to work :+1:  I'm not gonna comment on the SelectionInput changes, as long as it works I don't care, we should burn this whole things to the ground and code something better.

Needs a rebase :)

src/components/side_panel/chart/gauge_chart_panel/gauge_chart_config_panel.ts
```diff
@@ -22,8 +23,22 @@ export class GaugeChartConfigPanel extends Component<Props, SpreadsheetChildEnv>
...
 
+  setup() {
+    this.state.lastUpdatedDataRange = this.dataRange;
+    onWillUpdateProps((nextProps) => {
+      if (
+        this.dataRange === this.props.definition.dataRange &&
+        this.dataRange !== nextProps.definition.dataRange
+      ) {
+        this.dataRange = nextProps.definition.dataRange;
+        this.state.lastUpdatedDataRange = this.dataRange;
```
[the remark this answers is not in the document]
anhe-odoo: No, they are not always the same, as this.dataRange can be changed by the selectionInput

src/helpers/figures/charts/bar_chart.ts
anhe-odoo: No it's not the same as this._definition. We have to go through through this.getDataSets to "clean" the datasets (removing the sheetName in the range string if it's the same as the chart sheet name, removing invalid datasets, ...)

src/helpers/charts/abstract_chart.ts
```diff
@@ -18,16 +18,16 @@ import { Validator } from "../../types/validator";
...
+export abstract class AbstractChart<T extends ChartDefinition = ChartDefinition> {
   readonly title: string;
   abstract readonly type: ChartType;
-  protected readonly getters: CoreGetters;
 
-  constructor(definition: ChartDefinition, sheetId: UID, getters: CoreGetters) {
-    this.title = definition.title;
-    this.sheetId = sheetId;
-    this.getters = getters;
+  constructor(protected _definition: T, readonly sheetId: UID, readonly getters: CoreGetters) {
```
hokolomopo, line 25:
```suggestion
  constructor(protected _definition: T, readonly sheetId: UID,  protected readonly getters: CoreGetters) {
```

src/helpers/charts/bar_chart.ts
```diff
@@ -200,7 +199,7 @@ function getBarConfiguration(chart: BarChart, labels: string[]): ChartConfigurat
   const legend: ChartLegendOptions = {
     labels: { fontColor },
   };
-  if ((!chart.labelRange && chart.dataSets.length === 1) || chart.legendPosition === "none") {
+  if ((!chart.labelRange && chart.getDataSets().length === 1) || chart.legendPosition === "none") {
```
hokolomopo, line 202: we can probably give the datasets as argument of the method, to avoit building them 2 times

hokolomopo (commented):
👍

LucasLefevre (commented):
is it possible to split the commit in 2 commits? one for each issue described. Or is it chiant ?
It's easier to review when each change is isolated. It's also easier for the future when someone will try to understand it.

anhe-odoo:
@LucasLefevre : the fix is the same for both issues, but the code has been simplified as discussed together so it should be easier to review :)

src/helpers/charts/bar_chart.ts
```diff
@@ -142,7 +147,7 @@ export class BarChart extends AbstractChart {
   }
 
   getDefinition(): BarChartDefinition {
-    return this.getDefinitionWithSpecificDataSets(this.dataSets, this.labelRange);
+    return this.getDefinitionWithSpecificDataSets(this.fullDataSets, this.labelRange);
```
LucasLefevre, line 150: I don't understand the fix anymore :/
in a `DataSet`, `dataRange` includes `labelRange`...so it should end up being the same dataset string :thinking: 
I'll wait until you come back

LucasLefevre: Some data is lost inside `createDataSets` when there's a single cell range and it includes (only) the title => the dataset is dropped (it should not)


https://github.com/odoo/o-spreadsheet/pull/2442

src/helpers/charts/chart_common.ts
```diff
@@ -182,11 +182,6 @@ export function createDataSets(
           )
         );
       }
-    } else if (zone.left === zone.right && zone.top === zone.bottom) {
-      // A single cell. If it's only the title, the dataset is not added.
-      if (!dataSetsHaveTitle) {
-        dataSets.push(createDataSet(getters, dataSetSheetId, zone, undefined));
-      }
     } else {
       /* 1 row or 1 column */
```
LucasLefevre, line 186: also restore the previous comment :wink: 
```suggestion
      /* 1 cell, 1 row or 1 column */
```

src/helpers/charts/chart_common.ts
```diff
@@ -182,11 +182,6 @@ export function createDataSets(
           )
         );
       }
-    } else if (zone.left === zone.right && zone.top === zone.bottom) {
```
LucasLefevre, line 185: In the end, all the diff in this PR except this one is useless right ? (and the tests)
Could you revert the useless diff ?
Even if some changes are clearer, it's not worth it here IMO (especially for a bug fix in stable). Useless changes makes the review more difficult, forward-ports (and future forward-ports) more likely to conflict.

As a general rule of thumb, keep the diff minimal for bug fixes in stable :wink:  Refactorings are better in master and in different commits (not burried in a bug fix or a new feature)

LucasLefevre (commented):
only some cleaning I think :wink:

src/helpers/charts/chart_common.ts
anhe-odoo: Yes, you're right. I though the other diff with the `this.dataSetsHaveTitle` things were necessary, but not with this change :)

LucasLefevre (commented):
robodoo r+
Thanks :)

PR #2932, opened 2023-09-25
[IMP] evaluation: functions can handle errors

laa-odoo:
With those commits, functions can handle errors more easily. See individual commits for details.

The purpose of this commit is to allow the formulas
to return errors as results.

Functionally we want formulas like TRANSPOSE to be
able to transpose an error without invalidating the
entire evaluation process.

We also want formulas like LOOKUP not to be interrupted
when the search range has one or more errors.

Task: 2969401


src/functions/index.ts
```diff
@@ -107,6 +108,40 @@ function createComputeFunctionFromDescription(
...
+  if (e instanceof EvaluationError) {
+    e.message = e.message.replace("[[FUNCTION_NAME]]", functionName);
+    return { value: e };
+  }
+  // TODO check this
+  const error = new EvaluationError(
+    e.message.replace("[[FUNCTION_NAME]]", functionName),
+    e?.errorType,
+    e.logLevel !== undefined ? e.logLevel : CellErrorLevel.error
+  );
```
LucasLefevre, line 141: can be simplified. `e?.errorType` and `e.logLevel` are always undefined because we know `e` is not an `EvaluationError`
```ts
const error = new EvaluationError(e.message.replace("[[FUNCTION_NAME]]", functionName));
```

src/functions/module_info.ts
```diff
@@ -1,22 +1,19 @@
...
-  compute: function (value: () => Maybe<CellValue>): boolean {
-    try {
-      value();
-      return false;
-    } catch (e) {
-      return e?.errorType != CellErrorType.NotAvailable;
-    }
+  compute: function (value: Maybe<CellValue>): boolean {
+    const isErr = value instanceof EvaluationError;
+    const isna = value instanceof NotAvailableError;
```
LucasLefevre, line 15: `isNa`

tests/test_helpers/helpers.ts
```diff
@@ -638,10 +639,13 @@ export function getCellsObject(model: Model, sheetId: UID): Record<string, CellO
   const cells: Record<string, CellObject> = {};
   for (const cell of Object.values(model.getters.getCells(sheetId))) {
     const { col, row } = model.getters.getCellPosition(cell.id);
+
+    const evaluatedCell = model.getters.getEvaluatedCell({ sheetId, col, row });
```
LucasLefevre, line 643: revert this useless diff

src/plugins/ui_core_views/cell_evaluation/evaluation_plugin.ts
```diff
@@ -219,7 +220,13 @@ export class EvaluationPlugin extends UIPlugin {
     }
     const array = compiledFormula.execute(ranges, ...this.compilationParams);
     if (isMatrix(array)) {
-      return array.map((col) => col.map((row) => row.value));
+      throw new Error("results array are not supported outside the evaluation grid");
+    }
+    if (array.value instanceof EvaluationError) {
+      return array.value.errorType;
+    }
```
hokolomopo, line 227: Seems strange to only here treat error as strings, and not at any other part of the evaluation process.

Also if we want to do that we should test instanceof `Error` (or improve the wrapping of errors into EvaluationError) , because for example WrongNumberOfArgs will return an Error and not an EvaluationError

tests/evaluation/evaluation.test.ts
```diff
@@ -165,17 +165,19 @@ describe("evaluateCells", () => {
   test("error in some function calls", () => {
     const model = new Model();
     setCellContent(model, "A1", '=Sum("asdf")');
-    expect(getEvaluatedCell(model, "A1").value).toBe("#ERROR");
-    expect(getCellError(model, "A1")).toBe(
```
hokolomopo, line 169: why these change? That's a lot of diff and forward port breaking when nothing really had to change, no ? `getEvaluatedCell(model, "A1").value` didn't change, and I'm not sure there is a reason to change `getCellError`

src/plugins/ui_core_views/cell_evaluation/evaluator.ts
```diff
@@ -210,17 +216,21 @@ export class Evaluator {
     }
 
     const cellId = cell.id;
-
+    const localeFormat = { format: cell.format, locale: this.getters.getLocale() };
     try {
```
LucasLefevre, line 220: A next step would be to change every throw in this file to returns.

src/plugins/ui_core_views/cell_evaluation/evaluator.ts
```diff
@@ -507,7 +530,7 @@ class PositionBitsEncoder {
   private decodeSheet(sheetKey: PositionId): UID {
     const sheetId = this.inverseSheetMapping.get(sheetKey);
     if (sheetId === undefined) {
-      throw new Error("Sheet id not found");
+      throw { value: ERR_GENERIC, message: "Sheet id not found" };
```
LucasLefevre, line 533: This is a real programming error. It should remain an `Error`

src/plugins/ui_core_views/evaluation_data_validation.ts
```diff
@@ -228,9 +226,11 @@ export class EvaluationDataValidationPlugin extends UIPlugin {
         );
 
         const evaluated = this.getters.evaluateFormula(sheetId, translatedFormula);
-        return evaluated && !isMatrix(evaluated) ? evaluated.toString() : "";
+        return evaluated?.toString() || "";
```
LucasLefevre, line 229: what if it's a matrix? Is it safe to remove the check?

src/types/cells.ts
```diff
@@ -25,6 +24,7 @@ export type Cell = LiteralCell | FormulaCell;
 
 interface EvaluatedCellProperties {
   readonly format?: Format;
+  readonly message?: string;
```
LucasLefevre, line 27: I don't think we should have `message` here since it's specific to a cell value and we do have a specific type for it.
```suggestion
```

src/plugins/ui_core_views/cell_evaluation/compilation_parameters.ts
```diff
@@ -88,29 +88,30 @@ class CompilationParametersBuilder {
...
-      throw new Error(_t("Invalid sheet name"));
+      throw { value: ERR_GENERIC, message: _t("Invalid sheet name: %s") };
     }
     const evaluatedCell = this.getEvaluatedCellIfNotEmpty(position);
     if (evaluatedCell === undefined) {
       return { value: null, format: this.getters.getCell(position)?.format };
     }
-    return evaluatedCell;
+    const { value, format, message } = evaluatedCell;
+    return { value, format, message };
```
LucasLefevre, line 110: Let's avoid creating useless objects.
```suggestion
    return evaluatedCell
```

LucasLefevre (commented):
Nice !
I think we finally found a good way ! Good job :)

Some general comment (more detailed in individual comments)
- if we find a more meaningful name for `UnitData`, it would be nice
- throwing objects (and not Errors) makes me feel :raised_eyebrow: but :shrug: 
- I think there's some useless changes that could be reverted.
- Now that the functional side is working with this PR, a next step is to improve performance by using returns as much as possible (especially for `ODOO.*` functions)

src/formulas/code_builder.ts
```diff
@@ -3,10 +3,6 @@
  */
```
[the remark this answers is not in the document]
laa-odoo: fixed

src/functions/index.ts
```diff
@@ -74,7 +75,7 @@ class FunctionRegistry extends Registry<FunctionDescription> {
     }
     const descr = addMetaInfoFromArg(addDescr);
     validateArguments(descr.args);
-    this.mapping[name] = createComputeFunctionFromDescription(descr);
+    this.mapping[name] = addErrorHandling(createComputeFunctionFromDescription(descr), name);
```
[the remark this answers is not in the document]
laa-odoo: Indeed, to avoid that,  we can put the tryCatch directly in `buildComputeFunctionFromDescription`. @LucasLefevre are you agree with that ?

src/functions/helpers.ts
```diff
@@ -9,7 +9,7 @@ const SORT_TYPES_ORDER = ["number", "string", "boolean", "undefined"];
 
 export function assert(condition: () => boolean, message: string): void {
   if (!condition()) {
-    throw new EvaluationError(CellErrorType.GenericError, message);
+    throw new EvaluationError(message, CellErrorType.GenericError);
```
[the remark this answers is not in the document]
laa-odoo: That’s what we’re trying to get towards. The "asserts" are intended to be replaced by "returns" with string errors

src/helpers/cells/cell_evaluation.ts
```diff
@@ -60,31 +56,28 @@ export function createEvaluatedCell(
...
 
-function _createEvaluatedCell(value: CellValue | null, localeFormat: LocaleFormat): EvaluatedCell {
-  try {
-    for (const builder of builders) {
-      const evaluateCell = builder(value, localeFormat);
-      if (evaluateCell) {
-        return evaluateCell;
-      }
+function _createEvaluatedCell(value: CellValue, localeFormat: LocaleFormat): EvaluatedCell {
+  if (value instanceof EvaluationError) {
```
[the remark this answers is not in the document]
laa-odoo: I find it easier to read, less code to have in mind.
Actually, if it really bothers you, I'll go back.

tests/evaluation/evaluation.test.ts
laa-odoo: Indeed

src/helpers/cells/cell_evaluation.ts
```diff
@@ -155,46 +160,13 @@ function booleanCell(value: boolean, localeFormat: LocaleFormat): BooleanCell {
...
+    value,
+    message,
     isAutoSummable: false,
     defaultAlign: "center",
-    formattedValue: error.errorType,
+    formattedValue: value,
   };
 }
-
-const builders: EvaluatedCellBuilder[] = [
```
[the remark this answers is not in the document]
laa-odoo: we go from 51 lines to 15,
OK the previous code is elegant, it also gives the possibility of easily extending the code to the creation of new cell types.
I may be wrong but I don't think we're considering it.
Here, with 15 lines, in one vision, we understand all the possibilities.
But really I don't care, if you really want to go back then I will

src/plugins/ui_core_views/evaluation_data_validation.ts
laa-odoo: I don't know what we want exactly:

Here I removed the check because the try catch below takes care of this case.
The `evaluateFormula` getter has been modified to throw when an array is received.
Perhaps we should first ask ourselves what exactly we want to do with this getters when it returns an array ?

src/functions/helpers.ts
```diff
@@ -233,6 +230,9 @@ export function visitNumbers(args: ArgValue[], cb: (arg: number) => void, locale
       if (typeof cellValue === "number") {
         cb(cellValue);
       }
+      if (cellValue instanceof Error) {
+        throw cellValue;
+      }
```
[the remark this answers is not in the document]
laa-odoo: my bad, but it doesn't matter much, it is still managed correctly after

src/functions/module_array.ts
```diff
@@ -584,6 +584,23 @@ const TO_COL_ROW_ARGS = [
   ),
 ];
 
+function dataToKeep(ignore: number): (data: ValueAndFormat) => boolean {
```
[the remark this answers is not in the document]
laa-odoo: ok thx :)

src/functions/module_array.ts
```diff
@@ -594,20 +611,12 @@ export const TOCOL = {
     scanByColumn: Maybe<ValueAndFormat> = { value: TO_COL_ROW_DEFAULT_SCAN }
   ) {
     const _array = toMatrix(array);
-    const _ignore = toInteger(ignore.value, this.locale);
+    const _ignore = toNumber(ignore.value, this.locale);
```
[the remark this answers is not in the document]
laa-odoo: Done in the `shouldKeepValue` helper

src/functions/module_array.ts
```diff
@@ -594,20 +611,12 @@ export const TOCOL = {
...
     const _scanByColumn = toBoolean(scanByColumn.value);
 
-    assert(() => _ignore >= 0 && _ignore <= 3, _t("Argument ignore must be between 0 and 3"));
```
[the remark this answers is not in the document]
laa-odoo: indeed, but this will change again in a commit to turn all "assert" into "return"

src/functions/module_statistical.ts
```diff
@@ -66,6 +66,12 @@ function filterAndFlatData(
       flatDataY.push(valueY);
       flatDataX.push(valueX);
     }
+    if (valueY instanceof Error) {
```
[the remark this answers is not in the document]
laa-odoo: Currently all `visitAny` seem to throw when there are errors so yes, 
but let's hope we never need to use it on ranges with errors

src/functions/helpers.ts
```diff
@@ -73,41 +95,37 @@ export function tryToNumber(
...
-export function tryCastAsNumberMatrix(data: Matrix<any>, argName: string): Matrix<number> {
-  data.forEach((row) => {
-    row.forEach((cell) => {
-      if (typeof cell !== "number") {
-        throw new Error(
-          _t(
+export function toNumberMatrix(data: Matrix<UnitData>, argName: string): Matrix<number> {
+  return data.map((row) => {
+    return row.map((cell) => {
+      if (typeof cell.value !== "number") {
```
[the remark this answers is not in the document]
laa-odoo: `toNumber` will cast values (specially string values), 
and generally we don't want cast values when they come from matrix args

src/functions/module_filter.ts
```diff
@@ -133,19 +133,19 @@ export const FILTER = {
...
       _t("The arguments conditions must have the same dimensions."),
-      ..._conditionsMatrices
+      ...conditions
     );
     const _conditions = _conditionsMatrices.map((c) => c.flat());
 
     const mode = _conditionsMatrices[0].length === 1 ? "row" : "col";
     _array = mode === "row" ? transposeMatrix(_array) : _array;
-
+    conditions;
```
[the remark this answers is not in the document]
laa-odoo: thx

src/types/functions.ts
```diff
@@ -20,18 +19,19 @@ export type ArgType =
...
   type: ArgType[];
   default?: boolean;
   defaultValue?: any;
 }
 
-export type ComputeFunctionArg<T> = T | (() => T);
-export type ComputeFunction<T, R> = (this: EvalContext, ...args: ComputeFunctionArg<T>[]) => R;
+// export type ComputeFunctionArg<T> = T | (() => T);
+// export type ComputeFunction<T, R> = (this: EvalContext, ...args: ComputeFunctionArg<T>[]) => R;
+export type ComputeFunction<T, R> = (this: EvalContext, ...args: T[]) => R;
```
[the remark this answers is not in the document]
laa-odoo: indeed

src/functions/module_array.ts
```diff
@@ -340,22 +326,16 @@ export const MINVERSE = {
...
       _matrix
     );
-    if (!isNumberMatrix(_matrix)) {
-      throw new Error(_t("The argument square_matrix must be a matrix of numbers."));
-    }
-
     const { inverted } = invertMatrix(_matrix);
     if (!inverted) {
-      throw new Error(_t("The matrix is not invertible."));
+      throw { value: ERR_GENERIC, message: _t("The matrix is not invertible.") };
```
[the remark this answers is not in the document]
laa-odoo: Here, all the `throw` in `compute` are cases where `compute` is typed as returning a matrix

src/plugins/ui_core_views/cell_evaluation/evaluator.ts
laa-odoo: yes, my bad

src/types/misc.ts
```diff
@@ -181,13 +179,11 @@ export interface RangeCompiledFormula extends Omit<CompiledFormula, "dependencie
 }
 
 export type Matrix<T = unknown> = T[][];
-export type ValueAndFormat = { value: CellValue; format?: Format };
+export type UnitData = { value: CellValue; format?: Format; message?: string };
```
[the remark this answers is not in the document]
pro-odoo: `FunctionResult`

src/functions/index.ts
```diff
@@ -74,81 +73,60 @@ class FunctionRegistry extends Registry<FunctionDescription> {
...
-    );
-  }
+function addErrorHandling(
+  compute: ComputeFunction<Arg, Matrix<UnitData> | UnitData>,
+  functionName: string
+): ComputeFunction<Arg, Matrix<UnitData> | UnitData> {
+  return function (this: EvalContext, ...args: Arg[]): Matrix<UnitData> | UnitData {
+    try {
+      const computeFormula = compute.bind(this);
+      return computeFormula(...args);
```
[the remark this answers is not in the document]
laa-odoo: it is an ideal that we will try to achieve

src/types/misc.ts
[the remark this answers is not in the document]
laa-odoo: `FPayload`

src/plugins/ui_core_views/evaluation_data_validation.ts
LucasLefevre: Ok understood. After live discussion: `this.getters.evaluateFormula` shouldn't throw

src/formulas/compiler.ts
```diff
@@ -64,10 +64,10 @@ export function compileTokens(tokens: Token[]): CompiledFormula {
     const scope = new Scope();
 
     if (ast.type === "BIN_OPERATION" && ast.value === ":") {
-      throw new BadExpressionError(_t("Invalid formula"));
+      throw { value: CellErrorType.BadExpression, message: _t("Invalid formula") };
```
anhe-odoo, line 67: Personnally I prefere the old way to throw/return error. Why did you change this ?

laa-odoo: the goal would be that the formulas end up returning errors in the form of the normal payload rather than throwing errors, (performance reason)

LucasLefevre: I must admit I agree with Anthony. It feels very here where we are not in the context of a function implementation that would return a value (or an error).

laa-odoo: `Bad Expression Error(message)` and other errors could indeed result in `{value: "#SomeError", message}`

LucasLefevre (commented):
Yééé nice :)
I think we can merge tomorrow :)
I only left only minor comments (most of them translations)
I did not check commit by commit

LucasLefevre:
robodoo rebase-merge r+

LucasLefevre:
et maintenant ?
robodoo rebase-merge r+

LucasLefevre:
robodoo r+

PR #2968, opened 2023-10-05
[IMP] chart: export gauge and scorecard

anhe-odoo:
## Task Description

o-spreadsheet now allows to insert images in a spreadsheet, and we also allow to export them inside .xlsx file. However, while this works as intended in a standalone o-spreadsheet server, it doens't work correctly in Odoo as the data of the image are not found while we try ton convert the spreadsheet to an xlsx file. This PR aims to simplify the request made to get the binary data of the image file.

Moreover, this PR also add the export of Odoo chart as an image. The way these image are exported is slightly different than classical image, as the chart is converted to base64 data_url, and then directly converted from base64 to binary string, while classical image are loaded from a file to a binary string.

## Related tas/PR
- Task: : [3524473](https://www.odoo.com/web#id=3524473&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)
- https://github.com/odoo/enterprise/pull/47928

## Review Checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/helpers/figures/charts/scorecard_chart_designer.ts
```diff
@@ -0,0 +1,377 @@
...
+  style: {
+    font: string;
+    color: Color;
+    strikethrough?: boolean;
+    underline?: boolean;
+  };
+  position: PixelPosition;
+};
+
+export type ScorecardChartDesign = {
```
LucasLefevre, line 37: Now there are 3 data structures: definition, runtime and this. There's no clear difference between the last two. And there's no clear reason why they both exists.
Runtime should become this new structure.
And all this should be moved to `createScorecardChartRuntime`

src/helpers/figures/charts/scorecard_chart_designer.ts
```diff
@@ -0,0 +1,377 @@
...
+  private height: number;
+
+  constructor(figure: Figure, readonly runtime: ScorecardChartRuntime) {
+    const canvas = document.createElement("canvas");
+    this.width = canvas.width = figure.width;
+    this.height = canvas.height = figure.height;
+    this.context = canvas.getContext("2d")!;
+  }
+
+  computeDesign(): ScorecardChartDesign {
```
LucasLefevre, line 80: this method is way too long. Split it in smaller meaningful parts

LucasLefevre (commented):
quick partial review

src/helpers/figures/charts/scorecard_chart_designer.ts
anhe-odoo: I give it a try, but each part need information coming from previous part for the positioning of new element. I tried to make it shorter (reduced from 120 lines to 80) and clearer.

rrahir:
gauche > gauge ;-)

src/helpers/figures/charts/scorecard_chart_designer.ts
LucasLefevre: @anhe-odoo @hokolomopo 
We were discussing in private with @rrahir about that. He has a point about this split.
- it's true that runtime and designer being different things can make sense (one depends on the model, gets the colors, texts, etc., the other takes that structure and computes a more specialized structure with positions to draw on the canvas)
- however, this is the same thing for other (chart.js) charts: the runtime is already this specialized structure ready.

For consistency with other charts (chart.js), it's better to keep every flow (basically) the same. It will help the reader to understand how it works: he understands the flow for one chart type, he understands all charts :)

Merging scorecard runtime and designer doesn't mean we can't keep things neat and tidy in `createScorecardChartRuntime`

What do you think?

anhe-odoo:
@LucasLefevre : Don't know why but I can't reply directly on your last comment ...

I totally agree with Remi's idea to merge it. He was also going further with the idea of creating a single "Chart" component that could handle a runtime and draw the correct chart on a canvas, which, in my opinion, would be a good thing. But I don't have any idea of the cost vs benefits of this.

src/helpers/figures/charts/scorecard_chart_designer.ts
```diff
@@ -0,0 +1,376 @@
...
+export function getScorecardDesign(
+  figure: Figure,
+  runtime: ScorecardChartRuntime
+): ScorecardChartDesign {
+  const designer = new ScorecardChartDesigner(figure, runtime);
+  return designer.computeDesign();
+}
+
+class ScorecardChartDesigner {
+  private context;
```
rrahir, line 77:
```suggestion
  private context: CanvasRenderingContext2D;
```

src/helpers/figures/charts/scorecard_chart.ts
```diff
@@ -177,6 +179,102 @@ export class ScorecardChart extends AbstractChart {
...
+    context.moveTo(position.x, position.y - textHeight / 2 + lineWidth);
+    context.lineTo(position.x + textWidth, position.y - textHeight / 2 + lineWidth);
+    context.stroke();
+  }
+}
+
+export function drawScoreChart(structure: ScorecardChartDesign, canvas: HTMLCanvasElement) {
+  // Set canvas size
+  const ctx = canvas.getContext("2d")!;
+  canvas.width = structure.canvas.width;
```
rrahir, line 215: Now that we have the canvas that will take the full width/height of the container, I would like to get rid of the figure props altogether since it's only present for this specific type of charts.
Here we can infer the element size `ctx.canvas.getBoundingClientRect`.
So ctx could be passed as an argument to bother `drawScoreChart` and `ScorecardChartDesign`.
@LucasLefevre wdyt?

src/helpers/figures/charts/scorecard_chart_designer.ts
rrahir: nitpick ,the name kinda collides with ScorecardChartDesignPanel

tests/xlsx_export.test.ts
```diff
@@ -886,60 +891,63 @@ describe("Test XLSX export", () => {
...
-          type: "pie",
-        },
-        "1"
-      );
+      createScorecardChart(model, TEST_CHART_DATA.scorecard);
+      expect(getExportedExcelData(model).sheets[0].charts.length).toBe(0);
+      expect(getExportedExcelData(model).sheets[0].images.length).toBe(1);
+    });
+
+    test("Gauche Chart is exported as an image", () => {
```
rrahir, line 944: right

rrahir (commented):
First review, I haven't gone through the tests yet

src/helpers/figures/charts/scorecard_chart.ts
rrahir: > EDIT: image needs the figure id to get the image url :/
I meant for figureChart specifically ;-)

src/components/figures/chart/scorecard/chart_scorecard.ts
```diff
@@ -1,294 +1,51 @@
...
-    const str = chart.keyValue || "";
-    return this.measureTextWidth(ctx, str, fontSize);
+    let previousRuntime = this.runtime;
+    let previousWidth = this.props.figure.width;
+    let previousHeight = this.props.figure.height;
+    onPatched(() => {
+      const runtime = this.runtime;
+      const figure = this.props.figure;
+      if (
+        deepEquals(runtime, previousRuntime) &&
```
LucasLefevre, line 32: I think we can compare by reference instead of a deep comparison.
It would allow to use `useEffect` which does the comparison (by reference) for us.
The entire `onPatch` and `onMounted` could be replaced by:
```ts
useEffect(
      this.createChart.bind(this),
      () => [this.runtime, this.props.figure.width, this.props.figure.height, this.canvas.el]
);
```
Note that technically, `this.canvas.el` is also a dependency

src/helpers/figures/charts/scorecard_chart.ts
```diff
@@ -177,6 +179,103 @@ export class ScorecardChart extends AbstractChart {
...
+
+  // Draw  title
+  if (structure.title) {
+    ctx.font = structure.title.style.font;
+    ctx.fillStyle = structure.title.style.color;
+    ctx.fillText(structure.title.text, structure.title.position.x, structure.title.position.y);
+  }
+
+  // Draw baseline
+  if (structure.baseline) {
```
LucasLefevre, line 231: those comments are useless. The code is clear and easy enough.

src/helpers/figures/charts/scorecard_chart_config_builder.ts
```diff
@@ -0,0 +1,379 @@
...
+  bold: boolean | undefined = false,
+  italic: boolean | undefined = false
+): string {
+  const italicStr = italic ? "italic" : "";
+  const weight = bold ? "bold" : "";
+  return `${italicStr} ${weight} ${fontSize}px ${DEFAULT_FONT}`;
+}
+
+export function getScorecardConfiguration(
+  figure: Figure,
```
rrahir, line 69: I still have a problem with this. My take was that I really don't think that we should have to pass figure as a props to the chart component because the chart should be adapted to its container size without having to know what it's container is.
I'd rather have getScorecardConfiguration take {width, height} as argument.

src/components/figures/chart/scorecard/chart_scorecard.ts
```diff
@@ -1,294 +1,51 @@
...
   }
 
-  getElementMaxFontSize(availableHeight: Pixel, chart: ScorecardChart): number {
-    if (!chart.runtime) return 0;
-    const haveBaseline = chart.baseline !== "" || chart.baselineDescr;
-    const maxHeight = haveBaseline ? KEY_BOX_HEIGHT_RATIO * availableHeight : availableHeight;
-    return maxHeight / LINE_HEIGHT;
+  private createChart() {
+    const canvas = this.canvas.el as HTMLCanvasElement;
+    const config = getScorecardConfiguration(this.props.figure, this.runtime);
```
rrahir, line 47: extract the canvas dimension and pass it to `getScorecardConfiguration`

tests/figures/chart/scorecard_chart_component.test.ts
```diff
@@ -65,21 +42,35 @@ async function updateScorecardChartSize(width: number, height: number) {
...
-  let text = "";
-  for (let span of spans) {
-    text += span.textContent;
-  }
-  return text;
+function getFontSize(font: string) {
+  return Number(font.match(/([0-9\.]*)px/)?.[1]);
+}
+
+function getChartDesign(chartId: UID): ScorecardChartConfig {
```
LucasLefevre, line 51: The model could/should be a parameter instead of using the global `model`

src/components/figures/chart/scorecard/chart_scorecard.ts
rrahir, line 32: nitpick, I'd check the dimensions first.
Again, by not considering figure but rather the canvas size.

src/helpers/figures/charts/scorecard_chart.ts
rrahir: Bump on this one - we have to pass the chartId (resp. imageId) to `ChartFigure` (resp. ImageFigure) but that's it. the figure object is not required. I really think we can get rid of it and use the canvas as the container reference to compute the scorecard  config.

rrahir (changes requested):

src/helpers/figures/charts/scorecard_chart_config_builder.ts
```diff
@@ -0,0 +1,382 @@
...
+  bold: boolean | undefined = false,
+  italic: boolean | undefined = false
+): string {
+  const italicStr = italic ? "italic" : "";
+  const weight = bold ? "bold" : "";
+  return `${italicStr} ${weight} ${fontSize}px ${DEFAULT_FONT}`;
+}
+
+export function getScorecardConfiguration(
+  { width, height }: { width: Pixel; height: Pixel },
```
rrahir, line 69: DOMDimension

src/plugins/ui_feature/renderer.ts
```diff
@@ -26,7 +26,8 @@ import {
   computeTextFont,
   computeTextFontSizeInPixels,
   computeTextLinesHeight,
-  computeTextWidth,
+  drawDecoratedText,
+  //computeTextWidth,
```
rrahir, line 30: comment

src/plugins/ui_feature/renderer.ts
```diff
@@ -290,7 +291,7 @@ export class RendererPlugin extends UIPlugin {
   }
 
   private drawTexts(renderingContext: GridRenderingContext) {
-    const { ctx, thinLineWidth } = renderingContext;
+    const { ctx /*, thinLineWidth*/ } = renderingContext;
```
rrahir, line 294: comment

tests/figures/chart/scorecard_chart_component.test.ts
```diff
@@ -273,57 +264,238 @@ describe("Scorecard charts", () => {
...
+     * Be careful when writing new test using this, as the switch will mismatch elements
+     * that have the same text (ie baseline text = baselineDescr text = title text, ...).
+     * Be sure, when using this, to have all elements that are different in the test.
+     */
+    jest
+      .spyOn(MockCanvasRenderingContext2D.prototype, "fillText")
+      .mockImplementation(function (this: MockCanvasRenderingContext2D, text: string) {
+        const runtime = model.getters.getChartRuntime(chartId) as ScorecardChartRuntime;
+        switch (text) {
+          case runtime.baselineDisplay:
```
rrahir, line 405: I think you missed this one :) https://github.com/odoo/o-spreadsheet/pull/2968#discussion_r1361821451

rrahir:
@robodoo r+

LucasLefevre:
robodoo rebase-ff

PR #2997, opened 2023-10-12
[IMP] function: add CELL function

hokolomopo:
## Description:

Add the cell function, minus some functionnalities that we don't
support (eg. "color" parameter for format color) or some that we judge
not worth the work.

Also some commit to change slighly the function evaluation to make this function possible (change __originCellXC to __originCellPosition to have the sheetId, and manage to type the getters correctly in the function implementation)

Task: : [3484040](https://www.odoo.com/web#id=3484040&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/functions/module_info.ts
```diff
@@ -1,7 +1,58 @@
...
+  args: [
+    arg(
+      "info_type (string)",
+      _t('The type of information requested. Can be "address", "col", "contents" or "row".')
+    ),
+    arg("reference (meta)", _t("The reference to the cell.")),
+  ],
+  returns: ["ANY"],
+  compute: function (info: CellValue, reference: string) {
+    // NOTE: missing from Excel: "color", "filename", "format", "parentheses", "prefix", "protect", "type" and "width"
```
laa-odoo, line 24: 'format' and 'type' shouldn't be complex to implement.

'format' could be useful for us.

laa-odoo (commented):
I wonder if it wouldn't be better to store the sheet name information in the meta argument.
```refFn``` would construct the meta value. note that ```refFn``` already has the sheet name

File: src/helpers/reference_type.ts [hunk 2]
```diff
@@ -32,25 +32,25 @@ export function loopThroughReferenceType(token: Readonly<Token>): Token {
...
   return xc;
 }
 
 /**
- * Returns the given XC with the given reference type.
+ * Returns the given XC with the given reference type. The XC string should not contain a sheet name.
  */
-function setXcToReferenceType(xc: string, referenceType: FixedReferenceType): string {
+export function setXcToFixedReferenceType(xc: string, referenceType: FixedReferenceType): string {
   xc = xc.replace(/\$/g, "");
```
rrahir, line 54: could be nice to throw if there is a sheetname?

laa-odoo:
what final decision was made regarding passing the sheetName in the Meta arguments instead of in the evalual context ?

hokolomopo:
> what final decision was made regarding passing the sheetName in the Meta arguments instead of in the evalual context ?

@laa-odoo we said taht it was better to have the sheetName in the meta arguments. But thinking about it more I do need both the sheetName in the meta argument and the `__originSheetId`.

The function CELL returns the cell's address as a string with the sheet name if the referenced cell is in another sheet than the function. So I need both the sheetName in the range (which will give me the sheet the referenced cell is on), and the `__originSheetId` in the context (which will tell me the sheet the formula CELL is on)

src/helpers/references.ts
```diff
@@ -58,8 +58,16 @@ export function isSingleCellReference(xc: string): boolean {
...
+  }
   const parts = ref.split("!");
   const xc = parts.pop()!;
   const sheetName = getUnquotedSheetName(parts.join("!")) || undefined;
   return { sheetName, xc };
 }
+
+/** Return a reference SheetName!xc from the given arguments */
+export function getReference(xc: string, sheetName?: string): string {
+  return sheetName ? `${sheetName}!${xc}` : xc;
```
LucasLefevre, line 72: I think it would be safer to use `getCanonicalSheetName`. I don't think we have any guarantees  it's already canonical.
And some uses of `getCanonicalSheetName` can probably be changed to use this function

src/helpers/references.ts
```diff
@@ -58,8 +58,16 @@ export function isSingleCellReference(xc: string): boolean {
...
+    return { xc: ref };
+  }
   const parts = ref.split("!");
   const xc = parts.pop()!;
   const sheetName = getUnquotedSheetName(parts.join("!")) || undefined;
   return { sheetName, xc };
 }
+
+/** Return a reference SheetName!xc from the given arguments */
+export function getReference(xc: string, sheetName?: string): string {
```
LucasLefevre, line 71: what about `getFullReferenceString` or `toFullReferenceString` ?

LucasLefevre (commented):
woop woop

src/plugins/ui_core_views/cell_evaluation/evaluator.ts
```diff
@@ -405,10 +412,6 @@ export class Evaluator {
     return this.formulaDependencies().getCellsDependingOn(positionIds);
   }
 
-  private getCell(positionId: PositionId): Cell | undefined {
```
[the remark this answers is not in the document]
hokolomopo: Very small optimisation: if `this.getCell()` does the decoding internally, `computeCell()` would need to decode twice the positionId: once in the call of `getCell` and once when calling `computeFormulaCell()`

src/plugins/ui_core_views/cell_evaluation/evaluator.ts
```diff
@@ -505,3 +508,22 @@ class PositionBitsEncoder {
     return sheetId;
   }
 }
+
+export function evaluateFormula(
+  compiledFormula: CompiledFormula,
+  dependencies: Range[],
+  compilationParams: CompilationParameters,
+  sheetId: UID,
+  cellId?: UID
```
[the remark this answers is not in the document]
hokolomopo: Yeah, when evaluating from `getters.evaluateFormula()` we don't have a cell :x

src/helpers/references.ts
hokolomopo: Maybe `mergeReference` to be the opposite of `splitReference` ? Idk

LucasLefevre (commented):
small rebase then I think we are good to r+ :)

hokolomopo:
@LucasLefevre rebased

LucasLefevre (approved):
robodoo r+ rebase-ff

PR #3002, opened 2023-10-12
[FIX] clipboard: fix array formula copy

anhe-odoo:
## Task Description

When copying/cutting parts of the spread of an array formula, we currently copy an object with an evaluatedCell but without any cell, as there is no real cell to copy. Then, when pasting the copied values, we skip these parts as we look only for valid cell.

This commit aims to change this behaviour by pasting the evaluated cell when the cell is undefined.

## Related Task
- Task: [3541325](https://www.odoo.com/web#id=3541325&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## Review Checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/clipboard/clipboard_cells_state.ts
+++ b/src/helpers/clipboard/clipboard_cells_state.ts
@@ -38,7 +38,7 @@ interface CopiedTable {
 
 /** State of the clipboard when copying/cutting cells */
 export class ClipboardCellsState extends ClipboardCellsAbstractState {
-  private cells: ClipboardCell[][];
+  private readonly cells: ClipboardCell[][];
   private readonly copiedTables: CopiedTable[];
   private readonly zones: Zone[];
   private readonly uuidGenerator = new UuidGenerator();
@@ -82,11 +82,27 @@ export class ClipboardCellsState extends ClipboardCellsAbstractState {
       let cellsInRow: ClipboardCell[] = [];
       for (let col of columnsIndex) {
         const position = { col, row, sheetId };
+        const spreader = getters.getArrayFormulaSpreadingOn(position);
+        let cell = getters.getCell(position);
+        const evaluatedCell = getters.getEvaluatedCell(position);
+        if (spreader) {
+          const isSpreaderCopied =
+            rowsIndex.includes(spreader.row) && columnsIndex.includes(spreader.col);
+          const content = isSpreaderCopied
+            ? ""
+            : formatValue(evaluatedCell.value, { locale: getters.getLocale() });
+          cell = {
+            id: cell?.id || "",
+            style: cell?.style,
+            format: evaluatedCell.format,
+            content,
+            isFormula: false,
+          };
+        }
         cellsInRow.push({
-          cell: getters.getCell(position),
-          style: getters.getCellComputedStyle(position),
-          evaluatedCell: getters.getEvaluatedCell(position),
+          cell,
           border: getters.getCellBorder(position) || undefined,
+          evaluatedCell,
           position,
         });
       }
@@ -219,13 +235,6 @@ export class ClipboardCellsState extends ClipboardCellsAbstractState {
       });
     }
     this.pasteCopiedTables(target);
-    this.cells.forEach((row) => {
-      row.forEach((c) => {
-        if (c.cell) {
-          c.cell = undefined;
-        }
-      });
-    });
   }
 
   /**
@@ -309,7 +318,7 @@ export class ClipboardCellsState extends ClipboardCellsAbstractState {
   private clearClippedZones() {
     for (const row of this.cells) {
       for (const cell of row) {
-        if (cell.cell) {
+        if (cell?.cell) {
           this.dispatch("CLEAR_CELL", cell.position);
         }
       }
@@ -339,6 +348,9 @@ export class ClipboardCellsState extends ClipboardCellsAbstractState {
       const rowCells = this.cells[r];
       for (let c = 0; c < width; c++) {
         const origin = rowCells[c];
+        if (!origin) {
+          continue;
+        }
         const position = { col: col + c, row: row + r, sheetId: sheetId };
         // TODO: refactor this part. the "Paste merge" action is also executed with
         // MOVE_RANGES in pasteFromCut. Adding a condition on the operation type here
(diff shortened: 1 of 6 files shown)


src/helpers/clipboard/clipboard_cells_state.ts
```diff
@@ -75,13 +76,18 @@ export class ClipboardCellsState extends ClipboardCellsAbstractState {
...
     for (let row of rowsIndex) {
-      let cellsInRow: ClipboardCell[] = [];
+      let cellsInRow: Maybe<ClipboardCell>[] = [];
       for (let col of columnsIndex) {
         const position = { col, row, sheetId };
+        const spreader = getters.getArrayFormulaSpreadingOn(position);
+        if (spreader && rowsIndex.includes(spreader.row) && columnsIndex.includes(spreader.col)) {
+          cellsInRow.push(undefined);
+          continue;
+        }
```
laa-odoo, line 90: if we do that, copy/past cells with their own spreader will not copy/past the style and the border of the spreading cells, no ?

src/helpers/clipboard/clipboard_cells_state.ts
```diff
@@ -413,6 +416,10 @@ export class ClipboardCellsState extends ClipboardCellsAbstractState {
         style: origin.cell.style || null,
         format: origin.cell.format,
       });
+    } else if (origin.evaluatedCell && origin.evaluatedCell.type !== "empty") {
```
laa-odoo, line 419: we also need to consider onlyValue and onlyFormat cases in this part

src/helpers/clipboard/clipboard_cells_state.ts
```diff
@@ -369,58 +381,51 @@ export class ClipboardCellsState extends ClipboardCellsAbstractState {
...
-      if (origin.cell.isFormula && operation === "COPY") {
-        content = this.getters.getTranslatedCellFormula(
-          sheetId,
-          col - origin.position.col,
-          row - origin.position.row,
-          origin.cell.compiledFormula,
-          origin.cell.dependencies
-        );
-      }
+    if (clipboardOption?.pasteOption === "onlyFormat") {
```
laa-odoo, line 401: in this case, we will dispatch an UPDATE_CELL even if origin.cell is empty

laa-odoo: same comment concerning onlyValue

src/types/misc.ts
```diff
@@ -190,11 +190,12 @@ export function isMatrix(x: any): x is Matrix<any> {
 }
 
 export interface ClipboardCell {
+  content?: string;
```
LucasLefevre, line 193: Feels weird to have both `cell` and `content`. The first thing people reading this will ask themselves it what's the difference between `content` and `cell.content`. Same for `format`

src/helpers/clipboard/clipboard_cells_state.ts
```diff
@@ -502,14 +507,22 @@ export class ClipboardCellsState extends ClipboardCellsAbstractState {
...
     let htmlTable = '<table border="1" style="border-collapse:collapse">';
     for (const row of this.cells) {
       htmlTable += "<tr>";
       for (const cell of row) {
-        const cssStyle = cssPropertiesToCss(cellStyleToCss(cell.style));
+        if (!cell) {
+          continue;
+        }
+        const cssStyle = cssPropertiesToCss(
+          cellStyleToCss(this.getters.getCellComputedStyle(cell.position))
```
LucasLefevre, line 524: Isn't this another fix (that should be in another commit, and possibly backported to earlier versions) ? That because we don't copy the conditional formats, right ?

src/helpers/clipboard/clipboard_cells_state.ts
```diff
@@ -369,58 +381,51 @@ export class ClipboardCellsState extends ClipboardCellsAbstractState {
...
-        this.dispatch("UPDATE_CELL", {
-          ...target,
-          style: origin.cell.style,
-          format: origin.evaluatedCell.format,
-        });
-        return;
-      }
+    if (clipboardOption?.pasteOption === "onlyValue") {
+      const locale = this.getters.getLocale();
+      const content = formatValue(origin.evaluatedCell.value ?? "", { locale });
```
LucasLefevre, line 386: null coalescing is useless. `value` is never undefined nor null
```suggestion
      const content = formatValue(origin.evaluatedCell.value, { locale });
```

src/helpers/clipboard/clipboard_cells_state.ts
LucasLefevre: up

LucasLefevre (commented):
can target 17.0 :wink:

src/helpers/clipboard/clipboard_cells_state.ts
anhe-odoo: No, in fact this change is coming from the fact that, now, cell.style is not present anymore while, before, we were pushing `getCellComputedStyle(cell.position)` in the object when copying the cells. So it's not a fix, only a change caused by my other changes :)

src/helpers/clipboard/clipboard_cells_state.ts
anhe-odoo: While I agree'd with Alexis at first time, it looks like displaying UPDATE_CELL with empty origin is a desired behavior as we have a test to ensure that copying format from an empty cell clear the format of the paste target. And same things happen when pasting the value from an empty cell to clear the content of a target cell

LucasLefevre (commented):
revert those changes:
https://github.com/odoo/o-spreadsheet/pull/2473/commits/2faad1936326207d65dae533ffe4e7327a640cd2#diff-0207ac1d563650b2cfda792312504949500aa41ef99016cd31f58750249bf157L98
https://github.com/odoo/o-spreadsheet/pull/2473/commits/2faad1936326207d65dae533ffe4e7327a640cd2#diff-c71d8f0676e4628a0aac3d53bf055d95e80bcb3f582700abd3a77fd73984f892L33

It's n'importe quoi.

I think it will fix the strange system crash, since we won't even need to invalidate the cells

tests/clipboard/clipboard_plugin.test.ts
```diff
@@ -2052,6 +2052,79 @@ describe("clipboard: pasting outside of sheet", () => {
     expect(getEvaluatedCell(model, "A1").value).toBe(8.14);
   });
 
+  test("Can copy parts of the spreaded values", () => {
```
VincentSchippefilt, line 2055: spread in the past tense is spread

tests/clipboard/clipboard_plugin.test.ts
```diff
@@ -2052,6 +2052,79 @@ describe("clipboard: pasting outside of sheet", () => {
...
+    setCellContent(model, "A2", "2");
+    setCellContent(model, "A3", "3");
+    setCellContent(model, "B1", "=TRANSPOSE(A1:A3)");
+    copy(model, "C1:D1");
+    paste(model, "C2");
+    expect(getEvaluatedCell(model, "C2").value).toBe(2);
+    expect(getEvaluatedCell(model, "D2").value).toBe(3);
+  });
+
+  test("Cutting parts of the spreaded values will make a copy of the values", () => {
```
VincentSchippefilt, line 2067: spread in the past tense is spread

tests/clipboard/clipboard_plugin.test.ts
```diff
@@ -2052,6 +2052,79 @@ describe("clipboard: pasting outside of sheet", () => {
...
+    cut(model, "C1:D1");
+    paste(model, "C2");
+    expect(getEvaluatedCell(model, "B1").value).toBe(1);
+    expect(getEvaluatedCell(model, "C1").value).toBe(2);
+    expect(getEvaluatedCell(model, "C2").value).toBe(2);
+    expect(getEvaluatedCell(model, "D1").value).toBe(3);
+    expect(getEvaluatedCell(model, "D2").value).toBe(3);
+  });
+
+  test("can copy and paste format only from spreaded value", () => {
```
VincentSchippefilt, line 2082: spread in the past tense is spread

tests/clipboard/clipboard_plugin.test.ts
```diff
@@ -2052,6 +2052,79 @@ describe("clipboard: pasting outside of sheet", () => {
...
+    setCellContent(model, "A1", "=SUM(1+2)");
+
+    // formula with format seted on it
+    setCellContent(model, "A2", "=SUM(1+2)");
+    setCellFormat(model, "A2", "0%");
+
+    // formula that return value with format
+    setCellContent(model, "A3", "=DATE(2042,1,1)");
+
+    // formula that return value with format and other format seted on it
```
VincentSchippefilt, line 2095: set in the past is set

tests/clipboard/clipboard_plugin.test.ts
```diff
@@ -2052,6 +2052,79 @@ describe("clipboard: pasting outside of sheet", () => {
...
+    setCellContent(model, "A4", "=DATE(2042,1,1)");
+    setCellFormat(model, "A4", "0%");
+
+    // formula that return value with format infered from reference
```
VincentSchippefilt, line 2099: infered --> inferred

VincentSchippefilt:
robodoo r+

LucasLefevre:
robodoo r+

PR #3007, opened 2023-10-13
[FIX] find_and_replace: take array formula result into account

anhe-odoo:
## Task Description

Currently, in the find and replace results, we only look for matches in the cell directly, ignoring the values spread by an array formula. This PR aims to fix it, taking these values into account in the search, but also avoiding to replace them in a "FIND_AND_REPLACE_ALL" cycle, as writing a new value in these cells would avoid the array formula to spread.

## Related Task
- Task: [3413999](https://www.odoo.com/web#id=3413999&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## Review Checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_core_views/cell_evaluation/evaluation_plugin.ts
+++ b/src/plugins/ui_core_views/cell_evaluation/evaluation_plugin.ts
@@ -147,6 +147,7 @@ export class EvaluationPlugin extends UIPlugin {
     "getEvaluatedCell",
     "getEvaluatedCells",
     "getEvaluatedCellsInZone",
+    "getSpreadPositionsOf",
   ] as const;
 
   private shouldRebuildDependenciesGraph = true;
@@ -269,6 +270,10 @@ export class EvaluationPlugin extends UIPlugin {
     return positions(zone).map(({ col, row }) =>
       this.getters.getEvaluatedCell({ sheetId, col, row })
     );
+  }
+
+  getSpreadPositionsOf(position: CellPosition): CellPosition[] {
+    return this.evaluator.getSpreadPositionsOf(position);
   }
 
   // ---------------------------------------------------------------------------
--- a/src/plugins/ui_feature/find_and_replace.ts
+++ b/src/plugins/ui_feature/find_and_replace.ts
@@ -159,21 +159,28 @@ export class FindAndReplacePlugin extends UIPlugin {
     const sheetId = this.getters.getActiveSheetId();
     const cells = this.getters.getCells(sheetId);
     const matches: SearchMatch[] = [];
-    if (this.toSearch) {
+    if (this.toSearch && this.currentSearchRegex) {
       for (const cell of Object.values(cells)) {
         const { col, row } = this.getters.getCellPosition(cell.id);
+        const cellPosition = { sheetId, col, row };
         const isColHidden = this.getters.isColHidden(sheetId, col);
         const isRowHidden = this.getters.isRowHidden(sheetId, row);
         if (isColHidden || isRowHidden) {
           continue;
         }
-        if (
-          cell &&
-          this.currentSearchRegex &&
-          this.currentSearchRegex.test(this.getSearchableString({ sheetId, col, row }))
-        ) {
+        if (this.currentSearchRegex.test(this.getSearchableString(cellPosition))) {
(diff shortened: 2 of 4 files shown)


tests/find_and_replace/find_and_replace_plugin.test.ts
```diff
@@ -697,3 +718,25 @@ describe("replace", () => {
     expect(getActivePosition(model)).toBe("A1");
   });
 });
+
+test("simple search with array formula", () => {
```
hokolomopo, line 722: test name, we're testing replace. "replace do'nt replace spreaded results of array formula" or something like that

src/plugins/ui_feature/find_and_replace.ts
```diff
@@ -159,22 +159,28 @@ export class FindAndReplacePlugin extends UIPlugin {
...
-        if (
-          cell &&
-          this.currentSearchRegex &&
-          this.currentSearchRegex.test(this.getSearchableString({ sheetId, col, row }))
-        ) {
+        if (this.currentSearchRegex.test(this.getSearchableString({ sheetId, col, row }))) {
           const match: SearchMatch = { col, row, selected: false };
           matches.push(match);
         }
+        for (const spreadPosition of this.getters.getSpreadPositionsOf({ sheetId, col, row })) {
```
hokolomopo, line 174: nitpick: we're creating a { sheetId, col, row } object twice

hokolomopo (commented):
Nice :+1:

src/plugins/ui_core_views/cell_evaluation/evaluation_plugin.ts
```diff
@@ -269,6 +271,14 @@ export class EvaluationPlugin extends UIPlugin {
     );
   }
 
+  getArrayFormulaSpreadingOn(position: CellPosition): CellPosition | undefined {
```
rrahir, line 274: getCorrespondingFormulaCell can do the trick (it will be available after a small rebase)

src/plugins/ui_feature/find_and_replace.ts
```diff
@@ -237,9 +244,13 @@ export class FindAndReplacePlugin extends UIPlugin {
...
 
     const sheetId = this.getters.getActiveSheetId();
-    const cell = this.getters.getCell({ sheetId, ...selectedMatch });
     const { col, row } = selectedMatch;
 
+    if (this.getters.getArrayFormulaSpreadingOn({ sheetId, col, row })) {
+      return;
+    }
+
+    const cell = this.getters.getCell({ sheetId, ...selectedMatch });
```
rrahir, line 253: Alternatively to the use `getArrayFormulaSpreadingOn` we could check that cell has a content. I think it could make sense to compare the performances?

tests/find_and_replace/find_and_replace_plugin.test.ts
```diff
@@ -697,3 +718,25 @@ describe("replace", () => {
...
+    searchFormulas: false,
+  };
+  model.dispatch("UPDATE_SEARCH", { toSearch: "hello", searchOptions });
+  model.dispatch("REPLACE_ALL_SEARCH", { replaceWith: "kikou" });
+  const matches = model.getters.getSearchMatches();
+  const matchIndex = model.getters.getCurrentSelectedMatchIndex();
+  expect(matches).toHaveLength(0);
+  expect(matchIndex).toStrictEqual(null);
+  expect(getCellContent(model, "A2")).toBe("kikou");
+  expect(getCellContent(model, "C1")).toBe("kikou");
```
rrahir, line 740: this tests passes in 17.0 (I think we should fix it in 17.0 as well). this helpers checks the content of the evaluated cell, but not the content of the core cell.

anhe-odoo: It's totally normal that this test passes in 17.0 as is now, as we don't take array formula spreaded values into account when doing a search/replace, so it won't be replaced. This test doesn't have any sense as a standalone (without the behaviour modification I introduce in this fix) :)

anhe-odoo:
Closing this PR in favor of https://github.com/odoo/o-spreadsheet/pull/3138

PR #3029, opened 2023-10-17
[IMP] scorecard: handling CF font style on chart

somu-odoo:
## Description:

Before:
Any applied CF font color on key cell would not reflect on chart

After:
Text color of Key Value will reflect as follows:
- If no font color is set, then default text color will be applied.
- If CF font color is applied on cell, key value will reflect that text color.
- If user sets a font color on cell, then user applied color of key value 
will take precedence over CF font color

"getStyleOfSingleCellChart" method of evaluation chart is modified to
"getBackgroundOfSingleCellChart", which handles style of a single cell 
for chart at runtime .

Task: : [3543131](https://www.odoo.com/web#id=3543131&cids=2&menu_id=4720&action=333&active_id=2328&model=project.task&view_type=form)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui/evaluation_chart.ts
+++ b/src/plugins/ui/evaluation_chart.ts
@@ -1,5 +1,5 @@
 import { BACKGROUND_CHART_COLOR } from "../../constants";
-import { chartRuntimeFactory } from "../../helpers/charts";
+import { chartFontColor, chartRuntimeFactory } from "../../helpers/charts";
 import { Color, Range, UID } from "../../types";
 import { ChartRuntime } from "../../types/chart/chart";
 import {
@@ -9,8 +9,13 @@ import {
 } from "../../types/commands";
 import { UIPlugin } from "../ui_plugin";
 
+interface EvaluationChartStyle {
+  background: Color;
+  fontColor: Color;
+}
+
 export class EvaluationChartPlugin extends UIPlugin {
-  static getters = ["getChartRuntime", "getBackgroundOfSingleCellChart"] as const;
+  static getters = ["getChartRuntime", "getStyleOfSingleCellChart"] as const;
 
   readonly charts: Record<UID, ChartRuntime | undefined> = {};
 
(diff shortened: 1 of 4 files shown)


src/plugins/ui/evaluation_chart.ts
```diff
@@ -61,20 +66,31 @@ export class EvaluationChartPlugin extends UIPlugin {
...
    *  - the chart background color if one is defined
    *  - the fill color of the cell if one is defined
+   *  - the text color of the cell if one is defined
    *  - the fill color of the cell from conditional formats if one is defined
+   *  - the text color of the cell from conditional formats if one is defined
    *  - the default chart color if no other color is defined
+   *  - the default text color if no other color is defined
    */
-  getBackgroundOfSingleCellChart(
+  getStyleOfSingleCellChart(
```
LucasLefevre, line 75: the docstring first line can be updated :wink: It's no longer `Get the background color of...`

Also, the docstring is actually wrong (since the beginning ?). The CF style takes precedence over cell defined style.
I think the all this "order of priority" part can be removed. It's doomed to be outdated at some point. If someone wants to know how it's computed, they go and check `getCellComputedStyle`

src/plugins/ui/evaluation_chart.ts
```diff
@@ -61,20 +66,31 @@ export class EvaluationChartPlugin extends UIPlugin {
...
     }
     const col = mainRange.zone.left;
     const row = mainRange.zone.top;
-    const style = this.getters.getCellComputedStyle(mainRange.sheetId, col, row);
-    return style.fillColor || BACKGROUND_CHART_COLOR;
+    const sheetId = mainRange.sheetId;
+    const style = this.getters.getCellComputedStyle(sheetId, col, row);
+    return {
+      background: style.fillColor || BACKGROUND_CHART_COLOR,
+      fontColor: style.textColor || chartFontColor(style.fillColor),
```
LucasLefevre, line 93: shouldn't it fallback on BACKGROUND_CHART_COLOR
```suggestion
      fontColor: style.textColor || chartFontColor(style.fillColor || BACKGROUND_CHART_COLOR),
```
`style.fillColor || BACKGROUND_CHART_COLOR` can be extracted to a local `background` variable.

src/plugins/ui/evaluation_chart.ts
somu-odoo: Gotcha! Removed them :+1:

src/plugins/ui/evaluation_chart.ts
somu-odoo: Yep makes sense, I went with the approach of sending "null" value to it, so that it can auto assign the default color. But this one looks more understandable. :+1:

LucasLefevre (approved):
robodoo r+

LucasLefevre:
rodoboo r-
actually, it can target 17.0, not 16.0. It's not a blocking issue

LucasLefevre:
f*ck, I can't spell robodoo :sweat_smile:

somu-odoo:
Not an issue :smile:  
Anyway should I fix the FW PR or create a new one targeting 17.0?

PR #3044, opened 2023-10-20
[FIX] scorecard: fix baseline arrow

anhe-odoo:
## Task Description

Since a825627 the baseline arrow are drawn as an unicode character. However, this seems not to be supported by all browsers/OS, and we need to change this behavior to directly draw an arrow. This PR aims to draw the arrow using an SVG path and Path2D, fixing this issue.

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


rrahir:
@robodoo r+

PR #3057, opened 2023-10-23
[REF] chart_test: improve test speed/readability

hokolomopo:
## [REF] chart_test: improve test speed/readability

Slightly improve the testing speed of the chart component by
removing useless nextTick. Add helpers to open the chart side panels
without having 5 lines of codes and 5 nextTicks.

## [REF] tests: dispatch both input and change events

Currently in the test, we often only dispatch either an input or a change
event. But in practice, both events are fired when modifying an
input. This could lead to inconsistencies between the test and the
reality.

This commit changes the setInputValueAndTrigger to dispatch both events.
It also make the helper async with a nextTick at the end to shorten the
tests.

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/tests/colors/color_picker_component.test.ts
+++ b/tests/colors/color_picker_component.test.ts
@@ -9,7 +9,7 @@ import {
   setInputValueAndTrigger,
   simulateClick,
 } from "../test_helpers/dom_helper";
-import { mountComponent, nextTick } from "../test_helpers/helpers";
+import { mountComponent } from "../test_helpers/helpers";
 import { mockGetBoundingClientRect } from "../test_helpers/mock_helpers";
 
 mockGetBoundingClientRect({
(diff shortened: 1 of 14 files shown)


tests/colors/color_picker_component.test.ts
```diff
@@ -110,8 +109,7 @@ describe("Color Picker buttons", () => {
     await simulateClick(".o-color-picker-toggler");
     await simulateClick(".o-gradient");
     const color = "12ef78";
-    setInputValueAndTrigger(".o-custom-input-preview input", color, "input");
```
LucasLefevre, line 113: shouldn't all lines with (previously) "input" be converted to `"inputOnly"` ?

tests/test_helpers/dom_helper.ts
```diff
@@ -195,14 +195,20 @@ function dispatchEvent(selector: string | EventTarget, ev: Event) {
   }
 }
 
-export function setInputValueAndTrigger(
+export async function setInputValueAndTrigger(
   selector: DOMTarget,
   value: string,
-  eventType: string
-): void {
+  mode?: "onlyInput" | "onlyChange"
```
LucasLefevre, line 201: what do you think of two different functions: `changeInputValue` and `inputValue` ? would it be clearer than "onlyInput" | "onlyChange" ?

LucasLefevre (commented):
nice :+1:

tests/colors/color_picker_component.test.ts
hokolomopo: I don't think so, most of the times when we were only firing "input" on the tests, in a browser both event were fired. We're closer to reality that way. Only one test need to specifically test what happend when when an input is fired an not change

LucasLefevre (approved):
robodoo r+ rebase-ff

PR #3094, opened 2023-10-30
[IMP] Sheetview: expose Sheetview `Rect`

rrahir:
Introducing a new getter that allows to export the sheetview offset and dimensions (width/height).

Task: 3575484

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_stateful/sheetview.ts
+++ b/src/plugins/ui_stateful/sheetview.ts
@@ -99,6 +99,7 @@ export class SheetViewPlugin extends UIPlugin {
     "getSheetViewVisibleRows",
     "getFrozenSheetViewRatio",
     "isPositionVisible",
+    "getVisibleSheetViewRect",
   ] as const;
 
   readonly viewports: Record<UID, SheetViewports | undefined> = {};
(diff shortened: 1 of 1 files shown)


src/plugins/ui_stateful/sheetview.ts
```diff
@@ -316,6 +317,15 @@ export class SheetViewPlugin extends UIPlugin {
     };
   }
 
+  getVisibleSheetViewRect(): Rect {
```
hokolomopo, line 320: Are we ok with getters in o_spreadsheet that are only used in Odoo ? We could add this getter to a random odoo plugin

rrahir: this exposes the internal state of the plugin, so yeah I could use existing getters to infer it, it just looked super shitty.

hokolomopo: Yeah both solution sucks, I was just wondering our stance on "getter only used in odoo" vs "wonky getter only in odoo"

PR #3095, opened 2023-10-30
[IMP] locale: change date format with locale

hokolomopo:
## Description

This commit changes the date format of the cells when changing the locale.

This is currently done the dumbest way possible: check if the cell's format exactly match the previous locale's format, and if so, change it to the new locale's format. This works fine as long as the only date formats are the ones used in the topbar menu.

Further work can be done to support a conversion with more date formats that don't exactly match the locale's.

Task: : [3575516](https://www.odoo.com/web#id=3575516&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/core/settings.ts
+++ b/src/plugins/core/settings.ts
@@ -1,5 +1,12 @@
-import { isValidLocale } from "../../helpers/locale";
-import { CommandResult, CoreCommand, DEFAULT_LOCALE, Locale, WorkbookData } from "../../types";
+import { getDateTimeFormat, isValidLocale } from "../../helpers/locale";
+import {
+  CommandResult,
+  CoreCommand,
+  DEFAULT_LOCALE,
+  Format,
+  Locale,
+  WorkbookData,
+} from "../../types";
 import { CorePlugin } from "./../core_plugin";
 
 export class SettingsPlugin extends CorePlugin {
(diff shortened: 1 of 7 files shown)


src/plugins/core/settings.ts
```diff
@@ -26,6 +36,29 @@ export class SettingsPlugin extends CorePlugin {
...
   }
 
+  private changeCellsDateFormatWithLocale(oldLocale: Locale, newLocale: Locale) {
+    for (const sheetId of this.getters.getSheetIds()) {
+      for (const [cellId, cell] of Object.entries(this.getters.getCells(sheetId))) {
+        let formatToApply: Format | undefined;
+        if (cell.format === oldLocale.dateFormat) {
+          formatToApply = newLocale.dateFormat;
+        }
+        if (cell.format === getDateTimeFormat(oldLocale)) {
```
LucasLefevre, line 46: shouldn't we also check `timeFormat` while we are at it?

LucasLefevre (approved):

src/plugins/core/settings.ts
hokolomopo: True, I changed that

LucasLefevre (approved):
robodoo r+

PR #3125, opened 2023-11-03
[FIX] evaluation: fix incorrect invalidation

LucasLefevre:
## Description:

Steps to reproduce: look at the test.

Here is what happens:
- initially, all cells are being evaluated, in this order: [sheet1!A1, sheet2!A1, sheet2!A4]

1) When sheet!A1 is evaluated, it evaluates its dependency sheet2!A4. MEDIAN(A2:A3) function results with an error because A2:A3 is empty (the array formula MFILL is not spread yet)

2) Then sheet2!A1 is evaluated, which marks sheet2!A4 as to be re-evaluated at the next iteration because it depends on A2:A3 which was just filled with values.

3) The third cell in the list, sheet2!A4, is computed. But it's been already been computed at the first step (the result is an error). Nothing wrong at this point.
The issue comes when we store the value in `setEvaluatedCell`: we remove the cell from `setEvaluatedCell`! It means it will never be recomputed at the next iteration.

With this commit, because we no longer remove cells from the next iteration, the maximum iteration number is increased by one.

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_core_views/cell_evaluation/evaluator.ts
+++ b/src/plugins/ui_core_views/cell_evaluation/evaluator.ts
@@ -186,9 +186,6 @@ export class Evaluator {
   }
 
   private setEvaluatedCell(positionId: PositionId, evaluatedCell: EvaluatedCell) {
-    if (this.nextPositionsToUpdate.has(positionId)) {
-      this.nextPositionsToUpdate.delete(positionId);
-    }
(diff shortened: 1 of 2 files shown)


laa-odoo (approved):

laa-odoo:
@robodoo r+

PR #3129, opened 2023-11-03
[FIX] clipboard: traceback on paste from copy after deleting that sheet

khpa-odoo:
## Description:
If the sheet containing copy zone was deleted before paste, it lead to a
traceback.

The issue occurred since the method `getValidationRuleForCell` tried to access
the sheet of copied clipboard content, which now did not exist.

**Additional multi-user manipulation issues:**
i. "CREATE_SHEET"
ii."DUPLICATE_SHEET"

 **_Steps to reproduce_** :-
- User 1 performs (i) or (ii)
- User 2 copies some cells from newly created sheet and moves to some other sheet.
- User 1 performs undo (effectively deleting newly created sheet)
- User 2 hits "paste" => traceback due to same aforementioned reason.

This PR fixes these issues by tweaking the method to return the data validation
rules only if the sheet containing copy zone exists, else return undefined.
(just like how conditional format rules work)

Task: : [3580276](https://www.odoo.com/web#id=3580276&menu_id=4720&cids=2&action=333&active_id=2328&model=project.task&view_type=form)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/core/data_validation.ts
+++ b/src/plugins/core/data_validation.ts
@@ -134,6 +134,9 @@ export class DataValidationPlugin
   }
 
   getValidationRuleForCell({ sheetId, col, row }: CellPosition): DataValidationRule | undefined {
+    if (!this.rules[sheetId]) {
+      return undefined;
+    }
     for (const rule of this.rules[sheetId]) {
       for (const range of rule.ranges) {
         if (isInside(col, row, range.zone)) {
--- a/tests/clipboard/clipboard_plugin.test.ts
+++ b/tests/clipboard/clipboard_plugin.test.ts
@@ -240,6 +240,23 @@ describe("clipboard", () => {
     expect(getCell(model, "A2", sheet2Id)).toBe(undefined);
   });
 
+  test("can paste even if sheet containing copy zone has been deleted", () => {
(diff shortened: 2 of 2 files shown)


src/plugins/core/data_validation.ts
```diff
@@ -134,10 +134,12 @@ export class DataValidationPlugin
   }
 
   getValidationRuleForCell({ sheetId, col, row }: CellPosition): DataValidationRule | undefined {
-    for (const rule of this.rules[sheetId]) {
-      for (const range of rule.ranges) {
-        if (isInside(col, row, range.zone)) {
-          return rule;
+    if (this.rules[sheetId]) {
```
LucasLefevre, line 137: With this appraoach , we'll end up with many many and more and more indentation levels. It hurts readability
Prefer an early return in those cases.
```ts
if (!this.rules[sheetId]) {
  return;
}
```

khpa-odoo: Okay, noted!
Applied the change :+1:

LucasLefevre (approved):
robodoo r+
thanks

PR #3134, opened 2023-11-06
[IMP] evaluation: improve perfs of `getCorrespondingFormulaCell`

hokolomopo:
## [FIX] cell: update cell tokens and dependencies

Before this commit, when creating a new FormulaCell we compiled the
formula, and in the cell we stored the CompiledFormula, that contained
the tokens and the dependencies.

The problem was that the tokens were not updated when the formula
dependencies changed because of sheet manipulation. The cell's formula
dependencies wasn't updated either, and the dependencies info was
confusingly duplicated in `cell.compiledFormula.dependencies` (static
strings) and `cell.dependencies` (Ranges objects that we update).

This commit:
1) removes the `cell.dependencies` property, the dependencies are now
only stored in the CompiledFormula (as Ranges instead of strings)
2) replaces the raw `REFERENCE` tokens in the cell's formula by smarter
tokens that reference the range corresponding the the reference[1]. This
means that when a range is updated, the token content is updated too.

[1] we could also keep simple tokens, and update them/re-tokenize the
formulas on sheet manipulation, but this would be slower.

## [IMP] evaluation: improve perfs of `getCorrespondingFormulaCell`

The getter `getCorrespondingFormulaCell` was tokenizing the cell's content (via `compile`), which took a bit of time. But this is useless, as we already have the tokens in the formula cell.

Task: : [3584306](https://www.odoo.com/web#id=3584306&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/core/cell.ts
+++ b/src/plugins/core/cell.ts
@@ -1,5 +1,5 @@
 import { DEFAULT_STYLE } from "../../constants";
-import { compile, tokenize } from "../../formulas";
+import { Token, compile, tokenize } from "../../formulas";
 import { parseLiteral } from "../../helpers/cells";
 import {
   concat,
@@ -27,6 +27,7 @@ import {
   HeaderIndex,
   LiteralCell,
   Range,
+  RangeCompiledFormula,
   RangePart,
   Style,
   UID,
@@ -64,7 +65,7 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
     for (const sheet of Object.keys(this.cells)) {
       for (const cell of Object.values(this.cells[sheet] || {})) {
         if (cell.isFormula) {
-          for (const range of cell.dependencies) {
+          for (const range of cell.compiledFormula.dependencies) {
             if (!sheetId || range.sheetId === sheetId) {
               const change = applyChange(range);
               if (change.changeType !== "NONE") {
@@ -72,8 +73,9 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
                   "cells",
                   sheet,
                   cell.id,
-                  "dependencies" as any,
-                  cell.dependencies.indexOf(range),
+                  "compiledFormula" as any,
+                  "dependencies",
+                  cell.compiledFormula.dependencies.indexOf(range),
                   change.range
                 );
               }
@@ -296,13 +298,13 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
    */
   getFormulaCellContent(
     sheetId: UID,
-    cell: Pick<FormulaCell, "dependencies" | "compiledFormula">,
+    compiledFormula: RangeCompiledFormula,
     dependencies?: Range[]
   ): string {
-    const ranges = dependencies || cell.dependencies;
+    const ranges = dependencies || compiledFormula.dependencies;
     let rangeIndex = 0;
     return concat(
-      cell.compiledFormula.tokens.map((token) => {
+      compiledFormula.tokens.map((token) => {
         if (token.type === "REFERENCE") {
           const range = ranges[rangeIndex++];
           return this.getters.getRangeString(range, sheetId);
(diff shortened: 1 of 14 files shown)


src/plugins/core/cell.ts
```diff
@@ -618,25 +617,46 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
...
+      tokens,
+    };
+  }
 
   get content() {
-    return this.getFormulaCellContent(this.sheetId, {
-      dependencies: this.dependencies,
-      compiledFormula: this.compiledFormula,
-    });
+    return this.compiledFormula.tokens.map((token) => token.value).join("");
```
LucasLefevre, line 650: you can use `concat`, it's supposed to be faster

src/plugins/core/cell.ts
```diff
@@ -556,8 +554,10 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
       style,
       format,
       isFormula: true,
-      compiledFormula,
-      dependencies: [],
+      compiledFormula: {
+        ...compiledFormula,
+        dependencies: [],
+      },
```
LucasLefevre, line 560: spread and `dependencies: []` not needed.  We know `compiledFormula.dependencies` is already an empty array at this point.

src/plugins/core/cell.ts
```diff
@@ -618,25 +617,46 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
   }
 }
 
-class FormulaCellWithDependencies {
+class FormulaCellWithDependencies implements FormulaCell {
```
LucasLefevre, line 620: :+1:

src/plugins/ui_core_views/cell_evaluation/evaluation_plugin.ts
```diff
@@ -318,7 +319,7 @@ export class EvaluationPlugin extends UIPlugin {
     const cell = this.getters.getCell(position);
 
     if (cell && cell.content) {
-      if (cell.isFormula && !isBadExpression(cell.content)) {
+      if (cell.isFormula && !isBadExpression(cell.compiledFormula.tokens)) {
```
LucasLefevre, line 322: maybe we can recombine/reorder those two ifs and check `cell.isFormula` first. The goal would be to avoid `cell.content` for formulas since it isn't completely free since it rebuilds the range strings.

LucasLefevre (commented):
nice perf improvement and cleaning :)

src/plugins/core/cell.ts
hokolomopo: Needed for typing, compiledFormula.dependencies is as `Array<string>`, and we want an `Array<Range>` in the ForumlaCell :x

src/plugins/core/cell.ts
hokolomopo: Yeah I saw it afterward, I added it in one of the push force :)

src/plugins/ui_core_views/cell_evaluation/evaluation_plugin.ts
```diff
@@ -317,10 +318,9 @@ export class EvaluationPlugin extends UIPlugin {
   getCorrespondingFormulaCell(position: CellPosition): FormulaCell | undefined {
     const cell = this.getters.getCell(position);
 
-    if (cell && cell.content) {
-      if (cell.isFormula && !isBadExpression(cell.content)) {
-        return cell;
-      }
+    if (cell && cell.isFormula) {
+      return isBadExpression(cell.compiledFormula.tokens) ? undefined : cell;
+    } else if (cell && cell.content) {
```
LucasLefevre, line 323: could be `cell?.isFormula` and `cell?.content`

src/plugins/core/cell.ts
LucasLefevre: push force was push forced ? :smile:

LucasLefevre (commented):
Two minor comments.
LGTM otherwise :+1:

src/plugins/ui_core_views/cell_evaluation/evaluation_plugin.ts
hokolomopo: I think cell && ... is easier to read, no ?

LucasLefevre: I don't care. It's nitpicking

LucasLefevre (approved):
robodoo r+ rebase-ff

PR #3136, opened 2023-11-06
[REF] viewport: clean some dead code + improve typing

anhe-odoo:
## Task Description

This commit aims to remove some useless code in the viewports as well as change some typing like using DOMCoordinates and DOMDimension.

## Related Task
- Task: [3462055](https://www.odoo.com/web#id=3462055&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/helpers/internal_viewport.ts
```diff
@@ -153,7 +151,62 @@ export class InternalViewport {
...
+    this.adjustViewportZoneY();
+  }
+
+  /**
+   *
+   * @param zone
+   * @returns Computes the absolute coordinate of a given zone inside the viewport
+   */
+  getRect(zone: Zone): Rect | undefined {
+    const targetZone = intersection(zone, this);
```
hokolomopo, line 170: open question: is it worth to lose the git blame history just so the private and public methods are grouped together ?

LucasLefevre: > open question: is it worth to lose the git blame history just so the private and public methods are grouped together ?

Let's close the question: no, it's not worth it. It's much better to keep the diff history

That's one of the reason I don't like "sections" of code  with titles like `// private` where some methods are grouped together. The sections inevitably ends up being wrong.

Moving the methods for the sake of regrouping them also makes the review harder. Especially since it's mixed with other cleaning/fixes. I have to carefully compare each line before/after to check it anything changed

LucasLefevre (approved):
nice clean :)

rrahir:
@robodoo r+

PR #3207, opened 2023-11-17
[FIX] HeaderPositionsUIPlugin: cell wrapping for long text

dhrp-odoo:
## Description:

Previously, wrapping a cell with extensive text led to rendering issues in the spreadsheet. Additionally, using `overflow` and `clip` needlessly increased the row height

This PR fixes the issue by adding a command within the plugin to compute the header position of the sheet when applying formatting. It also corrects the `maxWidth` of the cell within the `getDefaultCellHeight` method.

Task: : [3586615](https://www.odoo.com/web#id=3586615&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_stateful/header_positions.ts
+++ b/src/plugins/ui_stateful/header_positions.ts
@@ -23,9 +23,11 @@ export class HeaderPositionsUIPlugin extends UIPlugin {
         }
         break;
       case "UPDATE_CELL":
-        if ("content" in cmd || "format" in cmd || cmd.style?.fontSize !== undefined) {
+        if ("content" in cmd || "format" in cmd) {
           this.headerPositions = {};
           this.isDirty = true;
+        } else {
+          this.headerPositions[cmd.sheetId] = this.computeHeaderPositionsOfSheet(cmd.sheetId);
         }
         break;
       case "UPDATE_FILTER":
--- a/src/helpers/text_helper.ts
+++ b/src/helpers/text_helper.ts
@@ -23,7 +23,7 @@ export function getDefaultCellHeight(
   colSize: number
 ) {
   if (!cell || !cell.content) return DEFAULT_CELL_HEIGHT;
-  const maxWidth = cell.style?.wrapping ? colSize - 2 * MIN_CELL_TEXT_MARGIN : undefined;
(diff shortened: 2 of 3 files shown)


src/plugins/ui_stateful/header_positions.ts
```diff
@@ -47,6 +47,7 @@ export class HeaderPositionsUIPlugin extends UIPlugin {
       case "UNGROUP_HEADERS":
       case "GROUP_HEADERS":
       case "CREATE_SHEET":
+      case "SET_FORMATTING":
```
hokolomopo, line 50: We can update the style of a cell without `SET_FORMATTING`, `SET_FORMATTING` is just a wrapper around `UPDATE_CELL`.

You should change the `UPDATE_CELL` of this switch. It should be
```js
      case "UPDATE_CELL":
        if ("content" in cmd) {
          this.headerPositions = {};
          this.isDirty = true;
        } else {
          this.headerPositions[cmd.sheetId] = this.computeHeaderPositionsOfSheet(cmd.sheetId);
        }
```

if the content of the cell changed, we need to mark the headers sizes as dirty while waiting for the evaluation. Otherwise if the style/format changed we only recompute the header sizes for this sheet

hokolomopo (commented):
Small comment and it should be good :)

src/plugins/ui_stateful/header_positions.ts
dhrp-odoo: Hello @hokolomopo,

When we change the format of content filtered in a data filter, It is not computing header positions correctly, Which results in keeping rows hidden inside the data filter and weird behavior between two plugins.

[Check out the failing test case here](https://github.com/odoo/o-spreadsheet/blob/17.0/tests/sheet/sheetview_plugin.test.ts#L846-#L861)

Should we keep the format to compute the header position after evaluation?

dhrp-odoo: I have implemented the suggested changes, Thanks :)

hokolomopo (commented):
@robodoo r+

PR #3229, opened 2023-11-21
[FIX] Spreadsheet: Allow shift-scroll to work on mac os

rrahir:
On mac os, using the wheel of a mouse while pressing the `shiftKey`of the keyboard will automatically swap the event deltaX and deltaY[1].

This revision proposes to let MacOs do the swapping so we can have the same behaviour cross-platform by scrolling while pressing the `shift` key. This approach is consistent with the experience with both Excel Online and Google Sheets.

[1] Note that this only applies to external mouses. The trackpad of a macbook will behave 'windows-like' but it can track in both directions by default.

Task: 3603771

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/helpers/dom_helpers.ts
+++ b/src/components/helpers/dom_helpers.ts
@@ -18,3 +18,7 @@ export function gridOverlayPosition() {
 export function getOpenedMenus(): HTMLElement[] {
   return Array.from(document.querySelectorAll<HTMLElement>(".o-spreadsheet .o-menu"));
 }
+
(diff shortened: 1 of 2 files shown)


hokolomopo:
Note that I don't think it's the same problem, but I cannot scroll horizontally on linux either

LucasLefevre (commented):
robodoo r+ then :green_apple:

PR #3230, opened 2023-11-21
[IMP] topbar: allow to insert checkbox in a cell

anhe-odoo:
## Task Description

Now that we have the datavalidation, this PR aims to add a simple item in the Insert menu to let the user directly insert a checkbox in a cell without having to use the Data-validation side-pannel. To mimic Google sheet behavior, we are now also able to remove a checkbox from a cell in the same way we can delete the content of a cell (e.g. with 'delete' or 'backspace' key).

## Related Task
- Task: [3175856](https://www.odoo.com/web#id=3175856&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/actions/menu_items_actions.ts
```diff
@@ -436,10 +436,27 @@ export const OPEN_CF_SIDEPANEL_ACTION = (env: SpreadsheetChildEnv) => {
 };
 
 export const INSERT_LINK = (env: SpreadsheetChildEnv) => {
-  let { col, row } = env.model.getters.getActivePosition();
+  const { col, row } = env.model.getters.getActivePosition();
   env.model.dispatch("OPEN_CELL_POPOVER", { col, row, popoverType: "LinkEditor" });
 };
 
+export const INSERT_CHECKBOX = (env: SpreadsheetChildEnv) => {
```
LucasLefevre, line 443: you can inline the implementation if it's only used once.

src/actions/menu_items_actions.ts
```diff
@@ -436,10 +436,27 @@ export const OPEN_CF_SIDEPANEL_ACTION = (env: SpreadsheetChildEnv) => {
...
+    sheetId,
+    rule: {
+      id: env.model.uuidGenerator.uuidv4(),
+      criterion: {
+        type: "isBoolean",
+        values: [],
+      },
+    },
+  });
+  env.model.dispatch("UPDATE_CELL", { sheetId, col, row, content: "FALSE" });
```
LucasLefevre, line 457: two dispatches means two history steps (you would need to hit Ctrl+Z twice to undo the change). We might need to add a new "helper" command to do that. e.g. INSERT_CHECKBOX

src/plugins/core/data_validation.ts
```diff
@@ -122,6 +122,24 @@ export class DataValidationPlugin
         this.addDataValidationRule(cmd.sheetId, { ...cmd.rule, ranges });
         break;
       }
+      case "DELETE_CONTENT": {
```
LucasLefevre, line 125: what about an UPDATE_CELL with `content: ""` ?

LucasLefevre (commented):
yop, quick look :)

src/plugins/core/data_validation.ts
anhe-odoo: It's a strange case in fact. For example, selecting the cell and emptying the composer should do nothing to the checkbox, but copying an empty cell over the "checkboxed" cell should clear it, so we should have a way to differentiate these things

LucasLefevre: ok, fair if it's done on purpose :+1: 
Is it tested ?

src/actions/menu_items_actions.ts
anhe-odoo: After a discussion with Adrien, we decided to change the way Checkbox is added in the datavalidation rule and always add "FALSE" as the default content

src/actions/insert_actions.ts
```diff
@@ -248,6 +248,26 @@ export const insertLink: ActionSpec = {
   icon: "o-spreadsheet-Icon.INSERT_LINK",
 };
 
+export const insertCheckbox: ActionSpec = {
+  name: _t("Checkbox"),
+  execute: (env) => {
+    const { sheetId, col, row } = env.model.getters.getActivePosition();
+    const ranges = [env.model.getters.getRangeDataFromXc(sheetId, toXC(col, row))];
```
hokolomopo, line 255: should be the whole selected range, not only the anchor cell no ? At least it work like that in GSheet

src/helpers/clipboard/clipboard_cells_state.ts
```diff
@@ -627,6 +627,10 @@ export class ClipboardCellsState extends ClipboardCellsAbstractState {
   private pasteDataValidation(origin: CellPosition, target: CellPosition) {
     const rule = this.getters.getValidationRuleForCell(origin);
     if (!rule) {
+      const targetRul = this.getters.getValidationRuleForCell(target);
```
hokolomopo, line 630:
```suggestion
      const targetRule = this.getters.getValidationRuleForCell(target);
```

src/plugins/core/data_validation.ts
```diff
@@ -122,6 +124,24 @@ export class DataValidationPlugin
...
+        const zones = cmd.target;
+        const sheetId = cmd.sheetId;
+        for (const zone of zones) {
+          for (let row = zone.top; row <= zone.bottom; row++) {
+            for (let col = zone.left; col <= zone.right; col++) {
+              const dataValidation = this.getValidationRuleForCell({ sheetId, col, row });
+              if (!dataValidation || dataValidation.criterion.type !== "isBoolean") {
```
hokolomopo, line 134: functionally is kinda strange that pressing "Delete" removes the checkboxes but not any other DV rule...

src/plugins/core/data_validation.ts
```diff
@@ -192,16 +212,19 @@ export class DataValidationPlugin
...
       const cell = this.getters.getCell(position);
-      const { sheetId, col, row } = position;
       const style: Style = {
         ...cell?.style,
         align: cell?.style?.align ?? "center",
         verticalAlign: cell?.style?.verticalAlign ?? "middle",
       };
-      this.dispatch("UPDATE_CELL", { sheetId, col, row, style });
+      let updatedCell: Omit<UpdateCellCommand, "type"> = { ...position, style };
+      if (cell?.content !== "true" && cell?.content !== "TRUE") {
```
hokolomopo, line 224:
```suggestion
      if (cell?.content.toLowerCase() !== "true") {
```
more robust if it's ever possible to have cell values be somthing like `TrUe`

tests/data_validation/data_validation_checkbox_component.test.ts
```diff
@@ -28,6 +33,24 @@ describe("Checkbox in model", () => {
     });
   });
 
+  test("Checkbox are removed when clearing the content of the cell", () => {
```
hokolomopo, line 36: Also test that other rule types aren't removed if it's intended to work like that

src/helpers/clipboard/clipboard_cells_state.ts
```diff
@@ -627,6 +627,10 @@ export class ClipboardCellsState extends ClipboardCellsAbstractState {
...
+      if (targetRul) {
+        this.adaptDataValidationRule(target.sheetId, targetRul, [], [toXC(target.col, target.row)]);
```
hokolomopo, line 632: Nitpick: i'd add a comment that we're removing the DV rule on the target, kinda hard to see what we're doing by reading the code normally. adaptDataValidationRule are kinda obscure ^^'

hokolomopo (commented):
Inserting checkboxes on a range doesn't work, otherwise some details and it looks fine :)

src/plugins/core/data_validation.ts
anhe-odoo: Gsheet has the same behavior except for the combobox, which should probably be insertable in the same way in another task, so I though we should only keep the checkbox for the moment, and see for the combobox when it will be insertable :)

src/helpers/clipboard/clipboard_cells_state.ts
```diff
@@ -627,6 +627,16 @@ export class ClipboardCellsState extends ClipboardCellsAbstractState {
   private pasteDataValidation(origin: CellPosition, target: CellPosition) {
     const rule = this.getters.getValidationRuleForCell(origin);
     if (!rule) {
+      const targetRule = this.getters.getValidationRuleForCell(target);
```
rrahir, line 630: Is this related to the new feature? I feel like it's more of a bug fix or maybe the situation could not occur beforehand?

src/plugins/core/data_validation.ts
rrahir: imo, the data validation update upon "delete content" is kind of unrelated to the insertion. If anything it is something that was overlooked in the first implementation and should be added in parallel, but not type by type

src/plugins/core/data_validation.ts
```diff
@@ -192,16 +212,19 @@ export class DataValidationPlugin
...
       const cell = this.getters.getCell(position);
-      const { sheetId, col, row } = position;
       const style: Style = {
         ...cell?.style,
         align: cell?.style?.align ?? "center",
         verticalAlign: cell?.style?.verticalAlign ?? "middle",
       };
-      this.dispatch("UPDATE_CELL", { sheetId, col, row, style });
+      let updatedCell: Omit<UpdateCellCommand, "type"> = { ...position, style };
+      if (cell?.content.toLowerCase() !== "true") {
```
rrahir, line 224: the data validation also works for the evaluated functions that return a boolean value.
 `A1:-> =1=1` should be ticked by default

src/helpers/clipboard/clipboard_cells_state.ts
anhe-odoo: It's not really a bug fix but a change in the desired behavior we discussed with Francois. But it should probably be added in a separated commit

src/components/side_panel/data_validation/data_validation_panel.ts
```diff
@@ -19,6 +20,21 @@ export class DataValidationPanel extends Component<Props, SpreadsheetChildEnv> {
...
+  setup(): void {
+    if (this.props.ruleID) {
+      this.onPreviewClick(this.props.ruleID);
+    }
+    onWillUpdateProps((nextProps: Props) => {
+      if (nextProps.ruleID !== this.props.ruleID) {
+        if (nextProps.ruleID) {
+          this.onPreviewClick(nextProps.ruleID);
+        } else {
+          this.onExitEditMode();
```
LucasLefevre, line 32: What about adding another entry in the side panel registry to be able to open the DV "editor" directly ?
`onWillUpdateProps` are always smelly ;)
It would also keep this component simpler

LucasLefevre (commented):
quick suggestion ;)

src/components/side_panel/data_validation/data_validation_panel.ts
anhe-odoo: Done :)

src/registries/side_panel_registry.ts
```diff
@@ -62,6 +63,11 @@ sidePanelRegistry.add("DataValidation", {
   Body: DataValidationPanel,
 });
 
+sidePanelRegistry.add("DataValidationEditor", {
+  title: _t("Data validation editor"),
```
LucasLefevre, line 67: I would have the same title. Such that the title is the same no matter how this side panel is open
```suggestion
  title: _t("Data validation"),
```

tests/menus/menu_items_registry.test.ts
```diff
@@ -929,6 +929,59 @@ describe("Menu Item actions", () => {
     restoreDefaultFunctions();
   });
 
+  describe("Insert -> Checkbox", () => {
+    test("In an empty cell", () => {
```
LucasLefevre, line 933: I'd move all those tests with the DV plugins tests (either core or evaluation, there's no obvious place since it involves both). Most of them are testing edge cases which are handled inside the plugins, and they are not specific to the menu item in the registry.
So instead of `doAction` => use `addDataValidation` like other tests in those files.

You can leave only one simple test case here to test the menu item itself.

EDIT: actually, the file `data_validation_checkbox_plugin.test.ts` is the perfect file ;)

Also, the file `data_validation_checkbox_component.test.ts` does not test the component and seem a vague copy-pasted version of the other file.

tests/data_validation/data_validation_checkbox_component.test.ts
LucasLefevre, line 36: move those to `data_validation_checkbox_plugin.test.ts`

tests/data_validation/data_validation_list_component.test.ts
```diff
@@ -233,6 +238,32 @@ describe("autocomplete in composer", () => {
...
       expect(model.getters.getEditionMode()).toBe("inactive");
     });
+
+    test("Dropdown lists are kept when clearing the content of a non-empty cell", () => {
+      expect(model.getters.getDataValidationListCellsPositions()).toEqual([
+        { sheetId: model.getters.getActiveSheetId(), col: 0, row: 0 },
+      ]);
+      setCellContent(model, "A1", "hello");
+      deleteContent(model, ["A1"]);
+      expect(model.getters.getDataValidationListCellsPositions()).toHaveLength(1);
```
LucasLefevre, line 248: you can check that "hello" was cleared

tests/data_validation/data_validation_list_component.test.ts
```diff
@@ -233,6 +238,32 @@ describe("autocomplete in composer", () => {
...
+    test("Dropdown lists are removed when clearing the content of an empty cell", () => {
+      expect(model.getters.getDataValidationListCellsPositions()).toEqual([
+        { sheetId: model.getters.getActiveSheetId(), col: 0, row: 0 },
+      ]);
+      setCellContent(model, "A1", "");
+      deleteContent(model, ["A1"]);
+      expect(model.getters.getDataValidationListCellsPositions()).toHaveLength(0);
+    });
+
+    test("Dropdown lists are removed when clearing the content of an empty cell", () => {
```
LucasLefevre, line 260: this test title is the same as the test above. So I'm not sure what it's supposed to test

LucasLefevre (commented):
only minor suggestions :)

src/plugins/core/data_validation.ts
```diff
@@ -122,6 +123,31 @@ export class DataValidationPlugin
...
+          for (let row = zone.top; row <= zone.bottom; row++) {
+            for (let col = zone.left; col <= zone.right; col++) {
+              const dataValidation = this.getValidationRuleForCell({ sheetId, col, row });
+              if (
+                !dataValidation ||
+                (dataValidation.criterion.type !== "isBoolean" &&
+                  !(
+                    dataValidation.criterion.type === "isValueInList" &&
+                    !this.getters.getCell({ sheetId, col, row })?.content
+                  ))
```
LucasLefevre, line 139: maybe written the other way around, the condition would be easier to read ?

tests/data_validation/data_validation_list_component.test.ts
```diff
@@ -233,6 +238,33 @@ describe("autocomplete in composer", () => {
       expect(getCellContent(model, "A1")).toBe("hello");
       expect(model.getters.getEditionMode()).toBe("inactive");
     });
+
+    test("Dropdown lists are kept when clearing the content of a non-empty cell", () => {
```
LucasLefevre, line 242: those tests could be move elsewhere than this file (component tests). I guess there are already other plugin tests for lists

anhe-odoo: While there were a plugin for the checkbox, I don't think there is any plugin tests for dropdown list, I'll create one :)

LucasLefevre (approved):
robodoo r+

PR #3235, opened 2023-11-21
[FIX] *: Support metaKey modifier in event handlers

rrahir:
The navigation & click handlers did not account for mac users. In general, the `ctrlKey` modifier is switched with `metaKey` for those users as some key combinations with `ctrlKey` are reserved by the system (i.e. `ctrlKey+click` triggers the context menu or `ctrlKey+ArrowLeft/Right` switches desktops by default).

This revision aims to adopt the same approcha as other spreadsheets (Excel Online & Google Sheets) to provide seamless experience and convince more people to adopt this library.

Task: 3606161

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/helpers/dom_helpers.ts
+++ b/src/components/helpers/dom_helpers.ts
@@ -1,3 +1,5 @@
+const macRegex = /Mac/i;
+
 /**
  * Return true if the event was triggered from
  * a child element.
@@ -20,5 +22,14 @@ export function getOpenedMenus(): HTMLElement[] {
 }
 
 export function isMacOS(): boolean {
-  return navigator.userAgent.toUpperCase().indexOf("MAC") >= 0;
+  return Boolean(macRegex.test(navigator.userAgent));
+}
+
+/**
+ * @param {KeyboardEvent | MouseEvent} ev
+ * @returns Returns true if the event was triggered with the "ctrl" modifier pressed.
+ * On Mac, this is the "meta" or "command" key.
+ */
+export function isCtrlKey(ev: KeyboardEvent | MouseEvent): boolean {
+  return isMacOS() ? ev.metaKey : ev.ctrlKey;
 }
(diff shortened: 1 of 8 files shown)


hokolomopo (commented):
Maybe add a test somewhere where we would test if things work on MAC ?

src/components/helpers/dom_helpers.ts
```diff
@@ -18,3 +20,16 @@ export function gridOverlayPosition() {
 export function getOpenedMenus(): HTMLElement[] {
   return Array.from(document.querySelectorAll<HTMLElement>(".o-spreadsheet .o-menu"));
 }
+
+function isMacOS(): boolean {
+  return Boolean(navigator.userAgent.match(macRegex));
```
LucasLefevre, line 25: `return macRegex.test(navigator.userAgent);`
not that it really matters here, but `test` is faster than `match`

LucasLefevre: actually, the same helper was also merged by https://github.com/odoo/o-spreadsheet/pull/3229/files

just needs a small rebase ^^

tests/components/spreadsheet.test.ts
```diff
@@ -176,6 +176,29 @@ describe("Spreadsheet", () => {
     expect(document.querySelectorAll(".o-sidePanel").length).toBe(0);
   });
 
+  test("Mac user use metaKey, not CtrlKey", async () => {
+    ({ model, parent, fixture } = await mountSpreadsheet({
+      model: new Model({ sheets: [{ id: "sh1" }] }),
+    }));
+    Object.defineProperty(window, "navigator", {
```
LucasLefevre, line 183: shouldn't it be cleaned ?

LucasLefevre (commented):
I believe you :green_apple:

tests/components/spreadsheet.test.ts
rrahir: it is now ^^

PR #3250, opened 2023-11-24
[FIX] tokenize: improve performance

LucasLefevre:
## Description:

Some performance improvements for the tokenizer (~-17% with both commits)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/formulas/tokenizer.ts
+++ b/src/formulas/tokenizer.ts
@@ -68,7 +68,7 @@ export function tokenize(str: string, locale = DEFAULT_LOCALE): Token[] {
 }
 
 function tokenizeDebugger(chars: TokenizingChars): Token | null {
-  if (chars.current() === "?") {
+  if (chars.current === "?") {
     chars.shift();
     return { type: "DEBUGGER", value: "?" };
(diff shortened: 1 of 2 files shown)


src/formulas/tokenizer.ts
```diff
@@ -227,21 +224,23 @@ function tokenizeInvalidRange(chars: TokenizingChars): Token | null {
 class TokenizingChars {
   private text: string;
   private currentIndex: number = 0;
+  current: string;
```
hokolomopo, line 227: might wanna have `private _current` and `get current() { return this._current; }` to avoid messing everything up by mistake. If it doesnt impact the perfs

LucasLefevre: I must admit I thought about, and didn't try. But it will inevitably impact the perf.
And since it's a very specialized data structure used only once here. I think it's ok to be less safe for the performance's sake

hokolomopo:
@robodoo r+ rebase-ff

PR #3304, opened 2023-11-30
[ADD] store engine: a state management system for components

LucasLefevre:
## Description:

This commit introduces a new way to manage state and business logic
of components based on owl reactivity.

Read `src/store_engine/README.md` for a full explanation and details.

The main goals are:

- Centralize the state used by several components across the UI.
- Move as much business logic as possible out of components (easier/faster
  to test).
- Interact with UI from the outside world (example: we currently cannot open
  a sidepanel from the outside of the app.)

Why not plugins?
===============
While these goals could be addressed by relocating state and business logic
to plugins, several disadvantages are associated with this approach:

- Excessive (deep) rendering occurs at every command dispatch, even if only a
  single component has actually changed.

- Plugins need boilerplate for getters and commands. It's also usually not
  worth the noise in getters and commands, all for the sake of a single component.

- When the state is specific to a single component, commands are only handled
  by one plugin. This means command multi-casting and shared getters doesn't
  bring any benefit.

- Unlike plugins, a single component can be mounted multiple times. Each
  component could possess its own state.

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


rrahir (commented):
comments of the day

src/store_engine/store.ts
```diff
@@ -0,0 +1,97 @@
...
+ * Remove the first element of a tuple
+ * @example
+ * type A = SkipFirst<[number, string, boolean]> // [string, boolean]
+ */
+type SkipFirst<T extends any[]> = T extends [any, ...infer U] ? U : never;
+
+/**
+ * TODO
+ */
+export function createMetaStore<T extends unknown>(storeName: string): StoreConstructor<T> {
```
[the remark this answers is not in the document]
LucasLefevre: not sure... "abstract" usually means something different (like an abstract class that you need to implement)
But I'm totally open to suggestions, "meta" isn't really great either.
`createInjectableExternalStore` ?

src/model.ts
```diff
@@ -590,10 +593,12 @@ export class Model extends EventBus<any> implements CommandDispatcher {
    * context. This is probably the way we should do if we want to be able to
    * freeze a part of the grid (so, we would need to render different zones)
    */
-  drawGrid(context: GridRenderingContext) {
-    // we make sure here that the viewport is properly positioned: the offsets
-    // correspond exactly to a cell
-    for (let [renderer, layer] of this.renderers) {
+  drawGrid(context: GridRenderingContext, layer: LAYERS) {
```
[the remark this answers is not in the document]
LucasLefevre: rename in all plugins

src/components/helpers/highlight_hook.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  useHighlights({
+    get highlights() {
+      return hoverState.hovered ? highlightProvider.highlights : [];
+    },
+  });
+  useEffect(
+    () => {
+      env.model.dispatch("RENDER_CANVAS");
+    },
```
VincentSchippefilt, line 19: in this case, I wonder why we still need a command that will be handled by only one component and have no side effects (history). Should we not have a simple function call to model.drawGrid() or useGridDrawing ?

LucasLefevre: you're right, we don't need this command. It's only dispatched to trigger a rendering. Which can be achieved with a simple function call :+1:

LucasLefevre: mouais, not that straight forward because it doesn't have a direct access to the canvas

VincentSchippefilt (approved):

VincentSchippefilt:
robodoo r+

LucasLefevre:
robodoo rebase-ff

PR #3312, opened 2023-12-01
[FIX] find&replace: Only switch sheet from the side panel

rrahir:
The recent improvement in [1] introduced a faulty behaviour, which is to try to dispatch commands in the finalize step of a dispatch. The functional intension was to let a user interact with the sheet and that such actions would automatically be mirrored in the f&r sidepanel. Ultimately, this would prove problematic as we need to perform the search in the finalize (because of the evaluation process) and that could lead to a tentative to switch sheets in order to let the user jump at the position of the match.

This revision addresses the issue by modifying the functional behaviour. As of now, if the currently selected match is not on the current sheet, the user will be redirected to the match position IIF they dispatch a "find&replace" related command (i.e. if they interact with the F&R sidepanel).

[1] https://github.com/odoo/o-spreadsheet/pull/2330

Task: 3618775

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_feature/find_and_replace.ts
+++ b/src/plugins/ui_feature/find_and_replace.ts
@@ -108,7 +108,7 @@ export class FindAndReplacePlugin extends UIPlugin {
 
   finalize() {
     if (this.isSearchDirty) {
-      this.refreshSearch();
+      this.refreshSearch(false);
       this.isSearchDirty = false;
     }
   }
@@ -161,10 +161,10 @@ export class FindAndReplacePlugin extends UIPlugin {
   /**
    * refresh the matches according to the current search options
    */
-  private refreshSearch() {
+  private refreshSearch(jumpToMatchSheet = true) {
     this.selectedMatchIndex = null;
     this.findMatches();
-    this.selectNextCell(Direction.current);
+    this.selectNextCell(Direction.current, jumpToMatchSheet);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/plugins/ui_feature/find_and_replace.ts
```diff
@@ -114,6 +116,7 @@ export class FindAndReplacePlugin extends UIPlugin {
     if (this.isSearchDirty) {
       this.refreshSearch();
       this.isSearchDirty = false;
+      this.jumpToMatchSheet = true;
```
hokolomopo, line 119: I would add an argument `jumpToMatchSheet` to `refreshSearch` and `selectNextCell`, and set it to false here. Seems way less fishy than handle a plugin state

src/plugins/ui_feature/find_and_replace.ts
```diff
@@ -112,8 +114,9 @@ export class FindAndReplacePlugin extends UIPlugin {
 
   finalize() {
     if (this.isSearchDirty) {
-      this.refreshSearch();
+      this.refreshSearch(this.jumpToMatchSheet);
```
LucasLefevre, line 117: I think Adrien is right, `this.jumpToMatchSheet` isn't needed :wink: 
```suggestion
      this.refreshSearch(false);
```

PR #3345, opened 2023-12-12
[FIX] sheet_interactive: rename sheet in readonly mode

dhrp-odoo:
## Description:

This PR addresses two distinct issues identified in the current codebase.

**Commit 1:**
- Resolves an issue where renaming a sheet with whitespace failed to trigger the sheet rename dialog.
- Implements whitespace trimming for sheet names to ensure the dialog opens as intended.

**Commit 2:**
- The commit rectifies an issue where users in readonly mode were able to invoke the sheet renaming dialog by double clicking on its name. However, they were unable to complete the renaming process in readonly mode.
- Introduces an early return, preventing the invocation of the renaming dialog altogether when the application is in readonly mode.

Task: : [3621086](https://www.odoo.com/web#id=3621086&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/ui/sheet_interactive.ts
+++ b/src/helpers/ui/sheet_interactive.ts
@@ -9,7 +9,7 @@ export function interactiveRenameSheet(env: SpreadsheetChildEnv, sheetId: UID, e
     if (name === null || name === placeholder) {
       return;
     }
-    if (name === "") {
+    if (name.trim() === "") {
       interactiveRenameSheet(env, sheetId, _lt("The sheet name cannot be empty."));
     }
     const result = env.model.dispatch("RENAME_SHEET", { sheetId, name });
--- a/src/helpers/ui/sheet_interactive.ts
+++ b/src/helpers/ui/sheet_interactive.ts
@@ -3,17 +3,21 @@ import { _lt } from "../../translation";
 import { CommandResult, SpreadsheetChildEnv, UID } from "../../types";
 
 export function interactiveRenameSheet(env: SpreadsheetChildEnv, sheetId: UID, errorText?: string) {
+  if (env.model.getters.isReadonly()) {
+    return;
+  }
+
   const placeholder = env.model.getters.getSheetName(sheetId);
   const title = _lt("Rename Sheet");
   const callback = (name: string | null) => {
     if (name === null || name === placeholder) {
       return;
     }
-    if (name.trim() === "") {
-      interactiveRenameSheet(env, sheetId, _lt("The sheet name cannot be empty."));
-    }
     const result = env.model.dispatch("RENAME_SHEET", { sheetId, name });
     if (!result.isSuccessful) {
+      if (result.reasons.includes(CommandResult.MissingSheetName)) {
+        interactiveRenameSheet(env, sheetId, _lt("The sheet name cannot be empty."));
+      }
       if (result.reasons.includes(CommandResult.DuplicatedSheetName)) {
         interactiveRenameSheet(
           env,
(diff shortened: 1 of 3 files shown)


src/plugins/core/sheet.ts
```diff
@@ -651,6 +651,9 @@ export class SheetPlugin extends CorePlugin<SheetState> implements SheetState {
     const { orderedSheetIds, sheets } = this;
     const name = cmd.name && cmd.name.trim().toLowerCase();
 
+    if (name && name.length > 100) {
```
rrahir, line 654: we could maybe take the same character limit as Excel, which is 31 chars.

Also I would flag this as an improvement unless you found a really blocking behaviour for the users.

dhrp-odoo: Should this now be merged into the master branch?

I've completed all the suggested changes :)

rrahir: let's target master, yes :)

src/plugins/core/sheet.ts
```diff
@@ -651,6 +651,9 @@ export class SheetPlugin extends CorePlugin<SheetState> implements SheetState {
     const { orderedSheetIds, sheets } = this;
     const name = cmd.name && cmd.name.trim().toLowerCase();
 
+    if (name && name.length > 31) {
```
LucasLefevre, line 654: You can create a constant `MAX_SHEET_NAME_LENGTH = 31` and use it here and interpolated in the error message.

Small question: does 31 come from somewhere or did you chose it arbitrarily (Gsheet seems to have chosen 51)

src/helpers/ui/sheet_interactive.ts
```diff
@@ -9,7 +9,7 @@ export function interactiveRenameSheet(env: SpreadsheetChildEnv, sheetId: UID, e
     if (name === null || name === placeholder) {
       return;
     }
-    if (name === "") {
+    if (name.trim() === "") {
```
LucasLefevre, line 16: Could you add a test for this ?

src/plugins/core/sheet.ts
```diff
@@ -651,6 +651,9 @@ export class SheetPlugin extends CorePlugin<SheetState> implements SheetState {
...
+      return CommandResult.SheetNameTooLong;
```
LucasLefevre, line 655: This can break existing spreadsheets.
Let's say the initial revisions look like this:
1. `{ type: "RENAME_SHEET", name: "A very very[...]long name" }`
1. `{ type: "UPDATE_CELL": content: "='A very very[...]long name'!A1" }`

With this commit, the rename would be refused, which would make the cell update invalid.
This would be better for master and it technically requires an upgrade script to "fix" the commands in the database.

dhrp-odoo: @LucasLefevre, I have a couple of questions:

1. If a sheet name exceeds the new length limit, what should be the updated name?

2. Do we need to modify all commands that reference sheet names to comply with the new character limit?

LucasLefevre: Hello @dhrp-odoo 

> Do we need to modify all commands that reference sheet names to comply with the new character limit?

Yes, that's what it implies ... and it's very cumbersome... :/

We discussed about it with the team. The real issue comes with excel export. An easy and acceptable solution (at least for stable) is to trigger an warning before exporting if any sheet have a name longer than 31 characters, asking the user to rename the sheet. Nobody ever complained about that limitation. In master, we might do at some point an automatic cropping at the excel export. But it's not a priority though.

rrahir:
@robodoo r+

LucasLefevre:
robodoo rebase-ff

PR #3357, opened 2023-12-18
popover store

rrahir:
## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/components/grid/hovered_cell_store.ts
```diff
@@ -0,0 +1,24 @@
...
+
+  handle(cmd: Command) {
+    switch (cmd.type) {
+      case "ACTIVATE_SHEET":
+        this.clear();
+    }
+  }
+
+  hover(position: Position) {
+    this.col = position?.col;
```
anhe-odoo, line 16: Is the "?" really needed ? Same question for the line under

src/components/popover/cell_popover_store.ts
```diff
@@ -1,59 +1,50 @@
...
+      (this.persistentPopover && { isOpen: true, ...this.persistentPopover }) || { isOpen: false }
+    );
+  }
+
+  get isOpen() {
+    return this.persistentPopover !== undefined;
   }
 
-  getCellPopover({ col, row }: Partial<Position>): ClosedCellPopover | PositionedCellPopover {
+  get cellPopover(): ClosedCellPopover | PositionedCellPopoverComponent {
```
anhe-odoo, line 47: I personally find it strange to mix `XXXPopover` with a `XXXPopoverComponent`. Is there a significant difference between them to explain the naming difference ?

src/types/cell_popovers.ts
```diff
@@ -42,9 +49,9 @@ type OpenCellPopover<C extends ComponentConstructor> = {
 
 export type CellPopoverComponent<
   C extends MaxSizedComponentConstructor = MaxSizedComponentConstructor
-> = ClosedCellPopover | OpenCellPopover<C>;
+> = ClosedCellPopover | OpenCellPopoverComponent<C>;
```
anhe-odoo, line 52: Another example of the type naming mix. Here we have a XPopoverComponent that can be a YPopover or a ZPopoverComponent. From this line, I understand that `ClosedCellPopover` is of the same kind of `OpenCellPopoverComponent`, while I would first thing that it should be something similar to `OpenCellPopover`, which is not the case. This is probably just some nitpick and/or a stupid comment, but I was thinking of it

src/components/grid/hovered_cell_store.ts
rrahir: indeed, it's a relicate of the dev. tahnks for spotting it :)

PR #3358, opened 2023-12-19
[FIX] format: fix chaining of FORMAT.LARGE.NUMBER

hokolomopo:
## Description

Chaining FORMAT.LARGE.NUMBER formulas was adding the postfix multiple times. This commit fixes the issue by adding a check to see if the postfix is already present.

Task: : [3633947](https://www.odoo.com/web#id=3633947&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/tests/functions/module_custom.test.ts
+++ b/tests/functions/module_custom.test.ts
@@ -1,7 +1,7 @@
 import { Model } from "../../src";
-import { setCellContent, setCellFormat } from "../test_helpers/commands_helpers";
+import { setCellContent, setCellFormat, setFormat } from "../test_helpers/commands_helpers";
 import { getCellContent, getCellError } from "../test_helpers/getters_helpers";
-import { evaluateCellText } from "../test_helpers/helpers";
+import { evaluateCellText, target } from "../test_helpers/helpers";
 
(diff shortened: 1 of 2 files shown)


tests/functions/module_custom.test.ts
```diff
@@ -129,4 +129,17 @@ describe("FORMAT.LARGE.NUMBER formula", () => {
 
     expect(evaluateCellText("A1", { A1: '=FORMAT.LARGE.NUMBER(100, "something")' })).toBe("#ERROR");
   });
+
+  test("Chaining FORMAT.LARGE.NUMBER does nothing", () => {
```
rrahir, line 133: maybe specify that it's specifically when we don't provide the second argument

hokolomopo: Yeah,  it's still broken if we chain the formula using different units ... Not sure if we can fix that. We could check that the suffix in the format looks like a large number suffix, but what if there's a currency that have the same look ?

LucasLefevre (approved):
robodoo r+

PR #3360, opened 2023-12-19
[FIX] evaluation: track dependencies with R-tree 

LucasLefevre:
## Description:

Currently, when building the dependency graph for a cell with the formula
`=SUM(A1:A1000)`, we iterate through each position in the range `A1:A1000` to
create the inverse dependency mapping. However, when the range is very large,
and many such formulas exist, the process can take a significant amount of
time (quadratic C x R, where C is the number of cells and R is the size of the
range(s)).

We found 3 spreadsheets on our production database which are unusable because
building the dependency graph never finishes. This commit brings that time
down to a fraction of a second.

One noteworthy observation is that a complete inverse dependency mapping is
unnecessary, as only a small subset of cells is likely to ever be updated.

The objective of this commit is to avoid the construction of the entire
inverse mapping while still maintaining an efficient method to identify
impacted cells when a zone on which it depends is modified.

The chosen strategy uses an R-tree[^1] data structure to index the dependency
relations and efficiently locate the cells when necessary.

Since implementing an R-tree is non-trivial and existing implementations are
available, we opt for the R-tree provided by the rbush[^2] library.
This library is ~500 lines of code, appears stable, and is widely used
(1.2 million weekly npm downloads).

[^1]: https://en.wikipedia.org/wiki/R-tree
[^2]: https://github.com/mourner/rbush

Task: : [3646902](https://www.odoo.com/web#id=3646902&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,8 @@
       "license": "LGPL-3.0-or-later",
       "dependencies": {
         "@odoo/owl": "2.2.6",
-        "bootstrap": "^5.1.3"
+        "bootstrap": "^5.1.3",
+        "rbush": "^3.0.1"
       },
       "devDependencies": {
         "@prettier/plugin-xml": "^2.0.1",
@@ -19,6 +20,7 @@
         "@types/chart.js": "2.9.3",
         "@types/jest": "^27.0.1",
         "@types/node": "^13.13.23",
+        "@types/rbush": "^3.0.3",
         "babel-eslint": "^10.1.0",
         "body-parser": "^1.19.0",
         "chart.js": "2.9.3",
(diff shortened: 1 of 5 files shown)


src/types/range.ts
```diff
@@ -18,6 +18,12 @@ export interface Range extends Cloneable<Range> {
   readonly rangeData: RangeData;
 }
 
+// TODO rename this
+export interface SmallRange {
+  sheetId: UID;
+  zone: Zone;
+}
```
LucasLefevre, line 25: Please help me.
I originally thought it would be great to be able to assign a `Range` to this `SmallRange` (that's the idea of having the two properties "sheetId" and "zone" properties). It's kinda a range but it lacks the (possible) unboundedness.

Should it be more like a "SheetyZone" such as ` { sheetId, left, right, top, bottom }` ?

pro-odoo: I would go for the idea to reduce the scope. If it's not use elsewhere for now, let's keep the name simple and clear. If necessary, it will still be possible to rename it later.

laa-odoo (approved):

rrahir:
@robodoo r+

PR #3371, opened 2024-01-02
[IMP] evaluation: `evaluateFormula` no longer throws

rrahir:
The getter `evaluateFormula` required to be wrapped in a try/catch statement to handle the evaluation errors but as this is quite error prone, we decided to change it to no longer throw. The getter will instead the value of the error that was previously thrown.

Task: 3576149

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui/evaluation_conditional_format.ts
+++ b/src/plugins/ui/evaluation_conditional_format.ts
@@ -125,41 +125,37 @@ export class EvaluationConditionalFormatPlugin extends UIPlugin {
     this.computedIcons[activeSheetId] = {};
     const computedStyle = this.computedStyles[activeSheetId];
     for (let cf of this.getters.getConditionalFormats(activeSheetId).reverse()) {
-      try {
-        switch (cf.rule.type) {
-          case "ColorScaleRule":
-            for (let range of cf.ranges) {
-              this.applyColorScale(range, cf.rule);
-            }
-            break;
-          case "IconSetRule":
-            for (let range of cf.ranges) {
-              this.applyIcon(range, cf.rule);
-            }
-            break;
-          default:
-            for (let ref of cf.ranges) {
-              const zone: Zone = this.getters.getRangeFromSheetXC(activeSheetId, ref).zone;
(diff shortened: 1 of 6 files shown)


src/plugins/ui/evaluation_conditional_format.ts
```diff
@@ -188,7 +184,7 @@ export class EvaluationConditionalFormatPlugin extends UIPlugin {
         return percentile(rangeValues, Number(threshold.value) / 100, true);
       case "formula":
         const value = threshold.value && this.getters.evaluateFormula(threshold.value);
-        return !(value instanceof Promise) ? value : null;
+        return typeof value == "number" ? value : null;
```
LucasLefevre, line 187: this is a behavior change, right ?
Technically, `value` could be a string before (even if it should not according to the method return type).
Are we missing a test where the formula evaluates to a string ?

src/plugins/ui/evaluation_conditional_format.ts
```diff
@@ -380,7 +376,7 @@ export class EvaluationConditionalFormatPlugin extends UIPlugin {
       if (cell && cell.evaluated.type === CellValueType.error) {
         return false;
       }
-      const values = rule.values.map(parsePrimitiveContent);
+      const values = rule.values?.map(parsePrimitiveContent) || [];
```
LucasLefevre, line 379: is this necessary ? The type definition states that `values` is always defined. But maybe it's lying :p

LucasLefevre (commented):
looking good :+1: 
Only 2 small questions

src/plugins/ui/evaluation_conditional_format.ts
rrahir: the data fed to `Model` unfortunately bypasses it :/. I feel like TS could warn me that the chaining `?.` is unnecessary since values is always defined 🤔

LucasLefevre (approved):
robodoo r+

PR #3372, opened 2024-01-03
Master export header groups adrm

hokolomopo:
## [IMP] xlsx export: don't export default row size:

The previous export of row sizes in the xlsx was inconsistent:
there was a comment saying that we always force our own row size in the
export, but 20 lines later we din't export the row if it was the default
size.

Now we only export the row size if the user has changed it manually.
We decided to do this rather than forcing our own row size in the xlsx,
because forcing a row/size would break the auto row size feature
when re-importing (since the row would be considered as manually sized).

## [IMP] xlsx: import and export header groups

Add the import and the export of the headers groups from/to xlsx.

Task: : [3637643](https://www.odoo.com/web#id=3637643&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/core/header_visibility.ts
+++ b/src/plugins/core/header_visibility.ts
@@ -207,7 +207,7 @@ export class HeaderVisibilityPlugin extends CorePlugin {
           if (sheet.rows[row] === undefined) {
             sheet.rows[row] = {};
           }
-          sheet.rows[row].isHidden = this.hiddenHeaders[sheet.id]["ROW"][row];
+          sheet.rows[row].isHidden ||= this.hiddenHeaders[sheet.id]["ROW"][row];
         }
       }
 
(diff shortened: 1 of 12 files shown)


src/plugins/core/header_visibility.ts
```diff
@@ -207,7 +207,8 @@ export class HeaderVisibilityPlugin extends CorePlugin {
           if (sheet.rows[row] === undefined) {
             sheet.rows[row] = {};
           }
-          sheet.rows[row].isHidden = this.hiddenHeaders[sheet.id]["ROW"][row];
+          sheet.rows[row].isHidden =
```
anhe-odoo, line 210: `sheet.rows[row].isHidden ||= this.hiddenHeaders[sheet.id]["ROW"][row];` ?

LucasLefevre (approved):
robodoo rebase-ff r+
Thanks :)

PR #3413, opened 2024-01-09
[FIX] side_panel: cf rule range bugs

khpa-odoo:
## Description:
_Steps to reproduce :-_

Bug 1 : 
- Change the range to ""  
- Focus out => selection input should have a red border even without clicking on 
  confirm

Bug 2 : 
- Enter "" as range and save
- Now change it to a valid range and confirm => red border and error message do 
  not disappear even if you confirm

Bug 3 :
- Enter an invalid range (eg. A2:B3:C3) and save
- Now change it to a valid range and confirm => red border disappears but the 
  error message does not disappear

Bug 4 :
- Enter an invalid range (with invalid sheet name) => red border appears 
  (correct), but there is no error message and you can still save the cf rule

With this commit the following amendments have been made:

Earlier, the condition checked to display selection input border in red 
considered 'empty range' as a valid input. That condition is changed to 
consider it as an invalid range.

The condition to check for invalid ranges, while dispatching command result 
'InvalidRange', is altered to use the getter `isRangeValid` so that cf rules 
with ranges containing invalid sheet name cannot be saved.

To ensure that the error message appears if and only if the input is invalid
(i.e. it disappears on entering a valid range as input, even without clicking on 
confirm or save), error messages for invalid range are now handled from 
the `onRangesChanged` method instead of `saveConditionalFormat` method.

Task: : [3651293](https://www.odoo.com/web#id=3651293&menu_id=4720&cids=2&action=333&active_id=2328&model=project.task&view_type=form)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/conditional_formatting/conditional_formatting.ts
+++ b/src/components/side_panel/conditional_formatting/conditional_formatting.ts
@@ -1,6 +1,6 @@
 import { Component, onWillUpdateProps, useExternalListener, useState } from "@odoo/owl";
 import { DEFAULT_COLOR_SCALE_MIDPOINT_COLOR } from "../../../constants";
-import { colorNumberString, rangeReference } from "../../../helpers/index";
+import { colorNumberString } from "../../../helpers/index";
 import { _t } from "../../../translation";
 import {
   CancelledReason,
@@ -381,10 +381,6 @@ export class ConditionalFormattingPanel extends Component<Props, SpreadsheetChil
     return this.env.model.getters.getConditionalFormats(this.env.model.getters.getActiveSheetId());
   }
 
-  get isRangeValid(): boolean {
-    return this.state.errors.includes(CommandResult.EmptyRange);
-  }
-
   errorMessage(error: CancelledReason): string {
     return CfTerms.Errors[error] || CfTerms.Errors.Unexpected;
   }
@@ -446,9 +442,10 @@ export class ConditionalFormattingPanel extends Component<Props, SpreadsheetChil
 
   saveConditionalFormat() {
     if (this.state.currentCF) {
-      const invalidRanges = this.state.currentCF.ranges.some((xc) => !xc.match(rangeReference));
-      if (invalidRanges) {
-        this.state.errors = [CommandResult.InvalidRange];
+      if (
+        this.state.errors.includes(CommandResult.EmptyRange) ||
+        this.state.errors.includes(CommandResult.InvalidRange)
+      ) {
         return;
       }
       const sheetId = this.env.model.getters.getActiveSheetId();
@@ -595,6 +592,15 @@ export class ConditionalFormattingPanel extends Component<Props, SpreadsheetChil
   }
 
   onRangesChanged(ranges: string[]) {
+    if (ranges.length === 0) {
+      this.state.errors = [CommandResult.EmptyRange];
+      return;
+    }
+    if (ranges.some((xc) => !this.env.model.getters.isRangeValid(xc))) {
+      this.state.errors = [CommandResult.InvalidRange];
+      return;
+    }
+    this.state.errors = [];
     if (this.state.currentCF) {
       this.state.currentCF.ranges = ranges;
     }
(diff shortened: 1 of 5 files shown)


src/components/side_panel/conditional_formatting/conditional_formatting.ts
```diff
@@ -446,9 +442,10 @@ export class ConditionalFormattingPanel extends Component<Props, SpreadsheetChil
 
   saveConditionalFormat() {
     if (this.state.currentCF) {
-      const invalidRanges = this.state.currentCF.ranges.some((xc) => !xc.match(rangeReference));
-      if (invalidRanges) {
-        this.state.errors = [CommandResult.InvalidRange];
+      if (
+        this.state.errors.includes(CommandResult.EmptyRange) ||
+        this.state.errors.includes(CommandResult.InvalidRange)
+      ) {
```
hokolomopo, line 448:
```suggestion
      if (this.state.errors.lenght) {
```
is there a situation where we want to allow saving if there is an error somewhere ?

tests/components/conditional_formatting.test.ts
```diff
@@ -363,7 +363,40 @@ describe("UI of conditional formats", () => {
...
+      triggerMouseEvent(selectors.buttonSave, "click");
+      await nextTick();
+      expect(dispatch).not.toHaveBeenCalledWith("ADD_CONDITIONAL_FORMAT");
+      const errorString2 = document.querySelector(selectors.error);
+      expect(errorString2!.textContent).toBe("The range is invalid");
+    });
+
+    test("display error message if and only if invalid range", async () => {
+      triggerMouseEvent(selectors.buttonAdd, "click");
+      await nextTick();
```
hokolomopo, line 377: use "await click" instead of triggerMouseEvent + nextTick in this whole test

tests/components/conditional_formatting.test.ts
```diff
@@ -363,7 +363,40 @@ describe("UI of conditional formats", () => {
...
+      await nextTick();
```
hokolomopo, line 378: is it necessary to haev 2 nextTicks ?

tests/components/selection_input.test.ts
```diff
@@ -343,10 +343,18 @@ describe("Selection Input", () => {
...
-  test("don't show red border initially", async () => {
-    await createSelectionInput();
+    await writeInput(0, "");
+    expect(fixture.querySelectorAll("input")[0].value).toBe("");
+    expect(fixture.querySelectorAll("input")[0].getAttribute("style")).toBe("color: #000;");
+    expect(fixture.querySelectorAll("input")[0].classList).toContain("o-invalid");
+    await writeInput(0, "C1");
+    expect(fixture.querySelectorAll("input")[0].value).toBe("C1");
+    expect(fixture.querySelectorAll("input")[0].getAttribute("style")).not.toBe("color: #000;");
     expect(fixture.querySelectorAll("input")[0].classList).not.toContain("o-invalid");
```
hokolomopo, line 353: kinda unnecessary, we're already testing A1 & B1, no need to test C1

hokolomopo (commented):
:+1:

tests/components/conditional_formatting.test.ts
rrahir: yup, we need a second tick for the selection input component to init itself. Based on its props, it dispatches a creation of selection input elements in the model then uses the new information to render the inputs created... typically something that will disappear with the stores.

src/components/side_panel/conditional_formatting/conditional_formatting.ts
rrahir: Then you'd try to save a CF with invalid arguments -> you have some errors because you forgot some argument, w/e. and since you have an errro ,you can't save anymore ^^you're stuck :p

rrahir:
@robodoo r+

PR #3456, opened 2024-01-17
[FIX] composer: Persistent composition on grid input

### LINKED ISSUE #1881: a Chinese Character Input Method is not support
when i use  Chinese Character Input Method like sougou, it doesn't work

rrahir:
## Description:

closes #1881 

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


rrahir:
@robodoo rebase-ff

src/components/composer/composer/composer.ts
```diff
@@ -314,24 +321,52 @@ export class Composer extends Component<Props, SpreadsheetChildEnv> {
...
       this.isKeyStillDown = true;
     }
   }
 
+  onPaste(ev: ClipboardEvent) {
+    if (this.env.model.getters.getEditionMode() === "inactive") {
+      return;
+    } else {
+      ev.stopPropagation();
+    }
```
hokolomopo, line 339:
```suggestion
    if (this.env.model.getters.getEditionMode() !== "inactive") {
      ev.stopPropagation();
    } ```

File: src/components/composer/content_editable_helper.ts [hunk 2]
```diff
@@ -1,9 +1,10 @@
 import { HtmlContent } from "./composer/composer";
 
 export class ContentEditableHelper {
+  // todo make el private and expose dedicated methods
```
hokolomopo, line 4: todo

src/components/composer/grid_composer/grid_composer.ts
```diff
@@ -43,53 +37,58 @@ export class GridComposer extends Component<Props, SpreadsheetChildEnv> {
...
-      this.composerState.delimitation = {
-        width: el!.parentElement!.clientWidth,
-        height: el!.parentElement!.clientHeight,
-      };
+    onWillUpdateProps(() => {
+      const isEditing = this.env.model.getters.getEditionMode() !== "inactive";
+      if (this.isEditing !== isEditing) {
+        this.isEditing = isEditing;
+        if (!isEditing) {
+          this.rect = this.defaultRect;
```
hokolomopo, line 53: defaultRect seems useless, we don't use this.rect for the composer style when we're not editing. And looking at how it's used in the Composer I think we can set it to undefined.

src/components/composer/grid_composer/grid_composer.ts
```diff
@@ -43,53 +37,58 @@ export class GridComposer extends Component<Props, SpreadsheetChildEnv> {
...
+          return;
+        }
+        const position = this.env.model.getters.getPosition();
+        const zone = this.env.model.getters.expandZone(
+          this.env.model.getters.getActiveSheetId(),
+          positionToZone(position)
+        );
+        this.rect = this.env.model.getters.getVisibleRect(zone);
```
hokolomopo, line 61: do we even need to bother checking if the edition mode changed and change this.rect at all ? Can't we juste creat `get rect(){ ...}` that does all that ? this.rect isn't in a state nor is it complex to compute

Would avoid having 2 variables `this.isEditing` and `this.rect`

src/components/grid/grid.xml
```diff
@@ -17,26 +18,14 @@
...
-      </t>
-      <t else="1">
-        <input
-          class="position-absolute"
-          style="z-index:-1000;"
-          t-on-input="onInput"
-          t-on-contextmenu="onInputContextMenu"
-          t-ref="hiddenInput"
-        />
-      </t>
```
hokolomopo, line 39: 🥳

src/components/grid/grid.ts
```diff
@@ -269,12 +267,12 @@ export class Grid extends Component<Props, SpreadsheetChildEnv> {
     PAGEUP: () => this.env.model.dispatch("SHIFT_VIEWPORT_UP"),
   };
 
-  focus() {
+  defaultFocus() {
```
hokolomopo, line 270:
```suggestion
  focusDefaultElement() {
```

src/helpers/focus_manager.ts
```diff
@@ -0,0 +1,11 @@
+export class FocusManager {
```
hokolomopo, line 1: Should find a better name, I'd expect more of a FocusManagar than to register a single element and expose a method to focus it.

tests/components/composer.test.ts
```diff
@@ -429,7 +429,15 @@ describe("ranges and highlights", () => {
 });
 
 describe("composer", () => {
-  test("starting the edition with enter, the composer should have the focus", async () => {
+  test("grid composer is not visible when not editing", async () => {
+    expect(model.getters.getEditionMode()).toBe("inactive");
+    const gridComposerEl = fixture.querySelector(".o-grid-composer") as HTMLDivElement;
+    expect(gridComposerEl.style.zIndex).toBe("-1000");
```
hokolomopo, line 435: Can't we use a less barbaric way to hide the input btw ? Or does display `hidden` and things like that don't work with IME ?

tests/components/composer.test.ts
```diff
@@ -612,11 +621,10 @@ describe("composer", () => {
     setCellContent(model, "A1", "[label](http://odoo.com)");
     await hoverCell(model, "A1", 400);
     expect(fixture.querySelector(".o-link-tool")).not.toBeNull();
-    await startComposition();
     await typeInComposerGrid(" updated");
     await keyDown("Enter");
     const cell = getCell(model, "A1") as LinkCell;
-    expect(cell.link.label).toBe("label updated");
+    // expect(cell.link.label).toBe("label updated");
```
hokolomopo, line 627: comment

hokolomopo (commented):
Yay no more random input focused 🥳

src/components/composer/content_editable_helper.ts
rrahir: that 's a cherry pick of a saas-16.2 (i think) commit, the todo still stands to thins day and frankly not blocking, I'd rather keepit this way and address in master after the store imp

src/helpers/focus_manager.ts
rrahir: I agree, it feels like manager means that we could do "lots" of stuff but I don't have a better idea right know.
`defaultFocus(Service)` maybe?
When you think "manager" what do you expect it to do more? (just fishing for some ideas, I really don't have a clue, and tbh, don't really care as we'll drop this with the store, though the discussion might arise witht the store as well)

tests/components/composer.test.ts
rrahir: I need to check this properly tomorrow, I'm fairly sure it was the issue, yes ^^

rrahir: I confirm that using `display: none` is a no go ^^

src/components/composer/grid_composer/grid_composer.ts
rrahir: note to myself I found:
that one was more of a "typing" hack as we'd have to force `this.rect!` (in `get composerProps` or  `get containerStyle` for example, so I figured it was best to set a "falsy" rect than anything else but I propose an improvement in the commit 'review 1 from hoko'

src/components/composer/grid_composer/grid_composer.ts
rrahir: it needs to be persistent because you want it to keep having the same position (DOM-wise) when we scroll (which in turn renders the whole app, the getter would then be updated and move :/

src/helpers/focus_manager.ts
LucasLefevre: maybe just `FocusableElement` ?
`env.focusableElement.focus()` feels natural

src/selection_stream/event_stream.ts
```diff
@@ -68,7 +70,7 @@ export class EventStream<Event> {
     if (this.observers.find((sub) => sub.owner === owner)) {
       throw new Error("You are already subscribed forever");
     }
-    if (this.mainSubscription?.owner) {
+    if (this.mainSubscription?.owner && this.mainSubscription.owner !== owner) {
```
LucasLefevre, line 73: could you explain a bit more in what way it's an issue to call the release callback on yourself ?
The commit message says it should not but doesn't really explain why it's an issue if it does.

LucasLefevre (commented):
I didn't test the IME itself, but the composer seems to work just like before :+1: 
I have only one naming suggestion for the focus manager.  Hopefully we can improve all this focus management in master :)

src/helpers/focus_manager.ts
rrahir: fine with me :)

src/components/spreadsheet/spreadsheet.ts
```diff
@@ -181,6 +182,7 @@ export class Spreadsheet extends Component<SpreadsheetProps, SpreadsheetChildEnv
       toggleSidePanel: this.toggleSidePanel.bind(this),
       _t: Spreadsheet._t,
       clipboard: navigator.clipboard,
+      FocusableElement: new FocusableElement(),
```
LucasLefevre, line 185: capitalised ? it looks like a class while it's actually an instance
```suggestion
      focusableElement: new FocusableElement(),
```

rrahir: sadge find & replace

LucasLefevre (commented):
robodoo r+ rebase-ff

PR #3474, opened 2024-01-17
[FIX] SelectionInput: Allow commands that target non focused input

rrahir:
Currently, dispatching a selectionInput related command can be rejected by the focused input even if the said command does not concern it.

Example:
- Set up 2 selection inputs, one of them with a single range (like for a line chart)
- Focus the single range input
- click on the "Add range" of the multi-range input -> nothing happens

The command are rejected as they pass through the currently focused input, which might not have the same properties (e.g. `hasSingleRange`) than the one targetted by the command.

Task: 3689504

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui/selection_inputs_manager.ts
+++ b/src/plugins/ui/selection_inputs_manager.ts
@@ -53,15 +53,36 @@ export class SelectionInputsManagerPlugin extends UIPlugin {
   allowDispatch(cmd: Command): CommandResult {
     switch (cmd.type) {
       case "FOCUS_RANGE":
+      case "CHANGE_RANGE":
+      case "ADD_EMPTY_RANGE":
+      case "REMOVE_RANGE":
+        if (!this.inputs[cmd.id]) {
+          return CommandResult.InvalidInputId;
+        }
+    }
+
+    switch (cmd.type) {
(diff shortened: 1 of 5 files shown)


src/plugins/ui/selection_inputs_manager.ts
```diff
@@ -51,17 +51,38 @@ export class SelectionInputsManagerPlugin extends UIPlugin {
...
+        const input = this.inputs[cmd.id];
+        if (input.inputHasSingleRange && cmd.value.split(",").length > 1) {
+          return CommandResult.MaximumRangesReached;
+        }
+        break;
+      }
     }
-    if (this.currentInput) {
-      return this.currentInput.allowDispatch(cmd);
-    }
```
LucasLefevre, line 64: Then we can remove the `allowDispatch` in `selection_input.ts`, right ?

rrahir: 100% correct, I amenede the commit in that sense ;-)

LucasLefevre (approved):
robodoo r+

PR #3536, opened 2024-01-26
[IMP] table: introduce table style

hokolomopo:
## [IMP] table: introduce table style

This commit improves the old concept of tables (filterTable) that only
allowed to filter the table. Now, the table can be configured and
styled in a side panel.

Introduce the TableStylePlugin that compute and store the style of
each cell of a table. The cellTableStyle is then combined with the
cell style and the CF style in ui_sheet.

The only styles available for now are the styles defined in the
TABLE_PRESET constants. These are based on the OpenXML specification
of table styles, with some color changes.

Further improvements might include:
 - allow to define custom table styles
 - dynamic tables that take the size of a spread formula result

## [FIX] style: fix getCellComputedStyle style composition

If a CF had a style with false attributes (eg. bold: false or
fillColor: undefined) they would overwrite the style of the cell
in the `getCellComputedStyle` getter.

It goes without saying that overwriting with undefined values made no
sense. Overwriting with false values was also wrong, because from an
user perspective there should be no difference between a cell with no
`bold` set, and a cell with `bold` toggled to true, then to false.

This would also impact the composition with table style in the following
commits.

## [REF] filters: de-classify FilterTable/Filter

Transform classes FilterTable and Filter into POJOs. Having class
instances made was convenient while the filter table were static,
but it becomes more annoying when we want to dynamically change
the filtered zone.

## [REF] filters: rename filterTable into table

With the introduction of table style, `filterTable` didn't make sense
anymore as a variable name, since atable did not necessarily have
filters. Rename all the occurrences of `filterTable` into `table`.:

## [REF] filters: rename plugin FilterPlugin to TablePlugin

Task: : [3636461](https://www.odoo.com/web#id=3636461&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/components/filters/filter_icon/filter_icon.ts
```diff
@@ -46,4 +60,10 @@ export class FilterIcon extends Component<Props, SpreadsheetChildEnv> {
   get isFilterActive(): boolean {
     return this.env.model.getters.isFilterActive(this.props.cellPosition);
   }
+
+  get getIconClass(): string {
+    const cellStyle = this.env.model.getters.getCellComputedStyle(this.props.cellPosition);
+    const lum = relativeLuminance(cellStyle.fillColor || "#fff");
```
LucasLefevre, line 66: `luminance`

src/components/format_as_table_widget/format_as_table_widget.ts
```diff
@@ -0,0 +1,76 @@
...
+import {
+  CustomTablePopoverMouseEvent,
+  TableStylePopover,
+} from "../side_panel/table/table_style_popover/table_style_popover";
+
+interface State {
+  popoverProps: PopoverProps | undefined;
+}
+
+css/* scss */ ``;
```
LucasLefevre, line 21: :eyes:

src/components/side_panel/table/table_panel.ts
```diff
@@ -0,0 +1,150 @@
...
+    }
+  }
+`;
+
+export class TablePanel extends Component<Props, SpreadsheetChildEnv> {
+  static template = "o-spreadsheet-TablePanel";
+  static components = { TableStylePicker, SelectionInput, ValidationMessages, Checkbox, Section };
+  static props = { onCloseSidePanel: Function };
+
+  _t = _t;
```
LucasLefevre, line 52: :eyes:

src/components/side_panel/table/table_panel.xml
```diff
@@ -0,0 +1,83 @@
+<templates>
+  <t t-name="o-spreadsheet-TablePanel">
+    <div class="o-table-panel">
+      <Section>
+        <t t-set-slot="title">Table options</t>
+        <div class="d-flex flex-row">
+          <div class="w-50">
+            <Checkbox
+              label="_t('Filter Button')"
```
LucasLefevre, line 9: this unfortunately doesn't work because the source term will never be extracted by the extraction script. It's not looking at the (string) value of xml attributes, interpret it as javascript and look for `_t`. The script is quite naive and is a simple static extraction.

Options are:
- move that to the typescript `this.filterButtonLabel = _t('Filter Button')`
- define a variable in the template `<t t-set="filter_button_label">Filter Button</t>`. Here because `Filter Button` looks like regular text content, it is extracted, just like if it was `<span>Hello world</span>`

src/components/side_panel/table/table_style_popover/table_style_popover.ts
```diff
@@ -0,0 +1,89 @@
...
+      background: #f5f5f5;
+    }
+
+    &:hover {
+      background: #ddd;
+    }
+  }
+`;
+
+export type CustomTablePopoverMouseEvent = MouseEvent & { hasClosedTableStylePopover?: boolean };
```
LucasLefevre, line 48: :roll_eyes: :see_no_evil:

src/plugins/index.ts
```diff
@@ -57,7 +58,7 @@ export const corePluginRegistry = new Registry<CorePluginConstructor>()
   .add("sheet", SheetPlugin)
   .add("header grouping", HeaderGroupingPlugin)
   .add("header visibility", HeaderVisibilityPlugin)
-  .add("filters", FiltersPlugin)
+  .add("filters", TablePlugin)
```
LucasLefevre, line 61:
```suggestion
  .add("tables", TablePlugin)
```

src/components/side_panel/table/table_style_popover/table_style_popover.ts
hokolomopo: :bullettrain_front: :fire:

hokolomopo:
TODO: take into account config.numberOfHeaders to compute the contentZone

src/plugins/ui_core_views/table_style.ts
```diff
@@ -0,0 +1,185 @@
...
+          absoluteTableStyle.styles[colInSheet][rowInSheet] = relativeTableStyle.styles[col][row];
+        }
+      }
+      return absoluteTableStyle;
+    });
+  }
+
+  /**
+   * Get the actual table config that will be used to compute the table style. It is different from the config of the table
+   * because of hidden rows and columns in the sheet. For example remove the hidden rows from config.numberOfHeaders.
```
LucasLefevre, line 111: could you reduce the length of those lines (to max ~80 chars) ? When I have two files next to each other in vs code, I can't read the end.

LucasLefevre (commented):
Really nice feature !
Some comments/questions/suggestions of the day. I didn't look the tests yet.

src/helpers/table_presets.ts
```diff
@@ -0,0 +1,365 @@
...
+  light: Color;
+  medium: Color;
+  dark: Color;
+  mediumBorder: Color;
+  highlight: Color;
+}
+
+type StyleTemplate = (colorSet: ColorSet) => TableStyle;
+
+const COLOR_SETS = {
```
LucasLefevre, line 37: From a functional POV, the colors here and the colors in the color picker should probably be the same. Or at least, the table colors should appear in the "custom colors" as soon as it's used.

We could ask the designer to review those colors to have nice combination.

src/helpers/table_helpers.ts
```diff
@@ -0,0 +1,298 @@
...
+        zones.push({ top: headerRows, left: i, bottom: lastRow - totalRows, right: i });
+      }
+      break;
+  }
+
+  return zones;
+}
+
+export function getTableStyleName(styleId: string, tableStyle: TableStyle): string {
+  return tableStyle.colorName ? `${tableStyle.colorName}, ${styleId}` : styleId;
```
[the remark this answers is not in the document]
hokolomopo: Yeah I wasn't sure about this either. But having the style id beats the name of the style being just "Purple" IMO. But maybe we should either have "real" style names, or just don't show any name at all.

I don't have a strong opinion on this.

src/components/side_panel/table/table_style_picker/table_style_picker.xml
```diff
@@ -0,0 +1,33 @@
...
+      <div class="d-flex flex-row overflow-hidden">
+        <div
+          class="o-table-style-list-item"
+          t-att-class="{ 'selected': styleId === props.table.config.styleId }"
+          t-foreach="getDisplayedTableStyles()"
+          t-as="styleId"
+          t-key="styleId"
+          t-att-title="getStyleName(styleId)"
+          t-on-click="() => this.onStylePicked(styleId)">
+          <div class="o-table-style-preview">
```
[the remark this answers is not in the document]
hokolomopo: The class have a different CSS in both components tho, so I'm not sure it make sense to put it in the child ?

src/helpers/table_canvas_helpers.ts
```diff
@@ -0,0 +1,110 @@
+import { ComputedTableStyle } from "../types/table";
+
+export function drawPreviewTable(
```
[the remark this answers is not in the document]
hokolomopo: TBH I didn't think about using HTML before I was already done with a canvas renderer 😅 I can test what it would look like in HTML if you want.

src/helpers/table_helpers.ts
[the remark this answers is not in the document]
LucasLefevre: I think no name at all is fine. We can ask the PO

src/helpers/table_helpers.ts
```diff
@@ -0,0 +1,298 @@
...
+  nOfCols: number,
+  nOfRows: number
+): Border[][] {
+  const borders: Border[][] = [];
+  for (let i = 0; i < nOfCols; i++) {
+    borders.push([]);
+    for (let j = 0; j < nOfRows; j++) {
+      borders[i].push({});
+    }
+  }
```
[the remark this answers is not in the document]
hokolomopo: Damn that's helpful.

Btw should I re-type everything to Matrix<Border> insteaed of Border[][] ? Or do we keep MAtrix<> only for functions ?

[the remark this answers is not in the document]
LucasLefevre: > Btw should I re-type everything to Matrix insteaed of Border[][] ? Or do we keep MAtrix<> only for functions ?

I don't really care

src/plugins/core/tables.ts
```diff
@@ -0,0 +1,487 @@
...
+      }
+      case "ADD_COLUMNS_ROWS":
+        this.onAddColumnsRows(cmd);
+        break;
+      case "REMOVE_COLUMNS_ROWS":
+        this.onDeleteColumnsRows(cmd);
+        break;
+      case "CREATE_TABLE": {
+        const zone = union(...cmd.target);
+        const id = this.uuidGenerator.uuidv4();
```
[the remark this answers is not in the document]
hokolomopo: Yeah it's not exported and user specific. A bit starnge, but it works 🤷

hokolomopo:
> probably linked to borders and not to these tables but the border of the total dépasse à droite image

Yeah I'm aware, it was already present before. I was planning of doing a dive in the renderer sometime after.

src/helpers/table_presets.ts
hokolomopo: Uh Francois gave me colors from a designer, I forgot to test them 😅

src/components/side_panel/table/table_panel.ts
```diff
@@ -0,0 +1,151 @@
...
+  static components = { TableStylePicker, SelectionInput, ValidationMessages, Checkbox, Section };
+  static props = { onCloseSidePanel: Function };
+
+  state!: State;
+
+  setup() {
+    const selectedTable = this.env.model.getters.getFirstTableInSelection();
+    if (!selectedTable) {
+      this.props.onCloseSidePanel();
+      return;
```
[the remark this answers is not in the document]
hokolomopo: What do you do about the typing then ? An exclamation point and trust that we really did check before opening the panel ? Check `if (state.table === undefined)` everywhere ? A t-if ?

src/components/side_panel/table/table_panel.xml
```diff
@@ -0,0 +1,85 @@
...
+        </div>
+      </Section>
+      <Section>
+        <TableStylePicker table="state.table"/>
+      </Section>
+      <Section>
+        <t t-set-slot="title">Data Range</t>
+        <SelectionInput
+          t-key="state.table.id"
+          ranges="() => [this.state.tableXc]"
```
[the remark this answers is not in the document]
hokolomopo: Still need to rebase 🙈

tests/table/tables_plugin.test.ts
```diff
@@ -0,0 +1,836 @@
...
+    });
+
+    test("Table zone is expanded when a cell on the row below the table but not below the table had a content", () => {
+      setCellContent(model, "B4", "Something");
+      createTable(model, "A1:A3");
+      setCellContent(model, "A4", "Something Else");
+      expect(zoneToXc(model.getters.getTables(sheetId)[0].zone)).toEqual("A1:A4");
+    });
+
+    test("Table zone isn't expanded when there was another table table below it", () => {
```
LucasLefevre, line 311: I would CTRL+H all `isn't` in the file to `is not` to avoid the single quote that breaks vscode jest extension

tests/table/tables_plugin.test.ts
```diff
@@ -0,0 +1,836 @@
...
+  describe("Grid manipulation", () => {
+    let model: Model;
+    let sheetId: UID;
+    beforeEach(() => {
+      model = new Model();
+      createTable(model, "C3:F6");
+      updateFilter(model, "C3", ["C"]);
+      updateFilter(model, "D3", ["D"]);
+      updateFilter(model, "E3", ["E"]);
+      updateFilter(model, "F3", ["F"]);
```
LucasLefevre, line 335: :angry: beforeEach with lots of setup.
I can already see myself trying to debug one of those tests because I broke something and trying to keep in my head the initial state of the spreadsheet with all those filtered values, and mentally updating my mental representation with the operations in the tests.
=> :exploding_head: :skull:

tests/table/tables_plugin.test.ts
```diff
@@ -0,0 +1,836 @@
...
+        expect(getFilterValues(model)).toEqual([
+          { zone: "C3:C6", value: ["C"] },
+          { zone: "D3:D6", value: ["D"] },
+          { zone: "E3:E6", value: ["E"] },
+          { zone: "F3:F6", value: ["F"] },
+        ]);
+      });
+    });
+
+    describe("Delete columns", () => {
```
LucasLefevre, line 409: Deleting all columns of the table is not tested (same for rows I guess)

tests/table/table_style_plugin.test.ts
```diff
@@ -0,0 +1,241 @@
...
+      expect(getCellStyle("B5")).toMatchObject({ fillColor: tableBackgroundColor }); // banded color is alternating even if a row is hidden
+    });
+  });
+
+  describe("Table borders", () => {
+    const outerBordersTableStyle = {
+      styleId: "TableStyleLight8",
+      numberOfHeaders: 0,
+      bandedRows: false,
+    };
```
LucasLefevre, line 110: To ensure no test mutates this object which would affect other tests in a non-deterministic way
```suggestion
    const outerBordersTableStyle = {
      styleId: "TableStyleLight8",
      numberOfHeaders: 0,
      bandedRows: false,
    } as const;
```

tests/table/table_helpers.test.ts
```diff
@@ -0,0 +1,409 @@
...
+}
+
+beforeEach(() => {
+  tableConfig = deepCopy(TEST_TABLE_CONFIG);
+  TABLE_PRESETS["TestStyle"] = deepCopy(TEST_TABLE_STYLE);
+  tableStyle = TABLE_PRESETS["TestStyle"];
+});
+
+describe("Table cell style", () => {
+  const wholeTableStyle = { fillColor: "#f00", textColor: "#f00" };
```
LucasLefevre, line 48: `as const`

LucasLefevre (commented):
some more comments :)

src/components/top_bar/top_bar.ts
```diff
@@ -164,7 +166,7 @@ export class TopBar extends Component<Props, SpreadsheetChildEnv> {
   menus: Action[] = [];
   EDIT = ACTION_EDIT;
   FORMAT = ACTION_FORMAT;
-  VIEW = ACTION_VIEW;
+  DATA = ACTION_DATA;
```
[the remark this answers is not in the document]
hokolomopo: Already exists :upside_down_face:

tests/table/tables_plugin.test.ts
hokolomopo: I tend to agree with you, but in this case all this setup is actually used and tested in every test. Copying that in each test would make it easier to read each test individually, but would also means plus 100 duplicated  lines of codes

src/components/side_panel/table/table_panel.ts
```diff
@@ -0,0 +1,151 @@
...
+      newTableZone,
+    }).reasons;
+  }
+
+  onRangeConfirmed() {
+    const newZone = toZone(this.state.tableXc);
+    const result = this.updateTable(undefined, newZone);
+
+    const position = { col: newZone.left, row: newZone.top };
+    this.env.model.selection.selectZone({
```
rrahir: I guess my comment about the zone belongs here instead - why should we change the selection upon confirmation?

src/components/side_panel/table/table_style_picker/table_style_picker.xml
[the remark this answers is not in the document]
rrahir: maybe don't use the same class name then

src/components/icons/icons.xml
```diff
@@ -955,4 +924,28 @@ https://fontawesome.com/license -->
       <circle cx="2" cy="12.5" r="1"/>
     </svg>
   </t>
+  <t t-name="o-spreadsheet-Icon.EDIT_TABLE">
+    <svg xmlns="http://www.w3.org/2000/svg" width="18" height="18">
```
rrahir, line 928: the margin on the side is less than a pixe, which I thinkg is rounded to 0, it looks like the styles is broken. Could it be possible to crop it a bit horizontally?

src/components/tables/table_dropdown_button/table_dropdown_button.ts
```diff
@@ -0,0 +1,74 @@
...
+    const tableConfig = { ...this.tableConfig, styleId };
+    const result = interactiveCreateTable(this.env, sheetId, tableConfig);
+    if (result.isSuccessful) {
+      this.env.openSidePanel("TableSidePanel", {});
+    }
+    this.closePopover();
+  }
+
+  onClick(ev: CustomTablePopoverMouseEvent) {
+    if (ev.hasClosedTableStylesPopover || this.state.popoverProps) {
```
rrahir, line 37: please don't :'( Is it really necessary or could you maintain a state to toggle it?

src/components/side_panel/table/table_panel.ts
hokolomopo: Yeah. The side panel is editing the table that is currently inside the selection. If we move the table, but don't change the selection, the side panel will close itself since there is no table in the selection :x

src/components/tables/table_dropdown_button/table_dropdown_button.ts
LucasLefevre: I think we have to find a better way globally (including for the existing similar case). I'd say we can copy that (shitty) way of managing this toggle and refactor it in a dedicated PR later (not sure how we should do it yet).

src/plugins/ui_feature/ui_sheet.ts
```diff
@@ -180,6 +189,33 @@ export class SheetUIPlugin extends UIPlugin {
...
+    const cellTableBorder = this.getters.getCellTableBorder(position) || {};
+
+    // Use removeFalsyAttributes to avoid overwriting borders with undefined values
+    const border = { ...cellTableBorder, ...removeFalsyAttributes(cellBorder) };
+
+    return isObjectEmptyRecursive(border) ? null : border;
+  }
+
+  getCellComputedStyle(position: CellPosition): Style {
+    const cell = this.getters.getCell(position);
```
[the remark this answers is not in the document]
hokolomopo: I guess there's no reason we couldn't. It's very fast, but it's called often.

There's no real impact on perfs tho, so if it's fine by you I'll do it in  afollow-up PR. I'd have to create another plugin (I don't want to add somewhat compex things in the fourre tout that is sheet_ui).

src/components/icons/icons.xml
hokolomopo: Added a bit of  apadding in the ActionButton. It sounds a bit dumb to have a space of 18px for icons, but that we cannot actually use icons of 18px because we need a padding inside the icon...

src/helpers/table_helpers.ts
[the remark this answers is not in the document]
hokolomopo: After discussing with PO:  let's keep the name, it between sometime useful and not harmful. Use case example: manager tells its employees to always use "Light8" style in their reports.

src/helpers/table_presets.ts
hokolomopo: A follow UP PR will be to put the same colors in the colorPicker that the ones in the tables. to have some kind of global "theme".

src/plugins/core/range.ts
```diff
@@ -284,6 +288,19 @@ export class RangeAdapter implements CommandHandler<CoreCommand> {
     });
   }
 
+  expandRange(range: Range, dimension: Dimension, quantity: number): Range {
```
LucasLefevre, line 291: What about `extendRange` instead? (similar to the `extendTableDown` in the table plugin).
I originally though this would expand the range to take merges into account just like what `expandZone` does.

src/components/side_panel/table_panel/table_panel.ts
```diff
@@ -53,23 +53,33 @@ export class TablePanel extends Component<Props, SpreadsheetChildEnv> {
   static props = { onCloseSidePanel: Function, table: Object };
 
   state!: State;
+  /** We want to save the state of the "hasFilter" checkbox so that toggling the header off (which
+   * disable the filters), then toggling it back on doesn't change the "hasFilter" state*/
+  filtersEnabledIfPossible: boolean = false;
```
LucasLefevre, line 58: could/should be in the (reactive) state. It's error prone and here's no reason not to move it there

src/components/side_panel/table_panel/table_panel.ts
```diff
@@ -53,23 +53,33 @@ export class TablePanel extends Component<Props, SpreadsheetChildEnv> {
...
       tableXc: this.env.model.getters.getRangeString(this.props.table.range, sheetId),
     });
+    this.filtersEnabledIfPossible = this.props.table.config.hasFilters;
 
     onWillUpdateProps((nextProps: Props) => {
       if (this.props.table.id !== nextProps.table.id) {
         const sheetId = this.env.model.getters.getActiveSheetId();
         this.state.tableXc = this.env.model.getters.getRangeString(nextProps.table.range, sheetId);
         this.state.tableZoneErrors = [];
+        this.filtersEnabledIfPossible = nextProps.table.config.hasFilters;
```
LucasLefevre, line 73: for later: I'm thinking about adding a way for a side panel to define a `t-key` (with the registry). It would allow to remove a bunch of `onWillUpdateProps`

src/components/side_panel/table_panel/table_panel.ts
```diff
@@ -81,18 +91,29 @@ export class TablePanel extends Component<Props, SpreadsheetChildEnv> {
...
+    const result = this.updateNumberOfHeaders(numberOfHeaders);
+
     if (!result.isSuccessful) {
       input.value = this.props.table.config.numberOfHeaders.toString();
     }
   }
 
+  private updateNumberOfHeaders(numberOfHeaders: number) {
+    const hasFilters =
+      !!numberOfHeaders && (this.tableConfig.hasFilters || this.filtersEnabledIfPossible);
```
LucasLefevre, line 109: `numberOfHeaders > 0` is easier to read (more direct) than `!!numberOfHeaders`

src/clipboard_handlers/tables_clipboard.ts
```diff
@@ -1,110 +1,181 @@
...
-      for (const i of range(0, table.filtersValues.length)) {
+  protected pasteZone(
+    sheetId: UID,
+    col: HeaderIndex,
+    row: HeaderIndex,
+    tableCells: TableCell[][],
+    clipboardOptions?: ClipboardOptions
+  ) {
+    for (const [r, rowCells] of tableCells.entries()) {
+      for (const [c, tableCell] of rowCells.entries()) {
```
LucasLefevre, line 146: `entries()` allocates new arrays. It's not be the best if there are lot's of them.

LucasLefevre (commented):
New commits are looking good.
Maybe I'm aveugle, but I don't see an "Insert table" menu in the top bar (I think you told me there was one)

src/components/side_panel/table_panel/table_panel.ts
hokolomopo: But won't it trigger another render when we update it in the onWillUpdateProps then ?

VincentSchippefilt (approved):

VincentSchippefilt:
robodoo rebase-ff r+

PR #3545, opened 2024-01-29
[IMP] chart: improve pie chart

anhe-odoo:
## Task Description

The aims of this task is to improve the user experience when creating a pie chart :
1. User can now plot a chart only based on the count of string occurrences
2. When hovering a pie part, the corresponding dataset are highlighted
3. When hovering a legend item, the corresponding pie part are highlighted

## Related Task:
- Task: : [3370683](https://www.odoo.com/web#id=3370683&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/chart/line_bar_pie_panel/config_panel.ts
+++ b/src/components/side_panel/chart/line_bar_pie_panel/config_panel.ts
@@ -91,7 +91,7 @@ export class LineBarPieConfigPanel extends Component<Props, SpreadsheetChildEnv>
       {
         name: "aggregated",
         label: _t("Aggregate"),
-        value: this.props.definition.aggregated,
+        value: this.props.definition.aggregated ?? false,
         onChange: this.onUpdateAggregated.bind(this),
       },
     ];
--- a/src/helpers/figures/charts/chart_factory.ts
+++ b/src/helpers/figures/charts/chart_factory.ts
@@ -3,6 +3,7 @@ import {
   DEFAULT_SCORECARD_BASELINE_COLOR_UP,
   DEFAULT_SCORECARD_BASELINE_MODE,
 } from "../../../constants";
+import { isEvaluationError } from "../../../functions/helpers";
 import { chartRegistry } from "../../../registries/chart_types";
 import { _t } from "../../../translation";
 import {
@@ -21,9 +22,11 @@ import {
 } from "../../../types/chart/chart";
 import { CoreGetters, Getters } from "../../../types/getters";
 import { Validator } from "../../../types/validator";
-import { getZoneArea, zoneToXc } from "../../zones";
+import { getZoneArea, zoneToDimension, zoneToXc } from "../../zones";
 import { AbstractChart } from "./abstract_chart";
+import { createDataSets } from "./chart_common";
 import { canChartParseLabels } from "./chart_common_line_scatter";
+import { getData } from "./chart_ui_common";
 
 /**
  * Create a function used to create a Chart based on the definition
@@ -118,7 +121,8 @@ export function getChartTypes(): Record<string, string> {
  */
 export function getSmartChartDefinition(zone: Zone, getters: Getters): ChartDefinition {
   let dataSetZone = zone;
-  if (zone.left !== zone.right) {
+  const singleColumn = zoneToDimension(zone).numberOfCols === 1;
+  if (!singleColumn) {
     dataSetZone = { ...zone, left: zone.left + 1 };
   }
   const dataSets = [zoneToXc(dataSetZone)];
@@ -159,7 +163,7 @@ export function getSmartChartDefinition(zone: Zone, getters: Getters): ChartDefi
   }
 
   let labelRangeXc: string | undefined;
-  if (zone.left !== zone.right) {
+  if (!singleColumn) {
     labelRangeXc = zoneToXc({
       ...zone,
       right: zone.left,
@@ -182,6 +186,21 @@ export function getSmartChartDefinition(zone: Zone, getters: Getters): ChartDefi
       dataSetsHaveTitle,
       verticalAxisPosition: "left",
       legendPosition: newLegendPos,
+    };
+  }
+  const _dataSets = createDataSets(getters, dataSets, sheetId, dataSetsHaveTitle);
+  if (
+    singleColumn &&
+    getData(getters, _dataSets[0]).every((e) => typeof e === "string" && !isEvaluationError(e))
+  ) {
+    return {
+      title: "",
+      dataSets,
+      aggregated: true,
+      labelRange: dataSets[0],
+      type: "pie",
+      legendPosition: "top",
+      dataSetsHaveTitle: false,
     };
   }
   return {
--- a/src/helpers/figures/charts/chart_ui_common.ts
+++ b/src/helpers/figures/charts/chart_ui_common.ts
@@ -1,6 +1,7 @@
 import type { BasePlatform, ChartConfiguration, ChartOptions, ChartType } from "chart.js";
 import { ChartTerms } from "../../../components/translations_terms";
 import { MAX_CHAR_LABEL } from "../../../constants";
+import { isEvaluationError } from "../../../functions/helpers";
 import { _t } from "../../../translation";
 import { Color, Figure, Format, Getters, LocaleFormat, Range } from "../../../types";
 import { GaugeChartRuntime, ScorecardChartRuntime } from "../../../types/chart";
@@ -231,9 +232,15 @@ export function getChartDatasetValues(getters: Getters, dataSets: DataSet[]): Da
           ? truncateLabel(cell.formattedValue)
           : (label = `${ChartTerms.Series} ${parseInt(dsIndex) + 1}`);
     } else {
-      label = label = `${ChartTerms.Series} ${parseInt(dsIndex) + 1}`;
+      label = `${ChartTerms.Series} ${parseInt(dsIndex) + 1}`;
     }
     let data = ds.dataRange ? getData(getters, ds) : [];
+    if (data.every((e) => typeof e === "string" && !isEvaluationError(e))) {
+      // In this case, we want a chart based on the string occurrences count
+      // This will be done by associating each string with a value of 1 and
+      // the using the classical aggregation method to sum the values.
+      data.fill(1);
+    }
     datasetValues.push({ data, label });
   }
   return datasetValues;
--- a/src/helpers/figures/charts/pie_chart.ts
+++ b/src/helpers/figures/charts/pie_chart.ts
@@ -104,7 +104,7 @@ export class PieChart extends AbstractChart {
       title: context.title || "",
       type: "pie",
       labelRange: context.auxiliaryRange || undefined,
-      aggregated: false,
+      aggregated: context.aggregated ?? false,
     };
   }
 
@@ -122,6 +122,7 @@ export class PieChart extends AbstractChart {
       auxiliaryRange: this.labelRange
         ? this.getters.getRangeString(this.labelRange, this.sheetId)
         : undefined,
+      aggregated: this.aggregated,
     };
   }
 
--- a/src/helpers/figures/charts/bar_chart.ts
+++ b/src/helpers/figures/charts/bar_chart.ts
@@ -98,7 +98,7 @@ export class BarChart extends AbstractChart {
       dataSets: context.range ? context.range : [],
       dataSetsHaveTitle: false,
       stacked: false,
-      aggregated: false,
+      aggregated: context.aggregated ?? false,
       legendPosition: "top",
       title: context.title || "",
       type: "bar",
@@ -117,6 +117,7 @@ export class BarChart extends AbstractChart {
       auxiliaryRange: this.labelRange
         ? this.getters.getRangeString(this.labelRange, this.sheetId)
         : undefined,
+      aggregated: this.aggregated,
     };
   }
 
--- a/src/helpers/figures/charts/line_chart.ts
+++ b/src/helpers/figures/charts/line_chart.ts
@@ -96,7 +96,7 @@ export class LineChart extends AbstractChart {
       verticalAxisPosition: "left",
       labelRange: context.auxiliaryRange || undefined,
       stacked: false,
-      aggregated: false,
+      aggregated: context.aggregated ?? false,
       cumulative: false,
     };
   }
@@ -140,6 +140,7 @@ export class LineChart extends AbstractChart {
       auxiliaryRange: this.labelRange
         ? this.getters.getRangeString(this.labelRange, this.sheetId)
         : undefined,
+      aggregated: this.aggregated,
     };
   }
 
--- a/src/helpers/figures/charts/scatter_chart.ts
+++ b/src/helpers/figures/charts/scatter_chart.ts
@@ -83,7 +83,7 @@ export class ScatterChart extends AbstractChart {
       type: "scatter",
       verticalAxisPosition: "left",
       labelRange: context.auxiliaryRange || undefined,
-      aggregated: false,
+      aggregated: context.aggregated ?? false,
     };
   }
 
@@ -124,6 +124,7 @@ export class ScatterChart extends AbstractChart {
       auxiliaryRange: this.labelRange
         ? this.getters.getRangeString(this.labelRange, this.sheetId)
         : undefined,
+      aggregated: this.aggregated,
     };
   }
 
--- a/src/types/chart/chart.ts
+++ b/src/types/chart/chart.ts
@@ -68,4 +68,5 @@ export interface ChartCreationContext {
   readonly title?: string;
   readonly background?: string;
   readonly auxiliaryRange?: string;
+  readonly aggregated?: boolean;
 }
--- a/tests/figures/chart/charts_component.test.ts
+++ b/tests/figures/chart/charts_component.test.ts
@@ -11,6 +11,7 @@ import {
   createGaugeChart,
   createScorecardChart,
   createSheet,
+  setCellContent,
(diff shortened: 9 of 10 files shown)


src/helpers/figures/charts/pie_chart.ts
```diff
@@ -216,25 +225,93 @@ function getPieConfiguration(
...
-    const xLabel = tooltipItem.label || tooltipItem.dataset.label;
     const yLabel = tooltipItem.parsed.y ?? tooltipItem.parsed;
     const toolTipFormat = !format && yLabel > 1000 ? "#,##" : format;
     const yLabelStr = formatValue(yLabel, { format: toolTipFormat, locale });
 
-    return xLabel ? `${xLabel}: ${yLabelStr} (${percentage}%)` : `${yLabelStr} (${percentage}%)`;
+    return `${yLabelStr} (${percentage}%)`;
+  };
+  config.options.onHover = onHoverPieChart;
+  //@ts-ignore
```
anhe-odoo, line 241: If someone has a way to do this without a ts-ignore, I'm in !

src/components/side_panel/chart/building_blocks/label_range/label_range.ts
```diff
@@ -5,19 +5,22 @@ import { SelectionInput } from "../../../../selection_input/selection_input";
 import { Checkbox } from "../../../components/checkbox/checkbox";
 import { Section } from "../../../components/section/section";
 
+export interface Option {
```
hokolomopo, line 8: Functional note (to discuss with PO): now when hovering a pie chart item, we don't show the dataset name at all. That can be a problem if we have a pie with multiple datasets, no ? (see pie chart in demo, I have no idea if the outer circle is the first or second dataset)

src/components/side_panel/components/checkbox/checkbox.xml
```diff
@@ -7,6 +7,7 @@
         t-att-checked="props.value"
         t-on-change="onChange"
         class="align-middle"
+        t-att-disabled="props.disabled"
```
hokolomopo, line 10: label should also be grayed out when the checkbox is disabled, otherwise it's not really visible the checkbox is disabled (at least for my colorblind eyes). See [here](https://github.com/odoo/o-spreadsheet/pull/3536/files#diff-93c872bfb48cc2863dbb3f42e071e10f05eae47c29a78e769653da51874a4097)

src/helpers/figures/charts/chart_factory.ts
```diff
@@ -184,6 +187,18 @@ export function getSmartChartDefinition(zone: Zone, getters: Getters): ChartDefi
       legendPosition: newLegendPos,
     };
   }
+  const _dataSets = createDataSets(getters, dataSets, sheetId, dataSetsHaveTitle);
+  if (singleColumn && getData(getters, _dataSets[0]).every((e) => typeof e === "string")) {
```
hokolomopo, line 191: That's for the point 1) in the spec right ? It doesn't seems to work
image

src/helpers/figures/charts/pie_chart.ts
hokolomopo:
```suggestion
  config.options.plugins!["eventPlugin"] = {
    events: ["mouseout"],
  };
```

if you want. But that's also just a way to ignore the typing 🤷 

Maybe add a comment on what aa eventPlugin even does 😅

src/helpers/figures/charts/pie_chart.ts
```diff
@@ -216,25 +225,93 @@ function getPieConfiguration(
...
+let lastHoveredIndex: number | undefined = undefined;
+
+function highlightItem(index: number, dataSets) {
+  dataSets.forEach((dataset) => {
+    const backgroundColors = dataset.backgroundColor;
+    if (!backgroundColors) {
+      return;
+    }
+    backgroundColors.forEach((color, i, colors) => {
+      const hexColor = toHex(color.length === 9 ? color.slice(0, -2) : color);
```
hokolomopo, line 270: I'd add a helper `setColorAlpha` instead of doing strange slice things on color strigns

src/helpers/figures/charts/pie_chart.ts
```diff
@@ -216,25 +225,93 @@ function getPieConfiguration(
...
+function onLeaveLegend(evt: ChartEvent, item: LegendItem, legend: LegendElement<"pie">) {
+  const datasets = legend.chart.data.datasets;
+  unHighlightItems(datasets);
+  legend.chart.update();
+}
+
+function onHoverPieChart(evt: ChartEvent, items: ActiveElement[], chart: Chart) {
+  const datasets = chart.data.datasets;
+  if (items[0]) {
+    if (items[0].index !== lastHoveredIndex) {
```
hokolomopo, line 304: using the `lastHoveredIndex` that is global and shared across all charts sounds _very_ suspicious... Is it really necessary ?

hokolomopo (commented):
ChartJS sounds like black magic 🧙 

Do you think it would simplify things if instead of highlighting the hovered pie part by making every other part transparent, you would use "easier" chartJs features to make the hovered items darker/add a border/make in bigger ([this](https://www.chartjs.org/docs/latest/charts/doughnut.html#interactions) ) ? 

You'd still need to do magic when hovering the legend tho...

src/helpers/figures/charts/pie_chart.ts
anhe-odoo: It's not really necessary, but it will avoid doing useless stuff when we move the mouse on the same pie part. But I'll try to find another way to do it

src/components/side_panel/chart/building_blocks/label_range/label_range.ts
hokolomopo: Did he think about multi-dataset pie chart ? 😅 But maybe we do'nt care

src/helpers/figures/charts/chart_factory.ts
anhe-odoo: =this part is working as intended, else you wont have a pie chart in your case :) But it looks like the data is not correctly rendered

src/components/side_panel/chart/building_blocks/label_range/label_range.ts
anhe-odoo: At first he didn't know we could make multi-dataset pie chart, and after I told him he said to keep the title hidden anyway :D

src/helpers/figures/charts/chart_factory.ts
anhe-odoo: Ok, found the issue :) Looks like I always tested with three different values at the first 3 cells when plotting my chart :D

src/helpers/color.ts
```diff
@@ -319,3 +319,7 @@ export function isSameColor(color1: Color, color2: Color, tolerance: number = 0)
   );
   return diff <= tolerance;
 }
+
+export function changeColorAlpha(color: string, alpha: number): string {
```
hokolomopo, line 323:
```suggestion
export function changeColorAlpha(color: Color, alpha: number): string {
```

LucasLefevre (commented):
would it be possible to split the 3 points in 3 different commits ?
It's easier to review now and for the future as well in case we need to dig in the history

src/components/side_panel/chart/building_blocks/label_range/label_range.ts
```diff
@@ -5,19 +5,22 @@ import { SelectionInput } from "../../../../selection_input/selection_input";
...
+  name: string;
+  label: string;
+  value: boolean;
+  onChange: (value: boolean) => void;
+  disabled?: boolean;
```
rrahir, line 13: Don't think that's used. Is it relevant for another/future purpose?

rrahir: or at least, not tested ;-)

src/helpers/figures/charts/pie_chart.ts
```diff
@@ -197,6 +204,56 @@ export class PieChart extends AbstractChart {
     const definition = this.getDefinitionWithSpecificDataSets(dataSets, labelRange);
     return new PieChart(definition, this.sheetId, this.getters);
   }
+
+  highlightItem(index: number, dataSets) {
```
rrahir, line 208: can dataset be typed?

src/helpers/figures/charts/chart_ui_common.ts
```diff
@@ -231,9 +251,12 @@ export function getChartDatasetValues(getters: Getters, dataSets: DataSet[]): Da
           ? truncateLabel(cell.formattedValue)
           : (label = `${ChartTerms.Series} ${parseInt(dsIndex) + 1}`);
     } else {
-      label = label = `${ChartTerms.Series} ${parseInt(dsIndex) + 1}`;
+      label = `${ChartTerms.Series} ${parseInt(dsIndex) + 1}`;
     }
     let data = ds.dataRange ? getData(getters, ds) : [];
+    if (data.every((e) => typeof e === "string" && !e.startsWith("#"))) {
```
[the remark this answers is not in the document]
LucasLefevre: Not all errors starts with `#` (Loading error in odoo)
You want to use `isEvaluationError` :wink:

src/components/side_panel/chart/line_chart/line_chart_config_panel.ts
```diff
@@ -31,6 +31,7 @@ export class LineConfigPanel extends LineBarPieConfigPanel {
         value: (this.props.definition as LineChartDefinition).labelsAsText,
         label: _t("Treat labels as text"),
         onChange: this.onUpdateLabelsAsText.bind(this),
+        disabled: false,
```
LucasLefevre, line 34: remove this. It's optional

src/components/side_panel/chart/scatter_chart/scatter_chart_config_panel.ts
```diff
@@ -29,6 +29,7 @@ export class ScatterConfigPanel extends LineBarPieConfigPanel {
         value: (this.props.definition as LineChartDefinition).labelsAsText,
         label: _t("Treat labels as text"),
         onChange: this.onUpdateLabelsAsText.bind(this),
+        disabled: false,
```
LucasLefevre, line 32: remove this. It's optional

src/helpers/figures/charts/chart_factory.ts
```diff
@@ -118,7 +120,8 @@ export function getChartTypes(): Record<string, string> {
  */
 export function getSmartChartDefinition(zone: Zone, getters: Getters): ChartDefinition {
   let dataSetZone = zone;
-  if (zone.left !== zone.right) {
+  const singleColumn = zone.left === zone.right;
```
LucasLefevre, line 123: `zoneToDimension(zone).numberOfCols === 1`

src/helpers/figures/charts/chart_factory.ts
```diff
@@ -184,6 +187,18 @@ export function getSmartChartDefinition(zone: Zone, getters: Getters): ChartDefi
       legendPosition: newLegendPos,
     };
   }
+  const _dataSets = createDataSets(getters, dataSets, sheetId, dataSetsHaveTitle);
+  if (singleColumn && getData(getters, _dataSets[0]).every((e) => typeof e === "string")) {
+    return {
+      title,
```
LucasLefevre, line 193: "A" is the title even though it's part of the dataset ?
image

src/helpers/figures/charts/chart_factory.ts
```diff
@@ -184,6 +187,18 @@ export function getSmartChartDefinition(zone: Zone, getters: Getters): ChartDefi
       legendPosition: newLegendPos,
     };
   }
+  const _dataSets = createDataSets(getters, dataSets, sheetId, dataSetsHaveTitle);
```
LucasLefevre, line 190: The dataset is built with `dataSetsHaveTitle` being `true` even though it's forced to `false` below. Can't it be an issue?

src/helpers/figures/charts/chart_ui_common.ts
```diff
@@ -213,6 +220,19 @@ export function getChartDatasetFormat(getters: Getters, dataSets: DataSet[]): Fo
   return undefined;
 }
 
+function countStringOccurrences(data) {
```
LucasLefevre, line 223: type for `data`

LucasLefevre, line 223: A more natural API for this function would be more something like `{ [string]: [count] }` IMHO (not coupled to the shape of a chart dataset).
It would also be easy to use for charts with `Object.keys()` for labels and `Object.values()` for values (I dare hope the order of `.keys()` and `.values()` is consistent)

src/helpers/figures/charts/chart_ui_common.ts
```diff
@@ -213,6 +220,19 @@ export function getChartDatasetFormat(getters: Getters, dataSets: DataSet[]): Fo
...
+  const stringCounts = {};
+  for (let value of data) {
```
LucasLefevre, line 225: `const`

src/helpers/figures/charts/chart_ui_common.ts
```diff
@@ -231,9 +251,12 @@ export function getChartDatasetValues(getters: Getters, dataSets: DataSet[]): Da
           ? truncateLabel(cell.formattedValue)
           : (label = `${ChartTerms.Series} ${parseInt(dsIndex) + 1}`);
     } else {
-      label = label = `${ChartTerms.Series} ${parseInt(dsIndex) + 1}`;
+      label = `${ChartTerms.Series} ${parseInt(dsIndex) + 1}`;
     }
     let data = ds.dataRange ? getData(getters, ds) : [];
+    if (data.every((e) => typeof e === "string" && !e.startsWith("#"))) {
+      data = countStringOccurrences(data).values;
```
LucasLefevre, line 258: aggregating the values should be done in `aggregateDataForLabels`, shouldn't it?

LucasLefevre (commented):
Some comments/suggestions on the first commit :) 

The second commit looks like a lot of code for a limited value. I also don't really understand why the datasets should be highlighted for pie charts but not other charts.
I think I'd cancel this part (see message in the task chatter)

src/components/side_panel/chart/line_chart/line_chart_config_panel.ts
anhe-odoo: Typescript gives me an error if a remove it. But I'll try to find another way to fix this error then

anhe-odoo: Git it, I needed to add the typing on the super definition

src/helpers/figures/charts/chart_factory.ts
anhe-odoo: I don't think so, as the `fullZone` will be the same with `dataSetsHaveTitle` set to `false` or `true`. But if you prefer, I can recall `createDataSets` in the `if` below ?

src/helpers/figures/charts/chart_ui_common.ts
```diff
@@ -231,9 +238,12 @@ export function getChartDatasetValues(getters: Getters, dataSets: DataSet[]): Da
           ? truncateLabel(cell.formattedValue)
           : (label = `${ChartTerms.Series} ${parseInt(dsIndex) + 1}`);
     } else {
-      label = label = `${ChartTerms.Series} ${parseInt(dsIndex) + 1}`;
+      label = `${ChartTerms.Series} ${parseInt(dsIndex) + 1}`;
     }
     let data = ds.dataRange ? getData(getters, ds) : [];
+    if (data.every((e) => typeof e === "string" && !isEvaluationError(e))) {
+      data.fill(1);
```
LucasLefevre, line 245: deserves a comment to explain why we fill the data with ones

src/helpers/figures/charts/chart_ui_common.ts
```diff
@@ -176,6 +177,12 @@ export function getChartLabelValues(
   labelRange?: Range
 ): LabelValues {
   let labels: LabelValues = { values: [], formattedValues: [] };
+  if (dataSets.length === 1) {
+    const data = getData(getters, dataSets[0]);
+    if (data.length && data.every((e) => typeof e === "string" && !isEvaluationError(e))) {
+      return { values: data, formattedValues: data };
+    }
+  }
```
LucasLefevre, line 185: I don't think this is needed anymore. At least no test fail if I remove this locally.

src/helpers/figures/charts/chart_factory.ts
LucasLefevre, line 191: why checking it's all string here but down below also check that it's not an error?

src/components/side_panel/chart/line_bar_pie_panel/config_panel.ts
```diff
@@ -86,13 +92,25 @@ export class LineBarPieConfigPanel extends Component<Props, SpreadsheetChildEnv>
...
+    let disabled = false;
+    if (this.isLabelInvalid || this.isDatasetInvalid) {
+      disabled = true;
+    } else if (this.labelRange && this.dataSeriesRanges.length) {
+      const zone = toUnboundedZone(this.dataSeriesRanges[0]);
+      disabled =
+        zone.left !== undefined &&
+        zone.right !== undefined &&
+        zone.left === zone.right &&
+        this.labelRange === this.dataSeriesRanges[0];
```
LucasLefevre, line 105: actually, I'm not sure why we should prevent the user to uncheck the aggregate.

LucasLefevre (approved):
robodoo r+

PR #3564, opened 2024-01-29
Master formulas lookup allow partial match laa

laa-odoo:
## Description:

    This commit introduces wildcard search
    functionality to the XLOOKUP formula,
    enhancing its search capabilities.
    
    In addition, it modifies the default search
    behavior for specific formulas - HLOOKUP,
    VLOOKUP, and MATCH - replacing strict search
    with wildcard search.
    
    The implemented wildcard search operates
    similarly to regular expression search,
    but only supports the generic "*" and "?"
    symbols.
    

Task: : [3703048](https://www.odoo.com/web#id=3703048&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/functions/helpers.ts
+++ b/src/functions/helpers.ts
@@ -479,10 +479,9 @@ type Operator = ">" | ">=" | "<" | "<=" | "<>" | "=";
 interface Predicate {
   operator: Operator;
   operand: number | string | boolean;
-  regexp?: RegExp;
 }
 
-function getPredicate(descr: string, isQuery: boolean, locale: Locale): Predicate {
+function getPredicate(descr: string, locale: Locale): Predicate {
   let operator: Operator;
   let operand: Maybe<CellValue>;
 
@@ -508,19 +507,18 @@ function getPredicate(descr: string, isQuery: boolean, locale: Locale): Predicat
     operand = toBoolean(operand);
   }
 
-  const result: Predicate = { operator, operand };
-
-  if (typeof operand === "string") {
-    if (isQuery) {
-      operand += "*";
-    }
-    result.regexp = operandToRegExp(operand);
-  }
-
-  return result;
+  return { operator, operand };
 }
 
-function operandToRegExp(operand: string): RegExp {
+/**
+ * Converts a search string containing wildcard characters to a regular expression.
+ *
+ * The function iterates over each character in the input string. If the character is a wildcard
+ * character ("?" or "*") and is not preceded by a "~", it is replaced by the corresponding regular
+ * expression.
+ * If the character is a special regular expression character, it is escaped with "\\".
+ */
+const wildcardToRegExp = memoize(function wildcardToRegExp(operand: string): RegExp {
   let exp = "";
   let predecessor = "";
   for (let char of operand) {
@@ -541,7 +539,7 @@ function operandToRegExp(operand: string): RegExp {
     predecessor = char;
   }
   return new RegExp("^" + exp + "$", "i");
-}
+});
 
 function evaluatePredicate(value: CellValue | undefined, criterion: Predicate): boolean {
   const { operator, operand } = criterion;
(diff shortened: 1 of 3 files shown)


src/functions/helpers.ts
```diff
@@ -810,10 +804,17 @@ export function linearSearch<T>(
...
   for (let i = 0; i < numberOfValues; i++) {
     const value = normalizeValue(getValue(data, i));
+
     if (value === _target) {
       return reverseSearch ? numberOfValues - i - 1 : i;
     }
-    if (mode === "nextSmaller") {
+    if (mode === "wildcard") {
+      if (typeof value === "string" && typeof _target === "string") {
+        if (operandToRegExp(_target).test(value)) {
```
LucasLefevre, line 813: creating a regexp can be quite slow (the regexp string is compiled to actual code).
Can it be created outside the for loop ?

src/functions/module_lookup.ts
```diff
@@ -528,10 +531,18 @@ export const XLOOKUP = {
       () => [-1, 1, -2, 2].includes(_searchMode),
       _t("searchMode should be a value in [-1, 1, -2, 2].")
     );
-    assert(() => [-1, 0, 1].includes(_matchMode), _t("matchMode should be a value in [-1, 0, 1]."));
+    assert(
+      () => [-1, 0, 1, 2].includes(_matchMode),
+      _t("matchMode should be a value in [-1, 0, 1, 2].")
```
LucasLefevre, line 536: the argument is named `match_mode` for the user
```suggestion
      _t("match_mode should be a value in [-1, 0, 1, 2].")
```

src/functions/module_lookup.ts
```diff
@@ -547,15 +558,22 @@ export const XLOOKUP = {
...
 
-    const mode = _matchMode === 0 ? "strict" : _matchMode === 1 ? "nextGreater" : "nextSmaller";
+    const mode =
+      _matchMode === 0
+        ? "strict"
+        : _matchMode === 1
+        ? "nextGreater"
+        : _matchMode === -1
+        ? "nextSmaller"
+        : "wildcard";
```
LucasLefevre, line 568: can we write this differently with a simple object mapping ? It would be easier to read I think
```ts
const MATCH_MODE = {
  0: "strict",
  1: "nextGreater",
  ...
}
```

src/functions/helpers.ts
LucasLefevre: Other performance consideration: if it is easy to detect if `_target` contains regexp like characters and, if not, avoid using the regexp all together ?

src/functions/helpers.ts
```diff
@@ -508,16 +507,7 @@ function getPredicate(descr: string, isQuery: boolean, locale: Locale): Predicat
...
-      operand += "*";
-    }
-    result.regexp = operandToRegExp(operand);
-  }
-
-  return result;
+  return { operator, operand };
 }
 
 function operandToRegExp(operand: string): RegExp {
```
LucasLefevre, line 513: It's probably just because I don't know this code:
- can we have a small docstring to explain this, with some examples ?
- the word `operand` looks misleading to me here. To me it would be `a` or `b` in `a+b`. But as I understand it, it's more like an "expression" or a "regexp-like" string. Am I right?

LucasLefevre (commented):
LGTM :)
as VLOOKUP is used a lot, and on large ranges, i'd try to make it as fast and possible. Specifically, not add overhead if the target is not wildcard-like

src/functions/helpers.ts
laa-odoo: my bad

laa-odoo:
@LucasLefevre, review was taken into account in a commit

laa-odoo:
@LucasLefevre `memoize` was placed on the `wildcardToRegExp` function

LucasLefevre (approved):
robodoo r+

PR #3573, opened 2024-01-30
[FIX] Selection: Loop selection based on content

rrahir:
Similarly to the issue addressed in PR #3214, we want `loopSelection` to consider seemingly empty cells when considering the expansion zone.

This revision splits the behaviour of the selection processor for `loopSelection` and the other navigation features (like cluster jumping) as we still want to ignore cells with an empty evaluation in those cases. The reason being that the user could be lost if, when navigating with their keyboards, they end up with their selection on empty cell.

As for #3214, the fix will need to be adapted to account for the spreaded formulas.

Task: 3709340

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


rrahir:
@fw-bot up to saas-16.3

PR #3579, opened 2024-01-31
[REF] clipboard: Rename handlers files

anhe-odoo:
## Description:

A simple commit to change the name of the clipboard handlers' files to avoid having once again a 'cell.ts', 'borders.ts', ... files


hokolomopo (commented):
@robodoo r+

PR #3599, opened 2024-02-02
[REF] SidePanel: allow to open from external code

LucasLefevre:
## Description:

This commit moves the side panel state from the `Spreadsheet` component to a dedicated store.

It simplifies the `Spreadsheet` component, but more importantly, it allows to
open the side panel from outside the `Spreadsheet` component
(=from a parent component).

We've done it once [^1] in a monstrous way, with the tools we had at the time. Let's get rid of that!

[^1]: 🙈 https://github.com/odoo/enterprise/blob/c7090769173bdc1940485365178c7738ff2a6518/spreadsheet_edition/static/src/bundle/actions/version_history/version_history_action.js#L75

Task: : [/](https://www.odoo.com/web#id=TASK_ID&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/grid/grid.ts
+++ b/src/components/grid/grid.ts
@@ -2,6 +2,7 @@ import {
   Component,
   onMounted,
   useChildSubEnv,
+  useEffect,
   useExternalListener,
   useRef,
   useState,
@@ -71,6 +72,7 @@ import { Menu, MenuState } from "../menu/menu";
 import { CellPopoverStore } from "../popover";
 import { Popover } from "../popover/popover";
 import { HorizontalScrollBar, VerticalScrollBar } from "../scrollbar/";
+import { SidePanelStore } from "../side_panel/side_panel/side_panel_store";
 import { HoveredCellStore } from "./hovered_cell_store";
 
 /**
@@ -101,7 +103,6 @@ const registries = {
 };
 
 interface Props {
-  sidePanelIsOpen: boolean;
   exposeFocus: (focus: () => void) => void;
 }
 
@@ -111,7 +112,6 @@ interface Props {
 export class Grid extends Component<Props, SpreadsheetChildEnv> {
   static template = "o-spreadsheet-Grid";
   static props = {
-    sidePanelIsOpen: Boolean,
     exposeFocus: Function,
   };
   static components = {
@@ -141,6 +141,7 @@ export class Grid extends Component<Props, SpreadsheetChildEnv> {
   onMouseWheel!: (ev: WheelEvent) => void;
   canvasPosition!: DOMCoordinates;
   hoveredCell!: Store<HoveredCellStore>;
+  sidePanel!: Store<SidePanelStore>;
 
   setup() {
     this.highlightStore = useStore(HighlightStore);
@@ -155,6 +156,7 @@ export class Grid extends Component<Props, SpreadsheetChildEnv> {
     this.composerStore = useStore(ComposerStore);
     this.composerFocusStore = useStore(ComposerFocusStore);
     this.DOMFocusableElementStore = useStore(DOMFocusableElementStore);
+    this.sidePanel = useStore(SidePanelStore);
 
     useChildSubEnv({ getPopoverContainerRect: () => this.getGridRect() });
     useExternalListener(document.body, "cut", this.copy.bind(this, true));
@@ -170,6 +172,15 @@ export class Grid extends Component<Props, SpreadsheetChildEnv> {
       this.hoveredCell.clear();
     });
     this.cellPopovers = useStore(CellPopoverStore);
+
+    useEffect(
+      () => {
+        if (!this.sidePanel.isOpen) {
+          this.DOMFocusableElementStore.focus();
+        }
+      },
+      () => [this.sidePanel.isOpen]
+    );
   }
 
   onCellHovered({ col, row }) {
--- a/src/components/side_panel/chart/main_chart_panel/main_chart_panel_store.ts
+++ b/src/components/side_panel/chart/main_chart_panel/main_chart_panel_store.ts
@@ -1,0 +1,36 @@
+import { Get } from "../../../../store_engine";
+import { SpreadsheetStore } from "../../../../stores";
+import { Command, UID } from "../../../../types";
+import { SidePanelStore } from "../../side_panel/side_panel_store";
+
+export class MainChartPanelStore extends SpreadsheetStore {
+  private sidePanel = this.get(SidePanelStore);
+
+  figureId: UID | null;
+  panel: "configuration" | "design" = "configuration";
+
+  constructor(get: Get, figureId: UID | null) {
+    super(get);
+    this.figureId = figureId;
+  }
+
+  protected handle(cmd: Command): void {
+    switch (cmd.type) {
+      case "DELETE_FIGURE":
+        if (this.sidePanel.componentTag === "ChartPanel" && this.figureId === cmd.id) {
+          this.sidePanel.close();
+        }
+        break;
+      case "SELECT_FIGURE":
+        if (cmd.id) {
+          this.figureId = cmd.id;
+        } else if (this.sidePanel.componentTag === "ChartPanel") {
+          this.sidePanel.close();
+        }
+    }
+  }
+
+  activatePanel(panel: "configuration" | "design") {
+    this.panel = panel;
+  }
+}
--- a/src/components/side_panel/side_panel/side_panel.ts
+++ b/src/components/side_panel/side_panel/side_panel.ts
@@ -1,8 +1,10 @@
-import { Component, onWillUpdateProps, useState } from "@odoo/owl";
+import { Component } from "@odoo/owl";
 import { BACKGROUND_HEADER_COLOR, FILTERS_COLOR } from "../../../constants";
-import { SidePanelContent, sidePanelRegistry } from "../../../registries/side_panel_registry";
+import { sidePanelRegistry } from "../../../registries/side_panel_registry";
+import { Store, useStore } from "../../../store_engine";
 import { SpreadsheetChildEnv } from "../../../types";
 import { css } from "../../helpers/css";
+import { SidePanelStore } from "./side_panel_store";
 
 css/* scss */ `
   .o-sidePanel {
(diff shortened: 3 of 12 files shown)


src/components/grid/grid.ts
```diff
@@ -170,6 +172,15 @@ export class Grid extends Component<Props, SpreadsheetChildEnv> {
       this.hoveredCell.clear();
     });
     this.cellPopovers = useStore(CellPopoverStore);
+
+    useEffect(
+      () => {
+        if (!this.sidePanel.isOpen) {
+          this.focus();
```
LucasLefevre, line 179: there's only one use of `focusGrid` in `Spreadsheet` component. Getting rid of this last use is my next step.

src/components/side_panel/side_panel/side_panel.ts
```diff
@@ -152,37 +154,25 @@ css/* scss */ `
...
-      panel: sidePanelRegistry.get(this.props.component),
-    });
-    onWillUpdateProps(
-      (nextProps: Props) => (this.state.panel = sidePanelRegistry.get(nextProps.component))
-    );
+    this.sidePanelStore = useStore(SidePanelStore);
+  }
+
+  get panel() {
+    return sidePanelRegistry.get(this.sidePanelStore.component);
```
hokolomopo, line 167: Make the store expose directly the component maybe ? No reason for the SidePanelComponent to worry about a registry if we want to make it dumber

src/components/side_panel/side_panel/side_panel.ts
```diff
@@ -152,37 +154,25 @@ css/* scss */ `
...
+  close() {
+    this.sidePanelStore.close();
   }
 
   getTitle() {
-    return typeof this.state.panel.title === "function"
-      ? this.state.panel.title(this.env)
-      : this.state.panel.title;
+    const panel = this.panel;
+    return typeof panel.title === "function" ? panel.title(this.env) : panel.title;
```
hokolomopo, line 176: same here, this coule be done in the store

src/components/side_panel/side_panel/side_panel.xml
```diff
@@ -3,21 +3,21 @@
     <div class="o-sidePanel">
       <div class="o-sidePanelHeader">
         <div class="o-sidePanelTitle" t-esc="getTitle()"/>
-        <div class="o-sidePanelClose" t-on-click="() => this.props.onCloseSidePanel()">✕</div>
+        <div class="o-sidePanelClose" t-on-click="close">✕</div>
```
hokolomopo, line 6:
```suggestion
        <div class="o-sidePanelClose" t-on-click="sidePanelStore.close">✕</div>
```
nitpick

src/components/side_panel/side_panel/side_panel.ts
hokolomopo: But the `sidePanelStore.component` as a string is still needed in the `t-key` of the templates ... Are the `t-key` still necessary ? it should be reactive now, no ?

src/components/side_panel/side_panel/side_panel.xml
```diff
@@ -3,21 +3,21 @@
...
-          onCloseSidePanel="props.onCloseSidePanel"
-          t-key="'Body_' + props.component"
+          t-component="panel.Body"
+          t-props="sidePanelStore.props"
+          onCloseSidePanel.bind="close"
+          t-key="'Body_' + sidePanelStore.component"
         />
       </div>
-      <div class="o-sidePanelFooter" t-if="state.panel.Footer">
+      <div class="o-sidePanelFooter" t-if="panel.Footer">
```
hokolomopo, line 16: Not related to the store, but that's the first time I realize we can define a Footer for the side panel 😅 . What's it for ? We never use it. And I have a hard time seeing where that'd be useful ? Probably some legacy stuff that we can remove, no?

src/components/side_panel/side_panel/side_panel_store.ts
```diff
@@ -0,0 +1,36 @@
...
+  isOpen: boolean = false;
+  props: SidePanelProps = {};
+  component: string = "";
+
+  open(panel: string, panelProps: SidePanelProps = {}) {
+    if (this.isOpen && panel !== this.component) {
+      this.props?.onCloseSidePanel?.();
+    }
+    this.component = panel;
+    this.props = panelProps;
```
hokolomopo, line 18: `this.props` make it look like it's a component, not a store 😅 

nitpick: Also `this.component` being a string sounds strange

src/components/side_panel/side_panel/side_panel_store.ts
```diff
@@ -0,0 +1,36 @@
+import { SpreadsheetStore } from "../../../stores";
+
+interface SidePanelProps {
+  onCloseSidePanel?: () => void;
+  [key: string]: unknown;
```
hokolomopo, line 5: random question: what's the difference between putting `unknown` vs `any` here ?

src/components/spreadsheet/spreadsheet.ts
```diff
@@ -293,8 +287,8 @@ export class Spreadsheet extends Component<SpreadsheetProps, SpreadsheetChildEnv
       loadCurrencies: this.model.config.external.loadCurrencies,
       loadLocales: this.model.config.external.loadLocales,
       isDashboard: () => this.model.getters.isDashboard(),
-      openSidePanel: this.openSidePanel.bind(this),
-      toggleSidePanel: this.toggleSidePanel.bind(this),
+      openSidePanel: this.sidePanel.open.bind(this.sidePanel),
+      toggleSidePanel: this.sidePanel.toggle.bind(this.sidePanel),
```
hokolomopo, line 291: I don't think those _need_ to be in the env now. The question is: do we want to keep them ? Bloating the env VS simpler syntax to open a side panel

src/components/spreadsheet/spreadsheet.xml
```diff
@@ -23,18 +23,10 @@
...
+            <Grid exposeFocus="(focus) => this._focusGrid = focus"/>
           </div>
         </div>
-        <SidePanel
-          t-if="sidePanel.isOpen"
-          onCloseSidePanel="() => this.closeSidePanel()"
-          component="sidePanel.component"
-          panelProps="sidePanel.panelProps"
-        />
+        <SidePanel t-if="sidePanel.isOpen"/>
```
hokolomopo, line 29: Maybe better to have the t-if in the component directly ?

hokolomopo (commented):
Byebye `sidePanelIsOpen` 🥳

hokolomopo:
Also the runbot is RED 🟥

src/components/grid/grid.ts
rrahir: Pretty sure it's not introduces by this PR bus since some side panels (pivot properties/list/ chart) close themselves up following a collaborative command:  If I'm currently editing a link for instance - the sidepanel closing will make me lose focus of my input box :'(

src/components/side_panel/side_panel/side_panel.xml
rrahir: Has it ever been used ^^I think the idea was to have a common structure with the Odoo dialog but we never properly used it, for instance for the Cancel/ confirm buttons

src/components/spreadsheet/spreadsheet.xml
rrahir: you'd end up with an empty div that might be subject to css styling, no?

src/components/side_panel/side_panel/side_panel_store.ts
LucasLefevre: with `any`, typescript let's you do anything. `this.props.foo.bar` would compile. With `unknown`, typescript will prevent everything unless you explicitly check what it is before (you'd have to check `"foo" in this.props` before accessing it).

Most of the time, you want to use `unknown`

src/components/spreadsheet/spreadsheet.ts
rrahir: we should then think about  how we'ed like to open a side panel from an action?

src/components/spreadsheet/spreadsheet.xml
hokolomopo: Probably possible to put the t-if on the top-most div, and nothing is displayed then ? Not sure.

src/components/side_panel/side_panel/side_panel.xml
LucasLefevre: it would need to be binded `sidePanelStore.close.bind(sidePanelStore)` :/

src/components/spreadsheet/spreadsheet.xml
LucasLefevre: Yes indeed. But that would mean the `SidePanel` component would still be created. Not critical, but since we already need the store, this t-if doesn't harm

src/components/side_panel/chart/main_chart_panel/main_chart_panel_store.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  constructor(get: Get, figureId: UID | null) {
+    super(get);
+    this.figureId = figureId;
+  }
+
+  protected handle(cmd: Command): void {
+    switch (cmd.type) {
+      case "DELETE_FIGURE":
+        if (this.sidePanel.componentTag === "ChartPanel") {
```
hokolomopo, line 20: Check if the figure deleted was the one selected ?

src/components/side_panel/chart/main_chart_panel/main_chart_panel_store.ts
```diff
@@ -0,0 +1,36 @@
...
+          this.sidePanel.close();
+        }
+        break;
+      case "SELECT_FIGURE":
+        if (cmd.id) {
+          this.figureId = cmd.id;
+        } else if (this.sidePanel.componentTag === "ChartPanel") {
```
hokolomopo, line 27: if this really happens, the typing of SELECT_FIGURE is wrong. SelectFigureCommand.id should always be defined

hokolomopo (commented):
I just came here from your comment on my table style PR, didn't read much

src/components/side_panel/chart/main_chart_panel/main_chart_panel_store.ts
LucasLefevre: It's wrong indeed :+1: good catch, thanks

src/components/side_panel/side_panel/side_panel.ts
LucasLefevre: the store doesn't have access to the env (and should not). Funnily enough, I don't think we ever use the dynamic title. It could probably be removed

src/components/side_panel/side_panel/side_panel.ts
LucasLefevre: I agree in principle. But it creates a circular dependency. We'd have to re-organize quite a bit of things

rrahir:
@robodoo r+

rrahir:
@robodoo r+

src/components/side_panel/chart/main_chart_panel/main_chart_panel_store.ts
rrahir: that line probably belongs to the release of selection plugin actually ...

PR #3609, opened 2024-02-05
[FIX] Find and replace: draw search results

hokolomopo:
The search results were not drawn since transforming the f&r plugin with a store (we defined a `draw` instead of a `drawLayer` method).

Took the opportunity to draw to use the `HighlightStore` to draw the search results instead of doing it manually.

## Description:

Task: : [3721321](https://www.odoo.com/web#id=3721321&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/find_and_replace/find_and_replace_store.ts
+++ b/src/components/side_panel/find_and_replace/find_and_replace_store.ts
@@ -1,11 +1,13 @@
 import { debounce, getSearchRegex, isInside, positionToZone } from "../../../helpers";
+import { HighlightProvider, HighlightStore } from "../../../stores/highlight_store";
+import { CellPosition, Color, Command, Highlight } from "../../../types";
+
+import { toRaw } from "@odoo/owl";
 import { Get } from "../../../store_engine";
 import { SpreadsheetStore } from "../../../stores";
-import { CellPosition, Color, Command, GridRenderingContext } from "../../../types";
 import { SearchOptions } from "../../../types/find_and_replace";
 
-const BORDER_COLOR: Color = "#8B008B";
-const BACKGROUND_COLOR: Color = "#8B008B33";
+const FIND_AND_REPLACE_HIGHLIGHT_COLOR: Color = "#8B008B";
 
 enum Direction {
   previous = -1,
@@ -13,7 +15,7 @@ enum Direction {
   next = 1,
 }
 
-export class FindAndReplaceStore extends SpreadsheetStore {
+export class FindAndReplaceStore extends SpreadsheetStore implements HighlightProvider {
   private allSheetsMatches: CellPosition[] = [];
   private activeSheetMatches: CellPosition[] = [];
   private specificRangeMatches: CellPosition[] = [];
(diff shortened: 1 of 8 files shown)


src/components/side_panel/find_and_replace/find_and_replace_store.ts
```diff
@@ -318,26 +315,31 @@ export class FindAndReplaceStore extends SpreadsheetStore {
...
-        ctx.fillStyle = BACKGROUND_COLOR;
-        ctx.fillRect(x, y, width, height);
-        if (index === this.selectedMatchIndex) {
-          ctx.strokeStyle = BORDER_COLOR;
-          ctx.strokeRect(x, y, width, height);
-        }
+        highlights.push({
+          sheetId,
+          zone: zoneWithMerge,
+          color: BORDER_COLOR,
```
LucasLefevre, line 323: BACKGROUND_COLOR changed to BACKGROUND_COLOR ? Was the change done on purpose ?

src/helpers/rendering.ts
```diff
@@ -11,11 +11,19 @@ export function drawHighlight(
...
+      ctx.lineWidth = 1;
+      ctx.strokeRect(x, y, width, height);
+    } else {
+      ctx.lineWidth = 2;
+      /** + 0.5 offset to have sharp lines. See comment in {@link RendererPlugin#drawBorder} for more details */
+      ctx.strokeRect(x + 0.5, y + 0.5, width, height);
+    }
+  }
   if (!highlight.noFill) {
+    ctx.globalCompositeOperation = "source-over";
```
LucasLefevre, line 26: what does this do ?

hokolomopo: Nothing, that's the default. It says "draw over exisiting things in canvas".

I moved the line, but we can remove it altogether

src/components/side_panel/find_and_replace/find_and_replace_store.ts
hokolomopo: BACKGROUND_COLOR was the same as BORDER_COLOR + an alpha. Since the alpha is set inside the highlight, there's no need for both colors to exist

LucasLefevre: The variable could be renamed then.
Also, the background is different before/after (the alpha is different I guess). We could use the `fillAlpha` you introduce in https://github.com/odoo/o-spreadsheet/pull/3639

src/components/side_panel/find_and_replace/find_and_replace_store.ts
```diff
@@ -46,10 +47,7 @@ export class FindAndReplaceStore extends SpreadsheetStore {
       this.model.dispatch("SET_FORMULA_VISIBILITY", { show: this.initialShowFormulaState });
       this.updateSearchContent.stopDebounce();
     });
-  }
-
-  get renderingLayers() {
-    return ["Search"] as const;
+    get(HighlightStore).register(this);
```
LucasLefevre, line 50: We should unregister to avoid memory leak. It may not be necessary here because the store is never used as a local store, but it's a good practice nonetheless to follow, in case this code is taken as an example and copy-pasted

LucasLefevre (approved):
robodoo r+

PR #3624, opened 2024-02-06
[IMP] tables: introduce dynamic tables

hokolomopo:
## [IMP] tables: introduce dynamic tables

This commit introduces the concept of dynamic table. A dynamic table
is a table that is bound to a array formula, instead of a static range.
The table will grow and shrink along the results of the array formula.

The notion is transparent to the user: if the users defines a table on
a single cell that contains an array formula, or all the cells of the
array formula, the table will be dynamic.

This means that the table zones are now an UI concept. The core plugin
now manages `InternalTable` objects, that can be either static or
dynamic (in which case they don't have a zone, but a single cell position).
The UI plugin `DynamicTablesPlugin` transform those `InternalTable` into
`Table` objects that every other plugin and component can easily use.

As the core plugin doesn't have access to the evaluation results, and
can't know if a cell contains an array formula, it's the responsability
of the components that create a table to determine if it's dynamic or not.

## [IMP] array formula: `getSpreadPositionsOf` now returns root position

Before this the getter `getSpreadPositionsOf` only returned the
position of the cells an array formula was spread to, minus the
cell the array formula was in. This made it impossible to differentiate
between a cell with a normal formula, and a cell with an array formula
returning a 1x1 matrix.

Now the getter returns the root position of the array formula as well.

There was also a bug where the spread relation weren't properly cleaned.

Task:  [3707037](https://www.odoo.com/web#id=3707037&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/ui/table_interactive.ts
+++ b/src/helpers/ui/table_interactive.ts
@@ -1,5 +1,6 @@
-import { TableTerms } from "../../components/translations_terms";
 import { CommandResult, DispatchResult, SpreadsheetChildEnv, UID } from "../../types";
+
+import { TableTerms } from "../../components/translations_terms";
 import { TableConfig } from "../../types/table";
 import { getZoneArea } from "../zones";
 
@@ -12,14 +13,22 @@ export function interactiveCreateTable(
   sheetId: UID,
   tableConfig?: TableConfig
 ): DispatchResult {
-  const selection = env.model.getters.getSelectedZones();
-  if (selection.length === 1 && getZoneArea(selection[0]) === 1) {
-    env.model.selection.selectTableAroundSelection();
-  }
-  const target = env.model.getters.getSelectedZones();
-  const ranges = target.map((zone) => env.model.getters.getRangeDataFromZone(sheetId, zone));
+  let target = env.model.getters.getSelectedZones();
+  let isDynamic = env.model.getters.canCreateDynamicTableOnZones(sheetId, target);
 
-  const result = env.model.dispatch("CREATE_TABLE", { ranges, sheetId, config: tableConfig });
+  if (target.length === 1 && !isDynamic && getZoneArea(target[0]) === 1) {
+    env.model.selection.selectTableAroundSelection();
+    target = env.model.getters.getSelectedZones();
+    isDynamic = env.model.getters.canCreateDynamicTableOnZones(sheetId, target);
+  }
+
+  const ranges = target.map((zone) => env.model.getters.getRangeDataFromZone(sheetId, zone));
+  const result = env.model.dispatch("CREATE_TABLE", {
+    ranges,
+    sheetId,
+    config: tableConfig,
+    tableType: isDynamic ? "dynamic" : "static",
+  });
   if (result.isCancelledBecause(CommandResult.TableOverlap)) {
     env.raiseError(TableTerms.Errors.TableOverlap);
   } else if (result.isCancelledBecause(CommandResult.NonContinuousTargets)) {
--- a/src/plugins/core/tables.ts
+++ b/src/plugins/core/tables.ts
@@ -8,22 +8,26 @@ import {
   isZoneInside,
   overlap,
   positions,
+  positionToZone,
   range,
   zoneToDimension,
+  zoneToTopLeft,
   zoneToXc,
 } from "../../helpers";
-import { getTableContentZone } from "../../helpers/table_helpers";
+import { createFilter } from "../../helpers/table_helpers";
 import { DEFAULT_TABLE_CONFIG, TABLE_PRESETS } from "../../helpers/table_presets";
 import {
   ApplyRangeChange,
   CellPosition,
   CommandResult,
   CoreCommand,
+  CoreTable,
+  CoreTableType,
+  DynamicTable,
   ExcelWorkbookData,
   Filter,
-  FilterId,
-  Position,
   Range,
+  StaticTable,
   Table,
   TableConfig,
   TableData,
@@ -34,31 +38,22 @@ import {
   WorkbookData,
   Zone,
 } from "../../types/index";
+
 import { CorePlugin } from "../core_plugin";
 
 interface TableState {
-  tables: Record<UID, Record<TableId, Table | undefined>>;
+  tables: Record<UID, Record<TableId, CoreTable | undefined>>;
 }
 
 export class TablePlugin extends CorePlugin<TableState> implements TableState {
-  static getters = [
-    "getFilter",
-    "getFilters",
-    "getTable",
-    "getTables",
-    "getTablesInZone",
-    "getTablesOverlappingZones",
-    "getFilterId",
-    "getFilterHeaders",
-    "isFilterHeader",
-  ] as const;
+  static getters = ["getCoreTable", "getCoreTables"] as const;
 
-  readonly tables: Record<UID, Record<TableId, Table | undefined>> = {};
+  readonly tables: Record<UID, Record<TableId, CoreTable | undefined>> = {};
 
   adaptRanges(applyChange: ApplyRangeChange, sheetId?: UID) {
     const sheetIds = sheetId ? [sheetId] : this.getters.getSheetIds();
     for (const sheetId of sheetIds) {
-      for (const table of this.getTables(sheetId)) {
+      for (const table of this.getCoreTables(sheetId)) {
         this.applyRangeChangeOnTable(sheetId, table, applyChange);
       }
     }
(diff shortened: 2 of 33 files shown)


anhe-odoo (commented):
Some minor comments :)

src/helpers/ui/table_interactive.ts
```diff
@@ -13,13 +14,21 @@ export function interactiveCreateTable(
...
   const selection = env.model.getters.getSelectedZones();
-  if (selection.length === 1 && getZoneArea(selection[0]) === 1) {
+  let target = env.model.getters.getSelectedZones();
+  const isDynamic = env.model.getters.canCreateDynamicTableOnZones(sheetId, selection);
+
+  if (selection.length === 1 && !isDynamic && getZoneArea(selection[0]) === 1) {
     env.model.selection.selectTableAroundSelection();
+    target = env.model.getters.getSelectedZones();
   }
-  const target = env.model.getters.getSelectedZones();
```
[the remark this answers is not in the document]
hokolomopo: 'cause I meant to use it in the `const isDynamic = ...` line instead of calling `env.model.getters.getSelectedZones()` twice :sweat:

src/plugins/core/tables.ts
```diff
@@ -82,19 +76,17 @@ export class TablePlugin extends CorePlugin<TableState> implements TableState {
...
           return CommandResult.TableNotFound;
         }
         return this.checkValidations(cmd, this.checkUpdatedTableZoneIsValid, (cmd) =>
           this.checkTableConfigUpdateIsValid(cmd.config)
         );
       case "ADD_MERGE":
-        for (const merge of cmd.target) {
-          for (const table of this.getTables(cmd.sheetId)) {
-            if (overlap(table.range.zone, merge)) {
+        for (let merge of cmd.target) {
```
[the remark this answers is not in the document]
hokolomopo: Rebase gone wrong for the let :smile:

src/plugins/core/tables.ts
```diff
@@ -157,14 +152,31 @@ export class TablePlugin extends CorePlugin<TableState> implements TableState {
...
+              : this.updateStaticTable(table, range, newConfig);
+            this.history.update("tables", cmd.sheetId, table.id, newTable);
+          }
         }
         break;
       }
       case "UPDATE_CELL": {
         const sheetId = cmd.sheetId;
-        for (const table of this.getTables(sheetId)) {
+        for (let table of this.getInternalTables(sheetId)) {
```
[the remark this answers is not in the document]
hokolomopo: rebase gone wrong :upside_down_face:

src/plugins/core/tables.ts
```diff
@@ -175,10 +187,10 @@ export class TablePlugin extends CorePlugin<TableState> implements TableState {
         break;
       }
       case "DELETE_CONTENT": {
-        const tables: Record<TableId, Table | undefined> = { ...this.tables[cmd.sheetId] };
+        const tables: Record<TableId, InternalTable | undefined> = { ...this.tables[cmd.sheetId] };
         for (const tableId in tables) {
```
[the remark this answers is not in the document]
anhe-odoo: Ok, I didn't know about this :)

rrahir:
THe first commit definitely deserves to be backported

src/plugins/core/tables.ts
```diff
@@ -130,24 +125,25 @@ export class TablePlugin extends CorePlugin<TableState> implements TableState {
...
+          : this.createStaticTable(id, union, config);
         this.history.update("tables", cmd.sheetId, newTable.id, newTable);
         break;
       }
       case "REMOVE_TABLE": {
-        const tables: Record<UID, Table> = {};
-        for (const table of this.getTables(cmd.sheetId)) {
-          if (cmd.target.every((zone) => !intersection(zone, table.range.zone))) {
+        const tables: Record<UID, InternalTable> = {};
+        for (const table of this.getInternalTables(cmd.sheetId)) {
```
rrahir, line 137: It was already present before but this command can technically delete several tables at a time. should the command have been targeting the exact table zone definition all along?

hokolomopo: Probably could have been a good way to do it, but at the time it would have required a bit more logic in the menu items for no real reason.

Now I'm stuck with it, I don't see how I would migrate the revisions correctly in python.

LucasLefevre (commented):
a few comments/suggestions :)

tests/evaluation/evaluation_formula_array.test.ts
```diff
@@ -740,4 +743,38 @@ describe("evaluate formulas that return an array", () => {
...
+      ]);
+    });
+
+    test("getSpreadPositionsOf is correct after the evaluation changed so the formula can spread again", () => {
+      setCellContent(model, "H1", "5");
+      setCellContent(model, "A1", "=MFILL(H1,H1,42)");
+
+      expect(model.getters.getSpreadPositionsOf({ sheetId, col: 0, row: 0 })).toHaveLength(25);
+
+      setCellContent(model, "A4", "Block spread");
```
LucasLefevre, line 773: I was running this test, trying to debug the evaluation cycle to asses what impacts your changes have. It turns out `"A1"` is now evaluated twice when setting content in "A4", even though it should not. One iteration with [A4, A1] (in that order), then a second iteration with A1 again (which is useless)
I'll check more in details what we should do instead tomorrow.

LucasLefevre: If I understand well, the motivation behind the changes in the evaluation was for `getSpreadPositionsOf` (which was not including the top-left cell).
Could we just change the implementation of `getSpreadPositionsOf` to return what we want without changing anything else to the evaluation? Also, it could actually return a `Zone`.

src/types/table.ts
```diff
@@ -27,6 +40,8 @@ export interface TableConfig {
   bandedRows: boolean;
   bandedColumns: boolean;
 
+  forceStaticTable?: boolean;
```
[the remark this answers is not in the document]
hokolomopo: Changed to `"dynamic "| "static" | "forceStatic"`, and removed `forceStatic` option in the config

src/plugins/ui_core_views/dynamic_tables.ts
```diff
@@ -0,0 +1,254 @@
...
+  }
+
+  private computeTables(sheetId: UID): Table[] {
+    const tables: Table[] = [];
+    const coreTables = this.getters.getCoreTables(sheetId);
+
+    // First we create the static tables, so we can use them to compute collision with dynamic tables
+    for (const table of coreTables) {
+      if (table.type === "dynamic") continue;
+      tables.push(this.coreTableToTable(sheetId, table));
```
LucasLefevre, line 87: call to `coreTableToTable` not useful here
```suggestion
      tables.push(table);
```

src/plugins/ui_core_views/dynamic_tables.ts
```diff
@@ -0,0 +1,254 @@
...
+    return this.getTables(sheetId).filter((table) =>
+      zones.some((zone) => overlap(table.range.zone, zone))
+    );
+  }
+
+  doesZonesContainFilter(sheetId: UID, zones: Zone[]): boolean {
+    return this.getTablesOverlappingZones(sheetId, zones).some((table) => table.config.hasFilters);
+  }
+
+  doesZonesContainTable(sheetId: UID, zones: Zone[]): boolean {
```
LucasLefevre, line 154: not used anymore

src/plugins/ui_core_views/dynamic_tables.ts
```diff
@@ -0,0 +1,254 @@
...
+  getFilterId(position: CellPosition): FilterId | undefined {
+    return this.getFilter(position)?.id;
+  }
+
+  getTable({ sheetId, col, row }: CellPosition): Table | undefined {
+    return this.getTables(sheetId).find((table) => isInside(col, row, table.range.zone));
+  }
+
+  /** Get the tables that are fully inside the given zone */
+  getTablesInZone(sheetId: UID, zone: Zone): Table[] {
```
LucasLefevre, line 137: not used anymore

tests/table/dynamic_table_plugin.test.ts
```diff
@@ -0,0 +1,222 @@
...
+  Object.assign(TABLE_PRESETS, oldTablePresets);
+});
+
+function getTables(model: Model, sheetId: UID) {
+  return model.getters
+    .getTables(sheetId)
+    .map((table) => ({ ...table, zone: zoneToXc(table.range.zone) }));
+}
+
+describe("Dynamic tables", () => {
```
LucasLefevre, line 47: add a test which duplicates a sheet with a dynamic table.  It doesn't seem there is one.

LucasLefevre (commented):
a few more comments. I think we're good after :)

tests/table/dynamic_table_plugin.test.ts
hokolomopo: And it was, in fact, bugged :sweat_smile:

LucasLefevre (approved):
robodoo r+ rebase-ff
:)

VincentSchippefilt:
robodoo r+ rebase-ff

PR #2631, opened 2023-06-26
[FIX] Filters: fix filter id on sheet duplication

rrahir:
The internal state of the filter plugin requires that the id in the table mapping matches the id of the related FilterTable object.

Task: 3384840

## Description:

description of this task, what is implemented and why it is implemented that way.

Odoo task ID : [3384840](https://www.odoo.com/web#id=3384840&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] feature is organized in plugin, or UI components
- [ ] support of duplicate sheet (deep copy)
- [ ] in model/core: ranges are Range object, and can be adapted (adaptRanges)
- [ ] in model/UI: ranges are strings (to show the user)
- [ ] undo-able commands (uses this.history.update)
- [ ] multiuser-able commands (has inverse commands and transformations where needed)
- [ ] new/updated/removed commands are documented
- [ ] exportable in excel
- [ ] translations (\_lt("qmsdf %s", abc))
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes
- [ ] doc is rebuild (npm run doc)
- [ ] status is correct in Odoo

PR #3629, opened 2024-02-07
[FIX] DataFilter: Fix overlapping filters

rrahir:
The fix proposed in PR #2631 was incomplete. While it prevented the *new* duplication of filter tables on overlapping zones (due to a mismatch inthe plugin internal mapping) it didn't fix the filters already duplicated and overlapping.

This revision adds a new migration to clear the duplicated filters based on the knowledge of how the bug could occur.

Before #2631, when duplicating a sheet containing a filter, the filter plugin would:
- Deep copy the entry of `this.tables[cmd.sheetIdFrom]` - an Object mapping the id of a `FilterTable` with the corresponding `FilterTable` object
- create a new entry for `cmd.sheetIdTo` in `this.tables` with the deep copy

However, deep copy would create a new `FilterTable` object with a new generated id (type uuid) which would mismatch with the internal mapping of the copied object.

e.g.
```typescript
// in sheet1
table1 = {id1: filter1 instanceof FilterTable}
id1 === filter1.id

// in sheet2
table2 = deepCopy(table1)
table2 ===  {id1: filter2 instanceof FilterTable}
id1 !== filter2.id
```

When trying to then alter the copied sheet structure and such modification altered the filters, the implementation did not account for such situation and would end up duplicating the filter table in `sheet2`, thus creating a sheet with filters with overlapping zones.

E.g.

``` typescript
tables["sheet2"] = {id1: filter2}
filter2.zone === "A1:A10"

deleteRow(5)

tables["sheet2"] = {id1: filter2, id3:filter3}
id1 !== filter2.id
id3 === filter3.id

// overlap in filte zones
filter2.zone === "A1:A10"
filter3.zone === "A1:A9"
```

Since Javascript guarantee `Object` property insertion order upon iteration (for equivalent objects [1]), we can safely say that in such situation the latest key of the object `tables["sheet2"]` is the truth and we can delete the other keyslinked to filters with overlapping zones.

One of the side-effects of these overlapping tables was a crash in the owl component `FilterIconsOverlay` due to some duplicated keys. As such, a regression test was added to cover this case.

[1] https://262.ecma-international.org/10.0/

Task: 3728009

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/migrations/data.ts
+++ b/src/migrations/data.ts
@@ -4,7 +4,15 @@ import {
   FORBIDDEN_IN_EXCEL_REGEX,
   FORMULA_REF_IDENTIFIER,
 } from "../constants";
-import { deepCopy, getItemId, toXC, toZone, UuidGenerator } from "../helpers/index";
+import {
+  deepCopy,
+  getItemId,
+  overlap,
+  toXC,
+  toZone,
+  UuidGenerator,
+  zoneToXc,
+} from "../helpers/index";
 import { StateUpdateMessage } from "../types/collaborative/transport_service";
 import {
   CoreCommand,
@@ -14,6 +22,7 @@ import {
   SheetData,
   UID,
   WorkbookData,
+  Zone,
 } from "../types/index";
 import { XlsxReader } from "../xlsx/xlsx_reader";
 import { normalizeV9 } from "./legacy_tools";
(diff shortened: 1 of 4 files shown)


LucasLefevre (approved):
robodoo r+

PR #3650, opened 2024-02-09
[IMP] chart: Improve scorecard design

anhe-odoo:
## Task Description

The aim of this task is to improve the design of the scorecards to take advangate of the available size instead of having some strange rule for the fontsize of some parts. The key value now take as much place as possible, and the baseline description is able to split into two lines to make the baseline value bigger.

## Related Task

- Task: [3716974](https://www.odoo.com/web#id=3716974&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/helpers/text_helper.ts
```diff
@@ -224,6 +224,40 @@ export function getContextFontSize(font: string): Pixel {
...
+  position: PixelPosition
+) {
+  const splitted = splitText(text);
+  const measure = context.measureText(splitted[1]);
+  context.fillText(splitted[1], position.x, position.y);
+  position.y -= measure.actualBoundingBoxAscent + measure.actualBoundingBoxDescent + 2;
+  context.fillText(splitted[0], position.x, position.y);
+}
+
+export function splitText(text: string): string[] {
```
hokolomopo, line 239:
```suggestion
export function splitTextInTwoLines(text: string): string[] {
```
or something like that

src/helpers/text_helper.ts
```diff
@@ -224,6 +224,40 @@ export function getContextFontSize(font: string): Pixel {
...
+  }
+  const length = Math.floor(text.length);
+  let leftSpace = Math.floor(length / 2),
+    rightSpace = Math.floor(length / 2);
+  while (leftSpace > 0 && text[leftSpace] !== " ") {
+    leftSpace--;
+  }
+  while (rightSpace < text.length && text[rightSpace] !== " ") {
+    rightSpace++;
+  }
```
hokolomopo, line 253: You cen definetly just trim() rather than looping there

src/helpers/text_helper.ts
```diff
@@ -224,6 +224,40 @@ export function getContextFontSize(font: string): Pixel {
...
+export function drawSplittedText(
+  context: CanvasRenderingContext2D,
+  text: string,
+  position: PixelPosition
+) {
+  const splitted = splitText(text);
+  const measure = context.measureText(splitted[1]);
+  context.fillText(splitted[1], position.x, position.y);
+  position.y -= measure.actualBoundingBoxAscent + measure.actualBoundingBoxDescent + 2;
+  context.fillText(splitted[0], position.x, position.y);
```
hokolomopo, line 236: Can have overlapping text :x 
image

src/helpers/figures/charts/scorecard_chart_config_builder.ts
```diff
@@ -218,27 +249,54 @@ class ScorecardChartConfigBuilder {
...
+        this.context.font = getDefaultContextFont(fontSize);
+        return this.context.measureText(this.baselineDescr).width;
+      }
+    );
+
+    let splittedBaseline = false;
+    if (baselineDescrFontSize < baselineValueFontSize / 2.5) {
+      splittedBaseline = true;
+      baselineDescrFontSize = Math.floor(baselineValueFontSize / 2.5);
+      for (const str of splitText(this.baselineDescr)) {
```
hokolomopo, line 285: We should probably put the splitted baseline in the `ScorecardChartConfig`,rather than hope that we split the exact same way in both this function and the text rendering function

hokolomopo (commented):
Better scorecards :partying_face:

hokolomopo:
Key value should eb centered if there is nothing else on the chart 

image

src/helpers/text_helper.ts
anhe-odoo: I'm not sure you understood the loop. Here I'm searching for the first space right and left of the middle to know where to split the line. I don't see how a trim can help me doing that ?

src/helpers/text_helper.ts
```diff
@@ -224,6 +224,40 @@ export function getContextFontSize(font: string): Pixel {
...
+}
+
+export function splitText(text: string): string[] {
+  let spaces = "";
+  while (text[0] === " ") {
+    spaces += " ";
+    text = text.slice(1);
+  }
```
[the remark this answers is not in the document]
anhe-odoo: The idea here was to keep the leading space, in case of the user has put some starting space in the baseline description to have a spacing between the value and the spacing. As this is something that can be done for the moment, I didn't want to change this behavior, even for multiline description :)

src/helpers/text_helper.ts
```diff
@@ -224,6 +228,90 @@ export function getContextFontSize(font: string): Pixel {
...
 }
 
+export function fittingString(ctx, str: string, maxWidth: number) {
+  var width = ctx.measureText(str).width;
+  var ellipsis = "…";
+  var ellipsisWidth = ctx.measureText(ellipsis).width;
+  if (width <= maxWidth || width <= ellipsisWidth) {
+    return str;
+  } else {
+    var len = str.length;
```
LucasLefevre, line 238: `var` :eyes: smells like something from stack overflow :smile: 
If it is, it's good not to re-invent the wheel and use what others have already done, but it requires the reference to the source to credit the original author :wink:

src/helpers/text_helper.ts
```diff
@@ -224,6 +228,90 @@ export function getContextFontSize(font: string): Pixel {
   return Number(font.match(pxRegex)?.[1]);
 }
 
+export function fittingString(ctx, str: string, maxWidth: number) {
+  var width = ctx.measureText(str).width;
+  var ellipsis = "…";
+  var ellipsisWidth = ctx.measureText(ellipsis).width;
```
LucasLefevre, line 234: measure text can be expensive. Can `computeTextWidth` be used here (it's cached)

src/helpers/text_helper.ts
```diff
@@ -224,6 +228,90 @@ export function getContextFontSize(font: string): Pixel {
   return Number(font.match(pxRegex)?.[1]);
 }
 
+export function fittingString(ctx, str: string, maxWidth: number) {
```
LucasLefevre, line 231:
```suggestion
export function clipText(ctx, str: string, maxWidth: number) {
```
maybe ? or `clipTextWithEllipsis` which is even clearer ?

src/helpers/text_helper.ts
```diff
@@ -224,6 +228,90 @@ export function getContextFontSize(font: string): Pixel {
...
+  var width = ctx.measureText(str).width;
```
LucasLefevre, line 232: doesn't the size also depends on the font weight (bold)?

src/helpers/figures/charts/scorecard_chart_config_builder.ts
```diff
@@ -269,37 +353,27 @@ class ScorecardChartConfigBuilder {
...
               color: this.runtime.baselineColor || this.secondaryFontColor,
             },
     };
   }
 
   /** Get the height of the chart minus all the vertical paddings */
-  private getDrawableHeight(): number {
-    const verticalPadding = 2 * this.chartPadding;
-    let availableHeight = this.height - verticalPadding;
+  private get getDrawableHeight(): number {
```
LucasLefevre, line 371: Why this change ?
Either it's a getter and the name should not start with `get` or it's a function.
```suggestion
  private getDrawableHeight(): number {
```

src/helpers/figures/charts/scorecard_chart_config_builder.ts
```diff
@@ -50,8 +46,19 @@ export type ScorecardChartConfig = {
     position: PixelPosition;
   };
   baseline?: ScorecardChartElement;
-  baselineDescr?: ScorecardChartElement;
```
LucasLefevre, line 53: could it be just changed to an array of elements ? `baselineDescr?: ScorecardChartElement[];`
There would be less ifs in the scorecard component such as `if (structure.baselineDescr.splitted)`)

The benefit would be to keep all the "complex" computations on the model/plugin side rather than in the component (because components are hard to test, even more when it's actually a canvas)

LucasLefevre (commented):
First batch of comments.
It looks like there's an issue in the text splitting mechanism (see screenshot in the task)

src/helpers/text_helper.ts
anhe-odoo: Yes, it has been copied from SO :D I added the link to the answer :)

src/helpers/text_helper.ts
anhe-odoo: Yes, but the canvas should have been correctly configurated to take that into account before

src/helpers/figures/charts/scorecard_chart_config_builder.ts
```diff
@@ -269,37 +368,27 @@ class ScorecardChartConfigBuilder {
           this.secondaryFontColor,
       },
       baselineDescr: {
-        font: getDefaultContextFont(baselineFontSize * BASELINE_DESCR_FONT_RATIO),
+        font: getDefaultContextFont(baselineDescrFontSize),
+        splitted: splittedBaseline,
```
LucasLefevre, line 372: `splitted` does not appear to be used in english. Could you rename all variables where it's used?

https://english.stackexchange.com/questions/467873/the-past-participle-of-split-split-or-splitted#:~:text=Splitted%20appears%20to%20be%20a,past%20participle%20of%20the%20verb.&text=Google%20Books%20shows%20very%20few%20usage%20instances%20of%20splitted%20compared%20to%20split.
```suggestion
        isSplit: isBaselineSplit,
```

src/helpers/figures/charts/scorecard_chart_config_builder.ts
```diff
@@ -216,29 +282,62 @@ class ScorecardChartConfigBuilder {
...
+      baselineValueFontSize,
+      (fontSize: number) => computeTextWidth(this.context, this.baselineDescr, { fontSize })
+    );
+
+    let splittedBaseline = false;
+    if (baselineDescrFontSize < baselineValueFontSize / 2.5) {
+      splittedBaseline = true;
+      baselineDescrFontSize = Math.floor(baselineValueFontSize / 2.5);
+      for (const str of this.splittedBaselineDescr) {
+        const strWidth = getFontSizeMatchingWidth(
```
LucasLefevre, line 327:
```suggestion
      for (const line of this.splittedBaselineDescr) {
        const lineWidth = getFontSizeMatchingWidth(
```

src/helpers/format.ts
```diff
@@ -168,6 +171,10 @@ function applyInternalFormat(
 }
 
 function applyInternalNumberFormat(value: number, format: InternalNumberFormat, locale: Locale) {
+  if (value === Infinity) {
```
LucasLefevre, line 174: also needs to handle `-Infinity`

src/helpers/figures/charts/scorecard_chart.ts
```diff
@@ -8,37 +8,123 @@ import { _t } from "../../../translation";
...
+    }
+    return baseline.formattedValue;
+  }
+  let { value, format } = baseline;
+  value = Math.abs(keyValue.value - value);
+  if (baselineMode === "percentage" && value !== 0) {
+    value = value / baseline.value;
+  }
+  if (baselineMode === "percentage") {
+    format = "0%";
```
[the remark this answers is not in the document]
anhe-odoo: I was thinking the same, but some test were failed if I kept one decimal so i though that having integer was a desired behavior. Anyway, one (or two) decimals are better in my opinion, even for a large percentage

src/helpers/figures/charts/scorecard_chart_config_builder.ts
```diff
@@ -102,41 +112,48 @@ class ScorecardChartConfigBuilder {
...
       this.baseline,
       style.baselineValue.font
     );
-    const { width: baselineDescrWidth } = this.getTextDimensions(
-      this.baselineDescr,
-      style.baselineDescr.font
-    );
+    if (!this.baseline) {
+      baselineHeight = this.getTextDimensions(this.baselineDescr, style.baselineDescr.font).height;
+    }
```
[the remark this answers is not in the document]
anhe-odoo: We can have a description without any baseline value. I even think that some dashboard have this kind of stuff :)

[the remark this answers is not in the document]
anhe-odoo: But, indeed, it's not tested. I'll add one

src/helpers/format.ts
anhe-odoo: No, this function is called with Math.abs(value), and after a "minus" sign is add to the result in case of negative value

src/helpers/figures/charts/scorecard_chart_config_builder.ts
```diff
@@ -156,14 +165,14 @@ class ScorecardChartConfigBuilder {
     }
 
     if (this.baselineDescr) {
-      if (style.baselineDescr.splitted) {
-        const description = this.splittedBaselineDescr;
+      const position = {
```
LucasLefevre, line 168: diff should be in the previous commit

src/helpers/format.ts
```diff
@@ -584,6 +584,18 @@ export function roundFormat(format: Format): Format {
   return convertInternalFormatToFormat(roundedFormat);
 }
 
+export function humanizeNumber({ value, format }: FPayload, locale: Locale): string {
```
LucasLefevre, line 587: move should be in previous commit

PR #3670, opened 2024-02-12
[IMP] index: export ClipboardHandlers Classes

anhe-odoo:
## Description

With the new clipboard refactoring, we are now able to implement clipboard handlers to extend the o-spreadsheet clipboard behavior. Then, if one's wants to implement a new handler, it should be better to use the abstract classes as a base class for the handler, so this commit aims to export them.

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -339,3 +339,6 @@ export type {
   FPayload,
 } from "./types";
 export type { FunctionRegistry };
+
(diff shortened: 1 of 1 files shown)


rrahir:
@robodoo r+

PR #3672, opened 2024-02-12
[PERF] several perf improvement to reduce memory allocation

LucasLefevre:
## Description:

With this PR, `model.exportData()` memory allocation is reduced by ~75%.
From ~995MB to ~240MB

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/misc.ts
+++ b/src/helpers/misc.ts
@@ -252,8 +252,8 @@ export function isObjectEmptyRecursive<T extends object>(argument: T | undefined
  * If the given item does not exist in the dictionary, it creates one with a new id.
  */
 export function getItemId<T>(item: T, itemsDic: { [id: number]: T }) {
-  for (let [key, value] of Object.entries(itemsDic)) {
-    if (deepEquals(value, item)) {
+  for (const key in itemsDic) {
+    if (deepEquals(itemsDic[key], item)) {
       return parseInt(key, 10);
     }
   }
--- a/src/helpers/misc.ts
+++ b/src/helpers/misc.ts
@@ -357,12 +357,14 @@ export function deepEquals(o1: any, o2: any): boolean {
   if (typeof o1 !== "object") return o1 === o2;
 
   // Objects can have different keys if the values are undefined
-  const keys = new Set<string>();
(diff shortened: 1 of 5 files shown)


rrahir:
@robodoo rebase-ff

rrahir (approved):

src/helpers/misc.ts
```diff
@@ -357,11 +357,13 @@ export function deepEquals(o1: any, o2: any): boolean {
   if (typeof o1 !== "object") return o1 === o2;
 
   // Objects can have different keys if the values are undefined
-  const keys = new Set<string>();
-  Object.keys(o1).forEach((key) => keys.add(key));
-  Object.keys(o2).forEach((key) => keys.add(key));
+  for (const key in o2) {
+    if (!(key in o1) && o2[key] !== undefined) {
```
VincentSchippefilt, line 361: isn't this testing that keys in `o2` that are defined must be in `o1` ?
what about keys in `o1` that might not be in `o2`?

src/helpers/misc.ts
```diff
@@ -357,11 +357,13 @@ export function deepEquals(o1: any, o2: any): boolean {
...
+      return false;
+    }
+  }
 
-  for (let key of keys) {
+  for (const key in o1) {
     if (typeof o1[key] !== typeof o1[key]) return false;
```
VincentSchippefilt, line 367: the intention was probably to compare the type of the values of `o1[key]` and `o2[key]`
```suggestion
    if (typeof o1[key] !== typeof o2[key]) return false;
```

src/helpers/misc.ts
LucasLefevre: > isn't this testing that keys in o2 that are defined must be in o1 ?

keys in `o2` that are not in `o1` must have their value `undefined` (otherwise it means the two objects are not equals)

> what about keys in o1 that might not be in o2?

In the for loop below, we iterate over keys of `o1` and check `if (o1[key] !== o2[key]) return false;`

VincentSchippefilt:
robodoo r+

PR #3675, opened 2024-02-12
[FIX] selection_inputs_manager: arrow keys when broken ranges

dhrp-odoo:
## Description:

Previously, the arrow keys failed to function as expected when broken ranges were present in the selection input.

### Steps to reproduce:

From the side panel of a chart with a SelectionInput containing two ranges:
- Enter a valid range in the second input.
- Corrupt the first input by entering "A11wwwz".
- Press Enter to confirm.
- Reselect the first range.
- Attempt to use the keyboard arrows to change the zone; however, this does not work.
- Select the second input.
- Attempt to use the keyboard arrows to change the zone; again, this does not work.

To address this issue, the if statement in command handling from the plugin was removed to ensure it allows capturing the range.

Task: : [3626171](https://www.odoo.com/web#id=3626171&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_feature/selection_inputs_manager.ts
+++ b/src/plugins/ui_feature/selection_inputs_manager.ts
@@ -98,9 +98,9 @@ export class SelectionInputsManagerPlugin extends UIPlugin {
         break;
       case "FOCUS_RANGE":
       case "CHANGE_RANGE":
-        if (cmd.id !== this.focusedInputId) {
-          const input = this.inputs[cmd.id];
-          const range = input.ranges.find((range) => range.id === cmd.rangeId);
+        const input = this.inputs[cmd.id];
+        const range = input.ranges.find((range) => range.id === cmd.rangeId);
+        if (range) {
           const sheetId = this.getters.getActiveSheetId();
           const zone = this.getters.getRangeFromSheetXC(sheetId, range?.xc || "A1").zone;
           this.selection.capture(
--- a/tests/plugins/selection_input.test.ts
+++ b/tests/plugins/selection_input.test.ts
@@ -700,4 +700,19 @@ describe("selection input plugin", () => {
     expect(model.getters.getSelectionInput(id2)[1].xc).toBe("F2");
     expect(model.getters.getSelectionInput(id2)[1].isFocused).toBe(true);
   });
+
+  test("Selection anchor updates when focusing on a new range within the same SelectionInput component", async () => {
+    model.dispatch("ENABLE_NEW_SELECTION_INPUT", { id, initialRanges: ["TEST", "B2"] });
+    model.dispatch("FOCUS_RANGE", { id, rangeId: idOfRange(model, id, 0) });
+
+    expect(model.getters.getSelectionInput(id)[0].xc).toBe("TEST");
+    model.selection.moveAnchorCell("down");
+    expect(model.getters.getSelectionInput(id)[0].xc).toBe("TEST");
+
+    model.dispatch("FOCUS_RANGE", { id, rangeId: idOfRange(model, id, 1) });
+
+    expect(model.getters.getSelectionInput(id)[1].xc).toBe("B2");
+    model.selection.moveAnchorCell("down");
+    expect(model.getters.getSelectionInput(id)[1].xc).toBe("B3");
+  });
 });


tests/components/selection_input.test.ts
```diff
@@ -263,6 +263,23 @@ describe("Selection Input", () => {
     app.destroy();
   });
 
+  test("Can use arrow keys to change ranges after input corruption", async () => {
```
rrahir, line 266: Since you are indirectly fixing the "handleEvent" function of the plugin  (unless I'm mistaken), the test should take place in the selection input plugin  ^^ no need to instanciate a whole component.

src/plugins/ui_feature/selection_inputs_manager.ts
```diff
@@ -98,9 +98,9 @@ export class SelectionInputsManagerPlugin extends UIPlugin {
         break;
       case "FOCUS_RANGE":
       case "CHANGE_RANGE":
-        if (cmd.id !== this.focusedInputId) {
-          const input = this.inputs[cmd.id];
-          const range = input.ranges.find((range) => range.id === cmd.rangeId);
+        const input = this.inputs[cmd.id];
```
rrahir, line 101: it seems weird to me  -  If we are already focused on the range, then dispatching a "focus_range" should have 0 effect (same goes for change range) since by definition, we already captured the selection.  Isn't there something else that we are missing? 
It looks like we currently can reach a state where we have `this.focusedInputId` defined but without having the control of the selection ,which would be the root problem and we are only masking it because we know that the components dispatch some specific commands. Let me know what you think about this :)

dhrp-odoo: If I understand correctly, `FOCUS_RANGE` makes sense even when clicking on different ranges in the same SelectionInput component. This is because we need to capture a zone that allows us to select the next cell whenever the user presses the arrow keys. right?

tests/components/selection_input.test.ts
```diff
@@ -263,6 +263,23 @@ describe("Selection Input", () => {
...
+    selectionInputEl[0].dispatchEvent(new KeyboardEvent("keydown", { key: "ArrowLeft" }));
+    await nextTick();
+    expect((selectionInputEl[0] as HTMLInputElement)?.value).toEqual("TEST");
+
+    focus(1);
+    await nextTick();
+    selectionInputEl[1].dispatchEvent(new KeyboardEvent("keydown", { key: "ArrowLeft" }));
+    await nextTick();
+    expect((selectionInputEl[1] as HTMLInputElement)?.value).toEqual("B2");
+    app.destroy();
```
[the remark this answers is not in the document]
dhrp-odoo: Maybe adding app.destroy() was a precautionary step for ensuring proper cleanup post-test execution. I included it because I observed similar cleanup steps in this file at 2-3 other places.

src/plugins/ui_feature/selection_inputs_manager.ts
rrahir: Yes, you are right - I (again) mistook the generic selection input with one of the range inputs.

tests/components/selection_input.test.ts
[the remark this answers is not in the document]
rrahir: If you have a look at the `mountComponent` helper, you will see that we actually register the app to be destroyed at the end of the test :) the only places where it makes sense to destroy the app by yourself are:
- testing a side effect of the destroy/unmount (like in the test `unmounting deletes the state`)
- when you want to mount several components in the same test like in test `The containerRect is the spreadsheet element if it is not explicitly in the popover props` 

Both practices should not be encouraged in general ,it has to stay an exception ^^



* the test `F2 alters edition mode` also does this but its not necessary anymore, probably a forward port that went bad

rrahir (commented):
The PR looks good :) if you could address the change for the test, we then could merge it :)

dhrp-odoo:
> The PR looks good :) if you could address the change for the test, we then could merge it :)

Changes made! Thanks :)

rrahir:
@robodoo r+

PR #3725, opened 2024-02-19
[FIX] side_panel: dynamic props

LucasLefevre:
## Description:

There are several problems in the current way side panels "props" are designed.

If the side panel receives an id as a prop (e.g. `figureId`, `pivotId`), we have to check that the underlying figure exists when the side panel is open. We also have to check at every render that it has not been deleted.

It's currently done with `onWillUpdateProps` which closes the side panel if the id is no longer valid.
Closing the side panel at this point (while rendering) breaks the mental model of `UI = f(data)`. It's also error prone as the state is changed in the middle of a render. It means `sidePanel.isOpen` might be `true` at the begining of the template but the side panel might actually be closed later in the template.

`onWillUpdateProps` should be avoided as much as possible.

This commit adds the ability to declare a function in the side panel registry entry to:
1) validate the props
2) generate props based on the model (e.g. the selected figure/pivot/...)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/figures/figure_chart/figure_chart.ts
+++ b/src/components/figures/figure_chart/figure_chart.ts
@@ -30,10 +30,8 @@ export class ChartFigure extends Component<Props, SpreadsheetChildEnv> {
   static components = {};
 
   onDoubleClick() {
-    const result = this.env.model.dispatch("SELECT_FIGURE", { id: this.props.figure.id });
-    if (result.isSuccessful) {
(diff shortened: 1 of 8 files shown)


LucasLefevre (commented):
TODO: clean selected figure in the selection plugin

rrahir:
@robodoo r+

PR #3726, opened 2024-02-19
[REF] filter: move filters icons inside grid overlay

hokolomopo:
## Description

This commit moves the filters icon overlay from `grid.xml/dashboard.xml` to `grid_overlay.xml`. This has several benefits:

- since the icons are positioned inside the grid overlay, they don't have to worry about the grid/dashboard position offset. The GridOverlay handles that.
- we can call the onMouseDown of grid from the FilterIconOverlay, selecting the cell with all its side effect (eg. closing the current grid composer on left click, opening the context menu on the correct cell on right click, ...)

The annoying thing is that we need to prevent the mouseDown to close the filter menu when clicking on the filter icon. This is handled on the click of the filter icon. Otherwise, the menu toggle won't work.

Task: : [3754487](https://www.odoo.com/web#id=3754487&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/filters/filter_icons_overlay/filter_icons_overlay.ts
+++ b/src/components/filters/filter_icons_overlay/filter_icons_overlay.ts
@@ -1,23 +1,20 @@
 import { Component } from "@odoo/owl";
-import { CellPosition, DOMCoordinates, SpreadsheetChildEnv } from "../../../types";
+import { CellPosition, SpreadsheetChildEnv } from "../../../types";
 import { GridCellIcon } from "../../grid_cell_icon/grid_cell_icon";
 import { FilterIcon } from "../filter_icon/filter_icon";
 
 interface Props {
-  gridPosition: DOMCoordinates;
+  onMouseDown: (ev: MouseEvent) => void;
 }
 
 export class FilterIconsOverlay extends Component<Props, SpreadsheetChildEnv> {
   static template = "o-spreadsheet-FilterIconsOverlay";
   static props = {
-    gridPosition: { type: Object, optional: true },
+    onMouseDown: Function,
   };
   static components = {
     GridCellIcon,
     FilterIcon,
-  };
-  static defaultProps = {
-    gridPosition: { x: 0, y: 0 },
   };
 
   getFilterHeadersPositions(): CellPosition[] {
--- a/src/components/grid_overlay/grid_overlay.xml
+++ b/src/components/grid_overlay/grid_overlay.xml
@@ -10,6 +10,7 @@
       t-on-contextmenu.stop.prevent="onContextMenu">
       <FiguresContainer onFigureDeleted="props.onFigureDeleted"/>
       <DataValidationOverlay/>
+      <FilterIconsOverlay onMouseDown="(ev) => this.onMouseDown(ev, {closePopover: false})"/>
       <GridAddRowsFooter
         t-if="!env.model.getters.isReadonly()"
         t-key="env.model.getters.getActiveSheetId()"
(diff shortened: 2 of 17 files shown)


src/components/filters/filter_icons_overlay/filter_icons_overlay.ts
```diff
@@ -1,28 +1,29 @@
...
-  };
 
   getFilterHeadersPositions(): CellPosition[] {
     const sheetId = this.env.model.getters.getActiveSheetId();
     const headerPositions = this.env.model.getters.getFilterHeaders(sheetId);
     return headerPositions.map((position) => ({ sheetId, ...position }));
   }
+
+  onMouseDown(ev: MouseEvent) {
+    this.props.onMouseDown(ev, false); // don't close popover on icon click
```
hokolomopo, line 27: Not a big fan of this. This is needed otherwise the mouse down would close the filter menu (grid onCellClicked), then the menu will be re-open on the onClick of the filter Icon.

src/components/grid_overlay/grid_overlay.xml
```diff
@@ -10,6 +10,7 @@
       t-on-contextmenu.stop.prevent="onContextMenu">
       <FiguresContainer onFigureDeleted="props.onFigureDeleted"/>
       <DataValidationOverlay/>
+      <FilterIconsOverlay onMouseDown.bind="this.onMouseDown"/>
```
rrahir, line 13:
```suggestion
      <FilterIconsOverlay onMouseDown.bind="(ev) => this.onMouseDown(ev, {closePopover: false})"/>
```
pass this version to the child component, it does not need to know about that specific option if it calls it with the same payload all the time

src/components/filters/filter_icons_overlay/filter_icons_overlay.ts
```diff
@@ -1,28 +1,29 @@
 import { Component } from "@odoo/owl";
-import { CellPosition, DOMCoordinates, SpreadsheetChildEnv } from "../../../types";
+import { CellPosition, SpreadsheetChildEnv } from "../../../types";
 import { GridCellIcon } from "../../grid_cell_icon/grid_cell_icon";
 import { FilterIcon } from "../filter_icon/filter_icon";
 
 interface Props {
-  gridPosition: DOMCoordinates;
+  onMouseDown: (ev: MouseEvent, modifiers: { closePopover: boolean }) => void;
```
rrahir, line 7: out of the scope of the task but I wonder if it's a good idea to pass our event to a handler of the parent. It feels like it could lead to weird side effects where the parent would end up trusting the event as if they were to actual target and start handling the ev.offset*/ev.client* which might not be aligned with the parent

hokolomopo: Yeah that could lead to errors. On the other hand, the parent trusting the event of a child on which it has no control of seems like a programming error. 

We could do something like sanitizing the event and removing the target/offset, whild keeping the "absolute" event modifiers. I don't think it's worth tho.

LucasLefevre (commented):
robodoo r+ (squashed the commits and run prettier)

src/components/filters/filter_icons_overlay/filter_icons_overlay.ts
rrahir: turns out it's not such. a good idea :/ https://www.odoo.com/web#id=3868838&cids=1&menu_id=4720&action=333&active_id=2328&model=project.task&view_type=form

hokolomopo: image

PR #3729, opened 2024-02-19
[ADD] data sources: introduce data sources

pro-odoo:
This commit introduces the concept of data sources. A data source is an instance of a class that is responsible for handling/processing/fetching data.

For now, there is no real implementation, but the concept is introduced as it will be necessary for the introduction of the new spreadsheet pivot table.

Note that this concept is coming from the odoo codebase.

Task: 3749157

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/data_sources.ts
+++ b/src/helpers/data_sources.ts
@@ -1,0 +1,29 @@
+import { EventBus } from "@odoo/owl";
+import { ModelConfig } from "../model";
+
+type DataSourceConstructor = new (custom: ModelConfig["custom"], params: unknown) => unknown;
+
+export class DataSources extends EventBus {
+  private dataSources: Record<string, unknown> = {};
+
+  constructor(private custom: ModelConfig["custom"]) {
+    super();
+  }
+
+  add(id: string, cls: DataSourceConstructor, params: unknown) {
+    this.dataSources[id] = new cls(this.custom, params);
+    return this.dataSources[id];
+  }
+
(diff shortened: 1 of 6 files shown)


src/helpers/data_sources.ts
```diff
@@ -0,0 +1,29 @@
+import { EventBus } from "@odoo/owl";
+import { ModelConfig } from "../model";
+
+type DataSourceConstructor = new (custom: ModelConfig["custom"], params: unknown) => unknown;
+
+export class DataSources extends EventBus {
```
LucasLefevre, line 6: shouldn't we also move `notifyWhenPromiseResolves`/`notify` here ?

LucasLefevre (commented):
Actually what's the purpose of `DataSources` now ? It looks like all the important/useful stuff has been moved to `OdooDataProvider` and `DataSources` is now only an object to store key/value pairs.
The only interesting part is `new cls(this.custom, params);` to instantiate "something".

Could we replace all the `dataSources.add(dataSourceId, OdooPivot, params)` (and same for list/charts) by `new OdooPivot(params)` and store it locally in the plugin that needs it.

I believe it would look more straight forward and less alambiqué than using this global (to the spreadsheet level) key/value object.
I can't help feeling this is now quite useless and we could do differently/simpler.

PR #3733, opened 2024-02-20
[FIX] headers_overlay: unhide cols/rows icon position broken

dhrp-odoo:
## Description:

- Previously, there was an issue with the position of the unhide cols/rows icon, which has now been rectified. This fix involves setting the icon's position to absolute, applying bootstrap classes, and removing CSS from the ts file for o-unhide-buttons.

- Furthermore, when the hiddenItems consist of the first or last column/row, bootstrap classes are now utilized to appropriately hide one of the icons.

Task: : [3696210](https://www.odoo.com/web#id=TASK_ID&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/headers_overlay/headers_overlay.ts
+++ b/src/components/headers_overlay/headers_overlay.ts
@@ -340,11 +340,6 @@ css/* scss */ `
       height: 10000px;
       background-color: ${SELECTION_BORDER_COLOR};
     }
-    .o-unhide-buttons {
-      width: fit-content;
-      gap: 5px;
-      transform: translate(-50%, 0);
-    }
     .o-unhide:hover {
       z-index: ${ComponentsImportance.Grid + 1};
       background-color: lightgrey;
@@ -536,10 +531,6 @@ css/* scss */ `
       width: 10000px;
       height: 1px;
       background-color: ${SELECTION_BORDER_COLOR};
-    }
-    .o-unhide-buttons {
-      height: fit-content;
-      transform: translate(0, -50%);
     }
     .o-unhide:hover {
       z-index: ${ComponentsImportance.Grid + 1};
--- a/src/components/headers_overlay/headers_overlay.xml
+++ b/src/components/headers_overlay/headers_overlay.xml
@@ -42,25 +42,22 @@
         t-as="hiddenItem"
         t-key="hiddenItem_index">
         <div
-          class="o-unhide-buttons position-relative float-end"
+          class="position-absolute end-0 translate-middle-y"
           t-att-style="getUnhideButtonStyle(hiddenItem[0])">
-          <t t-if="!hiddenItem.includes(0)">
-            <div
-              class="o-unhide rounded mb-1"
-              t-att-data-index="hiddenItem_index"
-              t-on-click="() => this.unhide(hiddenItem)">
-              <t t-call="o-spreadsheet-Icon.TRIANGLE_UP"/>
-            </div>
-          </t>
-          <t
-            t-if="!hiddenItem.includes(env.model.getters.getNumberRows(env.model.getters.getActiveSheetId())-1)">
-            <div
-              class="o-unhide rounded"
-              t-att-data-index="hiddenItem_index"
-              t-on-click="() => this.unhide(hiddenItem)">
-              <t t-call="o-spreadsheet-Icon.TRIANGLE_DOWN"/>
-            </div>
-          </t>
+          <div
+            class="o-unhide rounded mb-1"
+            t-att-class="{'invisible': hiddenItem.includes(0)}"
+            t-att-data-index="hiddenItem_index"
+            t-on-click="() => this.unhide(hiddenItem)">
+            <t t-call="o-spreadsheet-Icon.TRIANGLE_UP"/>
+          </div>
+          <div
+            class="o-unhide rounded"
+            t-att-class="{'invisible': hiddenItem.includes(env.model.getters.getNumberRows(env.model.getters.getActiveSheetId())-1)}"
+            t-att-data-index="hiddenItem_index"
+            t-on-click="() => this.unhide(hiddenItem)">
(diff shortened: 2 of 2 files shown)


src/components/headers_overlay/headers_overlay.ts
```diff
@@ -489,8 +491,20 @@ export class ColResizer extends AbstractResizer {
...
+  getUnhideButtonStyle(hiddenIndexes: HeaderIndex[]): string {
+    const lastIndex =
+      this.env.model.getters.getNumberCols(this.env.model.getters.getActiveSheetId()) - 1;
+
+    let leftOffset = this._getDimensionsInViewport(hiddenIndexes[0]).start;
+    if (hiddenIndexes.includes(0)) {
+      leftOffset += 10;
+    } else if (hiddenIndexes.includes(lastIndex)) {
+      leftOffset -= 10;
+    }
```
hokolomopo, line 503: Don't put arbitraty offset with no link to anything. The day we change exact sizes in the CSS this will become wrong. furthermore ATM the arrows are displaced with the transform CSS properties. Which is wrong for the extremities of the sheet. What your offset does in essence is to leave the wrong transform, but compensate the wrong transform so they become right.

We should either:
a) overwriting tranform CSS for the first/last col/row to it's correct
b) (rpobably better) leave the transform be, and apply CSS hidden to the arrows on the first/last col instead of the t-if. So the transform -50% will always be correct, since it takes both arrows into account. (and this can be done with a t-att-aclass in the tempalte directly)

src/components/headers_overlay/headers_overlay.ts
```diff
@@ -344,6 +345,7 @@ css/* scss */ `
       width: fit-content;
       gap: 5px;
       transform: translate(-50%, 0);
+      top: ${HEADER_HEIGHT / 2 - ICON_EDGE_LENGTH / 2 + "px"};
```
hokolomopo, line 348: Uh ? What's this offset for ? The icons are now not centered

src/components/headers_overlay/headers_overlay.ts
dhrp-odoo: Changes made as suggested. Thanks!

src/components/headers_overlay/headers_overlay.ts
hokolomopo: I meant that after your fix, the arrows are not centered vertically inside the headers. After your fix : 
image

Without the 4px top offset that you added. The arrows are now correctly centered vertically
image

dhrp-odoo: Oh, you want it centered?

I initially thought it should be positioned slightly below so that both the character and icon are visible if the column size is smaller.

I've made that adjustment. Thanks!

src/components/headers_overlay/headers_overlay.xml
```diff
@@ -42,25 +42,22 @@
         t-as="hiddenItem"
         t-key="hiddenItem_index">
         <div
-          class="o-unhide-buttons position-relative float-end"
+          class="o-unhide-buttons position-absolute float-end"
```
anhe-odoo, line 45: You can change this line to `position-absolute end-0 translate-middle-y`

src/components/headers_overlay/headers_overlay.xml
```diff
@@ -101,25 +98,22 @@
         t-as="hiddenItem"
         t-key="hiddenItem_index">
         <div
-          class="o-unhide-buttons position-relative h-100 d-flex align-items-center"
+          class="o-unhide-buttons position-absolute h-100 d-flex align-items-center"
```
anhe-odoo, line 101: You can change this line to `position-absolute h-100 d-flex align-items-center translate-middle-x gap-2`

anhe-odoo (changes requested):
For what I've seen, you can fix this issue with the two changes I propose here. All other changes are useless, and the two proposed lines will make you able to remove the `o-unhide-buttons` CSS style in headers_overlay.ts in favor to the use of bootstrap :)

dhrp-odoo:
> For what I've seen, you can fix this issue with the two changes I propose here. All other changes are useless, and the two proposed lines will make you able to remove the `o-unhide-buttons` CSS style in headers_overlay.ts in favor to the use of bootstrap :)

Ahhh, It seems you haven't consider the scenario involving icons on the first or last row/column. Right?

anhe-odoo:
@dhrp-odoo : Yes, indeed, I haven't tried the external limits, nice catch :) LGTM now

src/components/headers_overlay/headers_overlay.xml
```diff
@@ -42,25 +42,22 @@
...
-            <div
-              class="o-unhide rounded"
-              t-att-data-index="hiddenItem_index"
-              t-on-click="() => this.unhide(hiddenItem)">
-              <t t-call="o-spreadsheet-Icon.TRIANGLE_DOWN"/>
-            </div>
-          </t>
+          <div
+            class="o-unhide rounded mb-1"
+            t-att-class="{'invisible': hiddenItem.includes(0)}"
```
LucasLefevre, line 49: Any reason to change the `t-if` to `invisible` class ? Does it break the style with `t-if` ?
It's best to use `t-if` to avoid building the DOM (to make it invisible afterward)

dhrp-odoo: > Any reason to change the `t-if` to `invisible` class ? Does it break the style with `t-if` ?

Yes, there is. If the icon is positioned on the first or last row/column, it won't have one of the icons, leading to a misalignment.

I attempted to resolve this by adding some pixels, but then ADRM proposed using the invisible class instead. You can find the discussion [here](https://github.com/odoo/o-spreadsheet/pull/3733#discussion_r1497591212).

LucasLefevre (approved):
robodoo r+

PR #3743, opened 2024-02-26
[FIX] Grid: Context menu position was broken

dhrp-odoo:
## Description:

Previously, the context menu didn't open in the right spot due to a misplacement of t-on-contextmenu in grid.xml. This caused the menu's position to be recalculated unnecessarily.

This PR relocates t-on-contextmenu to its correct position.

Task: : [3746771](https://www.odoo.com/web#id=3746771&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/composer/composer/composer.ts
+++ b/src/components/composer/composer/composer.ts
@@ -100,6 +100,7 @@ export interface ComposerProps {
   focus: "inactive" | "cellFocus" | "contentFocus";
   onComposerContentFocused: () => void;
   onComposerCellFocused?: (content: String) => void;
+  onInputContextMenu?: (event: MouseEvent) => void;
   isDefaultFocus?: boolean;
 }
 
@@ -446,6 +447,12 @@ export class Composer extends Component<ComposerProps, SpreadsheetChildEnv> {
 
   onCompleted(text: string | undefined) {
     text && this.autoComplete(text);
+  }
+
+  onContextMenu(ev: MouseEvent) {
+    if (this.env.model.getters.getEditionMode() === "inactive") {
+      this.props.onInputContextMenu?.(ev);
+    }
   }
 
   // ---------------------------------------------------------------------------
--- a/src/components/grid/grid.ts
+++ b/src/components/grid/grid.ts
@@ -451,9 +451,8 @@ export class Grid extends Component<Props, SpreadsheetChildEnv> {
     } else if (this.env.model.getters.getActiveRows().has(row)) {
       type = "ROW";
     }
-    const { x, y, width, height } = this.env.model.getters.getVisibleRect(lastZone);
-
-    this.toggleContextMenu(type, x + width, y + height);
+    const { x, y, width } = this.env.model.getters.getVisibleRect(lastZone);
+    this.toggleContextMenu(type, this.canvasPosition.x + x + width, this.canvasPosition.y + y);
   }
 
   onCellRightClicked(col: HeaderIndex, row: HeaderIndex, { x, y }: DOMCoordinates) {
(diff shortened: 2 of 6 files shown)


src/components/grid/grid.xml
```diff
@@ -25,6 +24,7 @@
         onComposerContentFocused="props.onComposerContentFocused"
         onComposerCellFocused="props.onGridComposerCellFocused"
         focus="props.focusComposer"
+        t-on-contextmenu="onInputContextMenu"
```
hokolomopo, line 27: Yeah this won't work. You're trying to pass a props in a component with a `t-on-*` ... The props onInputContextMenu is always undefined then. 

Maybe write some tests ^^' 

Also is this part really necessary ? The task only mentions click position. I don't have a contextMenu button on my keyboard to check if it works or not in 16.0 and master.

dhrp-odoo: > Yeah this won't work. You're trying to pass a props in a component with a `t-on-*` ... The props onInputContextMenu is always undefined then.

Oops, that was the mistake. I copied a line from above and forgot to change the name of the props 😅
But it's still working? Weird.
 
> Also is this part really necessary ? The task only mentions click position. I don't have a contextMenu button on my keyboard to check if it works or not in 16.0 and master.

Shift+F10 ?

rrahir (changes requested):
I don't think it's necessary to pass the props up to the composer, it does not concern as it does not concern the behavior of the top bar composer. We should avoid this strategy whenever possible not to bloat the composer with too much props.

On another hand, I realized that right now, the native contextmenu is not available from the gridcomposer. I wulld be nice to fix this as well.

From a functional POV, I'd say that as long as the composer is opened(the edition is active) then we let the default component occur. otherwise, we trigger the Menu.

src/components/composer/composer/composer.ts
```diff
@@ -100,6 +100,7 @@ export interface ComposerProps {
   focus: "inactive" | "cellFocus" | "contentFocus";
   onComposerContentFocused: () => void;
   onComposerCellFocused?: (content: String) => void;
+  onInputContextMenu?: (event: MouseEvent) => void;
```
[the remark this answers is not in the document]
rrahir: scratch that, if we want to be able to decide whether or not to use the native contextmenu, the logic should probably be in the same place, which is the composer, not composergrid

dhrp-odoo:
> I realized that right now, the native contextmenu is not available from the gridcomposer. I would be nice to fix this as well.

Ah, you're absolutely right. I've fixed the issue with the native context menu in the grid composer especially when in edition mode :)

src/components/grid/grid.ts
```diff
@@ -328,6 +329,10 @@ export class Grid extends Component<Props, SpreadsheetChildEnv> {
     return this.hoveredCell.col === col && this.hoveredCell.row === row;
   }
 
+  private getGridRect(): Rect {
```
rrahir, line 332: in master , we introduced  the getter`getVisibleRectWithoutHeaders`  which I think does what you are looking for. it might be better to backport that getter, we never know if we might need it for ulterior bugfixes

dhrp-odoo: If I got it right, we need to include headers in the calculation. The discussion is about where to show the context menu when pressing Shift + F10. I'm just trying to replicate the behavior that's already in the main version.

Screenshot from 2024-04-04 11-17-52

rrahir:
@robodoo r+

PR #3746, opened 2024-02-26
[IMP] tables: store `getCellComputedStyle`

hokolomopo:
The actual style/borders of the cell is a combination of of the user defined style, the CF style, and the table style.

Before this commit, at each call to getCellComputedStyle we combined the three the get the actual style of the cell.

Now there is a new plugin `CellComputedStylePlugin` that caches this computed style.

Benchmark:
Tested by scrolling around on a sheet with 10000 rows and 100 cols,
and a bunch of tables/CFs.

`getCellComputedStyle` + `getCellComputedBorder` went from 5\~10% of
an animation frame (of 6ms) to almost nothing (<0.5%). The Garbage
Collection is also lighter, the two functions accounted for 15\~20% of
all the memory allocated when scrolling.

The tradeoff is that the memory usage is higher, since the style is now
stored in the `CellComputedStylePlugin` and not GC'd. We go up to 15%
of the heap occupied by the computed styles when scrolling all the way
down (about the same size as the CellPlugin).

Task: : [3768596](https://www.odoo.com/web#id=3768596&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_feature/cell_computed_style.ts
+++ b/src/plugins/ui_feature/cell_computed_style.ts
@@ -1,0 +1,100 @@
+import { LINK_COLOR } from "../../constants";
+import { isObjectEmptyRecursive, removeFalsyAttributes } from "../../helpers/index";
+import { Command, invalidateCFEvaluationCommands, invalidateEvaluationCommands } from "../../types";
+import { Border, CellPosition, Style, UID } from "../../types/misc";
+import { UIPlugin } from "../ui_plugin";
+import { doesCommandInvalidatesTableStyle } from "./table_style";
+
+export class CellComputedStylePlugin extends UIPlugin {
+  static getters = ["getCellComputedBorder", "getCellComputedStyle"] as const;
+
+  private styles: Record<UID, Record<number, Record<number, Style>>> = {};
+  private borders: Record<UID, Record<number, Record<number, Border | null>>> = {};
+
+  handle(cmd: Command) {
+    if (
+      invalidateEvaluationCommands.has(cmd.type) ||
+      cmd.type === "UPDATE_CELL" ||
+      cmd.type === "EVALUATE_CELLS"
+    ) {
+      this.styles = {};
+      this.borders = {};
+      return;
+    }
+
+    if (doesCommandInvalidatesTableStyle(cmd)) {
+      delete this.styles[cmd.sheetId];
+      delete this.borders[cmd.sheetId];
+      return;
+    }
+
+    if (invalidateCFEvaluationCommands.has(cmd.type)) {
+      this.styles = {};
+      return;
+    }
+  }
+
+  getCellComputedBorder(position: CellPosition): Border | null {
+    const { sheetId, row, col } = position;
+    if (this.borders[sheetId]?.[row]?.[col] !== undefined) {
+      return this.borders[sheetId][row][col];
+    }
+    if (!this.borders[sheetId]) {
+      this.borders[sheetId] = {};
+    }
+    if (!this.borders[sheetId][row]) {
+      this.borders[sheetId][row] = {};
+    }
+    if (!this.borders[sheetId][row][col]) {
+      this.borders[sheetId][row][col] = this.computeCellBorder(position);
+    }
+    return this.borders[sheetId][row][col];
+  }
+
+  getCellComputedStyle(position: CellPosition): Style {
+    const { sheetId, row, col } = position;
+    if (this.styles[sheetId]?.[row]?.[col] !== undefined) {
+      return this.styles[sheetId][row][col];
+    }
+    if (!this.styles[sheetId]) {
+      this.styles[sheetId] = {};
+    }
+    if (!this.styles[sheetId][row]) {
+      this.styles[sheetId][row] = {};
+    }
+    if (!this.styles[sheetId][row][col]) {
+      this.styles[sheetId][row][col] = this.computeCellStyle(position);
+    }
+    return this.styles[sheetId][row][col];
+  }
+
+  private computeCellBorder(position: CellPosition): Border | null {
+    const cellBorder = this.getters.getCellBorder(position) || {};
+    const cellTableBorder = this.getters.getCellTableBorder(position) || {};
+
+    // Use removeFalsyAttributes to avoid overwriting borders with undefined values
+    const border = {
+      ...removeFalsyAttributes(cellTableBorder),
+      ...removeFalsyAttributes(cellBorder),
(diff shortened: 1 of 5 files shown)


src/types/commands.ts
```diff
@@ -119,6 +119,25 @@ export const invalidateEvaluationCommands = new Set<CommandTypes>([
...
+  "UNHIDE_COLUMNS_ROWS",
+  "UNFOLD_HEADER_GROUP",
+  "FOLD_HEADER_GROUP",
+  "FOLD_ALL_HEADER_GROUPS",
+  "UNFOLD_ALL_HEADER_GROUPS",
+  "CREATE_TABLE",
+  "UPDATE_TABLE",
+  "UPDATE_FILTER",
+] as const;
+export const invalidateTableStyleCommandsSet = new Set<CommandTypes>(invalidateTableStyleCommands);
```
anhe-odoo, line 133: Is there a reason to declare this in this way instead of directly declaring the set from the Array (probably a useless comments, but it would keep the same declaration style as the others set)

anhe-odoo: Also, it doesn't need to be exported for the moment, thanks to your helper function :)

src/plugins/ui_feature/cell_computed_style.ts
```diff
@@ -0,0 +1,101 @@
...
+    if (!this.styles[sheetId][row]) {
+      this.styles[sheetId][row] = {};
+    }
+    if (!this.styles[sheetId][row][col]) {
+      this.styles[sheetId][row][col] = this._getCellComputedStyle(position);
+    }
+    return this.styles[sheetId][row][col];
+  }
+
+  private _getCellComputedBorder(position: CellPosition): Border | null {
```
anhe-odoo, line 75: I'm not sure for the `_getCellComputedBorder` (I would use `computeCellBorder`), but it's only my opinion.

src/plugins/ui_feature/cell_computed_style.ts
```diff
@@ -0,0 +1,101 @@
...
+    const cellBorder = this.getters.getCellBorder(position) || {};
+    const cellTableBorder = this.getters.getCellTableBorder(position) || {};
+
+    // Use removeFalsyAttributes to avoid overwriting borders with undefined values
+    const border = { ...cellTableBorder, ...removeFalsyAttributes(cellBorder) };
```
anhe-odoo, line 80: Maybe add a removeFalsyAttributes for `cellTableBorder` to avoid having undefined values in the final border object (just as you did it for the computedStyle below) ?

src/plugins/ui_feature/cell_computed_style.ts
```diff
@@ -0,0 +1,101 @@
...
+      cmd.type === "UPDATE_CELL" ||
+      cmd.type === "EVALUATE_CELLS"
+    ) {
+      this.styles = {};
+      this.borders = {};
+      return;
+    }
+
+    if (doesCommandInvalidatesTableStyle(cmd)) {
+      delete this.styles[cmd.sheetId];
```
anhe-odoo, line 30: Is it really needed to remove ALL the styles and border of a sheet when invalidating a table style ? Or is it some kind of security as we don't know the exact table size ? I understand that the add/remove col/row should invalidate everything, but is also the case for the update of a table/filter ?

anhe-odoo (commented):
Some randoms though, but else LGTM to go to the testing part :)

rrahir (commented):
Could you specify the test conditions you made?
- Was there abundant style set on every cell? combination of table+cf+hardcoded style
- 4% to 0.3% of what?  time spent? Do you have concrete values? (4% of 16ms or 200ms for instance)
- Did you check the memory cost when you loaded lots of cells through multiple sheets?

src/plugins/ui_feature/cell_computed_style.ts
hokolomopo: Hiding/unhiding rows will change the table style (banded rows are updated), so UPDATE_FILTER need to invalidate the style.

And after dynamic tables, we do'nt know the actual dimensions of the table added before the evaluation, so we cannot do much better than invalidating the whole sheet

src/plugins/ui_feature/cell_computed_style.ts
```diff
@@ -0,0 +1,101 @@
...
+
+export class CellComputedStylePlugin extends UIPlugin {
+  static getters = ["getCellComputedBorder", "getCellComputedStyle"] as const;
+
+  private styles: Record<UID, Record<number, Record<number, Style>>> = {};
+  private borders: Record<UID, Record<number, Record<number, Border | null>>> = {};
+
+  handle(cmd: Command) {
+    if (
+      invalidateEvaluationCommands.has(cmd.type) ||
```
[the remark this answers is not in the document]
hokolomopo: Yeah it's in a condition below. It's not on the same `if` because we don't need to invalidate the borders when the CF change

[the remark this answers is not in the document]
rrahir: Sorry I missed it completely

hokolomopo:
@rrahir added more precise benchmarck

rrahir:
I did a (ridiculously) big spreadsheet with 3000*26 cells with table/CF/hardcoded style and duplicated the sheet 10 times.
The heap of each sheet seems to be about 30mb, so it "can" keep growing but requires a ton of style& scrolling without any command to invalidate it to make it. I guess it's reasonable

src/types/commands.ts
```diff
@@ -119,6 +120,25 @@ export const invalidateEvaluationCommands = new Set<CommandTypes>([
...
+  "FOLD_HEADER_GROUP",
+  "FOLD_ALL_HEADER_GROUPS",
+  "UNFOLD_ALL_HEADER_GROUPS",
+  "CREATE_TABLE",
+  "UPDATE_TABLE",
+  "UPDATE_FILTER",
+] as const;
+const invalidateTableStyleCommandsSet = new Set<CommandTypes>(invalidateTableStyleCommands);
+
+export function doesCommandInvalidatesTableStyle(
```
LucasLefevre, line 136: I'd move this out of the `types` folder. There's isn't even a type associated with this function

LucasLefevre (commented):
LGTM, only a minor detail.
I hope the current test coverage is good and covers the cache invalidation

PR #3747, opened 2024-02-26
[REF] compilation_parameters: remove useless parameter

laa-odoo:
## Description:

next to https://github.com/odoo/o-spreadsheet/pull/3742

Since we treat the form of the arguments (matrix
or not) before each sub-formula, we can remove the
code doing the same thing across reference reading.

## review checklist

- [ ] feature is organized in plugin, or UI components
- [ ] support of duplicate sheet (deep copy)
- [ ] in model/core: ranges are Range object, and can be adapted (adaptRanges)
- [ ] in model/UI: ranges are strings (to show the user)
- [ ] undo-able commands (uses this.history.update)
- [ ] multiuser-able commands (has inverse commands and transformations where needed)
- [ ] new/updated/removed commands are documented
- [ ] exportable in excel
- [ ] translations (\_t("qmsdf %s", abc))
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes
- [ ] doc is rebuild (npm run doc)
- [ ] status is correct in Odoo

PR #3750, opened 2024-02-27
[FIX] evaluation: accept 1x1 result array in sub-formula

laa-odoo:
## Description:

Before this commit, when a formula function
expected a simple argument, and it received
an array, we transformed the array into a
simple argument if the array was of size 1x1

However, this process was only done when
reading the references to the ranges.

This commit extends the process when the result
array comes from a sub-formula and no longer
range references.

Task: : [3756474](https://www.odoo.com/web#id=3756474&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/functions/index.ts
+++ b/src/functions/index.ts
@@ -12,6 +12,7 @@ import {
   Matrix,
   ValueAndFormat,
 } from "../types";
+import { CellErrorType, EvaluationError } from "../types/errors";
 import { addMetaInfoFromArg, validateArguments } from "./arguments";
 import { matrixMap } from "./helpers";
 import * as array from "./module_array";
@@ -74,10 +75,36 @@ class FunctionRegistry extends Registry<FunctionDescription> {
     }
     const descr = addMetaInfoFromArg(addDescr);
     validateArguments(descr.args);
-    this.mapping[name] = createComputeFunctionFromDescription(descr);
+    this.mapping[name] = addInputHandling(descr, createComputeFunctionFromDescription(descr));
     super.add(name, descr);
     return this;
   }
+}
+
+function addInputHandling(
+  descr: FunctionDescription,
+  computeFunction: ComputeFunction<Arg, Matrix<ValueAndFormat> | ValueAndFormat>
+) {
+  function computeWithInputHandling(this: EvalContext, ...args: Arg[]) {
+    for (let i = 0; i < args.length; i++) {
+      const argDefinition = descr.args[descr.getArgToFocus(i + 1) - 1];
+      const arg = args[i];
+      if (isMatrix(arg) && !argDefinition.acceptMatrix) {
+        if (arg.length !== 1 || arg[0].length !== 1) {
+          throw new EvaluationError(
+            CellErrorType.GenericError,
+            _t(
+              "Function [[FUNCTION_NAME]] expects the parameter '%s' to be a single value or a single cell reference, not a range.",
+              argDefinition.name
+            )
+          );
+        }
+        args[i] = arg[0][0];
+      }
+    }
+    return computeFunction.apply(this, args);
+  }
+  return computeWithInputHandling;
 }
 
 function createComputeFunctionFromDescription(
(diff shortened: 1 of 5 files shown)


laa-odoo:
Possibility to go further and remove refFn in compilation parameters.
Since we treat the form of the arguments (matrix or not) before each
sub-formula, --> we can remove the code doing the same thing across
reference reading

POC https://github.com/odoo/o-spreadsheet/pull/3747

src/functions/index.ts
```diff
@@ -74,12 +75,43 @@ class FunctionRegistry extends Registry<FunctionDescription> {
...
+        if (arg.length !== 1 || arg[0].length !== 1) {
+          throw new EvaluationError(
+            CellErrorType.GenericError,
+            _t(
+              "Function [[FUNCTION_NAME]] expects the parameter '%s' to be a single value or a single cell reference, not a range.",
+              argDefinition.name
+            )
+          );
+        }
+        checkedArgs.push(arg[0][0]);
```
LucasLefevre, line 104: for maximum performance, it's worth mutating the original args instead of creating a new array, I think
```suggestion
        args[i] = arg[0][0];
```
and remove the else clause

src/functions/index.ts
```diff
@@ -74,12 +75,43 @@ class FunctionRegistry extends Registry<FunctionDescription> {
...
+      } else {
+        checkedArgs.push(arg);
+      }
+    }
+
+    const compute = computeFunction.bind(this);
+    return compute(...checkedArgs);
```
LucasLefevre, line 111: I believe this is faster since it doesn't create a new (binded) function and avoid the array destructuring.
```suggestion
    return computeFunction.apply(this, args);
```

LucasLefevre (commented):
a few performance suggestions.
Did you try measuring if it has a measurable impact?

src/functions/index.ts
LucasLefevre: probably worth using `.apply` in `buildComputeFunctionFromDescription` as well. We should measure the impact

src/functions/index.ts
```diff
@@ -74,12 +75,38 @@ class FunctionRegistry extends Registry<FunctionDescription> {
...
 }
 
+function addInputHandling(
+  descr: FunctionDescription,
+  computeFunction: ComputeFunction<Arg, Matrix<ValueAndFormat> | ValueAndFormat>
+) {
+  return function (this: EvalContext, ...args: Arg[]) {
+    for (let i = 0; i < args.length; i++) {
+      const argDefinition = descr.args[descr.getArgToFocus(i + 1) - 1];
+      const acceptMatrix = argDefinition.type.some((t) => t.startsWith("RANGE"));
```
LucasLefevre, line 91: I can measure a performance hit of ~+4.5% (on RNG's spreadsheet)
maybe we can try to move those two lines to check if the arg accepts a matrix outside of `return function (this: EvalContext, ...` to compute it only once (keep an array of booleans)

src/functions/index.ts
```diff
@@ -74,12 +75,38 @@ class FunctionRegistry extends Registry<FunctionDescription> {
...
     super.add(name, descr);
     return this;
   }
 }
 
+function addInputHandling(
+  descr: FunctionDescription,
+  computeFunction: ComputeFunction<Arg, Matrix<ValueAndFormat> | ValueAndFormat>
+) {
+  return function (this: EvalContext, ...args: Arg[]) {
```
LucasLefevre, line 88: adding a name helps to identify the function when profiling the code
```suggestion
  return function computeWithInputHandling(this: EvalContext, ...args: Arg[]) {
```

src/functions/index.ts
laa-odoo: @LucasLefevre  

to know if arg accept matrix or not, an attribute was added on `argFunctionDecription`.
The attribute is computed when formula is added to the formula registry.

If ok for you, I fixup the last commit

LucasLefevre (approved):
I couldn't measure any performance impact :+1: 
Thanks
robodoo r+

PR #3770, opened 2024-02-28
[IMP] data validation: performance of icons

hokolomopo:
## Description

A sheet with a lot of data validation icons was slow to render, because:

- we loop on all the cells of the data validation, and not only the visible cells
- we created components even for non-visible cells
- we use `getColDimensionsInViewport` which is slow

This commit fixes all of the above. We go from ~10fps to ~45fps when scrolling the sheet.

We could be even faster by not using components and only t-call to templates, but the world isn't ready yet.

Task: : [3773923](https://www.odoo.com/web#id=3773923&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/data_validation_overlay/data_validation_overlay.ts
+++ b/src/components/data_validation_overlay/data_validation_overlay.ts
@@ -9,13 +9,18 @@ export class DataValidationOverlay extends Component<{}, SpreadsheetChildEnv> {
   static components = { GridCellIcon, DataValidationCheckbox, DataValidationListIcon };
 
   get checkBoxCellPositions(): CellPosition[] {
-    return this.env.model.getters.getDataValidationCheckBoxCellPositions();
+    return this.env.model.getters
+      .getVisibleCellPositions()
+      .filter(this.env.model.getters.isCellValidCheckbox);
   }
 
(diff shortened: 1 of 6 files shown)


src/components/data_validation_overlay/data_validation_overlay.ts
```diff
@@ -9,13 +9,25 @@ export class DataValidationOverlay extends Component<{}, SpreadsheetChildEnv> {
...
   get checkBoxCellPositions(): CellPosition[] {
-    return this.env.model.getters.getDataValidationCheckBoxCellPositions();
+    const positions: CellPosition[] = [];
+    for (const position of this.env.model.getters.getVisibleCellPositions()) {
+      if (this.env.model.getters.isCellValidCheckbox(position)) {
+        positions.push(position);
+      }
+    }
+    return positions;
+    return this.env.model.getters
```
anhe-odoo, line 19: C-C-C-C-C-Combo (of return statement)

hokolomopo: :eyes:

PR #3776, opened 2024-02-29
[IMP] tables: set table color in custom colors

hokolomopo:
### [IMP] table: add table colors to custom colors

When inserting a table with a style in the sheet, it could be useful
for the user to be able to paint cells in the same colors as the table
cells.

This commit adds the table colors to the custom colors plugin, so they
are available to the user.

Another idea would have been to replace the color pickers colors by
the table colors. But then we end up with dull and sad colors in the
color picker so we gave up on that.

Also fixed some bugs and strange behaviors in the custom colors plugin:
- we compute and store the custom colors at each finalize, but the
getter `getCustomColors` was re-computing the colors instead of using
the stored ones.
- the colors stored by the plugin weren't loaded at import until a
command modifying the colors was executed.
- the colors were filtered/validated multiple times (once when computing
the colors, and once when adding them to the plugin state).
- the colors were sorted, then added to a POJO, making the sorting useless.

### [IMP] colors: reduce number of custom colors

This commit aims to reduce the number of custom colors during a
normal use of a spreadsheet and in the demo by replacing custom colors
by colors present in the color picker.

The following were changed:
- default scorecard baseline colors
- CF up/down arrows icons
- lot of demo data colors and their equivalent in the demo test xlsx

Task: : [3775936](https://www.odoo.com/web#id=3775936&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/plugins/ui_core_views/custom_colors.ts
```diff
@@ -90,6 +93,17 @@ export class CustomColorsPlugin extends UIPlugin<CustomColorState> {
       case "SET_ZONE_BORDERS":
       case "SET_FORMATTING":
         this.history.update("shouldUpdateColors", true);
+        break;
+      case "START":
+      case "CREATE_TABLE":
+      case "UPDATE_TABLE":
+      case "REMOVE_TABLE":
+        this.history.update("tableColors", this.getTableColors());
```
LucasLefevre, line 101: why have a different strategy for table colors than for other colors ?
table colors: have `this.tableColors` and update it directly
other colors: set `this.shouldUpdateColors` to true and compute the colors

hokolomopo: Because even if it isn't expensive, it seemed a bit dumb to recompute the table colors at each UPDATE_CELL.

That being said, we will need to recompute the table colors at each UPDATE_CELL after dynamic tables are introduced, so I can probably just remove the optimization.

LucasLefevre (approved):
robodoo r+ rebase-ff

PR #3799, opened 2024-03-08
[IMP] import/export table styles in xlsx

hokolomopo:
## [IMP] xlsx_import: use xlsx blob to test the import

Replace all the files in the `__xlsx__/xlsx_demo_data/` folder by an
xlsx file.

Now we can change the xlsx file directly when we want to add tests,
and the diff will be in  a zipped blob rather than spread across
multiple .xml files.

## [IMP] XLSX: Faulty default styles

Due to some mismatch between the style in the default exported construct
and the default values computed for a cell style, we were creating
useless style entries at export.

By setting a style index on a worksheet cell entry, our import code would
then apply this style to the cell even though that same cell, when
exported, had no style whatsoever.

Furthermore, we did not process Excel default color properly.
In Xlsx file, the color is sometimes set specifically to
"FF000000" in the style sheet. However, that color (black) is not
applied as a user-defined color in Excel (i.e. it's importance is lower
than the color applied by a table[^1]). Instead, a user-defined
"black" will be referred to via a theme property[^2].

Since we did not detect this "default color" pattern, we would consider
the attribute `rgb=FF000000` as a user-defined color an ultimately set
it in the plugin structure at import.

This revision detects the "default color" situation and discards the
given color to let the library (and specifically the CellPlugin as that
issue affects the cell font color in particular) use its own default
behaviour.

[^1]: To test this out:
      - Go to Excel and create a sheet on which you put
        a table.
      - Apply a style such that the text turns white
      - On the cells with white text, manually apply an italic format
      - Import the xlsx file inside o-spreadsheet
      ==> the text turned black in o-spreadsheet

[^2]: This situation seems to only concern that specific color and might
encompass a broader scope but I haven't been able to find proper
documentation about this.

Task: : [3789612](https://www.odoo.com/web#id=3789612&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/plugins/ui_stateful/filter_evaluation.ts
```diff
@@ -202,17 +201,12 @@ export class FilterEvaluationPlugin extends UIPlugin {
...
 
-          // In xlsx, filter header should ALWAYS be a string and should be unique in the table
-          const headerPosition = {
-            col: filter.col,
-            row: filter.rangeWithHeaders.zone.top,
-            sheetId,
-          };
-          const headerString = this.getters.getEvaluatedCell(headerPosition).formattedValue;
+          // In xlsx, column header should ALWAYS be a string and should be unique in the table
+          const headerString = this.getters.getEvaluatedCell(position).formattedValue;
```
anhe-odoo, line 205: Why can we now use `position` directly instead of `headerPosition` ? Is it just a previous mistake, or is there a reason why we can now do it ?

hokolomopo: It was useless before, headerPoisiton is the same as position 🙃

rrahir:
chti rebase ^^

hokolomopo:
@rrahir rebased

Rebasing the xlsx blob don't seems that bad (if you have Excel installed on your PC). Just taking the xlsx from master, look at wich of your new tests break, then edit the xlsx to have them work. It's actually probably easier than a rebase needed in the xml files.

I think the xlsx blob is the right direction :)

rrahir:
Excel online /vm are a good enough alternative ^^

src/xlsx/conversion/color_conversion.ts
```diff
@@ -86,3 +86,11 @@ export function hexaToInt(hex: Color) {
   }
   return parseInt(hex.replace("#", ""), 16);
 }
+
+/**
+ * When defining style (fontColor, borderColor for instance)
+ * Excel will specify rgb="FF000000"
+ * In that case, We should not consider this value asuser-defined but
```
LucasLefevre, line 93:
```suggestion
 * In that case, We should not consider this value as user-defined but
```

rrahir:
@robodoo r+

LucasLefevre:
robodoo rebase-ff

PR #3800, opened 2024-03-08
[FIX] evaluation: ensure value to evaluated cell

rrahir:
When the evaluation process fails, we catch the error and assume it is an `EvaluationError` but it's not always the case (i.e., `evaluateLiteral` can also fail in some situations [^1]) and the error.value can be undefined, which breaks the typing of `createEvaluatedFormula`. THis revision adds a default value to avoid this situation.

[^1]: Currently, this can happen if we try to define a link cell in which the label starts with an "=". That issue will be sorted out in another commit.

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_core_views/cell_evaluation/evaluator.ts
+++ b/src/plugins/ui_core_views/cell_evaluation/evaluator.ts
@@ -1,4 +1,5 @@
 import { compile } from "../../../formulas";
+import { implementationErrorMessage } from "../../../functions";
 import { matrixMap } from "../../../functions/helpers";
 import { forEachPositionsInZone, JetSet, lazy, toXC } from "../../../helpers";
 import { createEvaluatedCell, evaluateLiteral } from "../../../helpers/cells";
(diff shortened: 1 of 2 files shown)


src/plugins/ui_core_views/cell_evaluation/evaluator.ts
```diff
@@ -261,7 +261,7 @@ export class Evaluator {
         ? this.computeFormulaCell(cellPosition.sheetId, cell)
         : evaluateLiteral(cell.content, localeFormat);
     } catch (e) {
-      return createEvaluatedCell(e.value, localeFormat, e.message);
+      return createEvaluatedCell(e.value || CellErrorType.GenericError, localeFormat, e.message);
```
VincentSchippefilt, line 264:
```suggestion
      return createEvaluatedCell(e?.value || CellErrorType.GenericError, localeFormat, e?.message || "some default message");
```
it is possible to throw an undefined error as well

VincentSchippefilt:
robodoo r+

PR #3823, opened 2024-03-12
[PERF] evaluation: use custom data structures instead of Map and Set

LucasLefevre:
## Description:

This commit drastically improves the evaluation execution time (~43% faster with the large formula demo dataset)

`Map` and `Set` data structures are slow compared to objects and arrays (especially with `BigInt`, more on that later). Why use `Map` and `Set` then ?

We needed a set data structure to hold which positions should be evaluated. We obviously don't want duplicates (hence the set). However, we can't blindly add a position object `{ sheetId, col, row }` in a `Set` though because the same position could be added twice by adding the same position as two different objects that are not referentially equal.
That's why we came up with a way to encode the position object to a single value that can be added safely to the set without worrying about object references being equal. This encoded value is a `BigInt` (see `PositionBitsEncoder`). That's why we used `Set<BigInt>`. Since we had the position encoded as a single value, we naturally used that value as a key in a `Map` (regular `object` can only have `string` or `number` keys, not `BigInt`). Using this `BigInt` key also avoid having to decode the encoded value, then have a nested `[sheetId][col][row]` objects.

In the evaluation process, there's a lot going on. To the point were lookup in the map is a bottleneck.

This commit replaces the `Set` and `Map` in the evaluation by custom data structure `PositionSet` and `PositionMap<T>` with faster addition and lookup.

In addition to being faster, this makes the evaluation code simpler as we no longer need encoding/decoding everywhere. Those data structures are also easier to debug. With the `BigInt` encoding, it was hard to find back which position is the number `28416511n`.

Some measures
----------------------
*(all measures are averages over 8 runs)*

```js
console.time("eval");
o_spreadsheet.__DEBUG__.model.dispatch("EVALUATE_CELLS")
console.timeEnd("eval");
```
On the large formula demo data set (260k cells): -43%

|        	| Time  	| Allocated memory 	|
|--------	|-------	|------------------	|
| Before 	| 512ms 	| 190MB            	|
| After  	| 291ms 	| 122MB            	|

I also tried implementing `PositionSet` with a `PositionMap<boolean>` instead of `BinaryGrid`. It's ~10% slower and allocates ~20MB more (but it's 120 LoC less).

On RNG's Timesheet spreadsheet: -23%

|        	| Time  	| Allocated memory 	|
|--------	|-------	|------------------	|
| Before 	| 1894ms 	| 580MB            	|
| After  	| 1469ms 	| 460MB            	|

Basic tests on the raw data structures with 1M positions (10k rows and 100 columns)

|   	|                	| add   	| has   	|
|---	|----------------	|-------	|-------	|
| 1 	| `Set<BigInt>`    	| 314ms 	| 262ms 	|
| 2 	| `Set` + encoding 	| 440ms 	|       	|
| 3 	| `PositionSet`    	| 43ms  	| 46ms  	|

|   	|                     	| set   	| get   	|
|---	|---------------------	|-------	|-------	|
| 4 	| `Map<BigInt, number>` 	| 365ms 	| 304ms 	|
| 5 	| `PositionMap<number>` 	| 47ms  	| 51ms  	|

Setup code shared by all benchmarks:
```js
const sheetId = "1";
const postions = [];
for (let i = 0; i < 10000; i++) {
  for (let j = 0; j < 100; j++) {
    postions.push({ sheetId, row: i, col: j });
  }
}
```
1. `Set<BigInt>` 314ms	262ms
```js
const set = new Set();
const length = BigInt(postions.length);
performance.mark("add");
for (let i = 0n; i < length; i++) {
  set.add(i);
}
performance.measure("add", "add").duration);

performance.mark("has");
for (let i = 0n; i < length; i++) {
  set.has(i);
}
performance.measure("has", "has").duration;
```

2. `Set<BigInt> + encoding` 440ms
```js
const encoder = new PositionBitsEncoder();
const set = new Set();
performance.mark("add");
for (const position of postions) {
  set.add(encoder.encode(position));
}
performance.measure("add", "add").duration;
```

3. `PositionSet` 43ms	46ms
```js
const set = new PositionSet({
    [sheetId]: { rows: 10000, cols: 100 },
});
performance.mark("add");
for (const position of postions) {
  set.add(position);
}
performance.measure("add", "add").duration;

performance.mark("has");
let r;
for (const position of postions) {
  r = set.has(position);
}
performance.measure("has", "has").duration;
```

4. `Map<BigInt, number>` 365ms	304ms

```js
const map = new Map();
const length = BigInt(postions.length);
performance.mark("set");
for (let i = 0n; i < length; i++) {
  map.set(i, 4);
}
performance.measure("set", "set").duration;

performance.mark("get");
let r;
for (let i = 0n; i < length; i++) {
  r = map.get(i);
}
performance.measure("get", "get").duration;
```

5. `PositionMap` 47ms	51ms

```js
performance.mark("add");
for (const position of postions) {
  map.set(position, 4);
}
performance.measure("add", "add").duration;

performance.mark("get");
let r;
for (const position of postions) {
  r = map.get(position);
}
performance.measure("get", "get").duration;
```

Task: : [3802246](https://www.odoo.com/web#id=3802246&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/misc.ts
+++ b/src/helpers/misc.ts
@@ -461,22 +461,6 @@ export function isConsecutive(iterable: Iterable<number>): boolean {
   return true;
 }
 
-export class JetSet<T> extends Set<T> {
-  addMany(iterable: Iterable<T>): this {
-    for (const element of iterable) {
-      super.add(element);
-    }
-    return this;
-  }
-  deleteMany(iterable: Iterable<T>): boolean {
-    let wasDeleted = false;
-    for (const element of iterable) {
-      wasDeleted ||= super.delete(element);
-    }
-    return wasDeleted;
-  }
-}
-
 /**
  * Creates a version of the function that's memoized on the value of its first
  * argument, if any.
--- a/src/plugins/ui_core_views/cell_evaluation/binary_grid.ts
+++ b/src/plugins/ui_core_views/cell_evaluation/binary_grid.ts
@@ -1,0 +1,98 @@
+import { CellPosition } from "../../../types";
+
+type Bit = 0 | 1;
+
+/**
+ * Implements a fixed-sized grid or 2D matrix of bits.
+ * based on https://github.com/zandaqo/structurae
+ *
+ * The grid is implemented as a 1D array of 32-bit integers, where each bit represents a cell in the grid.
+ * It follows row-major order, with each row stored consecutively in 32-bit blocks.
+ * Pads the number of columns to the next power of 2 to allow quick lookups with bitwise operations.
+ *
+ * Key terminology:
+ * - bucket: Index of an item in the Uint32Array, a 32-bit integer.
+ * - bitPosition: The position of a bit within the bucket 32-bit integer.
+ */
+export class BinaryGrid extends Uint32Array {
+  private columnOffset = 0;
+  cols = 0;
+  rows = 0;
+
+  /**
+   * Creates a binary grid of specified dimensions.
+   */
+  static create(rows: number, columns: number): BinaryGrid {
+    const columnOffset = log2Ceil(columns);
+    const length = (rows << columnOffset) >> 5;
+    const grid = new this(length + 1);
+    grid.columnOffset = columnOffset;
+    grid.cols = columns;
+    grid.rows = rows;
+    return grid;
+  }
+
+  /**
+   * Returns the bit at given coordinates.
+   */
+  getValue(position: CellPosition): Bit {
+    const [bucket, bitPosition] = this.getCoordinates(position);
+    return ((this[bucket] >> bitPosition) & 1) as Bit;
+  }
+
+  /**
+   * Sets the bit at given coordinates.
+   */
+  setValue(position: CellPosition, value: Bit) {
+    const [bucket, bitPosition] = this.getCoordinates(position);
+    const currentValue = (this[bucket] >> bitPosition) & 1;
+    const hasBeenInserted = currentValue === 0 && value === 1;
+    this[bucket] = (this[bucket] & ~(1 << bitPosition)) | (value << bitPosition);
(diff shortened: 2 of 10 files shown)


src/helpers/position_map.ts
```diff
@@ -0,0 +1,41 @@
+import { CellPosition, UID } from "../types";
+
+export class PositionMap<T> {
```
LucasLefevre, line 3: We could also implement this with an unrolled 1-D array similar to the BinaryGrid.
I don't think it would be much much faster (lookups in objects are fast), but we might save quite a bit of memory since we save all the object keys.
Do we want to go that far?

src/helpers/binary_grid.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  private getCoordinates(position: CellPosition): [bucket: number, position: number] {
+    const { row, col } = position;
+    const index = (row << this.columnOffset) + col;
+    const bucket = index >> 5;
+    return [bucket, index - (bucket << 5)];
+  }
+}
+
+function log2Ceil(value) {
```
LucasLefevre, line 89: This is called only a few times so it won't actually make any difference in the overall picture, even if it's 10x faster
https://github.com/odoo/o-spreadsheet/commit/49c8e011fd57f8485ea8b64b03074d1b230fc692#r139640595

src/helpers/position_map.ts
LucasLefevre: A quick and dirty POC (see commit history) didn't show any meaningful difference :shrug:

VincentSchippefilt:
robodoo r+

PR #3843, opened 2024-03-15
[FIX] TextValueProvider: duplicate values in dropdown list

dhrp-odoo:
## Description:

Previously, using duplicate values in the dropdown list triggered an error when attempting to open the autocomplete dropdown. This was due to the utilization of the value as a key within a foreach loop. The solution involves utilizing the index as the key instead.

Task: : [3768392](https://www.odoo.com/web#id=3768392&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


tests/composer/edition_plugin.test.ts
```diff
@@ -1193,4 +1196,52 @@ describe("edition", () => {
...
   });
+
+  describe("Data validation autocomplete dropdown values", () => {
+    let model: Model;
+    beforeEach(async () => {
+      model = new Model();
+      await mountSpreadsheet({ model });
+    });
+
+    test("Duplicate values will be removed before sending proposals to the autocomplete dropdown in data validation with list", async () => {
```
hokolomopo, line 1207: Move these tests witht he other tests of the autocomplete for data validation (tests/data_validation/data_validation_list.test.ts). And use the `typeInComposer` helper rather than the triggerMouseEvent

src/plugins/ui_stateful/edition.ts
```diff
@@ -755,14 +755,18 @@ export class EditionPlugin extends UIPlugin {
 
     let values: string[];
     if (rule.criterion.type === "isValueInList") {
-      values = rule.criterion.values;
+      values = [...new Set(rule.criterion.values)];
```
hokolomopo, line 758: nitpick: filter duplicates values after the if/else rather than doing it in each case

hokolomopo (commented):
Looks good :+1:

tests/data_validation/data_validation_list_component.test.ts
```diff
@@ -254,6 +254,33 @@ describe("autocomplete in composer", () => {
...
+      displayStyle: "arrow",
+    });
+
+    ({ fixture, parent } = await mountComposerWrapper(model));
+    await typeInComposer("");
+    expect(model.getters.getAutoCompleteDataValidationValues()).toMatchObject(["ok", "hello"]);
+  });
+
+  test("Duplicate values will be removed before sending proposals to the autocomplete dropdown in data validation with list", async () => {
+    addDataValidation(model, "A1", "id", {
```
hokolomopo, line 274: also test that the data validation plugin has the filtered values since you filter them in the plugin now.

hokolomopo (commented):
Small comment then we're good :)

tests/data_validation/data_validation_list_component.test.ts
dhrp-odoo: I've added a test for the plugin, thanks!

hokolomopo (commented):
@robodoo r+

PR #3854, opened 2024-03-18
[IMP] formats: add custom formats to format menu

hokolomopo:
## Description

When a custom format was inserted in the spreadsheet (eg. custom currency), there was no indication for the user that the format was active on the cell, and no way to add it on another cell.

This commit adds the custom formats to the format menu, so the menu item has a "check" when it's used on the active cell, and can be applied to another cell easily.

Task: : [3799480](https://www.odoo.com/web#id=3799480&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/registries/menus/number_format_menu_registry.ts
+++ b/src/registries/menus/number_format_menu_registry.ts
@@ -1,66 +1,134 @@
-import { ActionSpec } from "../../actions/action";
+import { SpreadsheetChildEnv } from "../..";
+import { ActionSpec, createActions } from "../../actions/action";
 import * as ACTION_FORMAT from "../../actions/format_actions";
+import { isDateTimeFormat, memoize } from "../../helpers";
 import { _t } from "../../translation";
-import { MenuItemRegistry } from "../menu_items_registry";
+import { Format } from "../../types";
+import { Registry } from "../registry";
 
-export const numberFormatMenuRegistry = new MenuItemRegistry();
+export const numberFormatMenuRegistry = new Registry<ACTION_FORMAT.NumberFormatActionSpec>();
 
 numberFormatMenuRegistry
   .add("format_number_automatic", {
     ...ACTION_FORMAT.formatNumberAutomatic,
+    id: "format_number_automatic",
     sequence: 10,
   })
   .add("format_number_plain_text", {
     ...ACTION_FORMAT.formatNumberPlainText,
+    id: "format_number_plain_text",
     sequence: 15,
     separator: true,
   })
   .add("format_number_number", {
     ...ACTION_FORMAT.formatNumberNumber,
+    id: "format_number_number",
     sequence: 20,
   })
   .add("format_number_percent", {
     ...ACTION_FORMAT.formatNumberPercent,
+    id: "format_number_percent",
     sequence: 30,
     separator: true,
   })
   .add("format_number_currency", {
     ...ACTION_FORMAT.formatNumberCurrency,
+    id: "format_number_currency",
     sequence: 40,
   })
   .add("format_number_currency_rounded", {
     ...ACTION_FORMAT.formatNumberCurrencyRounded,
+    id: "format_number_currency_rounded",
     sequence: 50,
   })
   .add("format_custom_currency", {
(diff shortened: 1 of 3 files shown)


src/registries/menus/number_format_menu_registry.ts
```diff
@@ -1,9 +1,12 @@
...
+import { ActionSpec, createActions } from "../../actions/action";
 import * as ACTION_FORMAT from "../../actions/format_actions";
+import { isDateTimeFormat, memoize } from "../../helpers";
 import { _t } from "../../translation";
-import { MenuItemRegistry } from "../menu_items_registry";
+import { Format } from "../../types";
+import { Registry } from "../registry";
 
-export const numberFormatMenuRegistry = new MenuItemRegistry();
+export const numberFormatMenuRegistry = new Registry<ACTION_FORMAT.NumberFormatActionSpec>();
```
hokolomopo, line 9: Is this registry useful BTW ? We never extend it, or use it outside of this file. Shoudl this just be an array ?

src/registries/menus/number_format_menu_registry.ts
```diff
@@ -56,11 +59,64 @@ numberFormatMenuRegistry
...
+): ACTION_FORMAT.NumberFormatActionSpec[] {
+  const defaultFormats = new Set(
+    numberFormatMenuRegistry
+      .getAll()
+      .map((f) => (typeof f.format === "function" ? f.format(env) : f.format))
+  );
+
+  const customFormats = new Map<Format, ACTION_FORMAT.NumberFormatActionSpec>();
+  for (const sheetId of env.model.getters.getSheetIds()) {
+    const cells = env.model.getters.getEvaluatedCells(sheetId);
```
LucasLefevre, line 76: It's worth checking the performance on large spreadsheet. It won't be a nice experience if the menu takes 100+ms to open

hokolomopo: I tested a bit on a big spreadsheet, and it feels fine (<100ms).

Actually in my first implementation I stored the custom formats in a plugin that reacted to commands, but then every update cell command became slower 😅

LucasLefevre: I guess we can't have the beurre et l'argent du beurre...

src/registries/menus/number_format_menu_registry.ts
```diff
@@ -1,66 +1,134 @@
...
+    for (const cellId in cells) {
+      const cell = cells[cellId];
+
+      if (cell.format && !customFormats.has(cell.format) && !defaultFormats.has(cell.format)) {
+        const formatType = getNumberFormatType(cell.format);
+        if (formatType !== "number") {
```
LucasLefevre, line 94: the intention would be easier to understand if written the other way around
```suggestion
        if (formatType === "date" || formatType === "currency") {
```

tests/test_helpers/helpers.ts
```diff
@@ -713,7 +714,7 @@ export function getNode(
     if (path.length === 0) {
       return item;
     }
-    items = item.children({} as SpreadsheetChildEnv);
```
LucasLefevre, line 716: :1st_place_medal:

LucasLefevre (approved):
robodoo r+

rrahir:
tests did not pass, the r+ never occurred :/

hokolomopo:
@LucasLefevre rebased

LucasLefevre:
robodoo r+

src/registries/menus/number_format_menu_registry.ts
rrahir: I don't think it's very harmful to keep it as it is for now. We never know what kind of feature we might want to introduce in Odoo

src/registries/menus/number_format_menu_registry.ts
rrahir: on rng spreadsheet, the first hit is 200+ms but it still feels very usable  👍

PR #3856, opened 2024-03-18
[FIX] export: export non-Excel formulas as value

Dj0ulo:
__Current behavior before commit:__
Since the commit [`201aea7`][1], when `formulaCell` is a `FormulaCellWithDependencies`, `content` might be set to raw formula even if it is not readable by Excel.

__Description of the fix:__
Prevent `content` to be set to the raw formula if `isFormula === false`.

__Example of steps to reproduce the issue:__
- Write a number in **A1**
- Write `=FORMAT.LARGE.NUMBER(A1)` in **A2**
- Save as XLSX and open in Excel -> **A2** is not evaluated by Excel

opw-3782676

[1]: https://github.com/odoo/o-spreadsheet/commit/201aea7

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_core_views/cell_evaluation/evaluation_plugin.ts
+++ b/src/plugins/ui_core_views/cell_evaluation/evaluation_plugin.ts
@@ -314,7 +314,7 @@ export class EvaluationPlugin extends UIPlugin {
         ? getItemId<Format>(newFormat, data.formats)
         : exportedCellData.format;
       let content;
-      if (formulaCell instanceof FormulaCellWithDependencies) {
+      if (isFormula && formulaCell instanceof FormulaCellWithDependencies) {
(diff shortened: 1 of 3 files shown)


VincentSchippefilt:
Hello @Dj0ulo, 
I'm looking into why our tests didn't fail. there is something strange happening with some formulas that I need to check. I'll push on this branch the test adaptation :)

VincentSchippefilt:
@Dj0ulo I amended a commit in your branch with a test that shows that formulas with dependencies where exported incorrectly before your fix, and correct after your fix. Thank you for your work :)

VincentSchippefilt:
robodoo r+

Dj0ulo:
Nice, thanks for the review!

PR #3863, opened 2024-03-20
[IMP] stores: add `resetStores` method

hokolomopo:
## Description

Once the `DependencyContainer` was created and stores were injected, there was no way to remove them. It was a problem for the import of xlsx in Demo, because we created a new model and we'd want to inject a new ModelStore and re-create all the stores that had a reference to the old model.

This commit creates a `resetStores` method in the `DependencyContainer` that removes all the stores. Note that references to the old stores are not removed, so we need to destroy all the components that may use store to not end up in an incorrect state.

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/demo/main.js
+++ b/demo/main.js
@@ -189,6 +189,12 @@ class Demo extends Component {
             inputFiles[images[i]] = { imageSrc };
           }
           await this.initiateConnection(inputFiles);
+          stores.resetStores();
+          stores.inject(NotificationStore, {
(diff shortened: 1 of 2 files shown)


LucasLefevre (approved):
maybe find a better solution later, but it's blocking some of us right now.
robodoo r+

VincentSchippefilt:
robodoo retry

PR #3865, opened 2024-03-21
[IMP] tables: allow creation of custom table style

hokolomopo:
## [IMP] tables: allow creation of custom table style

This commit introduces the ability for the user to create custom
table styles. The table styles are now managed in a new `TableStyle`
plugin instead of being based on the constant `TABLE_PRESETS`. The
old table style plugin was renamed to `ComputedTableStyle`.

The option to create a custom table style is available in the table
style popover, that opens a new side panel to create a table style.

## [FIX] tests: prevent chaining of `makeTestEnv`

Since the introduction of stores, chaining `makeTestEnv` is wrong and
could cause hard to understand bugs in the tests.

We cannot chain `makeTestEnv` because the first one will create a new
DependencyContainer and inject the model/notification inside it. Then
the second one cannot inject a new model since a model is already in
the DependencyContainer, and cannot create a new DependencyContainer
since it would lose things injected in the first one.

Now chaining `makeTestEnv` will throw an error.

Also fixed a bug in `makeTestEnv` where `toggleSidePanel` was not
properly initialized.

Task: : [3802101](https://www.odoo.com/web#id=3802101&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/registries/menus/table_style_menu_registry.ts
```diff
@@ -0,0 +1,26 @@
+import { Action, createActions } from "../../actions/action";
+import { _t } from "../../translation";
+import { SpreadsheetChildEnv } from "./../../types/env";
+
+export function createTableStyleContextMenuActions(
```
LucasLefevre, line 5: From a functional POV, there's no hint that this context menu exists and you can delete the custom table from here

LucasLefevre: I didn't know excel so it didn't work for me. But I don't have any good idea either

rrahir:
@robodoo r+

LucasLefevre:
robodoo rebase-ff

PR #3876, opened 2024-03-21
[IMP] BottomBarStatistic: cache the statistics results 

rrahir:
## Description:
Currently, the statistics results for the bottom bar are recomputed at
each call, which occurs everytime we force a full rendering (i.e. when
we dispatch a command). This computaion can be very costy depending on
the size of the current selection.
This commit moves the concerned code into a dedicated store (no need to
be part of the selection plugin) and caches the results, only
invalidating them whenever the selection is updated or if a command
susceptible to affect the statistics.

Task: : [3820568](https://www.odoo.com/web#id=3820568&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/bottom_bar_statistic/aggregate_statistics_store.ts
+++ b/src/components/bottom_bar_statistic/aggregate_statistics_store.ts
@@ -1,0 +1,140 @@
+import { sum } from "../../functions/helper_math";
+import { average, countAny, countNumbers, max, min } from "../../functions/helper_statistical";
+import { Get } from "../../store_engine";
+import { SpreadsheetStore } from "../../stores";
+import { _t } from "../../translation";
+import {
+  CellValueType,
+  Command,
+  EvaluatedCell,
+  Locale,
+  invalidateEvaluationCommands,
+} from "../../types";
+
+export interface StatisticFnResults {
+  [name: string]: number | undefined;
+}
+
+interface SelectionStatisticFunction {
+  name: string;
+  compute: (data: EvaluatedCell[], locale: Locale) => number;
+  types: CellValueType[];
+}
+
+const selectionStatisticFunctions: SelectionStatisticFunction[] = [
+  {
+    name: _t("Sum"),
+    types: [CellValueType.number],
+    compute: (values, locale) => sum([[values]], locale),
+  },
+  {
+    name: _t("Avg"),
+    types: [CellValueType.number],
+    compute: (values, locale) => average([[values]], locale),
+  },
+  {
+    name: _t("Min"),
+    types: [CellValueType.number],
+    compute: (values, locale) => min([[values]], locale),
+  },
+  {
+    name: _t("Max"),
+    types: [CellValueType.number],
+    compute: (values, locale) => max([[values]], locale),
+  },
+  {
+    name: _t("Count"),
+    types: [CellValueType.number, CellValueType.text, CellValueType.boolean, CellValueType.error],
+    compute: (values) => countAny([[values]]),
+  },
+  {
+    name: _t("Count Numbers"),
+    types: [CellValueType.number, CellValueType.text, CellValueType.boolean, CellValueType.error],
+    compute: (values, locale) => countNumbers([[values]], locale),
+  },
+];
+
+export class AggregateStatisticsStore extends SpreadsheetStore {
+  statisticFnResults: StatisticFnResults = this._computeStatisticFnResults();
+  private isDirty = false;
+
+  constructor(get: Get) {
+    super(get);
+    this.model.selection.observe(this, {
+      handleEvent: this.handleEvent.bind(this),
+    });
+  }
+
+  handle(cmd: Command) {
+    if (
+      invalidateEvaluationCommands.has(cmd.type) ||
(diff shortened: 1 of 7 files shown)


src/components/bottom_bar_statistic/aggregate_statistics_store.ts
```diff
@@ -0,0 +1,130 @@
...
+    super(get);
+    this.model.selection.observe(this, {
+      handleEvent: this.handleEvent.bind(this),
+    });
+  }
+
+  handle(cmd: Command) {
+    if (
+      (isCoreCommand(cmd) && cmd.type !== "UPDATE_CELL") ||
+      (cmd.type === "UPDATE_CELL" && "content" in cmd)
```
hokolomopo, line 66: Why all the core commands rather than the commands invalidating the evaluation ? Kinda useless to invalidate the aggregate on an UPATE_LOCALE or ADD_TABLE

src/components/bottom_bar_statistic/aggregate_statistics_store.ts
```diff
@@ -0,0 +1,130 @@
...
+    }
+  }
+
+  private _computeStatisticFnResults(): StatisticFnResults {
+    const sheetId = this.getters.getActiveSheetId();
+    const cells = new Set<EvaluatedCell>();
+
+    const zones = this.getters.getSelectedZones();
+    for (const zone of zones) {
+      for (const { col, row } of positions(zone)) {
```
hokolomopo, line 100: should we try to replace the `for ... of positions(zone)` by double loops on col/rows to avoid allocating thousands of objects ?

src/components/bottom_bar_statistic/aggregate_statistics_store.ts
```diff
@@ -0,0 +1,130 @@
...
+    for (let fn of selectionStatisticFunctions) {
+      // We don't want to display statistical information when there is no interest:
+      // We set the statistical result to undefined if the data handled by the selection
+      // does not match the data handled by the function.
+      // Ex: if there are only texts in the selection, we prefer that the SUM result
+      // be displayed as undefined rather than 0.
+      let fnResult: number | undefined = undefined;
+      const evaluatedCells = [...cells].filter((c) => fn.types.includes(c.type));
+      if (evaluatedCells.length) {
+        fnResult = fn.compute(evaluatedCells, locale);
```
hokolomopo, line 124: we should use lazy values. 99.99% of the time we don't need to compute all the different aggregates

hokolomopo (commented):
It's nice to not have this feature stuffed into the selection plugin 👍

src/components/bottom_bar_statistic/aggregate_statistics_store.ts
rrahir: I guess we could get rid of positions altogether

src/components/bottom_bar_statistic/aggregate_statistics_store.ts
hokolomopo: up

rrahir: Sorry forgot to post my answer. I doubt that we need to do it as it's really not the costy part. I can investigate a bit further but the cost (changing everything to lazy and lose a bit of readability is such t hat if wze don't see relevant changes in perf, I 'd rather keep things as they are.

src/components/bottom_bar_statistic/aggregate_statistics_store.ts
```diff
@@ -0,0 +1,140 @@
...
+            continue; // Skip hidden cells
+          }
+
+          const evaluatedCell = getters.getEvaluatedCell({ sheetId, col, row });
+          if (evaluatedCell.type !== CellValueType.empty) {
+            cells.add(evaluatedCell);
+          }
+        }
+      }
+    }
```
VincentSchippefilt, line 122: would be better with `recomputeZones`, you would have only once each cell so you wouldn't need the set on line 106

src/components/bottom_bar_statistic/aggregate_statistics_store.ts
```diff
@@ -0,0 +1,140 @@
...
+    const locale = getters.getLocale();
+    let statisticFnResults: StatisticFnResults = {};
+    for (let fn of selectionStatisticFunctions) {
+      // We don't want to display statistical information when there is no interest:
+      // We set the statistical result to undefined if the data handled by the selection
+      // does not match the data handled by the function.
+      // Ex: if there are only texts in the selection, we prefer that the SUM result
+      // be displayed as undefined rather than 0.
+      let fnResult: number | undefined = undefined;
+      const evaluatedCells = [...cells].filter((c) => fn.types.includes(c.type));
```
VincentSchippefilt, line 132: the [...cells] could be done once outside the loop, `filter` returns a shallow copy so it can be used multiple times on the same structure

VincentSchippefilt:
robodoo r+

PR #3907, opened 2024-03-25
[FIX] *: Compute max/min on huge arrays

rrahir:
This is an extension of PR #1217. Navigators (or rather javascript engines) have a limit to the number of function arguments [1].

We sometimes have to use Math.max and Math.min functions (which are subject to this limitation) on every cell/column/row which means that we are functionally limited.

E.g. If we take the limitation of Chrome, we could only process 117k rows, which is not such a huge number if we consider the situation of a user importing their data.

This revision introduces wrappers `largeMax` and `largeMin` which take an  1-D array (a single argument) and find the maximum (resp. minimum) present in that list. For arrays with les than a 100K cells, the wrapper will still call Math.max as its implementation is most of the time faster.

[1]: Experimentally, that limit is 117k on Google Chrome and 501k on Firefox.

Task: 3802691

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/misc.ts
+++ b/src/helpers/misc.ts
@@ -274,3 +274,35 @@ export function deepEquals(o1: any, o2: any): boolean {
 
   return true;
 }
+
+/**
+ * Alternative to Math.max that works with large arrays.
+ * Typically useful for arrays bigger than 100k elements.
+ */
+export function largeMax(array: number[]) {
+  let len = array.length;
+
+  if (len < 100_000) return Math.max(...array);
+
+  let max: number = -Infinity;
+  while (len--) {
+    max = array[len] > max ? array[len] : max;
+  }
+  return max;
+}
+
+/**
+ * Alternative to Math.min that works with large arrays.
+ * Typically useful for arrays bigger than 100k elements.
+ */
+export function largeMin(array: number[]) {
+  let len = array.length;
+
+  if (len < 100_000) return Math.min(...array);
+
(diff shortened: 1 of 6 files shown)


src/helpers/misc.ts
```diff
@@ -274,3 +274,35 @@ export function deepEquals(o1: any, o2: any): boolean {
 
   return true;
 }
+
+/**
+ * Alternative to Math.max that works with large arrays.
+ * Typically useful for arrays bigger than 100k elements.
+ */
+export function largeMax(array: any[]) {
```
LucasLefevre, line 282: why `any` ? If we type this correctly, we can remove `Number(...)` below, which inevitably eats a bit of performance.
```suggestion
export function largeMax(array: number[]) {
```

rrahir: I'll move the number out of the loop then ? I wanted to make this wrapper as "close" to number as possible, and it supports strings as well (at the very least)

LucasLefevre: Accepting everything (and n'importe quoi) to get a random result is one of the bad parts of javascript though IMHO. I would not reproduce this (bad) behavior.

With numbers, it's crystal clear what's the result of 'max'. The max of strings however... is the result the last in the alphabetical order? is it the lengthier string?

LucasLefevre (approved):
robodoo r+

PR #3908, opened 2024-03-25
[IMP] functions: add #DIV/0! error

LucasLefevre:
## Description:

Task: : [3826335](https://www.odoo.com/web#id=3826335&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/functions/module_statistical.ts
+++ b/src/functions/module_statistical.ts
@@ -11,7 +11,7 @@ import {
   Maybe,
   isMatrix,
 } from "../types";
-import { EvaluationError, NotAvailableError } from "../types/errors";
+import { CellErrorType, EvaluationError, NotAvailableError } from "../types/errors";
 import { arg } from "./arguments";
 import { assertSameDimensions } from "./helper_assert";
 import { invertMatrix, multiplyMatrices } from "./helper_matrices";
@@ -81,7 +81,8 @@ function covariance(dataY: Arg, dataX: Arg, isSample: boolean): number {
 
   assert(
     () => count !== 0 && (!isSample || count !== 1),
-    _t("Evaluation of function [[FUNCTION_NAME]] caused a divide by zero error.")
+    _t("Evaluation of function [[FUNCTION_NAME]] caused a divide by zero error."),
+    CellErrorType.DivisionByZero
   );
 
   let sumY = 0;
@@ -119,7 +120,8 @@ function variance(args: Arg[], isSample: boolean, textAs0: boolean, locale: Loca
 
   assert(
     () => count !== 0 && (!isSample || count !== 1),
-    _t("Evaluation of function [[FUNCTION_NAME]] caused a divide by zero error.")
+    _t("Evaluation of function [[FUNCTION_NAME]] caused a divide by zero error."),
+    CellErrorType.DivisionByZero
   );
 
   const average = sum / count;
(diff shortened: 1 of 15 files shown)


laa-odoo:
for consistency, I find it strange that this new `CellErrorType` does not have its class which extends `EvaluationError`.
But I understand why.

This makes me say that this is not the right approach. 
I have the impression that with what has been done before,
it appears that each type of error corresponds to a different part of the evaluation process.

In other words, it appears that a type of error cannot be raised on the same scale/file as another error ?

Here it is an error which has the particularity of being a calculation error, --> an error which can naturally be caught by javascript. So I wonder if we should not let the calculations be done naturally, and in case of a javascript error, look if the type of the js error corresponds to a division error and act accordingly

LucasLefevre:
> it appears that a type of error cannot be raised on the same scale/file as another error ?

It may be a fact now, but why should it be something we want to enforce/keep ? Does it bring a value of any kind ?

> I wonder if we should not let the calculations be done naturally, and in case of a javascript error, look if the type of the js error corresponds to a division error and act accordingly

I don't think so because a division by zero is most probably something the developer should handle when implementing the function. Otherwise, it's a programming mistake/oversight and it should be fixed. What you propose would shadow those kind of mistakes.

laa-odoo:
>  It may be a fact now, but why should it be something we want to enforce/keep ? Does it bring a value of any kind ?

I find that doing this allows, in a process as complex as the evaluation, to maintain ease of abstraction.
For example, I find it simple to say that in formula modules, we only have to worry about throwing simple errors (Evaluation error)

> I don't think so because a division by zero is most probably something the developer should handle when implementing the function. Otherwise, it's a programming mistake/oversight and it should be fixed. What you propose would shadow those kind of mistakes.

Maybe I'm wrong but it doesn't seem so at the moment ?
and even if we need to handle division errors in the sense that certain formulas mechanically return division errors, then they can use the mechanics of returning division errors and not throwing division errors ?

laa-odoo:
It's just a thought

LucasLefevre:
I must admit I don't really understand your point. The different kind of evaluation errors is only a matter of display. Using one over the other is just as good IMO. There's nothing to think about in my mind.

src/functions/module_statistical.ts
```diff
@@ -80,7 +81,8 @@ function covariance(dataY: Arg, dataX: Arg, isSample: boolean): number {
 
   assert(
     () => count !== 0 && (!isSample || count !== 1),
-    _t("Evaluation of function [[FUNCTION_NAME]] caused a divide by zero error.")
+    _t("Evaluation of function [[FUNCTION_NAME]] caused a divide by zero error."),
```
hokolomopo, line 84: is the condition "(!isSample || count !== 1)" really a divide by zero error ?

src/functions/module_statistical.ts
```diff
@@ -118,7 +120,8 @@ function variance(args: Arg[], isSample: boolean, textAs0: boolean, locale: Loca
...
```
hokolomopo, line 123: same

src/functions/module_statistical.ts
LucasLefevre: :man_shrugging: not really but Google and/or Excel also behaves that way (I don't remember if I tried one or the other or both)

rrahir:
@robodoo r+

LucasLefevre (commented):
robodoo rebase-ff

PR #3924, opened 2024-03-26
[IMP] BottomBarStatistics: faster computation of stats

rrahir:
## Description:

130 > 48 et 200 à 80 ff

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/components/bottom_bar_statistic/aggregate_statistics_store.ts
```diff
@@ -135,10 +136,10 @@ export class AggregateStatisticsStore extends SpreadsheetStore {
       // does not match the data handled by the function.
       // Ex: if there are only texts in the selection, we prefer that the SUM result
       // be displayed as undefined rather than 0.
-      let fnResult: number | undefined = undefined;
+      let fnResult: Lazy<number> | undefined = undefined;
       const evaluatedCells = getCells(fn.types.join(","));
       if (evaluatedCells.length) {
-        fnResult = fn.compute(evaluatedCells, locale);
+        fnResult = lazy(() => fn.compute(evaluatedCells, locale));
```
LucasLefevre, line 147: we could even make the cell iteration and filter part of the lazy computation
```ts
    const fnResult: Lazy<number| undefined> = lazy(() => {
        const evaluatedCells = getCells(fn.types.join(","));
        if (evaluatedCells.length) {
          return fn.compute(evaluatedCells, locale);
        }
        return undefined;
      });
```

src/components/bottom_bar_statistic/aggregate_statistics_store.ts
```diff
@@ -107,12 +107,17 @@ export class AggregateStatisticsStore extends SpreadsheetStore {
     const sheetId = getters.getActiveSheetId();
     const cells = new Set<EvaluatedCell>();
 
+    const isColHidden = memoize((col: number) => getters.isColHidden(sheetId, col));
+    const isRowHidden = memoize((row: number) => getters.isRowHidden(sheetId, row));
```
LucasLefevre, line 111: does it really makes a significant difference ?
I believe the benchmark in the commit message is not the correct one. It's suspiciously the same as the previous commit

src/components/bottom_bar_statistic/aggregate_statistics_store.ts
rrahir: no we cannot. We check the presence of fnResult in the final object to know if we should display the statistics button or not. So we'd end up evaluating everything all the time

src/components/bottom_bar_statistic/aggregate_statistics_store.ts
hokolomopo: And if this getter really cause performance issue (which I'm not sure how since it really does nothing), shouldn't it be memoized inside the plugin rather than here ? That way every call to this getter in other plugin/components is faster

rrahir: I don't know what to say. Don't believe me? then try it for yourself. 
The reason I haven't set it everywhere right now is because I don't  want to have to invalidate the cache all the time and it's pretty much the only place where it's heavily used.

rrahir: Sorry I'm a moron and can't make proper commit messages apparemently. The gain was around 30° ms. I did another round of benchmark and will update the values accordingly

VincentSchippefilt: The time is mostly spent in isColFolder and isRowFolder (of header_grouping.ts). 
I think it would be wise to store on this plugin the list of cols and rows that are folded by sheet, so these getters becomes immediate.
image

pro-odoo:
https://github.com/odoo/o-spreadsheet/pull/3876

PR #3931, opened 2024-03-26
[FIX] Composer: pressing enter in the link editor

dhrp-odoo:
## Description:

Previously, pressing Enter in the link editor would inadvertently open the grid composer in edit mode, which was not the expected behavior. This PR addresses the issue by adding a check in the onInput method of the composer component.

Task: : [3796114](https://www.odoo.com/web#id=TASK_ID&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/link/link_editor/link_editor.ts
+++ b/src/components/link/link_editor/link_editor.ts
@@ -178,6 +178,7 @@ export class LinkEditor extends Component<LinkEditorProps, SpreadsheetChildEnv> 
         if (this.state.link.url) {
           this.save();
         }
+        ev.preventDefault();
         break;
       case "Escape":
         this.cancel();
--- a/tests/components/composer.test.ts
+++ b/tests/components/composer.test.ts
@@ -627,6 +627,19 @@ describe("composer", () => {
     expect(cell.link.url).toBe("http://odoo.com");
   });
 
+  test("Pressing Enter while editing a label does not open grid composer", async () => {
+    setCellContent(model, "A1", "[label](url.com)");
+    await simulateClick(".o-topbar-menu[data-id='insert']");
+    await simulateClick(".o-menu-item[data-name='insert_link']");
+    const editor = fixture.querySelector(".o-link-editor");
+    expect(editor).toBeTruthy();
+
+    editor!.querySelectorAll("input")[0].focus();
+    await keyDown("Enter");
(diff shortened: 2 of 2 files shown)


src/components/composer/composer/composer.ts
```diff
@@ -364,6 +364,9 @@ export class Composer extends Component<ComposerProps, SpreadsheetChildEnv> {
       const el = this.composerRef.el! as HTMLInputElement;
       content = el.childNodes.length ? el.textContent! : "";
     }
+    if (!content && this.props.focus === "inactive") {
```
hokolomopo, line 367: This probably works, but I think we should focus on the underlying cause of the issue: why the hell does the composer receive an input event when pressing enter on the link editor ?

rrahir: I think it's kind of the same beahaviour with the click event. [When you mousedown and and mouseup on differnet targets, the click event is ](https://developer.mozilla.org/en-US/docs/Web/API/Element/click_event#:~:text=If%20the%20button%20is%20pressed%20on%20one%20element%20and%20the%20pointer%20is%20moved%20outside%20the%20element%20before%20the%20button%20is%20released%2C%20the%20event%20is%20fired%20on%20the%20most%20specific%20ancestor%20element%20that%20contained%20both%20elements.)
Worth looking into it at least

dhrp-odoo: After digging into it, I noticed that when we press enter in the link editor, it triggers the input event in the composer. According to the documentation, this doesn't quite match how the click event behaves.

I tried adding an onInput listener higher up in the hierarchy, like a spreadsheet component, and ran into the same problem. EventPhase tells us that it is happening in the target phase, not during bubbling or capturing.

I am unsure if this is the default behavior or not. Using PreventDefault stops it, but it's likely to cause other problems.

rrahir: I'd argue taht preventDefault is exactly what you are looking for. You want to stop the browser chain of events since it makes no sense to keep the "keyboard input event" behaviour once we closed (and destroyed) the input .

rrahir:
@robodoo r+

PR #3939, opened 2024-03-29
[IMP] charts: preserve more options when changing types

LucasLefevre:
## Description:

Changing the type of a chart, many options are currently lost or reset,
such as "cumulative", "stacked", "Use row X as headers", etc.

With this commit, those options are preserved when changing the chart
type. The design options are still reset for now. It's less important because
the user is only likely to change the design once the chart definition is
properly setup and it won't change anymore.

Task: [3833884](https://www.odoo.com/web#id=3833884&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/figures/charts/bar_chart.ts
+++ b/src/helpers/figures/charts/bar_chart.ts
@@ -96,28 +96,26 @@ export class BarChart extends AbstractChart {
     return {
       background: context.background,
       dataSets: context.range ? context.range : [],
-      dataSetsHaveTitle: false,
-      stacked: false,
+      dataSetsHaveTitle: context.dataSetsHaveTitle ?? false,
+      stacked: context.stacked ?? false,
       aggregated: context.aggregated ?? false,
-      legendPosition: "top",
+      legendPosition: context.legendPosition ?? "top",
       title: context.title || "",
       type: "bar",
-      verticalAxisPosition: "left",
+      verticalAxisPosition: context.verticalAxisPosition ?? "left",
       labelRange: context.auxiliaryRange || undefined,
     };
   }
 
   getContextCreation(): ChartCreationContext {
     return {
-      background: this.background,
-      title: this.title,
+      ...this,
       range: this.dataSets.map((ds: DataSet) =>
         this.getters.getRangeString(ds.dataRange, this.sheetId)
       ),
       auxiliaryRange: this.labelRange
         ? this.getters.getRangeString(this.labelRange, this.sheetId)
         : undefined,
-      aggregated: this.aggregated,
     };
   }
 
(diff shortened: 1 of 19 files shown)


tests/figures/chart/charts_component.test.ts
LucasLefevre, line 1: I must admit I don't know what exactly to test. Testing every change combination for each type and each option is not realistic.
I tested one case in the goal of testing the business logic in the store. The rest is more like "configuration" of each chart type

hokolomopo: Maybe with a big ass `ChartCreationContext` with everything, and for each type of chart test the method `getDefinitionFromContextCreation()` ? + a store-specific test. Not sure. 

I'd at least a test specific to `Use row X as header` since that was the goal of the task

src/helpers/figures/charts/bar_chart.ts
```diff
@@ -96,8 +96,8 @@ export class BarChart extends AbstractChart {
     return {
```
hokolomopo, line 96: > The design options are still reset for now

This comment of the PR/commit is kinda wrong. We do preserve the background color for example. There's really no good reason not to also keep the other `design` options ? It's not more expansive to save them than the `config` options. I'd like to keep my axis to the right when switching from a line chart to a bar chart

hokolomopo (commented):
Hello there 👋

tests/figures/chart/bar_chart_plugin.test.ts
LucasLefevre, line 1: Note: I created a specific file for each chart type because I'd like to go away from the all mighty gigantic mixed-up file which is `chart_plugin.test.ts`.

tests/figures/chart/charts_component.test.ts
```diff
@@ -1168,6 +1163,21 @@ describe("charts", () => {
         expect(checkbox.checked).toBe(true);
       }
     });
+
+    test("dataSetsHaveTitle value is kept when changing to a chart without aggregate option then back again", async () => {
```
hokolomopo, line 1167: I don't think the testname is accurate, I don't see why the `aggregate` option matters here :sweat_smile:

hokolomopo (commented):
Small comment then it's good for  (need a small rebase tho)

tests/figures/chart/charts_component.test.ts
rrahir: I think it refers to the test just above which specifically tests the propagation of "aggregated".

rrahir (changes requested):
I rebased the PR but Adrien is right, we should also account for Waterfall charts :/
I'm pretty sure we'll forget to do this modification the next time we add a chart type ...

pro-odoo:
> I rebased the PR but Adrien is right, we should also account for Waterfall charts :/ I'm pretty sure we'll forget to do this modification the next time we add a chart type ...

The way to add a chart is not a big copy/paste ? So this will be included in the copy/paste and so we should not forget it 🪄

pro-odoo:
robodoo rebase-ff r+

PR #3941, opened 2024-03-29
[FIX] Composer: too long auto-complete dropdown

dhrp-odoo:
## Description:

Previously, when there were too many proposals in the auto-complete dropdown, it extended beyond the screen.

This PR addresses the problem by limiting the dropdown's height and making it scrollable.

Task: : [3838954](https://www.odoo.com/web#id=3838954&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/composer/composer/composer.ts
+++ b/src/components/composer/composer/composer.ts
@@ -196,6 +196,8 @@ export class Composer extends Component<ComposerProps, SpreadsheetChildEnv> {
         // render left
         assistantStyle.right = `0px`;
       }
+    } else if (this.props.delimitation) {
+      assistantStyle["max-height"] = `${this.props.delimitation.height}px`;
     }
     return cssPropertiesToCss(assistantStyle);
   }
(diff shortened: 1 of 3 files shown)


src/components/composer/composer/composer.ts
```diff
@@ -197,7 +200,9 @@ export class Composer extends Component<ComposerProps, SpreadsheetChildEnv> {
         assistantStyle.right = `0px`;
       }
     }
-    return cssPropertiesToCss(assistantStyle);
+
+    const assistantCss: string = cssPropertiesToCss(assistantStyle);
+    return assistantCss + this.props.composerAssistantStyle;
```
hokolomopo, line 205: I didn't read the composer code in much details so I might be wrong, but can't the max-height in topBar composer be acheived with the existing props (`props.delimitation`, `props.rect`) ? The gridComposer seems to only use them and be fine

dhrp-odoo: Currently, the top bar doesn't send these props, but I believe it should send `demarcation'. I need to include this send in the props and set the max height value here, as that is the exact value I was trying to get before.

LucasLefevre (approved):
robodoo r+

PR #3944, opened 2024-03-29
[IMP] charts: introduce waterfall chart

hokolomopo:
## Description:

This commit introduces Waterfall charts.

Implementation-wise, Waterfall charts are created with ChartJS floating
bars, instead of giving only the ending value of the bar we can give
ChartJS pairs (start, end) to create the floating bar.

The commit also contains a new ChartJS plugin to draw lines between the
floating bars (`waterfallLinesPlugin`).

Task: : [3834422](https://www.odoo.com/web#id=3834422&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/components/figures/chart/chartJs/chartjs.ts
```diff
@@ -60,10 +64,7 @@ export class ChartJsComponent extends Component<Props, SpreadsheetChildEnv> {
     } else {
       this.chart!.data.datasets = [];
     }
-    this.chart!.config.options!.plugins!.tooltip = chartData.options!.plugins!.tooltip;
-    this.chart!.config.options!.plugins!.legend = chartData.options!.plugins!.legend;
-    this.chart!.config.options!.scales = chartData.options?.scales;
-    // ?
+    this.chart!.config.options = chartData.options;
```
anhe-odoo, line 67: Can't we have something we would like to keep (not overwrite) in the config.options ? Shouldn't we do something like
```js
this.chart!.config.options = {...this.chart!.config.options, ...chartData.options}
```
to be sure ?

src/components/figures/chart/chartJs/chartjs_waterfall_plugin.ts
```diff
@@ -0,0 +1,70 @@
...
+      const nextBarRect = getBarElementRect(nextBar);
+      const rawBarValues = bar.$context.raw;
+      const value = rawBarValues[1] - rawBarValues[0];
+
+      const lineY = Math.round(value < 0 ? rect.bottom - 1 : rect.top);
+      const lineStart = Math.round(rect.right);
+      const lineEnd = Math.round(nextBarRect.left);
+
+      ctx.strokeStyle = "#999";
+      ctx.setLineDash([3, 2]);
```
anhe-odoo, line 45: Can't this be put outside of the while loop ? :)

src/components/figures/chart/chartJs/chartjs_waterfall_plugin.ts
```diff
@@ -0,0 +1,70 @@
...
+
+declare module "chart.js" {
+  interface PluginOptionsByType<TType extends ChartType> {
+    waterfallLinesPlugin?: WaterfallPluginOptions;
+  }
+}
+
+export const waterfallLinesPlugin: Plugin = {
+  id: "waterfallLinesPlugin",
+  beforeDraw(chart, args, options: WaterfallPluginOptions) {
```
anhe-odoo, line 15: Maybe add a comment here to explain that this part is only for drawing the connectors ? :)

src/components/figures/chart/chartJs/chartjs_waterfall_plugin.ts
```diff
@@ -0,0 +1,70 @@
...
+    const ctx = chart.ctx;
+    ctx.save();
+    const startIndex = drawData.findIndex((bar: any) => bar.height !== 0);
+    if (startIndex === -1) {
+      return;
+    }
+    let bar = drawData[startIndex];
+    let nextBar = getNextNonEmptyBar(drawData, startIndex);
+
+    while (nextBar) {
```
anhe-odoo, line 34: I'm not sure it would be a better approach, but can't we do this with something like : 
```js
for (const [i, nextBar] of drawData.entries()) {
  if (i === 0 || nextBar.height === 0) {
    continue;
  }
  ...
}
```
 ?
It will avoid using a while loop with `drawData.find(...)` each time

src/components/side_panel/chart/index.ts
```diff
@@ -60,4 +61,8 @@ chartSidePanelComponentRegistry
   .add("scorecard", {
     configuration: ScorecardChartConfigPanel,
     design: ScorecardChartDesignPanel,
+  })
+  .add("waterfall", {
+    configuration: LineBarPieConfigPanel,
```
anhe-odoo, line 66: Maybe we can rename the LineBarPieConfigPanel into something else (like `GeneralConfigPanel`) as it's starting to be used for more and more chart

src/components/side_panel/chart/line_bar_pie_panel/design_panel.ts
```diff
@@ -11,13 +12,21 @@ import { ChartTitle } from "../building_blocks/title/title";
 
 interface Props {
   figureId: UID;
-  definition: LineChartDefinition | BarChartDefinition | PieChartDefinition;
+  definition:
```
anhe-odoo, line 15: Maybe we can defined a main type like 
```js
const type AllChartDefinition = LineChartDefinition | BarChartDefinition | ...
```
to reuse it in the following lines and avoid having to change each line at hand each time ?

src/helpers/figures/charts/chart_common.ts
```diff
@@ -285,6 +286,7 @@ export function transformChartDefinitionWithDataSetsWithZone<
     | PieChartDefinition
     | ScatterChartDefinition
     | ComboChartDefinition
+    | WaterfallChartDefinition
```
anhe-odoo, line 289: Same question here for a global definition type ?

tests/test_helpers/commands_helpers.ts
```diff
@@ -155,6 +157,7 @@ export function createChart(
     id,
     sheetId,
     definition: {
+      ...data,
```
anhe-odoo, line 160: :+1:

src/registries/chart_types.ts
```diff
@@ -184,6 +189,22 @@ chartRegistry.add("scatter", {
...
+    new WaterfallChart(definition as WaterfallChartDefinition, sheetId, getters),
+  getChartRuntime: createWaterfallChartRuntime,
+  validateChartDefinition: (validator, definition: WaterfallChartDefinition) =>
+    WaterfallChart.validateChartDefinition(validator, definition),
+  transformDefinition: (
+    definition: WaterfallChartDefinition,
+    executed: AddColumnsRowsCommand | RemoveColumnsRowsCommand
+  ) => WaterfallChart.transformDefinition(definition, executed),
+  getChartDefinitionFromContextCreation: (context: ChartCreationContext) =>
+    WaterfallChart.getDefinitionFromContextCreation(context),
```
anhe-odoo, line 204: More like a question, but can't we write it like
```js
validateChartDefinition: WaterfallChart.validateChartDefinition,
transformDefinition: WaterfallChart.transformDefinition,
getChartDefinitionFromContextCreation: WaterfallChart.getDefinitionFromContextCreation,
```
to avoid using lamda ?

src/helpers/figures/charts/waterfall_chart.ts
```diff
@@ -0,0 +1,357 @@
...
+            text: _t("Subtotals"),
+            fontColor,
+            fillStyle: subTotalColor,
+          });
+        }
+        return legendValues;
+      },
+    },
+  };
+  if ((!chart.labelRange && chart.dataSets.length === 1) || chart.legendPosition === "none") {
```
anhe-odoo, line 230: I'm not sure about this condition. Unlike other charts, the legend should be visible here even if we have one dataset and no label, no ? To have the "negative", "positive", ... display

src/helpers/figures/charts/waterfall_chart.ts
```diff
@@ -0,0 +1,357 @@
...
+  const datasetValues: Array<[number, number]> = [];
+  const dataset: ChartDataset = {
+    label: "",
+    data: datasetValues,
+    backgroundColor,
+  };
+  const labelsWithSubTotals: string[] = [];
+  let lastValue = 0;
+  for (const dataSetsValue of dataSetsValues) {
+    for (let i = 0; i < dataSetsValue.data.length; i++) {
```
anhe-odoo, line 334: `for (const [i, data] of dataSetsValue.entries()) {` ? Or should we avoid this kind of stuff ?

anhe-odoo (commented):
Just some questions/small comments :)

src/components/figures/chart/chartJs/chartjs.ts
hokolomopo: Is that really something we want ? I feel like we want the options in the actual ChartJS chart to exactly match the options in the chart Runtime, and not keep leftover options from previous runtimes

src/registries/chart_types.ts
hokolomopo: Well yes, that's way shorter. I just copy/pasted it from the other chart registry entries. Not sure why we weren't doint it for the others 🤔

src/helpers/figures/charts/waterfall_chart.ts
hokolomopo: We should avoid using `entires()`, because it allocates new arrays. Here if we create a chart on a whole column, that's like 10K arrays allocated for nothing

src/helpers/figures/charts/waterfall_chart.ts
hokolomopo: You're right :)

src/components/side_panel/chart/line_bar_pie_panel/design_panel.ts
anhe-odoo: Not sure if it's a better idea, but we can also use something like `ChartWithAxisDefinition` ?

src/components/figures/chart/chartJs/chartjs.ts
rrahir: maybe @LucasLefevre will have an option on this but I think it was more about changing the least options possible?

LucasLefevre: I don't know why we only changed specifically tooltip, legend and scales rather then reset all the options.
Overriding the entire options seem good.
As Adrien said, the options from the runtime should be correct and complete, we don't want/need to preserve the ones chart.js has

src/types/chart/waterfall_chart.ts
```diff
@@ -0,0 +1,26 @@
...
+  readonly dataSetsHaveTitle: boolean;
+  readonly labelRange?: string;
+  readonly title: string;
+  readonly background?: Color;
+  readonly verticalAxisPosition: VerticalAxisPosition;
+  readonly legendPosition: LegendPosition;
+  readonly aggregated?: boolean;
+  readonly showSubTotals: boolean;
+  readonly showConnectorLines: boolean;
+  readonly firstValueAsSubtotal?: boolean;
```
LucasLefevre, line 17: :heart: thanks :)

rrahir:
@robodoo r+

LucasLefevre:
robodoo rebase-ff

PR #3946, opened 2024-03-29
[FIX]: Filters: Do not export 1-row filters to xlsx files

rrahir:
When exporting the data to Excel format,  the filters are processed as if they were tables but unfortunately, Excel does not tables that span over a single row.

As the very concept of a single row filter is not useful to the end-user, we just skip it when exporting the data to the xlsx format.

Task: 3839556

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/core/filters.ts
+++ b/src/plugins/core/filters.ts
@@ -353,6 +353,16 @@ export class FiltersPlugin extends CorePlugin<FiltersState> implements FiltersSt
   }
 
   exportForExcel(data: ExcelWorkbookData) {
-    this.export(data);
+    for (const sheet of data.sheets) {
+      for (const filterTable of this.getFilterTables(sheet.id)) {
+        if (zoneToDimension(filterTable.zone).height === 1) {
+          continue;
+        }
+        sheet.filterTables.push({
+          range: zoneToXc(filterTable.zone),
+          filters: [],
+        });
+      }
+    }
   }
 }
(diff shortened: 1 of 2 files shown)


src/plugins/core/filters.ts
```diff
@@ -353,6 +353,16 @@ export class FiltersPlugin extends CorePlugin<FiltersState> implements FiltersSt
...
   exportForExcel(data: ExcelWorkbookData) {
-    this.export(data);
+    for (const sheet of data.sheets) {
+      for (const filterTable of this.getFilterTables(sheet.id)) {
+        if (zoneToDimension(filterTable.zone).height === 1) {
+          continue;
+        }
+        sheet.filterTables.push({
+          range: zoneToXc(filterTable.zone),
+          filters: [],
```
LucasLefevre, line 363: The code in `export` (that was used before this commit) doesn't export `filters: [],`. Any reason to export that here ? Is it necessary ?

src/model.ts
```diff
@@ -537,7 +537,6 @@ export class Model extends EventBus<any> implements CommandDispatcher {
       }
     }
     data = deepCopy(data);
-
```
LucasLefevre, line 540: useless diff in `model.ts`

LucasLefevre (commented):
only a small question

src/plugins/core/filters.ts
rrahir: it's a typing issue. we were calling `export` which expects a `WorkbookData`, so the absence of filter was not noticed. I guess typescript cannot find this kind of issue because we modify in place, with no return type.

LucasLefevre (approved):
robodoo r+

PR #3947, opened 2024-03-29
[FIX] data menu: Auto-select adjacent cells on filter menu

rrahir:
This is a followup of PR #1970. There are two entry points to insert a filter from the interface and only one of them was amended.

Task: 3839869

## Description:

Task: : [3839869](https://www.odoo.com/web#id=3839869&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/actions/menu_items_actions.ts
+++ b/src/actions/menu_items_actions.ts
@@ -737,9 +737,12 @@ export const INSERT_LINK_NAME = (env: SpreadsheetChildEnv) => {
 //------------------------------------------------------------------------------
 
 export const FILTERS_CREATE_FILTER_TABLE = (env: SpreadsheetChildEnv) => {
+  const zones = env.model.getters.getSelectedZones();
+  if (zones.length === 1 && getZoneArea(zones[0]) === 1) {
+    env.model.selection.selectTableAroundSelection();
+  }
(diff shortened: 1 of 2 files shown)


src/actions/menu_items_actions.ts
```diff
@@ -737,6 +737,7 @@ export const INSERT_LINK_NAME = (env: SpreadsheetChildEnv) => {
 //------------------------------------------------------------------------------
 
 export const FILTERS_CREATE_FILTER_TABLE = (env: SpreadsheetChildEnv) => {
+  env.model.selection.selectTableAroundSelection();
```
hokolomopo, line 740: I don't know we want to change the behaviour in 16.3 (probably not?), but in master the behaviour is that we only do `selectTableAroundSelection` if there is a single cell selected. Otherwise it's impossible to create a filter on half the columns of a range for example

rrahir: my bad, i'll try to go with the current master behaviour

LucasLefevre (commented):
robodoo r+

PR #3958, opened 2024-04-02
[REF] autofill: simplify automatic autofill

LucasLefevre:
## Description:

Before commit e8f8fa321 an evaluated cell with the type "empty" could be empty or have a formula whose result is an empty string.

After the mention commit, only really empty cells (no content) have the evaluated type "empty". Empty string result is now mapped to the "text" type.

This change allow to slightly simplify the autofill code as we no longer needs to check the cell content or any potential array formula that would spread on the value.

The code is also now faster since checking `cell.content` can be expensice since it re-builds the entire formula string from the tokens and the dependencies.

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_feature/autofill.ts
+++ b/src/plugins/ui_feature/autofill.ts
@@ -298,10 +298,7 @@ export class AutofillPlugin extends UIPlugin {
 
     if (col > 0) {
       let leftPosition = { sheetId, col: col - 1, row };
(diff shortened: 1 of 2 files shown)


rrahir:
@robodoo r+

PR #3961, opened 2024-04-02
[PERF] helpers: faster default cell height

LucasLefevre:
## Description:

`cell.content` can be expensive for formulas as the content is a computed property which rebuilds the entire formula string based on the tokens and the dependency ranges.

With this commit, we first check if the cell is a formula, in which case we know it has a content.

When loading the Timesheet report on odoo.com, the total time spent in `getDefaultCellHeight` goes from ~280ms (~2.5%) to less than 5ms (0.0%)

Task: : [3845472](https://www.odoo.com/web#id=3845472&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/text_helper.ts
+++ b/src/helpers/text_helper.ts
@@ -22,7 +22,9 @@ export function getDefaultCellHeight(
   cell: Cell | undefined,
   colSize: number
 ) {
-  if (!cell || !cell.content) return DEFAULT_CELL_HEIGHT;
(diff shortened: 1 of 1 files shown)


pro-odoo (approved):
robodoo r+

PR #4024, opened 2024-04-09
[IMP] Chart: Reference cell as title in chart

dhrp-odoo:
## Description:

Now users can select a cell to set the chart title. This works for any chart type.

Task: : [3693157](https://www.odoo.com/web#id=3693157&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/components/side_panel/chart/building_blocks/title/title.ts
```diff
@@ -1,18 +1,60 @@
-import { Component } from "@odoo/owl";
```
hokolomopo, line 1: While you're modifing this file, could you rename it to `chart_title.ts` ? Also rename the .xml and the containing folder. In a separate commit preferably, so we don't haev issues with git losing the file history :)

src/components/side_panel/chart/bar_chart/bar_chart_design_panel.xml
```diff
@@ -12,7 +12,14 @@
...
       </Section>
-      <ChartTitle title="title" update.bind="updateTitle"/>
+      <ChartTitle
+        title="title"
+        update.bind="updateTitle"
+        isTitleCellReferenced="isTitleCellReferenced"
+        updateIsTitleCellReferenced.bind="updateIsTitleCellReferenced"
+        titleCellReference="titleCellReference"
+        updateTitleCellReference.bind="updateTitleCellReference"
+      />
```
hokolomopo, line 22: That's ... a lot or props. I'm sure those are necessary, but it doesn't really fir into the concept of "Generic & easily reusable building block component".

How about just giving ChartTitle the chartId as props and letting the component handle everything ? 

Another option would be to give ChartTitle  the same props as the DesignPanel (chartDefinition/updateChart/canUpdateChart). I think I prefer the first option, so ChartTitle is very easy to use with just one props needed, but both sounds fine.

src/components/side_panel/chart/building_blocks/title/title.ts
```diff
@@ -1,18 +1,60 @@
...
 
   updateTitle(ev: InputEvent) {
     this.props.update((ev.target as HTMLInputElement).value);
   }
+
+  handleSelectionChange(ranges: string[]) {
+    this.state.dataRange = ranges[0];
+  }
+
+  updateDataRange() {
```
hokolomopo, line 44:
```suggestion
  updateTitleReference() {
```
of updateTitleRangs, or something like that. I don't think the name "data" is fine for a title :)

src/types/chart/common_bar_combo.ts
```diff
@@ -6,6 +6,8 @@ export interface ComboBarChartDefinition {
   readonly dataSetsHaveTitle: boolean;
   readonly labelRange?: string;
   readonly title: string;
+  readonly isTitleCellReferenced?: boolean;
```
hokolomopo, line 9: nitpick: I think i'd rather have something like `titleType: "reference" | "rawValue"`

src/types/chart/common_bar_combo.ts
```diff
@@ -6,6 +6,8 @@ export interface ComboBarChartDefinition {
...
+  readonly titleCellReference?: string;
```
hokolomopo, line 10: I think there's a design problem here. I don't think we should have both `title` and `titleCellReference`. In the best case that's duplicated information, in the worth case it can very easily lead to issue beacuse of unhandled cases (like in you code).

There's a big missing part in your code. You keep using chart.title to generate the runtime of the chart. But chart.title is equal to the referencing cell content only when updating the title in the panel. Afterwards, when the cell's content changes, you don't update the title in the runtime with the new value.

I think this would have been easier to spot if we only had a single option `title` , which would be either a reference of a raw string.

src/helpers/figures/charts/abstract_chart.ts
```diff
@@ -21,11 +21,15 @@ import { Validator } from "../../../types/validator";
...
   readonly title: string;
+  readonly isTitleCellReferenced?: boolean;
+  readonly titleCellReference?: string;
   abstract readonly type: ChartType;
   protected readonly getters: CoreGetters;
 
   constructor(definition: ChartDefinition, sheetId: UID, getters: CoreGetters) {
     this.title = definition.title;
+    this.isTitleCellReferenced = definition.isTitleCellReferenced || false;
+    this.titleCellReference = definition.titleCellReference || "";
```
hokolomopo, line 32: In the definition I think it's better to have a single key for the title (see my previous comment), but here you don't haev a choice to have both. But `titleCellReference` should be a `Range`, and not a string. And it should be transormes like the other ranges of the chart. 

Currently you title reference doen't react at all to the addition/removal of columns rows, neither in the `updateRanges` nor in the collaborative transforms `transformDefinition`.

You're missing a lot of tests here.

hokolomopo (commented):
Thanks for the work, but there's some very big ussues.

The title is not updated dynamically at all, and the title reference does not react to sheet modifications.

src/types/chart/common_bar_combo.ts
dhrp-odoo: Hello, Firstly, I appreciate your input.

If we maintain

```js
readonly title: Range | string;
readonly titleType: "reference" | "value";
```

Then we have to include the title in the runtime of each chart. right? Otherwise, it will raise an error when retrieving the default Chart.js runtime because there is a chance that the title may contain Range. So when passing data we also need to pass title with it. Please correct me if i am wrong

And in case of excel export, while getting definition, if title type is reference then we need to set the title manually by getting getCellText, but is not possible with core getters right?

hokolomopo: - The range should be converted to the cell's value when creating the runtime
- For excel export, we can define `exportForExcel` in an UI plugin to overwrite the cell reference to its value

dhrp-odoo:
Hello @hokolomopo,

Thanks for your suggestions. I followed your advice to use only the property to ensure we don't miss anything. I've ensured that the title updates with all operations, just like other ranges in the definition.

I'm encountering a minor issue with TypeScript. I need the title's type to change based on the titleType property. If titleType is 'reference', the type of title should be a Range, and if it's 'value', it should be a string. I want to avoid having to explicitly cast the title as Range or as string every time I use it.

I tried implementing TypeScript's conditional types to achieve this, but it didn't work out. Can you suggest a better way to handle this scenario?

hokolomopo:
> I'm encountering a minor issue with TypeScript. I need the title's type to change based on the titleType property. If titleType is 'reference', the type of title should be a Range, and if it's 'value', it should be a string. I want to avoid having to explicitly cast the title as Range or as string every time I use it.

Hello :wave: 

Yeah that part was always gonna be the most annoying :x I' don't think this can be done using TS conditional type ? I could be wrong tho, I'm not very familiar with those.

In the same vein as what you did, you could use `typeof this.title` as the condition for your ternaries, it's almost the same thing but now you don't have to cast it. Something like `const str = typeof this.title === "string" ? this.title : getTitleRangeValue(...)`

A batter soluton would be to combine the title type and the title range/string into a single object, that way it could be typed. 
```ts
interface ChartTitle = { type: "reference", value: Range } | { type: "string", value: string }
```
that way when checking for title.type === "reference", typescript will detect that title.value is a Range.

src/helpers/figures/charts/chart_common.ts
```diff
@@ -111,6 +127,18 @@ export function copyLabelRangeWithNewSheetId(
...
+ * Copy a title reference. If the title is a reference to a range on the
+ * sheetIdFrom, the range will target sheetIdTo.
+ */
+export function copyTitleReferenceWithNewSheetId(
+  sheetIdFrom: UID,
+  sheetIdTo: UID,
+  title: Range | undefined
+): Range | undefined {
+  return title ? copyRangeWithNewSheetId(sheetIdFrom, sheetIdTo, title) : undefined;
+}
```
hokolomopo, line 140: The function should probably take directly the `ChartTitle` as argument. That way we wouldn't need to do a ternary each time we want to call it

src/helpers/figures/charts/chart_common.ts
```diff
@@ -388,3 +435,24 @@ export function getChartPositionAtCenterOfViewport(
 
   return position;
 }
+
+/**
+ * Retrieve the title for a chart from its definition.
+ * If the title is a cell reference, this function returns the formatted value of the referenced cell.
+ * If the title is provided as a direct value, it returns the title as is.
+ */
+export function getChartTitle(getters: Getters, title: ChartTitle): string {
```
hokolomopo, line 444:
```suggestion
export function getChartRuntimeTitle(getters: Getters, title: ChartTitle): string {
```
maybe

src/plugins/ui_core_views/evaluation_chart.ts
```diff
@@ -104,10 +105,22 @@ export class EvaluationChartPlugin extends UIPlugin<EvaluationChartState> {
...
-          figures.push({
-            ...figure,
-            data: figureData,
-          });
+          // Converts the chart title as cell value if it is a reference
+          if (figureData.titleType === "reference") {
+            const range = this.getters.getRangeFromSheetXC(sheet.id, figureData.title!);
+            if (range === undefined || range.invalidXc || range.invalidSheetName) {
+              throw new InvalidReferenceError();
+            }
```
hokolomopo, line 113: I think we'd want to set the title to an empty string rather than throwing and preventing the export

tests/figures/chart/charts_component.test.ts
```diff
@@ -270,7 +287,7 @@ describe("charts", () => {
         setInputValueAndTrigger(dataSeriesValues, "B2:B4");
         await nextTick();
         await simulateClick(".o-data-series .o-selection-ok");
-        const definition = model.getters.getChartDefinition(chartId) as ScorecardChart;
+        const definition = model.getters.getChartDefinition(chartId) as unknown as ScorecardChart; // TODO: fix typing
```
hokolomopo, line 290: TODO

tests/figures/chart/charts_component.test.ts
```diff
@@ -1165,6 +1182,109 @@ describe("charts", () => {
     });
   });
 
+  describe("Cell reference for title", () => {
```
hokolomopo, line 1185: The tests that need a mounted component should probably be in chart_title.test.ts (that should be renamed to `chart_title_component.test.ts` as per our test naming convention), and the one that only need the model should not mount a spreadsheet, and be either in `chart_plugin.test.ts` or maybe in a new file `chart_title_plugin.test.ts`

tests/side_panels/building_blocks/chart_title.test.ts
```diff
@@ -0,0 +1,48 @@
+import { Component, xml } from "@odoo/owl";
```
hokolomopo, line 1: the file should probably be moved inside the `chart` folder now that it clearly is only related to charts :)

tests/collaborative/collaborative_sheet_manipulations.test.ts
```diff
@@ -570,6 +570,7 @@ describe("Collaborative Sheet manipulation", () => {
       dataSets: ["A1:A3", "F1:F3"],
       labelRange: "F3",
       title: "chart title",
+      titleType: "string",
```
hokolomopo, line 573: should be a reference, and test that it's correclty modified in the test of this file

hokolomopo (commented):
Some comments on details, but the core seems good :slightly_smiling_face:

src/helpers/figures/charts/chart_common.ts
```diff
@@ -72,10 +79,19 @@ export function updateChartRangesWithDataSets(
     isStale = true;
     labelRange = range;
   }
+  let title = chartTitle;
+  if (title.type === "reference") {
+    const range = adaptChartRange(title.reference, applyChange);
+    if (range !== title.reference) {
```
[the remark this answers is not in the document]
hokolomopo: I'm pretty sure the `if` condition is always true. We compare two Range instance, so the reference is compared and not the actual value of the range. But since `applyChange` returns a copy of the range (that's the part i'm not 100% sure about), the comparison is useless because `range` is always !== `copyOfRange`.

Anyway that's ot something you introduced, so you can juste add a TODO so we will mayeb check if more later.

src/helpers/figures/charts/gauge_chart.ts
```diff
@@ -239,6 +264,13 @@ export class GaugeChart extends AbstractChart {
   getContextCreation(): ChartCreationContext {
     return {
       ...this,
+      title:
+        this.title.type === "reference"
+          ? this.title.reference
+            ? this.getters.getRangeString(this.title.reference, this.sheetId)
+            : ""
+          : this.title.value,
```
hokolomopo, line 272: you should always avoid using multi-level ternary statements. Since you ahve to repeat this ternary a bunch of times, I'd add a method (or helper) getTitle (do'nt forget to also adapt Odoo's PR)

LucasLefevre:
Hello @dhrp-odoo 
It looks like it requires a rebase :/ We merged a big task for chart customization which conflicts

src/components/side_panel/chart/bar_chart/bar_chart_design_panel.xml
dhrp-odoo: @hokolomopo, 

With the recent merging of ANHE's major improvement task, the overall picture has changed.

The `title.ts` file is now used not only for the Chart Title but also for the Axis Title.

Should we continue to accept only one prop (figureId) and let the component handle everything, or should we accept the necessary props from the parent? 

image

hokolomopo: Hello :slightly_smiling_face: 

I still feel like having as few props as possible is the way to go. But now since the ChartTile is used in multiple place, we need more props than the figureId indeed. 

Since we now already have 8 props for this component, you can either add the props you need and after this task we'll try to do a follow up to refactor and reduce the number of props, or you can refactor in this task. Do as you want :)

src/types/chart/chart.ts
```diff
@@ -81,6 +81,10 @@ export interface TitleDesign {
   readonly color?: Color;
 }
 
+export interface ChartTitleConfiguration extends Omit<TitleDesign, "text"> {
+  readonly title: ChartTitle;
```
anhe-odoo, line 85: Do we really need this object type ? Can't we simply use ChartTitle everywhere ? For me it's strange to have this nested structure that only contains one field

src/types/chart/chart.ts
```diff
@@ -103,6 +107,7 @@ export type ExcelChartType = "line" | "bar" | "pie" | "combo" | "scatter";
 
 export interface ExcelChartDefinition {
   readonly title?: TitleDesign;
+  readonly titleType?: ChartTitleType;
```
anhe-odoo, line 110: Maybe we can find a way to group the titleType and the titleDesign in one single object :) It will avoid having another field in the chart definition (and the contextCreation) that are already really long.

anhe-odoo (commented):
Nice feature :) And sorry for the changes you had to make because of my PR :/
I only have some suggestion :)

src/types/chart/chart.ts
dhrp-odoo: Yes, I believe we do. I discussed this with ADRM in this comment: https://github.com/odoo/o-spreadsheet/pull/4024#issuecomment-2074837702

anhe-odoo: That was not what I meant, but anyway, my question was wrong as I didn't figure we were extending the TitleDesign (I thoug you were creating a new object TitleConfiguration which only contains the ChartTitle, but it's not the case so you can ignore this comment :) )

src/types/chart/chart.ts
dhrp-odoo:
```ts

export interface WaterfallChartDefinition {
  readonly title: ChartTitleConfig;
  // other properties
}

export interface ChartTitleConfig extends TitleDesign {
  readonly type?: "string" | "reference";
}

/**
 * The AbstractChartTitle interface is primarily used within the AbstractChart
 * to handle the various types of title.
 */
export interface AbstractChartTitle extends Omit<ChartTitleConfig, "text"> {
  readonly title: ChartTitle;
}

export type ChartTitle =
  | { type: "reference"; reference: Range | undefined }
  | { type: "string"; value: string };

```

Do you suggest having a structure like this?

Grouping the design in one single object and the type in another might introduce unnecessary complexity, especially considering that we're already nesting one object in AbstractChart to manage types for the title properly with TypeScript errors.

anhe-odoo: I was thinking of something like
```js
export type ChartTitle = {
design: TitleDesign;
} & (
{type: "reference"; reference: Range | undefined}
| {type: "string"; value: string }
);

So that we will have a title object containing the design (where we can remove the `text` part) as well as the type and the reference/value :) But if it's too complex for the structure we can keep it as is

Comment (dhrp-odoo) [reply src/types/chart/chart.ts:110]: Hi 🙂
One last question before I start making changes: Do you mean to remove `text` directly from TitleDesign, or only when using it for this ChartTitle?

Comment (anhe-odoo) [reply src/types/chart/chart.ts:110]: Hello :) For me, it can be fully removed for TitleDesign, as long as we use ChartTitle in the definition instead :)

File: demo/data.js [hunk 18]
```diff
@@ -179,7 +179,7 @@ export const demoData = {
       rows: {},
       cols: {},
       merges: [],
-      cells: { B2: { content: "42" } },
+      cells: { B2: { content: "42" }, D2: { content: "Bar" } },
```
rrahir, line 182: If you encounter a situation shere you have to update the raw data in place, it means you need a migration for the said data.

rrahir (changes requested):

demo/data.js
dhrp-odoo: I didn't fully understand your point. Are you referring to the need to write a migration in `migration_step.ts`? Could you please elaborate further?

tests/figures/chart/charts_component.test.ts
```diff
@@ -584,6 +585,34 @@ describe("charts", () => {
     ]);
   });
 
+  test("Chart title updates after copy/pasting to another sheet if title is a cell reference", async () => {
```
hokolomopo, line 588: The test should be in the chart_plugin.test.ts file, not this one. You do not test a component here. Also the test name is a bit wrong, we do not update the range on the copy/paste, the title still references the same range.

There also should be a test on duplicate_sheet, there the references to the ranges of the duplcated sheet need to be updated.

hokolomopo (commented):
:wave: 

(runbot is red, a snapshot wasn't updated)

src/components/composer/standalone_composer/standalone_composer.ts
```diff
@@ -13,8 +13,10 @@ import { StandaloneComposerStore } from "./standalone_composer_store";
 css/* scss */ `
   .o-spreadsheet {
     .o-standalone-composer {
+      position: relative;
       min-height: 24px;
       box-sizing: border-box;
+      z-index: ${ComponentsImportance.StandaloneComposer};
```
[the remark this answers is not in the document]
dhrp-odoo: image

Found this: https://stackoverflow.com/questions/45366817/why-position-relative-without-z-index-creates-a-new-stacking-context

src/helpers/figures/charts/bar_chart.ts
```diff
@@ -137,26 +140,53 @@ export class BarChart extends AbstractChart {
...
+      this.sheetId,
+      sheetId,
+      this.title
+    );
+    const updatedAxesDesign = updateAxesDesignWithSheetReference(
+      this.getters,
+      this.sheetId,
+      sheetId,
+      this.axesDesign
+    );
```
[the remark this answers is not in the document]
dhrp-odoo: If I understand correctly, `copyInSheetId` is used when we `copy & paste` a chart within the same sheet or across different sheets. 

In cases where we paste the chart into a different sheet, we first need to process the chart's title. If the title contains a formula, the ranges within it should be updated to point to the original sheet ranges (the copied sheet) rather than the pasted sheet ranges. Is that correct?

On the other hand, `copyForSheetId` is used when `DUPLICATE_SHEET`, which involves cloning the ranges with a new `sheetId`. However, since the chart title contains a range string, it doesn't require this specific update. right?

Please correct me if I have misunderstood.

src/components/side_panel/chart/building_blocks/title/title.ts
```diff
@@ -117,4 +122,22 @@ export class ChartTitle extends Component<Props, SpreadsheetChildEnv> {
     this.state.activeTool = "";
     this.openedEl = null;
   }
+
+  toggleComposerMode(isComposerEnabled: boolean): void {
```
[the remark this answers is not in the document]
hokolomopo: Yes that's it, I'm pretty sure it's what vsc suggested :)

src/helpers/figures/charts/bar_chart.ts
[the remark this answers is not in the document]
hokolomopo: Actually I think we're both half right:

- You're right that we also need to adapt the formulas in `copyInSheetId`. If we copy the chart in another sheet, we need to change `=A1` to `=OriginalSheet!A1`. We didn't need to to that for the dataset because they are already converted to a Range, which point to a specific sheetId.
- But we also need to adapt the ranges on duplicate sheet. A title `=A1` can stay `=A1` in the duplicated sheet, but a title `=OriginalSheet!A1` would need to be transformed to `=DuplicatedSheet!A1`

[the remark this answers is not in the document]
dhrp-odoo: Oh, yes, that’s what we need here in the end! :smile:

src/helpers/figures/charts/bar_chart.ts
```diff
@@ -235,9 +273,9 @@ export function createBarChartRuntime(chart: BarChart, getters: Getters): BarCha
       ...CHART_COMMON_OPTIONS,
       indexAxis: chart.horizontal ? "y" : "x",
       layout: getBarChartLayout(definition),
-      scales: getBarChartScales(definition, chartData),
+      scales: getBarChartScales(getters, definition, chartData),
```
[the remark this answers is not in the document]
dhrp-odoo: Oh, I see! Can update the `definition` with the evaluated axis designs beforehand, 
```ts
  const updatedDefinition: GenericDefinition<BarChartDefinition> = {
    ...definition,
    axesDesign: evaluatedAxesDesign,
  };
```

so we don't have to pass it through the function parameter chain like `getChartAxis > getChartAxisTitleRuntime` to  set evaluated value?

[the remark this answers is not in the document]
hokolomopo: You should put the evaluated title inside `ChartRuntimeGenerationArgs` rather than updating the definition on the fly. That's where we now put the data extracted from the model (evaluated dataset, evaluated labels, locale, etc). You can add keys for your evaluated title/axis title.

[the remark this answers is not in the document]
dhrp-odoo: Yeah, I’ve already done that. I was taking about `getBarChartScales`, which accepts args which includes `evaluatedAxesDesign` now. We need to pass this to `getChartAxis > getChartAxisTitleRuntime` to set the evaluated value.  

Can we instead create an `updatedDefinition` object that includes the evaluated axesDesign, and then pass it like this:
```ts
    scales.y = getChartAxis(updatedDefinition, "left", "labels", options);
```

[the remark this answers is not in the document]
hokolomopo: personnaly, I would  add an argument instead of trying to update the definition on the fly I think

src/helpers/figures/charts/scorecard_chart.ts
```diff
@@ -451,11 +483,11 @@ export function createScorecardChartRuntime(
     chart.baselineMode === "progress" && isNumber(baselineDisplay, locale)
       ? toNumber(baselineDisplay, locale)
       : 0;
+  const chartTitle = getEvaluatedChartTitle(getters, chart.title);
   return {
     title: {
-      ...chart.title,
-      // chart titles are extracted from .json files and they are translated at runtime here
```
LucasLefevre, line 457: could you keep the comment about translation?

tests/figures/chart/chart_plugin.test.ts
```diff
@@ -1407,6 +1409,134 @@ describe("title", function () {
...
+      expect(definition.title.text).toEqual("=A1");
+
+      let options = getChartConfiguration(model, "1").options;
+      expect(options!.plugins!.title!.text).toEqual("Hello World");
+      expect(options!.scales!.x!.title.text).toEqual("Hello World");
+
+      moveColumns(model, "B", ["A"]);
+
+      definition = model.getters.getChartDefinition("1");
+      expect(definition.title.text).toEqual("=B1");
```
LucasLefevre, line 1471: you can also check `definition.axesDesign.x.title.text`

tests/figures/chart/chart_plugin.test.ts
```diff
@@ -1407,6 +1409,134 @@ describe("title", function () {
...
+      expect(definition.title).toEqual({ text: "='Copy of Sheet1'!A1" });
+      expect(definition.axesDesign).toEqual({
+        x: { title: { text: "='Copy of Sheet1'!A1" } },
+        y: { title: { text: "='Copy of Sheet1'!A1" } },
+      });
+    }
+  );
+
+  test.each(["line", "bar", "pyramid", "combo", "waterfall", "scatter"] as const)(
+    "Retains original sheet references when copy-pasting a chart with cell-referenced titles",
```
LucasLefevre, line 1511: The body of this test is exactly the same as the test above. Not sure what it was meant to test

tests/figures/chart/chart_plugin.test.ts
```diff
@@ -1407,6 +1409,134 @@ describe("title", function () {
...
+      options = getChartConfiguration(model, "1").options;
+      expect(options!.plugins!.title!.text).toEqual("Hello World");
+      expect(options!.scales!.x!.title.text).toEqual("Hello World");
+    }
+  );
+
+  test.each(["line", "bar", "pyramid", "combo", "waterfall", "scatter"] as const)(
+    "Updates chart title when duplicating a sheet with a cell-referencing chart title",
+    (type) => {
+      setCellContent(model, "A1", "Hello World");
```
LucasLefevre, line 1482: `setCellContent` useless for this test

tests/figures/chart/chart_plugin.test.ts
```diff
@@ -1407,6 +1409,134 @@ describe("title", function () {
...
+      createChart(
+        model,
+        {
+          dataSets: [{ dataRange: "C1:C4" }],
+          labelRange: "A2:A4",
+          type,
+          title: { text: "=Sheet1!A1" },
+          axesDesign: {
+            x: { title: { text: "=Sheet1!A1" } },
+            y: { title: { text: "=Sheet1!A1" } },
```
LucasLefevre, line 1492: could you also add references to another sheet (and check they do not change)

tests/figures/chart/waterfall/waterfall_panel_component.test.ts
```diff
@@ -91,7 +99,7 @@ describe("Waterfall chart side panel", () => {
       await openChartConfigSidePanel(model, env, chartId);
       await click(fixture, ".o-panel-design");
 
-      expect(getHTMLInputValue(".o-chart-title input")).toEqual("My Waterfall chart");
+      expect(textContent(".o-chart-title .o-composer")).toEqual("My Waterfall chart");
```
LucasLefevre, line 102: ADRM introduced `toHaveText` which what you cant here ;)
```suggestion
      expect(".o-chart-title .o-composer").toHaveText("My Waterfall chart");
```

tests/test_helpers/helpers.ts
```diff
@@ -729,6 +729,14 @@ export function textContentAll(cssSelector: string): string[] {
   return [...nodes].map((node) => node.textContent).filter((text): text is string => text !== null);
 }
 
+/**
+ * Return the text of the first node matching the selector
+ */
+export function textContent(cssSelector: string): string {
```
LucasLefevre, line 735: remove this in favor of `toHaveText`

src/plugins/core/range.ts
```diff
@@ -428,6 +432,27 @@ export class RangeAdapter implements CommandHandler<CoreCommand> {
     return { _sheetId: sheetId, _zone: zone };
   }
 
+  // TODO: Remove this getter once ADRM's PR has been merged
```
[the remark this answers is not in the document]
dhrp-odoo: Here it is :)
https://github.com/odoo/o-spreadsheet/pull/5181

tests/figures/chart/chart_plugin.test.ts
dhrp-odoo: 😅 I accidentally copy-pasted the test from above and forgot to modify it—apologies for that! 

Both tests are necessary to test the methods `copyForSheetId` and `copyInSheetId`.

dhrp-odoo:
decided to cancel the task

PR #4052, opened 2024-04-11
[FIX] clipboard: cannot paste chart with deleted ranges

hokolomopo:
## Description

Before this commit, when pasting a chart with ranges in its data/labels that were deleted between the copy and the paste, the clipboard would traceback.

That's because `getRangeString` crashed when the sheetId does not exist anymore. Fixed it by making the function return a string with an invalidSheetName error rather than crashing.

Also replace the uses of `getRangeString` with `getValidRangeString`,
which will return undefined for invalid ranges so the typing forces us
to handle it properly.

Task: : [3618758](https://www.odoo.com/web#id=3618758&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/core/range.ts
+++ b/src/plugins/core/range.ts
@@ -360,6 +360,9 @@ export class RangeAdapter implements CommandHandler<CoreCommand> {
     if (range.invalidXc) {
       return range.invalidXc;
     }
+    if (!this.getters.tryGetSheet(range.sheetId)) {
+      return INCORRECT_RANGE_STRING;
+    }
     if (range.zone.bottom - range.zone.top < 0 || range.zone.right - range.zone.left < 0) {
       return INCORRECT_RANGE_STRING;
     }
--- a/tests/plugins/range.test.ts
+++ b/tests/plugins/range.test.ts
@@ -563,6 +563,14 @@ describe("range plugin", () => {
       }
     );
   });
+
+  test("getRangeString does not crash with deleted sheet", () => {
+    const range = m.getters.getRangeFromSheetXC("s1", "A1");
+    expect(m.getters.getRangeString(range)).toBe("s1!A1");
+    createSheet(m, { sheetId: "s2" });
+    deleteSheet(m, "s1");
+    expect(m.getters.getRangeString(range)).toBe(INCORRECT_RANGE_STRING);
+  });
 });
 
 describe("Helpers", () => {
--- a/src/helpers/charts/bar_chart.ts
+++ b/src/helpers/charts/bar_chart.ts
@@ -23,7 +23,7 @@ import {
 import { LegendPosition, VerticalAxisPosition } from "../../types/chart/common_chart";
 import { Validator } from "../../types/validator";
 import { toXlsxHexColor } from "../../xlsx/helpers/colors";
-import { createRange } from "../range";
+import { createValidRange } from "../range";
 import { AbstractChart } from "./abstract_chart";
 import {
   ChartColors,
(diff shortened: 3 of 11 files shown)


src/plugins/core/range.ts
```diff
@@ -373,6 +373,8 @@ export class RangeAdapter implements CommandHandler<CoreCommand> {
     if (prefixSheet) {
       if (rangeImpl.invalidSheetName) {
         sheetName = rangeImpl.invalidSheetName;
+      } else if (!this.getters.tryGetSheet(rangeImpl.sheetId)) {
+        sheetName = INCORRECT_RANGE_STRING;
```
LucasLefevre, line 377: The root cause is that we don't adapt ranges that are in the clipboard, right?
(it doesn't harm that `getRangeString` is robust to missing sheets, but it may hide the root cause of a bug)

hokolomopo: Not sure but I don't think we want to adapt the ranges in the clipboard ? That's the whole discussion we already had 4 times about pasting things with the current state of the sheet VS the state at the time of pasting.

If we adapt the ranges inside the clipboard, that means that copy col A => delete col A => paste col A does not work

hokolomopo: Maybe we could put the chart Definiton instead of the abstract chart in the clipboard tho, that may solve the issues

rrahir (commented):
The fix seems good, just need a test for getRangeString specifically as it is fixed in itself.

tests/plugins/range.test.ts
```diff
@@ -563,6 +563,14 @@ describe("range plugin", () => {
       }
     );
   });
+
+  test("getRangeString does not crash with deleted sheet", () => {
+    const range = m.getters.getRangeFromSheetXC("s1", "A1");
+    expect(m.getters.getRangeString(range)).toBe("s1!A1");
+    createSheet(m, { sheetId: "s2" });
+    deleteSheet(m, "s1");
+    expect(m.getters.getRangeString(range)).toBe(INCORRECT_RANGE_STRING + "!A1");
```
rrahir, line 572: this seems weird the. I would simply expect "#REF" and nothing else

src/plugins/core/range.ts
```diff
@@ -401,6 +404,19 @@ export class RangeAdapter implements CommandHandler<CoreCommand> {
     return `${prefixSheet ? sheetName + "!" : ""}${rangeString}`;
   }
 
+  /** Same as getRangeString but return undefined for invalid ranges */
+  getValidRangeString(
```
rrahir, line 408: it's not the same as getRangeString, you allow to pass no range at all. you should follow the same signature don't you think?

hokolomopo: Allowing passing an undefined range is pretty useful, it removes 5+ ternaries. I thi k it's worth not havign the exact same signature for this. I can change the jsDoc tho if it bother you.

rrahir: imo, the 5 ternaries is not a valid point (again, just my pov). I think the getter should not be twisted so that it fits too much the usecase you have. By that logic, every ternary could/should be removed in favor of a function/getter that takes care of the garbage payload or corner cases. I going overboard with this but still, if I have 2 getters that pretty much do the same thing, I would expect the same payload.

hokolomopo: IMO I don't mind having a slighltu different signature if that means that it's easier to use. But I don't care that much and I'm not gonna fight for it, I updated it to have the exact same signature as getRangeString.

rrahir:
@robodoo r+

PR #4054, opened 2024-04-11
[IMP] formats: add quarter date formatting

LucasLefevre:
With this commit, dates can now be formatted by their quarter: "Q2 2023"

This also prepares the ground for the upcoming pivot features where we'll be able to group dates by quarter.

Tecnical note on the implementation: I initially wanted to reused the string part of formats [$...] to display the leading Q.
Format "q yyyy" would give "3 2023" and "[$Q] q yyyy" would give "Q3 2023" because of the leading string part. There's a translation issue though. How translator would be able to translate "Q" on its own, without any context? In some languages, it's also different than the first letter of the word "quarter". In Danish it's "3. kvt 2023".

Given that, "q" in a format automatically adds the leading "Q". It's less flexible, but it allows to have context for translators.

## Description:

Task: : [3864187](https://www.odoo.com/web#id=3864187&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/format.ts
+++ b/src/helpers/format.ts
@@ -388,7 +388,7 @@ export function isDateTimeFormat(format: Format) {
     return false;
   }
 }
-const allowedDateTimeFormatFirstChar = new Set(["h", "m", "y", "d"]);
+const allowedDateTimeFormatFirstChar = new Set(["h", "m", "y", "d", "q"]);
 
 export function applyDateTimeFormat(value: number, format: Format): FormattedValue {
   // TODO: unify the format functions for date and datetime
@@ -435,6 +435,10 @@ function formatJSDate(jsDate: DateTime, format: Format): FormattedValue {
           return MONTHS[jsDate.getMonth()];
         case "mmmmm":
           return MONTHS[jsDate.getMonth()].slice(0, 1);
+        case "qq":
+          return _t("Q%(quarter)s", { quarter: jsDate.getQuarter() });
+        case "qqqq":
+          return _t("Quarter %(quarter)s", { quarter: jsDate.getQuarter() });
         case "yy":
           const fullYear = String(jsDate.getFullYear()).replace("-", "").padStart(2, "0");
           return fullYear.slice(fullYear.length - 2);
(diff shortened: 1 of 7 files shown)


src/plugins/core/cell.ts
```diff
@@ -286,6 +286,24 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
 
   exportForExcel(data: ExcelWorkbookData) {
     this.export(data);
+    const incompatible: string[] = [];
+    for (const formatId in data.formats || []) {
+      if (!isExcelCompatible(data.formats[formatId])) {
```
hokolomopo, line 291: We could also "convert" incompatible formats (remove the incompatible parts). That was a format ""Q1 2023" would become "2023" which is a bit better than no format at all. I do similar stuff in the xlsx import. Not 100% sur if it's worth.

src/helpers/format.ts
```diff
@@ -429,6 +429,8 @@ function formatJSDate(jsDate: DateTime, format: Format): FormattedValue {
           return MONTHS[jsDate.getMonth()];
         case "mmmmm":
           return MONTHS[jsDate.getMonth()].slice(0, 1);
+        case "q":
+          return _t("Q%(quarter)s", { quarter: jsDate.getQuarter() });
```
hokolomopo, line 433: Regarding your comment in the commit message, how about having a format "q" => 3 and another format "qq" => Q3 (or uppercase "Q") to have flexibility ?

hokolomopo (commented):
Only suggestion, otherwise it looks good 🙂 

Now do the weeks ! 😈

src/helpers/format.ts
pro-odoo: If we add it, we should add `qqqq` for `Quarter 3`.

src/plugins/core/cell.ts
pro-odoo: IMHO, no format at all is better than removing part. It will be easier for the user to catch that a date has no format than just the year. So I think it's not worth.

tests/formats/format_helpers.test.ts
```diff
@@ -772,6 +783,34 @@ describe("formatValue on date and time", () => {
       ).toBe(result);
     });
 
+    test.each([
+      ["2023/01/01", "Q1"],
+      ["2023/02/01", "Q1"],
+      ["2024/02/29", "Q1"], // Leap year
```
pro-odoo, line 789: 👍

pro-odoo (approved):

pro-odoo:
Little type in the commit message: `Tecnical note` -> `Technical note`

LucasLefevre:
@pro-odoo done :)

pro-odoo (approved):
robodoo r+

PR #4063, opened 2024-04-12
[FIX] filter: recompute header position

LucasLefevre:
Steps to reproduce:
- create a spreadsheet
- add a second sheet
- on the first sheet, add values in A1,A2,A3
- create a filter on A1:A3
- filter one or more values
- go to the second sheet
- update a cell
- go back to the first sheet => the positions of headers/cells are broken

Reason: the filter evaluation plugin currently only store the filtered rows of the active sheet. If the sheet is inactive, all rows are considered as visible (early return `false`)

When you update a cell in the second sheet however, the header positions are recomputed for all sheets. When computing the header positions of the first sheet, the visibilty is therefore wrong (considered as visible even though it's filtered)

opw-3858512
Task: 3858512

## Description:

Task: : [3858512](https://www.odoo.com/web#id=3858512&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_stateful/filter_evaluation.ts
+++ b/src/plugins/ui_stateful/filter_evaluation.ts
@@ -34,7 +34,7 @@ export class FilterEvaluationPlugin extends UIPlugin {
 
   private filterValues: Record<UID, Record<FilterId, string[]>> = {};
 
-  hiddenRows: Set<number> = new Set();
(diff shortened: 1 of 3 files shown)


rrahir:
@robodoo r+

PR #4077, opened 2024-04-15
[IMP] model: allow to batch commands in one history step

hokolomopo:
## [IMP] model: allow to batch commands in one history step

We sometime need to create a UI plugin specifically to handle a command
so that its sub-commands are batched in a single history step.

This is kind of a problem when working with stores, because then we
need both a store to handle all the business logic, and a plugin that
handle a command created specifically to batch sub-commands (looking
at you find & replace).

We can fix that easily by creating a local command `BATCH_COMMANDS`
that take a callback as argument, and that batches every command
and sub-command executed in the callback within one history step.

Task: 3870119

## [IMP] f&r: remove find & replace plugin

Now that we can batch commands, the plugin find & replace
is now useless.

Task:  [3870119](https://www.odoo.com/web#id=3870119&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


anhe-odoo (commented):
I'm not sure for the `BATCH_COMMANDS` name, as this is more like a command that can run any kind of callback, and, by design, every commands in the callback will be batched into a single history step, but this command could be used for something else with some other kind of callback ...
But I can't find a good name ATM :/ Maybe something like RUN_CALLBACK ? (But I don't like this one either)

PR #4083, opened 2024-04-16
[IMP] highlight: add drag & drop to array formula highlight

hokolomopo:
##  [IMP] highlight: add drag & drop to array formula highlight

With this commit, it is now possible to drag & drop an array formula
by clicking on its highlight.

## [IMP] highlight: improve border hitbox

The hitbox for the borders of the highlight was not very well placed.
There was some margins computations, I'm assuming they were there
to avoid overlapping with the highlight corners but they we both
wrong (there was overlap) and useless (the corner is above the border).

The highlight border was also not centered on the cell border, and
was quite small, making it hard to click.

This commit removed the "margins", and widened and centered the borders.

## [REF] highlight: rename Highlight component:

The component that handle highlight's UI interactions is named
`Highlight`, which is the same as the type `Highlight`. This is annoying
because we always end up importing the wrong `Highlight`, and need
named imports if we want to use both the type and the component in a
file.

This commit renames the component to `HighlightOverlay`

Task: : [3869873](https://www.odoo.com/web#id=3869873&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/helpers/recompute_zones.ts
```diff
@@ -353,7 +353,7 @@ function constructZonesFromProfiles(
     pendingZones = nextPendingZones;
   }
   mergedZone.push(...pendingZones);
-  return mergedZone;
+  return mergedZone as T[];
```
hokolomopo, line 356: Not really happy about that but it's mighty annoying to type...

The rationale is that, as far as I know, calling recomputeZones on only Zones (and no unbounded zones) will always return Zones. So it'd be nice if it was typed as such.

The fact that it's typed as returning UnboundedZones make it very annoying to use for simple cases where we only handle zones. Do you have any good ideas @laa-odoo ?

laa-odoo: Indeed, this is never supposed to happen: recompute non-UnboundedZones will always give non-UnBoundedZones
But it will also be complicated to type the mechanics of recomputeZone.... The best would be to force the typing within recomputeZone itself so that it doesn't bother anyone. (
imo I am for)

src/stores/array_formula_highlight.ts
```diff
@@ -1,25 +1,126 @@
...
+            this,
+            { cell: { col: zone.left, row: zone.top }, zone },
+            {
+              handleEvent: this.handleEvent.bind(this),
+              release: this.cancelChangeHighlight.bind(this),
+            }
+          );
+        }
+        break;
+      case "STOP_CHANGE_HIGHLIGHT": {
```
rrahir, line 38: https://www.youtube.com/watch?v=4fndeDfaWCg&t=54s 

I was so happy to get rid of that command a year ago :'(

hokolomopo: Wdym ? The command was still in the codebase before this PR :sweat_smile:

hokolomopo: Nvm I'm stupid and looked at the wrong command. 

But yeah I'm not really happy about how the highlight d&d work.

rrahir: There is fundamental issue as we want to have a differed behaviour while using a direct channel.
So we add commands to add some differed component

Might not be better but at least no commands - Highlights can be defined as movable and resizable.
What about registering them with callbacks as well? 
```typescript
    const highlight = {
        ...,
        onMove: Object|undefined {start: () => {
            // do internal shit to the concerned store/plugin
        }, stop: () => {
            // finalize stuff inside the store/plugin
        }},
        onResize: Object|undefined {start: () => {
            // do internal shit to the concerned store/plugin
        }, stop: () => {
            // finalize stuff inside the store/plugin
        }},
```

This way, the store can implement it's prefered behaviour without the global command that starts with "if i'm the one listening to the selection, then..." 

I think it can work since highlight are only store related now?

WDYT? cc @LucasLefevre @pro-odoo

LucasLefevre: I also don't like the additional command and temporary state associated. I'm also not sure the feature cost/benefit ratio is worth it to be honest

hokolomopo: TBH we probably can cancel the task, without the dashboard reorganize feature, drag & drop array formula isn't that useful. We'll still need something similar to drag & drop the selection tho

rrahir: >  We'll still need something similar to drag & drop the selection tho

Why? It seems like the exact same stuff ?

rrahir: > I also don't like the additional command and temporary state associated. I'm also not sure the feature cost/benefit ratio is worth it to be honest

https://media1.tenor.com/m/TY-5qp551-AAAAAC/ron-ronswanson.gif

PR #4084, opened 2024-04-16
[FIX] compiler: translatable error message

LucasLefevre:
Steps to reproduce (in odoo, in stable):
- change your language to french
- type in a cell '=AVERAGEIFS(1,1,)' or '=AVERAGEIFS(1,1,)' => the end of the error message isn't translated

"La fonction AVERAGEIFS s'attend à ce que le paramètre 1 fasse référence à une cellule ou à une plage, et non à un number."

`currentArg.type.toLowerCase()` is not translated.

I removed the type of the actual parameter from the error message because it's difficult to have a sentence contruction that work well in all languages and for all types. Think about an empty argument, you don't want the "a": "...or range, not a empty" is wrong. But you want it if it's a number: "...or range, not a number"

I also changed the placeholder to named placeholder to help translators to re-order them if needed in a given language.

Task: : [3872025](https://www.odoo.com/web#id=3872025&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/formulas/compiler.ts
+++ b/src/formulas/compiler.ts
@@ -121,10 +121,8 @@ export function compileTokens(tokens: Token[]): CompiledFormula {
           if (!isRangeInput(currentArg)) {
             throw new BadExpressionError(
               _t(
-                "Function %s expects the parameter %s to be reference to a cell or range, not a %s.",
-                functionName,
-                (i + 1).toString(),
-                currentArg.type.toLowerCase()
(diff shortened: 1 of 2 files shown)


rrahir:
@robodoo r+

PR #4090, opened 2024-04-17
[IMP] Dashboard: Cache clickable cells

rrahir:
## Description:

The computation of the clickable cells visibility condition can
sometimes be costy (per cell). At the moment, we recompute those
conditions at each rendering request, which can become quickly costy
if we simply scroll the viewport.

This revision introduces a caching of this computations to smooth out
the user experience when scrolling a dashboard.

Task: 3869752

Task: : [3869752](https://www.odoo.com/web#id=3869752&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/dashboard/clickable_cell_store.ts
+++ b/src/components/dashboard/clickable_cell_store.ts
@@ -1,0 +1,88 @@
+import { markRaw } from "@odoo/owl";
+import { positionToZone, toXC } from "../../helpers";
+import { CellClickableItem, clickableCellRegistry } from "../../registries/cell_clickable_registry";
+import { SpreadsheetStore } from "../../stores/spreadsheet_store";
+import {
+  CellPosition,
+  Command,
+  Rect,
+  SpreadsheetChildEnv,
+  UID,
+  invalidateEvaluationCommands,
+} from "../../types";
+
+type Garbage = ((position: CellPosition, env: SpreadsheetChildEnv) => void) | false;
+
+export interface ClickableCell {
+  coordinates: Rect;
+  position: CellPosition;
+  action: (position: CellPosition, env: SpreadsheetChildEnv) => void;
+}
+
+export class ClickableCellsStore extends SpreadsheetStore {
+  private _clickableCells: Record<UID, Record<string, Garbage>> = markRaw({});
+  private _registryItems: CellClickableItem[] = markRaw(
+    clickableCellRegistry.getAll().sort((a, b) => a.sequence - b.sequence)
+  );
+
+  handle(cmd: Command) {
+    if (
+      invalidateEvaluationCommands.has(cmd.type) ||
+      cmd.type === "EVALUATE_CELLS" ||
+      (cmd.type === "UPDATE_CELL" && ("content" in cmd || "format" in cmd))
+    ) {
+      this._clickableCells = markRaw({});
+      this._registryItems = markRaw(
+        clickableCellRegistry.getAll().sort((a, b) => a.sequence - b.sequence)
+      );
+    }
+  }
+
+  private getClickableAction(position: CellPosition): Garbage {
+    const { sheetId, col, row } = position;
+    const clickableCells = this._clickableCells;
+    const xc = toXC(col, row);
+    if (!clickableCells[sheetId]) {
+      clickableCells[sheetId] = {};
+    }
+    if (!(xc in clickableCells[sheetId]!)) {
+      clickableCells[sheetId][xc] = this.findClickableAction(position);
+    }
+    return clickableCells[sheetId][xc];
+  }
+
+  private findClickableAction(position: CellPosition) {
+    const getters = this.getters;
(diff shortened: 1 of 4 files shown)


src/components/dashboard/clickable_cell_store.ts
```diff
@@ -0,0 +1,86 @@
...
+}
+
+export class ClickableCellsStore extends SpreadsheetStore {
+  private _clickableCells: Record<UID, Record<string, Garbage>> = markRaw({});
+
+  handle(cmd: Command) {
+    if (
+      invalidateEvaluationCommands.has(cmd.type) ||
+      cmd.type === "EVALUATE_CELLS" ||
+      (cmd.type === "UPDATE_CELL" && cmd.content)
```
hokolomopo, line 30: also need to check cmd.format theoritically (`=CELL` formula)

src/components/dashboard/clickable_cell_store.ts
```diff
@@ -0,0 +1,86 @@
...
+    }
+    if (!(xc in clickableCells[sheetId]!)) {
+      clickableCells[sheetId][xc] = this.findClickableAction(position);
+    }
+    return clickableCells[sheetId][xc];
+  }
+
+  private findClickableAction(position: CellPosition) {
+    const getters = this.getters;
+    for (const item of clickableCellRegistry.getAll().sort((a, b) => a.sequence - b.sequence)) {
```
hokolomopo, line 51: Probably a non-factor because the registry isn't big, be we create a new array + sort it at each call of this method

src/components/dashboard/clickable_cell_store.ts
```diff
@@ -0,0 +1,86 @@
...
+        const position = { sheetId, col, row };
+        const action = this.getClickableAction(position);
+        if (!action) {
+          continue;
+        }
+        let zone: Zone;
+        if (getters.isInMerge(position)) {
+          zone = getters.getMerge(position)!;
+        } else {
+          zone = positionToZone({ col, row });
```
hokolomopo, line 74:
```suggestion
          zone = positionToZone(position);
```
nitpick

hokolomopo: also nitpick: this could be a ternary

src/components/dashboard/clickable_cell_store.ts
```diff
@@ -0,0 +1,86 @@
...
+        }
+        const rect = getters.getVisibleRect(zone);
+        cells.push({
+          coordinates: rect,
+          position: position,
```
hokolomopo, line 79: still nitpicking
```suggestion
          position,
```

src/components/dashboard/clickable_cell_store.ts
```diff
@@ -0,0 +1,86 @@
...
+    for (const col of getters.getSheetViewVisibleCols()) {
+      for (const row of getters.getSheetViewVisibleRows()) {
+        const position = { sheetId, col, row };
+        const action = this.getClickableAction(position);
+        if (!action) {
+          continue;
+        }
+        let zone: Zone;
+        if (getters.isInMerge(position)) {
+          zone = getters.getMerge(position)!;
```
hokolomopo, line 72: In the case of emrges, we will re-compute  `this.getClickableAction` for each cell of the merge, and have a ClickableCell with the same zone for each cell right ? That doesn't seems right ... (we love merges)

hokolomopo (commented):
Faster dashboards :eyes: 

The Odoo's PR is missing tho

src/components/dashboard/clickable_cell_store.ts
rrahir: Could you explain why? Not sure how a change in the format could have an impact on the clickable cell features. 
Now, assuming you are right, I guess we should factorize this kind of condition because I think it's not clear :D

src/components/dashboard/clickable_cell_store.ts
rrahir: it could be better 👍

src/components/dashboard/clickable_cell_store.ts
rrahir: agreed that it's a waste of resource but not fundamentally wrong ^^

rrahir: it could be fixed in prior versions as well actually

src/components/dashboard/clickable_cell_store.ts
hokolomopo: https://support.microsoft.com/en-us/office/cell-function-51bd39a5-f338-4dbe-a33f-955d67c2b2cf

LucasLefevre (approved):
robodoo r+

PR #4091, opened 2024-04-17
[FIX] FigureComponent: Stop Ctrl+A propagation

dhrp-odoo:
## Description:

Previously, when a figure was selected, And pressing CTRL + A would select all rows and columns in the sheet, and pressing the Delete key would delete the selected figure. This occurred because all key events were bubbling up to the grid component.

This PR stops the propagation for the Ctrl+A key event in the figure component.

Task: : [3863300](https://www.odoo.com/web#id=3863300&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/figures/figure/figure.ts
+++ b/src/components/figures/figure/figure.ts
@@ -7,6 +7,7 @@ import {
 import { figureRegistry } from "../../../registries";
 import { Figure, Pixel, ResizeDirection, SpreadsheetChildEnv, UID } from "../../../types/index";
 import { css, cssPropertiesToCss } from "../../helpers/css";
+import { keyboardEventToShortcutString } from "../../helpers/dom_helpers";
 
 type ResizeAnchor =
   | "top left"
@@ -181,8 +182,9 @@ export class FigureComponent extends Component<Props, SpreadsheetChildEnv> {
 
   onKeyDown(ev: KeyboardEvent) {
     const figure = this.props.figure;
+    const keyDownShortcut = keyboardEventToShortcutString(ev);
 
-    switch (ev.key) {
+    switch (keyDownShortcut) {
       case "Delete":
         this.env.model.dispatch("DELETE_FIGURE", {
           sheetId: this.env.model.getters.getActiveSheetId(),
@@ -209,6 +211,21 @@ export class FigureComponent extends Component<Props, SpreadsheetChildEnv> {
           x: figure.x + delta[0],
           y: figure.y + delta[1],
         });
+        ev.preventDefault();
+        ev.stopPropagation();
+        break;
+      case "Ctrl+A":
+        // Maybe in the future we will implement a way to select all figures
+        ev.preventDefault();
+        ev.stopPropagation();
+        break;
+      case "Ctrl+Y":
+      case "Ctrl+Z":
+        if (keyDownShortcut === "Ctrl+Y") {
+          this.env.model.dispatch("REQUEST_REDO");
+        } else if (keyDownShortcut === "Ctrl+Z") {
+          this.env.model.dispatch("REQUEST_UNDO");
+        }
         ev.preventDefault();
         ev.stopPropagation();
         break;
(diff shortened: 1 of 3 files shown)


src/components/figures/figure/figure.ts
```diff
@@ -180,8 +181,14 @@ export class FigureComponent extends Component<Props, SpreadsheetChildEnv> {
   }
 
   onKeyDown(ev: KeyboardEvent) {
-    const figure = this.props.figure;
+    // Allow the event to bubble to the grid only if Ctrl + Y or Ctrl + Z is pressed
+    // Let the grid handle the undo/redo operations
+    if (isCtrlKey(ev) && (ev.key.toUpperCase() === "Y" || ev.key.toUpperCase() === "Z")) {
+      return;
+    }
```
hokolomopo, line 188:
```suggestion
    if (["CTRL+Z", "CTRL+Y"].includes(keyboardEventToShortcutString(ev)) {
      return;
    }
```
Nitpick. Use whichever one you find more readable

dhrp-odoo: Definitely, this one looks more readable :)

src/components/figures/figure/figure.ts
```diff
@@ -210,9 +209,18 @@ export class FigureComponent extends Component<Props, SpreadsheetChildEnv> {
...
         });
         ev.preventDefault();
-        ev.stopPropagation();
         break;
+      case "Ctrl+A":
+        ev.preventDefault();
+        break;
+      case "Ctrl+Y":
+      case "Ctrl+Z":
+        // Allow the event to bubble to the grid only if Ctrl+Y or Ctrl+Z is pressed
```
rrahir, line 218: since https://github.com/odoo/o-spreadsheet/pull/1803 we try an approach where every single component is responsible of their own action and do not rely on their parent potential reaction.

I would rather reimplement the call to REQUEST_UNDO/REDO (or maybe have a function defined in the spreadsheet env ) and call it explicitely

rrahir (changes requested):

src/components/figures/figure/figure.ts
dhrp-odoo: @rrahir, I believe I fixed this issue similarly in the initial PR #3678. However, I recall @hokolomopo advised me to allow events to propagate up the grid. Here is the discussion: [Link](https://github.com/odoo/o-spreadsheet/pull/3678#discussion_r1497524202). Perhaps, I will explore the option of reverting back to the previous implementation.

rrahir: I wrote a message on the PR and will try to bring  this discussion up on monday with the whole team present ;-) Your approach is not worse than it was before but I fear we took some liberties that we shouldn't have and now as as good time as any to correct things. Let's keep this PR on hold until we discuss this so you avoid doing the same work twice.

rrahir: We still haven't decided exactly how to handle things but I would like to follow this commit idea : https://github.com/odoo/o-spreadsheet/commit/3660bf27522121e720212f90d15309ba61ae510f#diff-91190cae1bee7c5780c36f202f34ede615a061931ff957bf339ba10eee08f9ee which was to simply handle what you should  leave the other behaviours alone. This means that you should not stop the propagation by default but only if you handle the shortcut.

dhrp-odoo: @rrahir, Changes made!

src/components/figures/figure/figure.ts
```diff
@@ -212,6 +214,21 @@ export class FigureComponent extends Component<Props, SpreadsheetChildEnv> {
         ev.preventDefault();
         ev.stopPropagation();
         break;
+      case "Ctrl+A":
+        // Maybe in the future we will implement a way to select all figures
```
rrahir, line 218: 👍

rrahir:
@robodoo rebase-ff r+

@dhrp-odoo Thanks a lot ! :)

PR #4104, opened 2024-04-19
[FIX] stores: bye bye reactivity 👋

LucasLefevre:
## Description:

Reactivity is very very (very) slow, with potential overhead exceeding 90%.

With a reactive store, every `this.something` call traverses the reactive
proxy, incurring significant cost.

This commit completely revisit the way store updates trigger app rendering.
Previously, this process relied entirely on Owl's reactivity system.

Instead, it introduces a new approach where store updates are triggered by
explicitly declared public methods (mutators)
Since stores follow the CQS principle all public methods actually
mutate state in a store, they cannot have any other effect since they
can't return anything.

The new approach also relies on proxies to track mutators calls, but
the proxy is only on the surface. A mutator is usually only called once
in the component. A properties is also usually accessed only once for
rendering. The internal working of the store is not affected and can
run at full speed without any overhead.

Note the loss of fined-grained rendering. However, as observed, bypassing
reactivity might result in faster overall app rendering compared to
relying on reactivity for fine-grained rendering, even if it means rendering
the entire app rather than a few components.

Reactivity had another unexpected error-prone issue: since everything
was proxied, we could never compare objects references without using
`toRaw` to retrieve the underlying un-proxied object. It was very
error prone and added noise in the code.

On the large number data set, open the find & replace panel and search for "1"
It currently takes 1080ms (search + rendering)
With this commit, that time is reduced to just 111ms, nearly a tenfold
improvement!

Task: : [3829125](https://www.odoo.com/web#id=3829125&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/composer/composer/composer_store.ts
+++ b/src/components/composer/composer/composer_store.ts
@@ -1,4 +1,3 @@
-import { toRaw } from "@odoo/owl";
 import { EnrichedToken, composerTokenize } from "../../../formulas/composer_tokenizer";
 import { POSTFIX_UNARY_OPERATORS } from "../../../formulas/tokenizer";
 import { parseLiteral } from "../../../helpers/cells";
@@ -68,6 +67,16 @@ export interface ComposerSelection {
 }
 
 export class ComposerStore extends SpreadsheetStore {
+  mutators = [
+    "startEdition",
+    "setCurrentContent",
+    "stopEdition",
+    "stopComposerRangeSelection",
+    "cancelEdition",
+    "cycleReferences",
+    "changeComposerCursorSelection",
+    "replaceComposerCursorSelection",
+  ] as const;
   private col: HeaderIndex = 0;
   private row: HeaderIndex = 0;
   editionMode: EditionMode = "inactive";
@@ -84,9 +93,9 @@ export class ComposerStore extends SpreadsheetStore {
 
   constructor(get: Get) {
     super(get);
-    this.highlightStore.register(toRaw(this));
+    this.highlightStore.register(this);
     this.onDispose(() => {
-      this.highlightStore.unRegister(toRaw(this));
+      this.highlightStore.unRegister(this);
     });
   }
 
@@ -229,7 +238,7 @@ export class ComposerStore extends SpreadsheetStore {
         if (this.isSelectingRange) {
           this.editionMode = "editing";
         }
-        this.model.selection.resetAnchor(toRaw(this), {
+        this.model.selection.resetAnchor(this, {
           cell: { col: left, row: top },
           zone: cmd.zone,
         });
@@ -247,7 +256,7 @@ export class ComposerStore extends SpreadsheetStore {
             row: activePosition.row,
           });
           const zone = this.getters.expandZone(cmd.sheetIdTo, positionToZone({ col, row }));
-          this.model.selection.resetAnchor(toRaw(this), { cell: { col, row }, zone });
+          this.model.selection.resetAnchor(this, { cell: { col, row }, zone });
         }
         break;
       case "DELETE_SHEET":
@@ -373,7 +382,7 @@ export class ComposerStore extends SpreadsheetStore {
   private startComposerRangeSelection() {
     if (this.sheetId === this.getters.getActiveSheetId()) {
       const zone = positionToZone({ col: this.col, row: this.row });
-      this.model.selection.resetAnchor(toRaw(this), {
+      this.model.selection.resetAnchor(this, {
         cell: { col: this.col, row: this.row },
         zone,
       });
@@ -404,7 +413,7 @@ export class ComposerStore extends SpreadsheetStore {
     this.colorIndexByRange = {};
     const zone = positionToZone({ col: this.col, row: this.row });
     this.model.selection.capture(
-      toRaw(this),
+      this,
       { cell: { col: this.col, row: this.row }, zone },
       {
         handleEvent: this.handleEvent.bind(this),
@@ -519,7 +528,7 @@ export class ComposerStore extends SpreadsheetStore {
       return;
     }
     this.editionMode = "inactive";
-    this.model.selection.release(toRaw(this));
+    this.model.selection.release(this);
   }
 
   /**
--- a/src/components/helpers/draw_grid_hook.ts
+++ b/src/components/helpers/draw_grid_hook.ts
@@ -9,7 +9,7 @@ import { DOMDimension, OrderedLayers } from "../../types";
 export function useGridDrawing(refName: string, model: Model, canvasSize: () => DOMDimension) {
   const canvasRef = useRef(refName);
   useEffect(drawGrid);
-  const rendererManager = useStore(RendererStore);
+  const rendererStore = useStore(RendererStore);
   useStore(GridRenderer);
 
   function drawGrid() {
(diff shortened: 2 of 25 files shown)


src/components/composer/composer/composer_store.ts
```diff
@@ -68,6 +67,16 @@ export interface ComposerSelection {
 }
 
 export class ComposerStore extends SpreadsheetStore {
+  mutators = [
```
LucasLefevre, line 70: any better name idea for `mutators` ?

src/components/helpers/draw_grid_hook.ts
```diff
@@ -39,6 +39,7 @@ export function useGridDrawing(refName: string, model: Model, canvasSize: () =>
 
     for (const layer of OrderedLayers()) {
       model.drawLayer(renderingContext, layer);
+      // @ts-ignore
       rendererManager.drawLayer(renderingContext, layer);
```
LucasLefevre, line 43: `drawLayer` is a method used during rendering. It's not a mutator.

src/store_engine/dependency_container.ts
```diff
@@ -1,9 +1,14 @@
+import { EventBus } from "../helpers/event_bus";
 import { Get, StoreConstructor, StoreParams } from "./store";
 
+interface StoreUpdateEvent {
+  type: "store-updated";
+}
```
LucasLefevre, line 6: TODO: use a separate event instead of `model.trigger('update')`. (It should be independent from `model`)

src/store_engine/store_hooks.ts
```diff
@@ -15,8 +15,8 @@ export function useStoreProvider() {
     __spreadsheet_stores__: container,
     getStore: <T extends StoreConstructor>(Store: T) => {
       const store = container.get(Store);
-      const mutators: readonly string[] = store.mutators;
-      return proxifyToRender(env, store, mutators);
+      // TODO check this, it shouldn't be/use a hook here
```
rrahir, line 18: why not?

src/store_engine/dependency_container.ts
rrahir: maybe have the bus linked to the storeContainer itself?

src/stores/grid_renderer_store.ts
```diff
@@ -57,20 +56,14 @@ import { RendererStore } from "./renderer_store";
 export const CELL_BACKGROUND_GRIDLINE_STROKE_STYLE = "#111";
 
 export class GridRenderer {
+  mutators = [] as const;
```
rrahir, line 59: is it mandatory to declare for stores? the typing seems to miss this.

src/store_engine/dependency_container.ts
LucasLefevre: Yes, that's what I had in mind

src/store_engine/store_hooks.ts
LucasLefevre: because `useStoreRenderProxy` uses `useCompent` which is only available in the `setup` of components. (right click on a cell, Insert link => boom)

rrahir: ah right :'(

src/components/composer/composer/composer_store.ts
pro-odoo: `modifiers` ? `state_mutators` ?
Actually, `mutators` is clear for me

src/components/composer/composer/composer_store.ts
```diff
@@ -84,9 +93,9 @@ export class ComposerStore extends SpreadsheetStore {
 
   constructor(get: Get) {
     super(get);
-    this.highlightStore.register(toRaw(this));
+    this.highlightStore.register(this);
```
pro-odoo, line 96: 🥳

src/stores/grid_renderer_store.ts
pro-odoo: Does a store makes sense without mutators ?

pro-odoo (commented):
Very nice 🥳  Bye bye `toRaw` and thanks for nothing

src/stores/grid_renderer_store.ts
LucasLefevre: > Does a store makes sense without mutators ?

currently yes because other stores need to register themselves to this store

rrahir:
Could be the occasion to get rid of "RENDER_CANVAS" command then since we have a new bus to work with?
```diff
diff --git a/src/components/helpers/highlight_hook.ts b/src/components/helpers/highlight_hook.ts
index 7c0edb7e0..4fb2116f5 100644
--- a/src/components/helpers/highlight_hook.ts
+++ b/src/components/helpers/highlight_hook.ts
@@ -1,12 +1,12 @@
-import { onMounted, useEffect, useEnv } from "@odoo/owl";
-import { useLocalStore } from "../../store_engine";
+import { onMounted, useEffect } from "@odoo/owl";
+import { useLocalStore, useStoreProvider } from "../../store_engine";
 import { HighlightProvider, HighlightStore } from "../../stores/highlight_store";
-import { Ref, SpreadsheetChildEnv } from "../../types";
+import { Ref } from "../../types";
 import { useHoveredElement } from "./listener_hook";
 
 export function useHighlightsOnHover(ref: Ref<HTMLElement>, highlightProvider: HighlightProvider) {
   const hoverState = useHoveredElement(ref);
-  const env = useEnv() as SpreadsheetChildEnv;
+  const stores = useStoreProvider();
 
   useHighlights({
     get highlights() {
@@ -15,7 +15,7 @@ export function useHighlightsOnHover(ref: Ref<HTMLElement>, highlightProvider: H
   });
   useEffect(
     () => {
-      env.model.dispatch("RENDER_CANVAS");
+      stores.trigger("store-updated");
     },
     () => [hoverState.hovered]
   );
``` @LucasLefevre WDYT?

**Comment by LucasLefevre** [comment]
@rrahir true. I pushed a commit :)

**Review by rrahir — APPROVED** [review]
Ok for me :)

**Review by pro-odoo — APPROVED** [review]
👍

File: src/helpers/misc.ts [hunk 7]
```diff
@@ -313,6 +313,27 @@ export function debounce<T extends (...args: any) => void>(
...
+  return async (...args) => {
+    if (!scheduled) {
+      scheduled = true;
+      await Promise.resolve();
+      scheduled = false;
+      callback(...args);
+    }
+  };
+}
+
```
VincentSchippefilt, line 336: shouldn't batched be linked to the nextanimationframe instead of the microtick ?

VincentSchippefilt:
I think it is a fantastic idea. good job !

src/helpers/misc.ts
LucasLefevre: I pushed a commit but it's :nauseated_face: just because we don't have a direct access to owl's `Scheduler` (same trick used in the test).
I'll go and talk to the framework-js team tomorrow to check if they can export `Scheduler` from owl

src/helpers/misc.ts
```diff
@@ -313,23 +314,29 @@ export function debounce<T extends (...args: any) => void>(
...
 }
 
+class Root extends Component {
+  static template = xml`<div/>`;
+}
+const Scheduler = new App(Root).scheduler.constructor as unknown as { requestAnimationFrame: any };
+
 /**
  * Creates a batched version of a callback so that all calls to it in the same
  * microtick will only call the original callback once.
```
VincentSchippefilt, line 324: THIS IS A LIE :)

VincentSchippefilt:
what would be less error prone:
```js
class MyStore extends Store {
 constructor() {...}
 static mutators = ["myMytator","myOtherMutator"]
 myMutator(value) { ... }
 myOtherMutator(value) { ... }
 myGetter() { return value; }
}
```
or 

```js
class MyStore extends Store {
 constructor() {...}
 mutators : {
    myMutator = function (value) { ... },
    myOtherMutator = function (value) { ... }
 }
 myGetter() { return value; } // if I write this.mutators.myMutator here, it's obviously a bug
}
```

rrahir:
@robodoo r+

PR #4106, opened 2024-04-22
[FIX] Evaluator: Prevent incorrect invalidations

rrahir:
## Description:
This commit solves an issue with the invalidation process of spreaded
formulas. While marking the positions invalidated by a given position,
we would mistakenly mark the latter to be recomputed as well,
effectively creating an infinite loop (position is invalid > invalidate
its dependencies > position is invalidated > etc ...).

This issue was introduced by a bugfix in https://github.com/odoo/o-spreadsheet/pull/3125
and was not discovered as the infinite loop is actually stopped by our
maximum iteration limit.

Avoiding these useless iterations implies a performance improvement of a
factor 30 when reevaluating a spreaded formula.

Co-authored-by: Lucas Lefèvre <lul@odoo.com>

Task:  3883954

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_core_views/cell_evaluation/evaluator.ts
+++ b/src/plugins/ui_core_views/cell_evaluation/evaluator.ts
@@ -122,7 +122,7 @@ export class Evaluator {
       }
       if (!content) {
         // The previous content could have blocked some array formulas
-        impactedPositionIds.addMany(this.getArrayFormulasBlockedByOrSpreadingOn(positionId));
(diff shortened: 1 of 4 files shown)


laa-odoo:
at first glance, this indeed seems necessary. 
Intuitively, I don't see any other change to make to correct the thing. 
The fix seems complete to me

laa-odoo (approved):

PR #4109, opened 2024-04-22
[FIX] borders: wrong borders on remove rows

hokolomopo:
## Description

When removing multiple headers, the borders are not shifted correctly, leading to incorrect borders.

This happens because we modify the borders for each deleted rows separately. But we don't sort the rows before modifying the borders. So we first delete the row 1, then the row 2. But the row 2 is now the row 1, since the old row 1 has been deleted. So we end up with wrong operations.

This is solved by simply sorting the rows in reverse order before modifying the borders.

Task: : [3884112](https://www.odoo.com/web#id=3884112&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/core/borders.ts
+++ b/src/plugins/core/borders.ts
@@ -79,7 +79,7 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
         this.clearBorders(cmd.sheetId, cmd.target);
         break;
       case "REMOVE_COLUMNS_ROWS":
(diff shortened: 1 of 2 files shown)


LucasLefevre (approved):
robodoo r+

PR #4157, opened 2024-04-30
[REF] clipboard: remove usage of `MOVE_RANGES`

hokolomopo:
## Description

This commit changes the clipboard to display a new command `MOVE_REFERENCE` instead of `MOVE_RANGES`. The new command only moves the ranges that are references to cells (in formulas in chart) instead of modifying all the existing ranges.

There is two main reasons:

- the handling of `MOVE_RANGE` is bugged in most plugins when it moves a range from a sheet to another. Fixing the command could lead to broken spreadsheets, so for now we prefer to remove it.

- cut/paste was in a strange state where some operations were done twice. For example a filter table was moved once when handling `MOVE_RANGES` and then moved again when the cut/paste of the table was actually done. It's better to handle everything inside the clipboard.

Task: : [3899595](https://www.odoo.com/web#id=3899595&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/helpers/clipboard/clipboard_cells_state.ts
```diff
@@ -316,12 +329,6 @@ export class ClipboardCellsState extends ClipboardCellsAbstractState {
       for (let c = 0; c < width; c++) {
         const origin = rowCells[c];
         const position = { col: col + c, row: row + r, sheetId: sheetId };
-        // TODO: refactor this part. the "Paste merge" action is also executed with
-        // MOVE_RANGES in pasteFromCut. Adding a condition on the operation type here
-        // is not appropriate
-        if (this.operation !== "CUT") {
-          this.pasteMergeIfExist(origin.position, position);
-        }
```
LucasLefevre, line 324: :+1:

src/plugins/core/range.ts
```diff
@@ -424,6 +427,33 @@ export class RangeAdapter implements CommandHandler<CoreCommand> {
     return new RangeImpl(rangeInterface, this.getters.getSheetSize);
   }
 
+  /**
+   * Get the changed range if the given zone is moved to the target position
+   */
+  getMovedRange(
```
LucasLefevre, line 433: ?
```suggestion
  moveRangeInsideZone(
```

src/helpers/clipboard/clipboard_cells_state.ts
hokolomopo: :grinning:

PR #4160, opened 2024-04-30
[REF] Master simplify model startup vsc

VincentSchippefilt:
## Description:

preliminary work to split the model into core_model and all_the_rest_model

Task: : 3637435 (https://www.odoo.com/web#id=3637435&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/model.ts
+++ b/src/model.ts
@@ -545,20 +545,20 @@ export class Model extends EventBus<any> implements CommandDispatcher {
           return result;
         }
         this.status = Status.Running;
+        const start = performance.now();
         const { changes, commands } = this.state.recordChanges(() => {
-          const start = performance.now();
           if (isCoreCommand(command)) {
             this.state.addCommand(command);
             this.dispatchToCoreOnly(command);
           }
-          this.dispatchToUI(command);
-          this.trigger("command-dispatched", command);
-          this.finalize();
-          const time = performance.now() - start;
-          if (time > 5) {
-            console.info(type, time, "ms");
-          }
         });
+        this.dispatchToUI(command);
+        this.trigger("command-dispatched", command);
+        this.finalize();
+        const time = performance.now() - start;
+        if (time > 5) {
+          console.info(type, time, "ms");
+        }
         this.session.save(command, commands, changes);
         this.status = Status.Ready;
         this.trigger("update");
--- a/src/model.ts
+++ b/src/model.ts
@@ -127,6 +127,7 @@ export class Model extends EventBus<any> implements CommandDispatcher {
   private corePlugins: CorePlugin[] = [];
 
   private featurePlugins: UIPlugin[] = [];
+  private uiLocalNotCoreAndStuff: UIPlugin[] = [];
 
   private statefulUIPlugins: UIPlugin[] = [];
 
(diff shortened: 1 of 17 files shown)


VincentSchippefilt:
robodoo rebase-ff

src/plugins/ui_plugin.ts
```diff
@@ -27,7 +27,7 @@ export interface UIPluginConfig {
   readonly custom: ModelConfig["custom"];
   readonly session: Session;
   readonly defaultCurrencyFormat?: Format;
-  readonly customColors: Color[];
+  // readonly customColors: Color[];
```
hokolomopo, line 30: comment ?

src/model.ts
```diff
@@ -249,8 +247,7 @@ export class Model extends EventBus<any> implements CommandDispatcher {
     this.session.loadInitialMessages(stateUpdateMessages);
 
     for (let Plugin of coreViewsPluginRegistry.getAll()) {
-      const plugin = this.setupUiPlugin(Plugin);
-      this.coreViewsPlugins.push(plugin);
+      const plugin = this.setupCoreUiPlugin(Plugin);
```
hokolomopo, line 250: I didn't think `coreViewPlugin` was a very good name, but IMO CoreUIPlugin is even more confusing.

Anyway the registry should also be renamed if we want to change the nomenclature

src/model.ts
```diff
@@ -324,6 +321,20 @@ export class Model extends EventBus<any> implements CommandDispatcher {
...
+    const plugin = new Plugin({ ...this.corePluginConfig, getters: this.getters });
+    for (let name of Plugin.getters) {
+      if (!(name in plugin)) {
+        throw new Error(`Invalid getter name: ${name} for plugin ${plugin.constructor}`);
+      }
+      if (name in this.getters) {
+        throw new Error(`Getter "${name}" is already defined.`);
+      }
+      this.getters[name] = plugin[name].bind(plugin);
+    }
```
hokolomopo, line 334: Any way to extract this loop in a method ? It's now copy/pasted 3 times in this file

hokolomopo (commented):
Tchou tchou :steam_locomotive: 

Not a fan of the name `CoreUIPlugin`, but I don't have ideas either :shrug:

src/plugins/ui_core_views/header_sizes_ui.ts
```diff
@@ -21,7 +21,7 @@ interface CellWithSize {
   size: Pixel;
 }
 
-export class HeaderSizeUIPlugin extends UIPlugin<HeaderSizeState> implements HeaderSizeState {
```
LucasLefevre, line 24: note that this plugin depends on DOM api to create a canvas element. We won't be able to move it to a Worker as it is now, we'll have to refactor a bit more here.
`private ctx = document.createElement("canvas").getContext("2d")`

LucasLefevre (commented):
> the core_ui_plugins where of type UIPlugin, which is not correct

could you elaborate in the commit message why it's not correct?
It's actually not clear to me why it's not correct. Core UI plugins still handle UI commands, can use all getters. Except what's available in the plugin config, nothing changed much.

They should probably not be able to handle UI commands, but then this commit is not enough (there's currently a few exceptions: START, EVALUATE_CELLS,  REFRESH_PIVOT, others?). 

Also note that moving the evaluation to a worker will be very hard, even impossible. It depends on a loooot of things, especially for odoo pivots (rpc, web pivot model, and more)

src/model.ts
```diff
@@ -587,13 +598,13 @@ export class Model extends EventBus<any> implements CommandDispatcher {
   private dispatchToHandlers(handlers: CommandHandler<Command>[], command: Command) {
     const isCommandCore = isCoreCommand(command);
     for (const handler of handlers) {
-      if (!isCommandCore && handler instanceof CorePlugin) {
+      if (!isCommandCore && this.corePlugins.includes(handler as any)) {
```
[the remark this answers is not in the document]
VincentSchippefilt: yes, I'm looking into it.

src/plugins/ui_core_views/header_sizes_ui.ts
LucasLefevre: I didn't know OffscreenCanvas were a thing

PR #4161, opened 2024-05-02
Master imp add formula vectorization laa

laa-odoo:
## Concerning performances:

I don't measure any significant change for the formulas dataset and for the array formulas dataset.

Regarding the addition of vectorized formulas, a large dataset was created too.
It has exactly the same data as in the formulas data set. See the difference above : 

| makeLargeDataset                    | Evaluate All Cells | Model Created In |
|---------------------|--------------------|------------------|
| **formulas**        | ~ 270 ms                | ~ 2000 ms             |
| **vectorizedFormulas** | ~ 350 ms                | ~ 380 ms              |

Task: : [3356482](https://www.odoo.com/web#id=3356482&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/formulas/compiler.ts
+++ b/src/formulas/compiler.ts
@@ -10,7 +10,8 @@ import { rangeTokenize } from "./range_tokenizer";
 
 const functions = functionRegistry.content;
 
-const OPERATOR_MAP = {
+export const OPERATOR_MAP = {
+  // export for test
   "=": "EQ",
   "+": "ADD",
   "-": "MINUS",
@@ -25,7 +26,8 @@ const OPERATOR_MAP = {
   "&": "CONCATENATE",
 };
 
-const UNARY_OPERATOR_MAP = {
+export const UNARY_OPERATOR_MAP = {
+  // export for test
   "-": "UMINUS",
   "+": "UPLUS",
   "%": "UNARY.PERCENT",
--- a/src/functions/index.ts
+++ b/src/functions/index.ts
@@ -83,14 +83,27 @@ export class FunctionRegistry extends Registry<FunctionDescription> {
     }
     const descr = addMetaInfoFromArg(addDescr);
     validateArguments(descr.args);
-    this.mapping[name] = addErrorHandling(
-      addInputHandling(descr, addResultHandling(descr.compute, name)),
-      name
-    );
+    this.mapping[name] = createComputeFunction(descr, name);
     super.add(name, descr);
     return this;
   }
 }
+
+export const functionRegistry: FunctionRegistry = new FunctionRegistry();
+
+for (let category of categories) {
+  const fns = category.functions;
+  for (let name in fns) {
+    const addDescr = fns[name];
+    addDescr.category = addDescr.category || category.name;
+    name = name.replace(/_/g, ".");
+    functionRegistry.add(name, { isExported: false, ...addDescr });
+  }
+}
+
+//------------------------------------------------------------------------------
+// CREATE COMPUTE FUNCTION
+//------------------------------------------------------------------------------
 
 type VectorArgType = "horizontal" | "vertical" | "matrix";
 
@@ -98,10 +111,18 @@ const notAvailableError = new NotAvailableError(
   _t("Array arguments to [[FUNCTION_NAME]] are of different size.")
 );
 
-function addInputHandling(
+function createComputeFunction(
   descr: FunctionDescription,
-  compute: ComputeFunction<FPayload | Matrix<FPayload>>
-): ComputeFunction<FPayload | Matrix<FPayload>> {
+  functionName: string
+): ComputeFunction<Matrix<FPayload> | FPayload> {
+  function computeFunction(this: EvalContext, ...args: Arg[]): Matrix<FPayload> | FPayload {
+    try {
+      return computeWithVectorization.apply(this, args);
+    } catch (e) {
+      return handleError(e, functionName);
+    }
+  }
+
   function computeWithVectorization(
     this: EvalContext,
     ...args: Arg[]
@@ -155,7 +176,7 @@ function addInputHandling(
     }
 
     if (countVectorizableCol === 1 && countVectorizableRow === 1) {
-      return compute.apply(this, args);
+      return computeToFPayloadObject.apply(this, args);
     }
 
     const argsVector: (i: number, j: number) => Arg[] = (i, j) =>
@@ -176,7 +197,7 @@ function addInputHandling(
       if (col > vectorizableColLimit - 1 || row > vectorizableRowLimit - 1) {
         return notAvailableError;
       }
-      const matrixElm = compute.apply(this, argsVector(col, row));
+      const matrixElm = computeToFPayloadObject.apply(this, argsVector(col, row));
       // In the case where the user tries to vectorize arguments of an array formula, we will get an
       // array for every combination of the vectorized arguments, which will lead to a 3D matrix and
       // we won't be able to return the values.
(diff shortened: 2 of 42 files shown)


src/formulas/compiler.ts
```diff
@@ -22,7 +22,7 @@ const OPERATOR_MAP = {
   "<=": "LTE",
   "<": "LT",
   "^": "POWER",
-  "&": "CONCATENATE",
+  "&": "CONCAT",
```
rrahir, line 25: What is the concrete difference of behaviour of the 2 functions? (concat vs concatenate)

laa-odoo: `CONCATENATE` takes ranges and not single arguments as input. Vectorization is only possible with simple arguments.

Besides, that's probably why both exist, they both do the same thing, but one will work as an aggregator and the other will vectorize

src/functions/index.ts
```diff
@@ -121,10 +152,37 @@ function addInputHandling(
...
+            return arg;
+        }
+      });
+
+    return generateMatrix(countVectorizableCol, countVectorizableRow, (col, row) => {
+      if (col > vectorizableColLimit - 1 || row > vectorizableRowLimit - 1) {
+        return notAvailableError;
+      }
+      const matrixElm = compute.apply(this, argsVector(col, row));
+      // to avoid mismatching behavior when user try to vectorize arguments on formula that spread
```
LucasLefevre, line 179: I'm not sure to understand what this means

src/functions/index.ts
```diff
@@ -83,33 +83,64 @@ export class FunctionRegistry extends Registry<FunctionDescription> {
...
       if (isMatrix(arg) && !argDefinition.acceptMatrix) {
-        if (arg.length !== 1 || arg[0].length !== 1) {
-          throw new EvaluationError(
-            _t(
-              "Function [[FUNCTION_NAME]] expects the parameter '%s' to be a single value or a single cell reference, not a range.",
-              argDefinition.name
-            )
-          );
+        // if argDefinition does not accept a matrix but arg is still a matrix
+        // --> triggers the vectorization of the arguments
```
LucasLefevre, line 123:
```suggestion
        // --> triggers the arguments vectorization
```

src/formulas/compiler.ts
LucasLefevre, line 25: maybe a little test ? :roll_eyes:

src/functions/index.ts
```diff
@@ -83,33 +83,64 @@ export class FunctionRegistry extends Registry<FunctionDescription> {
...
+          vectorArgsType.push("horizontal");
+          countVectorizableCol = Math.max(countVectorizableCol, arg.length);
+          vectorizableColLimit = Math.min(vectorizableColLimit, arg.length);
+        } else if (arg[0].length !== 1) {
+          vectorArgsType.push("vertical");
+          countVectorizableRow = Math.max(countVectorizableRow, arg[0].length);
+          vectorizableRowLimit = Math.min(vectorizableRowLimit, arg[0].length);
+        } else {
+          vectorArgsType.push("scalar");
+          args[i] = arg[0][0];
```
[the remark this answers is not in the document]
anhe-odoo: No, we need to update it here as, if we only have 1x1 elements, we won't go through the `argsVector` function as `countVectorizableCol == countVectorizableRow == 1`, but we need to have only scalar values in the args then.

src/functions/index.ts
```diff
@@ -83,33 +83,64 @@ export class FunctionRegistry extends Registry<FunctionDescription> {
...
         }
-        args[i] = arg[0][0];
+      } else {
+        vectorArgsType.push("scalar");
```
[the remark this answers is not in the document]
anhe-odoo: What about something like this : 
```js
let vectorArgsType: VectorArgType[] | undefined = undefined;
for (let i = 0; i < args.length; i++) {
  const argDefinition = descr.args[descr.getArgToFocus(i + 1) - 1];
  const arg = args[i];

  if (isMatrix(arg) && !argDefinition.acceptMatrix) {
    // if argDefinition does not accept a matrix but arg is still a matrix
    // --> triggers the arguments vectorization
    const nColumns = arg.length;
    const nRows = arg[0].length;
    if (nColumns !== 1 || nRows !== 1) {
      vectorArgsType ??= new Array(args.length);
      if (nColumns !== 1 && nRows !== 1) {
        vectorArgsType[i] = "matrix";
        countVectorizableCol = Math.max(countVectorizableCol, nColumns);
        countVectorizableRow = Math.max(countVectorizableRow, nRows);
        vectorizableColLimit = Math.min(vectorizableColLimit, nColumns);
        vectorizableRowLimit = Math.min(vectorizableRowLimit, nRows);
      } else if (nColumns !== 1) {
        vectorArgsType[i] = "horizontal";
        countVectorizableCol = Math.max(countVectorizableCol, nColumns);
        vectorizableColLimit = Math.min(vectorizableColLimit, nColumns);
      } else if (nRows !== 1) {
        vectorArgsType[i] = "vertical";
        countVectorizableRow = Math.max(countVectorizableRow, nRows);
        vectorizableRowLimit = Math.min(vectorizableRowLimit, nRows);
      }
    } else {
      args[i] = arg[0][0];
    }
  }
```

Then, if we only have 1x1 elements (like in the base case, where we use reference to cell as parameters), we won't create the arrray too. The switch can stay as you proposed : 
```js
const argsVector: (i: number, j: number) => Arg[] = (i, j) =>
  args.map((arg, index) => {
    switch (vectorArgsType?.[index]) {
      case "matrix":
        return arg![i][j];
      case "horizontal":
        return arg![i][0];
      case "vertical":
        return arg![0][j];
      case undefined:
        return arg;
    }
  });
```

src/functions/index.ts
anhe-odoo: // In the case where the user tries to vectorize arguments of an array formula, we will get an
// array for every combination of the vectorized arguments, which will lead to a 3D matrix and
// we won't be able to return the values.
// In this case, we keep the first element of each spreading part, just as Excel does, and
// create an array with these parts.
// For exemple, we have MUNIT(x) that return an unitary matrix of x*x. If we use it with a
// range, like MUNIT(A1:A2), we will get two unitary matrixes (one for the value in A1 and one
// for the value in A2). In this case, we will simply take the first value of each matrix and
// return the array [First value of MUNIT(A1), First value of MUNIT(A2)].

src/functions/index.ts
```diff
@@ -83,33 +83,65 @@ export class FunctionRegistry extends Registry<FunctionDescription> {
...
+    this.mapping[name] = addErrorHandling(
+      addInputHandling(descr, addResultHandling(descr.compute, name)),
+      name
+    );
     super.add(name, descr);
     return this;
   }
 }
 
+type VectorArgType = "scalar" | "horizontal" | "vertical" | "matrix";
```
LucasLefevre, line 95: remove scalar

tests/functions/vectorization.test.ts
```diff
@@ -121,4 +122,24 @@ describe("vectorization", () => {
...
+        functionDefinition.args.every((arg) =>
+          arg.type.every((t) => !t.startsWith("RANGE") && t !== "META")
+        )
+      );
+    }
+  });
+
+  test("unary operators should always accept vectors", () => {
+    // mean unary operators args should always be simple arg
+    for (let op in UNARY_OPERATOR_MAP) {
```
LucasLefevre, line 140: const

src/functions/index.ts
```diff
@@ -121,10 +153,44 @@ function addInputHandling(
...
+      // In the case where the user tries to vectorize arguments of an array formula, we will get an
+      // array for every combination of the vectorized arguments, which will lead to a 3D matrix and
+      // we won't be able to return the values.
+      // In this case, we keep the first element of each spreading part, just as Excel does, and
+      // create an array with these parts.
+      // For exemple, we have MUNIT(x) that return an unitary matrix of x*x. If we use it with a
+      // range, like MUNIT(A1:A2), we will get two unitary matrixes (one for the value in A1 and one
+      // for the value in A2). In this case, we will simply take the first value of each matrix and
+      // return the array [First value of MUNIT(A1), First value of MUNIT(A2)].
+      return isMatrix(matrixElm) ? matrixElm[0][0] : matrixElm;
```
VincentSchippefilt, line 189: maybe add a test for this case. Coverage shows matrixElm[0][0] is never executed

VincentSchippefilt (approved):
I hereby officially approve merging the vectorisation.

LucasLefevre (approved):
yougotthis-martini

robodoo r+ rebase-ff

PR #4162, opened 2024-05-02
[REF] zones: improve performances on 'positions'

laa-odoo:
Avoid using the `sort` function on a list of two arguments, divide the computation time on the line of code by 100

Task: [3901818](https://www.odoo.com/web#id=3901818&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/zones.ts
+++ b/src/helpers/zones.ts
@@ -476,14 +476,24 @@ export function isOneDimensional(zone: Zone): boolean {
  */
 export function positions(zone: Zone): Position[] {
   const positions: Position[] = [];
-  const [left, right] = [zone.right, zone.left].sort((a, b) => a - b);
-  const [top, bottom] = [zone.top, zone.bottom].sort((a, b) => a - b);
+
+  const { left, right, top, bottom } = reorderZone(zone);
   for (const col of range(left, right + 1)) {
     for (const row of range(top, bottom + 1)) {
       positions.push({ col, row });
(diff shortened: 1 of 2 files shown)


src/helpers/zones.ts
```diff
@@ -476,8 +476,8 @@ export function isOneDimensional(zone: Zone): boolean {
  */
 export function positions(zone: Zone): Position[] {
   const positions: Position[] = [];
-  const [left, right] = [zone.right, zone.left].sort((a, b) => a - b);
-  const [top, bottom] = [zone.top, zone.bottom].sort((a, b) => a - b);
+  const [left, right] = zone.left <= zone.right ? [zone.left, zone.right] : [zone.right, zone.left];
+  const [top, bottom] = zone.top <= zone.bottom ? [zone.top, zone.bottom] : [zone.bottom, zone.top];
```
LucasLefevre, line 480: We could even do something  better to avoid creating new arrays:
A helper function `orderZone` or something like that. That would reorder the zone, or return the original zone object if it's already ordered (which should be most of the time)
```suggestion
  const { left, right, top, bottom } = orderZone(zone);
```

LucasLefevre (approved):
robodoo r+

PR #4170, opened 2024-05-06
[IMP] clipboard: paste conditional format, data validation & merge content when original content is deleted

Rachico:
### [IMP] clipboard: paste conditional format, data validation & merge content when original content is deleted

Problem
-----
Before this commit, when copy/pasting cells  with CF/DV/merge, if we delete the CF/DV/merge before pasting the cells, the paste content will not contain the previously copied content.

Solution
-----
This commit solves the issue by adding  new attributes cfRules, merges, dataValidationRules in the clipboard content and using the content saved in this key to re-create the features.

Task: [3870289](https://www.odoo.com/web#id=3870289&cids=1&menu_id=4722&action=333&active_id=2328&model=project.task&view_type=form)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/clipboard_handlers/data_validation_clipboard.ts
+++ b/src/clipboard_handlers/data_validation_clipboard.ts
@@ -1,4 +1,4 @@
-import { UuidGenerator, deepEquals, isInside, positionToZone, recomputeZones } from "../helpers";
+import { UuidGenerator, deepEquals, positionToZone, recomputeZones } from "../helpers";
 import {
   CellPosition,
   ClipboardCellData,
@@ -6,18 +6,24 @@ import {
   ClipboardPasteTarget,
   DataValidationRule,
   HeaderIndex,
+  Maybe,
   UID,
   Zone,
 } from "../types";
 import { AbstractCellClipboardHandler } from "./abstract_cell_clipboard_handler";
 
-type ClipboardContent = {
-  cellPositions: CellPosition[][];
-};
+interface ClipboardDataValidationRule {
+  position: CellPosition;
+  rule: Maybe<DataValidationRule>;
+}
+
+interface ClipboardContent {
+  dvRules: Maybe<ClipboardDataValidationRule>[][];
+}
 
 export class DataValidationClipboardHandler extends AbstractCellClipboardHandler<
   ClipboardContent,
-  CellPosition
+  Maybe<ClipboardDataValidationRule>
 > {
   private readonly uuidGenerator = new UuidGenerator();
 
@@ -28,13 +34,18 @@ export class DataValidationClipboardHandler extends AbstractCellClipboardHandler
 
     const { rowsIndexes, columnsIndexes } = data;
     const sheetId = data.sheetId;
-    const cellPositions = rowsIndexes.map((row) =>
-      columnsIndexes.map((col) => ({ col, row, sheetId }))
-    );
+    const dvRules: Maybe<ClipboardDataValidationRule>[][] = [];
 
-    return {
-      cellPositions,
-    };
+    for (const row of rowsIndexes) {
+      const dvRuleInRow: Maybe<ClipboardDataValidationRule>[] = [];
+      for (const col of columnsIndexes) {
+        const position = { sheetId, col, row };
+        const rule = this.getters.getValidationRuleForCell(position);
+        dvRuleInRow.push({ position, rule });
+      }
+      dvRules.push(dvRuleInRow);
+    }
+    return { dvRules };
   }
 
   paste(
@@ -42,7 +53,7 @@ export class DataValidationClipboardHandler extends AbstractCellClipboardHandler
     clippedContent: ClipboardContent,
     options?: ClipboardOptions
   ) {
-    if (!clippedContent?.cellPositions) {
+    if (!clippedContent?.dvRules) {
       return;
     }
     if (options?.pasteOption) {
@@ -55,7 +66,7 @@ export class DataValidationClipboardHandler extends AbstractCellClipboardHandler
     const sheetId = target.sheetId;
 
     if (!options?.isCutOperation) {
-      this.pasteFromCopy(sheetId, zones, clippedContent.cellPositions);
+      this.pasteFromCopy(sheetId, zones, clippedContent.dvRules);
     } else {
       this.pasteFromCut(sheetId, zones, clippedContent);
     }
@@ -63,7 +74,7 @@ export class DataValidationClipboardHandler extends AbstractCellClipboardHandler
 
   private pasteFromCut(sheetId: UID, target: Zone[], content: ClipboardContent) {
     const selection = target[0];
-    this.pasteZone(sheetId, selection.left, selection.top, content.cellPositions, {
+    this.pasteZone(sheetId, selection.left, selection.top, content.dvRules, {
       isCutOperation: true,
     });
   }
@@ -72,10 +83,10 @@ export class DataValidationClipboardHandler extends AbstractCellClipboardHandler
     sheetId: UID,
     col: HeaderIndex,
     row: HeaderIndex,
-    positions: CellPosition[][],
+    dvRules: Maybe<ClipboardDataValidationRule>[][],
     clipboardOptions?: ClipboardOptions
   ) {
-    for (const [r, rowCells] of positions.entries()) {
+    for (const [r, rowCells] of dvRules.entries()) {
       for (const [c, origin] of rowCells.entries()) {
         const position = { col: col + c, row: row + r, sheetId };
         this.pasteDataValidation(origin, position, clipboardOptions?.isCutOperation);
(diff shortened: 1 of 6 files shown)


rrahir:
-> change the title as a [FIX] : the rationale is the following - we currently lose information when copy/pasting and the current result is different from the intended one.

src/clipboard_handlers/abstract_cell_clipboard_handler.ts
```diff
@@ -7,22 +7,28 @@ export class AbstractCellClipboardHandler<T, T1> extends ClipboardHandler<T> {
     return;
   }
 
-  pasteFromCopy(sheetId: UID, target: Zone[], content: T1[][], options?: ClipboardOptions) {
+  pasteFromCopy(
+    sheetId: UID,
+    target: Zone[],
+    content: T1[][],
+    options?: ClipboardOptions,
+    clipboardContent?: T
```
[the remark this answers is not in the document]
rrahir: After discussing with Anthony:
- the typing of T1 was set to be the minimal required information to manage the pasting
- We can change it to whatever we want. -> `content: T` is an acceptable change

src/clipboard_handlers/abstract_clipboard_handler.ts
```diff
@@ -16,12 +17,16 @@ export class ClipboardHandler<T> {
     return;
   }
 
-  paste(target: ClipboardPasteTarget, clippedContent: T, options: ClipboardOptions | undefined) {}
+  paste(
+    target: ClipboardPasteTarget,
+    clippedContent: T | MinimalClipboardData,
```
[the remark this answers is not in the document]
rrahir: After discussing with Anthony:
clippedContent was set as a generic such that every handler could define their own type and only deal with the data they wanted/cared about. It is a good approach and should stay that way.
On the other hand, nothing prevents us from having the type of the handler extending an interface representing the global keys found in the clipboard (more something like `ClipboardCellData` rahter that minimalClipboardData if possible).

rrahir (changes requested):
Nice work, the tests seem good :)
As discussed, there are some issues with the current state of the clipboard (regarding the development of this feature) which should be generally addressed before implementing your solution.

See the related task

tests/clipboard/clipboard_plugin.test.ts
```diff
@@ -1547,6 +1606,137 @@ describe("clipboard", () => {
     expect(getStyle(model, "C2")).toEqual({});
   });
 
+  test("copy cells with CF => remove origin CF => paste => it should paste with original CF", () => {
```
hokolomopo, line 1609: In my opinion, we should aim to make tests short and readable, rather than try to test absolutely everything. I don't think you need to test that every single command is successfully dispatched, nor do you need to test both than the cell content is pasted AND the CF is pasted AND that the Cf succefully is applied. That's a lot of bloating, making the test hard to understand at first glance IMO.

The test could be 15 lines instead of 60, only adding a CF => copy it => Remove the CF => paste => check the CF is pasted. We don't need to also test that we can paste cell values, nor that CF are correctly applied to the cells. It's tested somewhere else.

2 small general details:
- try to prettify thing on single lines when you can. `expect(getStyle(model, "A1")).toEqual({fillColor: "#00FF00" });` is short enough to be on a single line, no need to take 3 lines
- if you want to test that a command behaved correctly, avoid testing `toBeSuccessfullyDispatched` but rather the state of the model after the command. For example here `REMOVE_CONDITIONAL_FORMAT`would have been dispatched, but not actually removed the CF. But again, you don't need to test that every single command is successful, I'd say to only test it if it helps the readability of the test.

hokolomopo: (same comment for the next test)

hokolomopo (commented):
👋 

Soudns fine, with some minor details to fix. You should always avoid casting types, unless you really have no other choice. 🙂

src/clipboard_handlers/data_validation_clipboard.ts
```diff
@@ -84,40 +95,43 @@ export class DataValidationClipboardHandler extends AbstractCellClipboardHandler
...
+      if (isCutOperation) {
+        toRemoveZone.push(positionToZone(origin.position));
+      }
+      if (origin.position.sheetId === target.sheetId) {
+        const copyToRule = this.getDataValidationRuleToCopyTo(target.sheetId, rule, false);
+        this.adaptDataValidationRule(origin.position.sheetId, copyToRule, [zone], toRemoveZone);
+      } else {
+        /** If the origin sheet was deleted before pasting the data, no need to adapt the DV rules  */
+        if (this.getters.getSheetIds().includes(origin.position.sheetId)) {
+          this.adaptDataValidationRule(origin.position.sheetId, rule, [], toRemoveZone);
```
LucasLefevre, line 123: the zone should be removed from the original rule, not the one in the clipboard.
```suggestion
        const originRule = this.getters.getValidationRuleForCell(origin.position)
        if (originRule) {
          this.adaptDataValidationRule(origin.position.sheetId, originRule, [], toRemoveZone);
```
And it allows to remove `this.getters.getValidationRuleForCell(origin.position) ||` above

src/clipboard_handlers/data_validation_clipboard.ts
```diff
@@ -84,40 +95,43 @@ export class DataValidationClipboardHandler extends AbstractCellClipboardHandler
...
-    }
-    const zone = positionToZone(target);
-    for (const range of rule.ranges) {
-      if (isInside(origin.col, origin.row, range.zone)) {
-        const toRemoveZone: Zone[] = [];
-        if (isCutOperation) {
-          toRemoveZone.push(positionToZone(origin));
+    if (origin) {
+      const zone = positionToZone(target);
+      const rule = this.getters.getValidationRuleForCell(origin.position) || origin.rule;
```
LucasLefevre, line 104: Could you remind me the reason why we use the current validation rule and not `origin.rule` ? I think it deserves a comment in the code

Edit: it can be removed (see comment below)
```suggestion
      const rule = origin.rule;
```

src/clipboard_handlers/merge_clipboard.ts
```diff
@@ -1,95 +1,89 @@
...
   UID,
-  Zone,
 } from "../types";
 import { AbstractCellClipboardHandler } from "./abstract_cell_clipboard_handler";
 
-interface ClipboardContent {
-  cells: ClipboardCell[][];
-  zones: Zone[];
-  sheetId: UID;
+interface ClipboardContent extends MinimalClipboardData {
```
LucasLefevre, line 14: not needed
```suggestion
interface ClipboardContent {
```

src/clipboard_handlers/merge_clipboard.ts
```diff
@@ -1,95 +1,89 @@
...
+    return { merges };
   }
 
   /**
    * Paste the clipboard content in the given target
    */
   paste(target: ClipboardPasteTarget, content: ClipboardContent, options: ClipboardOptions) {
-    if (options?.isCutOperation || !("zones" in target) || !target.zones.length) {
+    if (
+      !content.merges ||
```
LucasLefevre, line 47: AFAICT, `content.merges` is always truthy. Remember that an empty array `[]` is truthy in javascript :wink: 

In this case, I would not add any check here. If the array is empty, it won't do anything anyway.

src/plugins/ui_stateful/clipboard.ts
```diff
@@ -31,7 +31,7 @@ interface InsertDeleteCellsTargets {
   paste: Zone[];
 }
 
-type MinimalClipboardData = {
+export type MinimalClipboardData = {
```
LucasLefevre, line 34: not needed
```suggestion
type MinimalClipboardData = {
```

tests/clipboard/clipboard_plugin.test.ts
```diff
@@ -1557,6 +1634,114 @@ describe("clipboard", () => {
...
+    const model = new Model();
+    setCellContent(model, "A1", "1");
+    setCellContent(model, "A2", "0");
+    setCellContent(model, "A3", "1");
```
LucasLefevre, line 1641: The test doesn't need content in the cell to be relevant. Keep it short and minimal
```suggestion
```

tests/clipboard/clipboard_plugin.test.ts
```diff
@@ -1557,6 +1634,114 @@ describe("clipboard", () => {
...
+      id: "cfId",
+      sheetId,
+    });
+    paste(model, "D1");
+    expect(model.getters.getConditionalFormats(model.getters.getActiveSheetId())).toMatchObject([
+      { ranges: ["D1:D3"], rule: cf.rule },
+    ]);
+    expect(getStyle(model, "D1")).toEqual({ fillColor: "#00FF00" });
+    expect(getStyle(model, "D2")).toEqual({});
+    expect(getStyle(model, "D3")).toEqual({ fillColor: "#00FF00" });
```
LucasLefevre, line 1660: The computed style is already tested elsewhere. You only care about the copy/paste behavior here, not the effect of the CF
```suggestion
```

tests/clipboard/clipboard_plugin.test.ts
```diff
@@ -1557,6 +1634,114 @@ describe("clipboard", () => {
...
+  });
+
+  test("copy cells with multiple independent CF => remove all copied CF => paste => it should paste with all original CF in the correct positions", () => {
+    const model = new Model();
+    setCellContent(model, "A1", "1");
+    setCellContent(model, "A2", "0");
+    setCellContent(model, "A3", "1");
+    setCellContent(model, "C1", "1");
+    setCellContent(model, "C2", "0");
+    setCellContent(model, "C3", "1");
```
LucasLefevre, line 1670:
```suggestion
```

tests/clipboard/clipboard_plugin.test.ts
```diff
@@ -1557,6 +1634,114 @@ describe("clipboard", () => {
...
+    expect(model.getters.getConditionalFormats(sheetId)).toMatchObject([
+      { ranges: ["E1:E3"], rule: cf1.rule },
+      { ranges: ["G1:G3"], rule: cf2.rule },
+    ]);
+    expect(getStyle(model, "E1")).toEqual({ fillColor: "#00FF00" });
+    expect(getStyle(model, "E2")).toEqual({});
+    expect(getStyle(model, "E3")).toEqual({ fillColor: "#00FF00" });
+    expect(getStyle(model, "G1")).toEqual({ fillColor: "#0000FF" });
+    expect(getStyle(model, "G2")).toEqual({});
+    expect(getStyle(model, "G3")).toEqual({ fillColor: "#0000FF" });
```
LucasLefevre, line 1703:
```suggestion
```

tests/clipboard/clipboard_plugin.test.ts
```diff
@@ -1557,6 +1634,114 @@ describe("clipboard", () => {
...
+  });
+
+  test("copy cells with multiple independent CF => remove origin sheet => paste => it should paste with all original CF in the correct positions", () => {
+    const model = new Model();
+    setCellContent(model, "A1", "1");
+    setCellContent(model, "A2", "0");
+    setCellContent(model, "A3", "1");
+    setCellContent(model, "C1", "1");
+    setCellContent(model, "C2", "0");
+    setCellContent(model, "C3", "1");
```
LucasLefevre, line 1713:
```suggestion
```

tests/clipboard/clipboard_plugin.test.ts
```diff
@@ -1557,6 +1634,114 @@ describe("clipboard", () => {
...
+    expect(model.getters.getConditionalFormats(newSheetId)).toMatchObject([
+      { ranges: ["E1:E3"], rule: cf1.rule },
+      { ranges: ["G1:G3"], rule: cf2.rule },
+    ]);
+    expect(getStyle(model, "E1")).toEqual({ fillColor: "#00FF00" });
+    expect(getStyle(model, "E2")).toEqual({});
+    expect(getStyle(model, "E3")).toEqual({ fillColor: "#00FF00" });
+    expect(getStyle(model, "G1")).toEqual({ fillColor: "#0000FF" });
+    expect(getStyle(model, "G2")).toEqual({});
+    expect(getStyle(model, "G3")).toEqual({ fillColor: "#0000FF" });
```
LucasLefevre, line 1742:
```suggestion
```

LucasLefevre (commented):
It's looking good :+1: 

Some cleanup suggestions :wink:

src/helpers/clipboard/clipboard_helpers.ts
```diff
@@ -46,11 +46,9 @@ function splitZoneForPaste(selection: Zone, splitWidth: number, splitHeight: num
 /**
  * Compute the complete zones where to paste the current clipboard
  */
-export function getPasteZones<T>(target: Zone[], content: T[][]): Zone[] {
-  if (!content.length || !content[0].length) {
+export function getPasteZones<T>(target: Zone[], height: number, width: number): Zone[] {
```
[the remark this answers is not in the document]
Rachico: Because we actually only need the height & width of the content.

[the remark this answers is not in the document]
LucasLefevre: But it makes the code everywhere else longer and more prone to errors.
`content.length, content[0].length` instead of just `content`

Also the change is unrelated to the commit.
=> revert this part :wink:

src/clipboard_handlers/merge_clipboard.ts
Rachico: The paste function here handles the pasting of merges. It doesn't make sense to move forward if they are no merges 🙂

LucasLefevre: > It doesn't make sense to move forward if they are no merges

but checking adds (useless) code. You don't add a check before every for loop to check if the iterated array is empty, right ? That's the same here. You can just let the flow go and it's perfectly fine.

No check is less code. If there's less code, it's easier to read and less bug prone :wink:

LucasLefevre (approved):
Thanks :)
robodoo r+

PR #4171, opened 2024-05-06
[REM] merge: remove useless topLeft attribute from Merge interface

Rachico:
## Description:

merge: remove useless topLeft attribute from Merge interface.

Context
-----
The "topLeft" attribute in the Merge interface was used as a shortcut to get the top left position of the merge zone. However, this was creating redundant code.

Task: [3912050](https://www.odoo.com/web#cids=1&menu_id=4722&action=333&active_id=2328&model=project.task&view_type=form&id=3912050)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/core/merge.ts
+++ b/src/plugins/core/merge.ts
@@ -254,11 +254,14 @@ export class MergePlugin extends CorePlugin<MergeState> implements MergeState {
   }
 
   getMainCellPosition(position: CellPosition): CellPosition {
-    if (!this.isInMerge(position)) {
-      return position;
-    }
-    const mergeTopLeftPos = this.getMerge(position)!.topLeft;
-    return { sheetId: position.sheetId, col: mergeTopLeftPos.col, row: mergeTopLeftPos.row };
+    const mergeZone = this.getMerge(position);
+    return mergeZone
+      ? {
+          sheetId: position.sheetId,
+          col: mergeZone.left,
+          row: mergeZone.top,
+        }
+      : position;
   }
 
   isMergeHidden(sheetId: UID, merge: Merge): boolean {
@@ -539,7 +542,6 @@ function exportMerges(merges: Record<number, Range | undefined>): string[] {
 function rangeToMerge(mergeId: number, range: Range): Merge {
   return {
     ...range.zone,
-    topLeft: { col: range.zone.left, row: range.zone.top },
     id: mergeId,
   };
 }
--- a/tests/model/model_import_export.test.ts
+++ b/tests/model/model_import_export.test.ts
@@ -486,7 +486,7 @@ describe("Import", () => {
     expect(Object.keys(getMerges(model))).toHaveLength(0);
     activateSheet(model, sheet1);
     expect(Object.keys(getMerges(model))).toHaveLength(1);
-    expect(Object.values(getMerges(model))[0].topLeft).toEqual(toCartesian("A2"));
+    expect(getMerges(model)[1]).toMatchObject(toZone("A2:B2"));
   });
 
   test("can import cell without content", () => {
--- a/src/plugins/ui_feature/autofill.ts
+++ b/src/plugins/ui_feature/autofill.ts
@@ -443,7 +443,7 @@ export class AutofillPlugin extends UIPlugin {
       }
     }
     const originMerge = this.getters.getMerge(originPosition);
-    if (originMerge?.topLeft.col === originCol && originMerge?.topLeft.row === originRow) {
+    if (originMerge?.left === originCol && originMerge?.top === originRow) {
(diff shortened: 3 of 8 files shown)


src/plugins/core/merge.ts
```diff
@@ -257,8 +257,12 @@ export class MergePlugin extends CorePlugin<MergeState> implements MergeState {
     if (!this.isInMerge(position)) {
       return position;
     }
-    const mergeTopLeftPos = this.getMerge(position)!.topLeft;
-    return { sheetId: position.sheetId, col: mergeTopLeftPos.col, row: mergeTopLeftPos.row };
+    const mergeZone = this.getMerge(position);
+    return {
+      sheetId: position.sheetId,
+      col: mergeZone?.left as HeaderIndex,
+      row: mergeZone?.top as HeaderIndex,
```
VincentSchippefilt, line 264:
```suggestion
      col: mergeZone!.left,
      row: mergeZone!.top,
```
No need to re-cast into HeaderIndex, you can tell typescript it has to have a value with `!`

VincentSchippefilt (commented):
small nit picking, otherwise all good. Thanks

src/plugins/core/merge.ts
hokolomopo, line 264: As a general rule we should avoid casting unless there is no other choice.

An `if(mergeZone)` just above instead of the `if(this.isInMerge)` would fix that.

Note that this wasn't needed because we had an `!` before, which is something we should also avoid

src/plugins/core/merge.ts
```diff
@@ -257,8 +257,12 @@ export class MergePlugin extends CorePlugin<MergeState> implements MergeState {
...
+    };
```
VincentSchippefilt, line 265:
```suggestion
    const mergeZone = this.getMerge(position);

    return mergeZone ? {
      sheetId: position.sheetId,
      col: mergeZone.left,
      row: mergeZone.top
    } : position;
```

src/plugins/core/merge.ts
```diff
@@ -254,11 +254,14 @@ export class MergePlugin extends CorePlugin<MergeState> implements MergeState {
...
-      return position;
-    }
-    const mergeTopLeftPos = this.getMerge(position)!.topLeft;
-    return { sheetId: position.sheetId, col: mergeTopLeftPos.col, row: mergeTopLeftPos.row };
+    const mergeZone = this.getMerge(position);
+    return mergeZone
+      ? {
+          sheetId: position.sheetId,
+          col: mergeZone?.left as HeaderIndex,
+          row: mergeZone?.top as HeaderIndex,
```
LucasLefevre, line 262: casting is useless here (and should always be avoided when possible).
```suggestion
          sheetId: position.sheetId,
          col: mergeZone.left,
          row: mergeZone.top,
```

tests/model/model_import_export.test.ts
```diff
@@ -486,7 +486,8 @@ describe("Import", () => {
     expect(Object.keys(getMerges(model))).toHaveLength(0);
     activateSheet(model, sheet1);
     expect(Object.keys(getMerges(model))).toHaveLength(1);
-    expect(Object.values(getMerges(model))[0].topLeft).toEqual(toCartesian("A2"));
+    const mergeZone = getMerges(model)[1];
+    expect({ col: mergeZone.left, row: mergeZone.top }).toEqual(toCartesian("A2"));
```
LucasLefevre, line 490: Check the entire merge and not only the first one
```suggestion
    expect(getMerges(model)).toEqual([toZone("A2:B2")]);
```

src/plugins/core/merge.ts
pro-odoo: History has shown us that each time we introduce a `!`, a future bug is also introduce. `!` is a way to tell typescript to have truth in you. It could be correct now, but could become incorrect in the future and we will not notice it. The right way to handle this is to check that `mergeZone` is defined, and you did it correctly. So, simply remove the cast and optional:
```
col: mergeZone.left,
row: mergeZone.top,
```

pro-odoo: Note that `!` can be used in tests

tests/model/model_import_export.test.ts
Rachico: @LucasLefevre it's not a list, it's an object with the ids of merges as keys
image

LucasLefevre: Ah ok,
then you can still check the entire merge
```ts
expect(mergeZone).toEqual(toZone("A2:B2"));
```

rrahir:
@robodoo r+

PR #4176, opened 2024-05-07
[IMP] evaluation: addDependencies now evaluate cell

anhe-odoo:
## Task Description

When adding a new dependency to a cell (like in the INDIRECT formula), we have to make sure the dependency have been evaluated before computing the formula result. To make sure it has at the first evaluation cycle, we need to force the evaluation of the dependency. This was done by triggering the evaluation of a new formula, by evaluating a string containing only a reference to the dependency (the XC path). This PR update the addDependency method to directly evaluate (if needed) the new dependency when adding it, in the evaluator, instead of using the hack of triggering an evaluation after.

## Related Task:

- Task: [3916213](https://www.odoo.com/web#cids=1&menu_id=4720&action=333&active_id=2328&model=project.task&view_type=form&id=3916213)

## Review Checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_core_views/cell_evaluation/evaluator.ts
+++ b/src/plugins/ui_core_views/cell_evaluation/evaluator.ts
@@ -89,8 +89,9 @@ export class Evaluator {
     this.formulaDependencies().addDependencies(position, dependencies);
   }
 
-  addDependencies(position: CellPosition, dependencies: Range[]) {
+  private addDependencies(position: CellPosition, dependencies: Range[]) {
     this.formulaDependencies().addDependencies(position, dependencies);
+    this.computeAndSave(position);
   }
 
   private updateCompilationParameters() {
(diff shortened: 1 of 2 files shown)


src/plugins/ui_core_views/cell_evaluation/evaluator.ts
```diff
@@ -91,6 +91,7 @@ export class Evaluator {
 
   addDependencies(position: CellPosition, dependencies: Range[]) {
```
LucasLefevre, line 92: could you make this private while we are at it ? :)

LucasLefevre (commented):
Ouuh nice :+1:

LucasLefevre:
robodoo r+

PR #4182, opened 2024-05-07
[IMP] dashboard: automatically reorganize a sheet

hokolomopo:
## [IMP] dashboard: automatically reorganize a sheet
This commit aims to simplify the process of creation of a dashboard
with two new features:

1) A context menu item to send figures and ranges to another sheet
2) A context menu to reorganize a sheet in a "dashboard-y" way.

Both features are implemented in the `ReorganizeSheetPlugin`. Sending
a figure/range to a sheet works by finding the bottom-most used cell
in the sheet and placing the figure/range below it. "Used" cells are
cells that contains some content/a merge/a table/overlap a figure.

Reorganizing a sheet works by:
- put the scorecards at the top of the sheet
- put the figures below the scorecards
- put the cell "clusters" (zones of contiguous cells) below them

Technically the reorganization is done by creating new rows at the
bottom of the sheet, cut/paste the clusters correctly ordered in
these rows, and then delete the original rows. See Odoo task's pad
for why we do it this way.

## [IMP] mergeContiguousZones: add test & fix diagonal
This commit moves the tests of mergeContiguousZones from Odoo
to o-spreadsheet.

It also fixes a bug where the algorithm was merging zones that were
diagonal to each other.

## [REF] clipboard: decouple clipboards and active sheet
This commit make the clipboards handlers independent from the active
sheet, thus allowing to use them to copy/paste things in any sheet.

## [IMP] clipboard: insert new rows when pasting charts
Currently, when pasting a chart at the edges of a sheet, the pasted
chart is shifted so that it doesn't go out of the sheet.

This commit changes the behaviour so that now the chart is always
pasted with its top-left corner at the top-left corner of the selection,
and insert new rows/columns if needed so that the chart fits in the
sheet.

Also take the occasion to factorize the code in common between the
chart clipboard and the image clipboard into the figure clipboard.

## [IMP] clipboard: instanciate clipboards only once
This commit make the different clipboards instance attributes of the
Clipboard plugin, instead of being created each time a paste or copy
action is done.

This:
- is a bit faster and less wasteful
- allows the clipboard instances to be exposed to other plugins
- will be necessary for a future refactoring of the clipboard plugin
where each clipboard will have its own copied data, instead of
having a big copied data object shared between all clipboards.

## [REF] clipboard: expose paste logic
This commit extract the logic of the paste operation of the
clipboard plugin into a new getter (`getPasteTarget`) and a new command
(`EXPAND_SHEET_FOR_ZONE`) so that it can be reused in other parts of the
code.

## [FIX] evaluation: `getEvaluatedCells` with spreading cells
The getter `getEvaluatedCells` does't return spreaded cells. It
also returns empty evaluated cells that have some style.

This is not the expected behaviour. `getEvaluatedCells` should return
all the non-empty evaluated cells of a sheet.

Task: : [3869372](https://www.odoo.com/web#id=3869372&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/actions/menu_items_actions.ts
```diff
@@ -562,3 +562,36 @@ export const CREATE_OR_REMOVE_FILTER_ACTION: ActionSpec = {
...
+    name: _t("New sheet"),
+    id: "new_sheet",
+    icon: "o-spreadsheet-Icon.PLUS",
+    execute: (env) => {
+      const position = env.model.getters.getSheetIds().length;
+      const sheetId = env.model.uuidGenerator.uuidv4();
+      const name = env.model.getters.getNextSheetName(_t("Sheet"));
+
+      env.model.dispatch("CREATE_SHEET", { sheetId, position, name });
+      sendItemToSheetCallback(env, sheetId);
```
LucasLefevre, line 593: That makes 2 history steps.
maybe `sheetId` could be optional in commands `SEND_SELECTION_TO_SHEET` and  `SEND_FIGURE_TO_SHEET` and it would create a new sheet in this case?

src/plugins/ui_feature/reorganize_sheet.ts
```diff
@@ -0,0 +1,572 @@
...
+  Pixel,
+  Position,
+  UID,
+  Zone,
+} from "../../types/index";
+import { UIPlugin } from "../ui_plugin";
+
+const INSERT_BOTTOM_MARGIN = 2; // in number of rows
+
+//ADRM TODO: send to sheet: manage borders
```
LucasLefevre, line 29: :eyes: borders... :roll_eyes:

src/registries/figure_registry.ts
```diff
@@ -146,5 +149,21 @@ function getDeleteMenuItem(
...
+    separator: true,
+  };
+}
+
+function getSendFigureToSheetMenuItem(figureId: UID): ActionSpec {
+  return {
+    sequence: 15,
+    id: "send_to_sheet",
+    name: _t("Send to sheet"),
+    icon: "o-spreadsheet-Icon.CUT", // ADRM TODO: icon
```
LucasLefevre, line 161: I kinda like the cut icon for this

src/plugins/ui_feature/reorganize_sheet.ts
hokolomopo: It was done (and tested), I just forgot the TODO

PR #4186, opened 2024-05-08
[FIX] functions: cast to matrix

LucasLefevre:
`=SUMIF(CHOOSECOLS(A1,-1),A2)` crashes because `visitMatchingRanges` blindly index `args[0][0]` even if `args[0]` is not a matrix. In this case, it's an error object because `CHOOSECOLS(A1,-1)` is not valid.

Task: 3918882

## Description:

description of this task, what is implemented and why it is implemented that way.

Task: : [TASK_ID](https://www.odoo.com/web#id=TASK_ID&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] feature is organized in plugin, or UI components
- [ ] support of duplicate sheet (deep copy)
- [ ] in model/core: ranges are Range object, and can be adapted (adaptRanges)
- [ ] in model/UI: ranges are strings (to show the user)
- [ ] undo-able commands (uses this.history.update)
- [ ] multiuser-able commands (has inverse commands and transformations where needed)
- [ ] new/updated/removed commands are documented
- [ ] exportable in excel
- [ ] translations (\_t("qmsdf %s", abc))
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes
- [ ] doc is rebuild (npm run doc)
- [ ] status is correct in Odoo

PR #4188, opened 2024-05-08
[FIX] Notificiation: set a default notification store

rrahir:
Following the introduction of stores, the minimal setup has evolved as it requires to instanciate a some stores before the creation of a spreadsheet app. The documentation was not updated accordingly.

More specifically, the only store that we required to instantiate manually was the `NotificationStore` and we had to add it to the App env as well. These last step are a pain for developpers that only want to set a quick and minimal integration.

This commit sets default notification methods (raiseError, notifyUser, askConfirmation) which can be updated by calling `NotificationStore.updateNotificationCallbacks`.

Task: 3919166

## Description:

Task: : [3919166](https://www.odoo.com/web#id=3919166&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/doc/integrating/integration.md
+++ b/doc/integrating/integration.md
@@ -9,13 +9,18 @@ const { Spreadsheet, Model } = o_spreadsheet;
 
 const model = new Model();
 const templates = await(await fetch("../dist/o_spreadsheet.xml")).text();
-const env = {
-  notifyUser: () => window.alert(content),
-  askConfirmation: (message, confirm, cancel) => confirm(),
-};
 const app = new owl.App(Spreadsheet, {
-  props: { model },
-  env,
+  props: {
+    model,
+    // optionals
+    notifyUser: () => window.alert(content),
+    askConfirmation: (message, confirm, cancel) => window.confirm(message),
+    raiseError: (content, callback) => {
+      window.alert(content);
+      callback?.();
+    },
+  },
+  {},
   templates,
 });
 app.mount(document.body);
@@ -25,8 +30,22 @@ app.mount(document.body);
 
 Spreadsheet component takes the following props:
 
+**required**:
+
 - `model`
   The spreadsheet model to be used with the component.
+
+**optional**:
+
+- `notifyUser`
+  A function used to notify the user. It supports several levels of severity as well as a sticky behaviour.
+  See its [`interface`](../../src/types/env.ts#L15)
+- `askConfirmation`
+  A function used to ask the user confirmation before applying a callback
+- `raiseError`
+  A function to warn the user when a manipulation error occurs.
+
+The optional props should implement [`NotificationStoreMethods`](../../src/stores/notification_store.ts#L3).
 
 ## Model creation
 
@@ -136,3 +155,7 @@ const model = new Model(data, {
   },
 });
 ```
+
+## Managing application state with stores
+
+See [Managing application state with stores](/src/store_engine/README.md)
--- a/src/components/spreadsheet/spreadsheet.ts
+++ b/src/components/spreadsheet/spreadsheet.ts
@@ -28,12 +28,12 @@ import {
   SEPARATOR_COLOR,
   TOPBAR_HEIGHT,
 } from "../../constants";
-import { batched } from "../../helpers";
+import { batched, deepEquals } from "../../helpers";
 import { ImageProvider } from "../../helpers/figures/images/image_provider";
 import { Model } from "../../model";
 import { Store, useStore, useStoreProvider } from "../../store_engine";
 import { ModelStore } from "../../stores";
-import { NotificationStore } from "../../stores/notification_store";
+import { NotificationStore, NotificationStoreMethods } from "../../stores/notification_store";
 import { _t } from "../../translation";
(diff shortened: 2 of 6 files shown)


src/stores/notification_store.ts
```diff
@@ -1,9 +1,28 @@
...
-  raiseError: (text: string, callback?: () => void) => any;
-  askConfirmation: (content: string, confirm: () => any, cancel?: () => any) => any;
+interface NotificationStoreMethods {
+  notifyUser: (notification: InformationNotification) => void;
+  raiseError: (text: string, callback?: () => void) => void;
+  askConfirmation: (content: string, confirm: () => void, cancel?: () => void) => void;
+}
+
+export class NotificationStore {
+  notifyUser: NotificationStoreMethods["notifyUser"] = (notification) => alert(notification);
```
hokolomopo, line 10:
```suggestion
  notifyUser: NotificationStoreMethods["notifyUser"] = (notification) => window.alert(notification);
```

for consistency (and ease of reading)

hokolomopo (commented):
This seems like a very good step in the right direction 😊

doc/integrating/integration.md
```diff
@@ -136,3 +132,43 @@ const model = new Model(data, {
...
+class Wrapper extends Component {
+  static template = xml/* xml */ `<div><Spreadsheet model="model"/></div>`;
+  static components = { Spreadsheet };
+
+  model = new Model();
+
+  setup() {
+    useStoreProvider();
+    const notificationStore = useStore(NotificationStore);
+    notificationStore.update({
```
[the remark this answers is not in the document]
rrahir: this would be circling back to where we were 3 years ago. I don't really mind (agree with the ease of use) but I have 2 points:
- The store approach allows to have at least one concrete example on how to interract with stores for external users that want to set an integration
- Back in the days, we wanted to alleviate the number of props and it feels way better to only have to provide model (look at version 15.0 ... 
image
)

doc/integrating/integration.md
```diff
@@ -25,9 +30,23 @@ app.mount(document.body);
...
 
+**required**:
+
 - `model`
   The spreadsheet model to be used with the component.
 
+**optional**:
+
+- `notifyUser`
+  A function used to notify the user. It supports several levels of severity as well as a sticky behaviour.
```
hokolomopo, line 41: seems strange to say the it supports "several levels of severity as well as a sticky behaviour" since the default one only does a window.alert with no nuanced behaviour.

src/types/env.ts
```diff
@@ -18,13 +18,7 @@ export interface InformationNotification {
   sticky: boolean;
 }
 
-export interface SpreadsheetEnv {
-  notifyUser: (notification: InformationNotification) => any;
-  raiseError: (text: string, callback?: () => void) => any;
-  askConfirmation: (content: string, confirm: () => any, cancel?: () => any) => any;
```
hokolomopo, line 24: Do we really want to remove these methods from the env ? Leaving them in the SpreadsheetChildEnv doesn't cost much, and makes interacting with them a bit less annoying.

hokolomopo (commented):
Needs a small rebase 🙂

doc/integrating/integration.md
rrahir: the API has to support it.

src/types/env.ts
rrahir: @VincentSchippefilt  disagreed on that point and I don't have any strong opinion on this. The bad bard is that is had to become
```typescript
 notifyUser: (notification: InformationNotification) => this.notificationStore.notifyUser(notification);
```
Everything we put in the env is set in stone as we pass direct references. If we want the notifications, or the model to act as actual props, we should have (nearly) every entry onf the env being dynamic references and not static or binded as we have today.

src/components/spreadsheet/spreadsheet.ts
```diff
@@ -314,6 +317,12 @@ export class Spreadsheet extends Component<SpreadsheetProps, SpreadsheetChildEnv
         this.composerFocusStore.focusGridComposerCell(content),
     });
 
+    this.notificationStore.updateNotificationCallbacks({ ...this.props });
+
+    onWillUpdateProps((nextProps) => {
+      this.notificationStore.updateNotificationCallbacks({ ...nextProps });
```
LucasLefevre, line 323: I think this is called at every rendering, whether the props actually changed or not.
Should we check that they changed `this.props !== nextProps` ? (not sure the reference are still the same if nothing changed)

LucasLefevre (commented):
nice nice nice ! It have been done a long time ago :)

To quote what a wise man once said:

> the ref should be done in master https://github.com/odoo/o-spreadsheet/pull/4186#issuecomment-2133502716

And it really looks like a ref :smile:

rrahir:
@LucasLefevre Moved it to master and adapted the willUpdateprops hook
@hokolomopo I put the call to the env back even though I fear it could still disappear in the future. BUt at least if aleviates this diff :p

LucasLefevre (approved):
robodoo r+

PR #4205, opened 2024-05-14
[IMP] evaluation: prevent array formula spread on merge

dhrp-odoo:
## Description:

This PR ensures that array formulas do not spread when users merge cells containing them. Instead, a `#SPILL` error is triggered if users attempt to merge cells containing array formulas.

Task: : [3888401](https://www.odoo.com/web#id=3888401&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_core_views/cell_evaluation/evaluation_plugin.ts
+++ b/src/plugins/ui_core_views/cell_evaluation/evaluation_plugin.ts
@@ -189,10 +189,6 @@ export class EvaluationPlugin extends UIPlugin {
           this.evaluator.updateDependencies(cmd);
         }
         break;
-      case "DUPLICATE_SHEET":
-      case "CREATE_SHEET":
-        this.shouldRebuildDependenciesGraph = true;
-        break;
       case "EVALUATE_CELLS":
         this.evaluator.evaluateAllCells();
         break;
--- a/src/plugins/ui_core_views/cell_evaluation/evaluator.ts
+++ b/src/plugins/ui_core_views/cell_evaluation/evaluator.ts
@@ -327,6 +327,7 @@ export class Evaluator {
     const nbRows = formulaReturn[0].length;
 
     forEachSpreadPositionInMatrix(nbColumns, nbRows, this.updateSpreadRelation(formulaPosition));
+    this.assertNoMergedCellsInSpreadZone(formulaPosition, formulaReturn);
     forEachSpreadPositionInMatrix(nbColumns, nbRows, this.checkCollision(formulaPosition));
     forEachSpreadPositionInMatrix(
       nbColumns,
@@ -369,6 +370,26 @@ export class Evaluator {
 
     throw new SplillBlockedError(
       _t("Result couldn't be automatically expanded. Please insert more columns and rows.")
+    );
+  }
+
+  private assertNoMergedCellsInSpreadZone(
+    { sheetId, col, row }: CellPosition,
+    matrixResult: Matrix<FPayload>
+  ) {
+    const mergedCells = this.getters.getMergesInZone(sheetId, {
+      top: row,
+      bottom: row + matrixResult.length,
+      left: col,
+      right: col + matrixResult[0].length,
+    });
+
+    if (mergedCells.length === 0) {
+      return;
+    }
+
+    throw new SplillBlockedError(
+      _t("Merged cells found in the spill zone. Please unmerge cells before using array formulas.")
     );
   }
 
--- a/src/types/commands.ts
+++ b/src/types/commands.ts
@@ -118,6 +118,7 @@ export const invalidateEvaluationCommands = new Set<CommandTypes>([
   "UNDO",
   "REDO",
   "ADD_MERGE",
+  "REMOVE_MERGE",
   "UPDATE_LOCALE",
   "ADD_PIVOT",
   "UPDATE_PIVOT",
--- a/src/types/commands.ts
+++ b/src/types/commands.ts
@@ -113,6 +113,7 @@ export const invalidateEvaluationCommands = new Set<CommandTypes>([
   "RENAME_SHEET",
   "DELETE_SHEET",
   "CREATE_SHEET",
+  "DUPLICATE_SHEET",
   "ADD_COLUMNS_ROWS",
   "REMOVE_COLUMNS_ROWS",
   "UNDO",
--- a/tests/evaluation/evaluation_formula_array.test.ts
+++ b/tests/evaluation/evaluation_formula_array.test.ts
@@ -11,9 +11,11 @@ import {
   deleteColumns,
   deleteContent,
   deleteRows,
+  merge,
   paste,
   setCellContent,
   setFormat,
+  unMerge,
 } from "../test_helpers/commands_helpers";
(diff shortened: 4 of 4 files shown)


tests/evaluation/evaluation_formula_array.test.ts
```diff
@@ -884,4 +886,39 @@ describe("evaluate formulas that return an array", () => {
...
+    test("do not spread result when collide", () => {
+      setCellContent(model, "A1", "=MFILL(2, 2, 42)");
+      merge(model, "A2:A3");
+      expect(getEvaluatedCell(model, "A1").value).toBe("#SPILL!");
+      expect(getEvaluatedCell(model, "A2").value).toBe(null);
+      expect(getEvaluatedCell(model, "B1").value).toBe(null);
+      expect(getEvaluatedCell(model, "B2").value).toBe(null);
+    });
+
+    test("spread result when remove collision", () => {
```
rrahir, line 912:
```suggestion
    test("spread result upon collision removal", () => {
```

src/plugins/ui_core_views/cell_evaluation/evaluation_plugin.ts
```diff
@@ -196,6 +196,9 @@ export class EvaluationPlugin extends UIPlugin {
       case "EVALUATE_CELLS":
         this.evaluator.evaluateAllCells();
         break;
+      case "REMOVE_MERGE":
```
rrahir, line 199: I think the "right" approach would be to add "REMOVE_MERGE" to. the set `invalidateEvaluationCommands` which already contains ADD_MERGE, it follows the same invalidation logic IMHO

About this @anhe-odoo or @laa-odoo is there a reason why  force `shouldRebuildDependenciesGraph` to true when handling `CREATE_SHEET` even though it's already a part of `invalidateEvaluationCommands`? see 6 lines above this

src/plugins/ui_core_views/cell_evaluation/evaluator.ts
```diff
@@ -317,6 +317,7 @@ export class Evaluator {
     const formulaPosition = this.getters.getCellPosition(cellId);
 
     this.assertSheetHasEnoughSpaceToSpreadFormulaResult(formulaPosition, formulaReturn);
+    this.assertNoMergedCellsInSpreadZone(formulaPosition, formulaReturn);
```
rrahir, line 320: this check should be done after line 324 ( -> forEachSpreadPositionInMatrix(nbColumns, nbRows, this.updateSpreadRelation(formulaPosition)); )

Regardless of the presence of merges, we could want the formula spread positions to be updated.
@laa-odoo @anhe-odoo  Wouldn't it also be the case for formulas that spread out of the sheet?

tests/evaluation/evaluation_formula_array.test.ts
```diff
@@ -884,4 +886,39 @@ describe("evaluate formulas that return an array", () => {
...
+      expect(getEvaluatedCell(model, "A1").value).toBe("#SPILL!");
+    });
+
+    test("throw error message when encountering a horizontally merged cell", () => {
+      setCellContent(model, "A1", "=MFILL(2, 2, 42)");
+      merge(model, "B1:C1");
+      expect(getEvaluatedCell(model, "A1").value).toBe("#SPILL!");
+    });
+
+    test("do not spread result when collide", () => {
```
rrahir, line 903:
```suggestion
    test("do not spread result upon collision", () => {
```

tests/evaluation/evaluation_formula_array.test.ts
```diff
@@ -884,4 +886,39 @@ describe("evaluate formulas that return an array", () => {
       expect(model.getters.getSpreadZone({ sheetId, col: 0, row: 0 })).toEqual(toZone("A1:B2"));
     });
   });
+
+  describe("result array can collides with merged cells", () => {
+    test("throw error message when encountering a vertically merged cell", () => {
+      setCellContent(model, "A1", "=MFILL(2, 2, 42)");
```
rrahir, line 892: not sure how this test and the one below are relevant. just testing the rpesence of a merge within the spreading range should be enough

rrahir (changes requested):

src/plugins/ui_core_views/cell_evaluation/evaluation_plugin.ts
anhe-odoo: I agree with the idea that ADD_MERGE should be in the `invalidateEvaluationCommands` set. About CREATE_SHEET, it clearly make a "double emploi" here. But is there any reason for the `CREATE_SHEET` command to be in the `invalidateEvaluationCommands` set ? I tried to remove it, and the only test that failed was for the composer, so I'm not sure

rrahir: I think the mistake was to put the check on 'invalidateEvaluationCommands'  in allowDispatch, which is bypassed in some cases (namely, I think it's not considered on subcommands and on remote commands from the server). Moving the check to dispatch instead should do the trick?

rrahir: My mistake, `invalidateEvaluationCommands` are handled in beforeHandle, not allowDispatch so it's ok.

Considering the tests that only fail in the composer, it then could simply be a lack of test coverage, which we could take the opportunity to add with this pull request.

src/types/commands.ts
```diff
@@ -118,6 +118,7 @@ export const invalidateEvaluationCommands = new Set<CommandTypes>([
   "UNDO",
   "REDO",
   "ADD_MERGE",
+  "REMOVE_MERGE",
```
rrahir, line 121: we should take this opportunity to update `invalidateEvaluationCommands` with the command `DUPLICATE_SHEET` as well. (**in a different commit**) 

This would allow to remove the useless lines here https://github.com/odoo/o-spreadsheet/blob/master/src/plugins/ui_core_views/cell_evaluation/evaluation_plugin.ts#L192 

But should also come with some tests for DUPLICATE_SHEET.

rrahir (changes requested):

src/types/commands.ts
dhrp-odoo: I think adding a test is unnecessary in this case. If the `DUPLICATE_SHEET` command is removed from EvaluationPlugin without being added to `invalidateEvaluationCommands`, numerous existing test cases will fail, and the UI will break.

rrahir: lk

src/plugins/ui_core_views/cell_evaluation/evaluation_plugin.ts
dhrp-odoo: Why did the composer-related tests pass for me? 🤔

rrahir:
@robodoo rebase-ff

dhrp-odoo:
@rrahir, can you r- it, I'm working on it.

rrahir:
it is not validated

rrahir:
@robodoo r+

PR #4220, opened 2024-05-15
[FIX] find_and_replace: fix previous command

Rachico:
### [FIX] find_and_replace: fix previous command

Problem
-----
The problem occurs when we perform a search using "Find & Replace". In fact, when going backward to a previous occurrence that is  located in the previous sheet, we match the first occurrence of that sheet, when actually we need to match on the last occurrence of that sheet.

Solution
-----
This commit solves the issue by checking whether the sheet switch is automatic (coming from the Previous button).

Task: [3839575](https://www.odoo.com/web#id=3839575&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/find_and_replace/find_and_replace_store.ts
+++ b/src/components/side_panel/find_and_replace/find_and_replace_store.ts
@@ -23,6 +23,7 @@ export class FindAndReplaceStore extends SpreadsheetStore implements HighlightPr
   private currentSearchRegex: RegExp | null = null;
   private isSearchDirty = false;
   private initialShowFormulaState: boolean;
+  private isAutoSheetSwitch: boolean = false;
 
   // fixme: why do we make selectedMatchIndex on top of a selected
   // property in the matches?
@@ -103,15 +104,20 @@ export class FindAndReplaceStore extends SpreadsheetStore implements HighlightPr
       case "ADD_COLUMNS_ROWS":
       case "EVALUATE_CELLS":
       case "UPDATE_CELL":
+        this.isSearchDirty = true;
+        break;
       case "ACTIVATE_SHEET":
         this.isSearchDirty = true;
+        this.findMatches();
         break;
     }
   }
 
   finalize() {
-    if (this.isSearchDirty) {
-      this.refreshSearch(false);
+    if (this.isSearchDirty && !this.isAutoSheetSwitch) {
+      this.isAutoSheetSwitch = true;
+      this.refreshSearch();
+      this.isAutoSheetSwitch = false;
       this.isSearchDirty = false;
     }
   }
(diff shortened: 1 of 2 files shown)


src/components/side_panel/find_and_replace/find_and_replace_store.ts
```diff
@@ -110,10 +111,11 @@ export class FindAndReplaceStore extends SpreadsheetStore implements HighlightPr
   }
 
   finalize() {
-    if (this.isSearchDirty) {
+    if (this.isSearchDirty && !this.isAutoSheetSwitch) {
```
LucasLefevre, line 114: There is an issue: `this.activeSheetMatches` is no longer updated when switching to the next match on the next sheet :/

Rachico: @LucasLefevre done.

src/components/side_panel/find_and_replace/find_and_replace_store.ts
```diff
@@ -110,10 +111,12 @@ export class FindAndReplaceStore extends SpreadsheetStore implements HighlightPr
...
       this.refreshSearch(false);
       this.isSearchDirty = false;
     }
+    this.findMatches();
```
LucasLefevre, line 118: `findMatches` is not necessarily fast if the spreadsheet is really large. It's not a good idea to do it for *every* command finalize.

I think I'd try to split `case "ACTIVATE_SHEET":` from the other cases and handle that there.

src/components/side_panel/find_and_replace/find_and_replace_store.ts
```diff
@@ -260,6 +263,7 @@ export class FindAndReplaceStore extends SpreadsheetStore implements HighlightPr
 
     // Switch to the sheet where the match is located
     if (jumpToMatchSheet && this.getters.getActiveSheetId() !== selectedMatch.sheetId) {
```
LucasLefevre, line 265: is `jumpToMatchSheet` still needed ? Wouldn't it work with something like:
```suggestion
    if (!this.isAutoSheetSwitch && this.getters.getActiveSheetId() !== selectedMatch.sheetId) {
```

src/components/side_panel/find_and_replace/find_and_replace_store.ts
```diff
@@ -110,10 +111,12 @@ export class FindAndReplaceStore extends SpreadsheetStore implements HighlightPr
   }
 
   finalize() {
-    if (this.isSearchDirty) {
+    if (this.isSearchDirty && !this.isAutoSheetSwitch) {
       this.refreshSearch(false);
       this.isSearchDirty = false;
     }
+    this.findMatches();
+    this.isAutoSheetSwitch = false;
```
LucasLefevre, line 119: It'd reset this to `false` at the same place where it's set to `true`

LucasLefevre (commented):
this needs to be fixed in saas-17.2 :wink:

Rachico:
> this needs to be fixed in saas-17.2 😉

https://github.com/odoo/o-spreadsheet/pull/4250

PR #4230, opened 2024-05-18
[PERF] history: reduce history memory usage

LucasLefevre:
## Description:

This commit reduces the amount of memory used to keep track of history
changes.

Currently, for each change, we keep the full path of keys from the root plugin
to the deepest data structure which is actually updated.
`[BordersPlugin, 'borders', 'sheet1', 0, 1, 'horizontal']`

This is useless, it is enough to keep a reference to the updated object and
the updated key.
Note that `after` was never used.

On the large formula data set, with 20k rows, here are the amounts of memory
used (after GC has run)

adding a column left of A:
- before 613Mb
- after  487Mb

adding borders on all cells, all sides:
- before 451Mb
- after 384Mb

Undoing a revision is also faster since we no longer needs to iterate over
all keys of the path.

Adding a column before A, then undo and reload. Replaying both commands
goes from `~1700ms` to `~1200ms`

Task: : [3940604](https://www.odoo.com/web#id=3940604&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/types/history.ts
+++ b/src/types/history.ts
@@ -9,9 +9,9 @@ export interface CreateRevisionOptions {
 }
 
 export interface HistoryChange {
-  path: [any, ...(number | string)[]];
+  key: string;
+  target: any;
   before: any;
-  after: any;
 }
 
 export interface WorkbookHistory<Plugin> {
--- a/src/history/factory.ts
+++ b/src/history/factory.ts
@@ -1,7 +1,6 @@
 import { transformAll } from "../collaborative/ot/ot";
 import { Revision } from "../collaborative/revisions";
 import { inverseCommand } from "../helpers/inverse_commands";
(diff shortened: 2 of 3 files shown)


src/types/history.ts
```diff
@@ -9,9 +9,9 @@ export interface CreateRevisionOptions {
 }
 
 export interface HistoryChange {
-  path: [any, ...(number | string)[]];
+  key: string;
+  obj: any;
```
LucasLefevre, line 13: `obj` :thinking:

pro-odoo: `target` ?

pro-odoo (approved):
robodoo r+
I wonder if it would be worth to backport it, what do you think ?

PR #4250, opened 2024-05-23
[FIX] find_and_replace: fix previous command

Rachico:
### [FIX] find_and_replace: fix previous command

Problem
-----
The problem occurs when we perform a search using "Find & Replace". In fact, when going backward to a previous occurrence that is  located in the previous sheet, we match the first occurrence of that sheet, when actually we need to match on the last occurrence of that sheet.

Solution
-----
This commit solves the issue by checking whether the sheet switch is automatic (coming from the Previous button).

Task: [3839575](https://www.odoo.com/web#id=3839575&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/find_and_replace/find_and_replace_store.ts
+++ b/src/components/side_panel/find_and_replace/find_and_replace_store.ts
@@ -30,6 +30,7 @@ export class FindAndReplaceStore extends SpreadsheetStore implements HighlightPr
   private currentSearchRegex: RegExp | null = null;
   private isSearchDirty = false;
   private initialShowFormulaState: boolean;
+  private preserveSelectedMatchIndex: boolean = false;
 
   // fixme: why do we make selectedMatchIndex on top of a selected
   // property in the matches?
@@ -157,7 +158,9 @@ export class FindAndReplaceStore extends SpreadsheetStore implements HighlightPr
    * refresh the matches according to the current search options
    */
   private refreshSearch(jumpToMatchSheet = true) {
-    this.selectedMatchIndex = null;
+    if (!this.preserveSelectedMatchIndex) {
+      this.selectedMatchIndex = null;
+    }
     this.findMatches();
     this.selectNextCell(Direction.current, jumpToMatchSheet);
   }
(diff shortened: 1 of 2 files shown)


src/types/commands.ts
```diff
@@ -684,6 +684,7 @@ export interface ActivateSheetCommand {
   type: "ACTIVATE_SHEET";
   sheetIdFrom: UID;
   sheetIdTo: UID;
+  autoSwitch?: boolean;
```
pro-odoo, line 687: It's not a good thing to add a flag in a widely used command only to handle a very specific use case in a very specific store. Also, the real issue is that `selectedMatchIndex` is reset in the method `refreshSearch`, and shouldn't in this use case. So, I suggest an alternative which is adding a flag **inside** the store:
```js
export class FindAndReplaceStore ...
...
private preserveSelectedMatchIndex: boolean = false;
...
finalize() {
// revert changes
}
...
private refreshSearch(jumpToMatchSheet = true) {
    if (!this.preserveSelectedMatchIndex) {
      this.selectedMatchIndex = null;
    }
    this.findMatches();
    this.selectNextCell(Direction.current, jumpToMatchSheet);
  }
...
if (jumpToMatchSheet && this.getters.getActiveSheetId() !== selectedMatch.sheetId) {
      this.preserveSelectedMatchIndex = true;
      this.model.dispatch("ACTIVATE_SHEET", {
        sheetIdFrom: this.getters.getActiveSheetId(),
        sheetIdTo: selectedMatch.sheetId,
        autoSwitch: true,
      });
      this.preserveSelectedMatchIndex = false;
      // We do not want to reset the selection at finalize in this case
      this.isSearchDirty = false;
    }
...
```

All tests are green with theses changes, the diff is smaller and the logic is kept inside the store and do not leak elsewhere.

Side note, `autoSwitch` and `isAutoSheetSwitch` are not really meaningful, it's important (and really difficult :D ) to have good names, it's easier to read the code (not only now, but in 2 or 3 years :) )

pro-odoo (changes requested):

src/components/side_panel/find_and_replace/find_and_replace_store.ts
```diff
@@ -267,10 +270,12 @@ export class FindAndReplaceStore extends SpreadsheetStore implements HighlightPr
 
     // Switch to the sheet where the match is located
     if (jumpToMatchSheet && this.getters.getActiveSheetId() !== selectedMatch.sheetId) {
+      this.preserveSelectedMatchIndex = true;
```
pro-odoo, line 273: You can add a little comment to explain why we need to do this.

pro-odoo (approved):
robodoo r+

PR #4272, opened 2024-05-27
[FIX] clipboard: do not change clipboard state on INSERT_CELLS

hokolomopo:
## Description

Commands `INSERT_CELLS`, `DELETE_CELLS`, `COPY_PASTE_CELLS_ABOVE`, and `COPY_PASTE_CELLS_ON_LEFT` and handled by the clipboard plugin. The problem is that since the last refactor of the clipboard plugin, these commands also invalidated the state of the copy/paste.

This commit fixes the issue by not changing the clipboard state when handling those commands.

Task: : [3874117](https://www.odoo.com/web#id=3874117&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/types/clipboard.ts
+++ b/src/types/clipboard.ts
@@ -8,9 +8,9 @@ export enum ClipboardMIMEType {
 export type ClipboardContent = { [type in ClipboardMIMEType]?: string };
 
 export interface ClipboardOptions {
+  isCutOperation: boolean;
   pasteOption?: ClipboardPasteOptions;
   selectTarget?: boolean;
-  isCutOperation?: boolean;
 }
 export type ClipboardPasteOptions = "onlyFormat" | "asValue";
 export type ClipboardOperation = "CUT" | "COPY";
--- a/src/clipboard_handlers/sheet_clipboard.ts
+++ b/src/clipboard_handlers/sheet_clipboard.ts
@@ -6,7 +6,6 @@ type ClipboardContent = {
   cells: any[][];
   zones: Zone[];
   sheetId: UID;
-  isCutOperation: boolean;
 };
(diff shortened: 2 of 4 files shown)


src/types/clipboard.ts
```diff
@@ -10,7 +10,7 @@ export type ClipboardContent = { [type in ClipboardMIMEType]?: string };
 export interface ClipboardOptions {
   pasteOption?: ClipboardPasteOptions;
   selectTarget?: boolean;
-  isCutOperation?: boolean;
+  isCutOperation: boolean;
```
rrahir, line 13: nitpick, we could move this at the top of the interface declaration since it's mandatory

rrahir (commented):
👍  I juste have a question. The task also refers to the command INSERT_CELL being an issue and I don't see a test regarding this. I have tested a bit on saas-17.2 and it doesn't seem buggy. Do you confirm that it was working properly before the fix?

rrahir:
Don't know how I f*'d my test but yes pretty broken. I also missed the insertcell test but to be fair, you test 2 different commands in the same test, I was expecting 2 different tests.

hokolomopo:
> Don't know how I f*'d my test but yes pretty broken. I also missed the insertcell test but to be fair, you test 2 different commands in the same test, I was expecting 2 different tests.

Delete_cells and insert_cells are kinda related so I didn't bother doing two separate tests, but I can split the tests if you want

rrahir:
@robodoo r+

PR #4280, opened 2024-05-27
[FIX] xlsx: Skip non-exported formula with no result

rrahir:
Our library, as long as Google Sheet, support the ability to manipulate apples and oranges (e.g. A1: "" , A2: =A1+2 - will work and yield 2 in our library) while Excel will reject it.

This revision adds some compatibility with the Excel export. In the case of a non-exportable formula which returns nothing as in : a falsy value which retains no information.

There are three types of falsy values:
- FALSE
- 0
- an empty string `""`

The two first cases contain information, as they can be the result nad boolean data. The third case could only be an empty formula result or the result of a string manipulation but in both cases, the information is the same as in an empty cell.
As such, this revision targets empty strings and skips them during the export (again only when they result from a formula non supported by Excel).

Task: 3491974

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


LucasLefevre (approved):
robodoo r+

PR #4296, opened 2024-05-28
[FIX] border: bottom sheet borders removed on DELETE_ROWS

Rachico:
### [FIX] border: bottom sheet borders removed on DELETE_ROWS

Problem
-----
Before this commit, when we add border at the bottom of the sheet (at very the last rows), if we delete some rows in a position before the border, some/all border disappear. The reason being that the sheet plugin updates the total rows before we loop through them to shift the border.

Solution
------
This commit fixes this behaviour by adding the total deleted rows to the loop that takes care of shifting the borders.

Task: [3911695](https://www.odoo.com/web#id=3911695&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/core/borders.ts
+++ b/src/plugins/core/borders.ts
@@ -105,7 +105,7 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
           if (cmd.dimension === "COL") {
             this.shiftBordersHorizontally(cmd.sheetId, el + 1, -1);
           } else {
-            this.shiftBordersVertically(cmd.sheetId, el + 1, -1);
+            this.shiftBordersVertically(cmd.sheetId, el + 1, -1, cmd.elements.length);
           }
         }
         break;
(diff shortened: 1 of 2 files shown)


src/plugins/core/borders.ts
```diff
@@ -329,7 +330,7 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
         destructive: false,
       });
     }
-    this.getRowsRange(sheetId)
+    range(0, this.getters.getNumberRows(sheetId) + totalDeletedRows + 1)
```
anhe-odoo, line 333: It seems a bit strange that we need something different here than for columns removing, while it should probably be done in the same way :/
We can maybe use something like the `this.getColumnsWithBorders(sheetId)` method used in the `shiftBordersHorizontally`, to avoid adding a parameter in this method, with something like
```js
  private getRowsWithBorders(sheetId: UID): HeaderIndex[] {
    const sheetBorders = this.borders[sheetId]?.filter(isDefined);
    if (!sheetBorders) return [];
    return Object.keys(sheetBorders[0]).map((index) => parseInt(index, 10));
  }
```
(juste make sure taking the first item of `sheetBorders` is enough)
Any idea about this suggestion @rrahir ?

anhe-odoo (changes requested):

src/plugins/core/borders.ts
rrahir: @anhe-odoo Agreed, there doesn't seem to be a good reason not to use the same approach for both the columns and rows. The suggested `getRowsWithBorders` seems good :)

tests/borders/border_plugin.test.ts
```diff
@@ -595,6 +596,17 @@ describe("Grid manipulation", () => {
     expect(getBorder(model, "A2")).toEqual({ top: b });
   });
 
+  test("Remove multiple rows before borders at the bottom of the sheet", () => {
+    createSheet(model, { activate: true, sheetId: "sh1", rows: 5, cols: 2 });
```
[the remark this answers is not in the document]
rrahir: Thought I had left a comment here: Pretty sure we can do without the createSheet. We should simply add the borders on the last rows (i.e. index 98 and 99) while deleting the first 2 rows.

Rachico:
New PR targeting 15.0: https://github.com/odoo/o-spreadsheet/pull/4325

PR #4297, opened 2024-05-29
[PERF] evaluation: faster spreading

LucasLefevre:
## Description:

This commit improves the speed of spreading a large array formula result.

Reference a large zone `=transpose(transpose(A1:E3000))`

before: 1.22s
after:  387ms

Instanciating a PositionSet is not free (it allocates memory)

Here, PositionSets were used but the properties of sets are actually not used
(uniquness of values in the set, O(1) lookup), and we add each position
only once and we never have to check if a position is in the set.
Using a simple arry is perfectly fine.

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


rrahir:
@robodoo r+

PR #4298, opened 2024-05-29
[FIX] composer: Prevent default paste in composer

rrahir:
Steps to reproduce:
- click on the column A
- copy/paste the column -> the layout is broken, some divs were shifted up and are hidden by the topbar. Namely, The column header is not visible anymore.

This issue is related to the one addressed in this pr[^1]. While we would prevent the handling of `insertFromPaste` inputs, the default browser behaviour would still occur and modify the `contentEditableHelper`. This would trigger a replacement of some components DOM elements, notably the header overlay. Since, from a position POV, the canvas is positioned just below the said overlay, it was also shifted.

This revision is taking a more aggressive approach than the one in [^1], we stop the paste event totally (not just the following input) in the composer when we are not editing.

[^1]: https://github.com/odoo/o-spreadsheet/pull/2254

Task: 3864000

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/tests/components/composer.test.ts
+++ b/tests/components/composer.test.ts
@@ -604,16 +604,12 @@ describe("composer", () => {
     await nextTick();
     expect(model.getters.getEditionMode()).toBe("inactive");
   });
-  test("input event triggered from a paste should not open composer", async () => {
-    gridInputEl.dispatchEvent(
-      new InputEvent("input", {
-        data: "d",
-        bubbles: true,
-        isComposing: false,
-        inputType: "insertFromPaste",
-      })
-    );
+
+  test("Default paste is prevented in a closed composer", async () => {
+    const pasteEvent = new Event("paste", { cancelable: true });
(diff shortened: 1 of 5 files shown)


tests/components/composer.test.ts
```diff
@@ -604,16 +604,12 @@ describe("composer", () => {
...
-    gridInputEl.dispatchEvent(
-      new InputEvent("input", {
-        data: "d",
-        bubbles: true,
-        isComposing: false,
-        inputType: "insertFromPaste",
-      })
-    );
+
+  test("Default paste is prevented in a closed popover", async () => {
```
hokolomopo, line 608:
```suggestion
  test("Default paste is prevented in a closed composer", async () => {
```

rrahir: :+1:

pro-odoo (approved):
robodoo r+

PR #4319, opened 2024-05-31
[ADD] formulas: introduce OFFSET function

Rachico:
### [ADD] formulas: introduce OFFSET function

This commit introduces the OFFSET function which returns a range reference shifted a specified number of rows and columns from a starting cell reference.

Task: [2837743](https://www.odoo.com/web#id=2837743&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/functions/module_lookup.ts
+++ b/src/functions/module_lookup.ts
@@ -252,7 +252,7 @@ export const INDEX: AddFunctionDescription = {
 };
 
 // -----------------------------------------------------------------------------
-// INDEX
+// INDIRECT
 // -----------------------------------------------------------------------------
 export const INDIRECT: AddFunctionDescription = {
   description: _t("Returns the content of a cell, specified by a string."),
--- a/src/functions/module_lookup.ts
+++ b/src/functions/module_lookup.ts
@@ -1,8 +1,9 @@
-import { getFullReference, range, toXC, toZone } from "../helpers/index";
+import { getFullReference, range, splitReference, toXC, toZone } from "../helpers/index";
 import { _t } from "../translation";
-import { AddFunctionDescription, CellPosition, FPayload, Matrix, Maybe } from "../types";
+import { AddFunctionDescription, CellPosition, FPayload, Matrix, Maybe, Zone } from "../types";
 import { CellErrorType, EvaluationError, InvalidReferenceError } from "../types/errors";
 import { arg } from "./arguments";
+import { assertPositive } from "./helper_assert";
 import {
   addPivotDependencies,
   assertDomainLength,
@@ -15,6 +16,7 @@ import {
   assertNumberGreaterThanOrEqualToOne,
   dichotomicSearch,
   expectNumberRangeError,
+  generateMatrix,
   isEvaluationError,
   linearSearch,
   strictToInteger,
@@ -840,5 +842,124 @@ export const PIVOT = {
       result[0][0] = { value: pivotTitle };
     }
     return result;
+  },
+} satisfies AddFunctionDescription;
+
+//--------------------------------------------------------------------------
+// OFFSET
+//--------------------------------------------------------------------------
+
+export const OFFSET = {
+  description: _t(
+    "Returns a range reference shifted by a specified number of rows and columns from a starting cell reference."
+  ),
+  args: [
+    arg(
+      "cell_reference (meta)",
+      _t("The starting point from which to count the offset rows and columns.")
+    ),
+    arg("offset_rows (number)", _t("The number of rows to offset by.")),
+    arg("offset_columns (number)", _t("The number of columns to offset by.")),
+    arg(
+      "height (number, default='height of cell_reference')",
+      _t("The number of rows of the range to return starting at the offset target.")
+    ),
+    arg(
+      "width (number, default='width of cell_reference')",
+      _t("The number of columns of the range to return starting at the offset target.")
+    ),
+  ],
+  compute: function (
+    cellReference: Maybe<{ value: string }>,
+    offsetRows: Maybe<FPayload>,
+    offsetColumns: Maybe<FPayload>,
+    height: Maybe<FPayload>,
+    width: Maybe<FPayload>
+  ) {
+    if (isEvaluationError(cellReference?.value)) {
+      return cellReference;
+    }
+
+    const _cellReference = cellReference?.value;
+    if (!_cellReference) {
+      throw new Error(
+        "In this context, the function OFFSET needs to have a cell or range in parameter."
+      );
+    }
+    const zone = toZone(_cellReference);
+
+    let offsetHeight = zone.bottom - zone.top + 1;
+    let offsetWidth = zone.right - zone.left + 1;
+
+    if (height) {
+      const _height = toNumber(height, this.locale);
+      assertPositive(
+        _t("Height value is %(_height)s. It should be greater than or equal to 1.", { _height }),
+        _height
+      );
+      offsetHeight = _height;
+    }
+
+    if (width) {
+      const _width = toNumber(width, this.locale);
+      assertPositive(
+        _t("Width value is %(_width)s. It should be greater than or equal to 1.", { _width }),
+        _width
+      );
+      offsetWidth = _width;
+    }
+
+    const { sheetName } = splitReference(_cellReference);
+
+    const sheetId =
+      (sheetName && this.getters.getSheetIdByName(sheetName)) || this.getters.getActiveSheetId();
+
+    const _offsetRows = toNumber(offsetRows, this.locale);
+    const _offsetColumns = toNumber(offsetColumns, this.locale);
+
+    let originPosition: CellPosition | undefined;
+    const __originCellXC = this.__originCellXC?.();
+    if (__originCellXC) {
+      const cellZone = toZone(__originCellXC);
+      originPosition = {
+        sheetId: this.__originSheetId,
+        col: cellZone.left,
+        row: cellZone.top,
+      };
+      this.updateDependencies?.(originPosition);
+    }
+
+    const startingCol = zone.left + _offsetColumns;
+    const startingRow = zone.top + _offsetRows;
+
+    if (startingCol < 0 || startingRow < 0) {
+      return new InvalidReferenceError(_t("OFFSET evaluates to an out of bounds range."));
+    }
+
+    const dependencyZone: Zone = {
+      left: startingCol,
+      top: startingRow,
+      right: startingCol + offsetWidth - 1,
+      bottom: startingRow + offsetHeight - 1,
+    };
+
+    const range = this.getters.getRangeFromZone(this.__originSheetId, dependencyZone);
+    if (range.invalidXc || range.invalidSheetName) {
+      return new InvalidReferenceError();
+    }
+    if (originPosition) {
+      this.addDependencies?.(originPosition, [range]);
+    }
+
+    return generateMatrix(
+      offsetWidth,
+      offsetHeight,
+      (col: number, row: number): FPayload =>
+        this.getters.getEvaluatedCell({
+          sheetId,
+          col: startingCol + col,
+          row: startingRow + row,
+        })
+    );
   },
 } satisfies AddFunctionDescription;
--- a/tests/functions/module_lookup.test.ts
+++ b/tests/functions/module_lookup.test.ts
@@ -1,6 +1,10 @@
 import { Model } from "../../src/model";
 import { activateSheet, createSheet, setCellContent } from "../test_helpers/commands_helpers";
-import { getCellContent, getEvaluatedCell } from "../test_helpers/getters_helpers";
+import {
+  getCellContent,
+  getEvaluatedCell,
+  getEvaluatedGrid,
+} from "../test_helpers/getters_helpers";
 import {
   createModelFromGrid,
   evaluateCell,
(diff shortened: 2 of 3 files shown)


laa-odoo (changes requested):
good approach ;)

a few small comments to take into account
can also add tests for each comment :)

src/functions/module_lookup.ts
```diff
@@ -858,3 +866,101 @@ export const PIVOT = {
...
+    offsetRows: Maybe<FPayloadNumber>,
+    offsetColumns: Maybe<FPayloadNumber>,
+    height: Maybe<FPayloadNumber>,
+    width: Maybe<FPayloadNumber>
+  ) {
+    const _cellReference = cellReference?.value;
+    const zone = toZone(_cellReference!);
+    const _offsetRows: number = offsetRows?.value || 0;
+    const _offsetColumns: number = offsetColumns?.value || 0;
+    const _height: number = height ? height.value - 1 : zone.bottom - zone.top;
```
laa-odoo, line 905: you can use an assert to check if height and width are positive or strictly positive ;)

src/functions/module_lookup.ts
```diff
@@ -858,3 +861,121 @@ export const PIVOT = {
...
+    if (width) {
+      const _width = toNumber(width, this.locale);
+      assertPositive(`width value is ${_width}. It should be greater than or equal to 1.`, _width);
+      offsetWidth = _width - 1;
+    }
+
+    let sheetId = this.getters.getActiveSheetId();
+
+    if (_cellReference?.includes("!")) {
+      const rangeSheetName = _cellReference.split("!").at(0);
```
rrahir, line 924: there already is a helper for this ;-) `splitReference`

src/functions/module_lookup.ts
```diff
@@ -858,3 +861,121 @@ export const PIVOT = {
...
+      assertPositive(
+        `Height value is ${_height}. It should be greater than or equal to 1.`,
+        _height
+      );
+      offsetHeight = _height - 1;
+    }
+
+    if (width) {
+      const _width = toNumber(width, this.locale);
+      assertPositive(`width value is ${_width}. It should be greater than or equal to 1.`, _width);
```
laa-odoo, line 917: _t(

src/functions/module_lookup.ts
```diff
@@ -858,3 +861,121 @@ export const PIVOT = {
...
+      throw cellReference;
+    }
+    const zone = toZone(_cellReference!);
+
+    let offsetHeight = zone.bottom - zone.top;
+    let offsetWidth = zone.right - zone.left;
+
+    if (height) {
+      const _height = toNumber(height, this.locale);
+      assertPositive(
```
laa-odoo, line 908: _t(

src/functions/module_lookup.ts
```diff
@@ -258,7 +261,7 @@ export const INDEX: AddFunctionDescription = {
 };
 
 // -----------------------------------------------------------------------------
-// INDEX
+// INDIRECT
```
anhe-odoo, line 264: :+1:

src/functions/module_lookup.ts
```diff
@@ -858,3 +861,121 @@ export const PIVOT = {
...
+      const cellZone = toZone(__originCellXC);
+      originPosition = {
+        sheetId: originSheetId,
+        col: cellZone.left,
+        row: cellZone.top,
+      };
+      this.updateDependencies?.(originPosition);
+    }
+
+    const startPosition = {
```
anhe-odoo, line 944: Maybe keep it as separated variables as creating an object is not useful here :)
```js
const startingCol = zone.left + _offsetColumns;
const startingRow = zone.top + _offsetRows;
```

src/functions/module_lookup.ts
```diff
@@ -858,3 +861,121 @@ export const PIVOT = {
...
+    ),
+  ],
+  compute: function (
+    cellReference: Maybe<{ value: string }>,
+    offsetRows: Maybe<FPayload>,
+    offsetColumns: Maybe<FPayload>,
+    height: Maybe<FPayload>,
+    width: Maybe<FPayload>
+  ): Matrix<FPayload> {
+    const _cellReference = cellReference?.value;
```
anhe-odoo, line 896: You can move this one after the `if (isEvaluationError...`

anhe-odoo (commented):
Nice formula :) And nice test coverage

laa-odoo (approved):

src/functions/module_lookup.ts
```diff
@@ -858,3 +860,122 @@ export const PIVOT = {
...
+      );
+      offsetWidth = _width - 1;
+    }
+
+    let sheetId = this.getters.getActiveSheetId();
+
+    const { sheetName } = splitReference(_cellReference!);
+
+    if (sheetName) {
+      sheetId = this.getters.getSheetIdByName(sheetName)!; // Bad sheet name is handled by isEvaluationError() function above
```
rrahir, line 928:
```suggestion
      sheetId = this.getters.getSheetIdByName(sheetName)!; // invalid references are caught upstream, we are guaranteed that the sheetId is found at this point.
```
Better to give the functional reason why we can make this assumption

src/functions/module_lookup.ts
```diff
@@ -858,3 +860,122 @@ export const PIVOT = {
...
+    width: Maybe<FPayload>
+  ): Matrix<FPayload> {
+    if (isEvaluationError(cellReference?.value)) {
+      throw cellReference;
+    }
+
+    const _cellReference = cellReference?.value;
+    const zone = toZone(_cellReference!);
+
+    let offsetHeight = zone.bottom - zone.top;
```
rrahir, line 902: I think the variable is not properly named. The height/width of the value returned will never be 0, so instead, I would set `let offsetHeight = zone.bottom - zone.top +1` such that it describes exactly what it does and remove the +1 in specific places, like when we compute `dependencyZone`

rrahir (changes requested):

src/functions/module_lookup.ts
```diff
@@ -858,3 +860,122 @@ export const PIVOT = {
...
+    const { sheetName } = splitReference(_cellReference!);
+
+    if (sheetName) {
+      sheetId = this.getters.getSheetIdByName(sheetName)!; // Bad sheet name is handled by isEvaluationError() function above
+    }
+
+    const _offsetRows = toNumber(offsetRows, this.locale);
+    const _offsetColumns = toNumber(offsetColumns, this.locale);
+
+    const originSheetId = this.__originSheetId;
```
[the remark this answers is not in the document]
Rachico: what exactly is not useful ?

tests/functions/module_lookup.test.ts
```diff
@@ -1622,3 +1622,196 @@ describe("INDIRECT formula", () => {
...
+    expect(evaluateCell("A1", { A1: "=OFFSET(A1:C5, 'hola')" })).toBe("#BAD_EXPR");
+    expect(evaluateCell("A1", { A1: "=OFFSET(A1:C5)" })).toBe("#BAD_EXPR");
+    expect(evaluateCell("A1", { A1: "=OFFSET(A1:C5, 0)" })).toBe("#BAD_EXPR");
+    expect(evaluateCell("A1", { A1: "=OFFSET(A1:C5, -1, -1)" })).toBe("#REF");
+    expect(evaluateCell("A1", { A1: "=OFFSET(A1:C5, 0, 0, -1, 0)" })).toBe("#ERROR");
+    expect(evaluateCell("A1", { A1: "=OFFSET(A1:C5, 0, 0, 0, -1)" })).toBe("#ERROR");
+    expect(evaluateCell("A1", { A1: "=OFFSET(Sheet100!A1:C5, , 0, -1)" })).toBe("#ERROR");
+  });
+
+  test("should select single cell", () => {
```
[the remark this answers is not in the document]
Rachico: it means that the OFFSET function should be able to select a zone containing one & only one cell.

src/functions/module_lookup.ts
Rachico: why can't the height/width be 0 ?

src/functions/module_lookup.ts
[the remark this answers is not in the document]
rrahir: creating a new variable for the 2 places where you use it. it's a direct fecerence, not something that needs to be computed so you could directly use `this.__originSheetId` and don't have to assign a new variable ^^

tests/functions/module_lookup.test.ts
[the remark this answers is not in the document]
rrahir: I feel like the wording could be la little more explicit. "should be able" is more like "can select a single cell" than "should".

src/functions/module_lookup.ts
rrahir: Based on the variable name, I would expect this value to be one of the dimension of the returned matrix of results. for instance in a matrix with `m` rows, I would expect offsetHeight to be `m`. I think we can agree that having a matrix with one of its dimension being 0 doesn't make sense^^.
But in the code you wrote, offsetHeight is set to `m-1` and then you call `generateMatrix` by incrementing the value of offsetHeight.

My comment was more about the code readability, to me the variable name is misleading, hence my suggestion.

src/functions/module_lookup.ts
```diff
@@ -848,3 +850,120 @@ export const PIVOT = {
...
+
+    const dependencyZone: Zone = {
+      left: startingCol,
+      top: startingRow,
+      right: startingCol + offsetWidth - 1,
+      bottom: startingRow + offsetHeight - 1,
+    };
+
+    const range = this.getters.getRangeFromZone(this.__originSheetId, dependencyZone);
+    if (range === undefined || range.invalidXc || range.invalidSheetName) {
```
rrahir, line 951: range cannot be undefined ;-)

src/functions/module_lookup.ts
```diff
@@ -848,3 +850,120 @@ export const PIVOT = {
...
+      throw cellReference;
+    }
+
+    const _cellReference = cellReference?.value;
+    const zone = toZone(_cellReference!);
+
+    let offsetHeight = zone.bottom - zone.top + 1;
+    let offsetWidth = zone.right - zone.left + 1;
+
+    if (height) {
```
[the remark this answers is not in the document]
Rachico: I need to check wether the user inputed a value for width and height or not

tests/functions/module_lookup.test.ts
```diff
@@ -1622,3 +1622,198 @@ describe("INDIRECT formula", () => {
...
+    const model = createModelFromGrid(grid);
+    expect(getEvaluatedCell(model, "D5").value).toBe("A3");
+    expect(getEvaluatedCell(model, "E5").value).toBe("B3");
+    expect(getEvaluatedCell(model, "F5").value).toBe("C3");
+    expect(getEvaluatedCell(model, "D6").value).toBe(null);
+    expect(getEvaluatedCell(model, "E6").value).toBe(null);
+    expect(getEvaluatedCell(model, "F6").value).toBe(null);
+    expect(getEvaluatedCell(model, "D7").value).toBe(null);
+    expect(getEvaluatedCell(model, "E7").value).toBe(null);
+    expect(getEvaluatedCell(model, "F7").value).toBe(null);
```
pro-odoo, line 1753:
```ts
expect(getEvaluatedGrid(model, "D5:F7")).toEqual([
        ["A3","B3","C3"],
        ["","",""],
        ["","",""],
    ]);
``` 

Easier to read. Use the same everywhere

src/functions/module_lookup.ts
```diff
@@ -845,3 +847,120 @@ export const PIVOT = {
...
+        _t("Height value is %(_height)s. It should be greater than or equal to 1.", { _height }),
+        _height
+      );
+      offsetHeight = _height;
+    }
+
+    if (width) {
+      const _width = toNumber(width, this.locale);
+      assertPositive(
+        _t("width value is %(_width)s. It should be greater than or equal to 1.", { _width }),
```
pro-odoo, line 904:
```suggestion
        _t("Width value is %(_width)s. It should be greater than or equal to 1.", { _width }),
```
Be consistent with the error message of `height`

pro-odoo (changes requested):
`=OFFSET(INDIRECT("A1"),1,1)` doest not work, but works on GSpreadsheet

laa-odoo:
> `=OFFSET(INDIRECT("A1"),1,1)` doest not work, but works on GSpreadsheet

This is a technical limitation that we have at the moment, 
we will not be able to solve this problem until it is possible to have references to the formula output,

Here what happens concretely is that `indirect` evaluates what is in `A1` and directly returns the result of A1 in the form of an object `{value, format}`.
As Offset is defined as accepting meta arguments, it does not understand when it receives a value/format object and not a reference.


Tell us if this functional case is important, 
IMO yes

pro-odoo:
> > `=OFFSET(INDIRECT("A1"),1,1)` doest not work, but works on GSpreadsheet
> 
> This is a technical limitation that we have at the moment, we will not be able to solve this problem until it is possible to have references to the formula output,
> 
> Here what happens concretely is that `indirect` evaluates what is in `A1` and directly returns the result of A1 in the form of an object `{value, format}` (it shouldn't) As Offset is defined as accepting meta arguments, it does not understand when it receives a value/format object and not a reference.
> 
> Tell us if this functional case is important, IMO yes

Ok, don't know if it's important or not but it's not blocking for merging this task

src/functions/module_lookup.ts
```diff
@@ -842,3 +844,120 @@ export const PIVOT = {
...
+      );
+      offsetWidth = _width;
+    }
+
+    let sheetId = this.getters.getActiveSheetId();
+
+    const { sheetName } = splitReference(_cellReference!);
+
+    if (sheetName) {
+      sheetId = this.getters.getSheetIdByName(sheetName)!; // invalid references are caught upstream, we are guaranteed that the sheetId is found at this point.
```
pro-odoo, line 912: It can works now, but the assumption you made can be wrong after another commit and detecting the error could be painful. You guarantee here that sheetId is defined, but if it's not the case anymore, there will be an obscur error later in the code (which assume that sheetId is defined). In my experience, each `!` we introduce (outside tests of course) will require a bugfix later. In addition, it breaks the power of typescript as it bypasses the type-checking.
I prefer to have a check with a proper error at the beginning of the function.

pro-odoo (approved):
robodoo rebase-ff r+

PR #4321, opened 2024-05-31
Master imp number computation time

laa-odoo:
Reduce computation time spent in IsNumber by 'tres beaucoup'

Task: : [3965387](https://www.odoo.com/web#id=3965387&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/format.ts
+++ b/src/helpers/format.ts
@@ -12,10 +12,8 @@ import {
   PLAIN_TEXT_FORMAT,
 } from "../types";
 import { EvaluationError } from "../types/errors";
-import { DEFAULT_LOCALE } from "./../types/locale";
 import { DateTime, INITIAL_1900_DAY, isDateTime, numberToJsDate, parseDateTime } from "./dates";
 import { escapeRegExp, memoize } from "./misc";
-import { isNumber } from "./numbers";
 
 /**
  *  Constant used to indicate the maximum of digits that is possible to display
@@ -532,10 +530,8 @@ export function detectDateFormat(content: string, locale: Locale): Format | unde
   return internalDate.format;
 }
 
+/** use this function only if the content corresponds to a number (means that isNumber(content) return true */
 export function detectNumberFormat(content: string): Format | undefined {
-  if (!isNumber(content, DEFAULT_LOCALE)) {
-    return undefined;
-  }
   const digitBase = content.includes(".") ? "0.00" : "0";
   const matchedCurrencies = content.match(/[\$€]/);
   if (matchedCurrencies) {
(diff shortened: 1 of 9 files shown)


src/plugins/core/cell.ts
```diff
@@ -571,16 +571,19 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
...
-      content = toString(parseLiteral(content, locale));
+      content = toString(parsedValue);
     }
     return {
       id,
       content,
       style,
       format,
       isFormula: false,
+      parsedValue,
```
LucasLefevre, line 590: we could maybe even store the full `EvaluatedCell` object. It would avoid recreating it every time (and garbage collecting the previous one).
WDYT?

rrahir: Not a big fan of putting an evaluation related object back to the core plugins ...

laa-odoo: Two steps further would be to do the same thing for formulas without references ? 🤔

LucasLefevre: > ld be to do the same thing for formulas without r

doesn't work because a formula without references can still perform RPCs and depends on non-core stuff

src/helpers/format.ts
```diff
@@ -532,10 +530,8 @@ export function detectDateFormat(content: string, locale: Locale): Format | unde
   return internalDate.format;
 }
 
+// use this function only if the content corresponds to a number (means that isNumber(content) return true
```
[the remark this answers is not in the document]
LucasLefevre: > I'm not a fan of the idea to have helpers that need to be conditionally called without any typing safeguard but I get that it's something we'd have to do if we want to optimize the library

We're on the same page

src/plugins/core/cell.ts
rrahir: Doesn't cost too much to see the impact on a real-world use case ;) I also am curious on the imact it'd have on rng's spreadsheet

laa-odoo: ### Test with `makeLargeDataset(26, 10_000, ["numbers"])`

|                            | Master | Changes on is Number | Changes on is number + store evaluated cell when literal cell |
|----------------------------|--------|----------------------|--------------------------------------------------------------|
| **Evaluate all cells**     | 254    | 148                  | 107                                                          |
| **Model created in**       | 1263   | 1045                 | 1114                                                         |

### Test on real usecase (the monstrous RNG spreadsheet)

|                            | Master | Changes on is Number | Changes on is number + store evaluated cell when literal cell |
|----------------------------|--------|----------------------|--------------------------------------------------------------|
| **Evaluate all cells**     | 743    | 719                  | 703                                                          |
| **Model created in**       | 4071   | 3792                 | 3664                                                         |


*All measurements are averages of 5 values.*

LucasLefevre (approved):
robodoo r+ rebase-ff

PR #4325, opened 2024-05-31
[FIX] border: bottom sheet borders removed on DELETE_ROWS

Rachico:
### [FIX] border: bottom sheet borders removed on DELETE_ROWS

Problem
-----
Before this commit, when we add borders at the bottom of a sheet (at very the last rows), if we delete some rows in a position before the borders, some/all borders disappear (depending on how many rows we deleted). The reason being that the sheet plugin updates the total rows before we loop through them to shift the borders.

Solution
-----
This commit fixes this behaviour by redefining the way we get the border to shift.

Task: [3911695](https://www.odoo.com/web#id=3911695&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/core/borders.ts
+++ b/src/plugins/core/borders.ts
@@ -1,5 +1,5 @@
 import { DEFAULT_BORDER_DESC } from "../../constants";
-import { deepEquals, range, toCartesian, toXC, toZone } from "../../helpers/index";
+import { deepEquals, isDefined, range, toCartesian, toXC, toZone } from "../../helpers/index";
 import {
   AddColumnsRowsCommand,
   Border,
@@ -220,6 +220,21 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
   }
 
   /**
+   * Get all the rows which contains at least a border
+   */
+  private getRowsWithBorders(sheetId: UID): number[] {
+    const sheetBorders = this.borders[sheetId]?.filter(isDefined);
+    if (!sheetBorders) return [];
+    const rowsWithBorders = new Set<number>();
+    for (const rowBorders of sheetBorders) {
+      for (const rowBorder in rowBorders) {
+        rowsWithBorders.add(parseInt(rowBorder, 10));
+      }
+    }
+    return Array.from(rowsWithBorders);
+  }
+
+  /**
    * Get the range of all the rows in the sheet
    */
   private getRowsRange(sheetId: UID): number[] {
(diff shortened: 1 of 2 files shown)


tests/plugins/borders.test.ts
```diff
@@ -470,4 +471,14 @@ describe("Grid manipulation", () => {
     expect(getBorder(model, "B3")).toEqual({ bottom: b, top: b, left: b, right: b });
     expect(getBorder(model, "C3")).toEqual({ left: b });
   });
+
+  test("Remove multiple rows before borders at the bottom of the sheet", () => {
+    const b = DEFAULT_BORDER_DESC;
```
rrahir, line 476: unfortunately the test passes even without de fix. As I recall from the test I did, the spec and the comment I left here: https://github.com/odoo/o-spreadsheet/pull/4296/files/f5e93842f0a1857ee5feab967354eba8ef768b5e#r1621232954 
the issue arises when we set the borders at the bottom of the sheet.

Make sure that your test is actually fixing the issue ;-) A good way to proceed is to:
- first write the test, make sure it is sound (doing what is should, no less, no more)
- ensure the test fails
- fix the issus
- ensure te test now passes

rrahir (changes requested):

src/plugins/core/borders.ts
```diff
@@ -219,6 +219,15 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
     return Object.keys(sheetBorders).map((index) => parseInt(index, 10));
   }
 
+  /**
+   * Get all the rows which contains at least a border
+   */
+  private getRowsWithBorders(sheetId: UID): number[] {
+    const sheetBorders = this.borders[sheetId]?.filter(isDefined);
+    if (!sheetBorders) return [];
+    return Object.keys(sheetBorders[0]).map((index) => parseInt(index, 10));
```
LucasLefevre, line 228: I'm afraid  this fix is not enough :(
If you add borders on the last row but *not in the first column* (`sheetBorders[0]`), the same issue is still there

Rachico: Good catch!

src/plugins/core/borders.ts
```diff
@@ -219,6 +219,21 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
...
 
+  /**
+   * Get all the rows which contains at least a border
+   */
+  private getRowsWithBorders(sheetId: UID): number[] {
+    const sheetBorders = this.borders[sheetId]?.filter(isDefined);
+    if (!sheetBorders) return [];
+    const rowsWithBorders = new Set<number>();
+    for (const border of sheetBorders) {
+      for (const rowBorder of Object.keys(border)) {
```
rrahir:
```suggestion
      for (const rowBorder in border) {
```
small nitpick,i'll change it myself, but you can just iterate on the index of an array with `(element **in** array)`

rrahir (approved):

rrahir:
@robodoo r+

PR #4334, opened 2024-06-03
[FIX] collaborative: snapshot when leaving spreadsheet

LucasLefevre:
Purpose
-------

Let's say a user works a lot on a spreadsheet and performs a lot of heavy operations generating very large revisions (with lots of commands), such as copy-pasting a very large zone (there's one UPDATE_CELL command per copy-pasted cell)

If the user does multiple such very large revision, one after the other, every thing is fine client side.

The next time the spreadsheet is open though: the server needs to load all those revisions to send them to the client. The revisions can be so large it can blow up the server memory limit.
The spreadsheet cannot be open anymore.

Solution
--------

The solution proposed here is to snapshot the spreadsheet when the client leaves the spreadsheet. Next time the spreadsheet is open, the snapshot will be loaded, and not the revivisions.

There's a catch: snapshotting a spreadsheet kills the local history (CTRL+Z). For this reason, we only snapshot if there's no other connected client as it would kill the other users history.

Task: : [3940465](https://www.odoo.com/web#id=3940465&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


pro-odoo (approved):
robodoo r+

PR #4355, opened 2024-06-05
[FIX] FindAndReplaceSidePanel: Fix range update

rrahir:
How to reproduce:
- Open the F&R SidePanel.
- Choose the 'Specific Range' Search Option from the dropdown.
- Attempt to pick a large range from the grid.

The component SelectionInput heavily relies on the fact its props are updated by the parent component after SelectionInput invoked `props.onSelectionChanged`.

The dataRange stored in the component `FindAndReplacePanel` was not observed, which means the after `SelectionInput` called `onSelectionChanged`, the update of the datarange would not update the props passed to `SelectionInput`. This would cause a desynchronized state which lead to an infinite loop because one of the conditions in `SelectionInput.onWillUpdateProps` would be triggered 100% of the time and the said condition unfortunately updates a store, then forcing a render(true), hence the infinite loop.

In this revision, we fix the bad state management of `FindAndReplace` but is seems again that `SelectionInput` is wobbly, so is using it. We should try to fix its behaviour once and for all.

Task: 3972409

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/find_and_replace/find_and_replace.ts
+++ b/src/components/side_panel/find_and_replace/find_and_replace.ts
@@ -1,4 +1,4 @@
-import { Component, onMounted, useRef } from "@odoo/owl";
+import { Component, onMounted, useRef, useState } from "@odoo/owl";
 import { zoneToXc } from "../../../helpers";
 import { Store, useLocalStore } from "../../../store_engine";
 import { _t } from "../../../translation";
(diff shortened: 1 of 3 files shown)


pro-odoo (approved):
robodoo r+

PR #3809, opened 2024-03-09
[IMP] LinkDisplay: update grid selection before opening the editor 

rrahir:
Currently, a user can open the link editor of a cell while hovering it but not selectd. This behaviour is counterintuitive as once the link editor is opened, we will close it if we change the selection.

This revision ensures that the link editor target and the selection are synchronized upon the initialization of the the former.

Task: 3793859

## Description:

description of this task, what is implemented and why it is implemented that way.

Task: : [TASK_ID](https://www.odoo.com/web#id=TASK_ID&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] feature is organized in plugin, or UI components
- [ ] support of duplicate sheet (deep copy)
- [ ] in model/core: ranges are Range object, and can be adapted (adaptRanges)
- [ ] in model/UI: ranges are strings (to show the user)
- [ ] undo-able commands (uses this.history.update)
- [ ] multiuser-able commands (has inverse commands and transformations where needed)
- [ ] new/updated/removed commands are documented
- [ ] exportable in excel
- [ ] translations (\_t("qmsdf %s", abc))
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes
- [ ] doc is rebuild (npm run doc)
- [ ] status is correct in Odoo

PR #4380, opened 2024-06-07
[FIX] ComposerStore: missing previous range

dhrp-odoo:
## Description:

The updateComposerRange function previously assumed that a previous range would always be found because it was only called when changing a highlight, which was a valid assumption under past conditions.

However, this assumption is no longer accurate. There are now two scenarios where this function can be called without a previous range:

1. With [PR#3809](https://github.com/odoo/o-spreadsheet/pull/3809), entering edition mode in a cell and opening the link editor causes a traceback due to the missing previous range.
2. Opening a cell in editing mode and creating a data filter while the cell is in edition mode also results in the same traceback.

To resolve this issue, we can return early if no update or replacement is needed within the composer range.

Task: : [3972392](https://www.odoo.com/web#id=3972392&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_stateful/edition.ts
+++ b/src/plugins/ui_stateful/edition.ts
@@ -618,10 +618,8 @@ export class EditionPlugin extends UIPlugin {
         return isEqual(this.getters.expandZone(activeSheetId, refRange.zone), oldZone);
       });
 
-    // this function assumes that the previous range is always found because
-    // it's called when changing a highlight, which exists by definition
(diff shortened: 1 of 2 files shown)


LucasLefevre (commented):
robodoo r+

PR #4385, opened 2024-06-10
[FIX] charts: fix padding computation

anhe-odoo:
## Task Description

Since the chart design refactoring, the value of the title is now stored in an object as chart.title.text, instead of the previous chart.title. During the introduction of the design customization, the computation of the padding hasn't been updated, considering now that the title is alsway defined, which is not the case. This PR aims to thix this behavior by changing the condition used in the padding computation.

## Related Task:

- Task: 0

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


anhe-odoo:
@rrahir : here is the Odoo version : https://github.com/odoo/odoo/pull/168639

PR #4404, opened 2024-06-10
[FIX] grid: prevent paste default

rrahir:
Followup of https://github.com/odoo/o-spreadsheet/pull/4298 The issue of the replacement of DOM elements still occured when we'd paste a chart is a specific way (see #How to reproduce). This commit ensures that the default paste behaviour is prevented when the lib has a custom handler, not just through the composer.

In this case, it is `FigureContainer` that is affected. After the paste effect, it replaces itself outside the visible part of its container `GridOverlay`. This issue would arise in Firefox and not Chromium-based browser. It to be linked to how the two handle the repositioning of [`static`](https://developer.mozilla.org/en-US/docs/Web/CSS/position#static) elements. This revision addresses this by setting the  position of `FigureContainer` to absolute and not static.

How to reproduce:
On Firefox,
- select a chart and open its sidepanel
- cut it (Ctrl-x)
- Focus a selectionInput in the sidepanel
- click on the grid to update the selectionInput
- immediatly paste the chart (Ctrl-v) -> the page layout is broken. On the demo spreadsheet, the charts all disappeared and the "Add rows" button is visible even if the bottom of the grid is way below the viewport.

Task: 3949903

## Description:

Task: : [3949903](https://www.odoo.com/web#id=3949903&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/grid/grid.ts
+++ b/src/components/grid/grid.ts
@@ -628,6 +628,7 @@ export class Grid extends Component<Props, SpreadsheetChildEnv> {
     }
 
     if (clipboardData.types.indexOf(ClipboardMIMEType.PlainText) > -1) {
+      ev.preventDefault();
       const content = clipboardData.getData(ClipboardMIMEType.PlainText);
       const target = this.env.model.getters.getSelectedZones();
       const clipboardString = this.env.model.getters.getClipboardTextContent();
--- a/src/components/figures/figure_container/figure_container.xml
+++ b/src/components/figures/figure_container/figure_container.xml
@@ -1,6 +1,6 @@
 <templates>
   <t t-name="o-spreadsheet-FiguresContainer">
-    <div>
+    <div class="position-absolute">
       <t t-foreach="containers" t-as="container" t-key="container.type">
         <div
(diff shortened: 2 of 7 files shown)


src/components/grid/grid.ts
```diff
@@ -617,6 +617,7 @@ export class Grid extends Component<Props, SpreadsheetChildEnv> {
   }
 
   async paste(ev: ClipboardEvent) {
+    ev.preventDefault();
```
LucasLefevre, line 620: this prevents from pasting in all regular inputs :/

LucasLefevre (commented):
Looks like the layout is broken with Firefox with the demo spreadsheet :(

image

src/components/grid/grid.ts
rrahir: Indeed put in a very pad place :/

LucasLefevre (approved):
robodoo r+

PR #4422, opened 2024-06-11
[FIX] charts: fix scorecard chard design

anhe-odoo:
## Task Description

Since the refactoring of the chart design customization, the padding and title size have been uniformized for all chart. In the dashboards, as we have very small scorecard, it doesn't look really good now. This task fix this design issue, as well as fixing the ellipsis mechanism of the scorecard title.

## Related Task

Task: : [3964015](https://www.odoo.com/web#id=3964015&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/figures/chart/scorecard/chart_scorecard.ts
+++ b/src/components/figures/chart/scorecard/chart_scorecard.ts
@@ -1,6 +1,7 @@
 import { Component, useEffect, useRef } from "@odoo/owl";
 import { drawScoreChart } from "../../../../helpers/figures/charts/scorecard_chart";
 import { getScorecardConfiguration } from "../../../../helpers/figures/charts/scorecard_chart_config_builder";
+import { _t } from "../../../../translation";
 import { Figure, SpreadsheetChildEnv } from "../../../../types";
 import { ScorecardChartRuntime } from "../../../../types/chart/scorecard_chart";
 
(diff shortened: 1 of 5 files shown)


src/helpers/figures/charts/scorecard_chart_config_builder.ts
```diff
@@ -318,7 +320,10 @@ class ScorecardChartConfigBuilder {
     );
 
     let isBaselineSplit = false;
-    if (baselineDescrFontSize < baselineValueFontSize / 2.5) {
+    if (
+      baselineDescrFontSize < baselineValueFontSize / 2.5 &&
+      this.baselineDescr.trim().includes(" ")
```
rrahir, line 325: why the "space" condition?

anhe-odoo: To avoid making a false split when we validate the font size condition but we only have one word in the baselineDescr. Without this condition, we would try to split, resulting in a two lines description with the first line containing the word and the second line being empty.

src/components/figures/chart/scorecard/chart_scorecard.ts
```diff
@@ -19,6 +19,10 @@ export class ScorecardChart extends Component<Props, SpreadsheetChildEnv> {
     return this.env.model.getters.getChartRuntime(this.props.figure.id) as ScorecardChartRuntime;
   }
 
+  get title(): string {
+    return this.env.model.getters.getChartDefinition(this.props.figure.id).title.text ?? "";
```
LucasLefevre, line 23: black magic
```suggestion
    const title = this.env.model.getters.getChartDefinition(this.props.figure.id).title.text ?? "";
    // there's dark magic exporting chart titles from .json files and it's translated at runtime here
    return _t(title)
```

LucasLefevre:
robodoo r+

PR #4424, opened 2024-06-11
[FIX] components: add missing static props

hokolomopo:
## Description

Some components were missing their static props declaration.

Enable the setting `warnIfNoStaticProps` when mounting the app in demo and `mountSpreadsheet` in the tests.

I didn't activate the setting for every `mountComponent` because that means that we have to correctly declare the props of every random component that we create for the tests. But that means that tests can still pass if a component is only testes when mounted in standalone.

Task: : [3984200](https://www.odoo.com/web#id=3984200&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/tests/test_helpers/helpers.ts
+++ b/tests/test_helpers/helpers.ts
@@ -111,6 +111,7 @@ export function addTestPlugin(
 const realTimeSetTimeout = window.setTimeout.bind(window);
 class Root extends Component {
   static template = xml`<div/>`;
+  static props = {};
 }
 const Scheduler = new App(Root).scheduler.constructor as unknown as { requestAnimationFrame: any };
 
@@ -228,6 +229,7 @@ class ParentWithPortalTarget<Props extends ComponentProps> extends Component<
       <t t-component="props.childComponent" t-props="props.childProps"/>
     </div>
   `;
+  static props = { "*": Object };
 }
 
 interface MountComponentArgs<Props extends ComponentProps> {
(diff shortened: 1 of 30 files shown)


tests/test_helpers/helpers.ts
```diff
@@ -265,7 +266,13 @@ export async function mountComponent<Props extends { [key: string]: any }>(
   model.drawLayer = () => {};
   const env = makeTestEnv({ ...optionalArgs.env, model: model });
   const props = optionalArgs.props || ({} as Props);
-  const app = new App(component, { props, env, test: true, translateFn: _t });
+  const app = new App(component, {
+    props,
+    env,
+    test: true,
+    translateFn: _t,
+    warnIfNoStaticProps: optionalArgs.warnIfNoStaticProps,
```
hokolomopo, line 274: Should I enable this for every `MountComponent` ? That menas that every random `Parent` component that we decalre in the tests would need to be correctly declared...

rrahir: I think we should. I already lost time multiple times trying to figure out why a test wouldn't pass just because we had some random invalid props that where either false from the very beginning or were not properly adapted.

hokolomopo: Sure, I added them to every test component then :)

rrahir:
@robodoo r+

PR #4426, opened 2024-06-11
[FIX] pivot: don't crash whole pivot if aggregator in error

hokolomopo:
## Description

Before this, if the aggregator function of a pivot returned an error (eg. division by zero in average), the whole pivot would be in error.

This commit fixes it, making only the cell in error to be in error.

Task: : [3977814](https://www.odoo.com/web#id=3977814&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/helpers/pivot/spreadsheet_pivot/spreadsheet_pivot.ts
```diff
@@ -224,10 +225,20 @@ export class SpreadsheetPivot implements Pivot<SpreadsheetPivotRuntimeDefinition
...
-      format: operator.format(values[0]),
-    };
+
+    try {
+      return {
+        value: values.length ? operator.fn([values], this.getters.getLocale()) : "",
+        format: operator.format(values[0]),
+      };
+    } catch (e) {
+      const errorFPayload = {
```
rrahir, line 235: I think we should just proces the "expected" evaluation errors. Programatic errors SHOULD make the code crash. Id add a filtering with `isEvaluationError`. WDYT?

hokolomopo: I'm not against the idea, but I copied the behaviour of the evaluation (`evaluator.ts computeCell`), so it might be an intended design choice ?

src/helpers/pivot/spreadsheet_pivot/spreadsheet_pivot.ts
```diff
@@ -224,10 +225,23 @@ export class SpreadsheetPivot implements Pivot<SpreadsheetPivotRuntimeDefinition
...
+      if (!(e instanceof EvaluationError)) {
+        throw e;
+      }
+      const errorFPayload = {
+        value: e.value ? e.value : CellErrorType.GenericError,
+        message: e.message,
+      };
+      replaceFunctionNamePlaceholder(errorFPayload, aggregator.toUpperCase());
+      return errorFPayload;
+    }
```
LucasLefevre, line 244: would it work to export and reuse `handleError` ?
```suggestion
    } catch (e) {
      return handleError(e, aggregator.toUpperCase();
    }
```

rrahir: hurray then ^^

hokolomopo: Didn't see the helper, it sounds like a good idea :D

LucasLefevre (approved):
robodoo r+

dberardo-com:
somehow i am still experiencing this issue after having pulled from 17.3.4. i also still see the issue on the live playground, which affects not only pivots but also charts

hokolomopo:
> somehow i am still experiencing this issue after having pulled from 17.3.4. i also still see the issue on the live playground, which affects not only pivots but also charts

Hello,

The live playground is not updated automatically, so it's not on the latest realase.

But the issue seems to be fixed for me in 17.3 ? Could you provide a more detailed bug report ? Also I don't see how this can  affects chart.

dberardo-com:
as states at the beginning of the issue, the bug affects both pivot and charts. it is enough to go to the live playground and try to include a charts and see the same error popping out.

as for my own version, i did pull from this repo like (git pull remote_repo 1.3.4) and indeed i can see the commit in the list: https://github.com/dberardo-com/o-spreadsheet/commits/17.3.4-custom/

after pulling and restarting the app i could still not use charts in the spreadsheet.

i am going to give it another try today just to be sure

hokolomopo:
For charts are you talking about https://github.com/odoo/o-spreadsheet/issues/4485 ? This PR has nothing to do with that ...

dberardo-com:
then i misunderstood the scope of this PR, since i did mention charts at the beginning of the issue: https://github.com/odoo/o-spreadsheet/issues/4485#issuecomment-2173804608

is there any PR tracking the chart issue ?

PR #4446, opened 2024-06-12
[FIX] collaborative: snapshot only when no pending changes

LucasLefevre:
Since 7bcff34394, we snapshot the spreadsheet when the user leaves the spreadsheet and there is no other connected users.

However, if the user updates the spreadsheet (which sends a revision to the server) and leaves immediatly without waiting for the server's response, the snapshot is made even though it's completely useless. It won't be accepted because it's not sent with the correct last (accepted) server revision id.

The correct server revision id to send would be the one from the first sent revision, but we don't know yet if it's accepted by the server or not.

Task: 0

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/collaborative/session.ts
+++ b/src/collaborative/session.ts
@@ -176,6 +176,9 @@ export class Session extends EventBus<CollaborativeEvent> {
    * Send a snapshot of the spreadsheet to the collaboration server
    */
   snapshot(data: WorkbookData) {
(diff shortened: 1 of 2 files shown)


pro-odoo:
robodoo r+

PR #4452, opened 2024-06-13
[IMP] pivot: add menu item to fix formulas

pro-odoo:
With this commit, a user can easily fix formulas of its pivot. Fixing a formula is simply replacing the =PIVOT formula with a formula by cell.

Task: 3987157

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/pivot/pivot_menu_items.ts
+++ b/src/helpers/pivot/pivot_menu_items.ts
@@ -15,3 +15,38 @@ export const pivotProperties: ActionSpec = {
   },
   icon: "o-spreadsheet-Icon.PIVOT",
 };
+
+export const FIX_FORMULAS: ActionSpec = {
+  name: _t("Convert to individual formulas"),
+  execute(env) {
+    const position = env.model.getters.getActivePosition();
+    const cell = env.model.getters.getCorrespondingFormulaCell(position);
+    const pivotId = env.model.getters.getPivotIdFromPosition(position);
+    if (!cell || !pivotId) {
+      return;
+    }
+    const { sheetId, col, row } = env.model.getters.getCellPosition(cell.id);
+    const pivot = env.model.getters.getPivot(pivotId);
+    pivot.init();
+    if (!pivot.isValid()) {
+      return;
+    }
+    env.model.dispatch("INSERT_PIVOT", {
+      sheetId,
+      col,
+      row,
+      pivotId,
+      table: pivot.getTableStructure().export(),
+    });
+  },
+  isVisible: (env) => {
+    const position = env.model.getters.getActivePosition();
+    const pivotId = env.model.getters.getPivotIdFromPosition(position);
+    if (!pivotId) {
+      return false;
+    }
+    const pivot = env.model.getters.getPivot(pivotId);
+    return pivot.isValid() && env.model.getters.isSpillPivotFormula(position);
+  },
+  icon: "o-spreadsheet-Icon.PIVOT",
+};
--- a/src/plugins/ui_core_views/pivot_ui.ts
+++ b/src/plugins/ui_core_views/pivot_ui.ts
@@ -33,6 +33,7 @@ export class PivotUIPlugin extends UIPlugin {
     "getPivotDomainArgsFromPosition",
     "isPivotUnused",
     "areDomainArgsFieldsValid",
+    "isSpillPivotFormula",
   ] as const;
 
   private pivots: Record<UID, Pivot> = {};
@@ -118,6 +119,15 @@ export class PivotUIPlugin extends UIPlugin {
       }
     }
     return undefined;
+  }
+
+  isSpillPivotFormula(position: CellPosition) {
+    const cell = this.getters.getCorrespondingFormulaCell(position);
+    if (cell && cell.isFormula) {
+      const pivotFunction = this.getFirstPivotFunction(cell.compiledFormula.tokens);
+      return pivotFunction?.functionName === "PIVOT";
+    }
+    return false;
   }
 
   getFirstPivotFunction(tokens: Token[]) {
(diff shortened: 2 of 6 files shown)


src/helpers/pivot/pivot_menu_items.ts
```diff
@@ -15,3 +15,30 @@ export const pivotProperties: ActionSpec = {
   },
   icon: "o-spreadsheet-Icon.PIVOT",
 };
+
+export const PIVOT_SWITCH_STATIC: ActionSpec = {
+  name: _t("Fix pivot formulas"),
+  execute(env) {
+    const position = env.model.getters.getActivePosition();
+    const cell = env.model.getters.getCorrespondingFormulaCell(position);
```
rrahir, line 23: you don't have to do this check as it's already done the the following getter. If we consider that it's better with the double check, I would have an early return if `!cell` to avoid useless computations.

tests/test_helpers/pivot_helpers.ts
```diff
@@ -2,11 +2,75 @@ import { DispatchResult, Model, UID } from "../../src";
...
 export function createModelWithPivot(range: string): Model {
   return new Model(pivotModelData(range));
 }
 
+/**
+ * Creates a Model instance with a predefined pivot table setup.
+ * It contains a "PIVOT" formula in the cell A8
+ *
+ */
+export function modelWithPivot() {
```
LucasLefevre, line 16: I'm half-half excited about this. It usually leads to tests that are hard to read and understand a you don't know what the pivot looks like when reading the test.
I personally prefer to manually setup the smallest pivot possible in each test, even if it takes more lines.
What's your opinion ?

src/helpers/pivot/pivot_menu_items.ts
```diff
@@ -15,3 +15,30 @@ export const pivotProperties: ActionSpec = {
   },
   icon: "o-spreadsheet-Icon.PIVOT",
 };
+
+export const FIX_FORMULAS: ActionSpec = {
+  name: _t("Fix pivot formulas"),
```
LucasLefevre, line 20: Is it a magic button to fix all bugs in pivot formulas !!!??
That's so niiiice :astonished:  :upside_down_face:

src/helpers/pivot/pivot_menu_items.ts
```diff
@@ -15,3 +15,30 @@ export const pivotProperties: ActionSpec = {
...
+    }
+    const { sheetId, col, row } = env.model.getters.getCellPosition(cell.id);
+    const pivot = env.model.getters.getPivot(pivotId);
+    pivot.init();
+    env.model.dispatch("INSERT_PIVOT", {
+      sheetId,
+      col,
+      row,
+      pivotId,
+      table: pivot.getTableStructure().export(),
```
LucasLefevre, line 36: doesn't it explode if the pivot isn't loaded ?

AFAICT, `isSpillPivotFormula` returns `true` on the main cell even if it's not loaded yet

tests/test_helpers/pivot_helpers.ts
pro-odoo: > you don't know what the pivot looks like when reading the test.
Without this kind of helper, it's a mess to create every single pivot test :/
I'm not 100% for, but I think it's necessary.
The setup a pivot is not straightforward. It requires at least 15-20 lines of codes, that reduces the readability of the code IMO.
I want to add an helper to print the pivot, something similar to `printDebugModel` to help us.

src/helpers/pivot/pivot_menu_items.ts
pro-odoo: François will have to find a good name :D

src/helpers/pivot/pivot_menu_items.ts
pro-odoo: If the pivot is not loaded, it will explode yes. But in this action `pivot.init()` is called, so we're safe.
But I forgot Odoo, I will add a check that the pivot is loaded, it will make the job

tests/test_helpers/pivot_helpers.ts
LucasLefevre: 11 lines :stuck_out_tongue: 
```ts
    const grid = {
      A1: "Date",       B1: "Price",
      A2: "2024-12-31", B2: "10",
      A3: "1995-04-14", B3: "30",
    };
    const model = createModelFromGrid(grid);
    addPivot(model, "A1:B3", {
      columns: [],
      rows: [{ name: "Date", granularity: "year_number" }],
      measures: [{ name: "Price", aggregator: "sum" }],
    });
```

> I want to add an helper to print the pivot, something similar to printDebugModel to help us.

could be helpful :)

pro-odoo: Mmm I didn't know `createModelFromGrid` :o 
Ok then I will use that, it's not painful now. I'll see in my task to move the tests from odoo if we need it

pro-odoo: Updated, and I created `printPivot` helper ;-)

LucasLefevre (commented):
nice and simple :ok_hand:

src/helpers/pivot/pivot_menu_items.ts
```diff
@@ -15,3 +15,33 @@ export const pivotProperties: ActionSpec = {
...
+  execute(env) {
+    const position = env.model.getters.getActivePosition();
+    const cell = env.model.getters.getCorrespondingFormulaCell(position);
+    const pivotId = env.model.getters.getPivotIdFromPosition(position);
+    if (!cell || !pivotId) {
+      return;
+    }
+    const { sheetId, col, row } = env.model.getters.getCellPosition(cell.id);
+    const pivot = env.model.getters.getPivot(pivotId);
+    pivot.init();
```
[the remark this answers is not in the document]
pro-odoo: The menu should be hidden if the pivot is not valid, I added it. But this code is still needed to ensure that the pivot is loaded, as it could be invalided between the menu is opened and the action is selected

LucasLefevre (approved):
robodoo r+

PR #4456, opened 2024-06-13
[IMP] documentation: improve commands documentation

LucasLefevre:
Task: 0

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/doc/extending/command.md
+++ b/doc/extending/command.md
@@ -1,60 +1,89 @@
 # Commands
 
-Commands are the way to make changes to the state. They are dispatched to the model, which relay them to each plugins.
+Commands are essential for modifying the spreadsheet state. They are dispatched to the model, which in turn relays them to each plugin.
+There are two types of commands: `CoreCommand` and `LocalCommand`.
 
-There are two kinds of commands: `CoreCommands` and `LocalCommands`.
+## Types of Commands
 
-1. `CoreCommands` are commands that
+### CoreCommand
 
-   - manipulate the imported/exported spreadsheet state
-   - are shared in collaborative environment
+- **Purpose**: Low-level commands for updating the core spreadsheet state.
+- **Handling**: Managed by core plugins to modify their state.
+- **Sub-Commands**: Can dispatch sub-core commands but not sub-local commands.
+- **Collaboration**: Broadcast to other connected users in a collaborative environment.
 
-1. `LocalCommands`: every other command
-   - manipulate the local state
-   - can be converted into CoreCommands
-   - are not shared in collaborative environment
+### LocalCommand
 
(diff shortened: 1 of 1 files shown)


doc/extending/command.md
```diff
@@ -1,60 +1,89 @@
...
-1. `LocalCommands`: every other command
-   - manipulate the local state
-   - can be converted into CoreCommands
-   - are not shared in collaborative environment
+### LocalCommand
 
-For example, "RESIZE_COLUMNS_ROWS" is a CoreCommand. "AUTORESIZE_COLUMNS" can be (locally) converted into a "RESIZE_COLUMNS_ROWS", and therefore, is not a CoreCommand.
-CoreCommands should be "device agnostic". This means that they should contain all the information necessary to perform their job. Local commands can use inferred information from the local internal state, such as the active sheet.
+- **Purpose**: Higher-level commands for managing the UI state or dispatching low level core commands.
+- **Handling**: Handle UI plugins. They cannot be handled by core plugins
```
Rachico, line 18:
```suggestion
- **Handling**: Managed by UI plugins. They cannot be handled by core plugins
```

Rachico:
Much clearer and comprehensive now, thanks!

doc/extending/command.md
```diff
@@ -72,16 +101,17 @@ export function repeatAddColumnsRowsCommand(
...
+import { repeatLocalCommandTransformRegistry, genericRepeat } from "@odoo/o-spreadsheet";
+
 repeatLocalCommandTransformRegistry.add("MY_LOCAL_COMMAND", genericRepeat);
 ```
 
-The difference with the core commands lies in the presence of a third argument in the transformation function: the core commands that were dispatched during the handling of the local command. This is useful if the result of the command depends on the state of an UI Plugin. In this case there's no guarantee that the state of the UI Plugin will be the same when the command is repeated. Adapting the child core commands is then sometimes a valid way to adapt the local command, as they don't depend on any internal state.
+For local commands, the transformation function includes a third argument: the core (sub)commands dispatched during the handling of the root local command. This is useful if the result depends on the UI plugin's state, as there is no guarantee that the UI plugin's state will be the same when the command is repeated. Adapting the child core commands can be a valid way to adjust the local command, as they do not depend on any internal state.
```
rrahir, line 114:
```suggestion
For local commands, the transformation function includes a third argument: the core (sub)commands dispatched during the handling of the root local command. This is useful if the result depends on the UI plugins' state, as there is no guarantee that the UI plugins' state will be the same when the command is repeated. Adapting the child core commands can be a valid way to adjust the local command, as they do not depend on any internal state.
```
assuming we are encompassing all ui plugins

rrahir:
@robodoo r+

PR #4458, opened 2024-06-13
[FIX] pivot: support empty date

pro-odoo:
Steps to reproduce:
- Open the demo sheet with the pivot
- Remove random date value => December appears

Task: 3989717

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/pivot/spreadsheet_pivot/date_spreadsheet_pivot.ts
+++ b/src/helpers/pivot/spreadsheet_pivot/date_spreadsheet_pivot.ts
@@ -8,6 +8,9 @@ export function createDate(dimension: PivotDimension, value: FieldValue["value"]
   if (!(granularity in MAP_VALUE_DIMENSION_DATE)) {
     throw new Error(`Unknown date granularity: ${granularity}`);
   }
+  if (value === null) {
+    return null;
+  }
   if (!MAP_VALUE_DIMENSION_DATE[granularity].set.has(value)) {
     MAP_VALUE_DIMENSION_DATE[granularity].set.add(value);
     const date = toJsDate(value, locale);
--- a/src/helpers/pivot/spreadsheet_pivot/spreadsheet_pivot.ts
+++ b/src/helpers/pivot/spreadsheet_pivot/spreadsheet_pivot.ts
@@ -188,7 +188,8 @@ export class SpreadsheetPivot implements Pivot<SpreadsheetPivotRuntimeDefinition
     if (!finalCell) {
       return { value: "" };
     }
-    if (finalCell.value === null) {
+    // Value can be null but stringified (e.g. an empty date, as for now every date is stringified)
+    if (finalCell.value === null || finalCell.value === `${null}`) {
       return { value: _t("(Undefined)") };
     }
     if (dimension.type === "date") {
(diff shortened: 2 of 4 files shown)


tests/pivots/spreadsheet_pivot/spreadsheet_pivot.test.ts
```diff
@@ -446,6 +446,23 @@ describe("Spreadsheet Pivot", () => {
...
+    updatePivot(model, "1", {
+      columns: [{ name: "Created on", granularity: "month_number", order: "asc" }],
+      rows: [],
+      measures: [{ name: "Expected Revenue", aggregator: "sum" }],
+    });
+    setCellContent(model, "A26", `=pivot(1)`);
+    expect(getEvaluatedGrid(model, "B26:F26")).toEqual([
+      ["February", "March", "April", "Total", ""],
+    ]);
+    setCellContent(model, "A4", ""); //The only cell in the range with month = February
```
LucasLefevre, line 460: typically what we discussed: create a pivot with lots of things, then delete stuff that we don't want :smile: 
this makes the test convoluted and hard to read.

not longer:
```ts
  test("Date dimensions should support empty cells", () => {
    const grid = {
      A1: "Date",
      A2: "",
      A3: "2024-03-01",
      A4: "=pivot(1)",
    }
    const model = createModelFromGrid(grid);
    addPivot(model, "A1:A3", {
      columns: [{ name: "Date", granularity: "month_number" }],
      measures: [{ name: "__count" }],
    });
    expect(getEvaluatedGrid(model, "B4:E4")).toEqual([
      ["March", "(Undefined)", "Total", ""],
    ]);
  });
```

pro-odoo: I fully agree.

src/helpers/pivot/spreadsheet_pivot/date_spreadsheet_pivot.ts
```diff
@@ -10,29 +10,33 @@ export function createDate(dimension: PivotDimension, value: FieldValue["value"]
   }
   if (!MAP_VALUE_DIMENSION_DATE[granularity].set.has(value)) {
```
hokolomopo, line 11: Shouldn't we early return `if(value === null) return null;` before this if ? Or is there an use to bother adding null to every granularity set, and to stringified null in the values?

pro-odoo: For now the `null` value should be returned as a string, but I could early return `"null"`

pro-odoo: > For now the null value should be returned as a string, but I could early return "null"

My bad, `createDate` should returns `null`. It's the caller of this function which stringify the value.
It's updated with your suggestion :)

src/helpers/pivot/spreadsheet_pivot/spreadsheet_pivot.ts
```diff
@@ -188,7 +188,8 @@ export class SpreadsheetPivot implements Pivot<SpreadsheetPivotRuntimeDefinition
     if (!finalCell) {
       return { value: "" };
     }
-    if (finalCell.value === null) {
+    // Value can be null but stringified (e.g. an empty date, as for now every date is stringified)
```
hokolomopo, line 191: Null dates are not stringified anymore with the last change right ? Or are they stringified somewhere else ?

pro-odoo: They are stringified by the caller

hokolomopo:
@robodoo r+

PR #4466, opened 2024-06-14
[IMP] pivot: normalize values

pro-odoo:
Before this commit, all values of a pivot (e.g. each arguments, each
values available for a field, etc.) were all stringified before being
used. This was done in order to be able to compare them easily during
the introduction of pivots.
Of course, this is not optimal as it implies a lot of useless conversion
(e.g to compare the year 2024 to the value 2024 in the function, we
stringified the two values, which was not efficient at all).

Here are some examples of the process, before this commit:

To filter the DataEntries based on a PivotNode (an pair field-value):
- Stringify the result of `toNormalizedPivotValue` (this function
  already parse the value to a CellValue)
- Stringify the each value of the field in the DataEntries.
- Compare them

From PIVOT.HEADER or PIVOT.VALUE formulas:
- Call `toString` on each arguments
- Filter the DataEntries based on the arguments for each node (see above)
- For a date time => Re-normalize the value in order to use the
  `PivotTimeAdapter` to get the format and value to return (as FPayload)

From PIVOT formula:
- Values to construct the SpreadsheetPivotTable are stringified
- Filter the DataEntries based on the arguments for each node (see above)
- For a date time => Re-normalize the value in order to use the
  `PivotTimeAdapter` to get the format and value to return (as FPayload)

Now, the values are directly in the right type (boolean, string or
number), based on the type of the field. It is done either in the
construction of the SpreadsheetPivotTable (the values from the cells are
directly in the right type => nothing to do) or in the PIVOT.HEADER or
PIVOT.VALUE formulas (by calling `toNormalizedPivotValue` on each
arguments).

So, the processes listed above are now:

To filter the DataEntries based on a PivotNode (an pair field-value):
- Compare the value to each value in the DataEntry. We now compare
  boolean with boolean, string with string and number with number.

From PIVOT.HEADER or PIVOT.VALUE formulas:
- Normalize each arguments, `toString` for the field, and
  `toNormalizedPivotValue` for the value
- Filter the DataEntries based on the arguments for each node (see above)
- Return the value, directly in the correct type.

From PIVOT formula:
- Values to construct the SpreadsheetPivotTable are not converted
- Filter the DataEntries based on the arguments for each node (see above)
- Return the value, directly in the correct type.

In addition, the type of the normalized value for date with granularity
`day` is now number instead of string. It reduced by 83% the time to
load a pivot with 5k rows grouped by a date field on granularity `day`.

To be able to support the positional arguments for the Odoo pivot, the
parsing of the domain is now done in the pivot itself. The validation
of the domain is also done in the pivot.A nice thing is that we can get
rid of the mention of `#` in `areDomainArgsFieldsValid` !.

This commit also removes some useless methods in the `PivotTimeAdapter`.

Task: 3991743

### DIFF
--- a/src/helpers/pivot/pivot_helpers.ts
+++ b/src/helpers/pivot/pivot_helpers.ts
@@ -14,7 +14,7 @@ import {
   PivotField,
   PivotTableCell,
 } from "../../types/pivot";
-import { isDefined } from "../misc";
+import { PivotRuntimeDefinition } from "./pivot_runtime_definition";
 import { pivotTimeAdapter } from "./pivot_time_adapter";
 
 const AGGREGATOR_NAMES = {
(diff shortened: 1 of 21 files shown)


src/helpers/pivot/pivot_helpers.ts
```diff
@@ -159,7 +159,7 @@ export function toPivotDomain(domainStr: string[]) {
 }
 
 export function flatPivotDomain(domain: PivotDomain) {
-  return domain.flatMap((arg) => [arg.field, arg.value]);
+  return domain.flatMap((arg) => [arg.field, arg.value === null ? "false" : arg.value]);
```
pro-odoo, line 162: Should be NA()

src/helpers/pivot/spreadsheet_pivot/table_spreadsheet_pivot.ts
```diff
@@ -158,7 +171,7 @@ export class SpreadsheetPivotTable {
   private getColMeasure(col: number) {
     const domain = this.getColHeaderDomain(col, this.columns.length - 1);
     const measure = domain?.at(-1)?.value;
-    if (measure === undefined) {
+    if (measure === undefined || measure === null) {
```
rrahir, line 174: small typo in the comment below

tests/pivots/pivot_helpers.test.ts
```diff
@@ -183,3 +186,76 @@ describe("toNormalizedPivotValue", () => {
...
+    expect(toFunctionPivotValue("11/2020", dimension)).toBe("44136");
+    expect(toFunctionPivotValue("2020", dimension)).toBe("2020");
+    expect(toFunctionPivotValue("01/11/2020", dimension)).toBe("43841");
+    expect(() => toFunctionPivotValue("false", dimension)).toThrow();
+    expect(toFunctionPivotValue(false, dimension)).toBe("0");
+    expect(() => toFunctionPivotValue("true", dimension)).toThrow();
+  });
+
+  test("Format values of time fields", () => {
+    for (const fieldType of ["date", "datetime"]) {
```
rrahir, line 218: nitpick - could be written as test.forEach

LucasLefevre (approved):
scroll scroll

rrahir:
@robodoo r+

rrahir:
@robodoo r+

PR #4482, opened 2024-06-17
[IMP] tables: add test for header inserted next to table

hokolomopo:
This commit adds a test that checks that a table is not expanded when inserting a header next to it.

This is more a limitation of our current behaviour of `adaptRanges` than a feature. In fact, this seems bugged when drag & drop a column of a table to the left of the table, we expect the table to include this column but it does not.

But changing this behaviour would either require tables to not use `Range`, or to have a complete overhaul of the way ranges work, which will probably break some spreadsheets and does not seems worth it ATM.

Task: 0

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


rrahir:
robodoo r+

PR #4497, opened 2024-06-19
[FIX] model: mark default translations as loaded

hokolomopo:
## Description

Currently the default translations function is marked as not loaded. That means that if no custom translation method is set, the translations will never be marked as loaded, and charts will crash because we try to deepCopy LazyTranslatedString.

However we can't mark the default translations as loaded by default, because otherwise top-level translations would always be translated before we can set the translation method.

The solution is to mark the default translations as loaded only when starting the model.

Task: : [3999116](https://www.odoo.com/web#id=3999116&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/translation.ts
+++ b/src/translation.ts
@@ -2,9 +2,11 @@ type SprintfValues = (string | String | number)[] | [{ [key: string]: string | n
 
 export type TranslationFunction = (string: string, ...values: SprintfValues) => string;
 
-// define a mock translation function, when o-spreadsheet runs in standalone it doesn't translate any string
-let _translate: TranslationFunction = (s) => s;
-let _loaded: () => boolean = () => false;
+const defaultTranslate: TranslationFunction = (s: string) => s;
+const defaultLoaded: () => boolean = () => false;
+
+let _translate = defaultTranslate;
+let _loaded = defaultLoaded;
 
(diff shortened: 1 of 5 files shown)


src/translation.ts
```diff
@@ -26,6 +29,19 @@ export function setTranslationMethod(tfn: TranslationFunction, loaded: () => boo
...
 }
 
+/**
+ * If no translation function has been set, this will mark the translation are loaded.
+ *
+ * By default, the translations should not be set as loaded, otherwise top-level translated constants will never be
+ * translated. But if by the time the model is instantiated no custom translation function has been set, we can set
+ * the default translation function as loaded so o-spreadsheet can be run in standalone with no translations.
+ */
+export function setDefaultTranslationAsLoaded() {
```
pro-odoo, line 39: IMO translation loaded is something internal, I would rename it to `setDefaultTranslationMethod`. It's clearer from the caller.

pro-odoo (commented):
Just nitpicking, feel free to ignore if you do not agree

pro-odoo (approved):
robodoo r+

PR #4499, opened 2024-06-20
[FIX] pivot: do not allow date dimension without granularity

pro-odoo:
When a date dimension is used in a pivot view, the granularity must be specified. Before this commit, it automatically defaulted to 'month_number'. With the experience of the odoo pivot, it's not a good idea to have a default granularity, as it's not easily migrable. In addition, supported a date dimension without granularity leads to lots of "if-else" in the code to handle this.

So, we prefer to avoid this situation and raise an error when a date dimension is used without a granularity.

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/demo/data.js
+++ b/demo/data.js
@@ -2524,7 +2524,7 @@ export const demoData = {
     1: {
       type: "SPREADSHEET",
       columns: [{ name: "Stage" }],
-      rows: [{ name: "Created on", order: "asc" }],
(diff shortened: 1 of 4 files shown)


LucasLefevre (approved):
robodoo r+

LucasLefevre:
robodoo r+

PR #4517, opened 2024-06-25
[IMP] FindAndReplace: warning when replacing formula

dhrp-odoo:
## Description:

This PR introduces a warning for users attempting to replace a formula or a part of a formula with a string.

Task: : [3942592](https://www.odoo.com/web#id=3942592&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/find_and_replace/find_and_replace_store.ts
+++ b/src/components/side_panel/find_and_replace/find_and_replace_store.ts
@@ -5,6 +5,8 @@ import { CellPosition, Color, Command, Highlight } from "../../../types";
 import { canonicalizeNumberContent } from "../../../helpers/locale";
 import { Get } from "../../../store_engine";
 import { SpreadsheetStore } from "../../../stores";
+import { NotificationStore } from "../../../stores/notification_store";
+import { _t } from "../../../translation";
 import { SearchOptions } from "../../../types/find_and_replace";
 
 const FIND_AND_REPLACE_HIGHLIGHT_COLOR: Color = "#8B008B";
@@ -32,6 +34,9 @@ export class FindAndReplaceStore extends SpreadsheetStore implements HighlightPr
   private isSearchDirty = false;
   private initialShowFormulaState: boolean;
   private preserveSelectedMatchIndex: boolean = false;
+  private irreplaceableMatchCount: number = 0;
+
+  private notificationStore = this.get(NotificationStore);
 
   // fixme: why do we make selectedMatchIndex on top of a selected
   // property in the matches?
@@ -119,6 +124,13 @@ export class FindAndReplaceStore extends SpreadsheetStore implements HighlightPr
         for (const match of cmd.matches) {
           this.replaceMatch(match, cmd.searchString, cmd.replaceWith, cmd.searchOptions);
         }
+
+        if (this.irreplaceableMatchCount > 0) {
+          this.showReplaceWarningMessage(cmd.matches.length, this.irreplaceableMatchCount);
+        }
+
+        this.irreplaceableMatchCount = 0;
+        break;
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/components/side_panel/find_and_replace/find_and_replace_store.ts
```diff
@@ -307,18 +312,57 @@ export class FindAndReplaceStore extends SpreadsheetStore implements HighlightPr
...
+    } else {
+      this.notificationStore.notifyUser({
+        type: "warning",
+        sticky: false,
+        text: _t(
+          replaceableMatches +
+            " match(es) replaced. " +
+            irreplaceableMatches +
+            " match(es) cannot be replaced as they are part of a formula."
+        ),
```
LucasLefevre, line 363: Dynamic strings can't be translated properly:

- only static strings are extracted from the code to be sent to translators
- they can't be translated as translators won't have translated every possible value of the dynamic parts. There won't be any match in the translation.

See https://www.odoo.com/documentation/17.0/developer/howtos/translations.html#variables. The doc is for python-side, but the same principles apply client-side

```suggestion
        text: _t("%(replaceable_count)s match(es) replaced. %(irreplaceable_count)s match(es) cannot be replaced as they are part of a formula.", { replaceable_count: ...,  irreplaceable_count: ...}),
```

Side note: when there are more than one placeholder, always use named placeholder (`%(replaceable_count)s`) instead of `%s`. It allows translators to reorder them if needed in some languages.

src/components/side_panel/find_and_replace/find_and_replace_store.ts
```diff
@@ -307,18 +312,58 @@ export class FindAndReplaceStore extends SpreadsheetStore implements HighlightPr
...
       replaceWith: this.toReplace,
-      matches: this.searchMatches,
+      matches: searchMatches,
       searchOptions: this.searchOptions,
     });
+
+    if (this.irreplaceableMatchCount > 0) {
+      this.showReplaceWarningMessage(searchMatches.length, this.irreplaceableMatchCount);
+    }
+    this.irreplaceableMatchCount = 0;
```
LucasLefevre, line 339: this little duplicated snippet of code could be moved at `case "REPLACE_SEARCH"`, right ?

tests/find_and_replace/find_and_replace_store.test.ts
```diff
@@ -808,3 +811,90 @@ test("Selecting a previous match located in a previous sheet will select the las
...
+
+describe("replace warnings", () => {
+  beforeEach(() => {
+    setCellContent(model, "A1", "2024");
+    setCellContent(model, "A2", "=DATE(2024, 1, 1)");
+
+    createSheet(model, { sheetId: "sh2", activate: true });
+
+    setCellContent(model, "A1", "2024");
+    setCellContent(model, "A2", "=DATE(2024, 1, 1)");
```
LucasLefevre, line 823: it's usually better to keep test setup as minimal as possible and tests as self-contained as possible. Here the second sheet is only useful in 2 tests out of 5 AFAICT.

I'd move the creation of the second sheet only in those 2 tests, even if a few lines are duplicated.

To be fair, I'd even completely remove this `beforeEach` and setup the minimal cells in each test (as an example, the first test only needs one cell to be relevant)

LucasLefevre (commented):
LGTM :+1: 
only two suggestions

src/components/side_panel/find_and_replace/find_and_replace_store.ts
dhrp-odoo: Yes, good idea! 🔄

Changes made, Thanks!

LucasLefevre (approved):
robodoo r+

PR #4531, opened 2024-07-01
Master perf get imported sheet size laa

laa-odoo:
Task: : [4022900](https://www.odoo.com/web#id=4022900&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)


src/plugins/core/sheet.ts
```diff
@@ -1002,12 +1002,18 @@ export class SheetPlugin extends CorePlugin<SheetState> implements SheetState {
...
-    let colNumber = data.colNumber;
-    for (let { col, row } of positions) {
-      rowNumber = Math.max(rowNumber, row + 1);
-      colNumber = Math.max(colNumber, col + 1);
+    let rowNumber = data.rowNumber - 1;
+    let colNumber = data.colNumber - 1;
+    for (const { col, row } of positions) {
+      if (row > rowNumber) {
+        rowNumber = row;
+      }
```
hokolomopo, line 1010: I tested a bit on the large dataset sheet, and didn't really notice any improvement between master and your PR :x

I really don't think that optimizing the number of "+1" that we do can have any impact on the performances, beside making the code a bit harder to read. 

I guess using `>` rather than Math.max could be slighthly faster, but again, I don't see a difference in practice.

I think we could optimize the function a bit by not looping twice on each cell (on in the main loop, and one in the `map` above). Maybe we loop thrice depending on how Object.keys is implemented ? 

What do you think ?

VincentSchippefilt: I agree with @hokolomopo here. A loop like 
```typescript
    for (const xc in data.cells) {
      const { col, row } = toCartesian(xc)
``` 
could avoid a guarantee of double allocation of an array of size `data.cells`. I don't know if it's a meaningful difference though...

laa-odoo: My bad, it was totally obvious...

Concerning the measurements, I have too large a standard deviation with my device to be able to observe a safe difference with `makeLargeDataset`
Here I observe a difference only by measuring isolated code.

Trying to save time on already low percentages is complicated to measure. 
But on several small percentages it may be worth it

PR #4534, opened 2024-07-01
[IMP] charts: add trending line

anhe-odoo:
## Task Description

This task aims to add the ability to add a trending line per data serie in a chart 'with axis' (line, bar, scatter, combo). We have now 4 types of trending line (polynomial, exponential, linear and logarithmic) and the ability to edit the color of the line and the order of the polynomial.

## Related Task

- Task: [3997500](https://www.odoo.com/web#id=3997500&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/chart/chart_with_axis/design_panel.ts
+++ b/src/components/side_panel/chart/chart_with_axis/design_panel.ts
@@ -1,11 +1,13 @@
 import { Component, useState } from "@odoo/owl";
-import { getNthColor, toHex } from "../../../../helpers";
+import { getNthColor, setColorAlpha, toHex } from "../../../../helpers";
 import { getDefinedAxis } from "../../../../helpers/figures/charts";
 import { _t } from "../../../../translation";
 import {
   ChartWithAxisDefinition,
+  Color,
   DispatchResult,
   SpreadsheetChildEnv,
+  TrendConfiguration,
   UID,
 } from "../../../../types/index";
 import { ChartTerms } from "../../../translations_terms";
@@ -64,11 +66,7 @@ export class ChartWithAxisDesignPanel extends Component<Props, SpreadsheetChildE
   }
 
   getDataSeries() {
-    const runtime = this.env.model.getters.getChartRuntime(this.props.figureId);
-    if (!runtime || !("chartJsConfig" in runtime)) {
-      return [];
-    }
-    return runtime.chartJsConfig.data.datasets.map((d) => d.label);
+    return this.props.definition.dataSets.map((d, i) => d.label ?? `${ChartTerms.Series} ${i + 1}`);
   }
 
   updateSerieEditor(ev) {
@@ -82,8 +80,10 @@ export class ChartWithAxisDesignPanel extends Component<Props, SpreadsheetChildE
   }
 
   updateDataSeriesColor(color: string) {
-    const dataSets = this.props.definition.dataSets;
-    if (!dataSets?.[this.state.index]) return;
+    const dataSets = [...this.props.definition.dataSets];
+    if (!dataSets?.[this.state.index]) {
+      return;
+    }
     dataSets[this.state.index] = {
       ...dataSets[this.state.index],
       backgroundColor: color,
@@ -93,15 +93,19 @@ export class ChartWithAxisDesignPanel extends Component<Props, SpreadsheetChildE
 
   getDataSerieColor() {
     const dataSets = this.props.definition.dataSets;
-    if (!dataSets?.[this.state.index]) return "";
+    if (!dataSets?.[this.state.index]) {
+      return "";
+    }
     const color = dataSets[this.state.index].backgroundColor;
     return color ? toHex(color) : getNthColor(this.state.index);
   }
 
   updateDataSeriesAxis(ev) {
     const axis = ev.target.value;
-    const dataSets = this.props.definition.dataSets;
-    if (!dataSets?.[this.state.index]) return;
+    const dataSets = [...this.props.definition.dataSets];
+    if (!dataSets?.[this.state.index]) {
+      return;
+    }
     dataSets[this.state.index] = {
       ...dataSets[this.state.index],
       yAxisId: axis === "left" ? "y" : "y1",
@@ -111,7 +115,9 @@ export class ChartWithAxisDesignPanel extends Component<Props, SpreadsheetChildE
 
   getDataSerieAxis() {
     const dataSets = this.props.definition.dataSets;
-    if (!dataSets?.[this.state.index]) return "left";
+    if (!dataSets?.[this.state.index]) {
+      return "left";
+    }
     return dataSets[this.state.index].yAxisId === "y1" ? "right" : "left";
   }
 
@@ -121,8 +127,10 @@ export class ChartWithAxisDesignPanel extends Component<Props, SpreadsheetChildE
 
   updateDataSeriesLabel(ev) {
     const label = ev.target.value;
-    const dataSets = this.props.definition.dataSets;
-    if (!dataSets?.[this.state.index]) return;
+    const dataSets = [...this.props.definition.dataSets];
+    if (!dataSets?.[this.state.index]) {
+      return;
+    }
     dataSets[this.state.index] = {
       ...dataSets[this.state.index],
       label,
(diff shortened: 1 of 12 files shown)


src/components/side_panel/chart/chart_with_axis/design_panel.ts
```diff
@@ -17,6 +21,12 @@ import {
 } from "../building_blocks/axis_design/axis_design_editor";
 import { GeneralDesignEditor } from "../building_blocks/general_design/general_design_editor";
 
+css/* scss */ `
+  .o-degree-input {
+    width: 50%;
```
hokolomopo, line 26: bootstrap w-50 ?

hokolomopo: actually you did use bootrap in the end, this css is unused :)

src/components/side_panel/chart/chart_with_axis/design_panel.ts
```diff
@@ -65,7 +76,8 @@ export class ChartWithAxisDesignPanel extends Component<Props, SpreadsheetChildE
     if (!runtime || !("chartJsConfig" in runtime)) {
       return [];
     }
-    return runtime.chartJsConfig.data.datasets.map((d) => d.label);
+    const datasets = runtime.chartJsConfig.data.datasets;
+    return this.props.definition.dataSets.map((d, i) => datasets[i].label);
```
hokolomopo, line 80: This seems very error prone, that will break if we ever add trendlines between datasets :x

src/components/side_panel/chart/chart_with_axis/design_panel.ts
```diff
@@ -131,4 +143,74 @@ export class ChartWithAxisDesignPanel extends Component<Props, SpreadsheetChildE
     const dataSets = this.props.definition.dataSets;
     return dataSets[this.state.index]?.label || this.getDataSeries()[this.state.index];
   }
+
+  updateDataTrend(display: boolean) {
```
hokolomopo, line 147:
```suggestion
  toggleDataTrend(display: boolean) {
```
maybe

src/components/side_panel/chart/chart_with_axis/design_panel.ts
```diff
@@ -131,4 +143,74 @@ export class ChartWithAxisDesignPanel extends Component<Props, SpreadsheetChildE
...
+    const dataSets = this.props.definition.dataSets;
+    if (!dataSets?.[this.state.index]) return;
```
hokolomopo, line 149: some people in the team don't like one line `if` statements ...

src/components/side_panel/chart/chart_with_axis/design_panel.ts
```diff
@@ -131,4 +143,74 @@ export class ChartWithAxisDesignPanel extends Component<Props, SpreadsheetChildE
...
+    dataSets[this.state.index] = {
+      ...dataSets[this.state.index],
+      trend: {
+        ...dataSets[this.state.index].trend,
+        color,
+      },
+    };
+    this.props.updateChart(this.props.figureId, { dataSets });
```
hokolomopo, line 214: This part seems repeated 4 time in the panel. Maybe a method `updateTrendValue(key, value)` would help ?

src/helpers/figures/charts/chart_common.ts
```diff
@@ -428,3 +437,118 @@ export function getDefinedAxis(definition: ChartWithAxisDefinition): {
...
+export function getTrendDatasetForLineChart(
+  config: TrendConfiguration,
+  dataset: any,
+  labelsAsText?: boolean
+) {
+  const values: number[] = [];
+  const labels: number[] = [];
+  const datasetLength = dataset.data.length;
+  const newLabels: number[] = [];
+  if (labelsAsText) {
```
hokolomopo, line 482: not the correct condition, labelAsText can be false whithout the chart having `{x, y}` points. You should test the chart axis type.

The trenline is breaking if there is non-number strings in the labels

src/helpers/figures/charts/chart_common.ts
```diff
@@ -428,3 +437,118 @@ export function getDefinedAxis(definition: ChartWithAxisDefinition): {
...
+      labels.push(point.x - 0);
+    }
+  }
+
+  const xmin = Math.min(...labels);
+  const xmax = Math.max(...labels);
+  const step = (xmax - xmin) / 20;
+  for (let i = 0; i < 21; i++) {
+    newLabels.push(xmin + i * step);
+  }
```
hokolomopo, line 499: can use `range()` helper

src/helpers/figures/charts/chart_common_line_scatter.ts
```diff
@@ -246,10 +245,12 @@ export function createLineOrScatterChartRuntime(
     config.options.scales!.x!.ticks!.callback = (value) =>
       formatValue(value, { format: labelFormat, locale });
     config.options.plugins!.tooltip!.callbacks!.title = (tooltipItem) => {
-      return formatValue(tooltipItem[0].parsed.x || tooltipItem[0].label, {
-        locale,
-        format: labelFormat,
-      });
+      return tooltipItem[0].datasetIndex! < dataSetsCount
```
hokolomopo, line 248: same as my first comment in the side panl, this will easily break if we change the order of datasets

hokolomopo (commented):
Hello :wave:

src/helpers/figures/charts/chart_common.ts
```diff
@@ -428,3 +437,80 @@ export function getDefinedAxis(definition: ChartWithAxisDefinition): {
...
   return { useLeftAxis, useRightAxis };
 }
+
+export function getTrendDatasetForBarChart(config: TrendConfiguration, dataset: any) {
+  const values: number[] = [];
+  const labels: number[] = [];
+  for (let i = 0; i < dataset.data.length; i++) {
+    values.push(dataset.data[i]);
+    labels.push(i + 1);
+  }
```
LucasLefevre, line 447: Is there any different behavior between this and line charts with `axisType` `"category"` ? could the same code be re-used ?

LucasLefevre (commented):
very very cool :)
(rebase required)

src/components/side_panel/chart/chart_with_axis/design_panel.ts
```diff
@@ -131,4 +142,64 @@ export class ChartWithAxisDesignPanel extends Component<Props, SpreadsheetChildE
...
+  toggleDataTrend(display: boolean) {
+    const dataSets = this.props.definition.dataSets;
+    if (!dataSets?.[this.state.index]) {
+      return;
+    }
+    dataSets[this.state.index] = {
+      ...dataSets[this.state.index],
+      trend: {
+        type: "polynomial",
+        order: 3,
```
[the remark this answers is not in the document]
anhe-odoo: There is no proof that the 3rd order polynomial would be a better choice but it will fit most dataset than a linear model. Anyway, going with the simplest model as default seem legit.

For the automatic choice, it would be realy heavy as we would have to make some matrix computation for each model to be compared for a data serie and then choose the best. I have another idea that will require some IA trained to automatically detect some parameters on the dataset and then apply the correct model (which would then be an option "Best fit"). It would be something not heavy at all (once trained, of course), but it will be for another task, one day :)

src/components/side_panel/chart/chart_with_axis/design_panel.ts
```diff
@@ -131,4 +142,64 @@ export class ChartWithAxisDesignPanel extends Component<Props, SpreadsheetChildE
...
+  onChangeTrendType(ev: InputEvent) {
+    const type = (ev.target as HTMLInputElement).value as TrendType;
+    this.updateTrendLineValue("type", type);
+  }
+
+  onChangePolynomialDegree(ev: InputEvent) {
+    const element = ev.target as HTMLInputElement;
+    const order = parseInt((ev.target as HTMLInputElement).value);
+    if (order < 1) {
+      element.value = `${this.getTrendLineConfiguration()?.order ?? 1}`;
```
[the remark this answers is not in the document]
anhe-odoo: Even with the "min="1", you can still write 0 or any negative value in the spinBox. The min anx max properties only limit the range that you can reach with up/down

src/components/side_panel/chart/chart_with_axis/design_panel.ts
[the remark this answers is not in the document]
LucasLefevre: Let's go for linear :)

src/helpers/figures/charts/chart_common.ts
```diff
@@ -428,3 +437,80 @@ export function getDefinedAxis(definition: ChartWithAxisDefinition): {
...
+      const order = config.order ?? 2;
+      const coeffs = polynomialRegression(values, labels, order, true).flat();
+      return newLabels.map((v) => evaluatePolynomial(coeffs, v, order));
+    }
+    case "linear": {
+      return predictLinearValues([values], [labels], [newLabels], true)[0];
+    }
+    case "exponential": {
+      const filteredValues: number[] = [];
+      const filteredLabels: number[] = [];
```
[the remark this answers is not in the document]
anhe-odoo: Ok for the `positiveLogValues`, but `labels` is already taken and here `filteredLabels` describe exactly what it does

src/helpers/figures/charts/chart_common.ts
```diff
@@ -428,3 +437,80 @@ export function getDefinedAxis(definition: ChartWithAxisDefinition): {
...
+  labels: number[],
+  newLabels: number[]
+): number[] {
+  switch (config.type) {
+    case "polynomial": {
+      const order = config.order ?? 2;
+      const coeffs = polynomialRegression(values, labels, order, true).flat();
+      return newLabels.map((v) => evaluatePolynomial(coeffs, v, order));
+    }
+    case "linear": {
```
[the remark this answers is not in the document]
anhe-odoo: Yes, it can be, but the computation method used here for the linear model is really faster than the one for the polynomial model. So, as the method were already there, I chose to separate both (just like GSheet and Excel does, they two have polynomial and linear models)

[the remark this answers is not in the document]
LucasLefevre: It makes sense in all cases to split both in the UI (people don't necessarily know what is a polynome and even less that a 1st order polynome is linear), but it could be only one thing in the background.
Shouldn't "polynomial order=1" also go through the linear case then if it's faster ?

src/helpers/figures/charts/chart_common.ts
anhe-odoo: Yes, there is a difference for the axis. For 10 values, the bar chart axis will go from 0.5 to 10.5, while the line chart will go from 1 to 10.

LucasLefevre (commented):
Some more comments.
There's a crash (see video in a comment below)

src/helpers/figures/charts/bar_chart.ts
```diff
@@ -354,6 +350,32 @@ export function createBarChartRuntime(chart: BarChart, getters: Getters): BarCha
...
+    /* We add a second x axis here to draw the trend lines, with the labels length being
+     * set so that the second axis points match the classical x axis
+     */
+    config.options.scales[TREND_LINE_XAXIS_ID] = {
+      ...xAxis,
+      labels: Array(maxLength).fill(""),
+      offset: false,
+      display: false,
+    };
+    /* DO NOT CHANGE THIS BEHAVIOR !!! These datasets must be inserted after the original
```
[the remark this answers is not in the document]
anhe-odoo: There is already a test for it :)

src/helpers/figures/charts/chart_common_line_scatter.ts
```diff
@@ -247,11 +317,19 @@ export function createLineOrScatterChartRuntime(
     config.options.scales!.x!.type = "linear";
     config.options.scales!.x!.ticks!.callback = (value) =>
       formatValue(value, { format: labelFormat, locale });
-    config.options.plugins!.tooltip!.callbacks!.title = (tooltipItem) => {
-      return formatValue(tooltipItem[0].parsed.x || tooltipItem[0].label, {
-        locale,
-        format: labelFormat,
-      });
+    config.options.plugins!.tooltip!.callbacks!.title = () => "";
```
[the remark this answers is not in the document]
anhe-odoo: Yes. The title was only containing the "x" value, which is now shown in the label

LucasLefevre (approved):
robodoo r+

PR #4572, opened 2024-07-03
[REV] chart: avoid useless chart updates

hokolomopo:
## Description

This reverts commit 6eb43533d.

It turns out that checking the deep equality of runtime while ignoring functions is not a good fix to avoid useless chart updates. The problem is that some runtime changes apply only to the callbacks (eg. changing the dataset format only change the ticks callback).

The only real alternative would be to add the variables that are used in the callbacks to the runtime, so the deepEquals would work. But this is very error prone: we'll 100% forget to add a variable at some point.

So we will accept the useless updates for now, until we see a real performance issue.

Task: : [4029016](https://www.odoo.com/web#id=4029016&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


LucasLefevre (approved):
robodoo r+

PR #4598, opened 2024-07-05
[IMP] CF: Capture invalid formula for cellIsRule

rrahir:
A user could input an invalid formula in for "cellIsRule" type of conditional formats. Up until recently, this did not pause any problem because the evaluation of cf was wrapped in a try catch and we would simply ignore the faulty formulas.
Unfortunately, following
https://github.com/odoo/o-spreadsheet/pull/3380, the try/catch was removed and the faulty CF would throw an error during its evaluation.

This commit reinstates the try-catch in case another part of the evaluation could throw in the future so that the end user would not see their spreadsheet chrash entirely and become unable to use it themselves.

It also adds a new condition to the CF allow dispatch to capture invalid formula in order to warn the user during the CF edition in the same way we already handle invalid formulas for colorScale and Iconset rules.

Task: 4036921

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/tests/conditional_formatting/conditional_formatting_plugin.test.ts
+++ b/tests/conditional_formatting/conditional_formatting_plugin.test.ts
@@ -505,7 +505,7 @@ describe("conditional format", () => {
     });
   });
 
-  test.skip("multiple conditional formats using stopIfTrue flag", () => {
+  test("multiple conditional formats using stopIfTrue flag", () => {
     setCellContent(model, "A1", "2");
 
     model.dispatch("ADD_CONDITIONAL_FORMAT", {
@@ -675,7 +675,7 @@ describe("conditional format", () => {
     });
   });
 
-  test("cannot send invalid arguments to conditional format rules command", () => {
+  test("cannot send invalid arguments to *move* conditional format rules command", () => {
     model.dispatch("ADD_CONDITIONAL_FORMAT", {
       cf: createEqualCF("1", { fillColor: "#FF0000" }, "idRule1"),
       ranges: toRangesData(sheetId, "A1"),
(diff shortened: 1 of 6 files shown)


tests/conditional_formatting/conditional_formatting_plugin.test.ts
```diff
@@ -675,7 +675,7 @@ describe("conditional format", () => {
     });
   });
 
-  test("cannot send invalid arguments to conditional format rules command", () => {
+  test("cannot send invalid arguments to *move* conditional format rules command", () => {
```
hokolomopo, line 678: \*move\* ?

src/plugins/ui_core_views/evaluation_conditional_format.ts
```diff
@@ -91,51 +91,55 @@ export class EvaluationConditionalFormatPlugin extends UIPlugin {
...
-                      sheetId,
-                      col - zone.left,
-                      row - zone.top,
-                      {
-                        ...compiledFormula,
-                        dependencies: compiledFormula.dependencies.map((d) =>
-                          this.getters.getRangeFromSheetXC(sheetId, d)
-                        ),
-                      }
+      try {
```
hokolomopo, line 94: What exactly throws  in this method ? The `compile` ? In that case we have the same problem with data validation. 

The actual evaluation won't throw since it's done with `getters.evaluateFormula` that have a try/catch

hokolomopo (commented):
pouet pouet

tests/conditional_formatting/conditional_formatting_plugin.test.ts
rrahir: erf, priority was the word I was looking for

src/plugins/ui_core_views/evaluation_conditional_format.ts
rrahir: It is indeed broken for the data validations as well :) I think we could do the same thing I did for this: 1 fix PR and 1 imp to capture these errors

hokolomopo:
Following lul PR https://github.com/odoo/o-spreadsheet/pull/4652 , this PR is obsolete no ?

LucasLefevre (commented):
nice little feedback for the user :)
:+1:

LucasLefevre (approved):
robodoo r+

PR #4625, opened 2024-07-08
[IMP] charts: make all series editable at once

anhe-odoo:
## Task Description

This commit aims to let the user edit some parameters of the data series once for all series (like the vertical axis) to avoid having to set it one by one.

## Related Task

- Task: 3972623

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/chart/building_blocks/general_design/general_design_editor.xml
+++ b/src/components/side_panel/chart/building_blocks/general_design/general_design_editor.xml
@@ -3,11 +3,14 @@
     <t t-set="chart_title">Chart title</t>
     <SidePanelCollapsible collapsedAtInit="false" title.translate="General">
       <t t-set-slot="content">
-        <Section class="'o-chart-background-color pt-0 pb-0'" title.translate="Background color">
-          <RoundColorPicker
-            currentColor="props.definition.background"
-            onColorPicked.bind="updateBackgroundColor"
-          />
+        <Section class="'o-chart-background-color pt-0 pb-0'">
+          <div class="d-flex align-items-center">
+            <span class="o-section-title mb-0 pe-2">Background color</span>
+            <RoundColorPicker
+              currentColor="props.definition.background"
+              onColorPicked.bind="updateBackgroundColor"
+            />
+          </div>
         </Section>
         <ChartTitle
           title="title.text"
--- a/src/components/side_panel/chart/building_blocks/generic_side_panel/config_panel.ts
+++ b/src/components/side_panel/chart/building_blocks/generic_side_panel/config_panel.ts
@@ -113,7 +113,16 @@ export class GenericChartConfigPanel extends Component<Props, SpreadsheetChildEn
    * button "confirm" is clicked
    */
   onDataSeriesRangesChanged(ranges: string[]) {
+    let yAxisId: string | undefined = undefined;
+    if (this.dataSeriesRanges.length) {
+      if (this.dataSeriesRanges.every((ds) => ds.yAxisId === "y")) {
+        yAxisId = "y";
+      } else if (this.dataSeriesRanges.every((ds) => ds.yAxisId === "y1")) {
+        yAxisId = "y1";
+      }
+    }
     this.dataSeriesRanges = ranges.map((dataRange, i) => ({
+      yAxisId,
       ...this.dataSeriesRanges?.[i],
       dataRange,
     }));
--- a/src/components/side_panel/chart/building_blocks/series_design/series_design_editor.ts
+++ b/src/components/side_panel/chart/building_blocks/series_design/series_design_editor.ts
@@ -1,11 +1,13 @@
 import { Component, useState } from "@odoo/owl";
 import { getColorsPalette, getNthColor, toHex } from "../../../../../helpers";
+import { _t } from "../../../../../translation";
 import {
   ChartWithDataSetDefinition,
   DispatchResult,
   SpreadsheetChildEnv,
   UID,
 } from "../../../../../types";
+import { ChartTerms } from "../../../../translations_terms";
 import { SidePanelCollapsible } from "../../../components/collapsible/side_panel_collapsible";
 import { RoundColorPicker } from "../../../components/round_color_picker/round_color_picker";
 import { Section } from "../../../components/section/section";
@@ -17,6 +19,7 @@ interface Props {
     figureID: UID,
     definition: Partial<ChartWithDataSetDefinition>
   ) => DispatchResult;
+  canEditAllSeries?: boolean;
   updateChart: (figureId: UID, definition: Partial<ChartWithDataSetDefinition>) => DispatchResult;
 }
 
@@ -33,20 +36,30 @@ export class SeriesDesignEditor extends Component<Props, SpreadsheetChildEnv> {
     updateChart: Function,
     canUpdateChart: Function,
     slots: { type: Object, optional: true },
+    canEditAllSeries: { type: Boolean, optional: true },
   };
 
-  protected state = useState({ index: 0 });
+  protected state: { index: "all" | number } = useState({
+    index: this.props.canEditAllSeries ? "all" : 0,
+  });
 
   getDataSeries() {
-    const runtime = this.env.model.getters.getChartRuntime(this.props.figureId);
-    if (!runtime || !("chartJsConfig" in runtime)) {
-      return [];
+    const dataSeries = this.props.definition.dataSets.map(
+      (d, i) => d.label ?? `${ChartTerms.Series} ${i + 1}`
+    );
+    if (!this.props.canEditAllSeries) {
+      return dataSeries;
     }
-    return runtime.chartJsConfig.data.datasets.map((d) => d.label);
+    return [_t("All series"), ...dataSeries];
   }
 
   updateSerieEditor(ev) {
-    this.state.index = ev.target.selectedIndex;
+    const selectedIndex = ev.target.selectedIndex;
+    if (this.props.canEditAllSeries) {
+      this.state.index = selectedIndex === 0 ? "all" : selectedIndex - 1;
+      return;
+    }
+    this.state.index = selectedIndex;
   }
 
   updateDataSeriesColor(color: string) {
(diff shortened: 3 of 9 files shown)


src/components/side_panel/chart/building_blocks/general_series/general_series_editor.ts
```diff
@@ -0,0 +1,123 @@
...
+} from "../../../../../types";
+import { SidePanelCollapsible } from "../../../components/collapsible/side_panel_collapsible";
+import { RoundColorPicker } from "../../../components/round_color_picker/round_color_picker";
+import { Section } from "../../../components/section/section";
+
+interface Props {
+  figureId: UID;
+  definition: ChartWithAxisDefinition;
+  updateChart: (figureId: UID, definition: Partial<ChartWithAxisDefinition>) => DispatchResult;
+  canChangeVerticalAxis?: boolean;
```
hokolomopo, line 19: Is this props useful ? The component is only used once where this props is "true"

src/components/side_panel/chart/building_blocks/general_series/general_series_editor.ts
```diff
@@ -0,0 +1,123 @@
...
+}
+
+export class GeneralSeriesEditor extends Component<Props, SpreadsheetChildEnv> {
```
hokolomopo, line 22:
```suggestion
export class ChartWithAxisSerieEditor extends Component<Props, SpreadsheetChildEnv> {
```
maybe ? since it's only used there

src/components/side_panel/chart/building_blocks/general_series/general_series_editor.ts
```diff
@@ -0,0 +1,123 @@
...
+    SidePanelCollapsible,
+    Section,
+    RoundColorPicker,
+  };
+  static props = {
+    figureId: String,
+    definition: Object,
+    updateChart: Function,
+    canChangeVerticalAxis: { type: Boolean, optional: true },
+    slots: { type: Object, optional: true },
```
hokolomopo, line 34: slots are enver used

src/components/side_panel/chart/building_blocks/general_series/general_series_editor.ts
```diff
@@ -0,0 +1,123 @@
...
+  };
+
+  private state = useState({ index: -1 });
+
+  getDataSeries() {
+    const runtime = this.env.model.getters.getChartRuntime(this.props.figureId);
```
hokolomopo, line 40: Do you think it's possible to not use the runtime in this component ? It kinda make the side panel depend on ChartJS, which isn't a problem per se, but not really amazing

src/components/side_panel/chart/building_blocks/general_series/general_series_editor.ts
hokolomopo: Edit: oh that's for radar isn't it ? and they aren'tmerged yet :upside_down_face:

src/components/side_panel/chart/building_blocks/general_series/general_series_editor.ts
```diff
@@ -0,0 +1,123 @@
...
+    const color = dataSets[this.state.index].backgroundColor;
+    return color ? toHex(color) : getNthColor(this.state.index);
+  }
+
+  updateDataSeriesAxis(ev) {
+    const yAxisId = ev.target.value === "left" ? "y" : "y1";
+    const dataSets = this.props.definition.dataSets;
+    const currentIndex = this.state.index;
+    for (let i = 0; i < dataSets.length; i++) {
+      if (currentIndex === -1 || currentIndex === i) {
```
hokolomopo, line 76: My opinion, but reading this loop is a bit confusing IMO. Id go with something like 
```
if(currentIndex === -1) dataset = datasets.map(...) 
else (dataset[currentIndex] = {...}
```

src/components/side_panel/chart/building_blocks/general_series/general_series_editor.ts
```diff
@@ -0,0 +1,123 @@
...
+      const { useLeftAxis, useRightAxis } = getDefinedAxis(this.props.definition);
+      if (useLeftAxis && useRightAxis) {
+        return "";
+      } else if (useLeftAxis) {
+        return "left";
+      }
+      return "right";
+    }
+    if (!dataSets?.[this.state.index]) return "left";
+    return dataSets[this.state.index].yAxisId === "y1" ? "right" : "left";
```
hokolomopo, line 98:
```suggestion
    return dataSets[this.state.index]?.yAxisId === "y1" ? "right" : "left";
```

src/components/side_panel/chart/building_blocks/general_series/general_series_editor.ts
```diff
@@ -0,0 +1,123 @@
...
+  };
+
+  private state = useState({ index: -1 });
+
+  getDataSeries() {
+    const runtime = this.env.model.getters.getChartRuntime(this.props.figureId);
+    if (!runtime || !("chartJsConfig" in runtime)) {
+      return [];
+    }
+    return [_t("All series"), ...runtime.chartJsConfig.data.datasets.map((d) => d.label)];
```
hokolomopo, line 44: I think there shouldn't be an "All series" category if there is nothing editable for all series

src/components/side_panel/chart/building_blocks/generic_side_panel/config_panel.ts
```diff
@@ -104,7 +104,26 @@ export class GenericChartConfigPanel extends Component<Props, SpreadsheetChildEn
...
+        useLeftAxis = true;
+      } else {
+        undefinedAxis = true;
+      }
+    }
+    let yAxisId: string | undefined = undefined;
+    if (useLeftAxis && !useRightAxis && !undefinedAxis) {
+      yAxisId = "y";
+    } else if (!useLeftAxis && useRightAxis && !undefinedAxis) {
+      yAxisId = "y1";
```
hokolomopo, line 123:
```suggestion
    let yAxisId: string | undefined = undefined;
    if (this.dataSeriesRanges.every(/* hasLeftAxis */) {
      yAxisId = "y";
    } else if (this.dataSeriesRanges.every(/* hasRightAxis */)) {
      yAxisId = "y1";
```
And the above for loop can be removed ? Unless I understand the function worng, it's kinda hard to read ^^'

hokolomopo (commented):
:wave: 

I think i'll red it until the radar charts are merged, it looks a bit strange in standalone

src/components/side_panel/chart/building_blocks/series_design/series_design_editor.ts
```diff
@@ -33,20 +36,27 @@ export class SeriesDesignEditor extends Component<Props, SpreadsheetChildEnv> {
...
     slots: { type: Object, optional: true },
+    canChooseAxis: { type: Boolean, optional: true },
   };
 
-  protected state = useState({ index: 0 });
+  get seriesOffset() {
+    return this.props.canChooseAxis ? 1 : 0;
+  }
+
+  protected state = useState({ index: -this.seriesOffset });
```
hokolomopo, line 46: How about getting rid of `seriesOffset` for some easy to understand boolean `cadEditAllSeries` and change `state.index: number` to `state.selectedSeries: "all" | number` ? 

I feel like index kinda sucks as a variable name (even more so if it is referenced outside the component) and having index === -1 being "all series selected" isn't helping the ease of understanding of this code IMO

tests/figures/chart/charts_component.test.ts
```diff
@@ -949,6 +980,42 @@ describe("charts", () => {
     ]);
   });
 
+  test("Vertical axis is take into account when adding new range", async () => {
```
hokolomopo, line 983:
```suggestion
  test("New range added to the chart will have the same axisId as existing ranges", async () => {
```

tests/figures/chart/charts_component.test.ts
```diff
@@ -1669,7 +1737,11 @@ describe("charts", () => {
           model,
           {
             dataSets: [
-              { dataRange: "B1:B4", trend: { type: "polynomial", order: 3, display: true } },
+              {
+                dataRange: "B1:B4",
+                label: "serie_1",
```
hokolomopo, line 1742: nitpick: there should be a default value that can be used in the `await setInputValueAndTrigger(".data-series-selector" ...)` without having to manually set a label to each dataset at each test no ? That'd remove half the changes you had to make in the tests

hokolomopo (commented):
pouet pouet

src/components/side_panel/chart/building_blocks/general_design/general_design_editor.xml
```diff
@@ -3,11 +3,14 @@
...
-            onColorPicked.bind="updateBackgroundColor"
-          />
+        <Section class="'o-chart-background-color pt-0 pb-0'">
+          <div class="d-flex align-items-center">
+            <span class="o-section-title mb-0 pe-2">Background color</span>
+            <RoundColorPicker
+              currentColor="props.definition.background"
+              onColorPicked.bind="updateBackgroundColor"
+            />
+          </div>
```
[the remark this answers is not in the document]
LucasLefevre: If we do it, we it should be in a separate commit (if not a different task) since it has nothing to do with the "all series editable" feature.

And I'd agree with Adrien. It'd keep it as the designers proposed

src/components/side_panel/chart/building_blocks/series_design/series_design_editor.ts
LucasLefevre: I agree. index -1 doesn't mean anything when reading the code. Let's make it more explicit

src/components/side_panel/chart/building_blocks/series_design/series_with_axis_design_editor.xml
```diff
@@ -1,6 +1,6 @@
 <templates>
   <t t-name="o-spreadsheet-SeriesWithAxisDesignEditor">
-    <SeriesDesignEditor t-props="props">
+    <SeriesDesignEditor t-props="props" canChooseAxis="!props.definition.horizontal">
```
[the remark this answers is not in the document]
anhe-odoo: Yes, but this props is explicitly for the axis (as maybe one day we will have some other global properties that can be edited once for all). What about something like a general `options` or `config` object that can contains a `canChooseAxis` boolean ?

src/components/side_panel/chart/building_blocks/general_design/general_design_editor.xml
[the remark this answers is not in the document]
LucasLefevre: Up, revert this :wink:

PR #4626, opened 2024-07-08
[IMP] locale: add start week

pro-odoo:
This commit adds the start week to the locale. It will be useful to introduce the granularity `day_of_week` for the pivots.

Task: 4031965

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


tests/model/model_import_export.test.ts
```diff
@@ -766,6 +772,21 @@ test("import localized date as string and detect the format", () => {
   expect(getEvaluatedCell(model, "A1")?.formattedValue).toBe("31/12/2020");
 });
 
+test("Week start is automatically added during migration", () => {
+  expect(
+    new Model({
+      version: 17,
```
rrahir, line 778: some reason to target 17 and not 19 ?

src/migrations/locale.ts
```diff
@@ -0,0 +1,94 @@
+export const WEEK_START = {
```
rrahir, line 1: Could add a FIXME to remove this map once FF starts supporting https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/getWeekInfo

rrahir (commented):
WDYT of taking the same strategy as the currencies and have a dedicated object that would be implemented by the  integrator instead of having to support everything in the library directly?

pro-odoo:
> WDYT of taking the same strategy as the currencies and have a dedicated object that would be implemented by the integrator instead of having to support everything in the library directly?

Not sure to understand what you mean...

tests/model/model_import_export.test.ts
pro-odoo: When I started this task, we were on version 17 :D

doc/integrating/integration.md
```diff
@@ -139,6 +139,7 @@ const locale = {
     code: "en_US",
     thousandsSeparator: ",",
     decimalSeparator: ".",
+    weekStart: 7,
```
rrahir, line 142: the 1-7 mapping could be explained in the doc as well

rrahir: Also maybe add that information in the Localisation sidepanel?

pro-odoo: Updated :)

rrahir:
@robodoo r+

PR #3893, opened 2024-03-24
[IMP] charts: improve design configuration

anhe-odoo:
## Description

This task aims to improve the design customization of the charts in o-spreadsheets, adding many customizable options like the title alignment/color/style, the axes configurations, the datasets colors, ...

## Related Task(s)

- Task [3792009](https://www.odoo.com/web#id=3792009&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## Review Checklist

- [ ] feature is organized in plugin, or UI components
- [ ] support of duplicate sheet (deep copy)
- [ ] in model/core: ranges are Range object, and can be adapted (adaptRanges)
- [ ] in model/UI: ranges are strings (to show the user)
- [ ] undo-able commands (uses this.history.update)
- [ ] multiuser-able commands (has inverse commands and transformations where needed)
- [ ] new/updated/removed commands are documented
- [ ] exportable in excel
- [ ] translations (\_t("qmsdf %s", abc))
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes
- [ ] doc is rebuild (npm run doc)
- [ ] status is correct in Odoo

PR #4627, opened 2024-07-08
[FIX] data: update version to 17

anhe-odoo:
## Description

Since https://github.com/odoo/o-spreadsheet/pull/3893 the current data version is 17, but the CURRENT_VERSION is still set to 16, as well as the EXCEL_IMPORT_VERSION, meaning the chart migration is never applied. This PR aims to fix this behavior.

- Related Task: N./A.

### DIFF
--- a/demo/data.js
+++ b/demo/data.js
@@ -6,7 +6,7 @@
  */
 
 export const demoData = {
-  version: 16,
+  version: 17,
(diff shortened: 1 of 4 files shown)


pro-odoo:
robodoo rebase-merge r+

PR #4649, opened 2024-07-11
[IMP] Tests: drop ts-jest in favor of swc-node/jest 

rrahir:
## Description:

SWC (speedy web compiler) is an rust-based faster alternative to the
traditional typescript transpiler `tsc`. SInce it does not support
typechecking, we cannot only rely on it in order to generate our
declaration files our build our library. We can however leverage the
already existing integration with jest, expecially the package
`@swc-node/jest` that works as an alternative to  `js-test`.

In order to preserve the typechecking of the test files, we alter the
`test` command to first run a simple typecheck with `tsc` and then run
the test suite with `@swc-node/jest` as a TS transpiler.

Some pros:
- Better execution time for the whole test suite
- running a standalone test way faster
- a pre-step of typecheck takes less time than currently beacause jest
  will keep running every test file and will not stop its execution
  because of a single file in error

Benchmark
---------

Running command `npm run test -- -w 50%` -  11178 tests:

Before:  131 secondes
After:   82 secondes (17 of typecheck + 68 of actual tests transpiled
and run

On Runbot

Before: 250 secondes
After:  122 secondes

Task: : [4049142](https://www.odoo.com/web#id=4049142&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,6 +18,7 @@
         "@prettier/plugin-xml": "^2.2.0",
         "@rollup/plugin-node-resolve": "^15.2.0",
         "@rollup/plugin-terser": "^0.4.3",
+        "@swc/jest": "^0.2.36",
         "@types/jest": "^27.0.1",
         "@types/node": "^13.13.23",
         "@types/rbush": "^3.0.3",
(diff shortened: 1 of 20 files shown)


LucasLefevre (approved):
robodoo r+
:heart:

PR #4668, opened 2024-07-16
[IMP] pivots: table type synced with pivot structure

rrahir:
If a user had linked a dynamic table to a pivot formula (`=PIVOT(1)`) the table would not be adapted accordingly when the user converted their pivot into "splitted" individual formulas. This revision detects if a dynamic table is attached the anchor of the dynamic pivot formula and converts it along the pivot such that it matches the the pivot range.

Task: 4039655

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_feature/insert_pivot.ts
+++ b/src/plugins/ui_feature/insert_pivot.ts
@@ -1,6 +1,6 @@
 import { PIVOT_TABLE_CONFIG } from "../../constants";
 import { SpreadsheetPivotTable } from "../../helpers/pivot/table_spreadsheet_pivot";
-import { getZoneArea } from "../../helpers/zones";
+import { getZoneArea, positionToZone } from "../../helpers/zones";
 import { _t } from "../../translation";
 import { HeaderIndex, PivotTableData, UID } from "../../types";
 import { Command } from "../../types/commands";
@@ -20,6 +20,8 @@ export class InsertPivotPlugin extends UIPlugin {
       case "INSERT_PIVOT_WITH_TABLE":
         this.insertPivotWithTable(cmd.sheetId, cmd.col, cmd.row, cmd.pivotId, cmd.table);
         break;
+      case "SPLIT_PIVOT_FORMULA":
+        this.splitPivotFormula(cmd.sheetId, cmd.col, cmd.row, cmd.pivotId, cmd.table);
     }
   }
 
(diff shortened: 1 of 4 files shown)


src/plugins/ui_feature/insert_pivot.ts
```diff
@@ -166,4 +168,41 @@ export class InsertPivotPlugin extends UIPlugin {
...
+      pivotId,
+      table: pivotTableData,
+    });
+    const table = this.getters.getCoreTable({ sheetId, col, row });
+    if (table?.type === "dynamic") {
+      const zone = positionToZone({ col, row });
+      const { cols, rows, measures, fieldsType } = pivotTableData;
+      const pivotTable = new SpreadsheetPivotTable(cols, rows, measures, fieldsType || {});
+      const colNumber = pivotTable.getNumberOfDataColumns() + 1;
+      const rowNumber = pivotTable.columns.length + pivotTable.rows.length;
```
anhe-odoo, line 192: Why do you add row.length + columns.length to compute the rowNumber ?

Also, nitpick, but adding +1 to colNumber to use colnumber-1 after is maybe useless ?

pro-odoo: > Why do you add row.length + columns.length to compute the rowNumber

Sadly, it's not crystal clear how the pivot works :/
Imagine the following pivot:
|  | Alice |  | Bob |  |  |
|---|---|---|---|---|---|
|  | TRUE | FALSE | TRUE | FALSE | Total |
|  | Expected Revenue | Expected Revenue | Expected Revenue | Expected Revenue | Expected Revenue |
| January | 10 | 20 | 30 | 40 | 100 |
| February | 1 | 2 | 3 | 4 | 10 |
| Total | 11 | 22 | 33 | 44 | 110 |

`columns` is an array of each row (also an array). So the first item of `columns` if the first row, etc.
So, `columns.length` is the height of the columns header (3 in this case)

`rows` is an array for each row of the body (so columns header not included), so 3 also in this case).

So the `rowNumber` of the table is the total of the columns header and the rows of the body, so 6.

I hope it's clearer with that, but we should improve it to make it more clear :)

LucasLefevre (approved):
robodoo r+

PR #4669, opened 2024-07-17
[PERF] evaluation: don't create closure for each cell

LucasLefevre:
## Description:

Currently, `__originCellXc` is a closure created for each evaluated cell.
It needs:
1) to be instanciated
2) to be garbage collected later.

It takes CPU time.

It's currently a closure to lazily build the XC (which is even costy performance-wise)

This commit changes `__originCellXc` to be the cell position object. The object is already created. So it doesn't require ab extra allocation (and GC)

`evaluateAllCells` on the large formula data set allocated 132MB before, and only 88MB with this commit.

The timing performance improvement is difficult to measure because there's a lot of variation, and GC is not deterministic.

It also improves the performance of all formulas using `__originCellXc` because, they no longer needs to parse the XC string.

Task: : [4057810](https://www.odoo.com/web#id=4057810&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/functions/helper_lookup.ts
+++ b/src/functions/helper_lookup.ts
@@ -1,6 +1,6 @@
-import { toZone, zoneToXc } from "../helpers";
+import { zoneToXc } from "../helpers";
 import { _t } from "../translation";
-import { CellPosition, EvalContext, FunctionResultObject, Getters, Maybe, UID } from "../types";
+import { EvalContext, FunctionResultObject, Getters, Maybe, UID } from "../types";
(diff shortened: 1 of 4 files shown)


laa-odoo (approved):

laa-odoo:
@robodoo r+

PR #4671, opened 2024-07-17
[REF] pivot: introduce id on measures

pro-odoo:
## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/pivot/pivot_helpers.ts
+++ b/src/helpers/pivot/pivot_helpers.ts
@@ -128,11 +128,11 @@ const DATE_FIELDS = ["date", "datetime"];
  * e.g "create_date:month" => { name: "create_date", granularity: "month" }
  */
 export function parseDimension(dimension: string): PivotCoreDimension {
-  const [name, granularity] = dimension.split(":");
+  const [fieldName, granularity] = dimension.split(":");
   if (granularity) {
-    return { name, granularity };
+    return { fieldName, granularity };
   }
-  return { name };
+  return { fieldName };
 }
 
 export function isDateField(field: PivotField) {
--- a/src/helpers/pivot/pivot_runtime_definition.ts
+++ b/src/helpers/pivot/pivot_runtime_definition.ts
@@ -36,33 +36,36 @@ export class PivotRuntimeDefinition {
     return dimension;
   }
 
-  getMeasure(name: string): PivotMeasure {
-    const measure = this.measures.find((measure) => measure.name === name);
+  getMeasure(id: string): PivotMeasure {
+    const measure = this.measures.find((measure) => measure.id === id);
     if (!measure) {
-      throw new EvaluationError(_t("Field %s is not a measure", name));
+      throw new EvaluationError(_t("Field %s is not a measure", id));
     }
     return measure;
   }
 }
 
 function createMeasure(fields: PivotFields, measure: PivotCoreMeasure): PivotMeasure {
-  const name = measure.name;
+  const fieldName = measure.fieldName;
   const field =
-    name === "__count"
+    fieldName === "__count"
       ? { name: "__count", string: _t("Count"), type: "integer", aggregator: "sum" }
-      : fields[name];
+      : fields[fieldName];
   const aggregator = measure.aggregator;
   return {
-    nameWithAggregator: name + (aggregator ? `:${aggregator}` : ""),
+    /**
+     * Get the id of the measure, as it is stored in the pivot formula
+     */
+    id: measure.id,
     /**
      * Display name of the measure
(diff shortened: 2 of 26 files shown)


src/migrations/migration_steps.ts
```diff
@@ -0,0 +1,358 @@
...
+import { getItemId } from "../helpers";
+import { toXC } from "../helpers/coordinates";
+import { getMaxObjectId } from "../helpers/pivot/pivot_helpers";
+import { overlap, toZone, zoneToXc } from "../helpers/zones";
+import { Registry } from "../registries/registry";
+import { CustomizedDataSet, DEFAULT_LOCALE, Format, Zone } from "../types";
+import { normalizeV9 } from "./legacy_tools";
+
+interface MigrationStep {
+  version: string;
```
VincentSchippefilt, line 15: just to be clearer, shouldn't it be either "fromVersion" or "toVersion" ?

src/helpers/pivot/pivot_helpers.ts
```diff
@@ -128,11 +128,11 @@ const DATE_FIELDS = ["date", "datetime"];
  * e.g "create_date:month" => { name: "create_date", granularity: "month" }
  */
 export function parseDimension(dimension: string): PivotCoreDimension {
```
LucasLefevre, line 130: Shouldn't `parseDimension` completely removed ?

Some leftover uses that I think are wrong:

- It is still used in the pivot layout configurator. It should be based on the id instead of the `nameWithranularity` (see `draggableIds`)
- in `parseArgsToPivotDomain` (and `getTypeOfDimension`): `fieldWithGranularity` is coming from the formula and is actually an id.
- `SpreadsheetPivotTable` manipulates `fieldWithGranularity`, it should be id (it needs to get the actual measure from the id but it doesn't have access to the definition atm)

src/helpers/pivot/pivot_runtime_definition.ts
```diff
@@ -36,33 +36,37 @@ export class PivotRuntimeDefinition {
...
-      : fields[name];
+      : fields[fieldName];
   const aggregator = measure.aggregator;
   return {
-    nameWithAggregator: name + (aggregator ? `:${aggregator}` : ""),
+    /**
+     * Get the id of the measure, as it is stored in the pivot formula
+     */
+    id: measure.id,
+    nameWithAggregator: fieldName + (aggregator ? `:${aggregator}` : ""),
```
LucasLefevre, line 60: `nameWithAggregator` can be removed. It's only used for a `t-key` and it'll be wrong as soon as we have duplicates measure + aggregator

src/types/pivot.ts
```diff
@@ -25,13 +25,14 @@ export type Granularity =
...
 export interface PivotCoreDimension {
-  name: string;
+  fieldName: string;
   order?: "asc" | "desc";
   granularity?: Granularity | string;
 }
 
 export interface PivotCoreMeasure {
-  name: string;
+  id: string; // Identifier of the measure, `technicalName:aggregator{:autoIncrementedNumber}`.
```
LucasLefevre, line 34: can be worth adding to the comment that it's used to identify measures in formulas

src/registries/auto_completes/pivot_auto_complete.ts
```diff
@@ -70,7 +70,7 @@ autoCompleteProviders.add("pivot_measures", {
 
     return definition.measures
       .map((measure) => {
-        if (measure.name === "__count") {
+        if (measure.fieldName === "__count") {
```
LucasLefevre, line 73: the measure autocomplete should auto-complete the id, but it autocompletes the field name

src/helpers/pivot/pivot_helpers.ts
pro-odoo: > It is still used in the pivot layout configurator. It should be based on the id instead of the nameWithranularity (see draggableIds)

No, col and row does not have `id` (they should, but in a future task ?)

> in parseArgsToPivotDomain (and getTypeOfDimension): fieldWithGranularity is coming from the formula and is actually an id.

It's for col and row, so it's correct

> SpreadsheetPivotTable manipulates fieldWithGranularity, it should be id (it needs to get the actual measure from the id but it doesn't have access to the definition atm)

For measure, it's useless to get the measure, the value is always a string (the field name of the measure). So no need to update `SpreadsheetPivotTable` :)

LucasLefevre (approved):
robodoo rebase-ff

LucasLefevre:
robodoo r-

LucasLefevre (commented):
robodoo r+

PR #4685, opened 2024-07-19
[FIX] pivot: support empty values

pro-odoo:
Steps to reproduce:
- Create a spreadsheet pivot, with some empty values in the data so that we have an (Undefined) header cell
- Copy/paste a cell child of the (undefined) header
- Freeze the pivot

=> The cells corresponding to empty values are empty.

Implementation note:
The pivot helper `toNormalizedPivotValue` takes `groupValue` as argument, but the type of `groupValue` is variable. It's either a CellValue or a FunctionResultObject. This is not ideal and should be fixed in the future.

Task: 4053215

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/pivot/pivot_helpers.ts
+++ b/src/helpers/pivot/pivot_helpers.ts
@@ -209,6 +209,9 @@ export function toNormalizedPivotValue(
     typeof groupValue === "boolean"
       ? toString(groupValue).toLocaleLowerCase()
       : toString(groupValue);
+  if (groupValueString === "null") {
+    return null;
+  }
   if (!pivotNormalizationValueRegistry.contains(dimension.type)) {
     throw new EvaluationError(
       _t("Field %(field)s is not supported because of its type (%(type)s)", {
@@ -236,6 +239,9 @@ export function toFunctionPivotValue(
   value: CellValue,
   dimension: Pick<PivotDimension, "type" | "granularity">
 ) {
+  if (value === null) {
+    return `"null"`;
+  }
   if (!pivotToFunctionValueRegistry.contains(dimension.type)) {
     return `"${value}"`;
   }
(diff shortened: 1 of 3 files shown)


src/helpers/pivot/pivot_helpers.ts
```diff
@@ -209,6 +209,9 @@ export function toNormalizedPivotValue(
     typeof groupValue === "boolean"
       ? toString(groupValue).toLocaleLowerCase()
       : toString(groupValue);
+  if (groupValueString === "null") {
```
hokolomopo, line 212: That should never happen because of the `if` above, no?

src/helpers/pivot/pivot_helpers.ts
```diff
@@ -236,6 +239,9 @@ export function toFunctionPivotValue(
   value: CellValue,
   dimension: Pick<PivotDimension, "type" | "granularity">
 ) {
+  if (value === null) {
+    return `"null"`;
```
hokolomopo, line 243: We already have the issue, and this seems like a very edge case, but the pivots are somewhat broken if there is cells with the string value `null` in the dataset

hokolomopo (commented):
@robodoo r+

src/helpers/pivot/pivot_helpers.ts
pro-odoo: It can be a FunctionResultObject :/

src/helpers/pivot/pivot_helpers.ts
pro-odoo: Task to fix in master: https://www.odoo.com/odoo/project/2328/tasks/4063595?cids=1

PR #4696, opened 2024-07-22
[IMP] chart: uniformize color with Odoo Charts

anhe-odoo:
## Description

This PR changes the `ColorGenerator` classes to match the colors used in Odoo Chart View.

## Related Task

- Task: 4028658


src/helpers/color.ts
```diff
@@ -367,21 +367,132 @@ const ColorsList = [
...
+  "#FFBC2C", // Yellow #2
+  "#C08A16", // Yellow #3
+];
+const COLORS_XL = [
+  "#4EA7F2", // Blue #1
+  "#3188E6", // Blue #2
+  "#056BD9", // Blue #3
+  "#155193", // Blue #4
+  "#A76DBC", // Violet #1
+  "#7F4295", // Violet #1
```
hokolomopo, line 424: I see you copied odoo/web, but there's probably a smarter way to do that that doesn't involve repeating 4 times Blue #1, 3 times Blue #2, 2 times Blue #3 ...

src/components/selection_input/selection_input_store.ts
```diff
@@ -201,7 +207,7 @@ export class SelectionInputStore extends SpreadsheetStore {
    * e.g. ["A1", "Sheet2!B3", "E12"]
    */
   get selectionInputs(): (RangeInputValue & { isFocused: boolean; isValidRange: boolean })[] {
-    const generator = new ColorGenerator(this.colors);
+    const generator = new ColorGenerator(this.colors, getPaletteName(this.ranges.length));
```
hokolomopo, line 210: All the palette logic could be hidden in the ColorGenerator maybe ? With an argument colors + argument dataset length

src/helpers/color.ts
```diff
@@ -367,21 +367,132 @@ const ColorsList = [
...
+      return COLORS_XL;
+  }
+}
+
+export function getNthColor(index: number, palette: string): Color {
+  const colors = getColors(palette);
+  return colors[index % ColorsList.length];
+}
+
+export function getPaletteName(quantity: number) {
```
hokolomopo, line 476: Do you have to go through a palette name ? Is returning the palette directly possible ?

src/helpers/figures/charts/pie_chart.ts
```diff
@@ -322,7 +322,11 @@ export function createPieChartRuntime(chart: PieChart, getters: Getters): PieCha
   const dataSetFormat = getChartDatasetFormat(getters, chart.dataSets);
   const locale = getters.getLocale();
   const config = getPieConfiguration(chart, labels, { format: dataSetFormat, locale });
-  const backgroundColor = getPieColors(new ColorGenerator(), dataSetsValues);
+  const dataSetsLength = dataSetsValues[0]?.data?.length ?? 0;
```
hokolomopo, line 325: What happens if the datasetes have different sizes ?

src/helpers/figures/charts/waterfall_chart.ts
```diff
@@ -242,14 +242,25 @@ function getWaterfallConfiguration(
...
+            fillStyle: negativeColor,
+            strokeStyle: negativeColor,
+          },
         ];
         if (chart.showSubTotals || chart.firstValueAsSubtotal) {
           legendValues.push({
             text: _t("Subtotals"),
             fontColor,
             fillStyle: subTotalColor,
+            strokeStyle: subTotalColor,
```
hokolomopo, line 263: Do these changes have something to do with the new colors ? Or was somehting broken before ?

hokolomopo (commented):
Pretty colors ! 😍

src/helpers/figures/charts/waterfall_chart.ts
anhe-odoo: image
This has beed added to avoid the gray border here

src/helpers/color.ts
anhe-odoo: It's complicated to find a way (apart defined the colors somewhere and using the constant), because the order of the colors change with the palette (SM = BRTOPY, MD = BTROPY, LG = XL = BVRTOPGY) so we can't rely on some rule like SM[i] = XL[4*i]. But we can have the XL array defined, and then defined the others according to this one, or having some constant like BLUE_1 and then building the array with these constant. What do you think ?

src/components/selection_input/selection_input_store.ts
hokolomopo: DO we ever want to do that ATM ?IF not, I think it's worth to simplify the use of the colorGenerator. We could always change it later if needed

src/helpers/color.ts
```diff
@@ -343,45 +343,113 @@ export function darkenColor(color: Color, percentage: number): Color {
...
 }
 
 export class ColorGenerator {
   private currentColorIndex = 0;
-  private colors: string[];
 
-  constructor(colors: string[] = []) {
-    this.colors = colors;
-  }
+  constructor(private colors: string[] = [], private palette: Color[] = []) {}
```
rrahir, line 448: we could maybe consider collisions between colors and palette?

src/helpers/color.ts
```diff
@@ -343,45 +343,116 @@ export function darkenColor(color: Color, percentage: number): Color {
...
+  "#936A12", // Yellow #4
 ];
 
-export function getNthColor(index: number): Color {
-  return ColorsList[index % ColorsList.length];
+export function getNthColor(index: number, palette: Color[]): Color {
+  return palette[index % palette.length];
+}
+
+export function getPalette(quantity: number) {
```
LucasLefevre, line 433:
```suggestion
export function getColorPalette(quantity: number) {
```

src/helpers/color.ts
```diff
@@ -343,45 +343,116 @@ export function darkenColor(color: Color, percentage: number): Color {
...
 
 export class ColorGenerator {
   private currentColorIndex = 0;
-  private colors: string[];
+  private palette: Color[];
 
-  constructor(colors: string[] = []) {
-    this.colors = colors;
+  constructor(private colors: (string | undefined)[] = [], _palette: Color[] = []) {
+    this.palette = _palette.filter((c) => !colors.includes(c));
```
LucasLefevre, line 450:
```suggestion
  constructor(palletSize: number, private colors: string[] = []) {
    this.palette = getPallet(palletSize).filter((c) => !colors.includes(c));
```

LucasLefevre (commented):
nice :star_struck:

LucasLefevre (commented):
robodoo r+

LucasLefevre:
robodoo r+

PR #4697, opened 2024-07-22
[FIX] auto_complete: keep dropdown open on exact match for functions

dhrp-odoo:
## Description:

Steps to reproduce:
 - Open a composer and type =SU, SUMIF appears in the autocomplete list.
 - Now type =SUM, the autocomplete closes itself, and you can't see the SUMIF function anymore.

Since PR#3401, the auto-complete dropdown closes when the user inputs an exact match of what is in the auto-complete proposals.

This behavior is not expected for functions, so a new property `keepOpenOnExactMatch` has been added to the auto-complete registry to control this.

By default, `keepOpenOnExactMatch` is set to `false` for all auto-completes. For functions, it is set to `true`, ensuring the dropdown doesn't close when the user inputs an exact match.

Task: 4061068

Task: : [4061068](https://www.odoo.com/odoo/project/2328/tasks/4061068?cids=2)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/registries/auto_completes/auto_complete_registry.ts
+++ b/src/registries/auto_completes/auto_complete_registry.ts
@@ -36,6 +36,7 @@ export interface AutoCompleteProviderDefinition {
   sequence?: number;
   autoSelectFirstProposal?: boolean;
   maxDisplayedProposals?: number;
+  keepOpenOnExactMatch?: boolean; // By default, the auto complete closes when the exact match is found.
   getProposals(
     this: { composer: ComposerStore; getters: Getters },
     tokenAtCursor: EnrichedToken,
--- a/src/components/composer/composer/composer_store.ts
+++ b/src/components/composer/composer/composer_store.ts
@@ -789,8 +789,11 @@ export class ComposerStore extends SpreadsheetStore {
       const exactMatch = proposals?.find((p) => p.text === tokenAtCursor.value);
       // remove tokens that are likely to be other parts of the formula that slipped in the token if it's a string
       const searchTerm = tokenAtCursor.value.replace(/[ ,\(\)]/g, "");
-      if (exactMatch && this._currentContent !== this.initialContent) {
-        // this means the user has chosen a proposal
+      if (
+        !provider.keepOpenOnExactMatch &&
+        exactMatch &&
+        this._currentContent !== this.initialContent
+      ) {
         return;
       }
       if (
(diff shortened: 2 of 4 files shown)


src/registries/auto_completes/auto_complete_registry.ts
```diff
@@ -36,6 +36,7 @@ export interface AutoCompleteProviderDefinition {
   sequence?: number;
   autoSelectFirstProposal?: boolean;
   maxDisplayedProposals?: number;
+  keepOpenOnExactMatch?: boolean; // By default, the auto complete closes when the exact match is found.
```
hokolomopo, line 39: I'm unsure we want to go in this direction, because I don't think that functionnaly we ever want to close the provider if we have an exact match ? I don't think it's something function specific.

For example if you have a pivot with id 1 and id 11, why would typing 1 remove the option 11 ? Same for data validation and other autocompleted stuff.

I don't think the exact match check make much sense. Maybe if we have an exact match AND there is no other matches we could close the automplete. But why bother? 

Looking at the original comment `this means the user has chosen a proposal` maybe the goal was to close the autocomplete when the use clicked on a proposal. In functions, it's done automatically because selecting a proposal also add a parenthesis. We could probably have a similar bheviour for the otehr autocomplete (autocompleting a pivot field would add a comma for example). 

@LucasLefevre what do you think ?

LucasLefevre: I agree with Adrien, it's not specific to functions.

> Maybe if we have an exact match AND there is no other matches we could close the automplete

close it...or display everything ? It looks like it's the behavior of the data validation dropdown list, but we can change the behavior to generalize and keep it simple.

hokolomopo: So @dhrp-odoo I think you can drop this task, I don't think it's a simple fix, we'll need to discuss it in more detail with the PO and experiment a bit to get a behaviour that fit all the use cases

dhrp-odoo: Hello @hokolomopo @LucasLefevre 👋

Sure, I've been experimenting with this since this morning and have identified some edge cases that need to be addressed:

- When typing =SUM, the composer assistant provides some values. However, if our intention is to write =SUMIFS, upon exact match, we receive an odd assistant suggestion. This is likely due to how we handle data validation assistant.
- Typing something unusual like =SUSUBUSBD also yields strange results, may be similar to the issue above.
- When typing =PIVOT.VALUE(1,"Expected Revenue","Stage", "Ne"), we are not getting any suggestions.

LucasLefevre: Nice, thanks!
I added your findings in the task to make sure they are addressed

dhrp-odoo:
Closing this PR for now. We'll address it later, considering all edge cases and discussing the complete behavior with the PO

PR #4708, opened 2024-07-24
[IMP] tests: drop ts-jest in favor of swc/jest

VincentSchippefilt:
SWC (speedy web compiler) is an rust-based faster alternative to the traditional typescript transpiler `tsc`. SInce it does not support typechecking, we cannot only rely on it in order to generate our declaration files our build our library. We can however leverage the already existing integration with jest, expecially the package `@swc/jest` that works as an alternative to  `js-test`.

In order to preserve the typechecking of the test files, we alter the `test` command to first run a simple typecheck with `tsc` and then run the test suite with `@swc/jest` as a TS transpiler.

Some pros:
- Better execution time for the whole test suite
- running a standalone test way faster
- a pre-step of typecheck takes less time than currently beacause jest will keep running every test file and will not stop its execution because of a single file in error

Benchmark
---------

Running command `npm run test -- -w 50%` -  11178 tests:

Before:  131 secondes
After:   82 secondes (17 of typecheck + 68 of actual tests transpiled
and run

On Runbot

closes odoo/o-spreadsheet#4649

Before: 250 secondes
After:  122 secondes
Task: 4049142

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -48,9 +48,9 @@
         "rollup": "^3.28.0",
         "rollup-plugin-dts": "^5.3.1",
         "rollup-plugin-typescript2": "^0.35.0",
-        "typedoc": "0.24.8",
-        "typedoc-plugin-markdown": "3.11.1",
-        "typescript": "^5.1.6",
+        "typedoc": "0.25.12",
+        "typedoc-plugin-markdown": "3.17.1",
+        "typescript": "^5.4.3",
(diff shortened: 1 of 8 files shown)


VincentSchippefilt:
robodoo fw=no

VincentSchippefilt:
robodoo rebase-ff

LucasLefevre (approved):
robodoo r+

PR #4716, opened 2024-07-25
[FIX] header_size: JS sparse array problem

dhrp-odoo:
## Description:

JavaScript methods can behave inconsistently with sparse arrays, often ignoring empty indexes. For example, Array.prototype.filter() skips empty indexes.

This commit resolves the issue by explicitly assigning undefined to all indexes when creating the sheet, ensuring that the insertion and removal of rows and columns work as expected.

Task: : [4074438](https://www.odoo.com/odoo/project/2328/tasks/4074438?cids=2)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/core/header_size.ts
+++ b/src/plugins/core/header_size.ts
@@ -15,7 +15,10 @@ export class HeaderSizePlugin extends CorePlugin<HeaderSizeState> implements Hea
   handle(cmd: Command) {
     switch (cmd.type) {
       case "CREATE_SHEET": {
-        this.history.update("sizes", cmd.sheetId, { COL: [], ROW: [] });
+        this.history.update("sizes", cmd.sheetId, {
+          COL: Array(this.getters.getNumberCols(cmd.sheetId)).fill(undefined),
+          ROW: Array(this.getters.getNumberRows(cmd.sheetId)).fill(undefined),
+        });
         break;
       }
       case "DUPLICATE_SHEET":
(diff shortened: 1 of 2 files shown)


src/plugins/core/header_size.ts
```diff
@@ -15,7 +15,10 @@ export class HeaderSizePlugin extends CorePlugin<HeaderSizeState> implements Hea
   handle(cmd: Command) {
     switch (cmd.type) {
       case "CREATE_SHEET": {
-        this.history.update("sizes", cmd.sheetId, { COL: [], ROW: [] });
+        this.history.update("sizes", cmd.sheetId, {
```
rrahir, line 18: the plugin `HeaderGroupingPlugin` has the stame structure. It just so happens that it uses array.map  where this plugin uses array.filter. I guess it's worth applying the same fix preemptively

rrahir (changes requested):

src/plugins/core/header_size.ts
dhrp-odoo: The cases differ because in HeaderSizePlugin, sizes are stored at specific indexes in ROW and COL arrays, making initialization crucial to avoid sparse array issues.

In HeaderGroupingPlugin, however, we store groups with start and end ranges rather than at specific indexes, so the array's structure isn’t impacted in the same way since we don't store values directly at particular indexes.

I just checked, and applying the same fix broke several tests.

rrahir: Thanks for your analysis and explanation :)

rrahir:
@robodoo r+

PR #4723, opened 2024-07-29
[FIX] composer: layout break when composer assistant is opened

dhrp-odoo:
In FW, a PR#[4721](https://github.com/odoo/o-spreadsheet/pull/4721) was mistakenly approved (r+) because it had no conflicts, but it did not resolve the intended issue in the master branch. This new PR addresses that issue.

Description:

When typing =S in the composer, the composer assistant opens. If there isn't enough space on the right (default 300px width), it breaks the layout.

This PR fixes the issue by adding a width check for the composer assistant. If the width condition is met, the style right: 0px is applied to prevent the layout from breaking.

Task: [4037213](https://www.odoo.com/odoo/project/2328/tasks/4037213?cids=2)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/composer/composer/composer.ts
+++ b/src/components/composer/composer/composer.ts
@@ -202,8 +202,8 @@ export class CellComposer extends Component<CellComposerProps, SpreadsheetChildE
         // render left
         assistantStyle.right = `0px`;
       }
-    } else if (this.props.delimitation) {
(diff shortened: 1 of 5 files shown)


hokolomopo:
@robodoo r+

PR #4728, opened 2024-07-30
[FIX] charts: Add missing "showValue" handler for Pie and waterfall c…

rrahir:
…harts

The typing of chart.js is NOT our friend and it shows again. Expecially the type `ChartOptions` that applies a deepPartial on each of its properties, including the plugin recently introduced: `chartShowValuesPlugin`.

Long story short, if provided with `showValues=true`, the plugin REQUIRES a callback function to format the data value. Unfortunately, the `DeepPartial` completely hides this requirement.

This bug was also missed because we do not directly test chart.js lib in our tests and it will be the subject of a task in the near future. In the meantime, we will strengthen the test case in our main integration (Odoo).

Task: 4081436

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/figures/charts/chart_common_line_scatter.ts
+++ b/src/helpers/figures/charts/chart_common_line_scatter.ts
@@ -7,7 +7,12 @@ import { getChartTimeOptions, timeFormatLuxonCompatible } from "../../chart_date
 import { ColorGenerator, colorToRGBA, rgbaToHex } from "../../color";
 import { formatValue } from "../../format";
 import { deepCopy, findNextDefinedValue } from "../../misc";
-import { chartFontColor, getChartAxisTitleRuntime, getDefinedAxis } from "./chart_common";
+import {
+  chartFontColor,
+  formatTickValue,
+  getChartAxisTitleRuntime,
+  getDefinedAxis,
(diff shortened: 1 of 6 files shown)


src/helpers/figures/charts/chart_common_line_scatter.ts
```diff
@@ -150,20 +155,12 @@ function getLineOrScatterConfiguration(
       title: getChartAxisTitleRuntime(chart.axesDesign?.x),
     },
   };
-  const formatCallback = (value) => {
```
LucasLefevre, line 153: There's still this piece of code duplicated in `combo_chart.ts`. It could be replaced with the new `formatTickValue`. in this PR or in master?

LucasLefevre (approved):
:+1: 
we can clean even further

src/helpers/figures/charts/chart_common_line_scatter.ts
rrahir: done here :)

LucasLefevre (approved):
robodoo r+

PR #4739, opened 2024-07-31
[FIX] chart: remove empty datasets for chart

anhe-odoo:
## Task Description

This task aims to undisplay empty datasets from a chart (when all the values are undefined, not when they are equals to 0). This has been done by ignoring the empty datasets during the runtime creation.

## Related Task
- Task: 4085864

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui/evaluation_chart.ts
+++ b/src/plugins/ui/evaluation_chart.ts
@@ -9,7 +9,7 @@ import {
 import { chartTerms } from "../../components/side_panel/translations_terms";
 import { MAX_CHAR_LABEL } from "../../constants";
 import { ChartColors } from "../../helpers/chart";
-import { recomputeZones, zoneToXc } from "../../helpers/index";
+import { isNumber, recomputeZones, zoneToXc } from "../../helpers/index";
 import { range } from "../../helpers/misc";
 import { Mode } from "../../model";
 import { Cell } from "../../types";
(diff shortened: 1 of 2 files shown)


src/plugins/ui/evaluation_chart.ts
```diff
@@ -231,6 +231,10 @@ export class EvaluationChartPlugin extends UIPlugin {
       }
     }
     for (const [dsIndex, ds] of Object.entries(definition.dataSets)) {
+      const data = ds.dataRange ? this.getData(ds, definition.sheetId) : [];
+      if (!data.filter(isDefined).length) {
```
hokolomopo, line 235:
```suggestion
      if (!data.every((cell) => cell === undefined)) {
```
nitpick. This won't reallocate a new array like `filter`

Also it seems strange to me that it works at all. Don't evaluated cells return empty strings for empty cells ? Why is checking that the value is undefined works then ? Or wasn't that the case in 15.0 ?

hokolomopo (commented):
Juste a small detail/question 🙂

src/plugins/ui/evaluation_chart.ts
anhe-odoo: Maybe it was only the case in 15.0, but here when we have empty cell in the dataset we get the "undefined" value. But I'll check in the foward-port if it still works for the next versions :)

rrahir:
@anhe-odoo could you close it if we agree that it's outdated?

anhe-odoo:
@rrahir I don't remember why we let this PR apart and restart from a fresh master-targeted PR (https://github.com/odoo/o-spreadsheet/pull/4842), but I suppose it was related to the "stable" policy as it's not really a bug but a change in behavior so, yes, we should close it.

PR #4741, opened 2024-08-01
[FIX] pivot: Prevent faulty pivot domain to crash

rrahir:
The getter `getPivotCellFromPosition` parses pivot cells domain args but this step assumes that the domain is valid (e.g. refers to a valid groupby,measure combination). Unfortunately, if the domain is invalid, the parsing will crash.

Similarly to the decision made for the getter `evaluateFormula`[^1] we wrap the parsing in a try/catch statement and in case of faulty evaluation, return the same result as for invalid/non-existing pivots, that is an empty Pivotcell.

[^1]: See https://github.com/odoo/o-spreadsheet/pull/3371

Task: 4088765

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


LucasLefevre (approved):
robodoo r+

PR #4768, opened 2024-08-02
[FIX] evaluation: handle throwing error on each vectorized position

laa-odoo:
Before this commit, if an error was throw when evaluating a vectorized formula, then the evaluation stopped and only the cell with the formula was marked as an error.
No other values ​​were spread.

This behavior is functionally false, because depending on the arguments passed, it is possible that on this vector position, the evaluation does not return an error.

This commit changes that by making sure to try for each vector position if the evaluation works. Even though it only spreads errors in certain cases.

Task: : [4091276](https://www.odoo.com/web#id=4091276&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/functions/index.ts
+++ b/src/functions/index.ts
@@ -117,17 +117,6 @@ function createComputeFunction(
   descr: FunctionDescription,
   functionName: string
 ): ComputeFunction<Matrix<FunctionResultObject> | FunctionResultObject> {
-  function runtimeCompute(
-    this: EvalContext,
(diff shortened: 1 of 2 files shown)


LucasLefevre (approved):
robodoo r+

PR #4789, opened 2024-08-08
[FIX] table: fix bug when edit table with invalid range

Rachico:
### [FIX] table: fix bug when edit table with invalid range

Before this commit, when editing a table and entering an invalid range, the entire spreadsheet stops working and causes the browser to crash. This is due to an infinite call to `onWillUpdateProps` of the `SelectionInput` component.

This commit fixes the problem by removing a useless range validity check when the input changes.

Task: [4102425](https://www.odoo.com/web#id=4102425&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/table_panel/table_panel.xml
+++ b/src/components/side_panel/table_panel/table_panel.xml
@@ -72,6 +72,7 @@
           t-key="props.table.id"
           ranges="[this.state.tableXc]"
           hasSingleRange="true"
+          isInvalid="this.state.tableZoneErrors.length !== 0"
           onSelectionChanged="(ranges) => this.onRangeChanged(ranges)"
           onSelectionConfirmed.bind="this.onRangeConfirmed"
         />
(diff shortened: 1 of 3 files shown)


src/components/side_panel/table_panel/table_panel.xml
```diff
@@ -84,6 +85,7 @@
         />
       </Section>
       <div class="o-sidePanelButtons">
+        <button t-on-click="resetTable" class="o-table-reset o-button">Reset table</button>
```
hokolomopo, line 88: Not sure we really want a "reset table" button ? It doesn't seems very useeful, does not functionnaly make sense that it would only reset the range and not all the other options, and there should aready be a reset option on the selection input.

Also you should add a test that would address the actual crash.

LucasLefevre (commented):
besides the adapted test, you should add a test that crashed before your fix and not after

PR #4793, opened 2024-08-08
[IMP] TextInput: Auto-select all text on focus

dhrp-odoo:
## Description:

Automatically selects all text in the TextInput component when focused, allowing users to quickly replace it.

Added support for saving with Enter and canceling with Escape.

Task: [4080323](https://www.odoo.com/odoo/project/2328/tasks/4080323)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/text_input/text_input.ts
+++ b/src/components/text_input/text_input.ts
@@ -58,17 +58,45 @@ export class TextInput extends Component<Props, SpreadsheetChildEnv> {
           this.save();
         }
       },
-      {
-        capture: true,
-      }
+      { capture: true }
     );
   }
 
-  onChange() {
-    this.save();
+  onKeyDown(ev: KeyboardEvent) {
+    switch (ev.key) {
+      case "Enter":
+        this.save();
+        ev.preventDefault();
+        ev.stopPropagation();
+        break;
+      case "Escape":
+        if (this.inputRef.el) {
+          this.inputRef.el.value = this.props.value;
+          this.inputRef.el.blur();
+        }
+        ev.preventDefault();
+        ev.stopPropagation();
+        break;
+    }
   }
 
   save() {
-    this.props.onChange((this.inputRef.el?.value || "").trim());
+    const currentValue = (this.inputRef.el?.value || "").trim();
+    if (currentValue !== this.props.value) {
+      this.props.onChange(currentValue);
+    }
+    this.inputRef.el?.blur();
+  }
+
+  focusInputAndSelectContent() {
+    const inputEl = this.inputRef.el;
+    if (!inputEl) return;
+
+    // The onFocus event selects all text in the input.
+    // The subsequent mouseup event can deselect this text,
+    // so t-on-mouseup.prevent.stop is used to prevent this
+    // default behavior and preserve the selection.
+    inputEl.focus();
+    inputEl.select();
   }
 }
--- a/src/components/text_input/text_input.xml
+++ b/src/components/text_input/text_input.xml
@@ -8,7 +8,11 @@
       t-att-id="props.id"
       t-att-placeholder="props.placeholder"
       t-att-value="props.value"
-      t-on-change="onChange"
+      t-on-change="save"
+      t-on-focus="focusInputAndSelectContent"
+      t-on-blur="save"
+      t-on-mouseup.prevent.stop=""
+      t-on-keydown="onKeyDown"
     />
   </div>
 </templates>
--- a/tests/components/text_input.test.ts
+++ b/tests/components/text_input.test.ts
@@ -63,4 +63,30 @@ describe("TextInput", () => {
     await keyDown({ key: "Enter" });
     expect(onChange).toHaveBeenCalledWith("world");
   });
+
+  test("selects input content upon focus", async () => {
+    await mountTextInput({ value: "hello", onChange: () => {} });
+    fixture.querySelector("input")!.focus();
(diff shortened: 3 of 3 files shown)


src/components/side_panel/pivot/editable_name/editable_name.ts
```diff
@@ -9,29 +9,67 @@ interface Props {
   onChanged: (name: string) => void;
 }
 
+interface State {
```
hokolomopo, line 12: The goal pf the other task I mentioned to you is to remove the `Edit` button, and to make the Pivot name into a `TextInput` so I think a lot of your changes in this file are unessecary since they will be deleted.

I think you can add the handling of Enter/Escape key directly inside the TextInputn and the rest of the commit is not needed. You can change the pivot renaming into a `TextInput` yourself without waiting for the other task if you want.

src/components/text_input/text_input.ts
```diff
@@ -56,17 +57,30 @@ export class TextInput extends Component<Props, SpreadsheetChildEnv> {
...
+
+  onBlur() {
+    this.save();
+  }
+
   save() {
     this.props.onChange((this.inputRef.el?.value || "").trim());
   }
+
+  private focusInputAndSelectContent() {
```
hokolomopo, line 80: This seems to only work half of the times somehow

https://drive.google.com/file/d/1OoW4AUPQMZ16YazZCCD694OlNYUqim4Z/view?usp=drive_link

dhrp-odoo: Hello, Sorry for the delay. I spent some time investigating the issue and working on other tasks.

I have identified that the root cause is related to the `mouseup` event. [When text is selected, the mouseup event causes the text to be deselected.](https://stackoverflow.com/questions/23407067/select-not-functioning-correctly-in-chrome)

Can we use `t-on-mouseup.prevent` to address this issue? Or, do you have any other solutions you have encountered in the past that could resolve this?

hokolomopo: In the `bottom_bar_sheet` `focusInputAndSelectContent` we change the selection manually isntead of using `.select()`. Maybe that'd work ? Otherwise yeah I guess a preventDefault is fine. I'd still add a comment somewhere explaining why we have to do this

dhrp-odoo: I’ve tried this as well, but since the `mouseup` event fires immediately after `focus`, it still causes an issue. I’ve added the comment, and the code is now ready for re-validation :)

src/components/text_input/text_input.ts
```diff
@@ -64,11 +64,44 @@ export class TextInput extends Component<Props, SpreadsheetChildEnv> {
...
     this.props.onChange((this.inputRef.el?.value || "").trim());
   }
+
+  onFocus() {
+    this.focusInputAndSelectContent();
+  }
+
+  onBlur() {
+    this.save();
+  }
```
hokolomopo, line 98: can probably remove those methods and directly call `focusInputAndSelectContent` and `save`

src/components/text_input/text_input.ts
```diff
@@ -64,11 +64,44 @@ export class TextInput extends Component<Props, SpreadsheetChildEnv> {
     );
   }
 
+  onKeyDown(ev: KeyboardEvent) {
+    switch (ev.key) {
+      case "Enter":
+        this.save();
+        this.inputRef.el!.blur();
```
hokolomopo, line 71: this might dispatch twice no ? Once on `save` and once on `onBlur` that will call `save`

src/components/text_input/text_input.ts
```diff
@@ -64,11 +64,44 @@ export class TextInput extends Component<Props, SpreadsheetChildEnv> {
...
+        ev.preventDefault();
+        ev.stopPropagation();
+        break;
+      case "Escape":
+        this.inputRef.el!.value = this.props.value;
+        this.inputRef.el!.blur();
```
hokolomopo, line 77: A bit the same here, `onBlur` will call `this.props.onChange`, that might call an `update_pivot` or whatever for nothing. Maybe the `save` could not do anything if the current content is the same as `props.content` ?

src/components/text_input/text_input.xml
```diff
@@ -1,5 +1,10 @@
 <templates>
   <div t-name="o-spreadsheet-TextInput" class="w-100">
+    <!--
```
hokolomopo, line 3: nitpick: not sure but I'd maybe put the comment in the `.ts`. The problem is that comments in the xml will end up in the DOM.

tests/components/text_input.test.ts
```diff
@@ -63,4 +63,30 @@ describe("TextInput", () => {
...
+
+  test("saves the value on input blur", async () => {
+    const onChange = jest.fn();
+    await mountTextInput({ value: "hello", onChange });
+    setInputValueAndTrigger(fixture.querySelector("input")!, "world");
+    fixture.querySelector("input")!.blur();
+    expect(onChange).toHaveBeenCalledWith("world");
+  });
+
+  test("can reset the value with escape key", async () => {
```
hokolomopo, line 82: also test Enter key

hokolomopo (commented):
Some small comments, the feature seems fine otherwise :)

LucasLefevre (approved):
Nice little UX improvements :+1: 
Thanks!
robodoo r+

PR #4820, opened 2024-08-09
[FIX] CF: 'Equal' & 'NotEqual' rules with empty values

dhrp-odoo:
## Description:

How to reproduce:

- Start with an empty spreadsheet.
- Create a new CF with the rule type cellIsRule set to "is Equal to".
- Leave the input field empty.
- Upon validation, all cells in the CF zone are colored.
- Export the spreadsheet as an .xlsx file and then re-import it.
- After re-importing, the cells are no longer colored because the CF rule was exported with a missing value, which was converted to "undefined".

This commit resolves the issue by disallowing the creation of CF rules with empty values for the 'Equal' and 'NotEqual' operators.

Task: : [4092056](https://www.odoo.com/odoo/project/2328/tasks/4092056)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/core/conditional_format.ts
+++ b/src/plugins/core/conditional_format.ts
@@ -341,6 +341,8 @@ export class ConditionalFormatPlugin
             "LessThan",
             "LessThanOrEqual",
             "NotContains",
+            "Equal",
+            "NotEqual",
           ]),
           this.checkOperatorArgsNumber(0, ["IsEmpty", "IsNotEmpty"]),
           this.checkCFValues
--- a/tests/conditional_formatting/conditional_formatting_plugin.test.ts
+++ b/tests/conditional_formatting/conditional_formatting_plugin.test.ts
@@ -532,18 +532,6 @@ describe("conditional format", () => {
     });
   });
 
-  test("Set conditionalFormat on empty cell", () => {
-    let result = model.dispatch("ADD_CONDITIONAL_FORMAT", {
-      cf: createEqualCF("", { fillColor: "#FF0000" }, "1"),
-      ranges: toRangesData(sheetId, "A1"),
-      sheetId,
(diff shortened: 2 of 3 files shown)


tests/plugins/conditional_formatting.test.ts
```diff
@@ -437,16 +437,14 @@ describe("conditional format", () => {
     });
   });
 
-  test("Set conditionalFormat on empty cell", () => {
+  test("Prevents creation of CF CellIsRule with operator 'Equal' and empty value", () => {
```
hokolomopo, line 440: Also add a test for `notEqual` . You could maybe add them both in the test `%s operator with missing first argument %s` where we seem to be testing all the others:)

src/plugins/core/conditional_format.ts
```diff
@@ -276,6 +276,8 @@ export class ConditionalFormatPlugin
             "LessThan",
             "LessThanOrEqual",
             "NotContains",
+            "Equal",
+            "NotEqual",
```
hokolomopo, line 280: That means that commands that were previously valid will now be dropped. To be 100% complete, we should have a migration. 

Well the worth case is that the user is now missing some CF, if he had some CF `isEqual` with no value. That's very niche, and won't break any sheet. So we probably can skip the migrations. But I would still target master instead of 15.0.

hokolomopo (commented):
👋

src/plugins/core/conditional_format.ts
dhrp-odoo: I've retargeted the PR to master 🙂

rrahir:
@robodoo r+

PR #4842, opened 2024-08-13
[FIX] chart: remove empty datasets for chart

anhe-odoo:
## Task Description

This task aims to undisplay empty datasets from a chart (when all the values are undefined, not when they are equals to 0). This has been done by ignoring the empty datasets during the runtime creation.

## Related Task

- Task: 4085864

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/figures/charts/chart_ui_common.ts
+++ b/src/helpers/figures/charts/chart_ui_common.ts
@@ -8,6 +8,7 @@ import { GaugeChartRuntime, ScorecardChartRuntime } from "../../../types/chart";
 import { ChartRuntime, DataSet, DatasetValues, LabelValues } from "../../../types/chart/chart";
 import { formatValue, isDateTimeFormat } from "../../format/format";
 import { deepCopy, range } from "../../misc";
+import { isNumber } from "../../numbers";
 import { recomputeZones } from "../../recompute_zones";
 import { AbstractChart } from "./abstract_chart";
 import { drawGaugeChart } from "./gauge_chart_rendering";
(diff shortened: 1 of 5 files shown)


src/helpers/figures/charts/chart_ui_common.ts
```diff
@@ -255,6 +256,14 @@ export function getChartDatasetValues(getters: Getters, dataSets: DataSet[]): Da
       label = `${ChartTerms.Series} ${parseInt(dsIndex) + 1}`;
     }
     let data = ds.dataRange ? getData(getters, ds) : [];
+    if (
+      data.every(
+        (cell) =>
+          cell === undefined || cell === null || !isNumber(cell.toString(), getters.getLocale())
```
LucasLefevre, line 262: doesn't this breaks what's just below where it looks like the dataset should *not* be ignored if it contains only strings.

tests/figures/chart/bar_chart_plugin.test.ts
```diff
@@ -59,7 +59,22 @@ describe("bar chart", () => {
...
+          {
+            name: "Sheet1",
+            colNumber: 10,
+            rowNumber: 10,
+            rows: {},
+            cells: {
+              B1: { content: "first column dataset" },
+              B2: { content: "10" },
+              B3: { content: "11" },
+              B4: { content: "12" },
```
LucasLefevre, line 73: avoid `beforeEach` with setup data please. Because it impacts all tests in the `describe` even if some/most of them do not need it. Tests setup should be minimal to test a given behavior.

src/helpers/figures/charts/chart_ui_common.ts
anhe-odoo: Yes, you're right. I'll invert the order of the tests

LucasLefevre (commented):
robodoo r+

LucasLefevre:
robodoo r+

rrahir:
@robodoo r+

PR #4862, opened 2024-08-19
[FIX] pivot: mark pivot as invalid if column name is empty

pro-odoo:
Steps to reproduce, given a range with some data:
* Replace the column title by a formula resulting in an empty string (="")
* select the entire range and insert a pivot table => unexpected error: An unexpected error occurred. Submit a support ticket at odoo.com/help. Field undefined does not exist

This was caused by the verification function of the column name, which only checked if the column name was empty, but a cell with a formula can return an empty string, which is not the same as an empty cell.

Task: 4116392

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


LucasLefevre (approved):
robodoo r+

PR #4878, opened 2024-08-22
[IMP] clipboard: copy whole pivot as dynamic pivot

hokolomopo:
## Description

When we copy part of a dynamic pivot, we don't copy the values but pivot formulas instead.

But when copying (or cutting) the whole pivot, we should copy the dynamic pivot and not the pivot formulas. The intent of the user in this case is most likely to copy the pivot as a whole and not to replace it with pivot formulas (particularly when cutting).

Task: : [4123272](https://www.odoo.com/web#id=4123272&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


LucasLefevre (approved):
robodoo r+

LucasLefevre (commented):
robodoo r+

PR #4897, opened 2024-08-22
Master composer refact colors laa

laa-odoo:
Task: [4123205](https://www.odoo.com/web#id=4123205&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/components/composer/composer/abstract_composer_store.ts
```diff
@@ -55,6 +55,7 @@ export const tokenColors = {
   RIGHT_PAREN: functionColor,
   ARG_SEPARATOR: functionColor,
   MATCHING_PAREN: "#000000",
+  EXTRA_PAREN: "#ff0000",
```
hokolomopo, line 57: nitpick: UNMATCHED_PAREN maybe ?

src/formulas/composer_tokenizer.ts
```diff
@@ -66,19 +67,55 @@ function enrichTokens(tokens: Token[]): EnrichedToken[] {
...
+      }
+      previousSymbolIndex = undefined;
     } else if (token.type === "RIGHT_PAREN") {
-      token.parenIndex = stack.pop();
+      token.parenthesesCode = code;
+      if (
+        code !== "" // necessary condition to not loose nextLvl information when to much parenthesis
+      ) {
+        // get the last character
+        nextLvl = Number(code.charAt(code.length - 1));
```
hokolomopo, line 108: `Number(code.at(- 1));`

src/formulas/composer_tokenizer.ts
```diff
@@ -66,19 +67,55 @@ function enrichTokens(tokens: Token[]): EnrichedToken[] {
...
+      if (previousSymbolIndex !== undefined) {
+        for (let previousIndex = previousSymbolIndex; previousIndex < i; previousIndex++) {
+          tokens[previousIndex].parenthesesCode = code;
+        }
+      }
+      previousSymbolIndex = undefined;
     } else if (token.type === "RIGHT_PAREN") {
-      token.parenIndex = stack.pop();
+      token.parenthesesCode = code;
+      if (
```
hokolomopo, line 104: the comments of this if don't seem that useful

src/formulas/composer_tokenizer.ts
```diff
@@ -200,5 +237,7 @@ function addArgsAST(tokens: EnrichedToken[]): EnrichedToken[] {
 export function composerTokenize(formula: string, locale: Locale): EnrichedToken[] {
   const tokens = rangeTokenize(formula, locale);
 
-  return addArgsAST(mapParentFunction(mapParenthesis(enrichTokens(tokens))));
+  const temp = mapParenthesisCode(enrichTokens(tokens));
+
+  return addArgsAST(mapParentFunction(temp));
```
hokolomopo, line 244: why this change ?

src/components/composer/composer/abstract_composer_store.ts
```diff
@@ -556,6 +559,51 @@ export abstract class AbstractComposerStore extends SpreadsheetStore {
...
+    }
+    return code;
+  }
+
+  private isTokenInCursorFormulaContext(
+    token: EnrichedToken,
+    // we refer to the previous symbol parenthesesCode and not directly the
+    // parenthesesCode of the token at the cursor because we don't want to
+    // match cases where the token at the cursor is between parentheses which
+    // are not function parentheses
```
hokolomopo, line 598: Why shoudn't we highlight the content of the parenthesis if it isn't a function ? It make sense no ?

hokolomopo: and the function could probably be inlined

src/formulas/composer_tokenizer.ts
```diff
@@ -30,19 +30,20 @@ interface FunctionContext {
...
 
 export interface EnrichedToken extends Token {
   start: number;
   end: number;
   length: number;
-  parenIndex?: number;
+  parenthesesCode?: string;
   functionContext?: FunctionContext;
   color?: Color;
+  isFormulaCursorContextToken?: boolean;
```
hokolomopo, line 47: I'd put `isHighlighted` or something. `isFormulaCursorContextToken` is a bit unclear, and the outside world doesn't care why the token is highlighted or not

hokolomopo: Or maybe the compsoer store could directly change the color

src/components/composer/composer/abstract_composer_store.ts
```diff
@@ -556,6 +559,51 @@ export abstract class AbstractComposerStore extends SpreadsheetStore {
     return highlight && highlight.color ? highlight.color : undefined;
   }
 
+  private computeFormulaCursorContext() {
```
hokolomopo, line 559: need a doctsring

src/components/composer/composer/abstract_composer_store.ts
```diff
@@ -556,6 +559,51 @@ export abstract class AbstractComposerStore extends SpreadsheetStore {
...
+          this.currentTokens[i].isFormulaCursorContextToken = this.isTokenInCursorFormulaContext(
+            this.currentTokens[i],
+            previousSymbolAtCursor
+          );
+        }
+        return;
+      }
+    }
+
+    for (let i = 0; i < this.currentTokens.length; i++) {
```
hokolomopo, line 577: wouldn't it make more sense to only change the tokens that need to have a different color ?

src/formulas/composer_tokenizer.ts
```diff
@@ -44,6 +45,7 @@ export interface EnrichedToken extends Token {
   functionContext?: FunctionContext;
   color?: Color;
   isFormulaCursorContextToken?: boolean;
+  isRelatedToCursorToken?: boolean;
```
hokolomopo, line 48: Same here, `isRelatedToCursorToken` doesn't mean much. We probably should have `highlight` and `lightenedColor` keys or something

src/components/composer/composer/abstract_composer_store.ts
```diff
@@ -609,6 +601,37 @@ export abstract class AbstractComposerStore extends SpreadsheetStore {
     return (token.parenthesesCode || "").startsWith(previousSymbolAtCursor.parenthesesCode || "");
   }
 
+  private computeParenthesisRelatedToCursor() {
+    if (
+      this.tokenAtCursor &&
+      ["LEFT_PAREN", "RIGHT_PAREN"].includes(this.tokenAtCursor.type) &&
```
hokolomopo: On Gsheet the parenthesis are highligted if the cursor is AFTER the opening parenthesis, or BEFORe the closing parenthesis

src/components/composer/composer/abstract_composer_store.ts
```diff
@@ -609,6 +601,37 @@ export abstract class AbstractComposerStore extends SpreadsheetStore {
...
+      this.tokenAtCursor.parenthesesCode !== ""
+    ) {
+      const matchingParenthesisIndexes = this.currentTokens.reduce((acc, t, index) => {
```
hokolomopo: I'm a bit tired so I didn't read all the function, but shouldn't this function be trivial ? If there is a parenthesis at the cursor, just loop on all the token to find a parenthesis with the same code and highlight them both.

I'm not sure how we could have multiple parenthesis with the same code ? Or your algorithm to add the code to the token is wrong ...

And we shoould mark the two parenthesis tokens and that's all, I don't think we should mark all the other tokens as not related to the cursor or w/e

src/components/composer/composer/abstract_composer_store.ts
```diff
@@ -601,37 +601,72 @@ export abstract class AbstractComposerStore extends SpreadsheetStore {
...
-            ? true
-            : false;
-        }
-        return;
+    if (this.tokenAtCursor) {
+      switch (this.tokenAtCursor.type) {
+        case "LEFT_PAREN":
+        case "RIGHT_PAREN":
+          if (this.tokenAtCursor.parenthesesCode) {
+            matchingTokenIndexes = this.currentTokens.reduce(
```
hokolomopo, line 612: a functionsimply looping on all the tokens would probably me easier to read than a  reduce

hokolomopo (commented):
Nice split of commits :+1: 

The whole thing seems very complex tho, there's probably a bit of room to make things simpler/easier to understand.

src/components/composer/composer/abstract_composer_store.ts
laa-odoo: I will say that 

1) It is important for the user to always have visual information on which sub-formula he is editing. 
Particularly when he is trying to edit a formula, and he moves his cursor inside it

2) That in the case where he edits a mathematical operation composed of several parentheses, then he can easily move his cursor to the nearest parenthesis to have another visual information on the corresponding matching parenthesis

inlined cases are taken into account

src/formulas/composer_tokenizer.ts
laa-odoo: what do you think of isBlurred ?
highlighted will be redundant with the ranges which are already highlighted on the colors

src/formulas/composer_tokenizer.ts
laa-odoo: an oversight used for debugging :D

src/components/composer/composer/abstract_composer_store.ts
laa-odoo: I don't like 'unmatched' because you can have an opening parenthesis while you haven't finished editing your formul --> in which case the parenthesis matches nothing, so you shouldn't highlight it and yet it's not a parenthesis too many

src/components/composer/composer/abstract_composer_store.ts
laa-odoo: you must first ensure that the parentheses exist as a duo to highlight them. A parenthesis without its girlfriend means nothing. It therefore has no reason to be highlighted. It's the sad life of parentheses :'(

ok to not mark all, (my bad)

src/formulas/composer_tokenizer.ts
laa-odoo: there is too much notion of different "light" on the composer, so I am not convinced by `lightenedColo`. For this also that I name by the functional meaning (but yes not necessarily better either).

To do with others maybe

src/components/composer/composer/abstract_composer_store.ts
hokolomopo: Yeah but you know the parenthesis at the cursor, and you have its code. So you could just loop on all the tokens and if you find another parenthesis with the same code you highlight them both then return. I 'ond think we need to loop multuple times on the tokens ?

laa-odoo: in fact (in a simple version of the code) I will still need to go through the tokens twice, 

here we are in a case where the tokens are not necessarily reset: when you simply move the position cursor, the parentheses which were previously highlighted (isRelatedToCursorToken to true) must return to false 

in any case the whole code changes drastically in the following commit, it is only a passing state

I don't know if that really answers your question.

src/components/composer/composer/abstract_composer_store.ts
laa-odoo: I don't see this behavior on gsheet, on excel perhaps ?

src/components/composer/composer/abstract_composer_store.ts
laa-odoo: I added a "[REF] to squash commit", tell me if you prefer this version

laa-odoo:
> Nice split of commits 👍
> 
> The whole thing seems very complex tho, there's probably a bit of room to make things simpler/easier to understand.

thx for your review :)

src/components/composer/composer/abstract_composer_store.ts
hokolomopo: MMh yeah it seems i'm hallucinating now ....

laa-odoo: perhaps, but tell me you see walls that go into the ground, castles that move and viking boats coming out of the water... I would believe you

src/components/composer/composer/abstract_composer_store.ts
```diff
@@ -487,6 +513,145 @@ export abstract class AbstractComposerStore extends SpreadsheetStore {
...
+        findAtLeastOneMatch = true;
+        this.currentTokens[i].isRelatedToCursorToken = true;
+        continue;
+      }
+      this.currentTokens[i].isRelatedToCursorToken = false;
+    }
+
+    if (findAtLeastOneMatch) {
+      this.tokenAtCursor!.isRelatedToCursorToken = true;
+    }
```
hokolomopo, line 652:
```suggestion
    for (let i = 0; i < this.currentTokens.length; i++) {
      if (
        doesTokenMatchCursorToken(this.currentTokens[i]) &&
        this.currentTokens[i] !== this.tokenAtCursor
      ) {
        this.tokenAtCursor!.isRelatedToCursorToken = true;
        this.currentTokens[i].isRelatedToCursorToken = true;
        continue;
      }
      this.currentTokens[i].isRelatedToCursorToken = false;
    }
```
can be simplified a bit. But it looks way better now :D

laa-odoo: indeed, thx

LucasLefevre, line 652:
```ts
    // reset everything first
    for (let i = 0; i < this.currentTokens.length; i++) {
      this.currentTokens[i].isRelatedToCursorToken = false;
    }
    for (let i = 0; i < this.currentTokens.length; i++) {
      if (
        doesTokenMatchCursorToken(this.currentTokens[i]) &&
        this.currentTokens[i] !== this.tokenAtCursor
      ) {
        this.currentTokens[i].isRelatedToCursorToken = true;
        this.tokenAtCursor!.isRelatedToCursorToken = true;
        continue;
      }
    }
```

src/components/composer/composer/composer.ts
```diff
@@ -33,6 +33,7 @@ const ASSISTANT_WIDTH = 300;
 const CLOSE_ICON_RADIUS = 9;
 
 export const selectionIndicatorClass = "selector-flag";
+export const backgroundIndicatorClass = "background-flag";
```
LucasLefevre, line 36: why "indicator" ?

LucasLefevre, line 36: doesn't need to be exported

src/formulas/composer_tokenizer.ts
```diff
@@ -34,6 +34,7 @@ interface FunctionContext {
  * - functionContext (only for tokens surrounded by a function)
  * - color (the base color of the token, without fading modification)
  * - isBlurred (indicates whether the text should be highlighted or not, useful to know if the token matches the same formula as the cursor token)
+ * - hasRelatedTokens (Useful for highlighting tokens related to the current token, such as matching parentheses or ranges)
```
LucasLefevre, line 37: not up-to-date

LucasLefevre (commented):
first quick review.

From a functional POV, I'd

- have the greyed elements less blurred (closer to what GSheet does?)
- the background less strong
- matching parenthesis in bold (if easy)
- not sure about the functional benefit of highlighting the same references

src/formulas/composer_tokenizer.ts
```diff
@@ -66,19 +67,53 @@ function enrichTokens(tokens: Token[]): EnrichedToken[] {
...
+      }
+      previousSymbolIndex = undefined;
     } else if (token.type === "RIGHT_PAREN") {
-      token.parenIndex = stack.pop();
+      token.parenthesesCode = code;
+      if (code !== "") {
+        // get the last character
+        nextLvl = Number(code.at(-1));
+        // remove the last character
+        code = code.slice(0, -1);
```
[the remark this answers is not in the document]
laa-odoo: very very very good comment (:

laa-odoo:
> first quick review.
> 
> From a functional POV, I'd
> 
> * have the greyed elements less blurred (closer to what GSheet does?)
> * the background less strong
> * matching parenthesis in bold (if easy)
> * not sure about the functional benefit of highlighting the same references

* grayed elements are less blurry
*highlighted references have been removed
* matching  parentheses in bold too complicated, to do in another task
* the background remains strong to compensate for the non-boldness of the parentheses

src/components/composer/composer/abstract_composer_store.ts
LucasLefevre: ORPHAN_RIGHT_PAREN ?

src/formulas/composer_tokenizer.ts
LucasLefevre: `isParenthesisLinkedToCursor` ? `isParenthesisHighlighted` ?

I'd add the notion of parenthesis in the name since it's only used for parenthesis (for now at least).
I agree that `isRelatedToCursorToken` doesn't mean much and you have to look at the implementation to know what it is/does.

src/components/composer/composer/composer.ts
```diff
@@ -642,64 +643,55 @@ export class Composer extends Component<CellComposerProps, SpreadsheetChildEnv>
...
-          result.push({ value: token.value, color: "#000" });
-          break;
+      result.push({ value: token.value, color: this.getTokenColor(token) });
+
+      if (
+        token.type === "REFERENCE" &&
+        this.props.composerStore.tokenAtCursor === token &&
+        this.props.composerStore.editionMode === "selecting"
+      ) {
+        result[result.length - 1].class = "text-decoration-underline";
```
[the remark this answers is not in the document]
laa-odoo: I explicitly wanted `getColoredToken` to return a color (Distinguishing color management from the rest)

Here `"text-decoration-underline"` (initially present) has nothing to do with color. In which case you will probably tell me that the function `getColoredTokens()` in which we are, is not suitable for what the function initially does. And I completely agree, we could rename it.

Maybe something like `getHtmlTokens` or `getHtmlContentFromTokens` ?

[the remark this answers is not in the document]
LucasLefevre: Yup, let's rename it.

also, it could also be moved to the store I think

src/formulas/composer_tokenizer.ts
```diff
@@ -65,19 +69,52 @@ function enrichTokens(tokens: Token[]): EnrichedToken[] {
...
+      }
+      previousSymbolIndex = undefined;
+    } else if (token.type === "SYMBOL") {
+      token.parenthesesCode = code.join(":");
+      previousSymbolIndex = i;
+    } else if (token.type === "SPACE") {
+      token.parenthesesCode = code.join(":");
+    } else {
+      token.parenthesesCode = code.join(":");
+      previousSymbolIndex = undefined;
```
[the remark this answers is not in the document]
laa-odoo: in fact `previousSymbolIndex` was introduced to specifically manage the link between 
the token of the opening parenthesis with the token of the associated function name. (see `if (previousSymbolIndex !== undefined)`)

Functionally, we want to keep the link between the opening parenthesis and the name of the function, even if space tokens are inserted between the name and the parenthesis.

This explains why ` else if (token.type === "SPACE") {` is a special case where we should not reset `previousSymbolIndex` to `undefined`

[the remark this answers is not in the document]
laa-odoo: an alternative without `previousSymbolIndex`

```typescript
function mapParenthesisCode(tokens: EnrichedToken[]): EnrichedToken[] {
  let code: number[] = [];
  let nextLevel = 0;
  return tokens.map((token, i) => {
    if (token.type === "LEFT_PAREN") {
      code.push(nextLevel + 1);
      nextLevel = 0;
      token.parenthesesCode = code.join(":");
      // allows to link the parentheses opening a function to the function
      for (let previousIndex = i - 1; previousIndex >= 0; previousIndex--) {
        if (tokens[previousIndex].type === "SPACE") {
          tokens[previousIndex].parenthesesCode = code.join(":");
          continue;
        }
        if (tokens[previousIndex].type === "SYMBOL") {
          tokens[previousIndex].parenthesesCode = code.join(":");
        }
        break;
      }
    } else if (token.type === "RIGHT_PAREN") {
      token.parenthesesCode = code.join(":");
      if (code.length) {
        nextLevel = code.pop()!;
      }
    } else {
      token.parenthesesCode = code.join(":");
    }
    return token;
  });
}
```

src/components/composer/composer/composer.ts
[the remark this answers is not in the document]
laa-odoo: if we have fun moving `"text-decoration-underline"` in the abstract, then `class` could become a token attribute, and we just have to spread token on HtmlContent in the composer, did you see it like that ?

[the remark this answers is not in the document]
LucasLefevre: I meant moving `getHtmlContentFromTokens` to the store. Not the `class` on the token

LucasLefevre (approved):
robodoo rebase-ff r+

PR #4954, opened 2024-09-05
Master fhe perf lul

LucasLefevre:
## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/functions/module_array.ts
+++ b/src/functions/module_array.ts
@@ -334,7 +334,7 @@ export const MINVERSE = {
       )
     ),
   ],
-  compute: function (matrix: Arg): Matrix<number> {
+  compute: function (matrix: Arg) {
     const _matrix = toNumberMatrix(matrix, "square_matrix");
     assertSquareMatrix(
       _t("The argument square_matrix must have the same number of columns and rows."),
@@ -342,7 +342,7 @@ export const MINVERSE = {
     );
     const { inverted } = invertMatrix(_matrix);
     if (!inverted) {
-      throw new EvaluationError(_t("The matrix is not invertible."));
+      return new EvaluationError(_t("The matrix is not invertible."));
     }
     return inverted;
   },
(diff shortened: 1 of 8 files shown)


laa-odoo:
in this case you might as well remove the return types from all the formulas, right? having a transient state where you don't know when you develop a formula whether you should put a return type or not will annoy me

laa-odoo:
if the goal is to gain performance this way. Might consider reviewing how the casting function returns the error as well

laa-odoo:
also in this approach I wonder when you replace that

```typescript
throw new EvaluationError(_t("The number of columns must be positive."));
```
by
```typescript
return new EvaluationError(_t("The number of columns must be positive."));
```
Wouldn't replacing instead with that also make it possible to gain more? 

```typescript
return  {value: CellErrorType.GenericError, message: _t("The number of columns must be positive.")}
```

(in any case this would allow us to keep a typing on the return of the function)

LucasLefevre:
> in this case you might as well remove the return types from all the formulas, right? having a transient state where you don't know when you develop a formula whether you should put a return type or not will annoy me

IMO return types are useful when they help, but can be omitted if it's too lengthy and we can let Typescript infer it

LucasLefevre:
> if the goal is to gain performance this way, might consider reviewing how the casting function returns the error as well

Any idea ? I only converted "easy" ones

LucasLefevre:
> Wouldn't replacing instead with that also make it possible to gain more?
return  {value: CellErrorType.GenericError, message: _t("The number of columns must be positive.")}

probably yes (especially given EvaluationError extends Error). But it's so much shorter and nicer with `return new EvaluationError()` :cry: 

> (in any case this would allow us to keep a typing on the return of the function)

It would be the same. What's annoying is with mixed return types (matrices and simple value/error)

laa-odoo:
> > if the goal is to gain performance this way, might consider reviewing how the casting function returns the error as well
> 
> Any idea ? I only converted "easy" ones


open task to do this :)
https://www.odoo.com/odoo/project/2328/tasks/4208351

LucasLefevre:
robodoo rebase-ff

PR #4956, opened 2024-09-06
[FIX] pivot: add suggestion message to broken pivot formulas

LucasLefevre:
## Description:

Since the pivot insertion is inserting static formulas, if the user changes the groups, the static formulas are no longer correct and are in error. The user might not understand what's happening and how to fix the issue.

This commit adds more info to the error message to help the user.

A similar issue happens with measures: when adding a new measure, nothing happens in the grid => no feedback.

Technical note: instead of changing the existing message, I added a second one, which I concatenate to the existing one. The reason is to avoid breaking existing translations in a stable version.

Task: [4169586](https://www.odoo.com/web#id=4169586&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/pivot/pivot_side_panel/pivot_side_panel_store.ts
+++ b/src/components/side_panel/pivot/pivot_side_panel/pivot_side_panel_store.ts
@@ -2,6 +2,7 @@ import { deepCopy, deepEquals } from "../../../../helpers";
 import { isDateField } from "../../../../helpers/pivot/pivot_helpers";
 import { pivotRegistry } from "../../../../helpers/pivot/pivot_registry";
 import { Get } from "../../../../store_engine";
+import { NotificationStore } from "../../../../stores/notification_store";
 import { SpreadsheetStore } from "../../../../stores/spreadsheet_store";
 import { _t } from "../../../../translation";
 import { Command, UID } from "../../../../types";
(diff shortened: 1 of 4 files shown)


src/components/side_panel/pivot/pivot_defer_update/pivot_defer_update.xml
```diff
@@ -1,25 +1,29 @@
...
-        onChange="(value) => props.toggleDeferUpdate(value)"
-      />
-      <div t-if="props.isDirty">
-        <i
-          class="btn btn-sm pe-0 fa fa-undo"
-          title="Discard all changes"
-          t-on-click="() => props.discard()"
+    <Section class="'border-top d-flex flex-column justify-content-between py-1'">
+      <t t-set="message">Consider using a dynamic pivot with =PIVOT(1)</t>
+      <ValidationMessages messages="[message]" msgType="'info'"/>
```
hokolomopo, line 5: image

Style is messed up:
- the validation show as an error with no background 
- only "defer update" has a gray background, it should be the whole panel footer

hokolomopo, line 5: Shoudn't we have a t-if of some sort ? The message is kinda useless if the only pivot formulas in the spreadsheet are dynamic

hokolomopo (commented):
:partying_face:

LucasLefevre:
oups, forgot to push the latest version :sweat_smile:

hokolomopo (commented):
@robodoo r+

PR #4960, opened 2024-09-10
[FIX] Find&Replace: always search cell content if `searchFormula`

rrahir:
The current implementation of a search with the option `serachFormula` will only look into the cell content if the cell is a valid formula cell and not a cell which results in an error following its evaluation.

e.g.  A1: '=SUM("a")' is invalid thus the related cell is not a formula cell but an error cell and will never be matched when looking for the term "sum".

Task: 4175968

## Description:

Task: [4175968](https://www.odoo.com/web#id=4175968&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui/find_and_replace.ts
+++ b/src/plugins/ui/find_and_replace.ts
@@ -151,11 +151,7 @@ export class FindAndReplacePlugin extends UIPlugin {
           cell &&
           this.currentSearchRegex &&
           this.currentSearchRegex.test(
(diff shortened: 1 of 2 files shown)


LucasLefevre (approved):
robodoo r+

PR #4961, opened 2024-09-10
[IMP] pivot: add indentation to row headers

hokolomopo:
### [IMP] pivot: add indentation to row headers

This commit adds indentation to row headers in the pivot table, based
on the depth of the groupBy.

### [IMP] format: add the handling of repeated character

This commit adds the handling of the `*` asterisk character in the format,
which repeats the character that comes after it enough times to fill
the column.

The asterisk is ignored in `cell.formattedValue`, and the character is
only repeated if the `formatValue` helper is given a `width` parameter,
which we now do only in the grid renderer.

The commit also adds this new format feature to the accounting format,
to match the behaviour of the other spreadsheet applications.

Task: [3965246](https://www.odoo.com/web#id=3965246&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


hokolomopo:
So the issue is that I only use the repeated character to fill the empty space of the column. But the char has a discrete width, so we end up with things that aren't perfectly aligned, and that can be a bit ugly
image

1. Just leave it like that ATM, and maybe in the future improve it
2. Do like GSheet & Excel, and add an "empty space" at the end of the repeated characters so that it's perfectly aligned. It sounds kind of annoying to implement TBH (maybe some kind of rich text ?)

LucasLefevre (approved):

src/types/pivot.ts
```diff
@@ -166,3 +166,9 @@ export interface DimensionTreeNode {
 }
 
 export type DimensionTree = DimensionTreeNode[];
+
+/** Pivot domain splitted for the domain related to the pivot's rows and columns  */
+export interface PivotColRowDomain {
```
LucasLefevre, line 171: On n'a pas déjà ça ?

hokolomopo: It wasn't merged yet when I coded the PR :) I'll remove it on rebase

src/helpers/format/format.ts
```diff
@@ -130,9 +145,46 @@ function applyTextInternalFormat(
...
+    const remainingWidth = availableWidth - timesToRepeat * charWidth;
+    const paddingChar = "\u2009"; // thin space
+    const paddingWidth = formatWidth.measureText(paddingChar);
+    const padding = paddingChar.repeat(Math.floor(remainingWidth / paddingWidth));
+
+    return { timesToRepeat, padding };
+  }
+
+  const { timesToRepeat, padding } = getTimesToRepeat();
+  return prefix + repeatedChar.repeat(timesToRepeat) + padding + suffix;
```
LucasLefevre, line 187: padding should be before the repeated chars

src/helpers/pivot/pivot_helpers.ts
```diff
@@ -14,6 +21,7 @@ import {
   PivotField,
   PivotTableCell,
 } from "../../types/pivot";
+import { domainToColRowDomain } from "./pivot_domain_helpers";
```
LucasLefevre, line 24: circular dependency :/

src/functions/module_lookup.ts
```diff
@@ -829,7 +830,8 @@ export const PIVOT = {
             result[col].push({ value: "" });
             break;
           case "HEADER":
-            result[col].push(pivot.getPivotHeaderValueAndFormat(pivotCell.domain));
+            const value = pivot.getPivotHeaderValueAndFormat(pivotCell.domain);
```
LucasLefevre, line 833: value => result ?
`value` makes me think it's only the value, but it's actually the object with the format

src/helpers/pivot/pivot_time_adapter.ts
```diff
@@ -183,7 +183,7 @@ const quarterNumberAdapter: PivotTimeAdapterNotNull<number> = {
   toValueAndFormat(normalizedValue) {
     return {
       value: _t("Q%(quarter_number)s", { quarter_number: normalizedValue }),
-      format: "0",
+      format: "@",
```
LucasLefevre, line 186: should also be adapter for new granularities: day of week, hours, minutes, seconds

src/helpers/format/format.ts
```diff
@@ -64,13 +66,22 @@ export const DAYS: Readonly<Record<number, string>> = {
...
+interface FormatWidth {
+  availableWidth: number;
+  measureText: (text: string) => number;
+}
+
 /**
  * Formats a cell value with its format.
  */
-export function formatValue(value: CellValue, { format, locale }: LocaleFormat): FormattedValue {
+export function formatValue(
```
LucasLefevre, line 77: can we have at least a few tests for this feature in `format_helpers.test.ts` ? (with mocked measures)

src/helpers/format/format.ts
```diff
@@ -130,9 +145,46 @@ function applyTextInternalFormat(
...
+    const charWidth = formatWidth.measureText(repeatedChar);
+    const availableWidth = formatWidth.availableWidth - widthTaken;
+    if (availableWidth <= 0) {
+      return { timesToRepeat: 0, padding: "" };
+    }
+
+    const timesToRepeat = Math.floor(availableWidth / charWidth);
+
+    const remainingWidth = availableWidth - timesToRepeat * charWidth;
+    const paddingChar = "\u2009"; // thin space
```
LucasLefevre, line 179: :+1: nice, thanks !

src/helpers/format/format.ts
hokolomopo: Why ? For a format like `*a0.00` it make more sense to have the padding after the repeated character. But there is probably an example where the padding make more sense before the repeated character.

I don't mind having the padding before, but both are equally problematic no ? Or do you see a specific reason it's better before ?

src/helpers/pivot/pivot_helpers.ts
hokolomopo: I think it make a bit more sense for `pivot_helpers` to be able to import `pivot_domain_helpers` ? I moved the constant `PREVIOUS_VALUE` inside `pivot_domain_helpers` to avoid havign somethign circular.

src/functions/module_lookup.ts
hokolomopo: `result` is already used in this scope. Repalced by `valueAndFormat`

src/helpers/format/format.ts
LucasLefevre: > Or do you see a specific reason it's better before ?

not really, except mimicking what Gsheet does

hokolomopo: aaaaaand Excel does add the padding after the repeated character 😅 I vote to leave it as it is 🤷‍♂️

LucasLefevre: :+1:

LucasLefevre (approved):
robodoo rebase-ff r+

PR #4965, opened 2024-09-11
[FIX] menu_items_actions: hide reinsert pivot menu

dhrp-odoo:
## Description:

Previously, attempting to reinsert a pivot with an invalid range resulted in a traceback error. This PR resolves the issue by hiding the menu item when the specified pivot range is invalid.

Task: [4137926](https://www.odoo.com/odoo/project/2328/tasks/4137926)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


rrahir:
While it's not related to the task itself, I realized that we don't have any test that check t he visibility (rather the invisibility) of the re-insert pivot menu items when there aren't any pivots in the sheet. I think the tests exist in Odoo but since we replace the items completely, it would make sense to add a test in here for the spreadsheet pivots :)

rrahir:
@robodoo r+

PR #4967, opened 2024-09-11
[IMP] custom color: handle all the chart colors

hokolomopo:
## Description

The custom color plugin wasn't handling data series color, or chart title color. Now we run a regex through the whole chart definition to get all the colors, so we don't have to think to add new handling in the custom color plugin every time we add a chart customization.

Task: [4178763](https://www.odoo.com/web#id=4178763&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_core_views/custom_colors.ts
+++ b/src/plugins/ui_core_views/custom_colors.ts
@@ -9,9 +9,10 @@ import {
   rgbaToHSLA,
   toHex,
 } from "../../helpers";
-import { GaugeChart, ScorecardChart } from "../../helpers/figures/charts";
-import { Color, CoreViewCommand, Immutable, RGBA, TableElementStyle, UID } from "../../types";
+import { Color, Command, Immutable, RGBA, TableElementStyle, UID } from "../../types";
 import { UIPlugin, UIPluginConfig } from "../ui_plugin";
+
+const chartColorRegex = /"(#[0-9a-fA-F]{6})"/g;
 
 /**
  * https://tomekdev.com/posts/sorting-colors-in-js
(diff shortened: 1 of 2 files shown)


src/plugins/ui_core_views/custom_colors.ts
```diff
@@ -157,29 +163,17 @@ export class CustomColorsPlugin extends UIPlugin<CustomColorState> {
...
-          chartsColors.add(colors.middleColor);
-          chartsColors.add(colors.upperColor);
-          break;
-        case "scorecard":
-          const scoreChart = chart as ScorecardChart;
-          chartsColors.add(scoreChart.baselineColorDown);
-          chartsColors.add(scoreChart.baselineColorUp);
-          break;
+    const stringifiedChart = JSON.stringify(chart.getDefinition());
+    const colors = stringifiedChart.match(chartColorRegex);
```
rrahir, line 173:
```suggestion
    const colors = stringifiedChart.matchAll(chartColorRegex);
```
will return the matching groups as well so you don"'t have to do string manipulation afterwards ;-)

src/plugins/ui_core_views/custom_colors.ts
```diff
@@ -76,16 +77,23 @@ export class CustomColorsPlugin extends UIPlugin<CustomColorState> {
...
   }
 
-  handle(cmd: CoreViewCommand) {
+  handle(cmd: Command) {
     switch (cmd.type) {
-      case "UPDATE_CELL":
+      case "START":
+        for (const sheetId of this.getters.getSheetIds()) {
+          for (const chartId of this.getters.getChartIds(sheetId)) {
+            this.tryToAddColors(this.getChartColors(chartId));
```
rrahir, line 88: shouldn't it concern every color source?

hokolomopo: the other ones are computed at every finalize, so we shouldn't need to bother with them there

rrahir:
@robodoo r+

rrahir:
@robodoo r-

src/plugins/ui_core_views/custom_colors.ts
```diff
@@ -220,13 +208,15 @@ export class CustomColorsPlugin extends UIPlugin<CustomColorState> {
...
-    if (!isColorValid(color)) {
-      return;
-    }
-    const formattedColor = toHex(color);
-    if (color && !COLOR_PICKER_DEFAULTS.includes(formattedColor)) {
-      this.history.update("customColors", formattedColor, true);
+  private tryToAddColors(colors: Color[]) {
+    for (const color of colors) {
+      if (!isColorValid(color)) {
+        return;
```
LucasLefevre, line 214: otherwise, as soon as one color is invalid, all the following colors are ignored
```suggestion
        continue;
```

LucasLefevre (commented):
`this.getChartColors(sheetId)` in `computeCustomColors` is completely wrong :stuck_out_tongue_closed_eyes: (sheetId vs expected ChartId).
It can be removed.
But actually, why are charts handled differently (not at finalize with the other stuff)?

hokolomopo:
> But actually, why are charts handled differently (not at finalize with the other stuff)?

Because it's way better performance-wise to not parse the chart every time we have some random command that triggers a finalize, but to only do that when we have a command that actually updates the chart. 

Actually I think I'll do another commit to compute every color this way, the plugin does not care at all about the evaluation, so it make no sense to do things in the finalize ? (I guess there was some older version of this plugin where it mattered). And handling the commands separatly would be way better than looping on every cell at each command.

LucasLefevre (approved):
robodoo r +

PR #4979, opened 2024-09-13
[FIX] export: Fix default style ignored at export

rrahir:
The default style values ignored at export were targetting the font color black (#000000) as the default value. However, the default value is the simple absence of value (or rather an empty string) as we should be able to differentiate it from a choice from the user to set the font color to black (e.g. when the cell is also receiving its style from a CF or a table).

Task: 4178743

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -162,7 +162,7 @@ export const DEFAULT_STYLE = {
   underline: false,
   fontSize: 10,
   fillColor: "",
-  textColor: "#000000",
+  textColor: "",
 } satisfies Required<Style>;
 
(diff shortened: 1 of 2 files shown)


src/constants.ts
```diff
@@ -162,7 +162,7 @@ export const DEFAULT_STYLE = {
   underline: false,
   fontSize: 10,
   fillColor: "",
-  textColor: "#000000",
```
hokolomopo, line 165: Maybe add a random test ? :)

rrahir: yup, done :)

LucasLefevre (commented):
robodoo r+

PR #5001, opened 2024-09-19
[IMP] chart: add moving average trend line type

Rachico:
### [IMP] chart: add moving average trend line type

This commits introduces a new trend line type for chart: moving average. A moving average is the mean of the previous k data points. For each value, the mean is taken from an equal number of data points (window) preceding it.

Task: [4086645](https://www.odoo.com/odoo/project/2328/tasks/4086645)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/types/chart/chart.ts
```diff
@@ -85,12 +85,13 @@ export interface TitleDesign {
   readonly color?: Color;
 }
 
-export type TrendType = "polynomial" | "exponential" | "logarithmic";
+export type TrendType = "polynomial" | "exponential" | "logarithmic" | "movingAverage";
```
LucasLefevre, line 88: Rename the type in case we implement other types of moving averages in the future
```suggestion
export type TrendType = "polynomial" | "exponential" | "logarithmic" | "trailingMovingAverage";
```

src/components/side_panel/chart/chart_with_axis/design_panel.xml
```diff
@@ -98,8 +98,20 @@
                     <option value="logarithmic" t-att-selected="trendType === 'logarithmic'">
                       Logarithmic
                     </option>
+                    <option value="movingAverage" t-att-selected="trendType === 'movingAverage'">
+                      Moving average
```
LucasLefevre, line 102: should we also make it clear for the user ? That we he could even google it if the user wants to better understand what it is/does.
```suggestion
                      Trailing moving average
```

LucasLefevre (approved):
robodoo r+

PR #5019, opened 2024-09-24
[FIX] xlsx: convert #REF at export to xlsx

hokolomopo:
### [FIX] xlsx: convert #REF at export to xlsx:

Reference errors in excel are #REF!, not #REF.

### [FIX] parser: inconsistent handling of #REF

Depending on how we get a formula with #REF, the parser can have
different behavior even if the formula is the same.

- If we write directly =#REF (or it comes from a snapshot), the parser
will throw an error.
- If we write =A1, then delete the row 1 to get =#REF, the parser will
return a REFERENCE token with an #REF value.

This is a bit of a problem when exporting the data to xlsx: since
we parse the formulas to know whether we need to export it, we get
a different result depending on how we got the #REF.

Task: [4207052](https://www.odoo.com/web#id=4207052&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/formulas/parser.ts
+++ b/src/formulas/parser.ts
@@ -1,8 +1,7 @@
 import { DEFAULT_ERROR_MESSAGE } from "../constants";
 import { parseNumber, removeStringQuotes } from "../helpers/index";
 import { _lt } from "../translation";
-import { InvalidReferenceError } from "../types/errors";
-import { UnknownFunctionError } from "./../types/errors";
+import { CellErrorType, UnknownFunctionError } from "./../types/errors";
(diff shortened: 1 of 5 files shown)


LucasLefevre (commented):
quel échec. In master, we should change it to #REF! as well, but... upgrade script :grimacing:

LucasLefevre:
robodoo rebase-ff r+

PR #5022, opened 2024-09-24
[IMP] charts: limit trending line degree range

anhe-odoo:
## Task Description

This task aims to add a maximum value for the trending line degree (for a polynomial model), fixed to the minimum value of 10 and 1 below of the number of point in the series.

## Related Task

- Task: 4207820

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/chart/chart_with_axis/design_panel.ts
+++ b/src/components/side_panel/chart/chart_with_axis/design_panel.ts
@@ -1,7 +1,8 @@
 import { Component, useState } from "@odoo/owl";
-import { getColorsPalette, getNthColor, setColorAlpha, toHex } from "../../../../helpers";
+import { getColorsPalette, getNthColor, range, setColorAlpha, toHex } from "../../../../helpers";
 import { CHART_AXIS_CHOICES, getDefinedAxis } from "../../../../helpers/figures/charts";
 import { _t } from "../../../../translation";
+import { ChartJSRuntime } from "../../../../types/chart";
 import {
   ChartWithAxisDefinition,
   Color,
@@ -74,6 +75,10 @@ export class ChartWithAxisDesignPanel<P extends Props = Props> extends Component
 
   getDataSeries() {
     return this.props.definition.dataSets.map((d, i) => d.label ?? `${ChartTerms.Series} ${i + 1}`);
+  }
+
+  getPolynomialDegrees(): number[] {
+    return range(1, this.getMaxPolynomialDegree() + 1);
   }
 
   updateSerieEditor(ev) {
@@ -205,12 +210,7 @@ export class ChartWithAxisDesignPanel<P extends Props = Props> extends Component
 
   onChangePolynomialDegree(ev: InputEvent) {
     const element = ev.target as HTMLInputElement;
-    const order = parseInt(element.value || "1");
-    if (order < 2) {
-      element.value = `${this.getTrendLineConfiguration()?.order ?? 2}`;
-      return;
-    }
-    this.updateTrendLineValue({ order });
+    this.updateTrendLineValue({ order: parseInt(element.value) });
   }
 
   getTrendLineColor() {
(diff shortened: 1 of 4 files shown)


src/components/side_panel/chart/chart_with_axis/design_panel.ts
```diff
@@ -2,6 +2,7 @@ import { Component, useState } from "@odoo/owl";
 import { getColorsPalette, getNthColor, setColorAlpha, toHex } from "../../../../helpers";
```
hokolomopo, line 2: nitpick: in commit message

fixed to the minimum value **between** 10 and 1 **below the** number of

src/components/side_panel/chart/chart_with_axis/design_panel.ts
```diff
@@ -232,4 +238,9 @@ export class ChartWithAxisDesignPanel extends Component<Props, SpreadsheetChildE
     };
     this.props.updateChart(this.props.figureId, { dataSets });
   }
+
+  getMaxPolynomialDegree() {
+    const runtime = this.env.model.getters.getChartRuntime(this.props.figureId) as LineChartRuntime;
```
hokolomopo, line 243: Do'nt we have a genetic type ChartJSRUntime of something like that ?

src/components/side_panel/chart/chart_with_axis/design_panel.ts
```diff
@@ -207,6 +208,11 @@ export class ChartWithAxisDesignPanel extends Component<Props, SpreadsheetChildE
       element.value = `${this.getTrendLineConfiguration()?.order ?? 2}`;
       return;
     }
+    const maxOrder = this.getMaxPolynomialDegree();
+    if (order > maxOrder) {
+      element.value = `${this.getTrendLineConfiguration()?.order ?? maxOrder}`;
+      return;
```
hokolomopo, line 214: Wait so we change the value of the input, but do not call updateTrendLineValue ? Won't we have a mismatch between the chart and the input ? (same for `if (order < 2)`)

hokolomopo (commented):
tchou tchou 🚂

src/components/side_panel/chart/chart_with_axis/design_panel.ts
hokolomopo: Oh yeah we try to use `this.getTrendLineConfiguration()?.order` (which I'm gressing is from the chart definition).

Still, shouldn't we set the max/min value instead of going back to the old value ? If I write degree 40 in my input i'd expect the polynomial go go to the max degree rather that goign back to 2. Or maybe it's just me ?

src/components/side_panel/chart/chart_with_axis/design_panel.ts
```diff
@@ -76,6 +77,10 @@ export class ChartWithAxisDesignPanel<P extends Props = Props> extends Component
     return this.props.definition.dataSets.map((d, i) => d.label ?? `${ChartTerms.Series} ${i + 1}`);
   }
 
+  getPolynomialDegrees(): number[] {
+    return Array.from({ length: Math.min(10, this.getMaxPolynomialDegree()) }, (_, i) => i + 1);
```
LucasLefevre, line 81: use the `range` helper which is much more readable IMO. Also the `Math.min(10` is already part of `getMaxPolynomialDegree`)
```suggestion
    return range(1, this.getMaxPolynomialDegree() +1);
```

src/components/side_panel/chart/chart_with_axis/design_panel.ts
```diff
@@ -210,6 +215,11 @@ export class ChartWithAxisDesignPanel<P extends Props = Props> extends Component
       element.value = `${this.getTrendLineConfiguration()?.order ?? 2}`;
       return;
     }
+    const maxOrder = this.getMaxPolynomialDegree();
+    if (order > maxOrder) {
+      element.value = `${this.getTrendLineConfiguration()?.order ?? maxOrder}`;
```
LucasLefevre, line 220: Is this still needed since it's a selection input?  (the user can't enter an arbitrary value)

LucasLefevre (approved):
robodoo r+

PR #5026, opened 2024-09-27
[FIX] pivot: finer grained cycle detection

LucasLefevre:
## Description:

Steps to reproduce:

- insert a fixed pivot (not dynamic!)
- insert a new calculated measure
- in the calculated measure formula, reference one of a cell which contains a header formula

=> the cell becomes a #CYCLE even though there's no circular dependendy
   between the calculated measure and the other dimension header

Task: [4214188](https://www.odoo.com/web#id=4214188&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/functions/helper_lookup.ts
+++ b/src/functions/helper_lookup.ts
@@ -2,7 +2,7 @@ import { zoneToXc } from "../helpers";
 import { _t } from "../translation";
 import { EvalContext, FunctionResultObject, Getters, Maybe, Range, UID } from "../types";
 import { EvaluationError, InvalidReferenceError } from "../types/errors";
-import { PivotCoreDefinition } from "../types/pivot";
+import { PivotCoreDefinition, PivotCoreMeasure } from "../types/pivot";
 
 /**
  * Get the pivot ID from the formula pivot ID.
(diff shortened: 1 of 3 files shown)


src/functions/helper_lookup.ts
```diff
@@ -37,7 +37,8 @@ export function assertDomainLength(domain: Maybe<FunctionResultObject>[]) {
 
 export function addPivotDependencies(
   evalContext: EvalContext,
-  coreDefinition: PivotCoreDefinition
+  coreDefinition: PivotCoreDefinition,
+  forMeasures: PivotCoreMeasure[]
 ) {
   //TODO This function can be very costly when used with PIVOT.VALUE and PIVOT.HEADER
```
hokolomopo, line 43: unrelated: what's the state of this TODO ?

src/functions/module_lookup.ts
```diff
@@ -813,7 +817,7 @@ export const PIVOT = {
     const pivotId = getPivotId(_pivotFormulaId, this.getters);
     const pivot = this.getters.getPivot(pivotId);
     const coreDefinition = this.getters.getPivotCoreDefinition(pivotId);
-    addPivotDependencies(this, coreDefinition);
+    addPivotDependencies(this, coreDefinition, coreDefinition.measures);
```
hokolomopo, line 820: So now we can reference headers of `PIVOT.HEADER()` but not headers originating from a spread pivot formula if I understand the code correctly. Do you think there's any way we can make that work ?

LucasLefevre: > Do you think there's any way we can make that work ?

yes, but it'd be quite difficult I think. Every cell would need to be computed in the correct dependency order, when they are needed.

src/functions/helper_lookup.ts
rrahir: Think it's still a todo until we find it to be a real bottleneck ;-)

rrahir:
@robodoo r+

PR #5028, opened 2024-09-27
[IMP] chart: avoid overlapping values on charts

hokolomopo:
## Description

The option "Show Values" on charts could lead to overlapping texts in the chart, which was quite ugly.

This commit improve the behaviour or the value labels:
- prevent overlapping value labels on stacked charts
- the bar chart value labels are now in the middle of the bar
- the value labels have now the same color as the dataset
- prevent value labels from going outside of the chart

Code-wise, the commit totally splits the handling of bar and horizontal bar charts. It make the code slightly longer, but avoid having to write a ternary for each computation of x/y position, which ultimately makes the code more readable.

Task: [4207889](https://www.odoo.com/web#id=4207889&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/figures/chart/chartJs/chartjs_show_values_plugin.ts
+++ b/src/components/figures/chart/chartJs/chartjs_show_values_plugin.ts
@@ -1,4 +1,5 @@
 import { ChartType, Plugin } from "chart.js";
+import { computeTextWidth } from "../../../../helpers";
 import {
   TREND_LINE_XAXIS_ID,
   chartFontColor,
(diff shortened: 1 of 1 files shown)


src/components/figures/chart/chartJs/chartjs_show_values_plugin.ts
```diff
@@ -29,74 +30,142 @@ export const chartShowValuesPlugin: Plugin = {
...
         }
       }
-    });
+      textsPositions[yPosition].push(xPosition);
 
-    ctx.restore();
-  },
-};
+      ctx.fillStyle = point.options.backgroundColor;
+      ctx.strokeStyle = options.background || "#ffffff";
```
rrahir, line 144: Not 100% sure about this one, if you have a very light grey dataset color and no background, the text value becomes unreadable. Maybe having the stroke computed based on `point.options.backgroundColor` would be better to ensure the text is visible?
fyi, there are 2 datasets in the picture :p
image

src/components/figures/chart/chartJs/chartjs_show_values_plugin.ts
```diff
@@ -29,74 +30,142 @@ export const chartShowValuesPlugin: Plugin = {
     if (!drawData) {
       return;
     }
-    const ctx = chart.ctx;
+    const ctx = chart.ctx as CanvasRenderingContext2D;
```
rrahir, line 33: think its not necessary.

rrahir (commented):
:+1: 

It might be worth targeting 18.0, maybe?

src/components/figures/chart/chartJs/chartjs_show_values_plugin.ts
```diff
@@ -29,74 +30,142 @@ export const chartShowValuesPlugin: Plugin = {
...
-            ctx.fillStyle = chartFontColor(bar.options.backgroundColor);
-            ctx.strokeStyle = chartFontColor(ctx.fillStyle);
-
-            const value = options.callback(dataset._parsed[i]);
-            ctx.strokeText(value, x, y);
-            ctx.fillText(value, x, y);
-          }
-          break;
+      for (const otherPosition of textsPositions[xPosition] || []) {
+        if (Math.abs(otherPosition - yPosition) < 13) {
```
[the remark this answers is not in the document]
rrahir: ok I just realized we already do this kind of pixel correction in the PieCharts ...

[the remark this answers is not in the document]
hokolomopo: In this particular case it should be fine no matter the scale of the chart, beaucase the font sizes are not scaled. Ideally i should measure the text height, but it's more costly/annoying

src/components/figures/chart/chartJs/chartjs_show_values_plugin.ts
hokolomopo: Not strictly necessary, but helpful when coding to have VSCode suggest the different canvas context's methods

src/components/figures/chart/chartJs/chartjs_show_values_plugin.ts
```diff
@@ -29,74 +30,142 @@ export const chartShowValuesPlugin: Plugin = {
...
+
+    for (let i = 0; i < dataset._parsed.length; i++) {
+      const value = dataset._parsed[i].y;
+      const point = dataset.data[i];
+
+      const xPosition = point.x;
+
+      let yPosition = 0;
+      if (chart.config.type === "line") {
+        yPosition = point.y - 10;
```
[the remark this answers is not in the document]
hokolomopo: text should be a bit aboce the data point for line chart, no ?

src/components/figures/chart/chartJs/chartjs_show_values_plugin.ts
hokolomopo: Having the text stroke be the same color as the background is _ususally_ prettier. Here for example if we do a white or black text bacgrkound it's not as pretty : 
image
image

But indeed there are cases where it'll be ugly.  I was thinking that if the user pick shitty colors that's his problem, GSheet does the same: 
image

I'll ask Francois

hokolomopo: image

hokolomopo:
> 👍
> 
> It might be worth targeting 18.0, maybe?

Yeah I wasn't sure what to target. I'd say either 17.4 where we introduced the feature or master

src/components/figures/chart/chartJs/chartjs_show_values_plugin.ts
rrahir: didn't realize it was types as nay, my bad  :+1:

src/components/figures/chart/chartJs/chartjs_show_values_plugin.ts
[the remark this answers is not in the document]
rrahir: it's more the arbitrary value of 10 that ticked me but it's the same as my remark below and your point makes sense.

hokolomopo:
backported to 16.4 https://github.com/odoo/o-spreadsheet/pull/5099

PR #5029, opened 2024-09-27
[PERF] Actions: Fill missing ids with fast unique values

rrahir:
The ActionsSpecs have a defaut behaviour where we are not obligated to provide an id in the actionSpec because "CreateAction" will end up generating it on demand with an uuidv4.

However, uuidv4 can become a problem if invoked to often and it so happens that some actionSpecs related to the topbar benefit from this automatic generation. Since the topbar items are regenerated at each rendering of a spreadsheet, they can become problematic.

For instance, in Odoo test suites, they take up to 2-3% of the computation time.

This revision bypasses the call to an actual uuidv4 by forcing a fast strategy in the uuidGenerator, which accomplishes the same result, having a unique identifier for each menu item generated.

Task: 4216427

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/actions/action.ts
+++ b/src/actions/action.ts
@@ -1,4 +1,3 @@
-import { UuidGenerator } from "../helpers";
 import { Color, SpreadsheetChildEnv } from "../types";
 
 /*
@@ -91,7 +90,7 @@ export function createActions(menuItems: ActionSpec[]): Action[] {
   return menuItems.map(createAction).sort((a, b) => a.sequence - b.sequence);
 }
 
-const uuidGenerator = new UuidGenerator();
+let nextItemId = 1;
 
 export function createAction(item: ActionSpec): Action {
   const name = item.name;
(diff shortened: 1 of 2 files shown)


src/actions/action.ts
```diff
@@ -92,6 +92,7 @@ export function createActions(menuItems: ActionSpec[]): Action[] {
 }
 
 const uuidGenerator = new UuidGenerator();
+uuidGenerator.setIsFastStrategy(true);
```
LucasLefevre, line 95: let's just use a incremented number here then. Instead of instantiating a `UuidGenerator` for no apparent reason.
Or it's the beginning of the [enterprise fizzbuzz](https://github.com/EnterpriseQualityCoding/FizzBuzzEnterpriseEdition) :smile:

LucasLefevre (commented):
:+1: 

I'm wondering how high the number can get with all the repeated renders though. And if there's even a remote chance of overflowing

rrahir:
> I'm wondering how high the number can get with all the repeated renders though. And if there's even a remote chance of overflowing

I was asking myself the same question, I think that for the time being it's quite controlled. each render was below 10 calls to UUIDGenerator and seeing the value of [max integer](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER). If it starts growing too much at some point, we can always force the id later on. As I said, i don't think it's too much of a bother.

hokolomopo (commented):
@robodoo r+

PR #5030, opened 2024-09-27
[REM] UUID: remove fast strategy

rrahir:
THe fast strategy for uuid generator was introduced to speed up the startup of a model, specifically when generating cells which could be numerous and the calls to crypto could become a hassle. However, we realised later on that a uuid was useless for that specific purpose. More specifically, a uuid is useful to generate unique identifiers that will be shared accross users to avoid collisions in multi-user context (i.e. concurrent updates).
In other cases, shortcutting the uuid generator to become a simple incremented integer value is either wrong or shows the uselessness of calling uuidGenerator in the first place.

Task: 4216816

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/uuid.ts
+++ b/src/helpers/uuid.ts
@@ -3,20 +3,9 @@
  * */
 
 export class UuidGenerator {
-  private isFastIdStrategy = false;
-
-  private fastIdStart = 0;
-
-  setIsFastStrategy(isFast: boolean) {
-    this.isFastIdStrategy = isFast;
-  }
-
   uuidv4(): string {
-    if (this.isFastIdStrategy) {
-      this.fastIdStart++;
-      return String(this.fastIdStart);
-      //@ts-ignore
-    } else if (window.crypto && window.crypto.getRandomValues) {
+    //@ts-ignore
+    if (window.crypto && window.crypto.getRandomValues) {
       //@ts-ignore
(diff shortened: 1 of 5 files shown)


src/helpers/uuid.ts
```diff
@@ -3,20 +3,9 @@
...
-    this.isFastIdStrategy = isFast;
-  }
-
   uuidv4(): string {
-    if (this.isFastIdStrategy) {
-      this.fastIdStart++;
-      return String(this.fastIdStart);
-      //@ts-ignore
-    } else if (window.crypto && window.crypto.getRandomValues) {
+    //@ts-ignore
```
hokolomopo, line 7: I think this ts-ignore is useless

rrahir: very much useful becasue ts won't let you sum an array of numbers with a number, even though of course it works in real life conditions ^^

hokolomopo: Was taI was talking about the ts-ignore on `if (window.crypto && window.crypto.getRandomValues)` :upside_down_face:

LucasLefevre (approved):
:+1: 

I pushed an additional commit to remove `this.uuidGenerator` from core plugins to discourage the use of uuids.
What do you think?

VincentSchippefilt:
robodoo r+

LucasLefevre:
robodoo rebase-ff

PR #5045, opened 2024-09-30
[FIX] pivot: remove forbidden chars when duplicating

LucasLefevre:
## Description:

Steps to reproduce:

- insert a new pivot
- rename the pivot with one of the following character in its name: ', *, ?, /, \, [, ]
- duplicate the pivot

=> boom

The command to create the new sheet is rejected (the name contains invalid characters). But we later tries to activate the sheet (which doesn doesn't exist since it was never created)

Task: [4221325](https://www.odoo.com/web#id=4221325&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_feature/insert_pivot.ts
+++ b/src/plugins/ui_feature/insert_pivot.ts
@@ -1,4 +1,4 @@
-import { PIVOT_TABLE_CONFIG } from "../../constants";
+import { FORBIDDEN_IN_EXCEL_REGEX, PIVOT_TABLE_CONFIG } from "../../constants";
 import { SpreadsheetPivotTable } from "../../helpers/pivot/table_spreadsheet_pivot";
 import { getZoneArea } from "../../helpers/zones";
 import { _t } from "../../translation";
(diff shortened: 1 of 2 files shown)


src/plugins/ui_feature/insert_pivot.ts
```diff
@@ -99,7 +101,7 @@ export class InsertPivotPlugin extends UIPlugin {
       name = `${pivotName} (${i})`;
       i++;
     }
-    return name;
+    return name.replace(new RegExp(FORBIDDEN_IN_EXCEL_REGEX, "g"), " ");
```
pro-odoo, line 104:
```suggestion
    let name = pivotName.replace(new RegExp(FORBIDDEN_IN_EXCEL_REGEX, "g"), " ");
    while (names.includes(name)) {
      name = `${pivotName} (${i})`;
      i++;
    }
    return name;
```

We could have a collision if the replacement is done after the check of existing sheet

LucasLefevre: Good catch !

Fixed and test added :)

pro-odoo (approved):
robodoo r+

PR #5050, opened 2024-10-01
[IMP] cf: apply panel update immediately

hokolomopo:
## Description

The previous workflow of the CF side panel was to edit the conditional format, then click on save to confirm. This wasn't the greatest, as it meant that you wouldn't see the result of your changes until you saved. And if you didn't like the change, you then needed to re-click on the CF to re-open its editor, and do the changes.

With this commit, the CF side panel now updates immediately when you change a value. This also mean that a new CF is created immediately when you click on the "Add" button. The changes made can be discarded by clicking on the "Cancel" button.

Task: [3720877](https://www.odoo.com/web#id=3720877&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/conditional_formatting/cf_editor/cf_editor.ts
+++ b/src/components/side_panel/conditional_formatting/cf_editor/cf_editor.ts
@@ -20,9 +20,12 @@ import {
   CommandResult,
   ConditionalFormat,
   ConditionalFormatRule,
+  ConditionalFormattingOperatorValues,
   DataBarRule,
   IconSetRule,
+  IconThreshold,
   SpreadsheetChildEnv,
+  ThresholdType,
 } from "../../../../types";
 import { ColorPickerWidget } from "../../../color_picker/color_picker_widget";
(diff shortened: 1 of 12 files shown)


src/components/side_panel/conditional_formatting/cf_editor/cf_editor.ts
```diff
@@ -280,6 +272,13 @@ export class ConditionalFormattingEditor extends Component<Props, SpreadsheetChi
     }
   }
 
+  onSave() {
```
LucasLefevre, line 275: Change something, let's say a color (which dispatches), then click on save: the exact same command is dispatched (and accepted). We could add an `allowDispatch` to rejects `ADD_CONDITIONAL_FORMAT` commands that wouldn't change anything to the CF

LucasLefevre (approved):
robodoo r+

PR #5056, opened 2024-10-07
[FIX] side_panel: align font weight with other side panels

dhrp-odoo:
## Description:

The font weight of this side panel was previously inconsistent with other side panels. This commit ensures that the font weight is now uniform across all side panels.

Task: [4194259](https://www.odoo.com/web#id=4194259&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/custom_currency/custom_currency.xml
+++ b/src/components/side_panel/custom_currency/custom_currency.xml
@@ -62,7 +62,7 @@
           <table>
             <t t-foreach="getFormatExamples()" t-as="example" t-key="example_index">
               <tr>
-                <td class="pe-3 fw-bolder" t-esc="example.label"/>
+                <td class="pe-3 o-fw-bold" t-esc="example.label"/>
                 <td t-esc="example.value"/>
               </tr>
             </t>
--- a/src/components/side_panel/components/collapsible/side_panel_collapsible.ts
+++ b/src/components/side_panel/components/collapsible/side_panel_collapsible.ts
@@ -4,7 +4,6 @@ import { css } from "../../../helpers";
 css/* scss */ `
   .o_side_panel_collapsible_title {
     font-size: 16px;
-    font-weight: bold;
     cursor: pointer;
     padding: 6px 0px 6px 6px !important;
 
--- a/src/components/side_panel/components/collapsible/side_panel_collapsible.xml
+++ b/src/components/side_panel/components/collapsible/side_panel_collapsible.xml
@@ -1,7 +1,7 @@
 <templates>
   <t t-name="o-spreadsheet-SidePanelCollapsible">
     <div class="" t-att-class="props.class">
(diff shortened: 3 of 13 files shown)


src/components/side_panel/custom_currency/custom_currency.xml
```diff
@@ -62,7 +62,7 @@
           <table>
             <t t-foreach="getFormatExamples()" t-as="example" t-key="example_index">
               <tr>
-                <td class="pe-3 fw-bolder" t-esc="example.label"/>
+                <td class="pe-3 fw-semibold" t-esc="example.label"/>
```
hokolomopo, line 65: I think I would avoid bootstrap classes altogether. If I was correct when writing the spec, some bootstrap classes had different font weight in odoo and in o-spreadsheet. Maybe `fw-medium` and `fw-semibold` don't have the same issue, but there's no garantee that is stay as it.

I think I would create a new class `o-fw-bold` and apply it everywhere. The class could use a constant common with `o_side_panel_collapsible_title` and `o-sidePanel-title` so we are sure to have the same font wight everywhere.

What do you think ?

dhrp-odoo: Yes, in Odoo, fw-bold applies a weight of 500, while in o-spreadsheet, it's set to 700.

If I understand correctly, Need to create a single class with a weight of 500 and then use that variable for elements like `o-sidePanelTitle`, `o_side_panel_collapsible_title`, `o-section-title`, `o-section-subtitle` ...

hokolomopo: Yeah I think the simplest if to create a constant `PANEL_BOLD_FONT_WEGHT` or something that would be used inside the CSS definition of `o-sidePanelTitle` and the other. For where we used bootstrap before and don't have a specific CSS class, I would create a `o-fw-bold` CSS class that use the font weight constant

rrahir: SInce you use o-fw-bold pretty much everywhere, I would add the class explicitely where it's required instead of adding a font-weight rule to the existing classes (like for `.o_side_panel_collapsible_title`)

dhrp-odoo: Changes made!

I believe we should maintain the font-weight rule for classes like `o-section-title` and `o-section-subtitle`, as they are used in multiple places and require consistent font weight throughout. What do you think?

hokolomopo: I 100% agree with @dhrp-odoo , I really don't want to have to copy/paste 5 different classes every time I want to create a `o-section-subtitle`

hokolomopo:
@robodoo r+

PR #5059, opened 2024-10-08
[FIX] conditional format: huge revisions on copy/paste cf

hokolomopo:
When copy/pasting a conditional format, the revision had one ADD_CONDITIONAL_FORMAT command per copied cell. That lead to huge revisions if, for example, a whole column with a CF was copied.

Now the revision only has a single ADD_CONDITIONAL_FORMAT command per modified CF.

The exact same work was done for data validation rules.

Benchmark:
-------------------
Copy/paste a column of 100 cells with a CF:
Before: 46.6KB revision
After: 526B revision (divided by 88!)

Task: [4240668](https://www.odoo.com/web#id=4240668&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/clipboard_handlers/conditional_format_clipboard.ts
+++ b/src/clipboard_handlers/conditional_format_clipboard.ts
@@ -26,6 +26,8 @@ export class ConditionalFormatClipboardHandler extends AbstractCellClipboardHand
   Maybe<ClipboardConditionalFormat>
 > {
   private readonly uuidGenerator = new UuidGenerator();
+  private queuedChanges: Record<UID, { toAdd: Zone[]; toRemove: Zone[]; cf: ConditionalFormat }[]> =
+    {};
 
   copy(data: ClipboardCellData): ClipboardContent | undefined {
     if (!data.zones.length) {
@@ -51,6 +53,7 @@ export class ConditionalFormatClipboardHandler extends AbstractCellClipboardHand
   }
 
   paste(target: ClipboardPasteTarget, clippedContent: ClipboardContent, options: ClipboardOptions) {
+    this.queuedChanges = {};
     if (options.pasteOption === "asValue") {
       return;
     }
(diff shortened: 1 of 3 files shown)


rrahir:
The commit title should refer to clipboard and not just CF since you also fix data validations. 

Also this seems to be the root cause why we have a weird order of clipboard handlers https://github.com/odoo/o-spreadsheet/blob/18.0/src/clipboard_handlers/index.ts#L23-L24. So might be worth backport this to saas-17.2?

src/clipboard_handlers/conditional_format_clipboard.ts
```diff
@@ -114,30 +119,55 @@ export class ConditionalFormatClipboardHandler extends AbstractCellClipboardHand
...
-    return cfInTarget ? cfInTarget : { ...originCF, id: this.uuidGenerator.uuidv4(), ranges: [] };
+    const queuedCfs = this.queuedChanges[targetSheetId];
+    if (!targetCF && queuedCfs) {
+      targetCF = queuedCfs.find(
+        (queued) =>
+          queued.cf.stopIfTrue === originCF.stopIfTrue && deepEquals(queued.cf.rule, originCF.rule)
+      )?.cf;
+    }
+
+    return targetCF ? targetCF : { ...originCF, id: this.uuidGenerator.uuidv4(), ranges: [] };
```
rrahir, line 171:
```suggestion
    return targetCF || { ...originCF, id: this.uuidGenerator.uuidv4(), ranges: [] };
```

src/clipboard_handlers/data_validation_clipboard.ts
```diff
@@ -119,16 +125,25 @@ export class DataValidationClipboardHandler extends AbstractCellClipboardHandler
...
+    if (!targetRule && queuedRules) {
+      targetRule = queuedRules.find(
+        (queued) =>
+          deepEquals(originRule.criterion, queued.rule.criterion) &&
+          originRule.isBlocking === queued.rule.isBlocking
+      )?.rule;
+    }
+
+    return targetRule
+      ? targetRule
```
rrahir, line 146:
```suggestion
      ||
```

rrahir (commented):
top top 👍

PR #5062, opened 2024-10-08
[FIX] table: insert table without filters

LucasLefevre:
## Description:

When inserting a table, it is inserted with the filter. It takes place on the cell and is only useful in a handful of use cases. It's very frequent that the very next step after inserting a table is to remove it.

So let's insert the table without the filter by default.

Feedback from RNG

Task: [4235172](https://www.odoo.com/web#id=4235172&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/migrations/migration_steps.ts
+++ b/src/migrations/migration_steps.ts
@@ -6,6 +6,7 @@ import {
 import { getItemId } from "../helpers";
 import { toXC } from "../helpers/coordinates";
 import { getMaxObjectId } from "../helpers/pivot/pivot_helpers";
+import { DEFAULT_TABLE_CONFIG } from "../helpers/table_presets";
 import { overlap, toZone, zoneToXc } from "../helpers/zones";
 import { Registry } from "../registries/registry";
 import { CustomizedDataSet, DEFAULT_LOCALE, Format, WorkbookData, Zone } from "../types";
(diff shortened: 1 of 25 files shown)


src/migrations/migration_steps.ts
```diff
@@ -420,6 +421,20 @@ migrationStepRegistry
       }
       return data;
     },
+  })
+  .add("migration_22", {
```
hokolomopo, line 425: If you want to do that you'd also want to create a migration for the revisions. 

TBH I wouldn't bother, the tables will lose their filters it's no big deal.

hokolomopo (commented):
That's a lot of test changes for a single boolean change in a constant 😳

src/migrations/migration_steps.ts
rrahir: Yup; either both the revision and data migration or none of them.
on one hand, the filtering is reset once you reload your sheet so you could drop them
on the other hand, if the main user only has read-only access to the sheet, it will piss them off...
I'd argue for adding a command migration :/

Why did the command had an optional config to begin with?

LucasLefevre: There's no need for a command migration, `"hasFilters":true` is part of the command :wink:

LucasLefevre: fuck, it's not part of the command when used from the menu :grimacing:

LucasLefevre:
done in master because it requires an upgrade script for revisions

src/migrations/migration_steps.ts
hokolomopo: > Why did the command had an optional config to begin with?

I believe my idea was to save space/network usage by having smaller revisions since most of the CREATE_TABLE would have the exact same DEFAULT_CONFIG in the payload. I also don't put it in the exported json. I'm saving the trees 😤

PR #5063, opened 2024-10-08
[FIX] clipboard: Fix clipboard cross-browser coverage

rrahir:
The current implementation of the cross-browser clipboard relies heavily on an API that is currently only available on Chromium based browsers (see [table](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API#api.clipboarditem)).

There is currently no definitive announce that the feature will be adopted by other browsers (namely FF and Safari) and since the feature is still marked as experimental, it'd be better to rely on a more generic approach.

This revision changes the flow to rely entirely on the `text/html` mimetype as it is supported by all modern browsers.

Task: 4241877

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/clipboard/clipboard_helpers.ts
+++ b/src/helpers/clipboard/clipboard_helpers.ts
@@ -1,4 +1,11 @@
-import { ClipboardCellData, UID, Zone } from "../../types";
+import {
+  ClipboardCellData,
+  ClipboardMIMEType,
+  OSClipboardContent,
+  ParsedOSClipboardContent,
+  UID,
+  Zone,
+} from "../../types";
 import { mergeOverlappingZones, positions } from "../zones";
 
 export function getClipboardDataPositions(sheetId: UID, zones: Zone[]): ClipboardCellData {
@@ -53,4 +60,24 @@ export function getPasteZones<T>(target: Zone[], content: T[][]): Zone[] {
   const width = content[0].length,
     height = content.length;
   return target.map((t) => splitZoneForPaste(t, width, height)).flat();
+}
+
+export function parseOSClipboardContent(content: OSClipboardContent): ParsedOSClipboardContent {
+  if (!content[ClipboardMIMEType.Html]) {
+    return {
+      text: content[ClipboardMIMEType.PlainText],
+    };
+  }
+  const htmlDocument = new DOMParser().parseFromString(
+    content[ClipboardMIMEType.Html],
+    "text/html"
+  );
+  const oSheetClipboardData = htmlDocument
+    .querySelector("div")
+    ?.getAttribute("data-osheet-clipboard");
+  const spreadsheetContent = oSheetClipboardData && JSON.parse(oSheetClipboardData);
+  return {
+    text: content[ClipboardMIMEType.PlainText],
+    data: spreadsheetContent,
+  };
 }
--- a/src/plugins/ui_stateful/clipboard.ts
+++ b/src/plugins/ui_stateful/clipboard.ts
@@ -6,11 +6,11 @@ import { getClipboardDataPositions } from "../../helpers/clipboard/clipboard_hel
 import { UuidGenerator, isZoneValid, union } from "../../helpers/index";
 import { CURRENT_VERSION } from "../../migrations/data";
 import {
-  ClipboardContent,
   ClipboardData,
   ClipboardMIMEType,
   ClipboardOptions,
   ClipboardPasteTarget,
+  OSClipboardContent,
 } from "../../types/clipboard";
 import {
   ClipboardCell,
@@ -39,6 +39,11 @@ type MinimalClipboardData = {
   figureId?: UID;
   [key: string]: unknown;
 };
+
+export interface SpreadsheetClipboardData extends MinimalClipboardData {
+  version?: number;
+  clipboardId?: string;
+}
 /**
  * Clipboard Plugin
  *
@@ -70,9 +75,7 @@ export class ClipboardPlugin extends UIPlugin {
         const zones = this.getters.getSelectedZones();
         return this.isCutAllowedOn(zones);
       case "PASTE_FROM_OS_CLIPBOARD": {
-        const copiedData = this.convertOSClipboardData(
-          cmd.clipboardContent[ClipboardMIMEType.PlainText] ?? ""
-        );
+        const copiedData = this.convertTextToClipboardData(cmd.clipboardContent.text ?? "");
         const pasteOption = cmd.pasteOption;
         return this.isPasteAllowed(cmd.target, copiedData, { pasteOption, isCutOperation: false });
       }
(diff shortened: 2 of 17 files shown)


src/helpers/clipboard/clipboard_helpers.ts
```diff
@@ -54,3 +61,20 @@ export function getPasteZones<T>(target: Zone[], content: T[][]): Zone[] {
     height = content.length;
   return target.map((t) => splitZoneForPaste(t, width, height)).flat();
 }
+
+export function gePastablePluginClipBoardContent(
```
rrahir, line 65: pls help. As a wise Lucas once said: "I love pasta but not in my code".

So yeah I wanted to make a difference between the osclipboard content  {key:string} and the useful stuff, that is the html already parsed. It seemed better to only have to invoke a domparser once and  pass around a typed object instead of a string.

src/helpers/clipboard/clipboard_helpers.ts
```diff
@@ -54,3 +61,20 @@ export function getPasteZones<T>(target: Zone[], content: T[][]): Zone[] {
...
+    content[ClipboardMIMEType.Html] ?? "<div></div>",
+    "text/html"
+  );
+  const oSheetClipboardData = htmlDocument
+    .querySelector("div")
+    ?.getAttribute("data-osheet-clipboard");
+  const spreadsheetContent = oSheetClipboardData && JSON.parse(oSheetClipboardData);
+  return {
+    [ClipboardMIMEType.PlainText]: content[ClipboardMIMEType.PlainText],
+    [ClipboardMIMEType.Html]: spreadsheetContent,
```
hokolomopo, line 78: I would change the HTML tag for somethnin else at this point, it's juste confusing to reference HTML content when it's just a JSON.

Also we might wanna use the actual HTML in the future, if we want to support copying a table from some webpage or things like that

src/plugins/ui_stateful/clipboard.ts
```diff
@@ -518,36 +524,35 @@ export class ClipboardPlugin extends UIPlugin {
...
+        htmlTable += "</tr>";
       }
-      htmlTable += "</tr>";
+      htmlTable += "</table>";
+      innerHTML = htmlTable;
     }
-    htmlTable += "</table></div>";
-    return htmlTable;
+    const serializedData = JSON.stringify(this.getgridData());
+    return `<div data-osheet-clipboard='${serializedData}'>${innerHTML}</div>`;
```
hokolomopo, line 555: I'm 95% sure this will break the HTML. Amprsand, quotes and other stuff need to be escaped at the very least

tests/clipboard/clipboard_plugin.test.ts
```diff
@@ -556,7 +561,10 @@ describe("clipboard", () => {
       setCellContent(model, "A2", "3");
       copy(model, "A1:B2");
       const htmlContent = model.getters.getClipboardContent()[ClipboardMIMEType.Html]!;
-      const expectedHtmlContent = `<div data-clipboard-id="${model.getters.getClipboardId()}"><table border="1" style="border-collapse:collapse"><tr><td style="">1</td><td style="">2</td></tr><tr><td style="">3</td><td style=""></td></tr></table></div>`;
+      const cbPlugin = getPlugin(model, ClipboardPlugin);
+      //@ts-ignore
+      const clipboardData = JSON.stringify(cbPlugin.getgridData());
```
hokolomopo, line 566: nitpick: IMO it's prettier to write `cbPlugin["getgridData"]()` instead of a ts-ignore if we're dead set on totally ignoring typing, ts-ignore are ugly 😞

src/helpers/clipboard/clipboard_helpers.ts
```diff
@@ -54,3 +61,20 @@ export function getPasteZones<T>(target: Zone[], content: T[][]): Zone[] {
...
+export function gePastablePluginClipBoardContent(
+  content: OSClipboardContent
+): ImportClipboardContent {
+  const htmlDocument = new DOMParser().parseFromString(
+    content[ClipboardMIMEType.Html] ?? "<div></div>",
+    "text/html"
+  );
+  const oSheetClipboardData = htmlDocument
+    .querySelector("div")
+    ?.getAttribute("data-osheet-clipboard");
```
hokolomopo, line 74: Just curious: are there any pro/cons of parsing the HTML rather than using a regex ? And of putting the data in an attribute rather than a HTML comment ?

hokolomopo (commented):
pouet pouet

src/helpers/clipboard/clipboard_helpers.ts
rrahir: yup I agree, I'm trying to come up with a better approach as it gets even funkier with the images

rrahir: note to self: change for text, data (and in the next PR, add a key image)

src/plugins/ui_stateful/clipboard.ts
```diff
@@ -476,26 +479,25 @@ export class ClipboardPlugin extends UIPlugin {
...
+  getClipboardContent(): OSClipboardContent {
     return {
       [ClipboardMIMEType.PlainText]: this.getPlainTextContent(),
       [ClipboardMIMEType.Html]: this.getHTMLContent(),
-      [ClipboardMIMEType.OSpreadsheet]: this.getSerializedGridData(),
     };
   }
 
-  private getSerializedGridData(): string {
+  private getgridData(): HtmlClipboardData {
```
hokolomopo, line 489:
```suggestion
  private getGridData(): HtmlClipboardData {
```
that's the name we had before so I'm fine if you want to keep it, but why grid ? What does the clipboard plugin knows about the grid ?

src/types/clipboard.ts
```diff
@@ -1,12 +1,17 @@
...
 export enum ClipboardMIMEType {
   PlainText = "text/plain",
   Html = "text/html",
-  OSpreadsheet = "web application/o-spreadsheet",
 }
 
-export type ClipboardContent = { [type in ClipboardMIMEType]?: string };
+export type OSClipboardContent = { [type in ClipboardMIMEType]?: string };
+
+export type ImportClipboardContent = {
```
hokolomopo, line 11:
```suggestion
export type ParsedOSClipboardContent = {
```
maybe ?

hokolomopo (commented):
A bunch of naming comments :see_no_evil: 

Otherwise looks good !

src/plugins/ui_stateful/clipboard.ts
rrahir: At first, the clipboard would return a table representing the values inside the sheet "grid". grid is not explicit in this case as it clearly stems from the visual representation, sheetData might be better?

hokolomopo:
Fixed typo in commit message

@robodoo r+

PR #5065, opened 2024-10-08
[IMP] pivot: allow to open side panel in readonly mode

LucasLefevre:
## Description:

In read-only mode, you can't open the pivots/lists/charts side panels. It's useful to know what you're looking at (range, domain, sorting, etc.), even if you can't edit anything.

Task: [4187705](https://www.odoo.com/web#id=4187705&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/tests/pivots/pivot_side_panel.test.ts
+++ b/tests/pivots/pivot_side_panel.test.ts
@@ -25,6 +25,19 @@ describe("Pivot side panel", () => {
     addPivot(model, "A1:B2", {}, "2");
   });
 
+  test("readonly panel is not clickable and greyed", async () => {
+    env.openSidePanel("PivotSidePanel", { pivotId: "2" });
+    await nextTick();
+    const panel = fixture.querySelector(".o-sidePanelBody > div");
+    expect(panel?.classList).not.toContain("pe-none");
+    expect(panel?.classList).not.toContain("opacity-50");
+    model.updateMode("readonly");
+    await nextTick();
+    expect(panel?.classList).toContain("pe-none");
+    expect(panel?.classList).toContain("opacity-50");
+    expect(panel?.getAttribute("inert")).toBe("1");
+  });
+
   test("It should open the pivot editor when pivotId is provided", async () => {
     env.openSidePanel("PivotSidePanel", { pivotId: "2" });
     await nextTick();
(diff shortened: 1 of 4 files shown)


tests/pivots/pivot_side_panel.test.ts
```diff
@@ -19,6 +19,13 @@ describe("Pivot side panel", () => {
     addPivot(model, "A1:D5", {}, "2");
   });
 
+  test("readonly panel snapshot", async () => {
```
hokolomopo, line 22: IMO snapshot tests are mostly useless, i'ts better to have actual testing of what's in the DOM

src/components/side_panel/pivot/pivot_layout_configurator/pivot_dimension/pivot_dimension.xml
```diff
@@ -20,7 +21,7 @@
           <t t-slot="upper-right-icons"/>
           <i
             class="o-button-icon fa fa-trash pe-1 ps-2"
-            t-if="props.onRemoved"
+            t-if="props.onRemoved and !isReadonly"
```
hokolomopo, line 24: Having to add an isReadonly condition on _every_ present and future button sounds like a nightmare and not somehting we want to do. You already forgot a button (hide/show measure), and to be complete we should be able to open the "Show measure as" side panel, but not to edit it.

Is it not possible to do something simpler, like a gray overlay above the panel that would stop all mouse events ?

hokolomopo (commented):
Havign `isReadOnly` everywhere doesn't sounds fun :disappointed:

tests/pivots/pivot_side_panel.test.ts
```diff
@@ -19,6 +19,18 @@ describe("Pivot side panel", () => {
...
```
hokolomopo, line 22: you forgot to change the test name 🙃

LucasLefevre: unacceptable !

pro-odoo (approved):
robodoo r+

PR #5069, opened 2024-10-09
[FIX] charts: fix trend line for datetime

anhe-odoo:
## Task Description

This commit aims to fix a wrong computation for trend line when using date (or datetime) as label. The issue can be reproduced following this:

1. Create a new spreadsheet
2. In A1:A4, enter 8/1, 9/1, 10/1 and 12/1
3. In B1:B4, enter 1, 4, 9 and 25
4. Create a new chart with A1:A4 as labels and B1:B4 as data series
5. Add a polynomial trend line of degree 2

We should have a perfect fit, as we have a quadradic progression with the month number, but it's not the case

## Related Task:

Task: 0

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


LucasLefevre (commented):
robodoo r+ rebase-ff

PR #5070, opened 2024-10-09
[FIX] CoreViewPlugin: Remove access to dispatch

rrahir:
The CoreView plugins have a local state that is directly derived from the core data. As such, they should not have any impact on plugins other than themselves, especially not on core plugins.

This revision removes the dispatch fro the core view plugins altogether as they don't and should not dispatch anything.

Task: 4241403

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/model.ts
+++ b/src/model.ts
@@ -15,6 +15,11 @@ import { BasePlugin } from "./plugins/base_plugin";
 import { RangeAdapter } from "./plugins/core/range";
 import { CorePlugin, CorePluginConfig, CorePluginConstructor } from "./plugins/core_plugin";
 import {
+  CoreViewPlugin,
+  CoreViewPluginConfig,
+  CoreViewPluginConstructor,
+} from "./plugins/core_view_plugin";
+import {
   corePluginRegistry,
   coreViewsPluginRegistry,
   featurePluginRegistry,
@@ -128,7 +133,7 @@ export class Model extends EventBus<any> implements CommandDispatcher {
 
   private statefulUIPlugins: UIPlugin[] = [];
 
-  private coreViewsPlugins: UIPlugin[] = [];
+  private coreViewsPlugins: CoreViewPlugin[] = [];
 
   private range: RangeAdapter;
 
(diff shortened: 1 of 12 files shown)


src/model.ts
```diff
@@ -14,6 +14,11 @@ import {
 import { BasePlugin } from "./plugins/base_plugin";
 import { RangeAdapter } from "./plugins/core/range";
 import { CorePlugin, CorePluginConfig, CorePluginConstructor } from "./plugins/core_plugin";
+import {
```
hokolomopo, line 17: typo in commit msg `dispatch fro the core view`

hokolomopo (commented):
That seems like  a breaking change so I would do it in master.

Wouldn't it cause problem with pivots/data source ? They need to dispatch EVALUATE_ALL_CELLS and aren't they coreView plugins? And in general removing the dispatch from the CoreView plugin sounds good in paper, but I think there might be relevant use case where a core_view would want to diapatch an UI command.

rrahir:
> That seems like a breaking change so I would do it in master.

It's already used in bad conditions, so removing it and fixing whatever uses it seems like a good approach - of course assuming no one is relying on this outside of Odoo. I however proposed https://github.com/odoo/o-spreadsheet/pull/5068 which fixed without "breaking".


> Wouldn't it cause problem with pivots/data source ? They need to dispatch EVALUATE_ALL_CELLS and aren't they coreView plugins? And in general removing the dispatch from the CoreView plugin sounds good in paper, but I think there might be relevant use case where a core_view would want to dispatch an UI command.

the datasource-related plugins don't dispatch anything, it's the datasource itself that sends a message to the their bus which in turn dispatches a global evaluation.
By blocking the access to dispatch to the coreView plugins we merely force the creation of 'dummy' ui plugins to handle the potential dispatches in order to prevent wrong and quite difficult to detect behaviours.

rrahir:
This could bring regressions and break existing code -> done in master in #5216

PR #5091, opened 2024-10-14
[IMP] xlsx: support import/export of data validation rules

Rachico:
This commit enhances XLSX handling capabilities by adding support for importing and exporting data validation rules, which helps to maintain data integrity and consistency when working with Excel files.

Task: [3637642](https://www.odoo.com/odoo/project/2328/tasks/3637642)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/xlsx/conversion/data_validation_conversion.ts
```diff
@@ -0,0 +1,125 @@
...
+  };
+}
+
+function convertDateRule(id: number, dv: XLSXDataValidation): DataValidationRuleData {
+  const values = [dv.formula1.toString()];
+  if (dv.formula2 && dv.formula2 !== "undefined") {
+    values.push(dv.formula2.toString());
+  }
+  let criterion = {
+    type: XLSX_DV_DATE_OPERATOR_TO_DV_TYPE_MAPPING[dv.operator] ?? "dateIsBetween",
```
hokolomopo, line 100: Should we really have a default ? The operator not being in the mapping mean that we do  not support it, we should skip the rule conversion altogether.

hokolomopo (commented):
🏎️ 

I didn't test that it worked in excel, but I trust you taht it works. Somethign I feel like I already told you is that _**we should always avoid typecasting and ts-ignore**_.  Typescript is there for a reason, and he is usually correct, even if it's annoying. The only casting you need is the one in `convertDecimalCriterionTypeToExcelOperator` as we cannot type object keys, the other ones can be avoided I think.

Otherwise it looks fine 👋

tests/xlsx/xlsx_import.test.ts
```diff
@@ -319,6 +320,86 @@ describe("Import xlsx data", () => {
...
+    const dvRule = getDataValidationBeginningAt(cellXc, testSheet);
+    switch (ruleType) {
+      case "number":
+      case "time":
+      case "textLength":
+        // Unsupported data validation type 'textLength'
+        expect(dvRule).toBeUndefined();
+        break;
+      case "decimal":
+        expect(dvRule).toMatchObject({
```
LucasLefevre, line 344: if you do a `.each` then switch/case on the value of the "each", then it should just be different tests. Even if it means repeating some setup lines

src/xlsx/functions/data_validation.ts
```diff
@@ -0,0 +1,132 @@
...
+  const rule = dvRule.criterion;
+  const formula1 = adaptFormulaToExcel(rule.values[0]);
+  const formula2 = rule.values[1] ? adaptFormulaToExcel(rule.values[1]) : undefined;
+  const operator = convertDateCriterionTypeToExcelOperator(dvRule.criterion.type);
+  const attributes = commonDataValidationAttributes(dvRule);
+  attributes.push(["type", "date"], ["operator", operator]);
+  if (formula2) {
+    return escapeXml/*xml*/ `
+      <dataValidation ${formatAttributes(attributes)}>
+        <formula1>${toSerialDate(formula1)}</formula1>
```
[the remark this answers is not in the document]
Rachico: Excel expects a serial date to be able to preserve the correct date value.

LucasLefevre (approved):
robodoo r+

PR #5098, opened 2024-10-15
[IMP] Clipboard: support images in the clipboard

rrahir:
## Description:
This commit adds the support of images in the clipboard. Namely,
- Pasting an image from the OS clipboard will automatically upload it to
  the filestore and dispatch a CREATE_IMAGE accordingly.
- Copying a chart or an image from the spreadsheet will add it to the
  oOS clipboard. It can then be pasted somewhere else, like a discuss
app, a text editor maybe?.

TODO
- [ ] support multi type of images (not just png) It technically works
  because we never check agains the actual type of the file we upload
but it's wrong. Maybe use a generic `image/*` mimetype instead of using
png. We still need to purge out dangerous mimetypes like image/svg(+xml)
- [ ] Currently uploads the image from the clipboard THEN only checks if the
  we are copy pasting from the same spreadsheet clipboard (potential
useless upload)
- [ ] Not supported by some apps like Office 365 online (word & powerpoint tested). Supposedly because our
  clipboard has some text/* types along the image/* so it ignores the
later. for a fully integrated feature, we should probably decorrelate it
from the standard copy paste and make a  menu "copy image to os
clipboard" ? tbd
- [ ] works on discord at least
- [ ] No test whatsoever
Task: [4235104](https://www.odoo.com/web#id=4235104&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/demo/file_store.js
+++ b/demo/file_store.js
@@ -21,4 +21,9 @@ export class FileStore {
   async delete(path) {
     console.warn("cannot delete file. Not implemented");
   }
+
+  async getFile(path) {
+    const response = await fetch(path);
+    return await response.blob();
+  }
 }
--- a/src/actions/menu_items_actions.ts
+++ b/src/actions/menu_items_actions.ts
@@ -55,13 +55,14 @@ async function paste(env: SpreadsheetChildEnv, pasteOption?: ClipboardPasteOptio
   const osClipboard = await env.clipboard.read();
   switch (osClipboard.status) {
     case "ok":
-      const clipboardContent = parseOSClipboardContent(osClipboard.content);
-      const clipboardId = clipboardContent.data?.clipboardId;
+      const clipboardId = env.model.getters.getClipboardId();
+      const osClipboardContent = parseOSClipboardContent(osClipboard.content, clipboardId);
+      const osClipboardId = osClipboardContent.data?.clipboardId;
 
       const target = env.model.getters.getSelectedZones();
 
-      if (env.model.getters.getClipboardId() !== clipboardId) {
-        interactivePasteFromOS(env, target, clipboardContent, pasteOption);
+      if (clipboardId !== osClipboardId) {
+        await interactivePasteFromOS(env, target, osClipboardContent, pasteOption);
       } else {
         interactivePaste(env, target, pasteOption);
       }
@@ -467,7 +468,7 @@ async function requestImage(env: SpreadsheetChildEnv): Promise<Image | undefined
     return await env.imageProvider!.requestImage();
   } catch {
     env.raiseError(_t("An unexpected error occurred during the image transfer"));
-    return undefined;
+    return;
   }
 }
 
@@ -477,7 +478,7 @@ export const CREATE_IMAGE = async (env: SpreadsheetChildEnv) => {
     const figureId = env.model.uuidGenerator.smallUuid();
     const image = await requestImage(env);
     if (!image) {
-      throw new Error("No image provider was given to the environment");
+      return;
     }
     const size = getMaxFigureSize(env.model.getters, image.size);
     const position = centerFigurePosition(env.model.getters, size);
--- a/src/components/helpers/dom_helpers.ts
+++ b/src/components/helpers/dom_helpers.ts
@@ -194,3 +194,29 @@ export function isCtrlKey(ev: KeyboardEvent | MouseEvent): boolean {
 export function isMiddleClickOrCtrlClick(ev: MouseEvent): boolean {
   return ev.button === 1 || (isCtrlKey(ev) && ev.button === 0);
 }
+
+export async function convertImageToPng(imageUrl: string): Promise<Blob | null> {
+  return new Promise((resolve, reject) => {
+    const image = new Image();
+    image.addEventListener("load", () => {
+      const canvas = document.createElement("canvas");
+      canvas.width = image.width;
+      canvas.height = image.height;
+
+      const ctx = canvas.getContext("2d");
+      ctx?.drawImage(image, 0, 0);
+      canvas.toBlob(resolve, "image/png");
+    });
+    image.addEventListener("error", reject);
+    image.src = imageUrl;
+  });
+}
+
+export function downloadFile(dataUrl: string, fileName: string) {
+  const a = document.createElement("a");
+  a.href = dataUrl;
+  a.download = fileName;
+  document.body.appendChild(a);
+  a.click();
+  document.body.removeChild(a);
+}
--- a/src/components/icons/icons.xml
+++ b/src/components/icons/icons.xml
@@ -82,6 +82,19 @@
       />
     </svg>
   </t>
+  <t t-name="o-spreadsheet-Icon.COPY_AS_IMAGE">
+    <svg class="o-icon">
+      <path
+        fill="currentColor"
+        d="m 14,15 v 1.5 H 2.5147386 c -0.5522847,0 -1,-0.447715 -1,-1 V 3 H 3 V 14 c 0,0.552285 0.4477153,1 1,1 h 10"
+      />
+      <path
+        fill="currentColor"
+        d="M 3.8194506,2.0480809 A 1.1084934,1.1763958 0 0 1 4.927944,0.87168505 H 15.273882 A 1.1084934,1.1763958 0 0 1 16.382375,2.0480809 V 13.027775 a 1.1084934,1.1763958 0 0 1 -1.108493,1.176396 H 4.927944 A 1.1084934,1.1763958 0 0 1 3.8194506,13.027775 V 2.0480809 H 4.927944 V 13.027775 H 15.273882 V 2.0480809 H 4.927944 M 12.3179,5.5772683 10.839908,9.1064553 9.7314149,7.93006 8.2534245,9.8907193 7.1449309,8.7143233 5.6669395,11.067114 H 14.534886 M 6.0364374,4.4008725 a 1.1084936,1.176396 0 0 1 2.2169871,0 1.1084936,1.176396 0 0 1 -2.2169871,0"
+        id="path1-5"
+      />
+    </svg>
+  </t>
   <t t-name="o-spreadsheet-Icon.PASTE">
     <svg class="o-icon">
       <path
@@ -969,6 +982,11 @@
   <t t-name="o-spreadsheet-Icon.HIDE_SHEET">
     <div class="o-icon">
       <i class="fa fa-eye-slash"/>
+    </div>
+  </t>
+  <t t-name="o-spreadsheet-Icon.DOWNLOAD">
+    <div class="o-icon">
+      <i class="fa fa-download"/>
     </div>
   </t>
 </templates>
--- a/src/helpers/clipboard/clipboard_helpers.ts
+++ b/src/helpers/clipboard/clipboard_helpers.ts
@@ -1,3 +1,4 @@
+import { SpreadsheetClipboardData } from "../../plugins/ui_stateful";
 import {
   ClipboardCellData,
   ClipboardMIMEType,
@@ -6,6 +7,7 @@ import {
   UID,
   Zone,
 } from "../../types";
+import { AllowedImageMimeTypes } from "../../types/image";
 import { mergeOverlappingZones, positions } from "../zones";
 
 export function getClipboardDataPositions(sheetId: UID, zones: Zone[]): ClipboardCellData {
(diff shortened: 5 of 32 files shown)


src/actions/menu_items_actions.ts
```diff
@@ -55,12 +56,13 @@ async function paste(env: SpreadsheetChildEnv, pasteOption?: ClipboardPasteOptio
   const osClipboard = await env.clipboard.read();
   switch (osClipboard.status) {
     case "ok":
-      const clipboardContent = parseOSClipboardContent(osClipboard.content);
-      const clipboardId = clipboardContent.data?.clipboardId;
+      const clipboardId = env.model.getters.getClipboardId();
+      const clipboardContent = await parseOSClipboardContent(env, osClipboard.content, clipboardId);
+      const contentClipboardId = clipboardContent.data?.clipboardId;
```
hokolomopo, line 61:
```suggestion
      const osClipboardContent = await parseOSClipboardContent(env, osClipboard.content, clipboardId);
      const osClipboardId = clipboardContent.data?.clipboardId;
```
nitpick. Mayeb those are better names ? Same in grid

src/actions/menu_items_actions.ts
```diff
@@ -465,8 +467,12 @@ export const REINSERT_STATIC_PIVOT_CHILDREN = (env: SpreadsheetChildEnv) =>
 async function requestImage(env: SpreadsheetChildEnv): Promise<Image | undefined> {
   try {
     return await env.imageProvider!.requestImage();
-  } catch {
-    env.raiseError(_t("An unexpected error occurred during the image transfer"));
+  } catch (e) {
+    env.raiseError(
+      e instanceof ImageImportError
+        ? e.message
+        : _t("An unexpected error occurred during the image transfer")
```
hokolomopo, line 474: Should we also put the actual error message ? For error reports & such

src/helpers/clipboard/clipboard_helpers.ts
```diff
@@ -62,22 +65,46 @@ export function getPasteZones<T>(target: Zone[], content: T[][]): Zone[] {
...
+  if (contentClipboardId !== clipboardId) {
+    const clipboardContent: ParsedOSClipboardContent = {
+      text: content[ClipboardMIMEType.PlainText],
+      data: spreadsheetContent,
+    };
+    for (const type of AllowedImageMimeTypes) {
+      if (content[type]) {
+        // TODO: support multiple import
+        try {
+          const imageData = await env.imageProvider?.uploadFile(content[type]!);
```
hokolomopo, line 95: If we paste multiple time the same image we upload it multiple times in the filestroe. Is there any way to fix that ?

src/plugins/ui_stateful/clipboard.ts
```diff
@@ -130,17 +142,37 @@ export class ClipboardPlugin extends UIPlugin {
       case "PASTE_FROM_OS_CLIPBOARD": {
         this._isCutOperation = false;
 
-        this.copiedData =
-          cmd.clipboardContent.data ||
-          this.convertTextToClipboardData(cmd.clipboardContent.text ?? "");
-
+        const htmlData = cmd.clipboardContent.data;
+        // TODO: support multiple image import
```
hokolomopo, line 146: maybe not necessary to have 3 times the same TODO

src/plugins/ui_stateful/clipboard.ts
```diff
@@ -517,11 +555,31 @@ export class ClipboardPlugin extends UIPlugin {
...
     if (!cells) {
-      innerHTML = "\t";
+      if (this.copiedData?.figureId) {
+        const figureId = this.copiedData.figureId;
+        const figureSheetId = this.getters.getFigureSheetId(figureId)!;
+        const figure = this.getters.getFigure(figureSheetId, figureId)!;
+        if (figure.tag == "image") {
+          if (!this.fileStore) {
+            innerHTML = "\t";
+          }
```
hokolomopo, line 569: the if is kinda pointless without some kind of break/return/else. We still go through the code that follow and the next if re-set the innerHTML

src/plugins/ui_stateful/clipboard.ts
```diff
@@ -549,6 +607,66 @@ export class ClipboardPlugin extends UIPlugin {
...
+          image.addEventListener("load", () => {
+            const canvas = document.createElement("canvas");
+            canvas.width = image.width;
+            canvas.height = image.height;
+
+            const ctx = canvas.getContext("2d");
+            ctx?.drawImage(image, 0, 0);
+            canvas.toBlob(resolve, "image/png");
+          });
+          image.addEventListener("error", () => reject);
```
hokolomopo, line 659: maybe extract this to a helper to not pollute too much the clipboard

src/registries/figure_registry.ts
```diff
@@ -58,6 +61,40 @@ function getChartMenu(
...
+        const figure = env.model.getters.getFigure(figureSheetId, figureId)!;
+        const chartType = env.model.getters.getChartType(figureId);
+        const runtime = env.model.getters.getChartRuntime(figureId);
+        const imageUrl = chartToImage(runtime, figure, chartType)!;
+        const innerHTML = `<img src="${xmlEscape(imageUrl)}" />`;
+        const blob = await chartToImageFile(runtime, figure, chartType)!;
+
+        env.clipboard.write({
+          "text/html": innerHTML,
+          "image/png": blob || undefined,
```
hokolomopo, line 80: is the || undefined useful ? chartToImageFile is typed to return File | undefined already

hokolomopo (commented):
boup boup

src/helpers/clipboard/clipboard_helpers.ts
rrahir: This is the responsibility of the filestore to check unicity, which is left to the integrator (e.g. the code inside Odoo) to manage duplicates (it could decide to detect already existing file and yield a link against that file). Inside Odoo, we have unicity of the file on disc, just a different attachment link).

src/plugins/ui_stateful/clipboard.ts
rrahir: it's more about "where should we act when we want to support several images"

src/plugins/ui_stateful/clipboard.ts
```diff
@@ -517,11 +553,31 @@ export class ClipboardPlugin extends UIPlugin {
...
+        if (figure.tag == "image") {
+          if (!this.fileStore) {
+            return "\t";
+          }
+          const imageUrl = this.getters.getImage(figureId).path;
+          const file = (await this.fileStore?.getFile(imageUrl)) || null;
+
+          if (file) {
+            const imageUrl = (await this.readFileAsDataURL(file)) as string;
+            innerHTML = `<img src="${xmlEscape(imageUrl)}" />`;
```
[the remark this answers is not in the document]
rrahir: most of the time, it'll probably be an actual public url, easier than encoding everything. Such an approach would require for the integrator to develop a "getPublicUrl" method to expose attachments and such to public requests. Technically feasible but it makes the feature hard to use in stand-alone, which is what I was trying to avoid.

src/helpers/figures/images/image_provider.ts
```diff
@@ -10,13 +15,25 @@ export class ImageProvider implements ImageProviderInterface {
   }
 
   async requestImage(): Promise<Image> {
-    const file = await this.getImageFromUser();
+    const file = await this.userImageUpload();
+    if (file.size > MAX_FILE_SIZE) {
+      throw new ImageImportError(_t("The file you are trying to upload is too large (> 5 mB)"));
```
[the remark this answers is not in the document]
rrahir: yup, it might not be useful at upload. The filestore implementation might reject files too big but it can be taken care of at that moment.

src/plugins/ui_stateful/clipboard.ts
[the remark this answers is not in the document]
rrahir: I realized I didn't directly answer the question but yes, if you get to inspect your clipboard after copying an image you'll have an html structure similar to this one, with an src that points to the file or an encoded one (see what I explain above). I didn't find the spec about this but I know that some programs (ms 365 suite for instance) solely rely on that information and completely ignore the DataTransfer items attached in the clipboard. You can test it out by copying an image in your PC and try to paste it in Word online.
image

src/components/helpers/dom_helpers.ts
```diff
@@ -87,3 +87,29 @@ export function isMacOS(): boolean {
 export function isCtrlKey(ev: KeyboardEvent | MouseEvent): boolean {
   return isMacOS() ? ev.metaKey : ev.ctrlKey;
 }
+
+export async function convertImageToPng(imageUrl: string): Promise<Blob | null> {
+  return await new Promise((resolve, reject) => {
+    const image = new Image();
+    image.src = imageUrl;
+    image.addEventListener("load", () => {
```
LucasLefevre, line 95: probably doesn't matter in practice, but should the event listener be attached before setting `src` ? If somehow `image.src = imageUrl;` is instant and synchronous, the event handler would never be called

src/components/helpers/dom_helpers.ts
```diff
@@ -87,3 +87,29 @@ export function isMacOS(): boolean {
...
+
+      const ctx = canvas.getContext("2d");
+      ctx?.drawImage(image, 0, 0);
+      canvas.toBlob(resolve, "image/png");
+    });
+    image.addEventListener("error", () => reject);
+  });
+}
+
+export function serveFile(dataUrl: string, fileName: string) {
```
LucasLefevre, line 108:
```suggestion
export function downloadFile(dataUrl: string, fileName: string) {
```

src/registries/figure_registry.ts
```diff
@@ -58,6 +61,40 @@ function getChartMenu(
     },
     getCopyMenuItem(figureId, env),
     getCutMenuItem(figureId, env),
+    {
+      id: "copy_as_image",
+      name: _t("Copy as image"),
+      icon: "o-spreadsheet-Icon.COPY_AS_IMAGE",
```
LucasLefevre, line 67: same as the existing `o-spreadsheet-Icon.COPY`
```suggestion
      icon: "o-spreadsheet-Icon.COPY",
```

src/components/icons/icons.xml
```diff
@@ -953,4 +966,9 @@
       <i class="fa fa-eye-slash"/>
     </div>
   </t>
+  <t t-name="o-spreadsheet-Icon.DOWNLOAD">
+    <svg class="o-icon">
+      <path fill="currentColor" d="M2 10.5h1.5V15h11v-4.5H16v6H2M5.5 8H8V1h2v7h2.5L9 11.5"/>
```
LucasLefevre, line 971: shorter and consistent with other download buttons in odoo
```suggestion
      <i class="fa fa-download"/>
```

src/components/icons/icons.xml
```diff
@@ -82,6 +82,19 @@
       />
     </svg>
   </t>
+  <t t-name="o-spreadsheet-Icon.COPY_AS_IMAGE">
```
LucasLefevre, line 85: not needed

src/plugins/ui_stateful/clipboard.ts
```diff
@@ -53,7 +58,7 @@ export interface SpreadsheetClipboardData extends MinimalClipboardData {
 export class ClipboardPlugin extends UIPlugin {
   static layers = ["Clipboard"] as const;
   static getters = [
-    "getClipboardContent",
+    "getOsClipboardContentAsync",
```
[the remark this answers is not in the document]
rrahir: sepcial request from Vincent as it's the only async getter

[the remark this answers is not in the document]
rrahir: @VincentSchippefilt  wdyt?

src/helpers/figures/charts/chart_ui_common.ts
```diff
@@ -72,6 +72,33 @@ export function chartToImage(
...
+export async function chartToImageFile(
+  runtime: ChartRuntime,
+  figure: Figure,
+  type: ChartType
+): Promise<File | undefined> {
+  const chartString = chartToImage(runtime, figure, type);
+  if (!chartString) {
+    return undefined;
+  }
+  return base64ToFile(chartString, "chart.png", "image/png");
```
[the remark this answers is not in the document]
rrahir: I did it a t first, can't recall why I abandonned that idea. I think it was a problem of file size.. i'll try to find something in my memory

demo/file_store.js
```diff
@@ -21,4 +21,9 @@ export class FileStore {
   async delete(path) {
     console.warn("cannot delete file. Not implemented");
   }
+
+  async getFile(path) {
+    const response = await fetch(path);
```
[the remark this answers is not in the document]
rrahir: The thing is we end up sharing the role of the image provider and the file store then. the file store could do random stuff on its own without the knowledge of the other bricks.

src/components/helpers/dom_helpers.ts
rrahir: I just couldn't make it work without that, i might have missed something...

src/registries/figure_registry.ts
rrahir: so twice the same icon in the same menu?

LucasLefevre: Oh sorry, I really thought it was exactly the same icon but I wasn't checking the right menu item

src/plugins/ui_stateful/clipboard.ts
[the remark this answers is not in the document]
VincentSchippefilt: I would rather name it for what it does than for it's usage: `getClipboardTextAndImageContent` and maybe add the async, but the linter tells us it's async so we don't need to specify it in the name.

src/actions/menu_items_actions.ts
```diff
@@ -465,8 +466,8 @@ export const REINSERT_STATIC_PIVOT_CHILDREN = (env: SpreadsheetChildEnv) =>
 async function requestImage(env: SpreadsheetChildEnv): Promise<Image | undefined> {
   try {
     return await env.imageProvider!.requestImage();
-  } catch {
-    env.raiseError(_t("An unexpected error occurred during the image transfer"));
+  } catch (e) {
+    env.raiseError(e.message || _t("An unexpected error occurred during the image transfer"));
```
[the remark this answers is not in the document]
rrahir: 👍  removed it as it only made sense to catch the image provide error if the file was too big but I removed that feature altogether.

src/helpers/figures/images/image_provider.ts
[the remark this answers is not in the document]
rrahir: => removed the upload limitation

LucasLefevre (commented):
mainly details

src/helpers/clipboard/clipboard_helpers.ts
```diff
@@ -62,22 +64,41 @@ export function getPasteZones<T>(target: Zone[], content: T[][]): Zone[] {
...
+      content[ClipboardMIMEType.Html],
+      "text/html"
+    );
+    const oSheetClipboardData = htmlDocument
+      .querySelector("div")
+      ?.getAttribute("data-osheet-clipboard");
+    spreadsheetContent = oSheetClipboardData && JSON.parse(oSheetClipboardData);
+    contentClipboardId = spreadsheetContent?.clipboardId;
+  }
+  if (contentClipboardId !== clipboardId) {
```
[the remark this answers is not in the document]
rrahir: not necessary indeed, it made sense when the helper was actively uploading to the server 👍

LucasLefevre (commented):
robodoo rebase-ff r+

PR #5101, opened 2024-10-17
[Find&Replace] A couple of imp

hokolomopo:
### [FIX] find and replace: do not update selection at each change

Previously, when doing any operation that would refresh the search
(eg. an update_cell), we would update the selection to the current
search result. This was preventing the user to have any work done
while the f&r panel was open.

This commit changes the behavior to only update the selection when
the user explicitly clicks on a search result or changes the sheet.

### [IMP] find and replace: focus search input on Ctrl+F

Pressing Ctrl+F with the f&r search panel open would close it. This
commit changes the behavior to focus the search input instead.

### [REF] spreadsheet: move Ctrl+F handling to Grid component

The handling of the Ctrl+F key combination was done in the Spreadsheet
component. This made no sense as:

- the Spreadsheet component never has the Dom focus
- we needed to explicitly disable the shortcuts in dashboard mode

This commit moves the handling of the Ctrl+F key combination to the Grid
along all the other keyboard shortcuts. It's less code, and the behavior
is now automatically disabled in dashboard mode (since there is no
Grid component in dashboard mode).

Task: [3973671](https://www.odoo.com/web#id=3973671&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/find_and_replace/find_and_replace_store.ts
+++ b/src/components/side_panel/find_and_replace/find_and_replace_store.ts
@@ -17,6 +17,11 @@ enum Direction {
   next = 1,
 }
 
+interface RefreshSearchOptions {
+  jumpToMatchSheet?: boolean;
+  updateSelection?: boolean;
+}
+
 export class FindAndReplaceStore extends SpreadsheetStore implements HighlightProvider {
   mutators = [
     "updateSearchOptions",
@@ -31,10 +36,12 @@ export class FindAndReplaceStore extends SpreadsheetStore implements HighlightPr
   private specificRangeMatches: CellPosition[] = [];
 
   private currentSearchRegex: RegExp | null = null;
-  private isSearchDirty = false;
   private initialShowFormulaState: boolean;
   private preserveSelectedMatchIndex: boolean = false;
   private irreplaceableMatchCount: number = 0;
+
+  private isSearchDirty = false;
+  private shouldFinalizeUpdateSelection = false;
 
   private notificationStore = this.get(NotificationStore);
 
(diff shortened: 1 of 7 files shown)


src/components/side_panel/find_and_replace/find_and_replace_store.ts
```diff
@@ -17,6 +17,11 @@ enum Direction {
   next = 1,
 }
 
+interface RefreshSearchOptions {
+  doNotJumpToMatchSheet?: boolean;
+  doNotSelectNextMatch?: boolean;
```
pro-odoo, line 22: Personal option, but I always prefer non-negative names, to avoid `!doNotJumpToMatchSheet` (not do not jump to match sheet). IMO, it's easier to read
```
if (jumpToMatchSheet) { ... }
```
than
 ```
 if (!doNotJumpToMatchSheet) { ... }
 ```

pro-odoo (approved):

src/components/side_panel/find_and_replace/find_and_replace_store.ts
hokolomopo: Yeah I tend to agree. I did that because I like truthy default arguments even less. But I can just make the arguments mandatory for the 4 calls to refreshSearch/selectNextCell, it's not a big deal

pro-odoo (approved):
robodoo rebase-ff r+

PR #5139, opened 2024-10-24
[FIX] chart: date line chart

dhrp-odoo:
## Description:

When creating a line chart with date labels, if 'Use row X as headers' is enabled, the first cell is empty. Currently, we're checking the format of this empty cell to determine if the labels are dates.

Task: [4268977](https://www.odoo.com/odoo/project/2328/tasks/4268977)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/figures/charts/line_chart.ts
+++ b/src/helpers/figures/charts/line_chart.ts
@@ -262,11 +262,7 @@ function canBeDateChart(labelRange: Range | undefined, getters: Getters): boolea
   if (!labelRange || !canBeLinearChart(labelRange, getters)) {
     return false;
   }
-  const labelFormat = getters.getEvaluatedCell({
-    sheetId: labelRange.sheetId,
-    col: labelRange.zone.left,
-    row: labelRange.zone.top,
-  }).format;
+  const labelFormat = getChartLabelFormat(getters, labelRange);
   return Boolean(labelFormat && timeFormatLuxonCompatible.test(labelFormat));
 }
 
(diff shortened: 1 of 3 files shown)


src/helpers/figures/charts/line_chart.ts
```diff
@@ -251,21 +257,25 @@ function getChartAxisType(chart: LineChart, getters: Getters): AxisType {
...
+  getters: Getters
+): boolean {
   if (!labelRange || !canBeLinearChart(labelRange, getters)) {
     return false;
   }
   const labelFormat = getters.getEvaluatedCell({
     sheetId: labelRange.sheetId,
     col: labelRange.zone.left,
-    row: labelRange.zone.top,
+    row: dataSetsHaveTitle ? labelRange.zone.top + 1 : labelRange.zone.top,
```
LucasLefevre, line 278: Hello :wave: 

There's an alternative implementation that might be less invasive in all the code (additional parameter to a lot of functions) and even better from a functional point of view.

The idea would be to check *all* cells in the label range, one after the other (starting with the first empty one if "dataSetsHaveTitle") and return the first format found.

dhrp-odoo: Thanks for the suggestion! I’ve removed the `dataSetsHaveTitle` parameter and updated the code accordingly :)

anhe-odoo: @LucasLefevre : What if we have different formats in the label range ? We still keep the first found ?

LucasLefevre: Yes. It's not worse than the current behavior (only from the first cell) and it's easy

LucasLefevre (approved):
Nice, thanks!
robodoo r+

PR #5140, opened 2024-10-24
[IMP] chart: improve bar chart borders

anhe-odoo:
## Task Description

Since the new design of chart color, when we try to draw a bar chart, we have a small difference between a classical bar chart and a combo bar where every series is a bar. This is due to the border of the bar in a bar chart that are drawn in a whitish color, while the border are still in the same color as the bar in a combo chart. This task aims to change the way we draw the border of a bar chart, only drawing a small white border between two stacked bar.

## Related Task
- Task: 4283212

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/figures/charts/bar_chart.ts
+++ b/src/helpers/figures/charts/bar_chart.ts
@@ -1,6 +1,6 @@
 import type { ChartDataset, LegendOptions } from "chart.js";
 import { DeepPartial } from "chart.js/dist/types/utils";
-import { BACKGROUND_CHART_COLOR, BORDER_CHART_COLOR } from "../../../constants";
+import { BACKGROUND_CHART_COLOR } from "../../../constants";
 import {
   AddColumnsRowsCommand,
   ApplyRangeChange,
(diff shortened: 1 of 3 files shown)


src/helpers/figures/charts/bar_chart.ts
```diff
@@ -304,7 +304,10 @@ export function createBarChartRuntime(chart: BarChart, getters: Getters): BarCha
       label,
       data,
       borderColor: BORDER_CHART_COLOR,
-      borderWidth: 1,
+      borderWidth: {
+        bottom: index === "0" || !chart.stacked ? 0 : 2,
+      },
+      borderSkipped: false,
```
hokolomopo, line 310: Thinking about it a bit more, it unfortunately doesn't work 😞 If the stacked bar chart contain negative values, we want the top border, not the bottom.

I'd say we should probably either enable all the borders of the stacked bar charts, or disable all the bar borders always and not bother too much.

src/helpers/figures/charts/bar_chart.ts
```diff
@@ -303,8 +303,8 @@ export function createBarChartRuntime(chart: BarChart, getters: Getters): BarCha
     const dataset: ChartDataset<"bar", number[]> = {
       label,
       data,
-      borderColor: BORDER_CHART_COLOR,
-      borderWidth: 1,
```
hokolomopo, line 307: A small test would be nice, so we do'nt break this again :slightly_smiling_face:

hokolomopo (commented):
Commit message is a bit strange with "Task description" and "Related Task" :upside_down_face:

LucasLefevre (approved):
robodoo r+

PR #5161, opened 2024-11-05
[FIX] pivot: skip cells on manipulated pivot array formulas

LucasLefevre:
## Description:

Steps to reproduce in odoo:

- insert a pivot with more rows than columns
- write in a cell =TRANSPOSE(ODOO.PIVOT.TABLE(1))
- right-click on the grand total value

=> boom

When computing which cell of the pivot table is clicked, we assume the matrix comes directly from the ODOO.PIVOT.TABLE(...) function to compute the offsets from the array formula. But it's completely wrong as the cell could at a completely different place if the matrix is manipulated by other functions before being outputted to the grid.

Forward-port of https://github.com/odoo/odoo/commit/d33a9f3

Task: [4292134](https://www.odoo.com/web#id=4292134&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_core_views/pivot_ui.ts
+++ b/src/plugins/ui_core_views/pivot_ui.ts
@@ -197,6 +197,12 @@ export class PivotUIPlugin extends UIPlugin {
     if (!pivot.isValid()) {
       return EMPTY_PIVOT_CELL;
     }
+    if (
+      functionName === "PIVOT" &&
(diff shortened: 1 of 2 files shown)


LucasLefevre:
robodoo r+

LucasLefevre:
non, c'est bon, everything is good
robodoo r+

PR #5163, opened 2024-11-05
[PERF] evaluation: early return errors

LucasLefevre:
## Description:

Throwing errors is super slow but the evaluation uses throw in many
places when the formula result is an error (historic). Returning
the error result is better than throwing.

Extensive work has been done in the past year to reduce `throw`, especially
the "Loading..." errors.

But if a function receives an error object as an argument and the
function uses one of the casting helpers (`toNumber`, `toString`, etc.)
the error is still thrown.

With this commit, if a function argument is an error, we by pass the
function `compute` and the error is directly returned.

That is unless the argument is declared as accepting the type `any`,
in which case it can use the error argument as part of its business logic.

We currently only check simple scalar arguments (and not matrices) for
performance reasons. Iterating over all matrices to check if there's any
error is costly and probably outweights the benefits (especially when there
isn't any error)

On one of the dashboard on odoo.com (id=116), the evaluation triggered by
setting a global filter is reduced by ~70% (when setting a filter,
everything is reloading and hence there are "Loading..." everywhere

before: 1707ms
after:   533ms

There's no significant performance hit on regular evaluations (when there's
no errors)

Many types of many function arguments had to be changed to remove `any`
where the function doesn't need to handle errors itself.

For some function expecting an matrix of a given type (let's say
`range<number>`), we added `range<any>`. The real type of those arguments
is `range<any>` but from a functional POV, they expect `range<number>` and
ignore any other values. We leave the `range<number>` only for an informative
reason.

Task: [4328215](https://www.odoo.com/web#id=4328215&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/functions/arguments.ts
+++ b/src/functions/arguments.ts
@@ -55,6 +55,10 @@ function makeArg(str: string, description: string): ArgDefinition {
     description,
     type: types,
   };
+  const acceptErrors = types.includes("ANY") || types.includes("RANGE");
+  if (acceptErrors) {
+    result.acceptErrors = true;
+  }
   if (isOptional) {
     result.optional = true;
   }
--- a/src/functions/index.ts
+++ b/src/functions/index.ts
@@ -215,6 +215,17 @@ function createComputeFunction(
     this: EvalContext,
     ...args: Arg[]
   ): Matrix<FunctionResultObject> | FunctionResultObject {
+    for (let i = 0; i < args.length; i++) {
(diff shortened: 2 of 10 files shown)


laa-odoo:
@robodoo r+

PR #5164, opened 2024-11-05
[IMP] conditional_format: improve CF data bar behavior for non-matching range sizes

Rachico:
Before this commit, defining a CF data bar rule based on the value of another range requires both ranges to have matching sizes, preventing the CF from being created otherwise. This restriction leads to a poor user experience when copying and pasting CFs, as the CF does not get pasted.

This commit removes the restriction, allowing CF creation in a "best effort" manner when the range sizes do not match. The CF style is therefore applied by comparing the matching cells.

Task: [4280720](https://www.odoo.com/odoo/project/2328/tasks/4280720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/core/conditional_format.ts
+++ b/src/plugins/core/conditional_format.ts
@@ -1,5 +1,5 @@
 import { compile } from "../../formulas/index";
-import { isInside, recomputeZones, toUnboundedZone, zoneToDimension } from "../../helpers/index";
+import { isInside, recomputeZones, toUnboundedZone } from "../../helpers/index";
 import {
   AddConditionalFormatCommand,
   ApplyRangeChange,
@@ -14,7 +14,6 @@ import {
   ConditionalFormatInternal,
   ConditionalFormattingOperatorValues,
   CoreCommand,
-  DataBarRule,
   ExcelWorkbookData,
   IconSetRule,
   IconThreshold,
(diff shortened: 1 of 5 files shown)


src/plugins/ui_core_views/evaluation_conditional_format.ts
```diff
@@ -320,6 +320,11 @@ export class EvaluationConditionalFormatPlugin extends UIPlugin {
   ) {
     const targetCol = col - zone.left + targetZone.left;
     const targetRow = row - zone.top + targetZone.top;
+    // if the target cell is outside the CF rangeValues
+    // we return -1 to avoid applying the data bar rule
```
LucasLefevre, line 324: I don't like the idea of a special case of data bar rule in a method that seems generic and supposed to return an evaluated cell.
I'd prefer if we inline this method where it's used

src/plugins/core/conditional_format.ts
```diff
@@ -614,21 +605,6 @@ export class ConditionalFormatPlugin
...
 
-  private checkDataBarRangeValues(rule: DataBarRule, ranges: RangeData[], sheetId: UID) {
-    if (rule.rangeValues) {
-      const { numberOfCols, numberOfRows } = zoneToDimension(
-        this.getters.getRangeFromSheetXC(sheetId, rule.rangeValues).zone
-      );
-      for (const range of ranges) {
-        const dimensions = zoneToDimension(this.getters.getRangeFromRangeData(range).zone);
-        if (numberOfCols !== dimensions.numberOfCols || numberOfRows !== dimensions.numberOfRows) {
-          return CommandResult.DataBarRangeValuesMismatch;
```
LucasLefevre, line 600: `CommandResult.DataBarRangeValuesMismatch` can be completely removed :wink:

LucasLefevre (commented):
I retargeted to 18.0 and applied my comment
robodoo r+

PR #5181, opened 2024-11-08
[IMP] gauge chart: allow formulas in gauge config

hokolomopo:
### [MOV] gauge: move gauge-specific panel tests

This commit moves the gauge-specific panel tests to a separate file
`gauge_panel_component.test.ts`.

A lot of gauge chart were also kinda duplicated: a test would test that
the input was marked as invalid when entering a wrong value, and another
test would test would test what the error message was when entering an
invalid value. Both can be done in the same test.

### [IMP] gauge: allow to have swap rangeMin and rangeMax

Currently, we have an allowDispatch checking that rangeMin is lower
than rangeMax. This is fine for literal values, but with the next
commit that allow formulas values for rangeMin and rangeMax, it is
harder to check.

This commit make the runtime swap rangeMin and rangeMax if rangeMin
is greater than rangeMax. It's simpler, more powerful, and in line
with what we do for the gauge thresholds that are swapped if needed.

### [REM] gauge: remove unused command result

There was a command result `GaugeLowerBiggerThanUpper` that was defined,
and checked in the gauge side panel. This command result was never
returned by the allowDispatch of the gauge charts. If the lower value
is bigger than the upper value, the gauge chart swap them.

### [IMP] gauge: prettify gauge chart plugin tests

This commit prettifies and improves the readability of the
`gauge_chart_plugin.test.ts` file:

- remove unused model creation data
- remove describe with only one test
- rename ids "1" or "2" to a more descriptive name
- prettify the code in one line when it doesn't affect the readability

### [IMP] gauge chart: allow formulas in gauge config

This commit adds the possibility to use formulas to configure the gauge
chart min/max ranges and the threshold values.

This make it so we cannot rely on allowDispatch now to check that the
values are number, because the only implemented allowDispatch of charts
is core (so no evaluation). The panel will try to filter the non-number
values, but they could change to non-number afterwards. In that case:

- the non-number thresholds will be dropped
- the chart will display an error if the range min/max is not a number

Task: [4236214](https://www.odoo.com/web#id=4236214&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


LucasLefevre (commented):
LGTM
except for ranges that should be adapted in `updateRanges` 

I'm ok with leaving the transformation (see [task 4319840](https://www.odoo.com/odoo/project/2328/tasks/4319840)) for now because it only happens in a concurrent and multi-user context, which is far less likely than just adding a column

src/components/side_panel/chart/gauge_chart_panel/gauge_chart_design_panel.ts
```diff
@@ -139,6 +145,12 @@ export class GaugeChartDesignPanel extends Component<Props, SpreadsheetChildEnv>
   }
 
   updateSectionRule(sectionRule: SectionRule) {
+    const invalidValueReasons = this.checkSectionRuleFormulasAreValid(this.state.sectionRule);
+    if (invalidValueReasons.length > 0) {
+      this.state.sectionRuleDispatchResult = new DispatchResult(invalidValueReasons);
+      return;
```
rrahir, line 151: this prevents the user from changing the general config onece there is an evaluated value taht does not fit the rule.. I think it should rather pipe the results together because I don't want to have to change the content of a cell or of my input juste to have the possibility to change the color of a threshold.

src/components/side_panel/chart/gauge_chart_panel/gauge_chart_design_panel.ts
```diff
@@ -139,6 +145,12 @@ export class GaugeChartDesignPanel extends Component<Props, SpreadsheetChildEnv>
   }
 
   updateSectionRule(sectionRule: SectionRule) {
+    const invalidValueReasons = this.checkSectionRuleFormulasAreValid(this.state.sectionRule);
+    if (invalidValueReasons.length > 0) {
+      this.state.sectionRuleDispatchResult = new DispatchResult(invalidValueReasons);
```
rrahir, line 150: Would it be possible to consider changing `sectionRuleDispatchResult` to something else like 
`sectionRuleErrors`? Creating a fake dispatch result just to be iable to use the ChartTerms looks so wrong :'(

rrahir (changes requested):

PR #5195, opened 2024-11-12
[IMP] model: create snapshot when loading xlsx file

Rachico:
Before this commit, when uploading an XLSX file, the code in the Odoo repository would create a spreadsheet document with data that is encoded in xml. Whenever the file is loaded, we parse the stringified xml to convert it to the o-spreadsheet json structure. The problem is that this conversion can be quite slow and costly when the XLSX file is large.

This commit mitigates the issue by saving a snapshot once the XLSX file is loaded for the first time.

Task: [4080514](https://www.odoo.com/odoo/project/2328/tasks/4080514)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/model.ts
+++ b/src/model.ts
@@ -233,7 +233,7 @@ export class Model extends EventBus<any> implements CommandDispatcher {
     this.session.loadInitialMessages(stateUpdateMessages);
     this.joinSession();
 
-    if (config.snapshotRequested) {
+    if (config.snapshotRequested || data["[Content_Types].xml"]) {
       this.session.snapshot(this.exportData());
     }
     // mark all models as "raw", so they will not be turned into reactive objects
--- a/tests/model.test.ts
+++ b/tests/model.test.ts
@@ -1,4 +1,5 @@
 import { CommandResult, CorePlugin } from "../src";
+import { MESSAGE_VERSION } from "../src/constants";
 import { toZone } from "../src/helpers";
 import { Model, ModelConfig } from "../src/model";
 import { corePluginRegistry, uiPluginRegistry } from "../src/plugins/index";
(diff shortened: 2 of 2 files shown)


tests/model.test.ts
```diff
@@ -212,4 +215,24 @@ describe("Model", () => {
     const model = new Model(modelData);
     expect(model.exportData()).toMatchSnapshot();
   });
+
+  test("snapshot when importing xlsx file", async () => {
+    const originalWarn = console.warn.bind(console.warn);
+    console.warn = () => {};
```
LucasLefevre, line 221: use can use `jest.spyOn` :wink: 
```suggestion
    jest.spyOn(console, "warn").mockImplementation(() => {});
```

tests/model.test.ts
```diff
@@ -212,4 +215,24 @@ describe("Model", () => {
...
+    let transport = new MockTransportService();
+    const spy = jest.spyOn(transport, "sendMessage");
+    const xlsx_data = getTextXlsxFiles();
+    new Model(xlsx_data, {
```
LucasLefevre, line 225:
```suggestion
    const xlsxData = getTextXlsxFiles();
    new Model(xlsxData, {
```

tests/model.test.ts
```diff
@@ -212,4 +215,24 @@ describe("Model", () => {
     const model = new Model(modelData);
     expect(model.exportData()).toMatchSnapshot();
   });
+
+  test("snapshot when importing xlsx file", async () => {
+    const originalWarn = console.warn.bind(console.warn);
+    console.warn = () => {};
+    let transport = new MockTransportService();
```
LucasLefevre, line 222:
```suggestion
    const transport = new MockTransportService();
```

LucasLefevre (commented):
a few details :wink:

src/model.ts
```diff
@@ -233,7 +233,7 @@ export class Model extends EventBus<any> implements CommandDispatcher {
     this.session.loadInitialMessages(stateUpdateMessages);
     this.joinSession();
 
-    if (config.snapshotRequested) {
+    if (config.snapshotRequested || data["[Content_Types].xml"]) {
```
LucasLefevre, line 236: This only works if the user has write access on the record. You also have to check if the model isn't in read-only mode.
I'd target 18.0 because I don't want to spend time fixing more bugs in 16.0 if this introduces an unexpected regression

Rachico:
This task will target the 18.0 branch instead.
https://github.com/odoo/o-spreadsheet/pull/5320

PR #5197, opened 2024-11-13
[REV] pivot: select pivot cells in composer

LucasLefevre:
This reverts commit 6c4a8e75631c41fd49fc297e995d95969c6060f3.

It is quite annoying to build a formula by referring a pivot value and getting the full pivot formula rather than its cell address.

So a good behavior IMO would be:

If you copy a value from a pivot and paste it somewhere else, you would get the full formula (i.e. =PIVOT(1,$$$,$$$,$$$))
If you select a value from a pivot while building a formula within the composer, you would get its address (i.e. =SUM(1+C3+4)) and not (=SUM(1+(PIVOT(1,$$$,$$$,$$$))+4)

Additional notes:

With the current behavior (full formula inserted) and if you actually want the reference, it's very annoying: you have to delete the full formula, carefully check the reference of the cell you want to target, then type the reference by hand. The other way around is slightly easier: either build the formula using the auto-complete assistant, or copy-paste the cell somewhere else and reference that cell with the fixed formula.

There are pros & cons with both behaviors, and they are incompatible. I say: if you use a dynamic pivot, we let you embrace dynamic all the way, otherwise just use a fixed formula pivot. Reverting full formula insertion is also less code and less surprising (not a different behavior just for pivots)

Task: 4189098

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/helpers/rendering.ts
```diff
@@ -16,7 +16,6 @@ export function drawHighlight(
   const color = highlight.color || HIGHLIGHT_COLOR;
 
   const { ctx } = renderingContext;
-  ctx.save();
```
hokolomopo, line 19: why is this removed ?

hokolomopo (commented):
+1, having formulas being automatically written when selecting a cell is really annoying IMO. Excel does it a lot with tables, and I hate it :imp:

src/helpers/rendering.ts
LucasLefevre: because it was added in the reverted commit, but I don't remember why :man_shrugging: 
It was probably a "bug" that only appeared in this case. I'll restore it

rrahir:
@robodoo r+

PR #5202, opened 2024-11-13
[FIX] charts: non-invertible matrix returns NaN as predicted data

anhe-odoo:
## Task Description

When trying to make a trending line, almost all model use matrix computation to get the predicted data, with some matrix multiplications and inversions. When the matrix aren't inversibles, the user get a traceback, which is not fine.
This PR aims to fix these issues by returning an array of NaN when there is an error in the computation of the predicted dataset.

## Related Task
- Task: 4328743

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/figures/charts/chart_common.ts
+++ b/src/helpers/figures/charts/chart_common.ts
@@ -535,41 +535,48 @@ export function interpolateData(
   const labelRange = labelMax - labelMin;
   const normalizedLabels = labels.map((v) => (v - labelMin) / labelRange);
   const normalizedNewLabels = newLabels.map((v) => (v - labelMin) / labelRange);
-  switch (config.type) {
-    case "polynomial": {
-      const order = config.order ?? 2;
-      if (order === 1) {
-        return predictLinearValues([values], [normalizedLabels], [normalizedNewLabels], true)[0];
-      }
-      const coeffs = polynomialRegression(values, normalizedLabels, order, true).flat();
-      return normalizedNewLabels.map((v) => evaluatePolynomial(coeffs, v, order));
-    }
-    case "exponential": {
-      const positiveLogValues: number[] = [];
-      const filteredLabels: number[] = [];
-      for (let i = 0; i < values.length; i++) {
-        if (values[i] > 0) {
-          positiveLogValues.push(Math.log(values[i]));
-          filteredLabels.push(normalizedLabels[i]);
+  try {
+    switch (config.type) {
+      case "polynomial": {
+        const order = config.order;
+        if (!order) {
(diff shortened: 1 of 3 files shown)


src/helpers/figures/charts/chart_common.ts
```diff
@@ -554,13 +558,24 @@ export function interpolateData(
...
-        predictLinearValues([positiveLogValues], [filteredLabels], [normalizedNewLabels], true)
-      )[0];
     }
     case "logarithmic": {
+      try {
+        return logM(
+          predictLinearValues([values], [normalizedLabels], [normalizedNewLabels], true)
+        )[0];
+      } catch (e) {
+        return [];
```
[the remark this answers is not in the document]
anhe-odoo: The empty array is a mistake, it should returns NaN every time. In fact, the idea of returning NaN was to see the series in the legend, while an empty array gives no entry in the legend.

src/helpers/figures/charts/chart_common.ts
```diff
@@ -541,8 +541,12 @@ export function interpolateData(
...
         return predictLinearValues([values], [normalizedLabels], [normalizedNewLabels], true)[0];
       }
-      const coeffs = polynomialRegression(values, normalizedLabels, order, true).flat();
-      return normalizedNewLabels.map((v) => evaluatePolynomial(coeffs, v, order));
+      try {
+        const coeffs = polynomialRegression(values, normalizedLabels, order, true).flat();
+        return normalizedNewLabels.map((v) => evaluatePolynomial(coeffs, v, order));
+      } catch (e) {
+        return Array.from({ length: newLabels.length }, () => NaN);
+      }
```
hokolomopo, line 549: try/catch should be on the whole switch instead of repeated in every switch case

src/helpers/figures/charts/chart_common.ts
```diff
@@ -535,41 +535,45 @@ export function interpolateData(
...
   const normalizedNewLabels = newLabels.map((v) => (v - labelMin) / labelRange);
-  switch (config.type) {
-    case "polynomial": {
-      const order = config.order ?? 2;
-      if (order === 1) {
-        return predictLinearValues([values], [normalizedLabels], [normalizedNewLabels], true)[0];
+  try {
+    switch (config.type) {
+      case "polynomial": {
+        const order = config.order ?? 2;
```
rrahir, line 541: shouldn't this crash early if the order is not provided?

LucasLefevre (commented):
robodoo r+

PR #5229, opened 2024-11-19
[IMP] sort: prevent sorting on range with array formula

hokolomopo:
### [IMP] sort: prevent sorting on range with array formula

Sorting on an array formula does not work since the spreaded values
cannot be reordered. But the "sort" button was still enabled,
leading to cases where the user tried to sort on an array formula,
which would fail without any feedback.

This commits displays a warning message when the user tries to sort
on a range with an array formula.

### [FIX] sort: use allowDispatch instead of throwing

The sort plugin would throw an error if the sort anchor wasn't in the
sort zone, instead of using the standard allowDispatch behaviour of
returning a DispatchResult.

Task: [4342539](https://www.odoo.com/web#id=4342539&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/tests/sort_plugin.test.ts
+++ b/tests/sort_plugin.test.ts
@@ -400,6 +400,15 @@ describe("Sorting allowDispatch", () => {
       sort(model, { zone: "A1:A5", anchor: "A1", direction: "ascending" })
     ).toBeCancelledBecause(CommandResult.InvalidSortZone);
   });
+
+  test("Sort with array formula", () => {
+    setCellContent(model, "A1", "8");
+    setCellContent(model, "A2", "4");
+    setCellContent(model, "A3", "=FILTER(A1:A2, A1:A2 > 1)");
+    expect(
+      sort(model, { zone: "A1:A4", anchor: "A1", direction: "ascending" })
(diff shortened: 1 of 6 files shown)


tests/sort_plugin.test.ts
```diff
@@ -375,18 +375,31 @@ describe("Basic Sorting", () => {
   });
 });
 
-describe("Trigger sort generic errors", () => {
+describe("Sorting allowDispatch", () => {
   const sheetId: UID = "sheet2";
+  beforeEach(() => {
+    model = new Model({ sheets: [{ id: sheetId, colNumber: 1, rowNumber: 6 }] });
```
LucasLefevre: I wouldn't restrict the number of cols/rows. Even more to only one column. It's error prone
Here it might be ok because there're only 3 tests and it's not "far", but if there are more tests in the future...

And if one (future) test requires 2 columns, it would need to redefine the model

LucasLefevre:
robodoo rebase-ff

PR #5248, opened 2024-11-21
[FIX] helpers: Add generic helper to sanitize sheet names

rrahir:
Some commands or functionalities can dispatch the creation of new sheets. Unfortunately, the sheet names cannot contain specific characters. This revision introduces a generic helper to sanitize the names before using it inside the CREATE_SHEET command.

Task: 4347719

## Description:

Task: [4347719](https://www.odoo.com/odoo/2328/tasks/4347719)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/misc.ts
+++ b/src/helpers/misc.ts
@@ -1,10 +1,12 @@
 //------------------------------------------------------------------------------
 // Miscellaneous
 //------------------------------------------------------------------------------
-import { NEWLINE } from "../constants";
+import { FORBIDDEN_SHEETNAME_CHARS_IN_EXCEL_REGEX, NEWLINE } from "../constants";
 import { ConsecutiveIndexes, Lazy, UID } from "../types";
 import { SearchOptions } from "../types/find_and_replace";
 import { Cloneable, DebouncedFunction } from "./../types/misc";
+
+const sanitizeSheetNameRegex = new RegExp(FORBIDDEN_SHEETNAME_CHARS_IN_EXCEL_REGEX, "g");
 
 /**
  * Remove quotes from a quoted string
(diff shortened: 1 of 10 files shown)


src/helpers/misc.ts
```diff
@@ -1,11 +1,13 @@
 //------------------------------------------------------------------------------
 // Miscellaneous
 //------------------------------------------------------------------------------
-import { NEWLINE } from "../constants";
+import { FORBIDDEN_IN_EXCEL_REGEX, NEWLINE } from "../constants";
```
hokolomopo, line 4: we might want to take this opportunity to rename this constant ? The name kinda sucks, it doesn't tell you what are forbidden (characters) or where (in sheet names).

src/helpers/misc.ts
```diff
@@ -109,6 +111,11 @@ export function getCanonicalSheetName(sheetName: string): string {
   return sheetName;
 }
 
+/** Replace the excel-excluded characters of a sheetName */
+export function sanitizeSheetName(sheetName: string, replacementChar: string = " "): string {
+  return sheetName.replace(sanitizeSheetNameRegex, replacementChar);
```
hokolomopo, line 116: We could use replaceAll rather than re-creating a regex with the global flag

hokolomopo (commented):
:steam_locomotive: :steam_locomotive:

src/helpers/misc.ts
rrahir: the regex is more efficient: https://www.measurethat.net/Benchmarks/Show/2396/0/replaceall-vs-regex-replace#latest_results_block

LucasLefevre (commented):
robodoo r+

PR #5250, opened 2024-11-22
[FIX] chart: labels should not be truncated in tooltip

hokolomopo:
## Description

It make sense to truncate labels in the chart axis ticks & in the legend, but it should not be truncated in the tooltip. Otherwise, the user has no way to know the full label.

Task: [4351338](https://www.odoo.com/odoo/2328/tasks/4351338)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/figures/charts/chart_common.ts
+++ b/src/helpers/figures/charts/chart_common.ts
@@ -1,4 +1,5 @@
 import { transformZone } from "../../../collaborative/ot/ot_helpers";
+import { MAX_CHAR_LABEL } from "../../../constants";
 import { _t } from "../../../translation";
 import {
   AddColumnsRowsCommand,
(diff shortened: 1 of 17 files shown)


rrahir:
requires a rebase :/

src/helpers/figures/charts/chart_common.ts
```diff
@@ -440,3 +442,13 @@ export function getPieColors(colors: ColorGenerator, dataSetsValues: DatasetValu
 
   return pieColors;
 }
+
+export function truncateLabel(label: string | undefined): string {
```
rrahir, line 446: seems to me that it's more 'formatTickValue' that should move to chart_ui_common rather than the other way around as it's clearly 'UI' but that split in the helpers seems already shady atm

rrahir:
merged with https://github.com/odoo/o-spreadsheet/pull/5246

PR #5252, opened 2024-11-22
[FIX] composer: overflow on composer at end of screen

hokolomopo:
## Description

The button to close the composer's formula assistant would overflow for the viewport and make an additional scrollbar appear when it's at the end of the screen (eg. standalone composer in the CF color scale panel).

This was due to two things:
1) there was a comment with `//`, which is invalid css and disabled the `width` property of the button
2) we were using a `CLOSE_ICON_RADIUS` to compute the position of the formula assistant. But this constant wasn't actually used in the CSS, so the button could have an arbitrary width unrelated to the constant. (in practice it was 18.4px vs 18px, but this would change if we changed the font size).

Task: [4315958](https://www.odoo.com/odoo/2328/tasks/4315958)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/composer/composer/composer.ts
+++ b/src/components/composer/composer/composer.ts
@@ -92,10 +92,10 @@ css/* scss */ `
     }
 
     .fa-stack {
-      // reset stack size which is doubled by default
-      width: 1em;
-      height: 1em;
-      line-height: 1em;
+      /* reset stack size which is doubled by default */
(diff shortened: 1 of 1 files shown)


rrahir:
@robodoo r+

PR #5256, opened 2024-11-22
[FIX] composer: topbar composer z-index

hokolomopo:
## Description

The topbar composer has a big z-index, so its formula assistant is displayed above the rest.

But this z-index is useless when the composer is not focused. In fact, it causes problems in further versions where we end up with grid popovers being displayed below the composer.

Also the css to color the border blue when the composer is focused wasn't working.

Task: [4246966](https://www.odoo.com/odoo/2328/tasks/4246966)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/composer/composer/composer.ts
+++ b/src/components/composer/composer/composer.ts
@@ -86,11 +86,11 @@ css/* scss */ `
   }
 
   /* Custom css to highlight topbar composer on focus */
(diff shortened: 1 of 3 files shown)


rrahir:
@robodoo r+

PR #5263, opened 2024-11-25
[IMP] chart: keep data series when switching chart types

hokolomopo:
## Description

Currently if you have a bar chart with 3 ranges, switch to a gauge chart that only accept a single range, then back to a bar chart you will lose the 2 other ranges.

This commit fixes that. The ranges are kept in the store so we can switch between chart types without losing data.

There is one issue however: we do not make any difference between ranges lost because the chart type changed, and ranges lost because the user manually removed them. So if a user change chart type => remove a range => change chart type again, the removed range will be back.

This commit also fixed another issue that the chart creation context was not dependant on which figure was edited.

Task: [4282798](https://www.odoo.com/odoo/2328/tasks/4282798)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


LucasLefevre (commented):
robodoo r+

PR #5268, opened 2024-11-26
[FIX] chart: trend line tooltip for scatter/line charts

hokolomopo:
## Description

For scatter/line chart whose labels were numbers, the trend line tooltip was showing something like `(15, 9)`. 9 was the correct Y value, but the 15 was referring to the fact that it was the 15th point in the generated trend line. Which made no sense to the user.

This commit changes the tooltip to show only the Y value.

Task: [4279972](https://www.odoo.com/odoo/2328/tasks/4279972)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


rrahir:
@robodoo r+

PR #5271, opened 2024-11-26
[FIX] pivot: prevent duplicating pivot in error

hokolomopo:
## Description

Trying to duplicate a pivot in error would raise a traceback when trying to call pivot.getTableStructure().

Task: [4361353](https://www.odoo.com/odoo/2328/tasks/4361353)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_feature/insert_pivot.ts
+++ b/src/plugins/ui_feature/insert_pivot.ts
@@ -5,11 +5,25 @@ import { SpreadsheetPivotTable } from "../../helpers/pivot/table_spreadsheet_piv
 import { getZoneArea, positionToZone } from "../../helpers/zones";
 import { _t } from "../../translation";
 import { CellPosition, HeaderIndex, PivotTableCell, PivotTableData, UID, Zone } from "../../types";
-import { Command } from "../../types/commands";
+import { Command, CommandResult } from "../../types/commands";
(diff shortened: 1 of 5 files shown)


src/plugins/ui_feature/insert_pivot.ts
```diff
@@ -4,12 +4,26 @@ import { SpreadsheetPivotTable } from "../../helpers/pivot/table_spreadsheet_piv
...
+import { Command, CommandResult } from "../../types/commands";
 import { UIPlugin } from "../ui_plugin";
 
 export class InsertPivotPlugin extends UIPlugin {
   static getters = [] as const;
 
+  allowDispatch(cmd: Command) {
+    switch (cmd.type) {
+      case "DUPLICATE_PIVOT_IN_NEW_SHEET":
+        if (!this.getters.isExistingPivot(cmd.pivotId)) {
```
rrahir, line 16: I might be restarting an old discussion (sorry if I do) but the "insert_new_pivot" command pretty much does the same thing but does not crash because the 'getTableStructure' won't crash before the pivot is evaluated.
Couldn't "getTableStructure" not throw at all?

rrahir:
@robodoo r+

PR #5273, opened 2024-11-27
[FIX] font_size_editor: dropdown and outline issues

dhrp-odoo:
### Issue 1:

Adjusting the font size with the **mouse wheel** while the dropdown was open caused unexpected behavior:
- Clicking a font size applied the **wheel-adjusted** value instead of the selected one, closing the dropdown.
- Keeping the dropdown open resulted in the font size being set twice.
- Disabled the **mouse wheel event** to prevent accidental changes.

### Issue 2:
The focus outline for the font size editor was not visible due to a global CSS reset in Odoo.
- To align with the design, the outline has been removed in o-spreadsheet.

Task: [4357023](https://www.odoo.com/odoo/2328/tasks/4357023)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/font_size_editor/font_size_editor.ts
+++ b/src/components/font_size_editor/font_size_editor.ts
@@ -1,6 +1,6 @@
 import { Component, useExternalListener, useRef, useState } from "@odoo/owl";
 import { setStyle } from "../../actions/menu_items_actions";
-import { DEFAULT_FONT_SIZE, FONT_SIZES, SELECTION_BORDER_COLOR } from "../../constants";
+import { DEFAULT_FONT_SIZE, FONT_SIZES } from "../../constants";
 import { clip } from "../../helpers/index";
 import { SpreadsheetChildEnv } from "../../types/index";
 import { css } from "../helpers/css";
@@ -20,7 +20,7 @@ css/* scss */ `
   .o-font-size-editor {
     height: calc(100% - 4px);
     input.o-font-size {
-      outline-color: ${SELECTION_BORDER_COLOR};
+      outline: none;
       height: 20px;
       width: 23px;
     }
--- a/tests/top_bar_component.test.ts
+++ b/tests/top_bar_component.test.ts
@@ -340,6 +340,18 @@ describe("TopBar component", () => {
     expect(getStyle(model, "A1").fontSize).toBe(8);
   });
 
+  test("prevents default behavior of mouse wheel event on font size input", async () => {
+    await mountParent();
+    const fontSizeInput = fixture.querySelector("input.o-font-size") as HTMLInputElement;
+
+    const event = new WheelEvent("wheel", { deltaY: 100 });
+    const preventDefaultSpy = jest.spyOn(event, "preventDefault");
+
+    fontSizeInput.dispatchEvent(event);
+
+    expect(preventDefaultSpy).toHaveBeenCalled();
+  });
+
   describe("horizontal align", () => {
     test.each([
       ["Left (Ctrl+Shift+L)", { align: "left" }],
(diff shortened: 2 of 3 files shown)


src/components/font_size_editor/font_size_editor.ts
```diff
@@ -110,6 +113,13 @@ export class FontSizeEditor extends Component<Props, SpreadsheetChildEnv> {
       this.props.onToggle();
     }
   }
+
+  onInputWheel(ev: WheelEvent) {
```
hokolomopo, line 117: That doesn't really fix the issue does it ? We can still have the same bug (onchange triggering when clicking on a menu item) if we input a value with the keyboard, don't confirm the input, then click a menu item.

Maybe we should go the other direction, and make the menu item change the input value ? (idk if the order of events allow this tho)

src/components/font_size_editor/font_size_editor.ts
```diff
@@ -20,9 +20,12 @@ css/* scss */ `
   .o-font-size-editor {
     height: calc(100% - 4px);
     input.o-font-size {
-      outline-color: ${SELECTION_BORDER_COLOR};
       height: 20px;
       width: 23px;
+
+      &:focus {
+        outline: ${SELECTION_BORDER_COLOR} auto 1px;
+      }
```
hokolomopo, line 28: TBH I think the no outline variant is prettier. I find the blue border quite ugly. But it's less clear that the input is focused. You should ask Francois which we should use.

hokolomopo (commented):
👋

src/components/font_size_editor/font_size_editor.ts
dhrp-odoo: Input `onChange` is triggered before the `onClick` of the dropdown menu item. As a result, when the user changes the input value and then clicks a menu item, the onChange sets the font size and closes the dropdown before the onClick can be processed, leading to the click being ignored.

Can we use maybe a 100ms debounce on the `setSize`. To ensure that if consecutive calls to `onChange` and `onClick` occurs, only the last one is applied?

hokolomopo: Debounce sounds like a good idea that would solve the problem :)

tests/top_bar_component.test.ts
```diff
@@ -331,13 +332,29 @@ describe("TopBar component", () => {
...
     expect(getStyle(model, "A1").fontSize).toBe(8);
+    jest.useRealTimers();
+  });
+
+  test("Font size selection from the dropdown overrides input value while editing", async () => {
+    jest.useFakeTimers();
+    const { model } = await mountParent();
+    await click(fixture, ".o-font-size-editor");
+    setInputValueAndTrigger(".o-font-size-editor input", "80", "onlyInput");
+    await click(fixture, '.o-dropdown-content [data-size="8"]');
```
hokolomopo, line 353: Is this test failing without your fix ? I'm pretty sure JSDOM doesn't really handle focus, and onChange is never called on the input. You could probably just remove the `onlyInput` and name the test "font size selection is debounced" or something.

src/components/font_size_editor/font_size_editor.ts
```diff
@@ -52,6 +52,8 @@ export class FontSizeEditor extends Component<Props, SpreadsheetChildEnv> {
   private inputRef = useRef("inputFontSize");
   private rootEditorRef = useRef("FontSizeEditor");
 
+  private debouncedSetSize = debounce(this.setSize.bind(this), 100);
```
hokolomopo, line 55: nitpick: I'd add a small comment explaining why we need to debouce :)

hokolomopo (commented):
Seems good !

tests/top_bar_component.test.ts
dhrp-odoo: Yes, I just checked :') Updated test accordingly

rrahir:
@robodoo r+

👍

PR #5283, opened 2024-11-28
[IMP] CellPlugin: Avoid useless UPDATE_CELL

rrahir:
The CellPlugin handler for "DELETE_CONTENT" will dispatch an UPDATE_CELL to empty its content regardless of the cell. On a big zone this can end up being quite costly as we pass for no reason in all the plugins and stores to apply 0 changes, it only adds some overhead.

the allowDispatch of UPDATE_CELL handles this case but in this case it's a subcommand and the allowDispatch is bypassed.

Benchmark
=========

Measure of the execution time of `model.Dispatch` on a 26x100 cells
sheet with **styled** cells.

cells with content (formula or text)
- before: 120~140ms
- after: 120~140ms

cells without content
- before 120~140ms
- 4~9ms

Task: 4367772

## Description:

Task: [4367772](https://www.odoo.com/odoo/2328/tasks/4367772)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/core/cell.ts
+++ b/src/plugins/core/cell.ts
@@ -162,7 +162,7 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
       for (let col = zone.left; col <= zone.right; col++) {
         for (let row = zone.top; row <= zone.bottom; row++) {
           const cell = this.getters.getCell({ sheetId, col, row });
-          if (cell) {
+          if (cell?.isFormula || cell?.content) {
             this.dispatch("UPDATE_CELL", {
               sheetId: sheetId,
               content: "",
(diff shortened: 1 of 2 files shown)


src/plugins/core/cell.ts
```diff
@@ -162,7 +162,7 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
       for (let col = zone.left; col <= zone.right; col++) {
         for (let row = zone.top; row <= zone.bottom; row++) {
           const cell = this.getters.getCell({ sheetId, col, row });
-          if (cell) {
+          if (cell?.content) {
```
LucasLefevre, line 165: note that `cell?.content` can also be quite costly if it's a formula. It rebuilds the entire formula string with the ranges.
```suggestion
          if (cell?.isFormula || cell?.content) {
```

LucasLefevre (commented):
Curious to see what are the perf before/after. Did you measure it ?

src/plugins/core/cell.ts
rrahir: 👍  completely forgot it was computed. I added the benchmark in the commit message but tl;dr  from 120ms -> 5ms

LucasLefevre (approved):
robodoo r+

PR #5299, opened 2024-12-03
[FIX] pivot: apply format on measure only a basic formula

LucasLefevre:
## Description:

Let's say I have a computed measure "weighted delay:sum", on which I applied an integer format.

I later use it in a formula such as:

=PIVOT.VALUE(1,"weighted delay:sum","#user_id",A3)/PIVOT.VALUE(1,"__count:sum","#user_id",A3)

This is no longer an integer. If I change the format on this formula ☝️, the format is applied on the entire measure, which is not what I want.

With this commit, we apply the format on the measure (rather than on the cell)
if and only if the value comes from a spilled PIVOT formula.

Task: [4377411](https://www.odoo.com/odoo/2328/tasks/4377411)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_feature/format.ts
+++ b/src/plugins/ui_feature/format.ts
@@ -11,6 +11,8 @@ import {
   CellValueType,
   Command,
   Format,
+  PivotTableCell,
+  PivotValueCell,
   Position,
   SetDecimalStep,
   UID,
@@ -43,7 +45,7 @@ export class FormatPlugin extends UIPlugin {
         for (let row = zone.top; row <= zone.bottom; row++) {
           const position = { sheetId, col, row };
           const pivotCell = this.getters.getPivotCellFromPosition(position);
-          if (pivotCell.type === "VALUE") {
+          if (this.isSpilledPivotValueFormula(position, pivotCell)) {
             measurePositions.push(position);
             const pivotId = this.getters.getPivotIdFromPosition(position) || "";
             measuresByPivotId[pivotId] ??= new Set();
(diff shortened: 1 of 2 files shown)


src/plugins/ui_feature/format.ts
```diff
@@ -81,6 +84,29 @@ export class FormatPlugin extends UIPlugin {
...
+    position: CellPosition,
+    pivotCell: PivotTableCell
+  ): pivotCell is PivotValueCell {
+    if (pivotCell.type !== "VALUE") {
+      return false;
+    }
+    const cell = this.getters.getCell(position);
+    if (!cell?.isFormula) {
+      return true;
+    }
```
hokolomopo, line 102: "If the cell is not a formula, the cell is a basic pivot value formula" ?

src/plugins/ui_feature/format.ts
```diff
@@ -43,7 +46,7 @@ export class FormatPlugin extends UIPlugin {
         for (let row = zone.top; row <= zone.bottom; row++) {
           const position = { sheetId, col, row };
           const pivotCell = this.getters.getPivotCellFromPosition(position);
-          if (pivotCell.type === "VALUE") {
+          if (this.isBasicPivotValueFormula(position, pivotCell)) {
```
hokolomopo, line 49: TBH I would just disable the functionnality on non spreaded pivot formula. It doens't make much sense for splitted formulas to me and sounds very much like black magic that people will not really understand. 

I don't expect changing the format on a single =PIVOT.VALUE cell that I put alone on a random sheet to affect the format of the whole pivot on all the other sheet. The point of having splitted pivot is that I can do things on it independantly from the pivot no ?

hokolomopo (commented):
👋

rrahir:
@robodoo r+

PR #5300, opened 2024-12-03
[FIX] borders: fix border on column/row deletion

fdamhaut:
- Fix issue when deleting the first/last row/colum not deleting the exterior borders

- Fix issue with wrong border movement when deleting row/col

- Fix not removing internal border when deleting row/col

Task: 4089281

## Description:

Task: [4089281](https://www.odoo.com/odoo/2328/tasks/4089281)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/core/borders.ts
+++ b/src/plugins/core/borders.ts
@@ -90,9 +90,55 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
         const elements = [...cmd.elements].sort((a, b) => b - a);
         for (const group of groupConsecutive(elements)) {
           if (cmd.dimension === "COL") {
-            this.shiftBordersHorizontally(cmd.sheetId, group[group.length - 1] + 1, -group.length);
+            if (group[0] >= this.getters.getNumberCols(cmd.sheetId)) {
+              for (let row: HeaderIndex = 0; row < this.getters.getNumberRows(cmd.sheetId); row++) {
+                this.history.update(
+                  "borders",
+                  cmd.sheetId,
+                  group[0] + 1,
+                  row,
+                  "vertical",
+                  undefined
+                );
+              }
+            }
+            if (group[group.length - 1] === 0) {
+              for (let row: HeaderIndex = 0; row < this.getters.getNumberRows(cmd.sheetId); row++) {
+                this.history.update("borders", cmd.sheetId, 0, row, "vertical", undefined);
+              }
+            }
+            const zone = this.getters.getColsZone(
+              cmd.sheetId,
+              group[group.length - 1] + 1,
+              group[0]
+            );
+            this.clearInsideBorders(cmd.sheetId, [zone]);
+            this.shiftBordersHorizontally(cmd.sheetId, group[0] + 1, -group.length);
           } else {
-            this.shiftBordersVertically(cmd.sheetId, group[group.length - 1] + 1, -group.length);
+            if (group[0] >= this.getters.getNumberRows(cmd.sheetId)) {
+              for (let col = 0; col < this.getters.getNumberCols(cmd.sheetId); col++) {
+                this.history.update(
+                  "borders",
+                  cmd.sheetId,
+                  col,
+                  group[0] + 1,
+                  "horizontal",
+                  undefined
+                );
+              }
+            }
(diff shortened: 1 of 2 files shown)


src/plugins/core/borders.ts
```diff
@@ -90,9 +90,55 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
...
+                  "borders",
+                  cmd.sheetId,
+                  group[0] + 1,
+                  row,
+                  "vertical",
+                  undefined
+                );
+              }
+            }
+            if (group[group.length - 1] == 0) {
```
hokolomopo, line 105: **always** use `===` in js. Typescript should mostly prevent issues, but `==` cast the variables before the comparison ... which lead to some interesting results. `0 == []` is true for example

src/plugins/core/borders.ts
```diff
@@ -90,9 +90,55 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
...
+              cmd.sheetId,
+              group[group.length - 1] + 1,
+              group[0]
+            );
+            this.clearInsideBorders(cmd.sheetId, [zone]);
+            this.shiftBordersHorizontally(cmd.sheetId, group[0] + 1, -group.length);
           } else {
-            this.shiftBordersVertically(cmd.sheetId, group[group.length - 1] + 1, -group.length);
+            if (group[0] == this.getters.getNumberRows(cmd.sheetId)) {
+              for (let col: HeaderIndex = 0; col < this.getters.getNumberCols(cmd.sheetId); col++) {
```
hokolomopo, line 119:
```suggestion
              for (let col = 0; col < this.getters.getNumberCols(cmd.sheetId); col++) {
```
nitpick: typing is kinda useless here

tests/plugins/borders.test.ts
```diff
@@ -533,6 +533,39 @@ describe("Grid manipulation", () => {
...
+    deleteColumns(model, ["B", "C", "D", "E"]);
+    expect(model.exportData().borders).toEqual({});
+  });
+
+  test("Remove top row => remove top border", () => {
+    setBorder(model, "top", "A1:J1");
+    deleteRows(model, [0]);
+    expect(model.exportData().borders).toEqual({});
+  });
+  test("Remove bottom row => remove bottom border", () => {
```
hokolomopo, line 554: nitpicks: missing newlienw between tests

tests/plugins/borders.test.ts
```diff
@@ -533,6 +533,39 @@ describe("Grid manipulation", () => {
     expect(getBorder(model, "C96")).toEqual({ bottom: b });
     expect(getBorder(model, "D96")).toEqual({ right: b, bottom: b });
   });
+
+  test("Remove multiple rows => remove internal borders", () => {
```
hokolomopo, line 537:
```suggestion
  test("Removing multiple rows removes the internal borders", () => {
```
nitpick. Try to do full sentences in test names where you can, it's more readable IMO

src/plugins/core/borders.ts
```diff
@@ -90,9 +90,55 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
         const elements = [...cmd.elements].sort((a, b) => b - a);
         for (const group of groupConsecutive(elements)) {
           if (cmd.dimension === "COL") {
-            this.shiftBordersHorizontally(cmd.sheetId, group[group.length - 1] + 1, -group.length);
+            if (group[0] == this.getters.getNumberCols(cmd.sheetId)) {
+              for (let row: HeaderIndex = 0; row < this.getters.getNumberRows(cmd.sheetId); row++) {
```
hokolomopo, line 94: That doesn't work if we remove the two last columns of the spreadsheet 😔 It seems a bit annoying since you compare the columns before they are removed from the spreadsheet (groups) and the numbre of columns of the sheet after the removal (`getNumberCols`, since the sheet plugin is already updated when you are here)

Same thing for the first cols and the rows

hokolomopo (commented):
Some nitpicks, and a thing that doesn't work unfortunately 😔

rrahir:
@robodoo r+

PR #5318, opened 2024-12-05
[IMP] chart: support import/export of dataset trendlines

Rachico:
This commit introduces the ability to import and export trendlines associated with datasets in chart configurations.

Task: [4319957](https://www.odoo.com/odoo/2328/tasks/4319957)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/xlsx/conversion/figure_conversion.ts
```diff
@@ -121,6 +130,30 @@ function convertExcelRangeToSheetXC(range: string, dataSetsHaveTitle: boolean):
   return getFullReference(sheetName, dataXC);
 }
 
+function convertExcelTrenline(
```
anhe-odoo, line 133: convertExcelTren**D**line

src/helpers/figures/charts/chart_common.ts
```diff
@@ -246,7 +248,11 @@ function createDataSet(
 /**
  * Transform a dataSet to a ExcelDataSet
  */
-export function toExcelDataset(getters: CoreGetters, ds: DataSet): ExcelChartDataset {
+export function toExcelDataset(
+  getters: CoreGetters,
+  ds: DataSet,
+  customDs?: CustomizedDataSet
```
anhe-odoo, line 254: Why do we now need to give the customizedDataSet to this function ? How comes that before we were able to have the correct colors, labels and axes without passing customDs here ? Isn't it possible to do it the same way for the trend line ?

Rachico: @anhe-odoo I need the trend to be add to the generated excel dataset. Therefore, I need the trend to be passed here. I can pass only the trend as a function argument instead. What do you think ?

src/helpers/figures/charts/chart_common.ts
```diff
@@ -272,6 +276,24 @@ export function toExcelDataset(getters: CoreGetters, ds: DataSet): ExcelChartDat
     };
   }
 
+  let trend: ExcelChartTrendConfiguration | undefined = undefined;
+  if (ds?.trend?.type) {
+    trend = {
+      type: CHART_TRENDLINE_TYPE_CONVERSION_MAP_REVERSE[ds.trend.type],
+      order: ds?.trend?.order,
```
anhe-odoo, line 283: You added a comment in another file saying the max degree in Excel is 6. Maybe we should check this here too ?

src/xlsx/extraction/chart_extractor.ts
```diff
@@ -139,13 +144,33 @@ export class XlsxChartExtractor extends XlsxBaseExtractor {
...
+  private extractChartTrendline(
+    chartDataElement: Element
+  ): ExcelChartTrendConfiguration | undefined {
+    const trendlineElement = this.querySelector(chartDataElement, "c:trendline");
+    if (!trendlineElement) {
+      return undefined;
+    }
+    const trendlineType = this.extractChildAttr(trendlineElement, "c:trendlineType", "val");
+    const trendlineColor = this.extractChildAttr(trendlineElement, "a:solidFill a:srgbClr", "val");
+    const trendlineOrder = this.extractChildAttr(trendlineElement, "c:order", "val");
```
anhe-odoo, line 164: Do we always need to extract the order ? Maybe we can do it only if we are in the "polynomial" case ?

src/xlsx/extraction/chart_extractor.ts
```diff
@@ -139,13 +144,33 @@ export class XlsxChartExtractor extends XlsxBaseExtractor {
...
+    const trendlineWindow = this.extractChildAttr(trendlineElement, "c:period", "val");
```
anhe-odoo, line 165: Do we always need the window, or only in the case of a moving average ?

tests/xlsx/xlsx_import.test.ts
```diff
@@ -721,18 +721,26 @@ describe("Import xlsx data", () => {
...
-        { dataRange: "Sheet1!B27:B35", backgroundColor: "#7030A0" },
+        {
+          dataRange: "Sheet1!B27:B35",
+          backgroundColor: "#7030A0",
+          trend: { type: "exponential", display: true },
+        },
         { dataRange: "Sheet1!C27:C35", backgroundColor: "#C65911" },
       ],
     ],
-  ])("Can import charts %s with dataset colors", (chartType, chartDatasets) => {
```
anhe-odoo, line 735: Can you make a separated test to check that we can only importe chart with trend line instead of modifying theses ?

src/helpers/figures/charts/chart_common.ts
```diff
@@ -272,6 +276,46 @@ export function toExcelDataset(getters: CoreGetters, ds: DataSet): ExcelChartDat
...
+    if (ds.trend.type === "polynomial" && ds.trend.order) {
+      if (ds.trend.order === 1) {
+        trend = {
+          ...trend,
+          type: "linear",
+        };
+      } else {
+        trend = {
+          ...trend,
+          order: toExcelPolynomialTrendlineOrder(ds.trend.order),
```
anhe-odoo, line 299: A function only used once doesn't seems really usefull. Maybe something like `Math.max(ds.trend.order, 6)` ? Or even with a const `MAX_EXCEL_POLYNOMIAL_DEGREE` as 6 if you want to make is clearer ?

Rachico: Fair enough!

src/helpers/figures/charts/chart_common.ts
```diff
@@ -35,6 +37,7 @@ import { rangeReference } from "../../references";
 import { getZoneArea, isFullRow, toUnboundedZone, zoneToDimension, zoneToXc } from "../../zones";
 
 export const TREND_LINE_XAXIS_ID = "x1";
+export const MAX_EXCEL_POLYNOMIAL_DEGREE = 6;
```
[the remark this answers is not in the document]
Rachico: @pro-odoo excel does not support a polynomial degree larger than 6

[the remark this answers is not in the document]
Rachico: https://www.officetooltips.com/excel_365/tips/polynomial_trend_equation_and_forecast#:~:text=The%20maximum%20degree%20of%20the%20polynomial%20trendline%20that%20can%20be%20displayed%20in%20an%20Excel%20chart%20is%206%3A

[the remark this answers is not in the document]
pro-odoo: You can add a comment with the link above, it will be enough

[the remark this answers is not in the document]
rrahir: I think it's worth linking against the actual spec about `ST_TrendlineType` and `ST_Order`. 
```xml
<xsd:simpleType name="ST_Order">
    <xsd:restriction base="xsd:unsignedByte">
        <xsd:minInclusive value="2"/>
        <xsd:maxInclusive value="6"/>
     </xsd:restriction>
</xsd:simpleType>
```
Found in the Appendix §A.5.1, which is related to section 21.2.3.50 ST_TrendlineType (Trendline Type) of the Open XML spec.

src/xlsx/functions/charts.ts
```diff
@@ -211,6 +211,63 @@ function insertTextProperties(
...
+
+function extractTrendlineCommonAttributes(trend: ExcelChartDataset["trend"]) {
+  if (!trend) {
+    return "";
+  }
+  return escapeXml/*xml*/ `
+    <c:spPr>
+      <a:ln w="19050" cap="rnd">
+          <a:solidFill>
+              <a:srgbClr val="${trend.color ? toXlsxHexColor(trend.color) : "000000"}" />
```
rrahir, line 260: if there is no color defined in the definition, we should probably apply the same logic we do for the runtime drawing and not put "black". Or maybe we don't have to specify the key.

src/xlsx/extraction/chart_extractor.ts
```diff
@@ -135,29 +141,69 @@ export class XlsxChartExtractor extends XlsxBaseExtractor {
...
+    const trendlineElement = this.querySelector(chartDataElement, "c:trendline");
+    if (!trendlineElement) {
+      return undefined;
+    }
+    const trendlineType = this.extractChildAttr(trendlineElement, "c:trendlineType", "val");
+    const trendlineColor = this.extractChildAttr(trendlineElement, "a:solidFill a:srgbClr", "val");
+    let trendlineConfig: ExcelChartTrendConfiguration = {
+      type: trendlineType ? (trendlineType.asString() as ExcelTrendlineType) : undefined,
+      color: trendlineColor ? `${toHex(trendlineColor.asString())}` : undefined,
+    };
```
hokolomopo, line 175:
```suggestion
    const trendlineConfig: ExcelChartTrendConfiguration = {
      type: this.extractChildAttr(trendlineElement, "c:trendlineType", "val")?.asString(),
      ...,
      order: this.extractChildAttr(trendlineElement, "c:order", "val")?.asNum(),
      window: ....
    };
```

do we really need all the if cases ? 

Or we could also simplfy the if cases with 

`trendlineConfig.window = this.extractChildAttr(trendlineElement, "c:period", "val")?.asNum()`

src/xlsx/extraction/chart_extractor.ts
```diff
@@ -170,6 +216,8 @@ export class XlsxChartExtractor extends XlsxBaseExtractor {
         return {
           label,
           range: this.extractChildTextContent(chartDataElement, "c:yVal c:f", { required: true })!,
+          trend: this.extractChartTrendline(chartDataElement),
+          backgroundColor: color ? `${toHex(color.toString())}` : undefined,
```
hokolomopo, line 220: same comment for color

src/helpers/figures/charts/chart_common.ts
```diff
@@ -273,6 +279,49 @@ export function toExcelDataset(getters: CoreGetters, ds: DataSet): ExcelChartDat
...
+  let trend: ExcelChartTrendConfiguration | undefined = undefined;
+  if (ds?.trend?.type) {
+    trend = {
+      type: CHART_TRENDLINE_TYPE_CONVERSION_MAP_REVERSE[ds.trend.type],
+    };
+    if (ds.trend.color) {
+      trend = {
+        ...trend,
+        color: ds.trend.color,
+      };
```
hokolomopo, line 291:
```suggestion
      trend.color = ds.trend.color;
```
are all the spread of this function really necessary ? There's no need to create a new object every time we want to set a property value

src/helpers/figures/charts/chart_common.ts
```diff
@@ -273,6 +279,49 @@ export function toExcelDataset(getters: CoreGetters, ds: DataSet): ExcelChartDat
...
+          type: "linear",
+        };
+      } else {
+        trend = {
+          ...trend,
+          order:
+            ds.trend.order > MAX_XLSX_POLYNOMIAL_DEGREE
+              ? MAX_XLSX_POLYNOMIAL_DEGREE
+              : ds.trend.order,
+        };
```
hokolomopo, line 306:
```suggestion
trend.order = Math.min(ds.trend.order, MAX_XLSX_POLYNOMIAL_DEGREE)
```

src/helpers/figures/charts/chart_common.ts
```diff
@@ -273,6 +279,49 @@ export function toExcelDataset(getters: CoreGetters, ds: DataSet): ExcelChartDat
...
+      }
+    }
+    if (ds.trend.type === "trailingMovingAverage" && ds.trend.window) {
+      trend = {
+        ...trend,
+        window: ds.trend.window,
+      };
+    }
+  }
+  if (trend) {
```
hokolomopo, line 316: just add the (possibly undefined) trend to the original return, there's no need to create a new return case with everything the same just to add the trend.

This whole block generating the trendLine can probably be simplified to 10 or so lines of code instead of 40

src/helpers/figures/charts/chart_common.ts
```diff
@@ -36,6 +38,8 @@ import { rangeReference } from "../../references";
 import { getZoneArea, isFullRow, toUnboundedZone, zoneToDimension, zoneToXc } from "../../zones";
 
 export const TREND_LINE_XAXIS_ID = "x1";
+// The possible values for the XLSX polynomial trendline order are defined by the ST_Order simple type (§21.2.3.29)
+export const MAX_XLSX_POLYNOMIAL_DEGREE = 6;
```
hokolomopo, line 42: nitpick: i'd put the variable inside a file with other xlsx constants

src/components/side_panel/chart/building_blocks/series_design/series_with_axis_design_editor.ts
```diff
@@ -112,9 +112,14 @@ export class SeriesWithAxisDesignEditor extends Component<Props, SpreadsheetChil
         break;
       case "exponential":
       case "logarithmic":
-      case "trailingMovingAverage":
         config = { type };
         break;
+      case "trailingMovingAverage":
+        config = {
+          type,
+          window: this.defaultWindowSize,
```
hokolomopo, line 120: This seems not needed ? The trailing average trendline is typed to have the window being optional. And I don't think this should have anything to do with a commit handling the xlsx export/import

tests/xlsx/__snapshots__/xlsx_export.test.ts.snap
```diff
@@ -35060,7 +35075,7 @@ exports[`Test XLSX export references with headers should be converted to referen
                         <!-- x-coordinate values -->
                         <c:numRef>
                             <c:f>
-                                Sheet1!B3:B100
+                                Sheet1!$B$3:$B$100
```
hokolomopo, line 35078: why are those suddenly fixed ranges ? I don't think exporting a trendline should change that

tests/xlsx/xlsx_import.test.ts
```diff
@@ -739,18 +739,75 @@ describe("Import xlsx data", () => {
     [
       "line",
       [
-        { dataRange: "Sheet1!B26:B35", backgroundColor: "#7030A0" },
+        {
+          dataRange: "Sheet1!B26:B35",
+          backgroundColor: "#7030A0",
+        },
```
hokolomopo, line 745: useless change ? same for the bar chart

tests/xlsx/xlsx_import.test.ts
```diff
@@ -763,7 +820,10 @@ describe("Import xlsx data", () => {
       "bar",
       "#fff",
       [
-        { dataRange: "Sheet1!B27:B35", backgroundColor: "#7030A0" },
+        {
+          dataRange: "Sheet1!B27:B35",
+          backgroundColor: "#7030A0",
+        },
```
hokolomopo, line 826: Is your prettier drunk ? 😅 Why all these changes 🙈

tests/xlsx/xlsx_import.test.ts
```diff
@@ -739,18 +739,75 @@ describe("Import xlsx data", () => {
...
+      "line",
+      [
+        {
+          dataRange: "Sheet1!B26:B35",
+          backgroundColor: "#7030A0",
+          trend: { type: "polynomial", order: 2, display: true, color: "#ED7D31" },
+        },
+        { dataRange: "Sheet1!C26:C35", backgroundColor: "#C65911" },
+      ],
+    ],
```
hokolomopo, line 777:
```suggestion
    [
      "line",
      [{ trend: { type: "polynomial", order: 2, display: true, color: "#ED7D31" } },],
    ],
```
you could simplify the test a bit by only testing the trendlines

hokolomopo (commented):
👋

src/xlsx/functions/charts.ts
```diff
@@ -211,6 +211,77 @@ function insertTextProperties(
...
+function extractTrendlineCommonAttributes(
+  trend: ExcelChartDataset["trend"],
+  dataSetColor: XlsxHexColor
+): XMLString {
+  if (!trend) {
+    return escapeXml/*xml*/ ``;
+  }
+  return escapeXml/*xml*/ `
+    <c:spPr>
+      <a:ln w="19050" cap="rnd">
```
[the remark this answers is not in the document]
Rachico: @hokolomopo  Excel defaults to `width=19050` i.e `1.5pt` for visually good line width. Otherwise, we would have a quite thin and ugly trend line.

src/helpers/figures/charts/chart_common.ts
Rachico: We cannot do `trend.color = ds.trend.color` because the `color` property is readonly

Rachico:
@hokolomopo It should be good now

src/components/side_panel/chart/building_blocks/series_design/series_with_axis_design_editor.ts
hokolomopo: Yeah it's now on a separate commit, but it doens't really address the issue. window is still typed as possibly undefined in a trendline. Either 1) it should be required, typed as such and we should do a migration to ensure it's always defined or 2) your code should also work when window is undefined, and defaults to `DEFAULT_WINDOW_SIZE` size. 

2) sounds way better

src/xlsx/functions/charts.ts
```diff
@@ -211,6 +211,72 @@ function insertTextProperties(
...
+  trend: ExcelChartDataset["trend"],
+  dataSetColor: XlsxHexColor
+): XMLString {
+  if (!trend) {
+    return escapeXml/*xml*/ ``;
+  }
+  const attrs: XMLAttributes = [
+    [
+      "val",
+      trend.color ? toXlsxHexColor(trend.color).slice(0, 6) : getTrendlineColor(dataSetColor),
```
hokolomopo, line 258: `toXlsxHexColor(trend.color).slice(0, 6)` is wrong, xlsx color have the alpha as the first two digits (AARRGGBB). You will slice the blue channel instead of the alpha channel here.

tests/xlsx/xlsx_import.test.ts
```diff
@@ -772,7 +797,10 @@ describe("Import xlsx data", () => {
       "combo",
       "#fff",
       [
-        { dataRange: "Sheet1!B27:B35", backgroundColor: "#1F77B4" },
+        {
+          dataRange: "Sheet1!B27:B35",
+          backgroundColor: "#1F77B4",
+        },
```
hokolomopo, line 803: useless change

hokolomopo (commented):
👋

src/xlsx/functions/charts.ts
```diff
@@ -253,10 +253,7 @@ function extractTrendlineCommonAttributes(
     return escapeXml/*xml*/ ``;
   }
   const attrs: XMLAttributes = [
-    [
-      "val",
-      trend.color ? toXlsxHexColor(trend.color).slice(0, 6) : getTrendlineColor(dataSetColor),
-    ],
+    ["val", trend.color ? toXlsxHexColor(trend.color).slice(-6) : getTrendlineColor(dataSetColor)],
```
rrahir, line 256: this is in the wrong commit

src/components/side_panel/chart/building_blocks/series_design/series_with_axis_design_editor.ts
rrahir: not sure of what this commit does. Is it a UI fix?  if so it probably does not belong to this PR but the content seems deeply intertwinded with excel export / import so I don't really udnerstand what it should do. Is the split between the commits correctly done?

src/xlsx/functions/charts.ts
Rachico: My bad, fixed now 👍

src/components/side_panel/chart/building_blocks/series_design/series_with_axis_design_editor.ts
rrahir: Then I think it's better not to make this modification.
If I understand correctly, the window is currently not required and the code can work as it is.  If you change the entry point that allowed the window to be undefined, we risk a future situation  where we will not understand how a user was able to generate a trendline with a window undefined since the code will show that it's not possible. Going back to the commit history to find this specific modification could take a lot of time for nothing.

I think that this modification should come in place if we decide that having trend.window undefined was a mistake and that we need to make a migration to make things right.

rrahir:
@robodoo r+

PR #5320, opened 2024-12-05
[IMP] model: create snapshot when loading xlsx file

Rachico:
Before this commit, when uploading an XLSX file, the code in the Odoo repository would create a spreadsheet document with data that is encoded in xml. Whenever the file is loaded, we parse the stringified xml to convert it to the o-spreadsheet json structure. The problem is that this conversion can be quite slow and costly when the XLSX file is large.

This commit mitigates the issue by saving a snapshot once the XLSX file is loaded for the first time.

Task: [4080514](https://www.odoo.com/odoo/2328/tasks/4080514)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/model.ts
+++ b/src/model.ts
@@ -281,7 +281,7 @@ export class Model extends EventBus<any> implements CommandDispatcher {
 
     this.joinSession();
 
-    if (config.snapshotRequested) {
+    if (config.snapshotRequested || (data["[Content_Types].xml"] && !this.getters.isReadonly())) {
       const startSnapshot = performance.now();
       console.debug("Snapshot requested");
       this.session.snapshot(this.exportData());
--- a/tests/model/model.test.ts
+++ b/tests/model/model.test.ts
@@ -1,9 +1,12 @@
 import { CommandResult, CorePlugin } from "../../src";
+import { MESSAGE_VERSION } from "../../src/constants";
 import { toZone } from "../../src/helpers";
 import { Model, ModelConfig } from "../../src/model";
 import { corePluginRegistry, featurePluginRegistry } from "../../src/plugins/index";
 import { UIPlugin } from "../../src/plugins/ui_plugin";
 import { Command, CommandTypes, CoreCommand, DispatchResult, coreTypes } from "../../src/types";
+import { MockTransportService } from "../__mocks__/transport_service";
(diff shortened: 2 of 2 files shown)


src/model.ts
```diff
@@ -281,7 +281,7 @@ export class Model extends EventBus<any> implements CommandDispatcher {
 
     this.joinSession();
 
-    if (config.snapshotRequested) {
+    if (config.snapshotRequested || data["[Content_Types].xml"]) {
```
LucasLefevre, line 284: it should not snapshot if the spreadsheet is in read-only mode (because the user most cannot update the spreadsheet in any way)

tests/model/model.test.ts
```diff
@@ -366,4 +369,22 @@ describe("Model", () => {
       ],
     });
   });
+
+  test("snapshot when importing xlsx file", async () => {
+    jest.spyOn(console, "warn").mockImplementation(() => {});
```
pro-odoo, line 374: Is there a reason ?

src/model.ts
```diff
@@ -281,7 +281,10 @@ export class Model extends EventBus<any> implements CommandDispatcher {
 
     this.joinSession();
 
-    if (config.snapshotRequested) {
+    if (
+      config.snapshotRequested ||
+      (data["[Content_Types].xml"] && this.config.mode !== "readonly")
```
pro-odoo, line 286: Add a test for the readonly flag

pro-odoo (changes requested):

tests/model/model.test.ts
rrahir: there is a global handler to make the tests fail in case of console.error or console.warn

pro-odoo: I know this global handler, but if I comment this line, the test does not fail. Expected ?

Rachico: That's weird, it was failing before 🤷

src/model.ts
LucasLefevre, line 286: mode dashboard is also considered readonly
```suggestion
      (data["[Content_Types].xml"] && !this.this.getters.isReadonly())
```

LucasLefevre (changes requested):

LucasLefevre (commented):
robodoo r+

PR #5327, opened 2024-12-06
[FIX] charts: fix logarithmic dataset trendline

Rachico:
Logarithmic trendline in charts was calculated incorrectly for datasets with null values. This commit fixes the issue by ensuring that only valid, positive values are considered for logarithmic dataset trend computation.

Task: [4385057](https://www.odoo.com/odoo/2328/tasks/4385057)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/figures/charts/runtime/chart_data_extractor.ts
+++ b/src/helpers/figures/charts/runtime/chart_data_extractor.ts
@@ -369,10 +369,18 @@ function interpolateData(
         )[0];
       }
       case "logarithmic": {
+        const positiveLogValues: number[] = [];
+        const filteredLabels: number[] = [];
+        for (let i = 0; i < values.length; i++) {
+          if (normalizedLabels[i] > 0) {
+            positiveLogValues.push(values[i]);
+            filteredLabels.push(Math.log(normalizedLabels[i]));
+          }
+        }
         return predictLinearValues(
-          [values],
-          logM([normalizedLabels]),
-          logM([normalizedNewLabels]),
+          [positiveLogValues],
+          [filteredLabels],
+          logM([normalizedNewLabels.slice(1)]), // We remove the first element to avoid log(0)
           true
         )[0];
       }
--- a/tests/figures/chart/chart_plugin.test.ts
+++ b/tests/figures/chart/chart_plugin.test.ts
@@ -3472,6 +3472,42 @@ describe("trending line", () => {
   });
 });
 
+test("logarithmic trending line", () => {
+  // prettier-ignore
+  setGrid(model, {
+      B1: "12", C1: "34",
+      B2: "45", C2: "23",
+      B3: "32", C3: "31",
+      B4: "29", C4: "54",
+      B5: "16", C5: "11",
+      B6: "19", C6: "89",
+      B7: "23", C7: "43",
+      B8: "64", C8: "42",
+    });
+  createChart(
+    model,
+    {
+      type: "scatter",
+      dataSets: [{ dataRange: "C1:C8", trend: { display: true, type: "logarithmic" } }],
+      labelRange: "B1:B8",
+      labelsAsText: false,
+      dataSetsHaveTitle: false,
+    },
+    "1"
+  );
+  function roundToFourDecimals(value) {
+    return Math.round(value * 10000) / 10000;
+  }
+  let runtime = model.getters.getChartRuntime("1") as LineChartRuntime;
+  // We round up to 4 decimals to avoid floating point errors
+  expect(runtime.chartJsConfig.data.datasets[1].data.map(roundToFourDecimals)).toEqual([
+    48.0254, 46.2824, 45.2628, 44.5393, 43.9782, 43.5197, 43.1321, 42.7963, 42.5001, 42.2351,
+    41.9954, 41.7766, 41.5754, 41.389, 41.2155, 41.0532, 40.9007, 40.757, 40.621, 40.4921, 40.3694,
+    40.2524, 40.1406, 40.0336, 39.9309, 39.8323, 39.7374, 39.6459, 39.5577, 39.4724, 39.39, 39.3101,
+    39.2328, 39.1577, 39.0848, 39.014, 38.945, 38.878, 38.8127, 38.749,
+  ]);
+});
+
 test("moving average trending line", () => {
   // prettier-ignore
   setGrid(model, {


src/helpers/figures/charts/runtime/chart_data_extractor.ts
```diff
@@ -402,10 +402,18 @@ function interpolateData(
         )[0];
       }
       case "logarithmic": {
+        const logValues: number[] = [];
+        const filteredLabels: number[] = [];
+        for (let i = 0; i < values.length; i++) {
+          if (normalizedLabels[i] > 0) {
+            logValues.push(values[i]);
```
anhe-odoo, line 409:
```js
filteredValues.push(values[i]);
logLabels.push(Math.log(normalizedLabels[i]));
```

src/helpers/figures/charts/runtime/chart_data_extractor.ts
```diff
@@ -402,10 +402,18 @@ function interpolateData(
...
       }
       case "logarithmic": {
+        const logLabels: number[] = [];
+        const filteredValues: number[] = [];
+        for (let i = 0; i < values.length; i++) {
+          if (normalizedLabels[i] > 0) {
+            logLabels.push(values[i]);
+            filteredValues.push(Math.log(normalizedLabels[i]));
+          }
+        }
```
anhe-odoo, line 412: Just a small though : here we can check if we have some values in `filteredValues` or if it's empty, cause as it's empty we can simply return an empty array. Same comment can be applied to the line 398 :)

src/helpers/figures/charts/runtime/chart_data_extractor.ts
```diff
@@ -402,10 +402,18 @@ function interpolateData(
         )[0];
       }
       case "logarithmic": {
+        const logLabels: number[] = [];
+        const filteredValues: number[] = [];
+        for (let i = 0; i < values.length; i++) {
+          if (normalizedLabels[i] > 0) {
+            logLabels.push(values[i]);
```
rrahir, line 409: the variable names are inversed ^^

src/helpers/figures/charts/runtime/chart_data_extractor.ts
```diff
@@ -402,10 +402,18 @@ function interpolateData(
...
+            filteredValues.push(Math.log(normalizedLabels[i]));
+          }
+        }
         return predictLinearValues(
-          [values],
-          logM([normalizedLabels]),
-          logM([normalizedNewLabels]),
+          [logLabels],
+          [filteredValues],
```
rrahir, line 415:
```suggestion
          [logM(filteredValues)],
```
you can keep the current call to logM, iI think it's more readable like this instead of having another point where we apply logarithms

tests/figures/chart/chart_plugin.test.ts
```diff
@@ -3491,6 +3491,42 @@ describe("trending line", () => {
...
+      labelsAsText: false,
+      dataSetsHaveTitle: false,
+    },
+    "1"
+  );
+  function roundToFourDecimals(value) {
+    return Math.round(value * 10000) / 10000;
+  }
+  let runtime = model.getters.getChartRuntime("1") as LineChartRuntime;
+  // We round up to 4 decimals to avoid floating point errors
```
rrahir, line 3521: 👍

rrahir:
should aim for 18.0

tests/figures/chart/chart_plugin.test.ts
rrahir: maybe worth adding a test that shows that we ignore non-positive values?

rrahir:
The PR should target version 18.0, where the bug originally appeared. You can retarget the pull request on Github in order to keep the comments we already left :)

tests/figures/chart/chart_plugin.test.ts
Rachico: @rrahir I think with the new normalization range, this test is not needed anymore as we do not slice the array of values. Let me know what you think :)

src/helpers/figures/charts/chart_common.ts
```diff
@@ -532,10 +532,19 @@ export function interpolateData(
...
   }
   const labelMin = Math.min(...labels);
-  const labelMax = Math.max(...labels);
-  const labelRange = labelMax - labelMin;
-  const normalizedLabels = labels.map((v) => (v - labelMin) / labelRange);
-  const normalizedNewLabels = newLabels.map((v) => (v - labelMin) / labelRange);
+  // We change the normalization technique for logarithmic trend lines for a better fit
+  const labelMax =
+    config.type === "logarithmic"
+      ? Math.max(...labels.map((l) => Math.abs(l)))
```
anhe-odoo, line 538: Nitpick : You can use `labels.map(Math.ab)` here to avoid creating a lambda :)

src/helpers/figures/charts/chart_common.ts
```diff
@@ -532,10 +532,19 @@ export function interpolateData(
     return [];
   }
   const labelMin = Math.min(...labels);
```
anhe-odoo, line 534: We don't need this anymore in the case of a logarithmic trend line, so it should be better to have it in a separated if (see below :) )

src/helpers/figures/charts/chart_common.ts
```diff
@@ -532,10 +532,19 @@ export function interpolateData(
...
-  const labelMax = Math.max(...labels);
-  const labelRange = labelMax - labelMin;
-  const normalizedLabels = labels.map((v) => (v - labelMin) / labelRange);
-  const normalizedNewLabels = newLabels.map((v) => (v - labelMin) / labelRange);
+  // We change the normalization technique for logarithmic trend lines for a better fit
```
anhe-odoo, line 535: Maybe you can explain a bit why it's better for the logarithmic model ? And also update your commit message as you're not filtering the "zeros" values anymore

anhe-odoo (changes requested):

LucasLefevre (commented):
The commit message (and the task description) doesn't explain well what the issue is and specifically how to reproduce it. I had to ask Anthony

> with null values

I originally thought it was null values coming from empty cells, but apparently, it's zeros coming from the normalization.

Could you add a clear list of steps to reproduce the issue and explain a bit what happens?
Thanks!

src/helpers/figures/charts/chart_common.ts
```diff
@@ -531,11 +531,26 @@ export function interpolateData(
...
+    const labelMax = Math.max(...labels.map(Math.abs));
+    normalizedLabels = labels.map((l) => l / labelMax);
+    normalizedNewLabels = newLabels.map((l) => l / labelMax);
+  } else {
+    const labelMax = Math.max(...labels);
+    const labelMin = Math.min(...labels);
+    const labelRange = labelMax - labelMin;
+    normalizedLabels = labels.map((l) => (l - labelMax) / labelRange);
+    normalizedNewLabels = newLabels.map((l) => (l - labelMax) / labelRange);
+  }
```
LucasLefevre, line 553: I think it's now a good idea to extract the normalization into a function

LucasLefevre (commented):
nitpicking :wink:

src/helpers/figures/charts/chart_common.ts
Rachico: But the logic is only being used once 🤔 , init ?

LucasLefevre (approved):
robodoo r+

PR #5329, opened 2024-12-06
[FIX] pivot: mouse events do not work in pivot text input

hokolomopo:
## Description

We added a behaviour to select the whole text of the measure name input when the input is focused. We select the whole input text on `focus` event, but we had to stop/prevent the `mouseup` events, otherwise the event would sometime unselect the text.

It turns out (surprise) that entirely disabling `mouseup` event on the input was a bad idea: it prevented the user from moving the cursor inside the input when all the text was selected.

This commit changes the way we handle all this:

- if the input is not focused, we disable the `mousedown` event on it. This prevent the `focus` event from firing.
- if the input is not focused and we catch a `mouseup` event on the input,  we select the whole text and focus the input manually.
- if the input is focused, we do not touch the events and let the browser handle them.

Task: [4350439](https://www.odoo.com/odoo/2328/tasks/4350439)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/text_input/text_input.ts
+++ b/src/components/text_input/text_input.ts
@@ -88,15 +88,21 @@ export class TextInput extends Component<Props, SpreadsheetChildEnv> {
     this.inputRef.el?.blur();
   }
 
-  focusInputAndSelectContent() {
-    const inputEl = this.inputRef.el;
-    if (!inputEl) return;
+  onMouseDown(ev: MouseEvent) {
+    // Stop the event if the input is not focused, we handle everything in onMouseUp
+    if (ev.target !== document.activeElement) {
+      ev.preventDefault();
+      ev.stopPropagation();
+    }
+  }
 
-    // The onFocus event selects all text in the input.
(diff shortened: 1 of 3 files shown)


hokolomopo:
> Have you considered simply relying on the focus event and see if ev.target is the same as ev.relatedTarget https://developer.mozilla.org/en-US/docs/Web/API/Element/focus_event#event_properties I think you achieve the same functional result but don't need the convoluted 'stop an event to prevent the following one'

@rrahir reading the spec, I'm not sure why I would compare ev.target and ev.related target ? relatedTarget is the element losing focus, so it should always be different from the element gaining focus (ev.target) right ? 

Anyway the reason we do not select the text on focus is that the mouseUp (that is fired after the focus) can sometimes unselect the text. That's why dhrp stopped/prevented the mouseUp event in his original PR :x

LucasLefevre (commented):
robodoo r+

PR #5344, opened 2024-12-13
[FIX] custom currency: panel on very long currency code

hokolomopo:
## Description

The custom currency panel was not displaying correctly when the
 currency code was very long. This commit fixes the CSS.

Task: [4402647](https://www.odoo.com/odoo/2328/tasks/4402647)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/custom_currency/custom_currency.xml
+++ b/src/components/side_panel/custom_currency/custom_currency.xml
@@ -59,11 +59,11 @@
           onChange.bind="toggleAccountingFormat"
         />
         <div class="o-format-examples mt-4" t-if="selectedFormat">
-          <table>
+          <table class="w-100">
             <t t-foreach="getFormatExamples()" t-as="example" t-key="example_index">
               <tr>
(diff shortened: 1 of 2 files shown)


src/components/side_panel/custom_currency/custom_currency.xml
```diff
@@ -59,11 +59,11 @@
           onChange.bind="toggleAccountingFormat"
         />
         <div class="o-format-examples mt-4" t-if="selectedFormat">
-          <table>
+          <table class="w-100">
```
rrahir, line 62: I think that it can sync nicely with https://developer.mozilla.org/en-US/docs/Web/CSS/table-layout#fixed without having the weird max-width = 1px

hokolomopo: good catch :+1:

pro-odoo (approved):
robodoo r+

PR #5345, opened 2024-12-13
[FIX] bottom_bar: ensure active sheet is scrolled into view

dhrp-odoo:
## Description:

Previously, selecting a sheet from the menu did not scroll the bottom bar's sheet list to bring the active sheet into view if it was outside the visible area. This commit resolves the issue by automatically scrolling the sheet list to ensure the active sheet is visible.

Task: [4377840](https://www.odoo.com/odoo/2328/tasks/4377840)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/tests/bottom_bar/bottom_bar_component.test.ts
+++ b/tests/bottom_bar/bottom_bar_component.test.ts
@@ -604,6 +604,23 @@ describe("BottomBar component", () => {
       simulateClick(".o-bottom-bar-arrow-left");
       expect(scrollTo).toBe(200);
     });
+
+    test("Selecting a sheet from the context menu scrolls to that sheet", async () => {
+      const mockScrollIntoView = jest.fn();
+      HTMLElement.prototype.scrollIntoView = mockScrollIntoView;
+
+      expect(model.getters.getActiveSheetId()).toBe("Sheet1");
+
+      await click(fixture, ".o-list-sheets");
+      await click(fixture, ".o-menu-item[data-name='Sheet6']");
+
+      expect(model.getters.getActiveSheetId()).toBe("Sheet6");
+
+      const sheet6Element = fixture.querySelector(".o-sheet[data-id='Sheet6']");
+      expect(mockScrollIntoView).toHaveBeenCalledWith({ behavior: "smooth", inline: "nearest" });
+      expect(mockScrollIntoView).toHaveBeenCalledTimes(1);
+      expect(mockScrollIntoView.mock.instances[0]).toBe(sheet6Element);
+    });
   });
 
   test("Display the statistic button only if no-empty cells are selected", async () => {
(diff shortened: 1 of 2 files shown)


src/components/bottom_bar/bottom_bar.ts
```diff
@@ -152,6 +152,7 @@ export class BottomBar extends Component<Props, SpreadsheetChildEnv> {
         textColor: sheet.isVisible ? undefined : "grey",
         execute: (env) => {
           env.model.dispatch("ACTIVATE_SHEET", { sheetIdFrom: from, sheetIdTo: sheetId });
+          this.scrollToSheet(sheetId);
```
hokolomopo, line 155: Can't you use [scrollIntoView](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) to scroll instead of having to manually compute the scroll ? Or does it not work ?

dhrp-odoo: Ah, you're right! I wasn't aware that such a straightforward solution exists. This is much simpler and exactly what we need here. However, we will need to mock it in order to test the behavior.

src/components/bottom_bar/bottom_bar.ts
```diff
@@ -152,6 +152,17 @@ export class BottomBar extends Component<Props, SpreadsheetChildEnv> {
         textColor: sheet.isVisible ? undefined : "grey",
         execute: (env) => {
           env.model.dispatch("ACTIVATE_SHEET", { sheetIdFrom: from, sheetIdTo: sheetId });
+
+          // Scrolls the sheet list to ensure the specified sheet is visible.
+          const sheetEl = this.bottomBarRef.el!.querySelector<HTMLElement>(
```
rrahir, line 157: you can move this code in a useEffect of BottomBarSheet, like so
```typescript
useEffect(
      (sheetId) => {
        if (this.props.sheetId === sheetId) {
          this.scrollToSheet();
        }
      },
      () => [this.env.model.getters.getActiveSheetId()]
```
I think it's better to let the component scroll itself instead of a querySelector which assumes the DOM structure of its child component.
Of course the arguments you added to the scrollIntoView could be updated the same way in `scrollToSheet`
    );
```

File: tests/bottom_bar/bottom_bar_component.test.ts [hunk 3]
```diff
@@ -604,6 +606,23 @@ describe("BottomBar component", () => {
...
+
+      await click(fixture, ".o-list-sheets");
+      await click(fixture, ".o-menu-item[data-name='Sheet6']");
+
+      expect(model.getters.getActiveSheetId()).toBe("Sheet6");
+
+      const sheet6Element = fixture.querySelector(".o-sheet[data-id='Sheet6']");
+      expect(mockScrollIntoView).toHaveBeenCalledWith({ behavior: "smooth", inline: "nearest" });
+      expect(mockScrollIntoView).toHaveBeenCalledTimes(1);
+      expect(mockScrollIntoView.mock.instances[0]).toBe(sheet6Element);
```
rrahir, line 624: if I understand, instances contains the object on which the spy function was called? Nice finding !

rrahir (changes requested):

tests/bottom_bar/bottom_bar_component.test.ts
dhrp-odoo: Yes, exactly! `instances` holds the DOM elements (in this case) where the spy function was invoked.

Thanks for the suggestion! I've updated the PR 😊

rrahir:
@robodoo r+

PR #5346, opened 2024-12-13
[FIX] spreadsheet: avoid parasitic renders

hokolomopo:
## Description

Clicking on a button on the topbar would trigger 3 renders:
- one on model update (expected)
- one because we received a collaborative event for our own revision (useless)
- one because the onClick of the topbar calls `DOMFocusableElementStore.focus`, which triggers a render because a store method was called (useless)

Task: [4373054](https://www.odoo.com/odoo/2328/tasks/4373054)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/collaborative/session.ts
+++ b/src/collaborative/session.ts
@@ -319,6 +319,9 @@ export class Session extends EventBus<CollaborativeEvent> {
       }
     }
     this.acknowledge(message);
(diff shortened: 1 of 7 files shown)


rrahir:
@robodoo r+

PR #5356, opened 2024-12-16
[IMP] figures: add anchor to figures

fdamhaut:
## Description:

 [IMP] figures: add anchor to figures

Add possibilities for chart to be anchored to the
top-left corner cell. If a cell is anchored, it
will be moving with the cell on modification of the sheet
(i.e. col/row addition/deletion).

The figure can still have a fixed position, this will act
as if the anchor is placed in the left-most, top-most cell.

Task: 4189170

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/types/figure.ts
```diff
@@ -1,14 +1,24 @@
...
+  offset: PixelPosition;
+  fixed_position: boolean;
   width: Pixel;
   height: Pixel;
   tag: string;
 }
 
+export interface FigureUI extends DOMCoordinates {
+  figure: Figure;
+}
```
pro-odoo, line 15: We use a lot `figureUI.figure.id`, this makes the code harder to read IMHO.
Also, I'm not sure about the naming of `FigureUI`
What do you think of:
```ts
export interface FigureWithDomCoordinates extends Figure {
  domCoordinates: DOMCoordinates;
}
```

demo/data.js
```diff
@@ -510,8 +518,14 @@ export const demoData = {
         },
         {
           id: "9",
+<<<<<<< HEAD
```
pro-odoo: A little rebase is needed :)
I made a PR to update demo data in v25 (https://github.com/odoo/o-spreadsheet/pull/5563/files) so you can remove the next commit. Then you can add a commit with an update to v26 (it will make the diff lower).
Thanks !

src/types/commands.ts
```diff
@@ -518,10 +524,8 @@ export interface UpdateChartCommand extends SheetDependentCommand {
 // Image
 //------------------------------------------------------------------------------
 
-export interface CreateImageOverCommand extends SheetDependentCommand {
+export interface CreateImageOverCommand extends SheetDependentCommand, SubFigureCommand {
```
pro-odoo, line 527: This will require an upgrade script in odoo. See here for an exemple:
https://github.com/odoo/upgrade/blob/master/migrations/spreadsheet_edition/saas~17.5.1.0/pre-migrate.py

pro-odoo (commented):
Some comments while scrolling :)

LucasLefevre (commented):
Also a partial review

src/components/side_panel/chart/building_blocks/general_design/general_design_editor.xml
```diff
@@ -20,6 +20,14 @@
           style="titleStyle"
           onFontSizeChanged.bind="updateChartTitleFontSize"
         />
+        <Section title.translate="Position">
```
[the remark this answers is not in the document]
fdamhaut: This was due to discussion with VSC, as both excel and google allow the user to choose between those 2 behaviours

src/types/commands.ts
```diff
@@ -498,42 +499,44 @@ export interface CreateFigureCommand extends SheetDependentCommand {
...
 export interface DeleteFigureCommand extends SheetDependentCommand {
   type: "DELETE_FIGURE";
-  id: UID;
+  figureId: UID;
+}
+
+interface BaseFigureCommand extends SheetDependentCommand {
+  figureId: UID;
+  anchor?: Position;
+  offset?: PixelPosition;
```
hokolomopo, line 515: why optional ?

src/clipboard_handlers/image_clipboard.ts
```diff
@@ -55,33 +55,26 @@ export class ImageClipboardHandler extends AbstractFigureClipboardHandler<Clipbo
...
 
     if (options.isCutOperation) {
       this.dispatch("DELETE_FIGURE", {
         sheetId: clippedContent.sheetId,
-        id: clippedContent.copiedFigure.id,
+        figureId: clippedContent.copiedFigure.id,
       });
     }
-    this.dispatch("SELECT_FIGURE", { id: figureId });
+    this.dispatch("SELECT_FIGURE", { figureId: figureId });
```
hokolomopo, line 77:
```suggestion
    this.dispatch("SELECT_FIGURE", { figureId });
```
there's probably other places in the code with this

src/types/commands.ts
```diff
@@ -498,42 +499,44 @@ export interface CreateFigureCommand extends SheetDependentCommand {
...
 
 export interface DeleteFigureCommand extends SheetDependentCommand {
   type: "DELETE_FIGURE";
-  id: UID;
+  figureId: UID;
+}
+
+interface BaseFigureCommand extends SheetDependentCommand {
+  figureId: UID;
+  anchor?: Position;
```
hokolomopo, line 514: I don't think `anchor` is transformed if someone were to add a row at the same time you add a figure. Would work w/o needing `ot_specific` stuff if you extends `PositionDependentCommand` and remove the anchor.

tests/figures/image/image_plugin.test.ts
```diff
@@ -59,8 +59,8 @@ describe("image plugin", function () {
       size: { width: 100, height: 100 },
       mimetype: "image/jpeg",
     };
-    createImage(model, { figureId: imageId, definition });
-    model.dispatch("SELECT_FIGURE", { id: imageId });
+    createImage(model, { id: imageId, definition });
```
hokolomopo, line 62: why those changes ? It's still a figureId as far as I can see

tests/figures/chart/menu_item_insert_chart.test.ts
```diff
@@ -128,19 +129,28 @@ describe("Insert chart menu item", () => {
...
-    payload.position = {
-      x: (width - DEFAULT_FIGURE_WIDTH) / 2,
-      y: (height - DEFAULT_FIGURE_HEIGHT) / 2,
+    payload.offset = {
+      x: 40,
+      y: 10.5,
+    };
+    payload.anchor = {
+      col: 2,
+      row: 14,
```
hokolomopo, line 138: All the changes to this test file are sus'. What are these hardcoded values and where do they come from ? I get it's supposed to be the center of the screen, but why do we need to hard code them instead of computing it now ? And why is another test `expect({ x: figureUI.x, y: figureUI.y }).toStrictEqual` now needed ? Is the payload test not enough ? I'm a bit confused.

tests/figures/chart/menu_item_insert_chart.test.ts
```diff
@@ -128,19 +129,28 @@ describe("Insert chart menu item", () => {
...
     }; // Position at the center of the viewport
     expect(dispatchSpy).toHaveBeenCalledWith("CREATE_CHART", payload);
+    const figureUI = model.getters.getVisibleFigures()[0];
+    expect({ x: figureUI.x, y: figureUI.y }).toStrictEqual({
```
hokolomopo, line 142:
```suggestion
    expect(model.getters.getVisibleFigures()[0]).toMatchObject({
```

tests/figures/figure_component.test.ts
```diff
@@ -330,106 +382,130 @@ describe("figures", () => {
...
 
     test.each([
-      [{ wheelX: 0, wheelY: 10 * DEFAULT_CELL_HEIGHT }],
-      [{ wheelX: 10 * DEFAULT_CELL_WIDTH, wheelY: 0 }],
-      [{ wheelX: 0, wheelY: 50 * DEFAULT_CELL_HEIGHT }], // scroll out of original viewport
-      [{ wheelX: 40 * DEFAULT_CELL_WIDTH, wheelY: 0 }], // scroll out of original viewport
+      [{ wheelX: 0, wheelY: 10 }],
+      [{ wheelX: 10, wheelY: 0 }],
+      [{ wheelX: 0, wheelY: 50 }], // scroll out of original viewport
+      [{ wheelX: 40, wheelY: 0 }], // scroll out of original viewport
```
hokolomopo, line 457: nitpick: variable names `wheelX` which looks like DOM event parameters being in `HeaderOffset` instead of `Pixel` is a bit confusing. Just add `additionalOffsetX` or keep wheelX and do things like `Math.floor(wheelX /DEFAULT_CELL_HEIGHT)` to compute the anchor/offset if you want to test those

hokolomopo (commented):
Xlsx import/export is totally broken, all the figures are on the top-left of the sheets :x

src/plugins/ui_stateful/sheetview.ts
```diff
@@ -296,6 +300,23 @@ export class SheetViewPlugin extends UIPlugin {
     return Math.max(...this.getSubViewports(sheetId).map((viewport) => viewport.getColIndex(x)));
   }
 
+  /**
+   * Return the index of a column given an offset x, based on the viewport left
+   * visible cell.
+   * If the position is outside of any viewport, return the column index on the main viewport.
+   */
+  getColIndexHidden(x: Pixel): HeaderIndex {
+    const colIndex = this.getColIndex(x);
```
rrahir, line 309: I have a hard time understanding the implementation :/
We should also find a clearer name imho

src/plugins/ui_stateful/sheetview.ts
```diff
@@ -306,6 +327,23 @@ export class SheetViewPlugin extends UIPlugin {
...
+   * If the position is outside of any viewport, return the row index on the main viewport.
+   */
+  getRowIndexHidden(y: Pixel): HeaderIndex {
+    const rowIndex = this.getRowIndex(y);
+    if (rowIndex >= 0) return rowIndex;
+    const sheetId = this.getters.getActiveSheetId();
+    let i = 0;
+    y += this.getActiveSheetScrollInfo().scrollY;
+    for (; y > 0; i += 1) {
+      y -= this.getters.getUserRowSize(sheetId, i) ?? DEFAULT_CELL_HEIGHT;
```
rrahir, line 342:
```suggestion
      y -= this.getters.getRowSize(sheetId, i);
```

src/plugins/ui_stateful/sheetview.ts
```diff
@@ -814,35 +852,74 @@ export class SheetViewPlugin extends UIPlugin {
...
       ) {
         continue;
       }
-      result.push(figure);
+      result.push(figureUI);
     }
     return result;
   }
 
+  getFigureUI(sheetId: string, figure: Figure): FigureUI {
```
rrahir, line 883:
```suggestion
  getFigureUI(sheetId: UID, figure: Figure): FigureUI {
```

src/plugins/ui_stateful/sheetview.ts
```diff
@@ -814,35 +852,74 @@ export class SheetViewPlugin extends UIPlugin {
...
+  getVisibleFigures(): FigureUI[] {
     const sheetId = this.getters.getActiveSheetId();
-    const result: Figure[] = [];
+    const result: FigureUI[] = [];
     const figures = this.getters.getFigures(sheetId);
-    const { scrollX, scrollY } = this.getActiveSheetScrollInfo();
-    const { x: offsetCorrectionX, y: offsetCorrectionY } =
-      this.getters.getMainViewportCoordinates();
     const { width, height } = this.getters.getSheetViewDimensionWithHeaders();
+    const { x: offsetCorrectionX, y: offsetCorrectionY } = this.getMainViewportCoordinates();
```
rrahir, line 860: useless diff

src/plugins/ui_stateful/sheetview.ts
```diff
@@ -814,35 +852,74 @@ export class SheetViewPlugin extends UIPlugin {
...
     return result;
   }
 
+  getFigureUI(sheetId: string, figure: Figure): FigureUI {
+    const x = figure.offset.x + this.getters.getColDimensions(sheetId, figure.anchor.col)["start"];
+    const y = figure.offset.y + this.getters.getRowDimensions(sheetId, figure.anchor.row)["start"];
+    return { ...figure, x, y };
+  }
+
+  getPositionAnchorOffset(sheetId: string, position: PixelPosition): AnchorOffset {
```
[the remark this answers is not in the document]
rrahir: I would suggest the follwing implementation that allows to avoid a certain number of new helpers:
I haven't tested it, but its to give the general idea. I would bet that it misses a scroll correction somewhere ^^ 
 ```typescript
const col = this.getColIndex(position.x) // find the cell we are part of
const row = this.getRowIndex(position.y)
const {x, y} = this.getRect(positionToZone({col, row})) // get the topLeft point of the cell (even if it's partially hidden by the scroll)

return {
      anchor: {col ,row},
      offset: {
        x: position.x - x,
        y: position.y - y,
      },
    };

rrahir (commented):
Hello 👋  

Mainly details and maybe one or two points that could be reworked for readability/ simplicity.

I haven't read the tests yet.

src/components/figures/figure_container/figure_container.ts
```diff
@@ -247,43 +245,64 @@ export class FiguresContainer extends Component<Props, SpreadsheetChildEnv> {
...
 
     const sheetId = this.env.model.getters.getActiveSheetId();
 
     const initialMousePosition = { x: ev.clientX, y: ev.clientY };
+    const initialScrollPosition = this.env.model.getters.getActiveSheetScrollInfo();
+
+    // switch to global viewport position
+    const container = this.getFigureContainer(figureUI);
+    if (["bottomLeft", "topLeft"].includes(container)) {
+      figureUI.x += initialScrollPosition.scrollX;
```
rrahir, line 301: I think the previous approach that would create a new object was better because here , you change the values in place but the object is passed as a props, so you actually change a props in place.

Was there a specific reason to get rid of `internalFigureToScreen` and its friends? I found the intention clearer with it (subjective af) but I might miss something related to the new anchor

rrahir (commented):
👋 New wave of comments, apparently  I missed several things at my first pas, i'll need to get back to it again and have a closer look at the drag and drop figure changes ;-)

rrahir:
@fdamhaut I forgot a point, we should have some tests that cover the functional issues I mentionned in the task pad (deleting the anchor header and the "MOVE_COLUMNS_ROWS" issue

fdamhaut:
> @fdamhaut I forgot a point, we should have some tests that cover the functional issues I mentionned in the task pad (deleting the anchor header and the "MOVE_COLUMNS_ROWS" issue

I already added some test in figures_plugin.test for that L457-830

src/types/commands.ts
```diff
@@ -493,47 +493,48 @@ export interface MoveConditionalFormatCommand extends SheetDependentCommand {
...
 }
 
 export interface DeleteFigureCommand extends SheetDependentCommand {
   type: "DELETE_FIGURE";
-  id: UID;
+  figureId: UID;
+}
+
+interface BaseFigureCommand extends PositionDependentCommand {
+  figureId: UID;
```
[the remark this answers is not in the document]
fdamhaut: More than alerting the dev, the goal is harmonizing command to have a similar structure. And from having either "id" or "figureId", i think the more explicit option is the best. (+ It's more like the rest of the command, sheet always use "sheetId", Pivot almost always use "pivotId",...)

src/components/figures/figure_container/figure_container.ts
fdamhaut: > Was there a specific reason to get rid of internalFigureToScreen and its friends? I found the intention clearer with it (subjective af) but I might miss something related to the new anchor
Here the code only move the figure coordinate if it's part of the frozen pane viewport so that is doesn't teleport to their "true" position, it's not really related to "internal -> screen" old behaviour.
However, i will isolate the resulting code in a method to be more explicit :+1:

rrahir:
> figures_plugin.test

SOrry I completely missed it :'(

src/plugins/ui_stateful/selection.ts
```diff
@@ -602,6 +604,43 @@ export class GridSelectionPlugin extends UIPlugin {
...
+          row: figure.row,
+        });
+      }
+      if (cmd.dimension == "ROW" && figure.row in moveDict) {
+        cmds.push({
+          type: "UPDATE_FIGURE",
+          sheetId: cmd.sheetId,
+          figureId: figure.id,
+          row: moveDict[figure.row],
+          col: figure.row,
```
rrahir, line 633:
```suggestion
          col: figure.col,
```
maybe a small test ^^

src/plugins/ui_stateful/sheetview.ts
```diff
@@ -817,35 +831,91 @@ export class SheetViewPlugin extends UIPlugin {
...
+   * It returns -1 if no row is found.
+   */
+  private getRowIndexTopOfMainViewport(y: Pixel): HeaderIndex {
+    if (y >= 0) return -1;
+    const sheetId = this.getters.getActiveSheetId();
+    let row = this.getActiveMainViewport().top;
+    let rowStart =
+      -this.getActiveSheetScrollInfo().scrollY - this.getters.getColRowOffset("ROW", row, 0);
+    for (; y < rowStart; row--) {
+      rowStart -= this.getters.getColSize(sheetId, row - 1);
```
rrahir, line 863:
```suggestion
      rowStart -= this.getters.getRowSize(sheetId, row - 1);
```
or probably getUserRowSize ?

rrahir (changes requested):
👋  Some small changes and a rebase and I think we're good to go :)

rrahir:
@robodoo r+

PR #5375, opened 2024-12-19
[FIX] popover: wrong position on updated popover

hokolomopo:
## Description

If a popover has a change of props, we don't clean its style before computing its new height/width, leading to a wrong position sometimes.

Task: [3814260](https://www.odoo.com/odoo/2328/tasks/3814260)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/popover/popover.ts
+++ b/src/components/popover/popover.ts
@@ -81,6 +81,10 @@ export class Popover extends Component<PopoverProps, SpreadsheetChildEnv> {
       this.currentDisplayValue = newDisplay;
 
       if (!anchor) return;
+      el.style.top = "";
+      el.style.left = "";
+      el.style["max-height"] = "";
+      el.style["max-width"] = "";
 
       const propsMaxSize = { width: this.props.maxWidth, height: this.props.maxHeight };
       let elDims = {


src/components/popover/popover.ts
```diff
@@ -81,6 +81,10 @@ export class Popover extends Component<PopoverProps, SpreadsheetChildEnv> {
       this.currentDisplayValue = newDisplay;
 
       if (!anchor) return;
+      el.style.top = "";
+      el.style.left = "";
+      el.style["max-height"] = "";
+      el.style["max-width"] = "";
```
LucasLefevre, line 87: you're resetting this here to have the correct `getBoundingClientRect` just below ? Is that right ?

LucasLefevre:
Actually, I can't reproduce the issue neither in 17.0 nor master :/ 
I might not do the right thing

src/components/popover/popover.ts
hokolomopo: yes

LucasLefevre (approved):
robodoo r+

PR #5400, opened 2025-01-02
[FIX] renderer: Fix Box rendering

rrahir:
When we implemented the frozen panes, we accidently destroyed the
distinction between the actual size of a cell/merge and its visual
dimensio in the viewport. i.e. the dimension of the part that we be
rendered.

However, the rest of the rendering process was relying on the box size
to compute several things (button , text positions for instance)
This means that all computations taht relied on the actual size of a
cell/merge are now false.

A good example is the merge. Create a merge with some text inside of it
and make sure it spans over several columns.
The text is aligned at the center of the merge. Now scroll a bit and the
text will stay centered in the VISIBLE part of the merge but not the
merge itself.

On another hand, we did not account for some text formatting, specifically
text aligned on the right can overflow outside of their pane.

This revision fixes the two issues:
- Reintroduce a getter `getRect` to properly compute the 
  positiontioning of text/icons of a box
- Change the rendering process of the grid by splitting the rendering
  one pane at a time and work in reverse such that a frozen pane takes
  precedence on the main viewport.

The second part of the approach is not the panacea as the mobile mode 
will push the concept of frozen pane further but as a fix it suits well.

Task: [4448426](https://www.odoo.com/odoo/2328/tasks/4448426)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/helpers/internal_viewport.ts
```diff
@@ -222,7 +222,7 @@ export class InternalViewport {
    * @param zone
```
hokolomopo, line 222: Typos in commit msg "dimensio", "taht"

tests/plugins/__snapshots__/renderer.test.ts.snap
```diff
@@ -115,13 +115,18 @@ Array [
   "context.moveTo(0, 989)",
   "context.lineTo(952, 989)",
   "context.stroke()",
+  "context.save()",
+  "context.beginPath()",
+  "context.rect(0, 0, 952, 974)",
+  "context.clip()",
```
hokolomopo, line 121: I would probably do a test that we clip what we dran on the frozen panes. It shouldn't be _too_ annoying

hokolomopo (commented):
:airplane:

tests/plugins/__snapshots__/renderer.test.ts.snap
rrahir: I added some tests related to the renderer and he new getter 'getRect'. Not sure about the naming but 'getRenderingRect' was definitely wrong. it's more used as a computational tool.

src/plugins/ui/sheetview.ts
```diff
@@ -457,7 +459,28 @@ export class SheetViewPlugin extends UIPlugin {
...
+    const x = Math.min(...viewportRects.map((rect) => rect.x));
+    const y = Math.min(...viewportRects.map((rect) => rect.y));
+    const width = Math.max(...viewportRects.map((rect) => rect.x + rect.width)) - x;
+    const height = Math.max(...viewportRects.map((rect) => rect.y + rect.height)) - y;
+    return {
+      x: x + this.gridOffsetX,
+      y: y + this.gridOffsetY,
+      width,
+      height,
+    };
```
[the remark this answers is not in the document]
rrahir: good call, I added it :)

LucasLefevre (commented):
robodoo r+

PR #5403, opened 2025-01-03
[IMP] xlsx: support exporting array formulas

Rachico:
This commit improves the export of vectorized & array formulas, including those that spill across cells, ensuring better compatibility with other spreadsheet applications. The adopted approach is to treat all formulas as array formulas & to add an additional xml file that contains metadata about array formula properties. As a result, we improve the integrity & usability of our generated XLSX files.

Task: [4134347](https://www.odoo.com/odoo/2328/tasks/4134347)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_core_views/cell_evaluation/evaluation_plugin.ts
+++ b/src/plugins/ui_core_views/cell_evaluation/evaluation_plugin.ts
@@ -345,8 +345,10 @@ export class EvaluationPlugin extends CoreViewPlugin {
 
       const formulaCell = this.getCorrespondingFormulaCell(position);
       if (formulaCell) {
+        const cell = this.getters.getCell(position);
+        isFormula = isExported && cell?.content === formulaCell.content;
         isExported = isExportableToExcel(formulaCell.compiledFormula.tokens);
-        isFormula = isExported;
+        // isFormula = isExported;
 
         // If the cell contains a non-exported formula and that is evaluates to
         // nothing* ,we don't export it.
@@ -374,7 +376,7 @@ export class EvaluationPlugin extends CoreViewPlugin {
       const spillZone = this.getSpreadZone(position);
       if (spillZone) {
         exportedSheetData.formulaSpillRanges[xc] = this.getters.getRangeString(
-          this.getters.getRangeFromZone(this.getters.getActiveSheetId(), spillZone),
+          this.getters.getRangeFromZone(position.sheetId, spillZone),
           position.sheetId
         );
       }
--- a/src/plugins/ui_core_views/cell_evaluation/evaluation_plugin.ts
+++ b/src/plugins/ui_core_views/cell_evaluation/evaluation_plugin.ts
@@ -331,10 +331,10 @@ export class EvaluationPlugin extends CoreViewPlugin {
   exportForExcel(data: ExcelWorkbookData) {
     for (const sheet of data.sheets) {
       sheet.cellValues = {};
+      sheet.formulaSpillRanges = {};
     }
     for (const position of this.evaluator.getEvaluatedPositions()) {
       const evaluatedCell = this.evaluator.getEvaluatedCell(position);
-
       const xc = toXC(position.col, position.row);
       const value = evaluatedCell.value;
       let isFormula = false;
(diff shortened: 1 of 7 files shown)


src/plugins/ui_core_views/cell_evaluation/evaluation_plugin.ts
```diff
@@ -372,6 +371,23 @@ export class EvaluationPlugin extends CoreViewPlugin {
...
       exportedSheetData.cells[xc] = content;
       exportedSheetData.cellValues[xc] = value;
+      const spillZone = this.getSpreadZone(position);
+      if (spillZone) {
+        exportedSheetData.arrayFormulaCells[xc] = this.getters.getRangeString(
+          this.getters.getRangeFromZone(this.getters.getActiveSheetId(), spillZone),
+          this.getters.getActiveSheetId()
+        );
+      } else {
+        exportedSheetData.arrayFormulaCells[xc] = this.getters.getRangeString(
```
LucasLefevre, line 381: This deserves a comment because as I read this, it's not an array formula but you still add it to `arrayFormulaCells` which looks louche

src/xlsx/functions/cells.ts
```diff
@@ -33,11 +34,10 @@ export function addFormula(
     return { attrs: [], node: escapeXml`` };
   }
 
-  const attrs: XMLAttributes = [["t", type]];
+  const attrs: XMLAttributes = [["cm", "1"]];
   const XlsxFormula = adaptFormulaToExcel(formula);
-
   const exportedValue = adaptFormulaValueToExcel(value);
-  const node = escapeXml/*xml*/ `<f>${XlsxFormula}</f><v>${exportedValue}</v>`;
+  const node = escapeXml/*xml*/ `<f t="array" ref="${arrayFormulaRange}">${XlsxFormula}</f><v>${exportedValue}</v>`;
```
LucasLefevre, line 40: also add a comment saying we are flagging all formulas as array because it works and it's easier (rather than detecting sub-expressions spilling)

src/xlsx/functions/worksheet.ts
```diff
@@ -110,7 +110,7 @@ export function addRows(
         let cellNode = escapeXml``;
         // Either formula or static value inside the cell
         if (content?.startsWith("=") && value !== undefined) {
-          const res = addFormula(content, value);
+          const res = addFormula(content, value, sheet.arrayFormulaCells[xc]);
```
LucasLefevre, line 113: Actually, I'd remove the `else` in the evaluation plugin and change this like this. I think it would also work
```suggestion
          const res = addFormula(content, value, sheet.arrayFormulaCells[xc] ?? xc);
```

src/plugins/ui_core_views/cell_evaluation/evaluation_plugin.ts
```diff
@@ -372,6 +371,23 @@ export class EvaluationPlugin extends CoreViewPlugin {
       }
       exportedSheetData.cells[xc] = content;
       exportedSheetData.cellValues[xc] = value;
+      const spillZone = this.getSpreadZone(position);
+      if (spillZone) {
+        exportedSheetData.arrayFormulaCells[xc] = this.getters.getRangeString(
+          this.getters.getRangeFromZone(this.getters.getActiveSheetId(), spillZone),
+          this.getters.getActiveSheetId()
```
LucasLefevre, line 378: The position might not be on the active sheet
```suggestion
          position.sheetId
```

LucasLefevre (commented):
could you back-port to 18.0 (or even to 17.0) ?
Thanks :)

Rachico:
> could you back-port to 18.0 (or even to 17.0) ? Thanks :)

Alrighty ;)

src/plugins/ui_core_views/cell_evaluation/evaluation_plugin.ts
Rachico: @LucasLefevre Let me know what you think about this variable name :)

```typescript
formulaSpillRanges
```

Rachico: I think it should be self-explanatory and therefore we don't need any comments here anymore :)

rrahir:
done in https://github.com/odoo/o-spreadsheet/pull/5555 ...

PR #5407, opened 2025-01-06
Formulas: add the SWITCH function

laa-odoo:
Task: [3829128](https://www.odoo.com/odoo/2328/tasks/3829128)

### DIFF
--- a/src/functions/module_logical.ts
+++ b/src/functions/module_logical.ts
@@ -3,7 +3,13 @@ import { AddFunctionDescription, Arg, FunctionResultObject, Maybe } from "../typ
 import { CellErrorType, EvaluationError } from "../types/errors";
 import { arg } from "./arguments";
 import { boolAnd, boolOr } from "./helper_logical";
-import { assert, conditionalVisitBoolean, isEvaluationError, toBoolean } from "./helpers";
+import {
+  assert,
+  conditionalVisitBoolean,
+  isEvaluationError,
+  toBoolean,
+  valueNotAvailable,
+} from "./helpers";
 
 // -----------------------------------------------------------------------------
 // AND
@@ -217,6 +223,53 @@ export const OR = {
     const { result, foundBoolean } = boolOr(logicalExpressions);
     assert(() => foundBoolean, _t("[[FUNCTION_NAME]] has no valid input data."));
     return result;
+  },
+  isExported: true,
+} satisfies AddFunctionDescription;
+
+// -----------------------------------------------------------------------------
+// SWITCH
+// -----------------------------------------------------------------------------
+
+export const SWITCH = {
+  description: _t("Returns a value by comparing cases to an expression."),
+  args: [
+    arg("expression (number, boolean, string)", _t("The value to be checked.")),
+    arg("case1 (number, boolean, string)", _t("The first case to be checked against expression.")),
+    arg("value1 (any)", _t("The corresponding value to be returned if case1 matches expression.")),
+    arg(
+      "case2 (any, repeating)",
+      _t("Additional cases to try if the previous ones don't match expression.")
+    ),
+    arg(
+      "value2 (any, repeating)",
+      _t("Additional values to be returned if their corresponding cases match expression.")
+    ),
+    arg(
+      `default (any, default="empty")`,
+      _t("An optional default value to be returned if none of the cases match expression.")
+    ),
+  ],
+  compute: function (
+    expression: Maybe<FunctionResultObject>,
+    ...casesAndValues: Maybe<FunctionResultObject>[]
+  ): FunctionResultObject {
+    const defaultValue =
+      casesAndValues.length % 2 === 0 ? valueNotAvailable(expression) : casesAndValues.pop();
+
+    for (let i = 0; i < casesAndValues.length; i += 2) {
+      const iCase = casesAndValues[i];
+
+      if (iCase && isEvaluationError(iCase.value)) {
+        return iCase;
+      }
+
+      if (expression?.value === iCase?.value) {
+        return casesAndValues[i + 1] || { value: 0 };
+      }
+    }
+
+    return defaultValue || { value: 0 };
   },
   isExported: true,
 } satisfies AddFunctionDescription;
--- a/src/functions/helpers.ts
+++ b/src/functions/helpers.ts
@@ -40,6 +40,13 @@ export function isEvaluationError(error: Maybe<CellValue>): error is ErrorValue 
   return typeof error === "string" && errorTypes.has(error);
 }
 
+export function valueNotAvailable(searchKey: Maybe<FunctionResultObject>): FunctionResultObject {
+  return {
+    value: CellErrorType.NotAvailable,
+    message: _t("Did not find value '%s' in [[FUNCTION_NAME]] evaluation.", toString(searchKey)),
+  };
+}
+
 // -----------------------------------------------------------------------------
 // FORMAT FUNCTIONS
 // -----------------------------------------------------------------------------
--- a/src/functions/module_lookup.ts
+++ b/src/functions/module_lookup.ts
@@ -25,19 +25,13 @@ import {
   toMatrix,
   toNumber,
   toString,
+  valueNotAvailable,
 } from "./helpers";
 
 const DEFAULT_IS_SORTED = true;
 const DEFAULT_MATCH_MODE = 0;
 const DEFAULT_SEARCH_MODE = 1;
 const DEFAULT_ABSOLUTE_RELATIVE_MODE = 1;
-
-function valueNotAvailable(searchKey: Maybe<FunctionResultObject>): FunctionResultObject {
-  return {
-    value: CellErrorType.NotAvailable,
-    message: _t("Did not find value '%s' in [[FUNCTION_NAME]] evaluation.", toString(searchKey)),
-  };
-}
 
 // -----------------------------------------------------------------------------
 // ADDRESS
--- a/tests/functions/module_logical.test.ts
+++ b/tests/functions/module_logical.test.ts
@@ -408,6 +408,65 @@ describe("OR formula", () => {
   });
 });
 
+describe("SWITCH formula", () => {
+  test("take 3 argument minimum", () => {
+    expect(evaluateCell("A1", { A1: "=SWITCH(42, 42, 42)" })).toBe(42);
+    expect(evaluateCell("A1", { A1: "=SWITCH(42, 42)" })).toBe("#BAD_EXPR");
+    expect(evaluateCell("A1", { A1: "=SWITCH(42)" })).toBe("#BAD_EXPR");
+    expect(evaluateCell("A1", { A1: "=SWITCH()" })).toBe("#BAD_EXPR");
+  });
+
+  test("functional tests on simple arguments", () => {
+    expect(evaluateCell("A1", { A1: "=SWITCH( 42, 23, 123)" })).toBe("#N/A");
+    expect(evaluateCell("A1", { A1: "=SWITCH( 23, 23, 123)" })).toBe(123);
+    expect(evaluateCell("A1", { A1: "=SWITCH( 42, 23, 123, 42, 321)" })).toBe(321);
+    expect(evaluateCell("A1", { A1: "=SWITCH( 42, 23, 123, 24, 321)" })).toBe("#N/A");
+    expect(evaluateCell("A1", { A1: "=SWITCH( 42, 23, 123, 24, 321, 42, 1111)" })).toBe(1111);
+
+    expect(evaluateCell("A1", { A1: "=SWITCH(TRUE, FALSE, 23)" })).toBe("#N/A");
+    expect(evaluateCell("A1", { A1: "=SWITCH(TRUE, TRUE, 23)" })).toBe(23);
+    expect(evaluateCell("A1", { A1: "=SWITCH(23, 23, TRUE)" })).toBe(true);
+
+    expect(evaluateCell("A1", { A1: '=SWITCH("str", 23, 1, "str2", 2, "str", 3)' })).toBe(3);
+    expect(evaluateCell("A1", { A1: '=SWITCH(25, 26, "str2", 25, "str")' })).toBe("str");
+
+    expect(evaluateCell("A1", { A1: "=SWITCH( , , 42)" })).toBe(42);
+    expect(evaluateCell("A1", { A1: "=SWITCH( 42, 42, )" })).toBe(0);
+
+    expect(evaluateCell("A1", { A1: '=SWITCH("", "", 42)' })).toBe(42);
+    expect(evaluateCell("A1", { A1: '=SWITCH(42, 42, "")' })).toBe("");
+
+    expect(evaluateCell("A1", { A1: '=SWITCH("true" , TRUE, 42)' })).toBe("#N/A");
+    expect(evaluateCell("A1", { A1: "=SWITCH(1, TRUE, 42)" })).toBe("#N/A");
+  });
+
+  test("functional with optional value after repeatable values", () => {
+    expect(evaluateCell("A1", { A1: "=SWITCH( 42, 23, 23, 123)" })).toBe(123);
+    expect(evaluateCell("A1", { A1: "=SWITCH( 42, 23, 23, 42, 321, 123)" })).toBe(321);
+    expect(evaluateCell("A1", { A1: "=SWITCH( 42, 23, 23, 45, 321, )" })).toBe(0);
+    expect(evaluateCell("A1", { A1: "=SWITCH( 42, 43, 43, )" })).toBe(0);
+  });
+
+  test("functional tests on cell arguments", () => {
+    expect(
+      evaluateCell("A1", { A1: "=SWITCH(A2, A3, A4)", A2: "test", A3: "TRUE", A4: "42" })
+    ).toBe("#N/A");
+    expect(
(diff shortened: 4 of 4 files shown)


anhe-odoo (changes requested):

src/functions/arguments.ts
```diff
@@ -132,32 +136,80 @@ export function addMetaInfoFromArg(addDescr: AddFunctionDescription): FunctionDe
...
-  if (!repeatingArg) {
-    return (argPosition) => argPosition;
-  }
-  if (repeatingArg === 1) {
-    return (argPosition) => Math.min(argPosition, countArg);
+export function argTargeting(
+  functionDescription: FunctionDescription,
+  nbrArgSupplied: number
+): (argPosition: number) => number | undefined {
+  if (nbrArgSupplied === functionDescription.minArgRequired) {
```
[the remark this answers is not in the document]
laa-odoo: Indeed :+1:

src/functions/module_filter.ts
```diff
@@ -221,12 +221,20 @@ export const SORTN: AddFunctionDescription = {
   compute: function (
```
anhe-odoo, line 221: should remove the isExported (or set to False)

src/functions/arguments.ts
```diff
@@ -96,67 +97,163 @@ export function addMetaInfoFromArg(addDescr: AddFunctionDescription): FunctionDe
...
+ * - they must be in a quantity greater than the optional arguments.
+ *
+ * The markdown tables below illustrate how values are mapped to positions based on the number of values supplied.
+ * Each table represents a different function configuration, with columns representing the number of values supplied as arguments
+ * and rows representing the correspondence with the argument index.
+ *
+ * The tables are built based on the following conventions:
+ * - `m`: Mandatory argument
+ * - `o`: Optional argument
+ * - `r`: Repeating argument
```
hokolomopo, line 142: I didn't read the details of the code, but are we not taking a wrong path here ? I'm not really sure why we are trying to handle exotic arguments combinatations (do they even exist in practice) at the price of an argument combination that _we know exist and was working fine_ (SORTN).

Reading only the commit msg and this function docstring my main takeway are:

1) SORTN arguments (`FUNC(m, o, o ,o, r, r)`) sounds entierly reasonable. I don't think we want a solution that does not support that. 
2) My very thrustworthy source (chatgpt) told me that mandatory arguments are always placed before optional arguments. Idk if he's right or not. And even if they are functions that do that, isn't the correct syntax to  have empty arguments in the arguments list ? eg. your configuration 4 we would not write `=FUNC(m1, m2, m3)` but `=FUNC(m1, , m2, , m3)` that way there is no confusion.

We should talk about this IRL, there's most likely tons of subtleties that I'm missing

laa-odoo: To be honest, I'm not entirely sure about the proposed solution here either.

It also bothers me for the point you raised, but I’m taking this opportunity to put in writing what has been discussed regarding this task.

To try and answer your question, the goal is not to specifically handle exotic cases, but the proposed solution allows us to manage all corner cases without having to constantly question how far we should go (or stop) in processing the arguments—at least, IMO, I don’t really know.

How did this reflection arise? It started from the need to implement the SWITCH function (like a switch case). It turns out that this function allows adding an argument at the end in addition to the repeatable arguments. This final argument corresponds to the default return value. Thus, the function follows an exotic pattern: [m, m, m, r, r, o]—exotic because an optional argument appears after repeatable arguments.

I call it exotic, but in reality, Google and Excel implemented SWITCH with the following pattern: [m, m, m, r, r]. It’s only through the user interface that the first repeatable argument is displayed as both repeatable and optional. From a user perspective, I find this quite confusing, as the UI keeps prompting the user to enter another repeatable parameter even after the optional argument.

Additionally, following the same approach as Excel and Google would have required finding another way to verify that the number of arguments passed during compilation is correct. With Google and Excel’s approach, it is no longer sufficient to simply apply a modulo operation on the number of repeatable/optional arguments to check whether the provided argument count is valid.

So, we asked ourselves if there was a solution that could handle cases like [m, m, m, r, r, o] to avoid compilation issues while also ensuring a proper representation of argument insertion in the UI.

This reflection also took into account that other formulas, such as LET and LAMBDA (which we don't yet support), have their own unique patterns: [m, m, r, r, m] or [r, m].

Considering all these cases - [m, m, m, r, r, o], [m, m, r, r, m], [r, m], and [m, o, o, o, r, r] (for SORTN) - there is probably a way to handle them properly. However, we don’t know if other exotic patterns should also be considered. For example, given the nature of the cases mentioned, could a pattern like [r, r, r, m, o] exist? It doesn’t seem far-fetched considering the other cases.

So, should we write specific code that we modify/restrict for each particular case we encounter? I don’t know.

In any case, the proposed solution aims to be universal and allows us to avoid worrying about a specific order of arguments. However, there are still some rules to follow, which conflict with SORTN.

Regarding SORTN, it turns out that this function was introduced by Google, not Excel. Since Google isn’t as "serious" as Excel, I concluded that they didn’t really know what they were doing and mistakenly defined their arguments, failing to make the second argument mandatory. This actually works in my favor, as it aligns with the rules of the new algorithm.

Now, I’m well aware that this solution could create confusion, such as in configuration 4, which you mentioned. However, configuration 4 has no functional relevance here—it is only meant to illustrate what the algorithm implies.

It’s worth noting that implementing such configurations will remain the responsibility of the formula developer rather than the end user. There is still a layer of abstraction before users encounter such cases and get confused about how argument interpretation actually works.

laa-odoo: The more general question that arises behind this problem is should we treat the cases `m, r, m` and `m, o, m` in the same way when we only provide two values

anhe-odoo: How would you like to treat them differently ?

laa-odoo: The solution proposed in this PR below doesn't treat them differently: 
when we provide only two arguments, they are treated in the same way,
--> the second value corresponds to the 3rd argument.

If we return to a classic implementation, they will be treated differently.
Because this does not make sense, we should then prohibit the mandatory arguments positioned after the optional arguments.

anhe-odoo: What was the final decision about this discussion ?

src/functions/module_filter.ts
anhe-odoo: Up

src/formulas/compiler.ts
```diff
@@ -316,43 +318,48 @@ function formulaArguments(tokens: Token[]) {
  * Check if arguments are supplied in the correct quantities
  */
 function assertEnoughArgs(ast: ASTFuncall) {
-  const nbrArg = ast.args.length;
+  const nbrArgSupplied = ast.args.length;
   const functionName = ast.value.toUpperCase();
   const functionDefinition = functions[functionName];
+  const nbrArgRepeating = functionDefinition.nbrArgRepeating;
```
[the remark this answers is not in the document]
laa-odoo: thx

src/functions/module_filter.ts
laa-odoo: to fix in previous versions

src/components/composer/composer/composer.ts
```diff
@@ -735,15 +737,78 @@ export class Composer extends Component<CellComposerProps, SpreadsheetChildEnv>
...
+        Math.max(Math.min(description.maxArgPossible, nbrArgSupplied), description.minArgRequired)
+      )(argPosition);
+      return focusedArg !== undefined ? [focusedArg] : [];
+    }
+
+    // Otherwise, the user is still typing the formula, so we don't know yet how many arguments the user will supply.
+    // Consequently, we need to compute not only the argument to focus for the current number of arguments supplied
+    // but also for all the possible numbers of arguments supplied.
+
+    const minArgsNumberPossibility = Math.max(nbrArgSupplied, description.minArgRequired);
```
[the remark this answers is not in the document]
laa-odoo: yep, thx

src/formulas/composer_tokenizer.ts
```diff
@@ -208,8 +208,8 @@ function addArgsAST(tokens: EnrichedToken[]): EnrichedToken[] {
       }
       for (const argTokens of argsTokens) {
         let tokens = argTokens;
-        if (tokens.at(-1)?.type === "ARG_SEPARATOR") {
-          tokens = tokens.slice(0, -1);
+        if (tokens.at(0)?.type === "ARG_SEPARATOR") {
```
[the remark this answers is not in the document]
laa-odoo: This change is implemented concurrently with the change on line 166 of the same file. Here, we had a problem: the position corresponding to the separator was miscalculated, which meant that when we encountered a formula ending with an empty argument followed by a parenthesis, the cursor position for the empty argument (in this configuration, after the separator) was incorrect.

src/functions/arguments.ts
laa-odoo: The team has decided to continue in this direction:

`m, o, m` is a subcase of `m, r, m` --> So we can try to keep this implementation.

src/functions/arguments.ts
```diff
@@ -96,67 +97,163 @@ export function addMetaInfoFromArg(addDescr: AddFunctionDescription): FunctionDe
...
-  let previousArgRepeating: boolean | undefined = false;
-  let previousArgOptional: boolean | undefined = false;
-  let previousArgDefault: boolean | undefined = false;
+export function validateArguments(
+  args: ArgDefinition[],
+  nbrArgOptional: number,
+  nbrArgRepeating: number
+) {
+  if (nbrArgRepeating && nbrArgOptional >= nbrArgRepeating) {
+    throw new Error(_t("Function ${name} has more optional arguments than repeatable ones."));
```
LucasLefevre, line 252: wait, `${name}` is never replaced by any real value, is it ? Also, this error message doesn't need to be translated, it's for developers defining wrong arguments

src/functions/arguments.ts
```diff
@@ -166,24 +263,18 @@ export function validateArguments(args: ArgDefinition[]) {
...
-      throw new Error(
-        _t(
-          "Function ${name} has at mandatory arguments declared after optional ones. All optional arguments must be after all mandatory arguments."
-        )
-      );
+    if (current.repeating) {
+      if (!consecutiveRepeating && foundRepeating) {
+        throw new Error(
+          _t(
+            "Function ${name} has non-consecutive repeating arguments. All repeating arguments must be declared consecutively."
```
LucasLefevre, line 270: same here

src/functions/index.ts
```diff
@@ -128,9 +128,11 @@ function createComputeFunction(
 
     let vectorArgsType: VectorArgType[] | undefined = undefined;
 
+    const getArgToFocus = argTargeting(descr, args.length);
```
LucasLefevre, line 131: it looks like the evaluation is significantly slower on this branch :/ (+10% to evaluate the large formula dataset)

before | after
-- | --
314.2ms | 346.6ms


I didn't check where exactly it comes from, but most probably this line here isn't free and it's executed *a lot* of times

src/functions/index.ts
```diff
@@ -217,7 +219,8 @@ function createComputeFunction(
   ): Matrix<FunctionResultObject> | FunctionResultObject {
     for (let i = 0; i < args.length; i++) {
       const arg = args[i];
-      const argDefinition = descr.args[descr.getArgToFocus(i + 1) - 1];
+      const getArgToFocus = argTargeting(descr, args.length);
```
LucasLefevre, line 222: and it's executed a second time

src/functions/arguments.ts
laa-odoo: indeed :)

src/functions/index.ts
laa-odoo: thx for measurements

src/formulas/compiler.ts
```diff
@@ -316,43 +318,47 @@ function formulaArguments(tokens: Token[]) {
...
+    const nbrValueRepeating =
+      nbrArgRepeating * Math.floor((nbrArgSupplied - minArgRequired) / nbrArgRepeating);
+    const nbrValueRemaining =
+      nbrArgSupplied - minArgRequired - nbrValueRepeating - functionDefinition.nbrArgOptional;
+
+    if (nbrValueRemaining > 0) {
       throw new BadExpressionError(
         _t(
-          "Invalid number of arguments for the %s function. Expected all arguments after position %s to be supplied by groups of %s arguments",
+          "Invalid number of arguments for the %s function. Repeatable arguments are expected to be supplied by groups of %s argument(s) with maximum %s optional argument(s), but got %s argument(s) too many.",
```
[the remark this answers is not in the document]
laa-odoo: done :)

src/formulas/compiler.ts
```diff
@@ -316,43 +318,53 @@ function formulaArguments(tokens: Token[]) {
...
-          "Invalid number of arguments for the %s function. Expected all arguments after position %s to be supplied by groups of %s arguments",
-          functionName,
-          unrepeatableArgs.toString(),
-          repeatableArgs.toString()
+          "Invalid number of arguments for the %(functionName)s function. Repeatable arguments are expected to be supplied by groups of %(nbrArgRepeating)s argument(s) with maximum %(nbrArgOptional)s optional argument(s), but got %(nbrValueRemaining)s argument(s) too many.",
+          {
+            functionName: functionName,
+            nbrArgRepeating: nbrArgRepeating.toString(),
+            nbrArgOptional: functionDefinition.nbrArgOptional.toString(),
+            nbrValueRemaining: nbrValueRemaining.toString(),
```
LucasLefevre, line 366: `.toString()` used to be required for `_t`, but not anymore
```suggestion
            nbrArgRepeating,
            nbrArgOptional: functionDefinition.nbrArgOptional,
            nbrValueRemaining,
```

src/formulas/compiler.ts
```diff
@@ -316,43 +318,53 @@ function formulaArguments(tokens: Token[]) {
...
+      nbrArgSupplied - minArgRequired - nbrValueRepeating - functionDefinition.nbrArgOptional;
+
+    if (nbrValueRemaining > 0) {
       throw new BadExpressionError(
         _t(
-          "Invalid number of arguments for the %s function. Expected all arguments after position %s to be supplied by groups of %s arguments",
-          functionName,
-          unrepeatableArgs.toString(),
-          repeatableArgs.toString()
+          "Invalid number of arguments for the %(functionName)s function. Repeatable arguments are expected to be supplied by groups of %(nbrArgRepeating)s argument(s) with maximum %(nbrArgOptional)s optional argument(s), but got %(nbrValueRemaining)s argument(s) too many.",
```
LucasLefevre, line 361: I think it would still be a valid sentence if we just remove the `argument(s)` ? I would be better if we didn't have this (s). It's not a particularly good UX to see "0 argument(s)"

src/formulas/compiler.ts
```diff
@@ -316,43 +318,53 @@ function formulaArguments(tokens: Token[]) {
...
       _t(
-        "Invalid number of arguments for the %s function. Expected %s minimum, but got %s instead.",
-        functionName,
-        functionDefinition.minArgRequired.toString(),
-        nbrArg.toString()
+        "Invalid number of arguments for the %(functionName)s function. Expected %(minArgRequired)s minimum, but got %(nbrArgSupplied)s instead.",
+        {
+          functionName: functionName,
+          minArgRequired: minArgRequired.toString(),
+          nbrArgSupplied: nbrArgSupplied.toString(),
```
LucasLefevre, line 333:
```suggestion
          functionName,
          minArgRequired,
          nbrArgSupplied,
```

src/formulas/compiler.ts
```diff
@@ -316,43 +318,53 @@ function formulaArguments(tokens: Token[]) {
...
       _t(
-        "Invalid number of arguments for the %s function. Expected %s maximum, but got %s instead.",
-        functionName,
-        functionDefinition.maxArgPossible.toString(),
-        nbrArg.toString()
+        "Invalid number of arguments for the %(functionName)s function. Expected %(maxArgPossible)s maximum, but got %(nbrArgSupplied)s instead.",
+        {
+          functionName: functionName,
+          maxArgPossible: functionDefinition.maxArgPossible.toString(),
+          nbrArgSupplied: nbrArgSupplied.toString(),
```
LucasLefevre, line 346:
```suggestion
          functionName,
          maxArgPossible: functionDefinition.maxArgPossible.toString(),
          nbrArgSupplied,
```

src/functions/module_logical.ts
```diff
@@ -221,6 +227,53 @@ export const OR = {
...
+// -----------------------------------------------------------------------------
+
+export const SWITCH = {
+  description: _t("Returns a value by comparing cases to an expression."),
+  args: [
+    arg("expression (number, boolean, string)", _t("The value to be checked.")),
+    arg("case1 (number, boolean, string)", _t("The first case to be checked again expression.")),
+    arg("value1 (any)", _t("The corresponding value to be returned if case1 matches expression.")),
+    arg(
+      "case2 (any, repeating)",
```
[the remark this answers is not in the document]
laa-odoo: Here, we don't specifically want to detect typing errors upstream of the function. Functionally, the function should not return an error on these "cases" until the function attempts to read the "case".

We only allowed "case1" to be typed as `number, boolean, string` because we need to read this case at a minimum for the function to work.

laa-odoo:
@LucasLefevre we are okay

LucasLefevre:
robodoo r+

LucasLefevre:
robodoo rebase-ff

PR #5413, opened 2025-01-07
[IMP] errors: display error origin position

LucasLefevre:
## Description:

Purpose
-------

Let's say a cell is evaluated to an error because one of its inputs is an error itself. The tooltip shows the error message, but you don't know where it comes from.

If the formula has many reference, you have to check every one of those references to find the original error.

As an example, if the formula =IF(OR(AND($D21<G$17,$D21>I$1),AND(I$1<$E21,I$1>$D21)),1,"") results with an error. Where does it come from ? D21 ? J1 ? I1 ? E21 ? 

Spec
----

Add below the error message the cell from which the error is coming from. The cell reference should be clickable to jump to cell.
Task: [4452745](https://www.odoo.com/odoo/2328/tasks/4452745)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/components/error_tooltip/error_tooltip.ts
```diff
@@ -29,28 +30,64 @@ export interface ErrorToolTipMessage {
...
+    let evaluationError: EvaluatedCell | undefined;
     if (cell.type === CellValueType.error && !!cell.message) {
-      errors.push({
-        title: _t("Error"),
-        message: cell.message,
-      });
+      evaluationError = cell;
     }
 
     const validationErrorMessage = getters.getInvalidDataValidationMessage(position);
```
rrahir, line 93: just inject that value directly to the errorTooltip component as you did with the EvaluationError. The builder is no longer relevant if it only manages a single case. Or you could keep the builder and have it yield a component... but the former seems easier

rrahir (commented):
the builder seems useless?

tests/evaluation/evaluation.test.ts
```diff
@@ -1086,8 +1091,51 @@ describe("evaluateCells", () => {
     expect(getCellError(model, "C1")).toBe("Invalid expression");
   });
 
-  // TO DO: add tests for exp format (ex: 4E10)
-  // RO DO: add tests for DATE string format (ex match: "28 02 2020")
```
LucasLefevre, line 1090: 5 years old TODOs... :stuck_out_tongue:

LucasLefevre:
robodoo rebase-ff

pro-odoo (approved):
robodoo r+

PR #5429, opened 2025-01-08
[FIX] pivot_side_panel: drag and drop of dimensions

dhrp-odoo:
## Description:

The pivot side panel did not scroll during drag-and-drop of dimensions when multiple dimensions were present.

This occurred because the dimensions container lacked a specific height. Adding a height to fill the available space would only enable scrolling within the dimensions list, not the entire side panel.

This fix resolves the issue by passing the pivot side panel's reference as the container to drag and drop hook for it's calculations, leveraging its `100vh` height. This approach ensures that scrolling affects the entire side panel, allowing drag-and-drop to function correctly.

Task: [3817565](https://www.odoo.com/odoo/2328/tasks/3817565)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/pivot/pivot_layout_configurator/pivot_layout_configurator.ts
+++ b/src/components/side_panel/pivot/pivot_layout_configurator/pivot_layout_configurator.ts
@@ -26,6 +26,7 @@ interface Props {
   unusedMeasureFields: PivotField[];
   unusedDateTimeGranularities: Record<string, Set<string>>;
   allGranularities: string[];
+  getScrollableContainerEl?: () => HTMLElement;
 }
 
 export class PivotLayoutConfigurator extends Component<Props, SpreadsheetChildEnv> {
@@ -43,6 +44,7 @@ export class PivotLayoutConfigurator extends Component<Props, SpreadsheetChildEn
     unusedMeasureFields: Array,
     unusedDateTimeGranularities: Object,
     allGranularities: Array,
+    getScrollableContainerEl: { type: Function, optional: true },
   };
 
   private dimensionsRef = useRef("pivot-dimensions");
@@ -74,7 +76,7 @@ export class PivotLayoutConfigurator extends Component<Props, SpreadsheetChildEn
       draggedItemId: dimension.nameWithGranularity,
       initialMousePosition: event.clientY,
       items: draggableItems,
-      scrollableContainerEl: this.dimensionsRef.el!,
+      scrollableContainerEl: this.props.getScrollableContainerEl?.() || this.dimensionsRef.el!,
       onDragEnd: (dimensionName, finalIndex) => {
         const originalIndex = draggableIds.findIndex((id) => id === dimensionName);
         if (originalIndex === finalIndex) {
@@ -124,7 +126,7 @@ export class PivotLayoutConfigurator extends Component<Props, SpreadsheetChildEn
       draggedItemId: measure.name,
       initialMousePosition: event.clientY,
       items: draggableItems,
-      scrollableContainerEl: this.dimensionsRef.el!,
+      scrollableContainerEl: this.props.getScrollableContainerEl?.() || this.dimensionsRef.el!,
       onDragEnd: (measureName, finalIndex) => {
         const originalIndex = draggableIds.findIndex((id) => id === measureName);
         if (originalIndex === finalIndex) {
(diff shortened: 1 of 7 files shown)


src/components/side_panel/pivot/pivot_layout_configurator/pivot_layout_configurator.ts
```diff
@@ -26,6 +26,7 @@ interface Props {
   unusedMeasureFields: PivotField[];
   unusedDateTimeGranularities: Record<string, Set<string>>;
   allGranularities: string[];
+  pivotSidePanelRef: Ref<HTMLElement>;
```
dhrp-odoo, line 29: I am uncertain whether passing props to the `PivotLayoutConfigurator` is the good approach. I explored alternatives but couldn't find a solution without significantly altering the existing code :/

hokolomopo: Not sure either TBH. I don't have a good solution that comes to mind. Maybe whoever reviews this will have a better idea ?

src/components/side_panel/pivot/pivot_layout_configurator/pivot_layout_configurator.ts
```diff
@@ -74,7 +76,7 @@ export class PivotLayoutConfigurator extends Component<Props, SpreadsheetChildEn
       draggedItemId: dimension.nameWithGranularity,
       initialMousePosition: event.clientY,
       items: draggableItems,
-      containerEl: this.dimensionsRef.el!,
+      containerEl: this.props.pivotSidePanelRef.el!,
```
hokolomopo, line 79: I'm not sure I understand your commit message perfectly. `This occurred because the dimensions container lacked a specific height` that's not really the problem is it ? We were passing the dragged element as argument instead of the container, which was plain wrong.

src/components/side_panel/pivot/pivot_layout_configurator/pivot_layout_configurator.ts
```diff
@@ -124,7 +126,7 @@ export class PivotLayoutConfigurator extends Component<Props, SpreadsheetChildEn
       draggedItemId: measure.name,
       initialMousePosition: event.clientY,
       items: draggableItems,
-      containerEl: this.dimensionsRef.el!,
+      containerEl: this.props.pivotSidePanelRef.el!,
```
hokolomopo, line 129: Maybe we should rename this argument `scrollableContainerEl`. Maybe it'd be slighlty better and easier to understand what it's for ? Not sure. What do you think ?

hokolomopo (commented):
Nice 👍 Cool to have this fixed

src/components/side_panel/pivot/pivot_layout_configurator/pivot_layout_configurator.ts
dhrp-odoo: Good point! Renaming `containerEl` to `scrollableContainerEl` makes its purpose more explicit. Do you think this change should instead be made in `master`?

hokolomopo: You can do it in master if you want it's slightly more correct, but I wouldn't bother to create another PR for that. Your choice :)

src/components/side_panel/pivot/pivot_layout_configurator/pivot_layout_configurator.ts
dhrp-odoo: Yes, that's correct. We were passing the wrong container, and it also didn't have any height, which caused the issue.

src/components/side_panel/pivot/pivot_layout_configurator/pivot_layout_configurator.ts
pro-odoo: I see two issues with this props: 
- increase of the coupling between `PivotLayoutConfigurator` and the side panel in which this component is used
- Owl will proxify an entire HTMLElement, with can be huge IMO.

To avoid this, my proposition is to replace this props by a callback `getParentElement` that would return the el of the side panel.
Not sure it's necessary, but we could make this props optional and use `dimensionRef` is the props is not defined.

What do you think ?

dhrp-odoo: Good point! Passing a callback seems better. I'll update it. Thanks!

pro-odoo (approved):
robodoo r+

pro-odoo:
robodoo rebase-ff
Thanks !

PR #5282, opened 2024-11-28
[FIX] chart: fix misalignment with trend line axis

anhe-odoo:
## Task Description

When inserting a trend line on a line chart, depending on the data we sometimes have a misplaced trend line axis giving an alignment issue between the trend line and the original plot. For example, if you have these data :

||A|B|
|-|-|-|
|1|1|1|
|2|2|2|
|3|3|3
|4|4|4|
|5|5|5|
|6|6|6|
|7|7|7|
|8|8|8|

When you try to insert a linear trend line, the line won't be aligned to the original line chart. This PR aims to fix this by changing the labels and axis of the trend line.

## Related Task

- Task: 4251695

## review checklist

- [ ] feature is organized in plugin, or UI components
- [ ] support of duplicate sheet (deep copy)
- [ ] in model/core: ranges are Range object, and can be adapted (adaptRanges)
- [ ] in model/UI: ranges are strings (to show the user)
- [ ] undo-able commands (uses this.history.update)
- [ ] multiuser-able commands (has inverse commands and transformations where needed)
- [ ] new/updated/removed commands are documented
- [ ] exportable in excel
- [ ] translations (\_t("qmsdf %s", abc))
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes
- [ ] doc is rebuild (npm run doc)
- [ ] status is correct in Odoo

PR #5430, opened 2025-01-08
[FIX] chart: fix trend line axis alignment

anhe-odoo:
## Task Description

Following https://github.com/odoo/o-spreadsheet/pull/5282, we now have the trend line data as an array of Point, and try to align the trend line axis to the real x-axis. This can't be done for categorical axis and time axis without treating the trend axis as a category axis, for both.

This PR aims to fix the axis for the datetime data, where the axis was considered as a datetime axis but can't be anymore. As this is caused by the way chartJs decides to align the axis, we don't have any other solution, neither have we a possibility to test the fix.


rrahir:
@robodoo r+

PR #5431, opened 2025-01-08
[FIX] pivot: notify only with static formulas

LucasLefevre:
## Description:

When we update a pivot from it's side panel with a static pivot present in the viewport, we have the warning "Pivot updates only work with dynamic pivot tables". But the warning is also present if you have no static pivot anywhere, but that there is no dynamic pivot in the viewport (eg. you scolled a bit after opening the pivot side panel).

The warning should only be displayed if you have a static pivot somewhere in the viewport.

Task: [4453844](https://www.odoo.com/odoo/2328/tasks/4453844)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


rrahir:
@robodoo r+

PR #5443, opened 2025-01-08
[FIX] Selection: selection follows the moved header

rrahir:
Following the refactoring of the clipboard in #2857, we also isolated the flow of `MOVE_COLUMNS_ROWS` so it would not interfere with the clipboard. By doing so, we lost the capacity to select the moved header during the handling of the command.

How to reproduce:
- select row 2
- paint it red to identify it
- move it to row 4

The selection is still on row 2 and not on row 4 as expected.

Task: 4461901

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_stateful/selection.ts
+++ b/src/plugins/ui_stateful/selection.ts
@@ -683,6 +683,11 @@ export class GridSelectionPlugin extends UIPlugin {
     ];
     handler.paste({ zones: pasteTarget }, data, { isCutOperation: true });
 
+    const selection = pasteTarget[0];
+    const col = selection.left;
+    const row = selection.top;
+    this.setSelectionMixin({ zone: selection, cell: { col, row } }, [selection]);
+
     const toRemove = isBasedBefore ? cmd.elements.map((el) => el + thickness) : cmd.elements;
     let currentIndex = cmd.base;
     for (const element of toRemove) {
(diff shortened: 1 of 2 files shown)


src/plugins/ui_stateful/selection.ts
```diff
@@ -683,6 +683,12 @@ export class GridSelectionPlugin extends UIPlugin {
     ];
     handler.paste({ zones: pasteTarget }, data, { isCutOperation: true });
 
+    const selection = pasteTarget[0];
+    const col = selection.left;
+    const row = selection.top;
+    this.selection.getBackToDefault();
```
hokolomopo, line 689: Is it even possible to drag & drop a header if we're in the composer of a selection input ? In the codebase we sometime use `getBackToDefault` before moving the selection, and sometime we don't, it all seems very random and arbitrary ...

LucasLefevre (commented):
robodoo r+

src/plugins/ui_stateful/selection.ts
rrahir: - I don't think it's random - I explicitely want the selection to be updated, not the other potential listeners. I know that by the end of my manipylation, it's the selection that should have the control. Should it be done here or before calling the command ?
Or even bypass the selection processor and update the selection internally since we happen to be in the right plugin ? I'm not sure.
I agree that the 2 last solutions seem to make more sense but editing the selection directly inside the plugin seems more straightforward with the other way the plugin works. I changed it accordinly

PR #5449, opened 2025-01-10
[FIX] selection: drag and drop resized cols and rows

dhrp-odoo:
## Description:

Previously, resizing a column or row and then dragging it to a new position (right or down) did not correctly resize the dropped column/row

This issue occurred because the resizing logic incorrectly used the original index of the column or row instead of adjusting for its new position. The correct approach requires using `index + 1` for "after" operations to account for the updated position, but the implementation used the unadjusted index, leading to the bug.

Task: [4454025](https://www.odoo.com/odoo/2328/tasks/4454025)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/tests/plugins/selection.test.ts
+++ b/tests/plugins/selection.test.ts
@@ -1072,22 +1072,30 @@ describe("move elements(s)", () => {
     const model = new Model();
     resizeColumns(model, ["A"], 10);
     resizeColumns(model, ["C"], 20);
-    moveColumns(model, "D", ["A"]);
+    moveColumns(model, "C", ["A"], "after");
     const sheetId = model.getters.getActiveSheetId();
     expect(model.getters.getColSize(sheetId, 0)).toEqual(DEFAULT_CELL_WIDTH);
     expect(model.getters.getColSize(sheetId, 1)).toEqual(20);
     expect(model.getters.getColSize(sheetId, 2)).toEqual(10);
+    moveColumns(model, "A", ["C"], "before");
+    expect(model.getters.getColSize(sheetId, 0)).toEqual(10);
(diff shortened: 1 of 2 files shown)


tests/plugins/selection.test.ts
```diff
@@ -1072,7 +1072,7 @@ describe("move elements(s)", () => {
     const model = new Model();
     resizeColumns(model, ["A"], 10);
     resizeColumns(model, ["C"], 20);
-    moveColumns(model, "D", ["A"]);
```
hokolomopo, line 1075: Maybe instead of modifying this line you could add some lines at the end of the test to test with `"after"`. That way we test both `before` and `after` behaviours.

tests/plugins/selection.test.ts
```diff
@@ -1083,7 +1083,7 @@ describe("move elements(s)", () => {
     const model = new Model();
     resizeRows(model, [0], 10);
     resizeRows(model, [2], 20);
-    moveRows(model, 3, [0]);
```
hokolomopo, line 1086: same here

hokolomopo (commented):
Small comment on the test, otherwise seems good :)

tests/plugins/selection.test.ts
dhrp-odoo: If I understand the code correctly, the original test case was incorrect. The "after" operation should only be used when moving down or to the right (to a higher row number or later column alphabet). Similarly, the "before" operation should only be used when moving up or to the left (to a lower row number or earlier column alphabet).

image `selection.ts`

I've updated the test to cover both operations. Please let me know if I’ve misunderstood the code.

LucasLefevre (approved):
robodoo r+
Thanks!

PR #5452, opened 2025-01-10
Master no shared borders rar

rrahir:
## Description:

Task: [4458245](https://www.odoo.com/odoo/2328/tasks/4458245)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/core/borders.ts
+++ b/src/plugins/core/borders.ts
@@ -1,5 +1,6 @@
 import { DEFAULT_BORDER_DESC } from "../../constants";
 import {
+  deepCopy,
   deepEquals,
   getItemId,
   groupConsecutive,
@@ -14,7 +15,6 @@ import {
   AddColumnsRowsCommand,
   Border,
   BorderDescr,
-  BorderDescription,
   BorderPosition,
   CellPosition,
   Color,
@@ -30,7 +30,7 @@ import {
 import { CorePlugin } from "../core_plugin";
 
 interface BordersPluginState {
-  readonly borders: Record<UID, (BorderDescription[] | undefined)[] | undefined>;
+  readonly borders: Record<UID, ((Border | undefined)[] | undefined)[] | undefined>;
 }
 /**
  * Formatting plugin.
@@ -70,7 +70,7 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
           // map and slice preserve empty values and do not set `undefined` instead
           const bordersCopy = borders
             .slice()
-            .map((col) => col?.slice().map((border) => ({ ...border })));
+            .map((col) => col?.slice().map((border) => deepCopy(border)));
           this.history.update("borders", cmd.sheetIdTo, bordersCopy);
         }
         break;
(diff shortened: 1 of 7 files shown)


src/plugins/core/borders.ts
```diff
@@ -105,53 +104,15 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
...
-              for (let col = 0; col < this.getters.getNumberCols(cmd.sheetId); col++) {
-                this.history.update("borders", cmd.sheetId, col, 0, "horizontal", undefined);
-              }
-            }
-            const zone = this.getters.getRowsZone(
-              cmd.sheetId,
-              group[group.length - 1] + 1,
-              group[0]
-            );
+            const start = Math.min(group[0], group[group.length - 1]);
```
anhe-odoo, line 113: Shouldn't we use this `start` value at line 117 instead of `group[0] + 1` ?

anhe-odoo: (same comment for below)

src/plugins/core/borders.ts
```diff
@@ -182,7 +143,7 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
       colLeftOfInsertion = cmd.base - 1;
       colRightOfInsertion = cmd.base + cmd.quantity;
     } else {
-      this.shiftBordersHorizontally(cmd.sheetId, cmd.base + 1, cmd.quantity, {
+      this.shiftBordersHorizontally(cmd.sheetId, cmd.base, cmd.quantity, {
```
anhe-odoo, line 146: Why did you change these lines but not the one above ? (l111 vs l146 and l117 vs l170) ?

LucasLefevre (commented):
noice

src/plugins/core/borders.ts
```diff
@@ -220,16 +181,8 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
...
-      top: this.borders[sheetId]?.[col]?.[row]?.horizontal,
-      bottom: this.borders[sheetId]?.[col]?.[row + 1]?.horizontal,
-      left: this.borders[sheetId]?.[col]?.[row]?.vertical,
-      right: this.borders[sheetId]?.[col + 1]?.[row]?.vertical,
-    };
-    if (!border.bottom && !border.left && !border.right && !border.top) {
-      return null;
-    }
-    return border;
+    const border = this.borders[sheetId]?.[col]?.[row];
```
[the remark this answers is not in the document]
rrahir: This is something I considered, but I figured it could wait for when we'll try to introduce infinite spreadsheets.

src/plugins/core/borders.ts
rrahir: apparently I can't read, the elements were sorted in descending order on purpose on line 104 so there was no need to redefin start/end to begin with.

src/plugins/core/borders.ts
rrahir: short: the tests pass so I didn't need to adapt it.
Long: it's been 2 hours and stillcan't figure out everything that happns. removing the +1 sees to have no impact, still on it

rrahir: So the functions 'shiftBorderXlly' seems to do some "scabreux" stuff depending on wheter or not we would provide a certain parameter, I guess this was kind of necessary when we had the previous data structure but now it doesn't seem required anymore. changed the behaviour so that the function is more strzaightforward on its purpose: shift all headers from a starting point from a given delta. Hence, the "before" insertion should act on the base header of the command (no +1) and the "after" insertion should only apply on the header after the insertion (hence the +1)

LucasLefevre (approved):
robodoo r+

LucasLefevre:
robodoo r+

PR #5456, opened 2025-01-13
[IMP] formulas: add support for '+' as first character for formulas

Rachico:
## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/composer/composer/cell_composer_store.ts
+++ b/src/components/composer/composer/cell_composer_store.ts
@@ -2,6 +2,7 @@ import { parseLiteral } from "../../../helpers/cells";
 import {
   formatValue,
   isDateTimeFormat,
+  isFormula,
   markdownLink,
   numberToString,
   parseDateTime,
@@ -82,7 +83,7 @@ export class CellComposerStore extends AbstractComposerStore {
         }
         break;
       case "ACTIVATE_SHEET":
-        if (!this._currentContent.startsWith("=")) {
+        if (!isFormula(this._currentContent)) {
           this._cancelEdition();
           this.resetContent();
         }
@@ -179,7 +180,7 @@ export class CellComposerStore extends AbstractComposerStore {
     if (content) {
       const sheetId = this.getters.getActiveSheetId();
       const cell = this.getters.getEvaluatedCell({ sheetId, col: this.col, row: this.row });
-      if (cell.link && !content.startsWith("=")) {
+      if (cell.link && !isFormula(content)) {
         content = markdownLink(content, cell.link.url);
       }
       this.addHeadersForSpreadingFormula(content);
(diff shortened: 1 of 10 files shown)


src/helpers/cells/cell_evaluation.ts
```diff
@@ -37,6 +37,7 @@ export function evaluateLiteral(
 
 export function parseLiteral(content: string, locale: Locale): CellValue {
   if (content.startsWith("=")) {
+    // if (content.startsWith("=") || content.startsWith("+")) {
```
laa-odoo, line 40: :eyes:

LucasLefevre (commented):
the repeated patter `.startsWith("=") || .startsWith("+")` could be extracted to a nice little function

LucasLefevre (commented):
actually, I think all this could be greatly simplified. Most of the code should not care about this except maybe the composer and maybe `canonicalizeFormula`

src/components/composer/grid_composer/grid_composer.ts
```diff
@@ -146,7 +147,7 @@ export class GridComposer extends Component<Props, SpreadsheetChildEnv> {
     if (this.composerStore.editionMode === "inactive") {
       return `z-index: -1000;`;
     }
-    const isFormula = this.composerStore.currentContent.startsWith("=");
+    const isFormula = _isFormula(this.composerStore.currentContent);
```
hokolomopo, line 150: nitpick: i'd rename the variable rather than renaming the import

tests/composer/composer_component.test.ts
```diff
@@ -100,6 +100,13 @@ describe("ranges and highlights", () => {
     expect(contentColors["SU"]).toBeSameColorAs("#000000");
   });
 
+  test("+SU, the + should be colored", async () => {
```
hokolomopo, line 103: I think you can just have 1-2 tests testing that the formula is recognized as such when starting with `+` and that it has highlight/formula helper/that the + is replaced by = on confirm. 

I'm not sure all of the tests you duplicated are that useful. This file is already the slowest file to run by far, let's not bloat it too much.

Or if you really want to have a lot of tests just to be sure, you could test the store instead of the component (`composer_store.test.ts`). It's way faster since we don't have to mount a component and bother with the DOM.

src/components/composer/composer/cell_composer_store.ts
```diff
@@ -179,7 +180,7 @@ export class CellComposerStore extends AbstractComposerStore {
     if (content) {
```
hokolomopo, line 180: There's still a `startsWith("=")` in this file.

hokolomopo (commented):
ThumbsUp.jpg

tests/composer/composer_component.test.ts
Rachico: We now have 11 tests in total for this new feature, I think it's fair @hokolomopo

pro-odoo (approved):
robodoo r+

PR #5457, opened 2025-01-13
[IMP] BorderEditorWidget: Reset position on close

rrahir:
Currently, once we used the border widget from the top bar to add a border and the border position is set (i.e. you already used the widget once), changing any other parameter will dispatch the change.

This is quite infuriating when trying to set very different borders in succession since if you only change the color, it will dispatch a change of border even though you did not have a chance to change the type of position beforehand. This creates situations where you need additional steps to get to the wanted result.

E.g.

    Add an external  border on A1
    Select A2 to add a bottom border in another color

-> you either have to change the color first and then it applied an external border or you first change the type of border and it applies it then you have to open the widget again to select the right color, which will end up giving the expected result.

This revision ensures that the position type is reset when closing the border widget so that the user can always choose which type of border to apply without frustrating and unexpected steps.

Task: 4475869

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/border_editor/border_editor_widget.ts
+++ b/src/components/border_editor/border_editor_widget.ts
@@ -1,10 +1,10 @@
-import { Component, useRef, useState } from "@odoo/owl";
+import { Component, onWillUpdateProps, useRef, useState } from "@odoo/owl";
 import { DEFAULT_BORDER_DESC } from "../../constants";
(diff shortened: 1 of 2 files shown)


LucasLefevre (approved):
robodoo r+

PR #5486, opened 2025-01-15
[FIX] TextValueProvider: Avoid duplicate key in template

rrahir:
The template of `TextValueProvider` would rely on proposal.text as an identification key but there are no unicity constraints on that value.

Task: 4483494

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/composer/autocomplete_dropdown/autocomplete_dropdown.xml
+++ b/src/components/composer/autocomplete_dropdown/autocomplete_dropdown.xml
@@ -4,7 +4,7 @@
       t-ref="autoCompleteList"
       t-att-class="{
           'o-autocomplete-dropdown': props.proposals.length}">
-      <t t-foreach="props.proposals" t-as="proposal" t-key="proposal.text">
+      <t t-foreach="props.proposals" t-as="proposal" t-key="proposal.text + proposal_index">
         <div
           class="d-flex flex-column text-start"
           t-att-class="{'o-autocomplete-value-focus': props.selectedIndex === proposal_index}"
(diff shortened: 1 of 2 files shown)


src/components/composer/autocomplete_dropdown/autocomplete_dropdown.xml
```diff
@@ -4,7 +4,7 @@
       t-ref="autoCompleteList"
       t-att-class="{
           'o-autocomplete-dropdown': props.proposals.length}">
-      <t t-foreach="props.proposals" t-as="proposal" t-key="proposal.text">
+      <t t-foreach="props.proposals" t-as="proposal" t-key="proposal_index">
```
LucasLefevre, line 7: I just asked Géry because I wasn't entirely sure because in practice it looks like this is working but...

*les clés _index, c'est uniquement ok si tu es sur a l'avance que ta liste est statique (ou éventuellement, si on ajoute des éléments, mais sans en enlever)
autrement, tu vas réconcilier des élém différents, et en général ca va marcher, en surface, sauf que ca risque de faire des trucs bizarres. par exemple, si l'utilisateur écrit dans un input, l'input va etre associé a un autre élément*

LucasLefevre: maybe a composite key would do the trick without the shortcomings of just `_index`
```suggestion
      <t t-foreach="props.proposals" t-as="proposal" t-key="proposal.text + proposal_index">
```

rrahir: yup forgot that the proposals change over time ><

PR #5494, opened 2025-01-16
[REF] tests: remove mock content editable helper

LucasLefevre:
## Description:

We have a class `ContentEditableHelper` which is a helper class designed
to manipulate the DOM and the selection of the composer element.

It has been mocked since the begining of o-spreadsheet, which means all the
business logic inside isn't tested at all!

It was mocked because jsdom (the lib implementing/mocking DOM for running
tests in nodejs) didn't support the Selection API at the time this project
was started.

However, the Selection API was implemented back in 2019 and we've updated
jsdom since then.

`ContentEditableHelper` has grown over the years and it now contains even more
business logic which is worth testing.
And it's probably meant to grow even more in the future as we add features
to the composer (as an example, see task 4471274).

This commit removes our mock and we now rely on jsdom implementation.

Caveats:
- jsdom doesn't update the selection when arrow keys (right or left) are
  pressed. We have to manual set the selection at the right place.
- jsdom doesn't support having the focusNode/offset before the anchorNode/offset
  (selecting multiple characters from right-to-left). Hence it can now longer
  be tested and one test had to be removed.

Task: [4491759](https://www.odoo.com/odoo/2328/tasks/4491759)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


tests/composer/composer_component.test.ts
```diff
@@ -896,7 +962,7 @@ describe("composer", () => {
       expect(composerStore.currentContent).toEqual("Azerty");
       expect(composerStore.composerSelection).toEqual({ start: 6, end: 6 });
 
-      composerStore.changeComposerCursorSelection(6, 4);
+      composerStore.changeComposerCursorSelection(4, 6);
```
LucasLefevre, line 964: right-to-left selection isn't supported

tests/composer/composer_component.test.ts
```diff
@@ -777,28 +850,20 @@ describe("composer", () => {
     expect(composerStore.editionMode).toBe("inactive");
   });
 
-  test("Select a right-to-left range with the keyboard", async () => {
```
LucasLefevre, line 780: right-to-left selection isn't supported

src/components/composer/content_editable_helper.ts
```diff
@@ -190,7 +194,7 @@ export class ContentEditableHelper {
     const focusedNode = document.getSelection()?.focusNode;
     if (!focusedNode || !this.el.contains(focusedNode)) return;
     const element = focusedNode instanceof HTMLElement ? focusedNode : focusedNode.parentElement;
-    element?.scrollIntoView({ block: "nearest" });
+    element?.scrollIntoView?.({ block: "nearest" });
```
rrahir, line 197: jsdom?

src/components/helpers/dom_helpers.ts
```diff
@@ -118,8 +118,6 @@ function findSelectionIndex(el: HTMLElement, nodeToFind: Node, nodeOffset: numbe
      * In this case, we return 0 because we don't know the size of the text before the selection.
      *
      * A known occurrence is triggered since the introduction of commit d4663158 (PR #2038).
-     *
-     * FIXME: find a way to test even though the selection API is not available in jsDOM.
```
rrahir: 🎆

src/components/composer/content_editable_helper.ts
LucasLefevre: yes, it's not implemented (`element.scrollIntoView` is undefined)

tests/test_helpers/helpers.ts
```diff
@@ -720,6 +736,61 @@ export async function editStandaloneComposer(
...
+      lineText += el.textContent;
+    }
+    currentTextPerLines.push(lineText);
+  }
+  /**
+   * This simulates the insertion of text in the composer DOM and sets the selection
+   * (focusNode, anchorNode and offsets) at a sensible position in the DOM.
+   * This does not 100% reflect the actual behavior of browsers (they all behave in slightly
+   * different ways anyway).
+   * If you want to test a very specific behavior about the selection, you should probably
```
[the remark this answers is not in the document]
LucasLefevre: Moved to docstring.
I wanted to make it clear the behavior should not be taken as "THE reality" if someone in the future uses the `insertText` without knowing the specifics/caveats of Selection and the result is not the one he expected

rrahir:
@robodoo r+

LucasLefevre:
robodoo rebase-ff

PR #5497, opened 2025-01-20
[IMP] formulas: add a cache on linear search functions

laa-odoo:
Benchmark for 10 000 Vlookup performed on a random array composed of integer values ​​between 1 and 10000
`=VLOOKUP(A1:A10000,A1:A10000,1,false)`

measures on "linearSearch" before the commit:
- 267 ms (48% of evaluateAllCells)
- 271 ms (48% of evaluateAllCells)
- 296 ms (50% of evaluateAllCells)

measures on "linearSearch" after the commit:
- 4 ms (2% of evaluateAllCells)
- 1 ms (0.5% of evalauteAllCells)
- 2 ms (3 % of evaluateAllCells)

Task: [4080146](https://www.odoo.com/odoo/2328/tasks/4080146)

### DIFF
--- a/src/plugins/ui_core_views/cell_evaluation/evaluator.ts
+++ b/src/plugins/ui_core_views/cell_evaluation/evaluator.ts
@@ -137,6 +137,10 @@ export class Evaluator {
     );
     this.compilationParams.evalContext.updateDependencies = this.updateDependencies.bind(this);
     this.compilationParams.evalContext.addDependencies = this.addDependencies.bind(this);
+    this.compilationParams.evalContext.lookupCaches = {
+      forwardSearch: new Map(),
+      reverseSearch: new Map(),
+    };
   }
 
   private createEmptyPositionSet() {
--- a/src/functions/helpers.ts
+++ b/src/functions/helpers.ts
@@ -15,6 +15,7 @@ import {
   isMatrix,
 } from "../types";
 import { CellErrorType, EvaluationError, errorTypes } from "../types/errors";
+import { LookupCaches } from "../types/functions";
 
 const SORT_TYPES_ORDER = ["number", "string", "boolean", "undefined"];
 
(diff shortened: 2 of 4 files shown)


src/plugins/ui_core_views/cell_evaluation/evaluator.ts
```diff
@@ -137,6 +138,10 @@ export class Evaluator {
     );
     this.compilationParams.evalContext.updateDependencies = this.updateDependencies.bind(this);
     this.compilationParams.evalContext.addDependencies = this.addDependencies.bind(this);
+    this.compilationParams.evalContext.lookupCaches = {
+      forwardSearch: new Map<any, Map<CellValue, number>>(),
+      reverseSearch: new Map<any, Map<CellValue, number>>(),
+    };
```
hokolomopo, line 144:
```suggestion
    this.compilationParams.evalContext.caches= {};
```
nitpick. And the the function can just set `this.cache = new Map()` That way each function can setup its cache as needed, and the evaluator doesn't have to know which function needs a cache.

LucasLefevre (commented):
When filled with data, I measured a 28% improvement on HBA's spreadsheet evaluation (~5.1s -> 3.7s).
When there's no data, there's no improvement because there's no exact match, but it's an unrealistic situation.

Looking good :+1:  Only nitpicking comments.

LucasLefevre:
Oh also: could you backport to 18.0 ?

src/plugins/ui_core_views/cell_evaluation/evaluator.ts
laa-odoo: I am looking for an even more general solution which will automatically include all calculation times that are too long at the scale of the evaluation. We can also discuss it again, I would be curious to have your opinion :)

LucasLefevre:
merged with https://github.com/odoo/o-spreadsheet/pull/5560

PR #5498, opened 2025-01-20
[IMP] Conditional Formatting: Have static reference by default

fdamhaut:
Task: 4498442

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/components/composer/standalone_composer/standalone_composer_store.ts
```diff
@@ -32,6 +34,17 @@ export class StandaloneComposerStore extends AbstractComposerStore {
...
 
+  /**
+   * Replace the current reference selected by the new one.
+   * */
+  protected getZoneReference(zone: Zone | UnboundedZone): string {
+    let res = super.getZoneReference(zone);
+    if (this.args().defaultStatic) {
+      res = setXcToFixedReferenceType(res, "colrow");
+    }
+    return res;
```
LucasLefevre, line 45:
```suggestion
    const ref = super.getZoneReference(zone);
    if (this.args().defaultStatic) {
      return setXcToFixedReferenceType(res, "colrow");
    }
    return ref;
```

VincentSchippefilt:
robodoo r+

src/components/composer/standalone_composer/standalone_composer_store.ts
```diff
@@ -32,6 +34,17 @@ export class StandaloneComposerStore extends AbstractComposerStore {
     return providersDefinitions;
   }
 
+  /**
+   * Replace the current reference selected by the new one.
+   * */
+  protected getZoneReference(zone: Zone | UnboundedZone): string {
+    const res = super.getZoneReference(zone);
+    if (this.args().defaultStatic) {
+      return setXcToFixedReferenceType(res, "colrow");
```
rrahir, line 43: the helper lied to you! it only handles since cell XC and you have to split the reference in two subparts and call the helper independently on each part like in https://github.com/odoo/o-spreadsheet/blob/18.0/src/helpers/reference_type.ts#L16-L23. I think the helper can be fixed :p

PR #5499, opened 2025-01-21
[IMP] package: add sass support

rrahir:
This commit adds sass support to the project.

Some points to note:
- All *.scss files inside `src/components` are taken into account.
- As this project is aimed to be used in Odoo, the scss files are bundled
  into two files `o_spreadsheet.scss` and `o_spreadsheet_variables.scss`
  which are then included in the Odoo assets.
  - `o_spreadsheet_variables.scss` contains all the variables used in the
    project. It corresponds to the `src/variables.scss` file. It should
    be included **before** `o_spreadsheet.scss` in Odoo assets.
  - `o_spreadsheet.scss` contains all the styles used in the project. It
    It should be included **after** `o_spreadsheet_variables.scss` in Odoo
    assets.
  - For demo, a single file `spreadsheet.css` is created which includes
    both `variables.scss` and all others `scss` files compiled.

Task: 4501078


package.json
```diff
@@ -64,6 +65,7 @@
     "@types/rbush": "^3.0.3",
     "babel-eslint": "^10.1.0",
     "body-parser": "^1.19.0",
+    "bundle-scss": "^1.5.4",
```
LucasLefevre, line 68: only 733 weekly download
Is it well maintained ? If it's not used a lot, it might not be. Is doesn't really inspire trust

package.json
```diff
@@ -13,15 +13,15 @@
...
+    "serve-static": "live-server --open=demo --watch=build/o_spreadsheet.iife.js,build/o_spreadsheet.xml,build/o_spreadsheet.css,main.css,demo",
     "dev": "npm-run-all --print-label bundle:dev --parallel server serve-static watch",
     "server": "node tools/server/main.cjs",
     "build:js": "tsc --module es6 --incremental",
     "bundle:iife": "rollup -c -m -- --format iife",
     "bundle:esm": "rollup -c -m -- --format esm",
     "bundle:xml": "node tools/bundle_xml/main.cjs",
-    "bundle:dev": "npm-run-all build:js bundle:iife \"bundle:xml -- --outDir build\"",
-    "dist": "tsc --module es6 --declaration --declarationDir dist/types && rollup -c && npm run bundle:xml -- --outDir dist",
+    "bundle:dev": "npm-run-all build:js bundle:iife \"bundle:xml -- --outDir build\" && node tools/bundle_scss/main.cjs --out build/o_spreadsheet.css --cssOnly",
```
LucasLefevre, line 23: bundling should also be done when running `npm run build`

package.json
LucasLefevre: Looks like https://www.npmjs.com/package/scss-bundle is more wildly used, although not updated since a long time as well

rrahir:
NOTE: before considering merging this, it will require odoo & spreadsheet-tools PR as well

tools/bundle_scss/scss.cjs
```diff
@@ -0,0 +1,38 @@
...
+  // scss files
+  const index = files.map((file) => `@use "${file}";`).join("\n");
+  const { css } = sass.compileString(index, {
+    loadPaths: ["."],
+  });
+  writeToFile(target, css + banner);
+
+  if (!cssOnly) {
+    const scssFile = target.replace(/\.css$/, ".scss");
+    let scss_contents = [];
```
LucasLefevre, line 25:
```suggestion
    let scssContents = [];
```

tools/bundle_scss/scss.cjs
```diff
@@ -0,0 +1,38 @@
...
+  const start = Date.now();
+  if (!target) {
+    throw new Error("Missing target!!");
+  }
+  const files = globSync("src/components/**/*.scss");
+  const banner = bundle.cssBanner();
+  // Build a fake index.scss file that imports all the components
+  // scss files
+  const index = files.map((file) => `@use "${file}";`).join("\n");
+  const { css } = sass.compileString(index, {
```
LucasLefevre, line 18: Instead of building a fake index file importing all others, I think we should build the `o_spreadsheet.scss` bundle file and then compile it to css.
That way, we guarantee both are equivalent.

Everytime we use `@use`, the `css` is valid, but the scss isn't-and it's silent.
And I actually think we'll want `@use` to import variables ?

It makes me reconsider using a lib that resolves all imports. (or not providing a `scss` bundle ?)

LucasLefevre: possible solutions are:
- no `@use`
- give css bundle to odoo
- declare all variables in a single file that we add as the first file in the bundle
- resolve `@use` ourselves
- give all scss files to odoo
   - all moved/flatten to a single directory (and rewrite `@use`
   - reproduce file hierarchy

LucasLefevre: option 1 seems to be the easiest short term. But in the long run, we'll want the scss in odoo too, to benefit from variable overwrites

package.json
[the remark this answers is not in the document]
pro-odoo: Now that we generate the `.css` file based on the `.scss` bundles, no need to have theses options anymore, we can generate both.

tools/bundle_scss/scss.cjs
pro-odoo: For the record:
Option 1: Works but we cannot override variables
Option 2: Works, but we will not allow to use `@use`, it will be much easier.
Option 3: odoo does not support `@use` and `@forward`, and forbid the use of `@import`, so not a possible solution

pro-odoo:
@LucasLefevre updated and green :)

LucasLefevre (approved):
robodoo r+

PR #5511, opened 2025-01-22
[FIX] chart tooltip: create app only on demand

pro-odoo:
Before this commit, a new App was created even when `renderToString` was never called. `renderToString` is for now only called to render a custom tooltip on a chart.

This caused a memory leak in the tests of Odoo, as a new App was created for each test, but never destroyed.

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


hokolomopo (commented):
@robodoo r+

PR #5512, opened 2025-01-22
[IMP] chart: allow multiples labels

anhe-odoo:
## Task Description

This PR aims to allow the usage of multiple ranges for a chart labels. This is a necessary step to have a full compatibility to the new treemap chart when switch from chart type, as well as a good way to "categorize" the labels with, for example, the first label ranges being the main category ("Art", "Sciences", ...) and the second label being the real label ("Math", "Chemistry", "Drawing", "Theatre", ...).

## Related Task

- Task: 4460827


src/helpers/figures/charts/runtime/chart_data_extractor.ts
```diff
@@ -44,11 +44,27 @@ import { recomputeZones } from "../../../recompute_zones";
...
+  if (labelRange && labelRange.length) {
+    for (const range of labelRange) {
+      const labelValues = getChartLabelValues(getters, dataSets, range);
+      const values = labelValues.formattedValues;
+      if (!labels.length) {
+        labels = values;
+      } else {
+        for (let i = 0; i < values.length; i++) {
+          if (values[i]) {
+            labels[i] += " \n " + values[i];
```
hokolomopo, line 60: Didn't you tell me that we do'nt need to do "hacky" \n and that chartJS accepts multiple labels ranges ? OR did I misunderstood ?

src/helpers/figures/charts/runtime/chart_data_extractor.ts
```diff
@@ -177,21 +213,45 @@ export function getLineChartData(
...
-    labels,
+    labels: applyLineBreakForLabels(labels),
     locale: getters.getLocale(),
     trendDataSetsValues,
     axisType,
   };
 }
 
+function applyLineBreakForLabels(labels: string[]): (string | string[])[] {
+  return labels.map((label) => (label.includes("\n") ? label.split("\n") : label));
```
hokolomopo, line 224: wait so you add newLines, and then you split  on the newLines to create an array ?  Why not create an array in the first place ? Or maybe I'm missing something

src/helpers/figures/charts/runtime/chart_data_extractor.ts
```diff
@@ -490,8 +566,11 @@ function isLinearChart(
   return !definition.labelsAsText && canBeLinearChart(labelRange, getters);
 }
 
-export function canChartParseLabels(labelRange: Range | undefined, getters: Getters): boolean {
-  return canBeDateChart(labelRange, getters) || canBeLinearChart(labelRange, getters);
+export function canChartParseLabels(labelRange: Range[] | undefined, getters: Getters): boolean {
+  if (!labelRange || labelRange.length !== 1) {
+    return false;
+  }
+  return canBeDateChart(labelRange[0], getters) || canBeLinearChart(labelRange[0], getters);
```
hokolomopo, line 573: I think if there's multiple label ranges, we should say that there cannot be a linear/date axis. Taking the first one feels very arbitrary.

hokolomopo (commented):
Your code seems fine, but I'm still not 100% convinced that we really want that functionally. That's a lot of changes. To discuss with the team.

Also every auxiliaryRange/labelRange variable should be changed to be plural. Which is even more changes :x

src/helpers/figures/charts/runtime/chart_data_extractor.ts
anhe-odoo: This is exactly what is done with the if above

src/helpers/figures/charts/runtime/chart_data_extractor.ts
anhe-odoo: Yeah, this is done to keep a "classical" string as label here, to be able to proceed the label shift and all the filtering stuff without too many changes. And then we resplit it in an array (your second comment)

src/helpers/figures/charts/runtime/chart_data_extractor.ts
hokolomopo: Oh yeah I cannot read sorry

PR #5513, opened 2025-01-23
[IMP] chart: add funnel chart

hokolomopo:
## Description

This commit adds the funnel chart type.

Technically it is implemented by adding a new type of chart to ChartJS that extends the bar chart. The only difference is that we draw trapezoids instead of rectangles. Otherwise, the runtime is the same as it would be for an horizontal bar chart with floating bars.

Task: [3619237](https://www.odoo.com/odoo/2328/tasks/3619237)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/types/index.ts
```diff
@@ -38,7 +38,7 @@ export * from "./workbook_data";
 
 declare global {
   interface Window {
-    Chart: typeof Chart;
+    Chart: typeof Chart & typeof Chart.Chart;
```
hokolomopo, line 41: window.Chart is both the class `Chart`, and all the exports of chartJS lib. A bit strange but 🤷

src/types/chart/funnel_chart.ts
```diff
@@ -0,0 +1,26 @@
...
+  readonly showValues?: boolean;
+  readonly funnelColors?: FunnelChartColors;
+}
+
+export type FunnelChartRuntime = {
+  chartJsConfig: ChartConfiguration;
+  background: Color;
+};
+
+export type FunnelChartColors = (Color | undefined)[];
```
LucasLefevre, line 26: in the command, I see `null` in the logs, maybe it's because it's been stringified

src/components/figures/chart/chartJs/chartjs_funnel_chart.ts
```diff
@@ -0,0 +1,121 @@
...
+  const startX = Math.min(x, base);
+  const startY = y - height / 2;
+
+  return {
+    x: startX + (width * 2) / 3,
+    y: startY + height / 2,
+  };
+};
+
+window.Chart.Tooltip.positioners.funnelTooltipPositioner = funnelTooltipPositioner;
```
LucasLefevre, line 104: how does chartjs knows it must use `funnelTooltipPositioner` ? is it a convention `<chart type>TooltipPositioner` ?

src/components/figures/chart/chartJs/chartjs_funnel_chart.ts
```diff
@@ -0,0 +1,121 @@
...
+
+export class FunnelChartController extends window.Chart?.BarController {
+  static id = "funnel";
+  static defaults = { ...window.Chart?.BarController.defaults, dataElementType: "funnel" };
+
+  updateElements(rects, start, count, mode) {
+    super.updateElements(rects, start, count, mode);
+    for (let i = start; i < start + count; i++) {
+      const rect = rects[i];
+      this.updateElement(rect, i, { nextElementWidth: rects[i + 1]?.width || 0 }, mode);
```
LucasLefevre, line 18: could you add a small comment to explain what this does? for people who never went into chartjs internals

src/helpers/figures/charts/runtime/chartjs_tooltip.ts
```diff
@@ -220,6 +220,28 @@ export function getGeoChartTooltip(
...
+    position: "funnelTooltipPositioner",
+    callbacks: {
+      title: () => "",
+      label: function (tooltipItem) {
+        const label = tooltipItem.label;
+        const yLabel = tooltipItem.parsed.x;
+
+        const axisId = tooltipItem.dataset.xAxisID;
+        const yLabelStr = formatChartDatasetValue(args.axisFormats, args.locale)(yLabel, axisId);
+        return label ? `${label}: ${yLabelStr}` : yLabelStr;
```
LucasLefevre, line 239: you might want to update this with the new `beforeLabel` approach

src/helpers/misc.ts
```diff
@@ -512,7 +512,7 @@ export function insertItemsAtIndex<T>(array: readonly T[], items: T[], index: nu
 
 export function replaceItemAtIndex<T>(array: readonly T[], newItem: T, index: number): T[] {
   const newArray = [...array];
-  newArray.splice(index, 1, newItem);
```
LucasLefevre, line 515: :+1: 
why should it be simple when we can make it more complex... :roll_eyes:

src/components/figures/chart/chartJs/chartjs_funnel_chart.ts
hokolomopo: image

We just give it as argument of the tooltip 😅

src/registries/chart_types.ts
```diff
@@ -398,4 +411,11 @@ chartSubtypeRegistry
     chartType: "geo",
     category: "misc",
     preview: "o-spreadsheet-ChartPreview.GEO_CHART",
+  })
+  .add("funnel", {
```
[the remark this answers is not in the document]
hokolomopo: Will be tacked here https://www.odoo.com/odoo/2328/tasks/4570622

LucasLefevre (approved):
robodoo r+

PR #5526, opened 2025-01-23
[FIX] spreadsheet: adapt CF formulas

fdamhaut:
Adapt CF formula on sheet change

Task: 4510426

## Description:

Task: [4510426](https://www.odoo.com/odoo/2328/tasks/4510426)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/core/conditional_format.ts
+++ b/src/plugins/core/conditional_format.ts
@@ -1,4 +1,4 @@
-import { compile } from "../../formulas/index";
+import { compile } from "../../formulas/compiler";
 import { deepEquals, isInside, recomputeZones, toUnboundedZone } from "../../helpers/index";
 import {
   AddConditionalFormatCommand,
@@ -58,37 +58,92 @@ export class ConditionalFormatPlugin
 
   readonly cfRules: { [sheet: string]: ConditionalFormatInternal[] } = {};
 
-  loopThroughRangesOfSheet(sheetId: UID, applyChange: ApplyRangeChange) {
-    for (const rule of this.cfRules[sheetId]) {
-      if (rule.rule.type === "DataBarRule" && rule.rule.rangeValues) {
-        const change = applyChange(rule.rule.rangeValues);
-        switch (change.changeType) {
-          case "REMOVE":
+  adaptCFFormulas(applyChange: ApplyRangeChange) {
+    for (const sheetId in this.cfRules) {
+      for (const rule of this.cfRules[sheetId]) {
+        if (rule.rule.type === "DataBarRule" && rule.rule.rangeValues) {
+          const change = applyChange(rule.rule.rangeValues);
+          switch (change.changeType) {
+            case "REMOVE":
+              this.history.update(
+                "cfRules",
+                sheetId,
+                this.cfRules[sheetId].indexOf(rule),
+                "rule",
+                //@ts-expect-error
+                "rangeValues",
+                undefined
+              );
+              break;
+            case "RESIZE":
+            case "MOVE":
+            case "CHANGE":
+              this.history.update(
+                "cfRules",
+                sheetId,
+                this.cfRules[sheetId].indexOf(rule),
+                "rule",
+                //@ts-expect-error
+                "rangeValues",
+                change.range
+              );
+              break;
+          }
+        } else if (rule.rule.type === "CellIsRule") {
+          for (let i = 0; i < rule.rule.values.length; i++) {
             this.history.update(
               "cfRules",
               sheetId,
               this.cfRules[sheetId].indexOf(rule),
               "rule",
               //@ts-expect-error
-              "rangeValues",
-              undefined
+              "values",
+              i,
+              this.getters.adaptFormulaStringDependencies(sheetId, rule.rule.values[i], applyChange)
             );
-            break;
-          case "RESIZE":
-          case "MOVE":
-          case "CHANGE":
-            this.history.update(
-              "cfRules",
-              sheetId,
-              this.cfRules[sheetId].indexOf(rule),
-              "rule",
-              //@ts-expect-error
-              "rangeValues",
-              change.range
-            );
-            break;
+          }
+        } else if (rule.rule.type === "IconSetRule") {
+          for (const inflectionPoint of ["lowerInflectionPoint", "upperInflectionPoint"] as const) {
+            if (rule.rule[inflectionPoint].type === "formula") {
+              this.history.update(
+                "cfRules",
+                sheetId,
+                this.cfRules[sheetId].indexOf(rule),
+                "rule",
+                //@ts-expect-error
+                inflectionPoint,
+                "value",
+                this.getters.adaptFormulaStringDependencies(
+                  sheetId,
+                  rule.rule[inflectionPoint].value,
+                  applyChange
+                )
+              );
+            }
+          }
+        } else if (rule.rule.type === "ColorScaleRule") {
+          for (const value of ["minimum", "maximum", "midpoint"] as const) {
+            const ruleValue = rule.rule[value];
+            if (ruleValue?.type === "formula" && ruleValue?.value) {
+              this.history.update(
+                "cfRules",
+                sheetId,
+                this.cfRules[sheetId].indexOf(rule),
+                "rule",
+                //@ts-expect-error
+                value,
+                "value",
+                this.getters.adaptFormulaStringDependencies(sheetId, ruleValue.value, applyChange)
+              );
+            }
+          }
         }
       }
+    }
+  }
+
+  adaptCFRanges(sheetId: UID, applyChange: ApplyRangeChange) {
+    for (const rule of this.cfRules[sheetId]) {
       for (const range of rule.ranges) {
         const change = applyChange(range);
         switch (change.changeType) {
@@ -126,13 +181,11 @@ export class ConditionalFormatPlugin
   }
 
   adaptRanges(applyChange: ApplyRangeChange, sheetId?: UID) {
-    if (sheetId) {
-      this.loopThroughRangesOfSheet(sheetId, applyChange);
-    } else {
-      for (const sheetId of Object.keys(this.cfRules)) {
-        this.loopThroughRangesOfSheet(sheetId, applyChange);
-      }
+    const sheetIds = sheetId ? [sheetId] : Object.keys(this.cfRules);
+    for (const sheetId of sheetIds) {
+      this.adaptCFRanges(sheetId, applyChange);
     }
+    this.adaptCFFormulas(applyChange);
   }
 
   // ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


tests/conditional_formatting/conditional_formatting_plugin.test.ts
```diff
@@ -679,6 +680,199 @@ describe("conditional format", () => {
...
+
+    test.each(CFs)("On column deletion %s", (name, cfrule) => {
+      const expected = formulasToCFs(
+        [
+          ["=A1", "=A5"],
+          ["=D1", "=D5"],
+          ["=A5", "=D5"],
+          ["=$D$5", "=$D$10"],
+          ["=D5*5", "=D5*10"],
+          ["=SUM(A1:D5)", "=SUM(A1:D5)"],
```
hokolomopo, line 802: The tests are kinda hard to read with the original formulas 100+ lines above the expected result. Maybe we could chanhe the structure with the test.each having a list of "(formula, expectedResult) " and the test body looping on every CF type. What do you think ?

tests/conditional_formatting/conditional_formatting_plugin.test.ts
```diff
@@ -679,6 +680,199 @@ describe("conditional format", () => {
...
+        rule,
+        ranges: ["A1"],
+      };
+    }
+
+    function formulasToCFs(formulas: string[][], cfrule: any): ConditionalFormat[] {
+      return formulas.map((fn) => cfrule(fn[0], fn[1])).map(ruleToCF);
+    }
+
+    const CFs = [
```
hokolomopo, line 760:
```suggestion
    const cfGenerators = [
```
nitpick. Or something similar

hokolomopo (commented):
boup boup :wave:

src/plugins/core/conditional_format.ts
```diff
@@ -88,6 +125,86 @@ export class ConditionalFormatPlugin
             );
             break;
         }
+      } else if (rule.rule.type === "CellIsRule") {
+        if (rule.rule.values.length > 0) {
```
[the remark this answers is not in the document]
fdamhaut: But I want to trigger it if there is 1 OR 2 elements.

tests/conditional_formatting/conditional_formatting_plugin.test.ts
```diff
@@ -679,6 +680,199 @@ describe("conditional format", () => {
...
+
+    function ruleToCF(rule: ConditionalFormatRule): ConditionalFormat {
+      return {
+        id: "1",
+        rule,
+        ranges: ["A1"],
+      };
+    }
+
+    function formulasToCFs(formulas: string[][], cfrule: any): ConditionalFormat[] {
```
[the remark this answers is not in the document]
fdamhaut: With how test naming is done (taking the 1st argument given AFAIK), if i want to have CFs be [string, fn] pair, I need this function to accept any (or have a new type + as <newtype> everywhere) and i'm not sure it's worth the additional verbosity for a test

src/plugins/core/conditional_format.ts
```diff
@@ -88,6 +88,57 @@ export class ConditionalFormatPlugin
             );
             break;
         }
+      } else if (rule.rule.type === "CellIsRule") {
```
hokolomopo, line 91: Doesn't work for formulas referencing a range on another sheet (`=SomeOtherSheet!A1`). You should loop on all the CFs, and not on only the CFs onf the sheet that has had the range change.

Same for data validation.

Data bars also kinda have the issue, but entering a sheet on another range is totally broken. That can probably be fixed [here](https://www.odoo.com/odoo/2328/tasks/4072640)

src/plugins/core/conditional_format.ts
```diff
@@ -58,37 +58,99 @@ export class ConditionalFormatPlugin
...
-              //@ts-expect-error
-              "rangeValues",
-              undefined
-            );
-            break;
-          case "RESIZE":
-          case "MOVE":
-          case "CHANGE":
+  loopThroughAllRanges(applyChange: ApplyRangeChange) {
+    for (const sheetId of Object.keys(this.cfRules)) {
```
LucasLefevre, line 62: Equivalent but without `Object.keys` (which allocates a new array, better avoid it if we can)
```suggestion
    for (const sheetId in this.cfRules) {
```

src/plugins/core/conditional_format.ts
```diff
@@ -58,37 +58,99 @@ export class ConditionalFormatPlugin
...
+                this.cfRules[sheetId].indexOf(rule),
+                "rule",
+                //@ts-expect-error
+                "rangeValues",
+                change.range
+              );
+              break;
+          }
+        } else if (rule.rule.type === "CellIsRule") {
+          for (let arg = 0; arg < rule.rule.values.length; arg++) {
```
LucasLefevre, line 93: `arg` ? Not sure what is means here ?
```suggestion
          for (let i = 0; i < rule.rule.values.length; i++) {
```

src/plugins/core/conditional_format.ts
```diff
@@ -58,37 +58,99 @@ export class ConditionalFormatPlugin
...
+                sheetId,
+                rule.rule.values[arg],
+                applyChange
+              )
             );
-            break;
+          }
+        } else if (rule.rule.type === "IconSetRule") {
+          for (const arg of ["lowerInflectionPoint", "upperInflectionPoint"]) {
+            if (arg in rule.rule && rule.rule[arg]?.type === "formula") {
```
LucasLefevre, line 111: according to the type, it's always defined
```suggestion
            if (rule.rule[arg].type === "formula") {
```

src/plugins/core/conditional_format.ts
```diff
@@ -58,37 +58,99 @@ export class ConditionalFormatPlugin
...
+              this.getters.adaptFormulaStringDependencies(
+                sheetId,
+                rule.rule.values[arg],
+                applyChange
+              )
             );
-            break;
+          }
+        } else if (rule.rule.type === "IconSetRule") {
+          for (const arg of ["lowerInflectionPoint", "upperInflectionPoint"]) {
```
LucasLefevre, line 110:
```suggestion
          for (const arg of ["lowerInflectionPoint", "upperInflectionPoint"] as const) {
```
otherwise `rule.rule[arg]`'s type can't be inferred and it's typed as `any`

Also, `arg` doesn't mean much
```suggestion
          for (const inflectionPoint of ["lowerInflectionPoint", "upperInflectionPoint"]) {
```

src/plugins/core/conditional_format.ts
```diff
@@ -58,37 +58,99 @@ export class ConditionalFormatPlugin
...
+                this.getters.adaptFormulaStringDependencies(
+                  sheetId,
+                  rule.rule[arg].value,
+                  applyChange
+                )
+              );
+            }
+          }
+        } else if (rule.rule.type === "ColorScaleRule") {
+          for (const arg of ["minimum", "maximum", "midpoint"]) {
```
LucasLefevre, line 129:
```suggestion
          for (const arg of ["minimum", "maximum", "midpoint"] as const) {
```
and it reveals that `rule.rule[arg].value,` can be `undefined` for min and max and boom

src/plugins/core/conditional_format.ts
```diff
@@ -126,13 +188,11 @@ export class ConditionalFormatPlugin
...
-    if (sheetId) {
+    const sheetIds = sheetId ? [sheetId] : Object.keys(this.cfRules);
+    for (const sheetId of sheetIds) {
       this.loopThroughRangesOfSheet(sheetId, applyChange);
-    } else {
-      for (const sheetId of Object.keys(this.cfRules)) {
-        this.loopThroughRangesOfSheet(sheetId, applyChange);
-      }
     }
+    this.loopThroughAllRanges(applyChange);
```
LucasLefevre, line 195: looking at this code is very strange. You don't understand why you loop first over sheets and then a second time through all ranges.
I think `loopThroughRangesOfSheet` and `loopThroughAllRanges` should be renamed to reflect what they do

tests/conditional_formatting/conditional_formatting_plugin.test.ts
```diff
@@ -679,6 +680,275 @@ describe("conditional format", () => {
     });
   });
 
+  describe("Grid Manipulation Formulas", () => {
+    function cellIsRuleEqual(f: string): ConditionalFormatRule {
```
LucasLefevre, line 684: could you rename `f` to `formula` everywhere ?

tests/conditional_formatting/conditional_formatting_plugin.test.ts
```diff
@@ -679,6 +680,275 @@ describe("conditional format", () => {
...
+          };
+        });
+    }
+
+    const cfGenerators = [
+      ["cellIsRuleEqual", cellIsRuleEqual],
+      ["cellIsRuleBetween", cellIsRuleBetween],
+      ["colorScaleRule", colorScaleRule],
+      ["IconSetRule", iconSetRule],
+    ];
```
LucasLefevre, line 761: you can add a better type for `cfrule`. And then you have to add `as const` otherwise Typescript infers `(string | () => ...)[]` instead of `[string, () => ...]`
```suggestion
    function formulasToCFs(formulas: string[], cfrule: (formula) => ConditionalFormatRule): ConditionalFormat[] {
      return formulas
        .map((f) => cfrule(f))
        .map((rule) => {
          return {
            id: "1",
            rule,
            ranges: ["A1"],
          };
        });
    }

    const cfGenerators = [
      ["cellIsRuleEqual", cellIsRuleEqual],
      ["cellIsRuleBetween", cellIsRuleBetween],
      ["colorScaleRule", colorScaleRule],
      ["IconSetRule", iconSetRule],
    ] as const;
```

src/plugins/core/conditional_format.ts
```diff
@@ -58,37 +58,95 @@ export class ConditionalFormatPlugin
...
+                "cfRules",
+                sheetId,
+                this.cfRules[sheetId].indexOf(rule),
+                "rule",
+                //@ts-expect-error
+                value,
+                "value",
+                this.getters.adaptFormulaStringDependencies(
+                  sheetId,
+                  rule.rule[value]!.value!,
```
LucasLefevre, line 137: coud you move `rule.rule[value]!.value!` to a variable before the if statement and use it here and in the if condition.
It'll allow to get rid of `!`

tests/conditional_formatting/conditional_formatting_plugin.test.ts
```diff
@@ -679,6 +680,278 @@ describe("conditional format", () => {
...
+          };
+        });
+    }
+
+    const cfGenerators = [
+      ["cellIsRuleEqual", cellIsRuleEqual],
+      ["cellIsRuleBetween", cellIsRuleBetween],
+      ["colorScaleRule", colorScaleRule],
+      ["IconSetRule", iconSetRule],
+    ] as [string, (s: string) => ConditionalFormatRule][];
```
LucasLefevre, line 764:
```suggestion
    ] as const;
```
Slightly easier and shorter and works as well

LucasLefevre (commented):
only details

src/plugins/core/conditional_format.ts
fdamhaut: value is only present based on the type, having the const would then mean i have to split the if condition in 2.
I feel like ! is clearer than two nested ifs with a const in between them

LucasLefevre:
robodoo r+

PR #5527, opened 2025-01-23
[FIX] formulas: binary search returning incorrect index for multiple exact matches

laa-odoo:
Ensure binary search in exact match mode returns
the first occurrence when multiple matches exist.

Task: [4328300](https://www.odoo.com/odoo/2328/tasks/4328300)


LucasLefevre (commented):
robodoo r+

PR #5535, opened 2025-01-24
[FIX] ui_sheet: batch resize commands in `AUTORESIZE_ROWS`

hokolomopo:
## Description

When auto-resizing rows, we would dispatch a `RESIZE_COLUMNS_ROWS` for each row in the selection. But we could simply be dispatching a single `RESIZE_COLUMNS_ROWS` command all the rows, since the size would be the same for all the rows (`null` to mark that they can resize dynamically). This lighten up the network load when auto-resizing multiple rows.

Note that the same cannot be done for columns, since the size of auto-resized columns can be different.

Task: [4504918](https://www.odoo.com/odoo/2328/tasks/4504918)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui/ui_sheet.ts
+++ b/src/plugins/ui/ui_sheet.ts
@@ -61,14 +61,12 @@ export class SheetUIPlugin extends UIPlugin {
         }
         break;
       case "AUTORESIZE_ROWS":
-        for (let row of cmd.rows) {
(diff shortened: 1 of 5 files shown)


LucasLefevre (commented):
robodoo r+

PR #5558, opened 2025-01-28
[FIX] auto_complete: search pivot value with quotes

LucasLefevre:
## Description:

In the demo sheet, When typing =PIVOT.VALUE(1,"Expected Revenue","Stage", "Ne), we are not getting any suggestions.

The reason is the double quotes, which is not part of the value fuzzy search key and hence the value is filtered out.

Task: [4061068](https://www.odoo.com/odoo/2328/tasks/4061068)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


rrahir:
@robodoo r+

PR #5572, opened 2025-01-29
[FIX] Borders: Fix border continuity

rrahir:
The code that would compare the common sides of two borders was not adapted when we improved the border structure (from a single value to an object with color and style properties). It would make a reference comparison which would sometimes fail, sometimes succeed depending on how the borders were added in the first place.

Task: 4523890

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/tests/borders/border_plugin.test.ts
+++ b/tests/borders/border_plugin.test.ts
@@ -741,6 +741,66 @@ describe("Grid manipulation", () => {
   });
 });
 
+describe("Border continuity", () => {
+  const border = {
+    top: DEFAULT_BORDER_DESC,
+    left: DEFAULT_BORDER_DESC,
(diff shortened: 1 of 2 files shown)


tests/borders/border_plugin.test.ts
```diff
@@ -741,6 +741,66 @@ describe("Grid manipulation", () => {
   });
 });
 
+describe("Border continuity", () => {
+  const border = {
```
hokolomopo, line 745:
```suggestion
  const allBorders = {
```
nitpick

LucasLefevre (approved):
robodoo r+

PR #5573, opened 2025-01-29
[FIX] locale: wrong parsing with "." thousand separator

hokolomopo:
## Description

The parsing of strings to number was wrong if the thousand separator was a ".". We would first replace the locale's thousand separator with JS's thousand separator ("."), and then remove all thousand separators in the string. Thus removing the actual decimal separator.

Task: [4525746](https://www.odoo.com/odoo/2328/tasks/4525746)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


pro-odoo (approved):
robodoo r+

PR #5589, opened 2025-01-30
[IMP] OT: add registry for operation values transform

fdamhaut:
## Description:

This task add an adaptRange-like feature to command operational transform.
The goal is to simplify the adaptability of formulas and range in OT and avoid the 
N² function required in the current OT for range adaptation using the same ApplyRangeChange
method.

Task: [4319840](https://www.odoo.com/odoo/2328/tasks/4319840)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/collaborative/ot/ot.ts
+++ b/src/collaborative/ot/ot.ts
@@ -6,6 +6,7 @@ import {
   moveHeaderIndexesOnHeaderAddition,
   moveHeaderIndexesOnHeaderDeletion,
 } from "../../helpers/index";
+import { deepCopy } from "../../helpers/misc";
 import { otRegistry } from "../../registries/ot_registry";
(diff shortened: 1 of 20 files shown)


src/collaborative/ot/ot.ts
```diff
@@ -61,12 +65,30 @@ const transformations: {
  */
 export function transform(
   toTransform: CoreCommand,
-  executed: CoreCommand
+  executed: CoreCommand,
+  coreGetters: CoreGetters
```
LucasLefevre, line 69: transformation cannot depend on getters because they should not depend on the current state of the spreadsheet. Different users can be in different states (before everyone have received every revision). So different states would mean different transformation, and ultimately, all clients would not converge to the same state.

PR #5603, opened 2025-01-31
[FIX] pivot: display column values in sorted pivot

LucasLefevre:
## Description:

Steps to reproduce:

- Insert an odoo pivot with row groups, but no column group.
- sort the measure on one column => the formula =PIVOT(1,,FALSE) doesn't show any data, even though it should show the column

Task: [4536075](https://www.odoo.com/odoo/2328/tasks/4536075)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/tests/pivots/spreadsheet_pivot/spreadsheet_pivot.test.ts
+++ b/tests/pivots/spreadsheet_pivot/spreadsheet_pivot.test.ts
@@ -660,6 +660,33 @@ describe("Spreadsheet Pivot", () => {
     );
   });
 
+  test("sorted PIVOT include_total=FALSE with no column groups includes column data", () => {
+    // prettier-ignore
+    const grid = {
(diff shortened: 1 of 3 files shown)


tests/pivots/spreadsheet_pivot/spreadsheet_pivot.test.ts
```diff
@@ -660,6 +660,33 @@ describe("Spreadsheet Pivot", () => {
     );
   });
 
+  test("sorted PIVOT include_total=FALSE with no column groups includes column data", () => {
```
LucasLefevre, line 663: note: this test was passing before the fix thanks to the combination of both "off-by-one" errors

VincentSchippefilt:
robodoo r+

PR #5606, opened 2025-02-03
[FIX] formulas: accept simple values on search range for lookup formulas

laa-odoo:
Accept simple values on search range and return range for all lookup formulas.

Task: [4543381](https://www.odoo.com/odoo/2328/tasks/4543381)

### DIFF
--- a/src/functions/module_lookup.ts
+++ b/src/functions/module_lookup.ts
@@ -2,6 +2,8 @@ import { getCanonicalSheetName, toXC, toZone } from "../helpers/index";
 import { _t } from "../translation";
 import {
   AddFunctionDescription,
+  Arg,
+  ArgValue,
   CellValue,
   Matrix,
   Maybe,
@@ -18,6 +20,7 @@ import {
   linearSearch,
   strictToInteger,
   toBoolean,
+  toMatrix,
   toNumber,
   toString,
 } from "./helpers";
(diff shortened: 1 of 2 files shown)


src/functions/module_lookup.ts
```diff
@@ -153,7 +156,7 @@ export const HLOOKUP = {
   args: [
     arg("search_key (any)", _t("The value to search for. For example, 42, 'Cats', or I24.")),
     arg(
-      "range (range)",
+      "range (any, range)",
```
LucasLefevre, line 159: Do they also accept error values ?
It doesn't matter in 17.0, but it does in the version in which we did the "early return" optimization
```suggestion
      "range (string,number,boolean,range)",
```

tests/functions/module_lookup.test.ts
```diff
@@ -764,6 +784,15 @@ describe("VLOOKUP formula", () => {
     setCellContent(model, "A1", "=VLOOKUP(5, B1, 1)");
     expect(getCellError(model, "A1")).toBe("Did not find value '5' in VLOOKUP evaluation.");
   });
+
+  test("Accept simple values on the search range or the return range", () => {
+    const grid = evaluateGrid({
+      A1: "=HLOOKUP(42, 42, 1)",
```
rrahir, line 790:
```suggestion
      A1: "=VLOOKUP(42, 42, 1)",
```
same below :)

rrahir:
@robodoo r+
@laa-odoo I corrected the test

laa-odoo:
> @robodoo r+ @laa-odoo I corrected the test

thx

PR #5607, opened 2025-02-03
[FIX] composer: long composer asisstant breaks the layout

hokolomopo:
### [FIX] composer: broken layout for long list suggestions

If the composer TextProvider displayed a very long list of possible
values, it broke the spreadsheet layout.

This commit changes the composer assistant to a Popover component,
which have build-in code to be limited by the spreadsheet size.

The assistant was also positioned relative to the `o-composer` element,
where it should have been positioned relative to the parent element,
because it's the one that has the borders/padding.

### [FIX] composer: wrong size for autocomplete dropdown

In the grid composer we pass a `rect` props to the composer to make
the formula assistant have the same size as the underlying cell.

This makes:
1) the FunctionDescriptionProvider ugly for large cells
2) the TextValueProvider buggy for large cells, as its CSS define a
`max-width` that conflicts with the `min-width` that the composer sets.

This commit caps the `min-width` of the composer to `ASSISTANT_WIDTH`.

### [FIX] popover: scroll lost on re-render

The commit c49ed07780 tried to fix the positioning of the `Popover`
when it was re-rendered with possibly different props. But setting the
`max-height` of the popover to undefined, even temporarily, was causing
the scroll to be lost.

This reverts the changes of c49ed07780, and adds a `t-key` to the
`Popover` in the `Menu` to fix the original issue by forcing a
total re-render of the `Popover` when the user opens a different menu.

### [MOV] test: move some composer tests in correct files

Some tests in the `composer_component.test.ts` were mounting a
spreadsheet after a ComposerWrapper was already mounted in the
beforeEach.

It caused issue with the following commit that adds a Portal target in
the ComposerWrapper. This led to two Portal targets in the document and
a test failing.

This commit moves these tests to the `composer_integration_component.test.ts`
file where they belong.

Task: [4456224](https://www.odoo.com/odoo/2328/tasks/4456224)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/composer/composer/composer.ts
+++ b/src/components/composer/composer/composer.ts
@@ -172,7 +172,8 @@ export class Composer extends Component<ComposerProps, SpreadsheetChildEnv> {
     const composerRect = this.composerRef.el!.getBoundingClientRect();
     const assistantStyle: CSSProperties = {};
 
-    assistantStyle["min-width"] = `${this.props.rect?.width || ASSISTANT_WIDTH}px`;
+    const minWidth = Math.min(this.props.rect?.width || Infinity, ASSISTANT_WIDTH);
+    assistantStyle["min-width"] = `${minWidth}px`;
     if (this.autoCompleteState.type === "function") {
       assistantStyle.width = `${ASSISTANT_WIDTH}px`;
     }
--- a/src/components/composer/composer/composer.ts
+++ b/src/components/composer/composer/composer.ts
@@ -2,21 +2,20 @@ import { Component, onMounted, useEffect, useRef, useState } from "@odoo/owl";
 import { COMPOSER_ASSISTANT_COLOR, DEFAULT_FONT, NEWLINE } from "../../../constants";
 import { EnrichedToken } from "../../../formulas/index";
 import { functionRegistry } from "../../../functions/index";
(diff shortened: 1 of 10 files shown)


src/components/composer/composer/composer.ts
```diff
@@ -811,15 +781,22 @@ export class Composer extends Component<ComposerProps, SpreadsheetChildEnv> {
     this.autoCompleteState.values = values.map((value) => ({ text: value, description: "" }));
     this.autoCompleteState.getHtmlContent = (value) => [{ value }];
   }
+
+  get popoverProps(): PopoverProps {
+    return {
+      anchorRect: this.props.getContainerRect?.() || getBoundingRectAsPOJO(this.composerRef.el!),
```
LucasLefevre, line 787: do we really need the props `props.getContainerRect` ? Can't we always use the rect of `this.composerRef.el` ?
It's slightly "décallé" but it's not choquant. It can be seen as a design choice

hokolomopo: I think it's a bit uglier but sure

rrahir:
@robodoo rebase-ff

LucasLefevre (approved):
robodoo r+ rebase-ff

PR #5620, opened 2025-02-04
[FIX] chart: label header prevents creating linear/time charts

hokolomopo:
## Description

The helpers to check if the labels of a chart can be used to create a linear/time axis weren't taking into account that the first label should sometimes be ignored because of `chart.dataSetsHaveTitle`.

It was impossible to create a linear/time chart when the first label was a (non-number) title.

Also the helper `getChartLabelFormat` only supported columns as label ranges, where the label range can be rows as well.

Task: [4543496](https://www.odoo.com/odoo/2328/tasks/4543496)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/figures/charts/chart_factory.ts
+++ b/src/helpers/figures/charts/chart_factory.ts
@@ -13,6 +13,7 @@ import {
   UID,
   Zone,
 } from "../../../types";
+import { LineChartDefinition } from "../../../types/chart";
 import {
   ChartCreationContext,
   ChartDefinition,
@@ -23,7 +24,7 @@ import { CoreGetters, Getters } from "../../../types/getters";
 import { Validator } from "../../../types/validator";
 import { getZoneArea, zoneToXc } from "../../zones";
 import { AbstractChart } from "./abstract_chart";
-import { canChartParseLabels } from "./line_chart";
+import { LineChart, canChartParseLabels } from "./line_chart";
 
 /**
  * Create a function used to create a Chart based on the definition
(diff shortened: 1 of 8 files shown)


src/helpers/figures/charts/chart_factory.ts
```diff
@@ -168,21 +169,22 @@ export function getSmartChartDefinition(zone: Zone, getters: Getters): ChartDefi
...
+    stacked: false,
+    aggregated: false,
+    cumulative: false,
+    labelRange: labelRangeXc,
+    type: "line",
+    dataSetsHaveTitle,
+    verticalAxisPosition: "left",
+    legendPosition: newLegendPos,
+  };
+  const chart = new LineChart(lineChartDefinition, sheetId, getters);
```
anhe-odoo, line 185: I personnaly find it useless to create a LineChart here just so that we have the correct object to pass to `canChartPaseLabels`, while this function will only use labelRange, dataSets[0] and dataSetsHaveTitle from the full chart.
Maybe we can create a temporary object if you really want to have only one argument, but I would go with 4 distincs arguments (`labelRange`, `dataSet = dataSets[0]`, `dataSetsHaveTitle` and `getters`)

anhe-odoo: Moreover, the only reason you need these new arguments is for the "shouldRemoveFirstLabel" function calls. So maybe we can call it here before calling canChartParseLabels, and then call the later with `labelRange`, `removeFirstLabel` and `getters` ?

tests/figures/chart/charts_component.test.ts
```diff
@@ -1056,7 +1056,12 @@ describe("charts", () => {
 
     test("Side panel correctly reacts to has_header checkbox check/uncheck (with only one point)", async () => {
       createTestChart("basicChart");
-      updateChart(model, chartId, { type: "line", labelRange: "C2", dataSets: ["A1"] });
+      updateChart(model, chartId, {
+        type: "line",
+        labelRange: "C2",
+        dataSets: ["A1"],
+        dataSetsHaveTitle: false,
```
anhe-odoo, line 1063: Why do we need this now ? Was it a mistake not to have this property set here before ?

src/helpers/figures/charts/chart_factory.ts
hokolomopo: We don't have actual `DataSet` objects here, only zones/XCs. Se we'd have to call `createDataSets` first. And I'm not a big fan of calling 3 helpers just to get the parameters to call another helpers because 1) it's annoying and 2) there's no guarantee that we call the same helpers/in the same way that the actual chart does. So it's way easier for errors to slip by. I'd much prefer just create an actual LineChart.

tests/figures/chart/charts_component.test.ts
hokolomopo: Yeah we had a dataset of one cell with dataSetsHaveTitle = true ... there's no dataset there ...

rrahir:
@robodoo r+

hokolomopo:
@robodoo rebase-ff

PR #5633, opened 2025-02-04
[IMP] perfs: improve performance of `includesAll`

hokolomopo:
## Description

The helper `includesAll` is used to check if all the visible rows are selected when opening the row context menu. But it's implementation was very naive (O(n^2)) and was causing performance issues when the number of rows was high.

This commit uses a set in the implementation of `includesAll` to make the check O(n). We also make an assumption that the two arrays passed to `includesAll` do not contain duplicates, which make it basically free when there is less rows selected than there are visible rows.

Benchmark on spreadsheet with 122K rows:
- Before: ~25 seconds taken by `includesAll`
- After: ~110ms taken by `includesAll`

Task: [4548168](https://www.odoo.com/odoo/2328/tasks/4548168)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/misc.ts
+++ b/src/helpers/misc.ts
@@ -482,9 +482,17 @@ export function deepEquals(o1: any, o2: any): boolean {
   return true;
 }
 
-/** Check if the given array contains all the values of the other array. */
+/**
+ * Check if the given array contains all the values of the other array.
+ * It makes the assumption that both array do not contain duplicates.
+ */
 export function includesAll<T>(arr: T[], values: T[]): boolean {
-  return values.every((value) => arr.includes(value));
+  if (arr.length < values.length) {
+    return false;
+  }
+
+  const set = new Set(arr);
+  return values.every((value) => set.has(value));
 }
 
 /**


src/helpers/misc.ts
```diff
@@ -482,9 +482,17 @@ export function deepEquals(o1: any, o2: any): boolean {
...
+ * Check if the given array contains all the values of the other array.
+ * It makes the assumption that both array do not contain duplicates.
+ */
 export function includesAll<T>(arr: T[], values: T[]): boolean {
-  return values.every((value) => arr.includes(value));
+  if (arr.length < values.length) {
+    return false;
+  }
+
+  const set = new Set(arr);
```
rrahir, line 494: soon in your browser:  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/intersection

rrahir: actually, we can benefit from it if we bump typescript package

hokolomopo: It's slower !

pro-odoo (approved):
robodoo r+

PR #5641, opened 2025-02-05
[IMP] css: remove global `content-box` directive

hokolomopo:
## Description

We have a global css directive that sets `box-sizing: content-box` everywhere. This is a bit stupid, as it override bootstrap default `border-box` directive.

And it also make using `width/height` directive less intuitive, as the size you set does not includes border/padding.

We had 32 occurrences of `box-sizing: border-box` in our codebase, to override the global directive.

Task: [4550753](https://www.odoo.com/odoo/2328/tasks/4550753)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/spreadsheet/spreadsheet.ts
+++ b/src/components/spreadsheet/spreadsheet.ts
@@ -33,7 +33,6 @@ import {
   PRIMARY_BUTTON_ACTIVE_BG,
   PRIMARY_BUTTON_BG,
   PRIMARY_BUTTON_HOVER_BG,
-  SCROLLBAR_WIDTH,
   SEPARATOR_COLOR,
   TEXT_BODY,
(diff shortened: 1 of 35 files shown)


src/components/spreadsheet/spreadsheet.ts
```diff
@@ -225,18 +222,9 @@ css/* scss */ `
     }
 
     > canvas {
+      box-sizing: content-box;
       border-bottom: 1px solid #e2e3e3;
     }
-    .o-scrollbar {
```
hokolomopo, line 230: was duplicated in scrollbar.ts

LucasLefevre (approved):
robodoo r+

PR #5644, opened 2025-02-05
[FIX] formula : fix SORT formula error

anhe-odoo:
## Description

This commit aims to fix the error message of the SORT formula when one of the repeatable parameter is undefined. This issue was using "%d" instead of "%s" in _t

Task: 0

### DIFF
--- a/src/functions/module_filter.ts
+++ b/src/functions/module_filter.ts
@@ -27,8 +27,8 @@ function sortMatrix(
     assert(
       () => value !== undefined,
       _t(
-        "Value for parameter %d is missing, while the function [[FUNCTION_NAME]] expect a number or a range.",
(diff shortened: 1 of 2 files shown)


src/functions/module_filter.ts
```diff
@@ -27,8 +27,8 @@ function sortMatrix(
     assert(
       () => value !== undefined,
       _t(
-        "Value for parameter %d is missing, while the function [[FUNCTION_NAME]] expect a number or a range.",
-        i + 1
+        "Value for parameter %s is missing, while the function [[FUNCTION_NAME]] expect a number or a range.",
+        (i + 1).toString()
```
hokolomopo, line 31: the toString isn't needed no ?

PR #5647, opened 2025-02-06
[IMP] TopBar: Add responsiveness

rrahir:
## Description:

Task: [4562073](https://www.odoo.com/odoo/2328/tasks/4562073)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/remove_duplicates/remove_duplicates.ts
+++ b/src/components/side_panel/remove_duplicates/remove_duplicates.ts
@@ -1,4 +1,4 @@
-import { Component, onWillUpdateProps, useState } from "@odoo/owl";
+import { Component, onMounted, onWillUpdateProps, useState } from "@odoo/owl";
 import { numberToLetters, zoneToDimension } from "../../../helpers";
 import { _t } from "../../../translation";
 import { HeaderIndex, SpreadsheetChildEnv } from "../../../types/index";
@@ -33,6 +33,7 @@ export class RemoveDuplicatesPanel extends Component<Props, SpreadsheetChildEnv>
   });
 
   setup() {
+    onMounted(() => this.updateColumns());
     onWillUpdateProps(() => this.updateColumns());
   }
 
--- a/src/components/top_bar/number_formats_tool/number_formats_tool.ts
+++ b/src/components/top_bar/number_formats_tool/number_formats_tool.ts
@@ -1,0 +1,50 @@
+import { Component, useRef, useState } from "@odoo/owl";
+import { Action, createAction } from "../../../actions/action";
+import { formatNumberMenuItemSpec } from "../../../registries";
+import { DOMCoordinates, SpreadsheetChildEnv } from "../../../types";
+import { ActionButton } from "../../action_button/action_button";
+import { ToolBarDropdownStore, useToolBarDropdownStore } from "../../helpers/top_bar_tool_hook";
+import { Menu } from "../../menu/menu";
+
+interface Props {
+  class: string;
+}
+
+interface State {
+  menuItems: Action[];
+  position: DOMCoordinates;
+}
+
+export class NumberFormatsTool extends Component<Props, SpreadsheetChildEnv> {
+  static template = "o-spreadsheet-NumberFormatsTool";
+  static components = { Menu, ActionButton };
+  static props = { class: String };
+  formatNumberMenuItemSpec = formatNumberMenuItemSpec;
+  topBarToolStore!: ToolBarDropdownStore;
+
+  buttonRef = useRef("buttonRef");
+  state: State = useState({
+    position: { x: 0, y: 0 },
+    menuItems: [],
+  });
+
+  setup() {
+    this.topBarToolStore = useToolBarDropdownStore();
+  }
+
+  toggleMenu() {
+    if (this.isActive) {
+      this.topBarToolStore.closeDropdowns();
+    } else {
+      const menu = createAction(this.formatNumberMenuItemSpec);
+      this.state.menuItems = menu.children(this.env).sort((a, b) => a.sequence - b.sequence);
+      const { x, y, height } = this.buttonRef.el!.getBoundingClientRect();
+      this.state.position = { x, y: y + height };
+      this.topBarToolStore.openDropdown();
+    }
+  }
+
+  get isActive() {
+    return this.topBarToolStore.isActive;
+  }
+}
--- a/src/components/top_bar/top_bar.scss
+++ b/src/components/top_bar/top_bar.scss
@@ -1,0 +1,12 @@
+@media (max-width: 900px) {
+  div.o-topbar-responsive {
+    flex-direction: column !important;
+  }
+}
+
+@media (max-width: 768px) {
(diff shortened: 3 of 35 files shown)


src/components/top_bar/top_bar.ts
```diff
@@ -291,4 +289,228 @@ export class TopBar extends Component<Props, SpreadsheetChildEnv> {
...
+      ],
+      class: "o-hoverable-button o-menu-item-button o-toolbar-button",
+      childClass: "o-hoverable-button",
+    },
+    sequence: 3,
+  })
+  .add("misc")
+  .addChild("misc", {
+    component: TableDropdownButton,
+    props: { class: "toolbar-button" },
```
pro-odoo, line 506: `o-toolbar-button` ? 
I'm pretty sure that it's the only element that has not the `o-hoverable-button`, is it expected ?
If it's not intended, I would move these two classes (`o-hoverable-button` and `o-toolbar-button`) inside the template, and use a div or a span with these classes in the t-foreach ?

src/components/top_bar/top_bar.ts
```diff
@@ -164,31 +152,59 @@ export class TopBar extends Component<Props, SpreadsheetChildEnv> {
     PaintFormatButton,
     BorderEditorWidget,
     TableDropdownButton,
+    Popover,
```
pro-odoo, line 155: Remove the unused one (ColorPickerWidget, ColorPicker, FontSizeEditor, ActionButton, PaintFormatButton, BorderEditorWidget and TableDropdownButton I think)

src/components/top_bar/top_bar.ts
```diff
@@ -102,52 +124,18 @@ css/* scss */ `
...
-              }
-            }
-          }
-        }
       }
     }
   }
 `;
 
+export const topbarToolsWidthThresholds = [750, 710, 560, 480];
```
pro-odoo, line 134: What do you think of: 
`export const topbarToolsWidthThresholds = [715, 675, 545, 480];`

src/components/top_bar/top_bar_tool_store.ts
```diff
@@ -0,0 +1,18 @@
+import { Component } from "@odoo/owl";
+
+export class TopBarToolStore {
```
pro-odoo, line 3:
```suggestion
/**
 * This store is used to manage the dropdown that is currently opened after clicking an item on the toolbar. It can only have one displayed at a time.
 *
 */
export class TopBarToolStore {
```

pro-odoo (commented):
Nice one 👍 It's easier to read the topbar template now, and we can add items from odoo !
I have one concern, if we add an item from odoo, it will break the display with the hardcoded threshold :/ 
Didn't read all the tests, my comments are mainly nitpick.
I added a screenshot in the pad.

rrahir:
> I have one concern, if we add an item from odoo, it will break the display with the hardcoded threshold :/

we could alternatively decide to set the section size in the registry itself, with a sequence. Otherwise we could decide to apply a rule by counting the number of items in each section and use it to define the thresholds. I don't know if we need it right now though

src/components/side_panel/remove_duplicates/remove_duplicates.ts
```diff
@@ -33,6 +33,7 @@ export class RemoveDuplicatesPanel extends Component<Props, SpreadsheetChildEnv>
   });
 
   setup() {
+    onMounted(() => this.updateColumns());
```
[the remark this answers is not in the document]
rrahir: it works be chance in stable, because if we end up in this menu by opening the sidepanel, it triggers a global render and therefore the state is update. But it's also already broken because if you open a chart sidepanel and then switch tot the remove duplicates, you will see that the content is wrong and contains and error because you never properly populated the columns.

I don't remember the reason why it's highlighted in my PR, probably a global render that is skipped following a change in the top bar behaviour?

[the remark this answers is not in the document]
rrahir: Actully, I remember now.
Following my ref, I actually trigger a full rendering, which did not occur in the tests for the reason metioned above (in the test ,the grid resizing following the `openSidePanel` is not triggered)
The test that was failing was this one
`if no columns selected --> display error message and disable`

In this test, you are checking that an error message appears if you click on a given button but actually, the error message is already present before you click on it. we just never tested it properly.

src/components/top_bar/top_bar.ts
rrahir: it's been a battle already to use it because some of those components alreay contain some classes that make it quite hard to work. The hoverable button class was designed to interact with o-menu-item, and that class is present in some of the sub component (like ActionButton)
I've ben trying for an hour an can't make it work don't know if it's worth it

src/components/top_bar/top_bar.scss
```diff
@@ -3,3 +3,10 @@
     flex-direction: column !important;
   }
 }
+
+@media (max-width: 768px) {
```
rrahir, line 7: this is required to unf_k the text in a small screen.

src/components/top_bar/number_formats_tool/number_formats_tool.ts
```diff
@@ -0,0 +1,52 @@
+import { Component, useRef, useState } from "@odoo/owl";
+import { Action, createAction } from "../../../actions/action";
+import { formatNumberMenuItemSpec } from "../../../registries";
+import { DOMCoordinates, SpreadsheetChildEnv } from "../../../types";
+import { ActionButton } from "../../action_button/action_button";
+import { useToolBarDropdownStore } from "../../helpers/top_bar_tool_hook";
+import { Menu } from "../../menu/menu";
+
+formatNumberMenuItemSpec;
```
LucasLefevre, line 9: :eyes:

src/components/top_bar/top_bar.ts
```diff
@@ -118,7 +116,7 @@ css/* scss */ `
 /**
  * find the first index bigger than the current width and retract it from  the registry entries
  */
-export const topbarToolsWidthThresholds = [750, 710, 560, 480];
+// export const topbarToolsWidthThresholds = [750, 600]; //[750, 710, 560, 480];
```
pro-odoo, line 119: remove

src/components/top_bar/top_bar.ts
```diff
@@ -164,20 +167,50 @@ export class TopBar extends Component<Props, SpreadsheetChildEnv> {
...
     onMounted(() => {
-      resizeObserver.observe(this.toolbarRef.el!);
+      resizeObserver.observe(this.toolBarContainerRef.el!);
     });
     onWillUnmount(() => {
       resizeObserver.disconnect();
     });
+    // useEffect(() => {
+    //   this.setVisiblityToolsGroups();
+    // });
```
pro-odoo, line 177: 👀

pro-odoo:
robodoo r-

hokolomopo:
@robodoo r+

PR #5656, opened 2025-02-06
poc cache for each sub formula

laa-odoo:
Task: [4618834](https://www.odoo.com/odoo/2328/tasks/4618834)

### DIFF
--- a/src/functions/index.ts
+++ b/src/functions/index.ts
@@ -281,10 +281,22 @@ function createComputeFunction(
     let subCache = cache.get(functionName);
 
     for (const arg of args) {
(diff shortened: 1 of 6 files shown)


src/functions/index.ts
```diff
@@ -55,6 +55,7 @@ const categories: Category[] = [
 ];
 
 const functionNameRegex = /^[A-Z0-9\_\.]+$/;
+const resultNameKey = "ouuuuuuWAAAAAAAAAAAAAAAAAAAAAH";
```
rrahir, line 58: 👌

PR #5660, opened 2025-02-10
[PERF] import/export XLSX

laa-odoo:
Task: [4563258](https://www.odoo.com/odoo/2328/tasks/4563258)

### DIFF
--- a/src/helpers/data_normalization.ts
+++ b/src/helpers/data_normalization.ts
@@ -53,7 +53,7 @@ export function* iterateItemIdsPositions(sheetId: UID, itemIdsByZones: Record<st
   }
 }
 
-function getCanonicalRepresentation(item: any): string {
+export function getCanonicalRepresentation(item: any): string {
   if (item === null) return "null";
   if (item === undefined) return "undefined";
   if (typeof item !== "object") return String(item);
--- a/src/helpers/data_normalization.ts
+++ b/src/helpers/data_normalization.ts
@@ -1,25 +1,32 @@
 import { Position, UID } from "../types";
-import { deepEquals, largeMax } from "./misc";
 import { recomputeZones } from "./recompute_zones";
 import { positionToZone, toZone, zoneToXc } from "./zones";
+
+type ReverseLookup = Map<string, number>;
+type ItemsDic<T> = { [id: number]: T };
+const globalReverseLookup = new Map<ItemsDic<any>, ReverseLookup>();
 
 /**
  * Get the id of the given item (its key in the given dictionary).
  * If the given item does not exist in the dictionary, it creates one with a new id.
  */
-export function getItemId<T>(item: T, itemsDic: { [id: number]: T }) {
-  for (const key in itemsDic) {
-    if (deepEquals(itemsDic[key], item)) {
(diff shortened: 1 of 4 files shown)


src/helpers/data_normalization.ts
```diff
@@ -45,3 +59,29 @@ export function* iterateItemIdsPositions(sheetId: UID, itemIdsByZones: Record<st
...
+
+  if (Array.isArray(item)) {
+    const len = item.length;
+    let result = "[";
+    for (let i = 0; i < len; i++) {
+      if (i > 0) result += ",";
+      result += getCanonicalRepresentation(item[i]);
+    }
+    return result + "]";
+  }
```
anhe-odoo, line 76: Maybe we can do something like `item.map(getCanonicalRepresentation).join(',')`, but I'm not sure about the performance

anhe-odoo: (as you want)

src/helpers/data_normalization.ts
```diff
@@ -45,3 +59,29 @@ export function* iterateItemIdsPositions(sheetId: UID, itemIdsByZones: Record<st
     }
   }
 }
+
+function getCanonicalRepresentation(item: any): string {
```
anhe-odoo: Maybe we can use number as the canonical representation of number and have number | string as the reverseLookup map key :)

src/helpers/data_normalization.ts
```diff
@@ -1,25 +1,39 @@
...
   const maxId = ids.length === 0 ? 0 : largeMax(ids.map((id) => parseInt(id, 10)));
+  const newId = maxId + 1;
+  itemsDic[newId] = item;
+  reverseLookup.set(canonical, newId);
+  return newId;
+}
 
-  itemsDic[maxId + 1] = item;
-  return maxId + 1;
+export function createReverseLookup<T>(itemsDic: { [id: number]: T }): Map<string, number> {
```
LucasLefevre, line 29: most probably not useful

src/helpers/data_normalization.ts
```diff
@@ -1,25 +1,39 @@
...
 /**
  * Get the id of the given item (its key in the given dictionary).
  * If the given item does not exist in the dictionary, it creates one with a new id.
  */
-export function getItemId<T>(item: T, itemsDic: { [id: number]: T }) {
-  for (const key in itemsDic) {
-    if (deepEquals(itemsDic[key], item)) {
-      return parseInt(key, 10);
-    }
+export function getItemId<T>(
```
LucasLefevre, line 10: a global cache helps to keep the calling code simpler

src/xlsx/functions/conditional_formatting.ts
```diff
@@ -26,6 +26,7 @@ export function addConditionalFormatting(
 ): XMLString[] {
   // Conditional Formats
   const cfNodes: XMLString[] = [];
+
```
LucasLefevre, line 29: useless diff

laa-odoo:
done here: https://github.com/odoo/o-spreadsheet/pull/5819

PR #5674, opened 2025-02-10
[FIX] spreadsheet: handle inverse increment when autofilling upwards/left

Rachico:
Prior to this commit, autofilling upwards or to the left only correctly inverted the increment for numbers, but not when given a group of cells. This led to incorrect results when autofilling mixed content.

The issue occurred because the rule generator always processed values in the same order, regardless of the autofill direction. When a single numeric value was present in a group of cells, the autofill logic failed to apply a negative increment when filling upwards.

To address this, the rule generator now considers the autofill direction.

Steps to reproduce:
1. Insert the following values in a column:
   - A10: 'test'
   - A11: 'test2'
   - A12: '4'
2. Select A10:A12 and autofill upwards.
3. Expected: A9 should be '3'. Before this fix, it incorrectly showed '5'.

Task: [4199568](https://www.odoo.com/odoo/2328/tasks/4199568)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/registries/autofill_rules.ts
+++ b/src/registries/autofill_rules.ts
@@ -7,7 +7,7 @@ import {
   isDateTimeFormat,
 } from "../helpers";
 import { evaluateLiteral } from "../helpers/cells";
-import { AutofillModifier, Cell, CellValueType, DEFAULT_LOCALE } from "../types/index";
+import { AutofillModifier, Cell, CellValueType, DEFAULT_LOCALE, DIRECTION } from "../types/index";
 import { EvaluatedCell, LiteralCell } from "./../types/cells";
 import { Registry } from "./registry";
 
@@ -20,7 +20,7 @@ import { Registry } from "./registry";
  */
 export interface AutofillRule {
   condition: (cell: Cell, cells: (Cell | undefined)[]) => boolean;
-  generateRule: (cell: Cell, cells: (Cell | undefined)[]) => AutofillModifier;
+  generateRule: (cell: Cell, cells: (Cell | undefined)[], direction: DIRECTION) => AutofillModifier;
   sequence: number;
 }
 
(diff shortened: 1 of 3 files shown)


Rachico:
@robodoo r+

src/registries/autofill_rules.ts
```diff
@@ -20,7 +20,11 @@ import { Registry } from "./registry";
  */
 export interface AutofillRule {
   condition: (cell: Cell, cells: (Cell | undefined)[]) => boolean;
-  generateRule: (cell: Cell, cells: (Cell | undefined)[]) => AutofillModifier;
+  generateRule: (
+    cell: Cell,
+    cells: (Cell | undefined)[],
+    direction?: DIRECTION | undefined
```
pro-odoo, line 26: `?` and `undefined` are the same.
By the way, not sure it can be undefined..

pro-odoo (commented):
A odoo branch is required, to handle list and pivot autofill

src/registries/autofill_rules.ts
Rachico: The class attribute can be undefined. Therefore, I need it to type as such to a avoid type compatibility issues.

image

pro-odoo: The class attribute can be undefined in the plugin, because during anything else than an autofill, the state of the plugin has no direction, which is coherent. But in the getRule function, we should have a direction, otherwise it's a bug. So I would make it required in `generateRule`, with a `if` before the call to `generateRule` to ensure that `this.direction` is not undefined. 
We already do a similar thing (but not good because we used a `!`) when a new AutofillGenerator is created (L393)

PR #5679, opened 2025-02-10
[IMP] DV, CF: Enforce range on CF & DV

fdamhaut:
## Description:

Enforce that the ranges on CF & DV belong to the same sheet they do.

Task: [4072640](https://www.odoo.com/odoo/2328/tasks/4072640)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/packages/o-spreadsheet-engine/src/plugins/core/range.ts
+++ b/packages/o-spreadsheet-engine/src/plugins/core/range.ts
@@ -217,7 +217,8 @@ export class RangeAdapter implements CommandHandler<CoreCommand> {
   getRangeString(
     range: Range,
     forSheetId: UID,
-    options: RangeStringOptions = { useBoundedReference: false, useFixedReference: false }
+    options: RangeStringOptions = { useBoundedReference: false, useFixedReference: false },
+    newSheetNames?: Record<UID, string>
   ): string {
     if (!range) {
       return CellErrorType.InvalidReference;
@@ -228,7 +229,15 @@ export class RangeAdapter implements CommandHandler<CoreCommand> {
     if (!this.getters.tryGetSheet(range.sheetId)) {
       return CellErrorType.InvalidReference;
     }
-    return getRangeString(range, forSheetId, this.getters.getSheetName, options);
+    return getRangeString(range, forSheetId, this.sheetNameGetter(newSheetNames), options);
+  }
+
+  sheetNameGetter(newSheetNames?: Record<UID, string>) {
+    if (!newSheetNames) return this.getters.getSheetName;
+    return (sheetId: UID): string => {
+      if (newSheetNames[sheetId]) return newSheetNames[sheetId];
+      return this.getters.getSheetName(sheetId);
+    };
   }
 
   getRangeDataFromXc(sheetId: UID, xc: string): RangeData {
@@ -313,14 +322,24 @@ export class RangeAdapter implements CommandHandler<CoreCommand> {
     if (!formula.startsWith("=")) {
       return formula;
     }
-
+    const newSheetNames: Record<UID, string> = {};
     const compiledFormula = compile(formula);
     const updatedDependencies = compiledFormula.dependencies.map((dep) => {
       const range = this.getters.getRangeFromSheetXC(sheetId, dep);
       const changedRange = applyChange(range);
+      if (changedRange.changeType === "CHANGE" && "newName" in changedRange) {
+        newSheetNames[range.sheetId] = changedRange.newName;
+      }
       return changedRange.changeType === "NONE" ? range : changedRange.range;
     });
-    return this.getters.getFormulaString(sheetId, compiledFormula.tokens, updatedDependencies);
+
+    return this.getters.getFormulaString(
+      sheetId,
+      compiledFormula.tokens,
+      updatedDependencies,
+      false,
+      newSheetNames
+    );
   }
 
   /**
--- a/packages/o-spreadsheet-engine/src/plugins/core/range.ts
+++ b/packages/o-spreadsheet-engine/src/plugins/core/range.ts
@@ -327,7 +327,7 @@ export class RangeAdapter implements CommandHandler<CoreCommand> {
     const updatedDependencies = compiledFormula.dependencies.map((dep) => {
       const range = this.getters.getRangeFromSheetXC(sheetId, dep);
       const changedRange = applyChange(range);
-      if (changedRange.changeType === "CHANGE" && "newName" in changedRange) {
+      if (changedRange.changeType === "RENAME" && "newName" in changedRange) {
         newSheetNames[range.sheetId] = changedRange.newName;
       }
       return changedRange.changeType === "NONE" ? range : changedRange.range;
--- a/src/components/helpers/formulas.ts
+++ b/src/components/helpers/formulas.ts
@@ -1,0 +1,22 @@
+import { rangeTokenize } from "@odoo/o-spreadsheet-engine";
+import { Getters, UID } from "../..";
+
+export function adaptFormulaToSheet(
+  getters: Getters,
+  formula: string,
+  toSheetId: UID,
+  fromSheetId?: UID
+) {
+  return rangeTokenize(formula)
+    .map((token) => {
+      if (token.type === "REFERENCE") {
+        const range = getters.getRangeFromSheetXC(
+          fromSheetId ?? getters.getActiveSheetId(),
+          token.value
+        );
+        return getters.getRangeString(range, toSheetId);
+      }
+      return token.value;
+    })
+    .join("");
+}
--- a/src/components/selection_input/selection_input.ts
+++ b/src/components/selection_input/selection_input.ts
@@ -3,7 +3,7 @@ import { SpreadsheetChildEnv } from "@odoo/o-spreadsheet-engine/types/spreadshee
 import { Component, onWillUpdateProps, useEffect, useRef, useState } from "@odoo/owl";
 import { deepEquals, range } from "../../helpers";
 import { Store, useLocalStore } from "../../store_engine";
-import { Color } from "../../types";
+import { Color, UID } from "../../types";
 import { useDragAndDropListItems } from "../helpers/drag_and_drop_dom_items_hook";
 import { updateSelectionWithArrowKeys } from "../helpers/selection_helpers";
 import { RangeInputValue, SelectionInputStore } from "./selection_input_store";
@@ -13,6 +13,7 @@ interface Props {
   hasSingleRange?: boolean;
   required?: boolean;
   autofocus?: boolean;
+  availableOnSheetId?: UID;
   isInvalid?: boolean;
   class?: string;
   onSelectionChanged?: (ranges: string[]) => void;
@@ -52,6 +53,7 @@ export class SelectionInput extends Component<Props, SpreadsheetChildEnv> {
     hasSingleRange: { type: Boolean, optional: true },
     required: { type: Boolean, optional: true },
     autofocus: { type: Boolean, optional: true },
+    availableOnSheetId: { type: String, optional: true },
     isInvalid: { type: Boolean, optional: true },
     class: { type: String, optional: true },
     onSelectionChanged: { type: Function, optional: true },
@@ -62,6 +64,7 @@ export class SelectionInput extends Component<Props, SpreadsheetChildEnv> {
     disabledRanges: { type: Array, optional: true, default: [] },
     disabledRangeTitle: { type: String, optional: true },
   };
+
   private state: State = useState({
     isMissing: false,
     mode: "select-range",
@@ -76,7 +79,7 @@ export class SelectionInput extends Component<Props, SpreadsheetChildEnv> {
   }
 
   get canAddRange(): boolean {
-    return !this.props.hasSingleRange;
+    return !this.props.hasSingleRange && !this.isReadonly;
   }
 
   get isInvalid(): boolean {
@@ -84,11 +87,17 @@ export class SelectionInput extends Component<Props, SpreadsheetChildEnv> {
   }
 
   get isConfirmable(): boolean {
-    return this.store.isConfirmable;
+    return this.store.isConfirmable && !this.isReadonly;
   }
 
   get isResettable(): boolean {
-    return this.store.isResettable;
+    return this.store.isResettable && !this.isReadonly;
+  }
+
+  get isReadonly(): boolean {
+    return this.props.availableOnSheetId
+      ? this.env.model.getters.getActiveSheetId() !== this.props.availableOnSheetId
+      : false;
   }
 
   get hasDisabledRanges(): boolean {
@@ -105,7 +114,8 @@ export class SelectionInput extends Component<Props, SpreadsheetChildEnv> {
       this.props.ranges,
       this.props.hasSingleRange || false,
       this.props.colors,
-      this.props.disabledRanges
+      this.props.disabledRanges,
+      this.props.availableOnSheetId
     );
     if (this.props.autofocus) {
       this.store.focusById(this.store.selectionInputs[0]?.id);
@@ -172,7 +182,7 @@ export class SelectionInput extends Component<Props, SpreadsheetChildEnv> {
   }
 
   getColor(range: SelectionRange) {
-    if (!range.color) {
+    if (!range.color || this.isReadonly) {
       return "";
     }
     return cssPropertiesToCss({ color: range.color });
--- a/src/components/selection_input/selection_input.xml
+++ b/src/components/selection_input/selection_input.xml
@@ -7,7 +7,8 @@
         t-key="range.id"
         class="o-selection-input d-flex flex-row"
         t-att-style="dragAndDrop.itemsStyle[range.id]"
-        t-att-class="props.class">
+        t-att-class="props.class"
+        t-if="!isReadonly || range.xc">
         <span
           t-if="ranges.length > 1 and props.onSelectionReordered"
           title="Reorder range"
@@ -23,14 +24,15 @@
             t-on-input="(ev) => this.onInputChanged(range.id, ev)"
             t-on-focus="() => this.focus(range.id)"
             t-on-keydown="onKeydown"
+            t-att-disabled="isReadonly"
             t-att-value="range.xc"
             t-att-style="getColor(range)"
             class="o-input mb-2"
             t-att-class="{
               'o-disabled-ranges' : range.disabled and !range.isFocused,
-              'o-focused' : range.isFocused,
+              'o-focused' : range.isFocused and !isReadonly,
               'o-invalid border-danger position-relative': isInvalid || !range.isValidRange,
-              'text-decoration-underline': range.xc and range.isFocused and state.mode === 'select-range'
+              'text-decoration-underline': range.xc and range.isFocused and state.mode === 'select-range' and !isReadonly
             }"
             t-ref="{{range.isFocused ? 'focusedInput' : 'unfocusedInput' + range_index}}"
           />
@@ -49,12 +51,13 @@
         </div>
         <button
           class="border-0 bg-transparent fw-bold o-remove-selection o-button-icon pe-0"
-          t-if="ranges.length > 1"
+          t-if="ranges.length > 1 and !isReadonly"
           t-on-click="() => this.removeInput(range.id)">
           <t t-call="o-spreadsheet-Icon.TRASH_FILLED"/>
         </button>
       </div>
-      <div class="d-flex flex-row w-100 o-selection-input">
+
+      <div t-if="!isReadonly" class="d-flex flex-row w-100 o-selection-input">
         <button class="o-button o-add-selection" t-if="canAddRange" t-on-click="addEmptyInput">
           Add range
         </button>
--- a/src/components/side_panel/conditional_formatting/cf_editor/cf_editor.ts
+++ b/src/components/side_panel/conditional_formatting/cf_editor/cf_editor.ts
@@ -1,3 +1,4 @@
+import { compile } from "@odoo/o-spreadsheet-engine";
 import { ICONS, ICON_SETS } from "@odoo/o-spreadsheet-engine/components/icons/icons";
 import { CfTerms } from "@odoo/o-spreadsheet-engine/components/translations_terms";
 import { DEFAULT_COLOR_SCALE_MIDPOINT_COLOR } from "@odoo/o-spreadsheet-engine/constants";
@@ -8,7 +9,7 @@ import { SpreadsheetChildEnv } from "@odoo/o-spreadsheet-engine/types/spreadshee
 import { hexaToInt } from "@odoo/o-spreadsheet-engine/xlsx/conversion";
 import { Component, ComponentConstructor, useExternalListener, useState } from "@odoo/owl";
 import { Action } from "../../../../actions/action";
-import { colorNumberToHex, colorToNumber, isColorValid, rangeReference } from "../../../../helpers";
+import { colorNumberToHex, colorToNumber, isColorValid } from "../../../../helpers";
 import { cycleFixedReference } from "../../../../helpers/reference_type";
 import {
   criterionComponentRegistry,
@@ -29,11 +30,13 @@ import {
   IconSetRule,
   IconThreshold,
   ThresholdType,
+  UID,
   availableConditionalFormatOperators,
 } from "../../../../types";
 import { ColorPickerWidget } from "../../../color_picker/color_picker_widget";
 import { StandaloneComposer } from "../../../composer/standalone_composer/standalone_composer";
 import { cssPropertiesToCss, getTextDecoration } from "../../../helpers";
+import { adaptFormulaToSheet } from "../../../helpers/formulas";
 import { IconPicker } from "../../../icon_picker/icon_picker";
 import { SelectionInput } from "../../../selection_input/selection_input";
 import { ValidationMessages } from "../../../validation_messages/validation_messages";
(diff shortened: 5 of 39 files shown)


src/components/side_panel/conditional_formatting/conditional_formatting.ts
```diff
@@ -46,18 +46,18 @@ export class ConditionalFormattingPanel extends Component<Props, SpreadsheetChil
...
+      if (this.state.mode == "list") {
+        this.activeSheetId = this.env.model.getters.getActiveSheetId();
+      }
+      if (nextProps.selection !== this.props.selection) {
+        const rules = this.env.model.getters.getRulesSelection(
+          this.activeSheetId,
+          nextProps.selection || []
+        );
+        const cf =
+          rules.length === 1 ? this.conditionalFormats.find((c) => c.id === rules[0]) : undefined;
```
hokolomopo, line 58: why change this ?

src/components/side_panel/conditional_formatting/conditional_formatting.ts
```diff
@@ -46,18 +46,18 @@ export class ConditionalFormattingPanel extends Component<Props, SpreadsheetChil
...
-      } else if (nextProps.selection !== this.props.selection) {
-        const sheetId = this.env.model.getters.getActiveSheetId();
-        const rules = this.env.model.getters.getRulesSelection(sheetId, nextProps.selection || []);
-        if (rules.length === 1) {
-          const cf = this.conditionalFormats.find((c) => c.id === rules[0]);
-          if (cf) {
-            this.editConditionalFormat(cf);
-          }
+      if (this.state.mode == "list") {
+        this.activeSheetId = this.env.model.getters.getActiveSheetId();
```
hokolomopo, line 50: `this.activeSheetId` is useless no ? it's only used when we're in lsit mode, and this make sure that it's always equal to `getters.getActiveSheetId`. Just use `getters.getActiveSheetId` directly.

src/components/translations_terms.ts
```diff
@@ -28,6 +28,7 @@ export const CfTerms = {
     [CommandResult.ValueCellIsInvalidFormula]: _t(
       "At least one of the provided values is an invalid formula"
     ),
+    [CommandResult.TargetOutOfSheet]: _t("The range cannot target out of the current sheet."),
```
hokolomopo, line 31: Message isn't very clear IMO. Since we can leave the side panel open while changing sheet, talking about current sheet is dubious. "The range" when there is no clear way to know about which range we are talking about is also not very clear.

"All the ranges should be in the same sheet as the conditional format" maybe ?

hokolomopo (commented):
pouet pouet :trumpet:

src/components/side_panel/conditional_formatting/conditional_formatting.ts
fdamhaut: It's also used as the defaultSheetId for transforming stringXC to range on save (not in list mode, so still to the value of the rule "main" sheet.

src/components/translations_terms.ts
hokolomopo: Still valid, the message 
image

does not seems very english/clear. Maybe it's just me. And we should probably put the range between quotes.

src/components/side_panel/conditional_formatting/cf_editor/cf_editor.ts
```diff
@@ -198,6 +200,7 @@ export class ConditionalFormattingEditor extends Component<Props, SpreadsheetChi
   setup() {
     this.state = useState<State>({
       errors: [],
+      sheetId: this.env.model.getters.getActiveSheetId(),
```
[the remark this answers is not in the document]
hokolomopo: not addressed

src/components/side_panel/conditional_formatting/cf_editor/cf_editor.ts
```diff
@@ -225,7 +224,33 @@ export class ConditionalFormattingEditor extends Component<Props, SpreadsheetChi
...
 
   get errorMessages(): string[] {
-    return this.state.errors.map((error) => CfTerms.Errors[error] || CfTerms.Errors.Unexpected);
+    return this.state.errors.map((error) => this.errorMessage(error));
+  }
+
+  get firstInvalidRangeString(): string | undefined {
+    const sheetId = this.sheetId;
+    return this.state.ranges.find((xc) => {
+      const range = this.env.model.getters.getRangeDataFromXc(sheetId, xc);
```
hokolomopo, line 233:
```suggestion
      const range = this.env.model.getters.getRangeFromSheetXc(sheetId, xc);
```
don't use RangeData when you do'nt have to. They are a bit ugly, and only meant for commands. We *shouldn't* want to access range._sheetId, it's "private"

src/components/translations_terms.ts
```diff
@@ -5,30 +5,30 @@ import { CommandResult } from "../types/index";
...
+    [CommandResult.SecondArgMissing]: () =>
+      _t("The second argument is missing. Please provide a value"),
+    [CommandResult.MinNaN]: () => _t("The minpoint must be a number"),
+    [CommandResult.MidNaN]: () => _t("The midpoint must be a number"),
+    [CommandResult.MaxNaN]: () => _t("The maxpoint must be a number"),
+    [CommandResult.ValueUpperInflectionNaN]: () => _t("The first value must be a number"),
+    [CommandResult.ValueLowerInflectionNaN]: () => _t("The second value must be a number"),
+    [CommandResult.MinBiggerThanMax]: () => _t("Minimum must be smaller then Maximum"),
+    [CommandResult.MinBiggerThanMid]: () => _t("Minimum must be smaller then Midpoint"),
+    [CommandResult.MidBiggerThanMax]: () => _t("Midpoint must be smaller then Maximum"),
```
hokolomopo, line 19: while you're changing this: you could fix the message. "than" rather than "then". And remove random uppercasing of words

src/plugins/core/conditional_format.ts
```diff
@@ -144,6 +144,8 @@ export class ConditionalFormatPlugin
       case "ADD_CONDITIONAL_FORMAT":
         return this.checkValidations(
           cmd,
+          this.checkEmptyRange,
+          this.checkValidRange,
           this.checkCFRule,
           this.checkEmptyRange,
```
hokolomopo, line 150: this.checkEmptyRange is there twice

src/plugins/core/conditional_format.ts
```diff
@@ -389,6 +391,18 @@ export class ConditionalFormatPlugin
...
 
+  private checkValidRange(cmd: AddConditionalFormatCommand): CommandResult {
+    const ranges = cmd.ranges.map((range) => this.getters.getRangeFromRangeData(range));
+    if (ranges.some((range) => range.sheetId != cmd.sheetId)) {
+      return CommandResult.TargetOutOfSheet;
+    }
+    const stringRanges = ranges.map((range) => this.getters.getRangeString(range, cmd.sheetId));
+    if (stringRanges.some((xc) => !this.getters.isRangeValid(xc))) {
+      return CommandResult.InvalidRange;
+    }
```
hokolomopo, line 402: probably wrong way to do that/useless. We're coming from a RangeData, so either it's a normal rangeData and the XC will always be valid, or it's really malformed and the conversion to range/string will throw and we'll never return `CommandResult.InvalidRange`.

And we already test for invalid zones inside the sheet core plugin.

hokolomopo: Oh you copied what was already in the data validation. Well either i'm wrong and there is cases where you return `CommandResult.InvalidRange` here, or you can remove it from both.

tests/conditional_formatting/conditional_formatting_panel_component.test.ts
```diff
@@ -427,8 +427,27 @@ describe("UI of conditional formats", () => {
...
+
+    test("cannot create a new CF with out of sheet range", async () => {
+      createSheet(model, { sheetId: "s2", name: "Sheet2", activate: false });
+      await click(fixture, selectors.buttonAdd);
+      await nextTick();
+
+      setInputValueAndTrigger(selectors.ruleEditor.range, "Sheet2!A1");
+
+      const dispatch = spyModelDispatch(model);
+      //  click save
```
hokolomopo, line 441: nitpick: comment looks useless

tests/data_validation/data_validation_generics_side_panel_component.test.ts
```diff
@@ -152,7 +152,26 @@ describe("data validation sidePanel component", () => {
...
     expect(errorMessageEl).toBeTruthy();
-    expect(errorMessageEl?.textContent).toContain("The range is invalid.");
+    expect(errorMessageEl?.textContent).toContain("The range A1:HOLA is invalid.");
+  });
+
+  test("Out of sheet range", async () => {
+    createSheet(model, { sheetId: "ID", name: "Sheet2", activate: false });
+    await simulateClick(".o-dv-add");
+    await nextTick();
+    await changeCriterionType("dateIs");
```
hokolomopo, line 162: useless no ?

tests/data_validation/data_validation_generics_side_panel_component.test.ts
```diff
@@ -18,7 +18,7 @@ mockGetBoundingClientRect({
   "o-dv-type": () => dataValidationSelectBoundingRect,
 });
 
-export async function mountDataValidationPanel(model?: Model) {
+async function mountDataValidationPanel(model?: Model) {
```
hokolomopo, line 21: remove this and use the new helper

hokolomopo (commented):
:+1:

src/plugins/core/conditional_format.ts
fdamhaut: I didn't look at it in details but your comment makes sense.

tests/data_validation/data_validation_generics_side_panel_component.test.ts
fdamhaut: What is 'the new helper' ?

hokolomopo: Oh you duplicated the helper in both test files, I though you defined a new helper that you could import everywhere. 

Maybe you should define it somewhere so you can import it from both files ? But probably we don't care, it's only like 4 lines.

src/components/helpers/formulas.ts
```diff
@@ -0,0 +1,23 @@
+import { UID } from "../..";
+import { rangeTokenize } from "../../formulas";
+import { Getters } from "../../types/getters";
+
+export function adaptFormulaToSheet(
```
hokolomopo, line 5: nitpick: maybe could be a getter in range ?

src/components/selection_input/selection_input.ts
```diff
@@ -83,6 +85,10 @@ export class SelectionInput extends Component<Props, SpreadsheetChildEnv> {
     onSelectionRemoved: { type: Function, optional: true },
     colors: { type: Array, optional: true, default: [] },
   };
+  static defaultProps = {
+    readonly: false,
```
hokolomopo, line 89: nitpick: it's a bit useless , undefined is already falsy. And if you have a default props, the props should be typed as non-optional in `Props` (because it will always have a value in practice)

src/components/selection_input/selection_input.ts
```diff
@@ -97,19 +103,19 @@ export class SelectionInput extends Component<Props, SpreadsheetChildEnv> {
   }
 
   get canAddRange(): boolean {
-    return !this.props.hasSingleRange;
+    return !this.props.hasSingleRange && !this.props.readonly;
```
hokolomopo, line 106: Some issues with the readonly style:

- why does the reset/add range button disappear, but the confirm button stays here and is disabled ? We should stay consistent.
- the "trash" button should definitely not be there in readonly
- the selection input should probably lose its colors in readonly. Maybe cancel/confirm the edition to lose focus so we bother lose the colors and the buttons ?
- probably should change the border color rather than the border size when disabling the inputs, otherwise they all jump one pixel up

src/components/selection_input/selection_input.xml
```diff
@@ -15,14 +15,15 @@
           class="o-drag-handle d-flex align-items-center mb-2 o-button-icon">
           <t t-call="o-spreadsheet-Icon.SHORT_THIN_DRAG_HANDLE"/>
         </span>
-        <div class="position-relative w-100">
+        <div class="position-relative w-100" t-att-disabled="props.readonly">
```
hokolomopo, line 18: is this disabled useful ?

src/components/helpers/formulas.ts
```diff
@@ -0,0 +1,23 @@
+import { UID } from "../..";
```
hokolomopo, line 1: see with mehdi because it's probably related to https://www.odoo.com/odoo/2328/tasks/4589387, but the thrash icon does not work when you enter a range from another sheet

src/components/side_panel/conditional_formatting/cf_editor/cf_editor.ts
```diff
@@ -570,14 +605,18 @@ export class ConditionalFormattingEditor extends Component<Props, SpreadsheetChi
     const isInvalid = valueIndex === 0 ? this.isValue1Invalid : this.isValue2Invalid;
     return {
       onConfirm: (str: string) => {
-        this.state.rules.cellIs.values[valueIndex] = str;
+        this.state.rules.cellIs.values[valueIndex] = adaptFormulaToSheet(
+          this.env.model.getters,
+          str,
+          this.props.sheetId
+        );
```
hokolomopo, line 612: what does this do ? When testing it never changes anything ? And I don't get why we could need to modify the ranges of the formula ?

src/components/side_panel/conditional_formatting/cf_editor/cf_editor.ts
```diff
@@ -365,12 +395,17 @@ export class ConditionalFormattingEditor extends Component<Props, SpreadsheetChi
...
+        this.invalidFormulas.includes(this.state.rules.cellIs.values[0]))
     );
   }
 
   get isValue2Invalid(): boolean {
-    return this.state.errors.includes(CommandResult.SecondArgMissing);
+    return (
+      this.state.errors.includes(CommandResult.SecondArgMissing) ||
+      (this.state.errors.includes(CommandResult.ValueCellIsInvalidFormula) &&
+        this.invalidFormulas.includes(this.state.rules.cellIs.values[1]))
```
hokolomopo, line 407: testing both sounds redundant

src/components/side_panel/data_validation/dv_editor/dv_editor.ts
```diff
@@ -134,6 +144,24 @@ export class DataValidationEditor extends Component<Props, SpreadsheetChildEnv>
...
+    switch (reason) {
+      case CommandResult.TargetOutOfSheet:
+      case CommandResult.InvalidRange:
+        return DVTerms.Errors[reason](this.firstInvalidRangeString);
+      default:
+        return DVTerms.Errors[reason]() || DVTerms.Errors.Unexpected();
+    }
+  }
+
+  get firstInvalidRangeString(): string | undefined {
```
hokolomopo, line 160: why stop at the first ? it doesn't cost anything to return them all and join them in the error string no ?

src/components/translations_terms.ts
```diff
@@ -5,30 +5,36 @@ import { CommandResult } from "../types/index";
...
+    [CommandResult.MidInvalidFormula]: () => _t("Invalid midpoint formula"),
+    [CommandResult.ValueUpperInvalidFormula]: () => _t("Invalid upper inflection point formula"),
+    [CommandResult.ValueLowerInvalidFormula]: () => _t("Invalid lower inflection point formula"),
+    [CommandResult.EmptyRange]: () => _t("A range needs to be defined"),
+    [CommandResult.ValueCellIsInvalidFormula]: (formula: string[]) =>
+      _t("Invalid formulas: '%s'", formula.join("', '")),
+    [CommandResult.TargetOutOfSheet]: (sheetName: string, ranges: string[]) =>
+      _t(
+        "Ranges '%s' should target the sheet on which the conditional format is defined (%s)",
+        ranges.join("', '"),
```
hokolomopo, line 34: joint with `"` instead. `'` is used for sheet names with spaces (eg. `'Split Panes'!A1` )

tests/conditional_formatting/conditional_formatting_panel_component.test.ts
```diff
@@ -427,8 +427,38 @@ describe("UI of conditional formats", () => {
...
+
+    test("cannot edit a CF range on another sheet", async () => {
+      createSheet(model, { sheetId: "s2", name: "Sheet2", activate: false });
+      await click(fixture, selectors.buttonAdd);
+      await nextTick();
+
+      model.dispatch("ACTIVATE_NEXT_SHEET");
+      await nextTick();
+
+      const range = document.querySelector(selectors.ruleEditor.range) as HTMLInputElement;
```
hokolomopo, line 458:
```suggestion
      const range = fixture.querySelector<HTMLInputElement>(selectors.ruleEditor.range);
```
nitpick

tests/data_validation/data_validation_generics_side_panel_component.test.ts
```diff
@@ -152,7 +152,25 @@ describe("data validation sidePanel component", () => {
     expect(fixture.querySelector(".o-selection-input .o-invalid")).toBeTruthy();
     const errorMessageEl = fixture.querySelector(".o-validation-error");
     expect(errorMessageEl).toBeTruthy();
-    expect(errorMessageEl?.textContent).toContain("The range is invalid.");
+    expect(errorMessageEl?.textContent).toContain("The range A1:HOLA is invalid.");
+  });
+
+  test("Out of sheet range", async () => {
```
hokolomopo, line 158: mayeb also a test ·`cannot edit a DV range on another sheet`

tests/data_validation/data_validation_list_component.test.ts
```diff
@@ -36,6 +37,13 @@ beforeEach(async () => {
...
+async function mountDataValidationPanel(model?: Model) {
+  return mountComponentWithPortalTarget(DataValidationPanel, {
+    model: model || new Model(),
+    props: { onCloseSidePanel: () => {} },
+  });
+}
+
 describe("Edit criterion in side panel", () => {
   describe("Value in list", () => {
     beforeEach(async () => {
```
hokolomopo, line 49: missing a test for the new props readonly of the selection input

hokolomopo (commented):
Some issues with the style of the disabled selection input. And I'm probably dumb, but I don't get what `adaptFormulaToSheet` does in practice/why it's needed ?  

Mostly nitpicks otherwise :)

src/components/side_panel/conditional_formatting/cf_editor/cf_editor.ts
fdamhaut: If you start editing while one the other sheet, the composer use the current sheet as the 'defaultRangeSheetId' (Meaning A1 = s2!A1) which should be prefixed when saving the string

src/components/helpers/formulas.ts
fdamhaut: There are others functions that affect formulas coming to this file (in others PRs). I feel like range is already big enough and doesn't need more bloating.

src/components/selection_input/selection_input.xml
fdamhaut: not anymore

src/components/helpers/formulas.ts
fdamhaut: It's not related to range from other sheet, in master it simply doesn't work at all :/.
I'll see with mehdi if he already has a fix

src/components/side_panel/data_validation/dv_editor/dv_editor.ts
```diff
@@ -134,6 +144,25 @@ export class DataValidationEditor extends Component<Props, SpreadsheetChildEnv>
...
 
   get errorMessages(): string[] {
-    return this.state.errors.map((error) => DVTerms.Errors[error] || DVTerms.Errors.Unexpected);
+    return this.state.errors.map((error) => this.errorMessage(error));
+  }
+
+  errorMessage(reason: CancelledReason): string {
+    switch (reason) {
+      case CommandResult.TargetOutOfSheet:
+        return DVTerms.Errors[reason](this.props.sheetId, this.invalidRangeString);
```
hokolomopo, line 153: sheetName, not sheetId

src/components/spreadsheet/spreadsheet.ts
```diff
@@ -190,7 +190,14 @@ css/* scss */ `
     }
 
     .o-input[type="text"] {
-      border-width: 0 0 1px 0;
+      &:not([disabled]) {
+        border-width: 0 0 1px 0;
+      }
+
+      &:disabled {
+        border-width: 0 0 0 0;
```
hokolomopo, line 198: not needed anymore. You overwrite it back in the selection input css

src/components/spreadsheet/spreadsheet.ts
```diff
@@ -190,7 +190,14 @@ css/* scss */ `
...
+        cursor: not-allowed;
```
hokolomopo, line 199: Either put all the `disabled` css in the selection input css, or all the css here, I don't think there's a point of having half the css in each place

src/components/spreadsheet/spreadsheet.ts
```diff
@@ -190,7 +190,14 @@ css/* scss */ `
     }
 
     .o-input[type="text"] {
-      border-width: 0 0 1px 0;
+      &:not([disabled]) {
```
hokolomopo, line 193: same here, not needed anymore

hokolomopo (commented):
some small issues,  otherwise LGTM

src/components/side_panel/conditional_formatting/conditional_formatting.ts
```diff
@@ -46,18 +46,18 @@ export class ConditionalFormattingPanel extends Component<Props, SpreadsheetChil
...
-        this.switchToList();
-      } else if (nextProps.selection !== this.props.selection) {
-        const sheetId = this.env.model.getters.getActiveSheetId();
-        const rules = this.env.model.getters.getRulesSelection(sheetId, nextProps.selection || []);
-        if (rules.length === 1) {
-          const cf = this.conditionalFormats.find((c) => c.id === rules[0]);
-          if (cf) {
-            this.editConditionalFormat(cf);
-          }
+      if (this.state.mode == "list") {
```
rrahir:
```suggestion
      if (this.state.mode === "list") {
```
eslint is not happy ^^

src/helpers/reference_type.ts
```diff
@@ -53,7 +54,7 @@ function getTokenNextReferenceType(xc: string): string {
 export function setXcToFixedReferenceType(xc: string, referenceType: FixedReferenceType): string {
   let sheetName;
   ({ sheetName, xc } = splitReference(xc));
-  sheetName = sheetName ? sheetName + "!" : "";
+  sheetName = sheetName ? getCanonicalSymbolName(sheetName) + "!" : "";
```
rrahir: It should probably be a `FIX` commit and we should add a test as well :)

src/components/side_panel/data_validation/dv_editor/dv_editor.ts
```diff
@@ -190,7 +190,7 @@ export class DataValidationEditor extends Component<Props, SpreadsheetChildEnv>
     const sheetId = this.props.sheetId;
     return this.state.rule.ranges.filter((xc) => {
       const range = this.env.model.getters.getRangeDataFromXc(sheetId, xc);
-      return range._sheetId != sheetId || !this.env.model.getters.isRangeValid(xc);
```
rrahir: all the eslint corrections should probably be part of the commit in which they were introduced but not that big of a deal

rrahir (changes requested):

packages/o-spreadsheet-engine/src/plugins/core/range.ts
```diff
@@ -228,7 +229,15 @@ export class RangeAdapter implements CommandHandler<CoreCommand> {
     if (!this.getters.tryGetSheet(range.sheetId)) {
       return CellErrorType.InvalidReference;
     }
-    return getRangeString(range, forSheetId, this.getters.getSheetName, options);
+    return getRangeString(range, forSheetId, this.sheetNameGetter(newSheetNames), options);
+  }
+
+  sheetNameGetter(newSheetNames?: Record<UID, string>) {
```
LucasLefevre, line 235: If it's used only once, make it closure instead of a (public) method

packages/o-spreadsheet-engine/src/plugins/core/range.ts
```diff
@@ -313,14 +322,24 @@ export class RangeAdapter implements CommandHandler<CoreCommand> {
     if (!formula.startsWith("=")) {
       return formula;
     }
-
+    const newSheetNames: Record<UID, string> = {};
     const compiledFormula = compile(formula);
     const updatedDependencies = compiledFormula.dependencies.map((dep) => {
       const range = this.getters.getRangeFromSheetXC(sheetId, dep);
       const changedRange = applyChange(range);
+      if (changedRange.changeType === "RENAME" && "newName" in changedRange) {
```
LucasLefevre, line 330: according to the type, `newName` is always defined
```suggestion
      if (changedRange.changeType === "RENAME") {
```

packages/o-spreadsheet-engine/src/plugins/core/range.ts
```diff
@@ -313,14 +322,24 @@ export class RangeAdapter implements CommandHandler<CoreCommand> {
     if (!formula.startsWith("=")) {
       return formula;
     }
-
+    const newSheetNames: Record<UID, string> = {};
```
LucasLefevre, line 325: I really think we should try to have the sheet plugin handling `RENAME_SHEET` before `applyChange` are called everywhere.
I would be much simpler everywhere

PR #5687, opened 2025-02-11
[FIX] composer: multiline composer breaks the topbar layout

hokolomopo:
## Description

Opening a multi-line composer in the topbar would break the layout of the topbar since the topbar does not have a fixed height anymore with 6bfcd98.

This commit fixes that by adding a container around the composer with a fixed height.

Task: [4521789](https://www.odoo.com/odoo/2328/tasks/4521789)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


rrahir:
@robodoo r+

PR #5689, opened 2025-02-11
[FIX] UUID: make `smallUuid` stronger

rrahir:
The recent introduction of `smallUuid` highlighted an unexpected use of the uuid's. In the internal mapping of plugins, we create objects which keys are uuids and we expect the Object to respect the order of insertion.

Unfortunately, the order of insertion in an Object is not guaranteed all the time. Specifically, the keys that are `number`or that can be coerced as a number have a priority over `string`.

Since `uuidv4` contains a dash (`-`), it cannot be coerced as a number but `smallUuid` surely can. It turns out its use can yield to unexpected behaviours and indeterministic tests.

Task: 0

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/uuid.ts
+++ b/src/helpers/uuid.ts
@@ -26,13 +26,13 @@ export class UuidGenerator {
       return String(this.fastIdStart);
       //@ts-ignore
     } else if (window.crypto && window.crypto.getRandomValues) {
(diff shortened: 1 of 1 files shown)


rrahir:
@robodoo fw=no

PR #5713, opened 2025-02-13
[FIX] tests: remove mock uuid

hokolomopo:
## Description

We have a file that mocks the uuid generator for the tests. This was used to make some test deterministic. We can (and should) remove this because:

- Most of the times we didn't even need to mock the uuid generator
- In most of the test that did need it, we don't care about the generated id and can use expect.any(String) instead
- In the few cases where we do care about the generated id, it was not reliable. A change that created a new uuid between the moment of the mock and the generation of the tested id would break the test

Task: [0](https://www.odoo.com/odoo/2328/tasks/0)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


VincentSchippefilt:
robodoo r+

PR #5734, opened 2025-02-14
[FIX] selection input: highlight color not sync with range color

hokolomopo:
## Description

After re-ordering the selection inputs of a chart, the highlight colors would not be the same as the range colors.

This was because the store mutator `updateColors` would not update the store range colors. And `store.highlights` is based on `ranges.colors` while `store.selectionInputs` ignores the colors of `store.ranges`.

The test `update of colors are taken into account` was also a bit sketchy. We would modify the `props.colors` array in place, thus the condition `nextProps.colors !== props.colors` would never be satisfied, and `store.updateColors` would never be called.

Task: [4577803](https://www.odoo.com/odoo/2328/tasks/4577803)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/selection_input/selection_input_store.ts
+++ b/src/components/selection_input/selection_input_store.ts
@@ -31,7 +31,7 @@ export class SelectionInputStore extends SpreadsheetStore {
     "confirm",
     "updateColors",
   ] as const;
-  ranges: RangeInputValue[] = [];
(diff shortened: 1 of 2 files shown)


rrahir:
@robodoo r+

PR #4838, opened 2024-08-13
[IMP] Charts: ignore hidden row/cols in datasets and labels

anhe-odoo:
## Task Description

This Task aims to ignore the hidden columns/rows when computing the data values and labels in the charts runtime.

## Related Task
- Task: 3976159

## review checklist

- [ ] feature is organized in plugin, or UI components
- [ ] support of duplicate sheet (deep copy)
- [ ] in model/core: ranges are Range object, and can be adapted (adaptRanges)
- [ ] in model/UI: ranges are strings (to show the user)
- [ ] undo-able commands (uses this.history.update)
- [ ] multiuser-able commands (has inverse commands and transformations where needed)
- [ ] new/updated/removed commands are documented
- [ ] exportable in excel
- [ ] translations (\_t("qmsdf %s", abc))
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes
- [ ] doc is rebuild (npm run doc)
- [ ] status is correct in Odoo

PR #5741, opened 2025-02-14
[FIX] Charts: Hide datasets instead of filtering them

rrahir:
Currently, we filter out datasets that do not contain visible data in
the chart runtime data (see https://github.com/odoo/o-spreadsheet/pull/4838). Unfortunately, we inadvertently
created desynchronization between the datasets in the runtime and the
ones in the chart definition, which can lead to some errors while
generating the rest of the runtime (e.g. each dataset has some design
attributes and by filtering the runtime dataset uphill, we no longer
know which design belongs to which dataset).

This revision suggest another approach: Keeping the datasets intact
and mark them as 'hidden' to maintain the '1 to 1' relation between the
core definition and runtime arbitrarily filter out the "hidden" datasets
during the rendering of the chart. We align ourselves with the
data structure[^1] of chartJs to make it easier.

[^1]: https://www.chartjs.org/docs/latest/general/data-structures.html#dataset-configuration

## Description:

Task: [4577744](https://www.odoo.com/odoo/2328/tasks/4577744)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/figures/charts/bar_chart.ts
+++ b/src/helpers/figures/charts/bar_chart.ts
@@ -53,6 +53,7 @@ import {
   filterEmptyDataPoints,
   getChartDatasetFormat,
   getChartDatasetValues,
+  getChartJsLegend,
   getChartLabelValues,
   getDefaultChartJsRuntime,
 } from "./chart_ui_common";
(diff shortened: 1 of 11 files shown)


LucasLefevre (approved):
robodoo r+

PR #5756, opened 2025-02-18
Master sunburst chart adrm

hokolomopo:
### [IMP] chart: add sunburst chart

This commit adds a new chart type: sunburst chart. The sunburst chart
is used to display hierarchical data in a circular chart.

Technically, the chart is implemented using a doughnut chart with
multiple levels and invisible segments.

### [IMP] chart: add `ChartWithAxisDefinition` type

Before this, we used the type `ChartWithDataSetDefinition` in the
component `AxisDesignEditor`. Which was wrong because there are charts
with datasets but no axis (geo, pie). It forced those chart to have a
unused key `axisDesign` in their definition.

The `aggregated` key in the geo chart definition is also never used
(they are always aggregated).

### [IMP] tests: define default creation context for chart tests

Every type of chart has a test that checks what happens if we
create a chart from a creation context. But every time we add something
new to to creation context, we have to change every single test, even
if the chart doesn't use the new key of the context.

This commits adds a constant with a default creation context to
simplify the tests.

### [IMP] chart: extract `ChartTitle` tools into `TextStyler`

This commit extract the style editing tools of the `ChartTitle`
component into a new component called `TextStyler`. This will be used
in the following commits to have a more generic component that can be
used to style any text in the chart.

It also adds add the possibility ot edit a background color or a
vertical alignment in this `TextStyler` component.

The commit also simplifies the TextStyler compared to the previous
behaviour of `ChartTitle` by:

- deleting all the props updateColor, toggleBold, ... in favor of a
single callback updateStyle
- using ActionButton components instead of custom HTML that looks like
an action button

### [MOV] chart: rename `title` files to `chart_title`

Task: [4575651](https://www.odoo.com/odoo/2328/tasks/4575651)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


demo/data.js
```diff
@@ -406,9 +406,74 @@ export const demoData = {
...
+        V12: "Drama",
+        V13: "Drama",
+        V14: "Drama",
+        V15: "Drama",
+        V16: "Drama",
+        V17: "Action",
+        V18: "Animation",
+        V19: "Drama",
+        V20: "Action",
+        V21: "Animation",
```
rrahir, line 429: V22: "Misc",
V23: "Misc"

Add a category with a single level to display the full sunburst visuals

demo/data.js
```diff
@@ -406,9 +406,74 @@ export const demoData = {
...
+        X12: "2201000000",
+        X13: "1074000000",
+        X14: "678200000",
+        X15: "910000000",
+        X16: "533000000",
+        X17: "1347000000",
+        X18: "935300000",
+        X19: "701800000",
+        X20: "1215000000",
+        X21: "807800000",
```
rrahir, line 471: X22: "3215000000",
X23: "1212000000"

demo/data.js
```diff
@@ -683,6 +748,29 @@ export const demoData = {
...
+          y: 1800,
+          width: 400,
+          height: 400,
+          tag: "chart",
+          data: {
+            type: "sunburst",
+            dataSetsHaveTitle: true,
+            dataSets: [
+              {
+                dataRange: "V1:V21",
```
rrahir, line 763:
```suggestion
                dataRange: "V1:V23",
```

demo/data.js
```diff
@@ -683,6 +748,29 @@ export const demoData = {
...
+              },
+              {
+                dataRange: "W1:W21",
```
rrahir, line 766:
```suggestion
                dataRange: "W1:W23",
```

demo/data.js
```diff
@@ -683,6 +748,29 @@ export const demoData = {
...
+              },
+            ],
+            legendPosition: "top",
+            labelRange: "X1:X21",
```
rrahir, line 770:
```suggestion
            labelRange: "X1:X23",
```

demo/data.js
```diff
@@ -700,6 +788,21 @@ export const demoData = {
             styleId: "TableStyleMedium7",
           },
         },
+        {
+          range: "V1:X21",
```
rrahir, line 792:
```suggestion
          range: "V1:X23",
```

src/components/figures/chart/chartJs/chartjs_sunburst_labels_plugin.ts
```diff
@@ -0,0 +1,108 @@
...
+  callback: (value: number, axisId?: string) => string;
+}
+
+declare module "chart.js" {
+  interface PluginOptionsByType<TType extends ChartType> {
+    sunburstLabelsPlugin?: ChartSunburstLabelsPluginOptions;
+  }
+}
+
+const Y_PADDING = 3;
```
rrahir, line 25: I trust you :D

src/helpers/figures/charts/runtime/chartjs_dataset.ts
```diff
@@ -368,6 +375,132 @@ export function getFunnelLabelColors(labels: string[], colors?: FunnelChartColor
...
+  const dataSets: SunburstChartJSDataset[] = [];
+  for (let i = data.length - 1; i >= 0; i--) {
+    const dataset: SunburstChartJSDataset = {
+      groupColors,
+      parsing: { key: "value" },
+      data: data[i] as any,
+      borderColor: (ctx) => {
+        const data = ctx.type === "data" ? (ctx.raw as SunburstChartRawData) : undefined;
+        if (!data || data.label === GHOST_SUNBURST_VALUE) {
+          return "rgba(0, 0, 0, 0)";
```
VincentSchippefilt, line 403: nit: should probably be a constant like GHOST_SUNBURST_STYLE

src/helpers/figures/charts/runtime/chart_data_extractor.ts
```diff
@@ -275,6 +276,34 @@ export function getGeoChartData(
...
+  const removeFirstLabel = shouldRemoveFirstLabel(
+    labelRange,
+    dataSets[0],
+    definition.dataSetsHaveTitle || false
+  );
+  if (removeFirstLabel) {
+    labels.shift();
+  }
+  ({ labels, dataSetsValues } = filterInvalidHierarchicalPoints(labels, dataSetsValues));
+  ({ labels, dataSetsValues } = filterValuesWithDifferentSigns(labels, dataSetsValues));
```
VincentSchippefilt, line 297: do you think we should switch those 2 lines : the complexity of `filterInvalidHierarchicalPoints` is bigger than the one of `filterValuesWithDifferentSigns`, so it might be faster to restrict first based on sign

hokolomopo: sure, it should be slightly better

rrahir:
@robodoo rebase-ff

PR #5765, opened 2025-02-19
[IMP] clipboard: ignore hidden cells when copying ranges

anhe-odoo:
## Task Description

When copying a range, we should ignore the hidden rows/columns and not paste their contents. This should be done only for the copy operation, not the cut.

## Related Task
- Task: [3872420](https://www.odoo.com/odoo/2328/tasks/3872420)

### DIFF
--- a/src/clipboard_handlers/abstract_clipboard_handler.ts
+++ b/src/clipboard_handlers/abstract_clipboard_handler.ts
@@ -12,7 +12,7 @@ import {
 export class ClipboardHandler<T> {
   constructor(protected getters: Getters, protected dispatch: CommandDispatcher["dispatch"]) {}
 
-  copy(data: ClipboardData): T | undefined {
+  copy(data: ClipboardData, isCutOperation: boolean): T | undefined {
     return;
   }
 
--- a/src/clipboard_handlers/tables_clipboard.ts
+++ b/src/clipboard_handlers/tables_clipboard.ts
@@ -41,7 +41,7 @@ export class TableClipboardHandler extends AbstractCellClipboardHandler<
   ClipboardContent,
   TableCell
 > {
-  copy(data: ClipboardCellData): ClipboardContent {
+  copy(data: ClipboardCellData, isCutOperation?: boolean): ClipboardContent {
     const sheetId = data.sheetId;
 
     const { rowsIndexes, columnsIndexes, zones } = data;
@@ -69,8 +69,20 @@ export class TableClipboardHandler extends AbstractCellClipboardHandler<
           zones.some((z) => isZoneInside(tableZone, z))
         ) {
           copiedTablesIds.add(table.id);
+          let { numberOfRows } = zoneToDimension(tableZone);
+          for (let rowIndex = tableZone.top; rowIndex <= tableZone.bottom; rowIndex++) {
+            if (!isCutOperation && !rowsIndexes.includes(rowIndex)) {
+              numberOfRows--;
+            }
+          }
+          const range = coreTable.range;
+          const newRange = this.getters.extendRange(
+            coreTable.range,
+            "ROW",
+            range.zone.top + numberOfRows - 1 - range.zone.bottom
+          );
           copiedTable = {
-            range: this.getters.getRangeData(coreTable.range),
+            range: this.getters.getRangeData(newRange),
             config: coreTable.config,
             type: coreTable.type,
           };
--- a/tests/clipboard/clipboard_plugin.test.ts
+++ b/tests/clipboard/clipboard_plugin.test.ts
@@ -32,11 +32,14 @@ import {
   createSheet,
   createSheetWithName,
   createTable,
+  createTableWithFilter,
   cut,
   deleteCells,
   deleteColumns,
   deleteRows,
   deleteSheet,
+  hideColumns,
+  hideRows,
   insertCells,
   merge,
   paste,
@@ -52,6 +55,7 @@ import {
   setZoneBorders,
   unMerge,
   undo,
+  updateFilter,
   updateLocale,
 } from "../test_helpers/commands_helpers";
 import {
@@ -2026,6 +2030,86 @@ describe("clipboard", () => {
     cut(model, "B1");
     paste(model, "C3");
     expect(getCell(model, "C3")!.content).toBe(expectedInvalidFormula);
+  });
+
+  test("filtered rows are ignored when copying range", () => {
+    //prettier-ignore
+    const model = createModelFromGrid({
+      B2: "b2", C2: "c2", D2: "d2",
+      B3: "b3", C3: "c3", D3: "d3",
+      B4: "b4", C4: "c4", D4: "d4",
+    });
+
+    createTableWithFilter(model, "B2:D4");
+    updateFilter(model, "C3", ["c3"]);
+    copy(model, "B2:D4");
+    paste(model, "B5");
+
+    expect(getEvaluatedGrid(model, "B5:D7")).toEqual([
+      ["b2", "c2", "d2"],
+      ["b4", "c4", "d4"],
+      ["", "", ""],
+    ]);
+  });
+
+  test("filtered rows are ignored when cutting range", () => {
+    //prettier-ignore
+    const model = createModelFromGrid({
+      B2: "b2", C2: "c2", D2: "d2",
+      B3: "b3", C3: "c3", D3: "d3",
+      B4: "b4", C4: "c4", D4: "d4",
+    });
+
+    createTableWithFilter(model, "B2:D4");
+    updateFilter(model, "C3", ["c3"]);
+    cut(model, "B2:D4");
+    paste(model, "B5");
+
+    expect(getEvaluatedGrid(model, "B5:D7")).toEqual([
+      ["b2", "c2", "d2"],
+      ["b3", "c3", "d3"],
+      ["b4", "c4", "d4"],
+    ]);
+  });
+
+  test("hidden rows/columns are taken into account when copypasting range", () => {
+    //prettier-ignore
+    const model = createModelFromGrid({
+      B2: "b2", C2: "c2", D2: "d2",
+      B3: "b3", C3: "c3", D3: "d3",
+      B4: "b4", C4: "c4", D4: "d4",
+    });
+
+    hideRows(model, [2]);
+    hideColumns(model, ["C"]);
+    copy(model, "B2:D4");
+    paste(model, "E1");
+
+    expect(getEvaluatedGrid(model, "E1:G3")).toEqual([
+      ["b2", "c2", "d2"],
+      ["b3", "c3", "d3"],
+      ["b4", "c4", "d4"],
+    ]);
+  });
+
+  test("hidden rows/columns are taken into account when cutpasting range", () => {
+    //prettier-ignore
+    const model = createModelFromGrid({
+      B2: "b2", C2: "c2", D2: "d2",
+      B3: "b3", C3: "c3", D3: "d3",
+      B4: "b4", C4: "c4", D4: "d4",
+    });
+
+    hideRows(model, [2]);
+    hideColumns(model, ["C"]);
+    cut(model, "B2:D4");
+    paste(model, "E1");
+
+    expect(getEvaluatedGrid(model, "E1:G3")).toEqual([
+      ["b2", "c2", "d2"],
+      ["b3", "c3", "d3"],
+      ["b4", "c4", "d4"],
+    ]);
   });
 
   test("copying a spread pivot cell results in the fixed pivot formula", () => {
--- a/src/components/paint_format_button/paint_format_store.ts
+++ b/src/components/paint_format_button/paint_format_store.ts
@@ -70,7 +70,7 @@ export class PaintFormatStore extends SpreadsheetStore {
 
     const copiedData = {};
     for (const handler of this.clipboardHandlers) {
-      Object.assign(copiedData, handler.copy(getClipboardDataPositions(sheetId, zones)));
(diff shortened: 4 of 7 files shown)


src/clipboard_handlers/merge_clipboard.ts
```diff
@@ -18,12 +18,15 @@ export class MergeClipboardHandler extends AbstractCellClipboardHandler<
...
   Maybe<Merge>
 > {
-  copy(data: ClipboardCellData): ClipboardContent | undefined {
+  copy(data: ClipboardCellData, isCutOperation?: boolean): ClipboardContent | undefined {
     const sheetId = this.getters.getActiveSheetId();
     const { rowsIndexes, columnsIndexes } = data;
     const merges: Maybe<Merge>[][] = [];
 
     for (const row of rowsIndexes) {
+      if (!isCutOperation && this.getters.isRowFiltered(sheetId, row)) {
```
hokolomopo, line 27: merges with half their rows filtered are still copied in full. Not sure if it's/what the correct behaviour is

tests/clipboard/clipboard_plugin.test.ts
```diff
@@ -2001,6 +2005,127 @@ describe("clipboard", () => {
...
+    paste(model, "B5");
+
+    expect(getEvaluatedGrid(model, "B5:D7")).toEqual([
+      ["b2", "c2", "d2"],
+      ["b4", "c4", "d4"],
+      ["", "", ""],
+    ]);
+  });
+
+  test("hidden row are taken into account when copypasting range", () => {
```
hokolomopo, line 2033:
```suggestion
  test("hidden rows are taken into account when copypasting range", () => {
```

tests/clipboard/clipboard_plugin.test.ts
```diff
@@ -2001,6 +2005,127 @@ describe("clipboard", () => {
...
+    paste(model, "E1");
+
+    expect(getEvaluatedGrid(model, "E1:G3")).toEqual([
+      ["b2", "c2", "d2"],
+      ["b3", "c3", "d3"],
+      ["b4", "c4", "d4"],
+    ]);
+  });
+
+  test("hidden row are taken into account when cutpasting range", () => {
```
hokolomopo, line 2057:
```suggestion
  test("hidden rows are taken into account when cutpasting range", () => {
```

hokolomopo: should also test that filted rows are kept on cut/paste

nitpick: could also test both hidden rows/cols in a single test if we want to gain some lines of code

tests/clipboard/clipboard_plugin.test.ts
```diff
@@ -2001,6 +2005,127 @@ describe("clipboard", () => {
     expect(getCell(model, "C3")!.content).toBe(expectedInvalidFormula);
   });
 
+  test("filtered rows are ignored when copying range", () => {
+    const model = createModelFromGrid({
```
hokolomopo, line 2009: nitpick: would be prettier/shorter with a  prettier-ignore and a grid-like display

hokolomopo (commented):
:wave:

src/clipboard_handlers/abstract_cell_clipboard_handler.ts
```diff
@@ -4,7 +4,7 @@ import { ClipboardCellData, ClipboardOptions, HeaderIndex, UID, Zone } from "../
 import { ClipboardHandler } from "./abstract_clipboard_handler";
 
 export class AbstractCellClipboardHandler<T, T1> extends ClipboardHandler<T> {
-  copy(data: ClipboardCellData): T | undefined {
+  copy(data: ClipboardCellData, isCutOperation?: boolean): T | undefined {
```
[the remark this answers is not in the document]
anhe-odoo: Looks like it doesn't make boom :)

src/clipboard_handlers/merge_clipboard.ts
anhe-odoo: Probably not the best solution (but can't figure what the best solution should be), but we have the same behavior as GSheet

src/plugins/ui_stateful/selection.ts
```diff
@@ -546,7 +546,7 @@ export class GridSelectionPlugin extends UIPlugin {
 
     const sheetId = this.getActiveSheetId();
     const handler = new CellClipboardHandler(this.getters, this.dispatch);
-    const data = handler.copy(getClipboardDataPositions(sheetId, target));
+    const data = handler.copy(getClipboardDataPositions(sheetId, target), true);
```
hokolomopo, line 549: nitpick: maybe be should change the boolean arg  isCutOperation to `"copy" | "paste"` so random calls to the method are more readable

src/clipboard_handlers/borders_clipboard.ts
```diff
@@ -26,6 +26,9 @@ export class BorderClipboardHandler extends AbstractCellClipboardHandler<
     const borders: (Border | null)[][] = [];
 
     for (const row of rowsIndexes) {
+      if (!isCutOperation && this.getters.isRowFiltered(sheetId, row)) {
+        continue;
+      }
```
hokolomopo, line 31: idea: remove filtered rows directly from from rowsIndexes, that way we do'nt ahve to repeat this every clipbboard hanldelr

src/clipboard_handlers/abstract_clipboard_handler.ts
```diff
@@ -12,7 +12,7 @@ import {
 export class ClipboardHandler<T> {
```
hokolomopo, line 12: strange behaviour with data cleanup, remove duplicate of filtered rows don't work, but from hidden rows do work. Idk if that's meant to be the case.

hokolomopo (commented):
:wave:

src/clipboard_handlers/abstract_clipboard_handler.ts
anhe-odoo: Fixed :) No we have the same behavior

hokolomopo:
Copying fulll table with filtered rows leads to blank spaces in the table
image

src/clipboard_handlers/tables_clipboard.ts
```diff
@@ -166,16 +168,27 @@ export class TableClipboardHandler extends AbstractCellClipboardHandler<
...
   ) {
     if (tableCell.table && !options?.pasteOption) {
       const { range: tableRange } = tableCell.table;
-      const zoneDims = zoneToDimension(this.getters.getRangeFromRangeData(tableRange).zone);
+      const zone = this.getters.getRangeFromRangeData(tableRange).zone;
+      const zoneDims = zoneToDimension(zone);
+      let { numberOfCols, numberOfRows } = zoneDims;
+      if (!options?.isCutOperation) {
+        for (let r = zone.top; r <= zone.bottom; r++) {
+          if (this.getters.isRowFiltered(sheetId, r)) {
```
hokolomopo, line 182: should be done in the `copy`, not on the `paste`. We can filter/unfilter rows between the moment of the copy and the paste, in which case this won't work.

src/clipboard_handlers/tables_clipboard.ts
```diff
@@ -69,10 +70,20 @@ export class TableClipboardHandler extends AbstractCellClipboardHandler<
           zones.some((z) => isZoneInside(tableZone, z))
         ) {
           copiedTablesIds.add(table.id);
+          let { numberOfCols, numberOfRows } = zoneToDimension(tableZone);
+          for (let r = tableZone.top; r <= tableZone.bottom; r++) {
+            if (
+              !isCutOperation &&
+              (this.getters.isRowFiltered(sheetId, r) || !rowsIndexes.includes(r))
```
hokolomopo, line 77: `this.getters.isRowFiltered(sheetId, r)` condition is useless no ? filtered rows aren't in rowIndexes

src/clipboard_handlers/tables_clipboard.ts
```diff
@@ -166,16 +179,20 @@ export class TableClipboardHandler extends AbstractCellClipboardHandler<
...
+      const zone = this.getters.getRangeFromRangeData(tableRange).zone;
+      const zoneDims = zoneToDimension(zone);
+      const { numberOfCols, numberOfRows } = dimension || zoneDims;
       const newTableZone = {
         left: position.col,
         top: position.row,
-        right: position.col + zoneDims.numberOfCols - 1,
-        bottom: position.row + zoneDims.numberOfRows - 1,
+        right: position.col + Math.min(numberOfCols, nCols) - 1,
+        bottom: position.row + Math.min(numberOfRows, nRows) - 1,
```
hokolomopo, line 195: In what cases using tableCell.table.dimensions isn't enough ? Why do we also need to check the size of the original table zone and the number of tableCells ?

hokolomopo (commented):
some small things, otherwise LGTM :)

src/clipboard_handlers/tables_clipboard.ts
anhe-odoo: You're right. It was needed before the last version, when I wasn't recomputing `tableCell.table.dimensions` but using the original table zone.

src/plugins/ui_stateful/selection.ts
```diff
@@ -552,7 +552,7 @@ export class GridSelectionPlugin extends UIPlugin {
 
     const sheetId = this.getActiveSheetId();
     const handler = new CellClipboardHandler(this.getters, this.dispatch);
-    const data = handler.copy(getClipboardDataPositions(sheetId, target));
+    const data = handler.copy(getFilteredClipboardDataPositions(sheetId, target, this.getters));
```
LucasLefevre, line 555: I think this change is useless since it a cut operation here. Which means `getFilteredClipboardDataPositions` is only used once and it can be inlined

src/clipboard_handlers/tables_clipboard.ts
```diff
@@ -69,10 +70,17 @@ export class TableClipboardHandler extends AbstractCellClipboardHandler<
           zones.some((z) => isZoneInside(tableZone, z))
         ) {
           copiedTablesIds.add(table.id);
+          let { numberOfCols, numberOfRows } = zoneToDimension(tableZone);
+          for (let r = tableZone.top; r <= tableZone.bottom; r++) {
+            if (!isCutOperation && !rowsIndexes.includes(r)) {
+              numberOfRows--;
+            }
+          }
```
LucasLefevre, line 78: Could you instead just reduce (a copy of) `coreTable.range` here ? instead of adding this new `dimension` ?

tests/clipboard/clipboard_plugin.test.ts
```diff
@@ -2028,6 +2032,86 @@ describe("clipboard", () => {
...
+    paste(model, "B5");
+
+    expect(getEvaluatedGrid(model, "B5:D7")).toEqual([
+      ["b2", "c2", "d2"],
+      ["b4", "c4", "d4"],
+      ["", "", ""],
+    ]);
+  });
+
+  test("filtered rows are ignored when copying range", () => {
```
LucasLefevre, line 2055: *"filtered rows are ignored when **cutting** range"*

src/clipboard_handlers/tables_clipboard.ts
```diff
@@ -69,8 +69,16 @@ export class TableClipboardHandler extends AbstractCellClipboardHandler<
           zones.some((z) => isZoneInside(tableZone, z))
         ) {
           copiedTablesIds.add(table.id);
+          let { numberOfRows } = zoneToDimension(tableZone);
+          for (let r = tableZone.top; r <= tableZone.bottom; r++) {
```
LucasLefevre, line 73: We can spend 2 more characters for the sake of readability
```suggestion
          for (let row = tableZone.top; row <= tableZone.bottom; row++) {
```

src/clipboard_handlers/tables_clipboard.ts
```diff
@@ -69,8 +69,16 @@ export class TableClipboardHandler extends AbstractCellClipboardHandler<
...
+            if (!isCutOperation && !rowsIndexes.includes(r)) {
+              numberOfRows--;
+            }
+          }
+          const range = this.getters.getRangeData(coreTable.range);
+          range._zone = { ...range._zone, bottom: range._zone.top + numberOfRows - 1 };
```
LucasLefevre, line 79: avoid using and manipulating `_zone`. The `_` indicates it's (almost) private and shouldn't be used.
You can use `getters.extendRange` to reduce the range.

```ts
          const range = coreTable.range;
          const newRange = this.getters.extendRange(coreTable.range, "ROW", range.zone.bottom - range.zone.top - numberOfRows + 1)
          copiedTable = {
            range: this.getters.getRangeData(newRange),
```

src/clipboard_handlers/tables_clipboard.ts
```diff
@@ -170,12 +178,14 @@ export class TableClipboardHandler extends AbstractCellClipboardHandler<
   ) {
     if (tableCell.table && !options?.pasteOption) {
       const { range: tableRange } = tableCell.table;
-      const zoneDims = zoneToDimension(this.getters.getRangeFromRangeData(tableRange).zone);
+      const zone = this.getters.getRangeFromRangeData(tableRange).zone;
+      const zoneDims = zoneToDimension(zone);
+      const { numberOfCols, numberOfRows } = zoneDims;
```
LucasLefevre, line 183: this diff can be reverted now

src/clipboard_handlers/tables_clipboard.ts
anhe-odoo: This loop is inside another for loop that use the "row" variable ...

rrahir:
@robodoo rebase-ff

PR #5766, opened 2025-02-19
[FIX] chart: tooltip issues

hokolomopo:
### [FIX] chart: handle tooltip with colons in the label

To split the labels and values in the tooltips, we did a simple split
by colon, since our labels had the format `label: value`. But this
obviously breaks when the label itself contains a colon.

With this commit, we now use the `beforeLabel` tooltip callback to
get the label, and the `label` callback to get the value instead of
doing everything in the `label` callback.

### [FIX] chart: remove empty title in tooltip

If there is no title in the tooltip, we still displayed the div
containing the title. That led to a small empty space at the top
of the tooltip.

Task: [4586683](https://www.odoo.com/odoo/2328/tasks/4586683)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/figures/charts/runtime/chartjs_tooltip.ts
+++ b/src/helpers/figures/charts/runtime/chartjs_tooltip.ts
@@ -35,8 +35,8 @@ export function getBarChartTooltip(
           ? undefined
           : "";
       },
+      beforeLabel: (tooltipItem) => tooltipItem.dataset?.label || tooltipItem.label,
       label: function (tooltipItem) {
-        const xLabel = tooltipItem.dataset?.label || tooltipItem.label;
         const horizontalChart = definition.horizontal;
         let yLabel = horizontalChart ? tooltipItem.parsed.x : tooltipItem.parsed.y;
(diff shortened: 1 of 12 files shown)


src/helpers/figures/charts/runtime/chartjs_tooltip.ts
```diff
@@ -117,19 +114,17 @@ export function getPieChartTooltip(
       title: function (tooltipItems) {
         return tooltipItems[0].dataset.label;
       },
+      beforeLabel: (tooltipItem) => tooltipItem.dataset?.label || tooltipItem.label,
```
LucasLefevre, line 117: I think it should be reversed
`tooltipItem.dataset?.label || tooltipItem.label` should be `tooltipItem.label || tooltipItem.dataset.label`.

copy-paste went too far :smile: 
image

LucasLefevre (approved):
robodoo r+

LucasLefevre:
robodoo rebase-ff

PR #5768, opened 2025-02-19
[FIX] chart: only spread relevant options in datasets

hokolomopo:
## Description

In the chart side panel, if we enter a range like `A1:B3` it will be spread into multiple datasets (`A1:A3` and `B1:B3`). The problem is that the dataset settings were also copied for each dataset. So They had the same color, same label, ... which obviously does not make sense.

Now only the `yAxisId` option is copied to the new dataset.

Task: [4577783](https://www.odoo.com/odoo/2328/tasks/4577783)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


rrahir:
@robodoo r+

PR #5776, opened 2025-02-20
[PERF] borders: avoid call to getCellBorder in setBorder

laa-odoo:
getCellBorder performs a deepCopy to avoid changing the reference to the border. But applied to a lot of data, deepCopy is heavy.

As in setBorder we are not reusing the result of
getCellBorder, we can afford to use a version without deepCopy of getCellBorder.

For the application of the borders of a sheet of 1000 rows and 1000 columns, measurements show a total time saving of 50%

Task: [4591697](https://www.odoo.com/odoo/2328/tasks/4591697)

### DIFF
--- a/src/plugins/core/borders.ts
+++ b/src/plugins/core/borders.ts
@@ -412,7 +412,7 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
       if (
         border?.left &&
         col > 0 &&
-        !deepEquals(this.getCellBorder({ sheetId, col: col - 1, row })?.right, border?.left)
+        !deepEquals(this.borders[sheetId]?.[col - 1]?.[row]?.right, border?.left)
       ) {
         this.history.update("borders", sheetId, col - 1, row, "right", undefined);
       }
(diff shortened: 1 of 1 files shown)


src/plugins/core/borders.ts
```diff
@@ -412,7 +412,7 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
       if (
         border?.left &&
         col > 0 &&
-        !deepEquals(this.getCellBorder({ sheetId, col: col - 1, row })?.right, border?.left)
+        !deepEquals(this.getRawCellBorder({ sheetId, col: col - 1, row })?.right, border?.left)
```
pro-odoo, line 415:
```suggestion
        !deepEquals(this.borders[sheetId]?.[col-1]?.[row]?.right, border?.left)
```
Even better, no ? Don't have to create a function that has to be named, required a comment on why we do not use deepCopy in this case, ...

What do you think ?

LucasLefevre:
can it be easily back-ported to 18.0 ? (not sure since Rémi refactored in the mean time)

src/plugins/core/borders.ts
laa-odoo: Yes of course,

why make it simple when you can make it complicated ;)

laa-odoo:
> can it be easily back-ported to 18.0 ? (not sure since Rémi refactored in the mean time)

in fact it is not suitable here

pro-odoo (approved):
robodoo r+

PR #5012, opened 2024-09-20
Master f2 composer rar

rrahir:
## Description:

description of this task, what is implemented and why it is implemented that way.

Task: [TASK_ID](https://www.odoo.com/web#id=TASK_ID&action=333&active_id=2328&model=project.task&view_type=form&cids=1&menu_id=4720)

## review checklist

- [ ] feature is organized in plugin, or UI components
- [ ] support of duplicate sheet (deep copy)
- [ ] in model/core: ranges are Range object, and can be adapted (adaptRanges)
- [ ] in model/UI: ranges are strings (to show the user)
- [ ] undo-able commands (uses this.history.update)
- [ ] multiuser-able commands (has inverse commands and transformations where needed)
- [ ] new/updated/removed commands are documented
- [ ] exportable in excel
- [ ] translations (\_t("qmsdf %s", abc))
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes
- [ ] doc is rebuild (npm run doc)
- [ ] status is correct in Odoo

PR #5777, opened 2025-02-20
[FIX] composer: broken right-to-left selection

hokolomopo:
## Description:

CONTAINS BACKPORT OF [#5494](https://github.com/odoo/o-spreadsheet/pull/5494)

Bug:
- Enter a formula "=A1+SUM(A2)"
- Put the cursor between the U and M
- Try to select from U to A
- The selection will reset upon mouseUp

Using `setEnd` in the contentEditableHelper does not work if the end
node is before the start node (in the DOM).

See https://developer.mozilla.org/en-US/docs/Web/API/Range/setEnd

Note that the bug didn't appear if the colors/classes of the
contentEditableHelper didn't change, because then the selection
wasn't reset and didn't need to be re-applied.

Task: [4592612](https://www.odoo.com/odoo/2328/tasks/)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/components/composer/content_editable_helper.ts
```diff
@@ -48,8 +52,16 @@ export class ContentEditableHelper {
...
-      range.setEnd(endNode.node, endNode.offset);
+
+      // setEnd (setStart) will result in a collapsed range if the end point is before the start point
+      // https://developer.mozilla.org/en-US/docs/Web/API/Range/setEnd
+      if (start <= end) {
+        range.setStart(startNode.node, startNode.offset);
+        range.setEnd(endNode.node, endNode.offset);
+      } else {
+        range.setStart(endNode.node, endNode.offset);
+        range.setEnd(startNode.node, startNode.offset);
```
hokolomopo, line 63: @rrahir Do you recall why you changed `extend` to `setEnd` [here](https://github.com/odoo/o-spreadsheet/pull/5012/files#diff-0c54baeea98218a3cf48db2bfac7a4cb31c4a8ee0406e314c08f996bf7524e73) ? I'm guessing you had a good reason

rrahir: I really dont. If I had to guess, because I could'nt make it work as I wanted in FF for some reason.. i'll have a look

tests/composer/composer_component.test.ts
```diff
@@ -878,6 +878,24 @@ describe("composer", () => {
...
+    composerEl = fixture.querySelector<HTMLElement>("div.o-composer")!;
+
+    composerStore.changeComposerCursorSelection(5, 5);
+    await nextTick();
+
+    composerStore.changeComposerCursorSelection(5, 2);
+    await nextTick();
+    const selection = document.getSelection()!;
+    await nextTick();
+    expect(selection?.toString()).toBe("1+S");
```
LucasLefevre, line 894: I didn't know `selection.toString()` !

LucasLefevre (commented):
robodoo r+

LucasLefevre:
robodoo rebase-ff

PR #5798, opened 2025-02-25
[FIX] Chart: fix data series reordering

anhe-odoo:
## Task Description

When trying to move data series for more than one position (above or below), we face a bug where the final order is not right. For example, with 3 dataseries (lets call them d1, d2 and d3), if the user tries to put d3 at the first position, he will end with [d2, d3, d1] instead of [d3, d1, d2].

## Related Task

- Task: [4600194](https://www.odoo.com/odoo/2328/tasks/4600194)


hokolomopo:
@robodoo r+

PR #5820, opened 2025-02-28
[FIX] HeaderOverlay: Fix unhide button position

rrahir:
The visibility of the button to unhide headers never properly accounted for the presence of the frozen pane. Furthermore, recent fixes in the viewport made the situation worse because the offset of the last header was badly computed when it was hidden.

Task: 4548264

## Description:

Task: [4548264](https://www.odoo.com/odoo/2328/tasks/4548264)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/headers_overlay/headers_overlay.xml
+++ b/src/components/headers_overlay/headers_overlay.xml
@@ -9,7 +9,7 @@
 
   <t t-name="o-spreadsheet-RowResizer" owl="1">
     <div
-      class="o-row-resizer"
+      class="o-row-resizer overflow-hidden"
       t-on-mousemove.self="onMouseMove"
       t-on-mouseleave="onMouseLeave"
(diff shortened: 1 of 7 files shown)


src/components/headers_overlay/headers_overlay.xml
```diff
@@ -9,7 +9,7 @@
 
   <t t-name="o-spreadsheet-RowResizer" owl="1">
     <div
-      class="o-row-resizer"
+      class="o-row-resizer overflow-hidden"
```
hokolomopo, line 12: overflow isn't needed with the t-if, no ?

hokolomopo (commented):
boup boup

hokolomopo:
@robodoo r+

PR #5824, opened 2025-03-03
[FIX] spreadsheet: handle inverse increment when autofilling upwards/left

Rachico:
Prior to this commit, autofilling upwards or to the left only correctly inverted the incrementation for numbers, but not when given a group of cells.  This led to incorrect results when autofilling mixed content.

The issue occurred because the rule generator always processed values in the same order, regardless of the autofill direction. When a single numeric value was present in a group of cells, the autofill logic failed to apply a negative increment when filling upwards.

To address this, the rule generator now considers the autofill direction.

Steps to reproduce:
1. Insert the following values in a column:
   - A10: 'test'
   - A11: 'test2'
   - A12: '4'
2. Select A10:A12 and autofill upwards.
3. Expected: A9 should be '3'. Before this fix, it incorrectly showed '5'.

Task: [4199568](https://www.odoo.com/odoo/2328/tasks/4199568)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


hokolomopo:
@robodoo r+

PR #5826, opened 2025-03-04
Master chart dashboard animation adrm

hokolomopo:
### [IMP] chart: improve funnel chart default animation

Change the default animation duration of funnel chart elements
from a constant of 1000ms to a scaling value, the smaller the
element the shorter the animation.

### [FIX] chart: fix funnel chart when animation are enabled

The funnel charts didn't work properly when chart animations were
enabled. The `width` property is not accessible during the animation,
we have to compute is manually from `x` and `base`.

Passing `nextElementWidth` at `updateElements` is also not enough,
since it will only be called at the first render, and not at each
animation frame. We can give the reference to the next element instead.

### [IMP] chart: unify `ChartJsComponent` methods arguments

This commit makes the `createChart` and `updateChart` methods of the
`ChartJsComponent` class have the same arguments so they are easier
to override.

Task: [4624050](https://www.odoo.com/odoo/2328/tasks/4624050)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/figures/chart/chartJs/chartjs.ts
+++ b/src/components/figures/chart/chartJs/chartjs.ts
@@ -68,7 +68,7 @@ export class ChartJsComponent extends Component<Props, SpreadsheetChildEnv> {
           this.chart?.destroy();
           this.createChart(deepCopy(runtime.chartJsConfig));
         } else {
-          this.updateChartJs(deepCopy(runtime));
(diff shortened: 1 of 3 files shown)


pro-odoo:
robodoo rebase-ff r+

PR #5830, opened 2025-03-05
[FIX] xlsx: correctly import labels from scatter charts

Rachico:
When importing an XLSX file containing a scatter chart into o-spreadsheet, the labels were not imported correctly. Instead, the actual range was used as labels by default, leading to incorrect visualization.

Steps to reproduce:
    1. Create an Excel file with a chart.
    2. Export the file as XLSX.
    3. Import the file into o-spreadsheet.
    4. The labels are not imported; instead, the data range is incorrectly used as labels.

Task: [4626572](https://www.odoo.com/odoo/2328/tasks/4626572)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


hokolomopo (commented):
Not the same behavior as Excel  :x 

When there is string data in the labels of a scatter chart, Excel ignores the labels and replace them by their indexes (1,2,3,4). When importing with your fix, we get the labels back. To a behavior closer to Excel, we should probably remove the label range ? Or maybe accept that our scatter chart don't behave the same as Excel and keep your fix ? Not sure ...

image
image

PR #5836, opened 2025-03-06
[FIX] Config: check stylesheets scope

rrahir:
The style that we define in our component can sometimes interfere within Odoo. This revision adds a step to force the style to be scoped to 'o-spreadsheet'.

Task: 4629441

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -25,10 +25,7 @@ if [ "$HUSKY_PRE_COMMIT" != 0 ]; then
             exec git diff --cached --color | grep -nE $consoleregexp -A 2 -B 2
             read -p "There are some occurrences of forbidden patterns at your modification. Are you sure want to continue? (y/n)" yn
             echo $yn | grep ^[Yy]$
-            if [ $? -eq 0 ]
-            then
-                exit 0; #THE USER WANTS TO CONTINUE
-            else
+            if [ $? -ne 0 ]; then
                 exit 1; # THE USER DONT WANT TO CONTINUE SO ROLLBACK
             fi
         else
@@ -36,4 +33,7 @@ if [ "$HUSKY_PRE_COMMIT" != 0 ]; then
             exec git diff --cached --color | grep -nE $consoleregexp -A 2 -B 2
         fi
     fi
+
+    # Css wrap check
+    node .husky/scss_check.js
 fi
--- a/.husky/scss_check.js
+++ b/.husky/scss_check.js
@@ -1,0 +1,48 @@
+import { execSync } from "child_process";
+import { existsSync, readFileSync } from "fs";
+import process from "process";
+
+const commentPattern = /\/\/.*|\/\*[\s\S]*?\*\//g;
+const firstLevelSelectorPattern = /^(?!(\s|\$|\})).+/gm;
+
+// Get scss files in diff
+const files = execSync("git diff --name-only --cached")
+  .toString()
+  .split("\n")
+  .filter((file) => file.endsWith(".scss"));
+
+const faultyFiles = [];
+
+for (const file of files) {
+  if (!existsSync(file)) continue;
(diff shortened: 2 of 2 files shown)


.husky/pre-commit
```diff
@@ -25,15 +25,40 @@ if [ "$HUSKY_PRE_COMMIT" != 0 ]; then
             exec git diff --cached --color | grep -nE $consoleregexp -A 2 -B 2
             read -p "There are some occurrences of forbidden patterns at your modification. Are you sure want to continue? (y/n)" yn
             echo $yn | grep ^[Yy]$
-            if [ $? -eq 0 ]
-            then
-                exit 0; #THE USER WANTS TO CONTINUE
-            else
+            if [ $? -ne 0 ]; then
```
rrahir, line 28: -ne  <-> not equal

.husky/pre-commit
```diff
@@ -25,15 +25,40 @@ if [ "$HUSKY_PRE_COMMIT" != 0 ]; then
...
             echo -e "\033[1;31mWARNING: Some occurences of forbidden patterns were detected \033[0m"
             exec git diff --cached --color | grep -nE $consoleregexp -A 2 -B 2
         fi
     fi
+
+    # Css wrap check
+    scssregexp='.*\.scss'
+    ignoreFileInstruction='\/\* husky-ignore \*\/'
+    filelist=()
+    for file in $(git diff --cached --name-only | grep -E $scssregexp); do
```
hokolomopo, line 41: check is not enought :x This is not flagged as invalid css

image

rrahir: I know, we discusswed this limitation yesterday? it's more about having a generic check so the dev stays alert rather than hafin a full parser just for the fun.. or else I need to make a js script because standard grep is not smart enough to get it done but it seemed not necessary. @LucasLefevre

hokolomopo: Grep can probably still do the job in this case I think. We could maybe have a regex `(startOfLine)(somethingThatIsn't.o-spreadshhet)` Will only work is the CSS is correctly prettified but w/e.

This should work ? 

`^(\.(?!o-spreadsheet)|[a-zA-Z]+).*\{`

rrahir: you need to handle multi line grep which does not work on macos for instance, so no it does not work.

rrahir: i'll try to do a correct js script whenever I have the time

.husky/scss_check.js
```diff
@@ -0,0 +1,46 @@
+import { execSync } from "child_process";
+import { existsSync, readFileSync } from "fs";
+import process from "process";
+
+const commentPattern = /\/\/.*|\/\*[\s\S]*?\*\//g;
+// const firstLevelSelectorPattern = /^(\.(?!o-spreadsheet\s)|[a-zA-Z]+).*\{/gm;
```
hokolomopo, line 6: comment

src/variables.scss
```diff
@@ -1 +1 @@
-$os-text-body: #374151;
```
hokolomopo, line 1: blue-da-ba-dee

.husky/scss_check.js
```diff
@@ -0,0 +1,46 @@
...
+Faulty files:
+
+${fileList}
+
+Your file should start with:
+.o-spreadsheet {
+    ...
+}
+
+and the selector should encompass all the style.\x1b[0m
```
hokolomopo, line 43: nitpick: not strictly what the condition is, we can have as many .o-spreadsheet selector as we want in the file

hokolomopo (commented):
Yo, listen up here's a story
About a little guy
That lives in a blue world

hokolomopo:
@robodoo r+

PR #5846, opened 2025-03-06
[FIX] pivot: sorting issue for (Undefined) columns in table

dhrp-odoo:
## Description:

Previously, sorting columns labeled as (Undefined) did not work correctly in the pivot table. The `isSortedColumnValid` function only checked whether a column was among the possible values, returning `false` for `null` values.

This commit resolves the issue by adding a check for `null` values in the `isSortedColumnValid` function.

Task: [4614177](https://www.odoo.com/odoo/2328/tasks/4614177)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


hokolomopo (commented):
Thanks for the PR

@robodoo r+

PR #5877, opened 2025-03-10
[FIX] chart: fix broken moving average trendline when adding another trendline

Rachico:
Prior to this commit, adding a trendline to a second dataset caused the moving average trendline of the first dataset to break. This issue occurred due to a misconfiguration in the trendline axis when multiple trendlines were applied.

Steps to reproduce
1. Create a chart with two datasets.
2. Add a moving average trendline to the first dataset.
   - The trendline displays correctly.
3. Add a trendline to the second dataset (any type).
   - The moving average trendline for the first dataset is now broken.

Task: [4632945](https://www.odoo.com/odoo/2328/tasks/4632945)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/figures/charts/chart_common.ts
+++ b/src/helpers/figures/charts/chart_common.ts
@@ -35,6 +35,7 @@ import { rangeReference } from "../../references";
 import { getZoneArea, isFullRow, toUnboundedZone, zoneToDimension, zoneToXc } from "../../zones";
 
 export const TREND_LINE_XAXIS_ID = "x1";
+export const MOVING_AVERAGE_TREND_LINE_XAXIS_ID = "xMovingAverage";
 
 /**
  * This file contains helpers that are common to different charts (mainly
@@ -446,4 +447,8 @@ export function getPieColors(colors: ColorGenerator, dataSetsValues: DatasetValu
   }
 
   return pieColors;
+}
+
+export function isTrendLineAxis(axisID: string) {
+  return axisID === TREND_LINE_XAXIS_ID || axisID === MOVING_AVERAGE_TREND_LINE_XAXIS_ID;
 }
--- a/src/helpers/figures/charts/runtime/chartjs_dataset.ts
+++ b/src/helpers/figures/charts/runtime/chartjs_dataset.ts
@@ -30,7 +30,12 @@ import {
   rgbaToHex,
   setColorAlpha,
 } from "../../../color";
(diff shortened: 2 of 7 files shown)


src/helpers/figures/charts/runtime/chartjs_dataset.ts
```diff
@@ -346,7 +346,7 @@ function getTrendingLineDataSet(
 
   return {
     type: "line",
-    xAxisID: TREND_LINE_XAXIS_ID,
+    xAxisID: config.type !== "trailingMovingAverage" ? TREND_LINE_XAXIS_ID : "x",
```
hokolomopo, line 349: At multiple place in the code we rely on `dataSet.xAxisID !== TREND_LINE_XAXIS_ID` to know whether the dataset is a trendline or not. This'll break all of those.  (show values, scatter chart, legend, tooltip, ...)

src/helpers/figures/charts/chart_common.ts
```diff
@@ -447,3 +448,7 @@ export function getPieColors(colors: ColorGenerator, dataSetsValues: DatasetValu
 
   return pieColors;
 }
+
+export function isTrendlineAxis(axisID: string) {
```
hokolomopo, line 452:
```suggestion
export function isTrendLineAxis(axisID: string) {
```
nitpick

src/helpers/figures/charts/chart_common.ts
```diff
@@ -35,6 +35,7 @@ import { rangeReference } from "../../references";
 import { getZoneArea, isFullRow, toUnboundedZone, zoneToDimension, zoneToXc } from "../../zones";
 
 export const TREND_LINE_XAXIS_ID = "x1";
+export const MOVING_AVERAGE_TREND_LINE_XAXIS_ID = "xMA";
```
hokolomopo, line 38:
```suggestion
export const MOVING_AVERAGE_TREND_LINE_XAXIS_ID = "xMovingAverage";
```
nitpick. That way it's easier to know what it does if we're ever debugging a chart runtime

hokolomopo (commented):
Might want to add a test :) 

Nitpicks otherwise, + a bug in the odoo pad

LucasLefevre (commented):
robodoo r+

PR #5878, opened 2025-03-10
[PERF] worksheet: Move PositionMap constant declarations outside of f…

laa-odoo:
…or loop in addRows

Previously, heavy PositionMap constants were declared within the for loop, leading to unnecessary overhead on each iteration. This change moves their declaration outside the loop to optimize performance.

Task: [4563258](https://www.odoo.com/odoo/2328/tasks/4563258)


LucasLefevre (approved):
robodoo r+

PR #5887, opened 2025-03-11
[IMP] sheet: remove useless translation

LucasLefevre:
This error is never supposed to be thrown and even less be shown to the end-user.

Translating the message is useless and adds useless work for translators.

Task: 0

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


hokolomopo (commented):
@robodoo r+

PR #5905, opened 2025-03-12
[IMP] spreadsheet_edition: move to user

fdamhaut:
Add a move to user when clicking on an user
on the "connected user" popover.

task-id: 4607572

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_feature/collaborative.ts
+++ b/src/plugins/ui_feature/collaborative.ts
@@ -1,10 +1,17 @@
 import { ClientDisconnectedError } from "../../collaborative/session";
 import { DEFAULT_FONT, DEFAULT_FONT_SIZE } from "../../constants";
 import { AlternatingColorGenerator } from "../../helpers";
-import { Client, ClientPosition, Color, GridRenderingContext, UID } from "../../types";
+import {
+  Client,
+  ClientId,
+  ClientPosition,
+  ClientWithPosition,
+  Color,
+  GridRenderingContext,
+} from "../../types";
 import { UIPlugin, UIPluginConfig } from "../ui_plugin";
 
-interface ClientToDisplay extends Required<Client> {
+interface ClientToDisplay extends ClientWithPosition {
   color: Color;
 }
 
@@ -12,12 +19,13 @@ export class CollaborativePlugin extends UIPlugin {
   static getters = [
     "getClientsToDisplay",
     "getClient",
+    "getCurrentClient",
     "getConnectedClients",
     "isFullySynchronized",
   ] as const;
   static layers = ["Selection"] as const;
   private availableColors = new AlternatingColorGenerator(12);
-  private colors: Record<UID, Color> = {};
+  private colors: Record<ClientId, Color> = {};
   private session: UIPluginConfig["session"];
 
   constructor(config: UIPluginConfig) {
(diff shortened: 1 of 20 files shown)


src/plugins/ui_feature/collaborative.ts
```diff
@@ -32,8 +42,53 @@ export class CollaborativePlugin extends UIPlugin {
...
+        this.showClientTag = true;
+        break;
+      case "HIDE_CLIENT_TAG":
+        this.showClientTag = false;
+        break;
+    }
+  }
+
+  jumpToClient(clientId: ClientId) {
+    const clients = this.getConnectedClients();
```
LucasLefevre, line 60: instead of looping over all clients
```suggestion
    const client = this.getClient(clientId);
    if (client.position) {
```

src/plugins/ui_feature/collaborative.ts
```diff
@@ -32,8 +42,53 @@ export class CollaborativePlugin extends UIPlugin {
...
+        this.dispatch("SCROLL_TO_CELL", { col: client.position.col, row: client.position.row });
+      }
+    }
+  }
+
+  focusClient(client: Client) {
+    if (client.focusTimeout) {
+      clearTimeout(client.focusTimeout);
+    }
+    client.focusTimeout = setTimeout(() => {
```
LucasLefevre, line 77: I'm a bit skeptical about this timeout not being cleared properly when we no longer needs the model.
It's asking for trouble.
As an example, it means all models survive for at least 3 seconds in tests.

Could you move all this code in a local store (where you can register a cleanup callback with `this.onDispose`) ?

rrahir (changes requested):
The commit name refered spreadsheet_edition but we are in o-spreadsheet ;-)

src/stores/client_focus_store.ts
```diff
@@ -0,0 +1,71 @@
...
+export class ClientFocusStore extends SpreadsheetStore {
+  mutators = [
+    "focusClient",
+    "unfocusClient",
+    "showClientTag",
+    "hideClientTag",
+    "jumpToClient",
+  ] as const;
+
+  private _showClientTag: Boolean = false;
```
LucasLefevre, line 14: `Boolean` => `boolean`, use the typescript literal type instead of the Javascript Boolean constructor.
And in this case, TS infers the type for you

src/plugins/ui_feature/collaborative.ts
```diff
@@ -32,12 +40,20 @@ export class CollaborativePlugin extends UIPlugin {
...
+    return this.session.getClient(clientId);
   }
 
-  getConnectedClients() {
-    return this.session.getConnectedClients();
+  getCurrentClient() {
+    return this.session.getCurrentClient();
+  }
+
+  getConnectedClients(): Set<Client> {
```
LucasLefevre, line 51: I don't see any reason (anymore) for which it should be a Set rather than a simple array

src/types/collaborative/session.ts
```diff
@@ -7,6 +8,11 @@ export interface Client {
   id: ClientId;
   name: string;
   position?: ClientPosition;
+  color?: Color;
```
LucasLefevre, line 11: I'm not sure why `color?` is needed here ? Everything builds without it.
If you remove it, `ClientWithPosition` can also be removed (and revert associated changes)

src/types/collaborative/session.ts
```diff
@@ -1,3 +1,4 @@
+import { Color } from "chart.js";
```
LucasLefevre, line 1: wrong import from chartjs

LucasLefevre (commented):
mainly small details before it can be merged

LucasLefevre (commented):
robodoo rebase-ff r+

PR #5906, opened 2025-03-13
[FIX] sheet: autoresize doesn't work on evaluated multiline cell

hokolomopo:
## Description

If a cell is a formula returning a multi-line content, the autoresize do nothing.

Task: [4609545](https://www.odoo.com/odoo/2328/tasks/4609545)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_feature/ui_sheet.ts
+++ b/src/plugins/ui_feature/ui_sheet.ts
@@ -1,7 +1,13 @@
-import { GRID_ICON_MARGIN, ICON_EDGE_LENGTH, PADDING_AUTORESIZE_HORIZONTAL } from "../../constants";
+import {
+  DEFAULT_CELL_HEIGHT,
+  GRID_ICON_MARGIN,
+  ICON_EDGE_LENGTH,
+  PADDING_AUTORESIZE_HORIZONTAL,
+} from "../../constants";
 import {
   computeIconWidth,
   computeTextWidth,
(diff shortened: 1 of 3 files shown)


src/plugins/ui_feature/ui_sheet.ts
```diff
@@ -170,4 +171,53 @@ export class SheetUIPlugin extends UIPlugin {
...
+        });
+      }
+    } else {
+      this.dispatch("RESIZE_COLUMNS_ROWS", {
+        elements: rows,
+        dimension: "ROW",
+        size: null,
+        sheetId,
+      });
+    }
```
LucasLefevre, line 221: you could also group rows by size to avoid dispatching once for every row in case there's an evaluated bazar to resize.

Something like
```ts
const groupedSizes = new Map<number | null, HeaderIndex[]>(rowSizes.map((size) => [size, []]));
for (let i = 0; i < rowSizes.length; i++) {
  groupedSizes.get(rowSizes[i])?.push(rows[i]);
}
for (const [size, rows] of groupedSizes) {
  this.dispatch("RESIZE_COLUMNS_ROWS", {
    elements: rows,
    dimension: "ROW",
    size,
    sheetId,
  });
}
```

LucasLefevre (commented):
robodoo r+

PR #5915, opened 2025-03-14
[FIX] clipboard: cross-sheet cut/paste is broken for tables/cfs

hokolomopo:
## Description:

Preface: the handling of `MOVE_RANGES` is broken in multiple plugins. When calling `adaptRanges`, we don't check that the resulting range is in the same sheet as the original range.

Fixing that in stable is probably not a good idea. This would mean that suddenly tables/cfs/merges could appear where they were previously not, and could break existing sheets.

With the clipboard refactoring, the `MOVE_RANGES` command is now dispatched at the start of the paste handling (in `CellClipboardHandler`). At this dispatch, the CFs/Tables/Merges are moved to the wrong sheet. So in a cut, when the time comes to delete the original Table, the table is not found at the copy position, and the delete fails.

This commit fixes that by dispatching the `MOVE_RANGES` command at the very end of the paste handling, after the tables/cfs/merges have been moved/deleted by their respective handlers.

Note: the cut/paste of filter values does not work anymore. It somewhat was working before because of the `MOVE_RANGES`, but doesn't work anymore since `UPDATE_FILTER` is not working before the next evaluation because of dynamic tables.

Task: [3905618](https://www.odoo.com/odoo/2328/tasks/3905618)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/clipboard_handlers/cell_clipboard.ts
+++ b/src/clipboard_handlers/cell_clipboard.ts
@@ -179,13 +179,6 @@ export class CellClipboardHandler extends AbstractCellClipboardHandler<
     this.clearClippedZones(content);
     const selection = target[0];
     this.pasteZone(sheetId, selection.left, selection.top, content.cells, options);
-    this.dispatch("MOVE_RANGES", {
-      target: content.zones,
-      sheetId: content.sheetId,
-      targetSheetId: sheetId,
-      col: selection.left,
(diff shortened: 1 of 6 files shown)


VincentSchippefilt:
robodoo r+

PR #5916, opened 2025-03-14
[FIX] clipboard: cross-sheet cut/paste is broken for merges

hokolomopo:
## Description

Preface: the handling of `MOVE_RANGES` is broken in multiple plugins. When calling `adaptRanges`, we don't check that the resulting range is in the same sheet as the original range.

Fixing that in stable is probably not a good idea. This would mean that suddenly cfs/merges could appear where they were previously not, and could break existing sheets.

The handling of copy/paste of merges is done either in `pasteMergeIfExist` (for copy/paste) or by `MOVE_RANGES` (for cut/paste). The latter is broken.

This commit fixes that by handling all the copy/cut cases of the merges in `pasteMergeIfExist` rather than relying on `MOVE_RANGES`. The `MOVE_RANGES` command is still dispatched, but it now do not affect merges since they are already deleted from their original location.

Task: [3905618](https://www.odoo.com/odoo/2328/tasks/3905618)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/clipboard/clipboard_cells_state.ts
+++ b/src/helpers/clipboard/clipboard_cells_state.ts
@@ -316,12 +316,7 @@ export class ClipboardCellsState extends ClipboardCellsAbstractState {
       for (let c = 0; c < width; c++) {
         const origin = rowCells[c];
         const position = { col: col + c, row: row + r, sheetId: sheetId };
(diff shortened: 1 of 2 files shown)


VincentSchippefilt:
robodoo up to 17.0 r+

PR #5932, opened 2025-03-17
[FIX] selection_input: fix range deletion

Rachico:
Prior to this commit, clicking the trash icon next to a range in DV/CF did not remove the range due to a missing event handler. This bug was introduced by this commit: 011e2237a56f0fa082d066a2aaea717b32d4c0e5

This commit ensures that the deletion action is correctly triggered, allowing a proper management of CF/DV rules

Steps to reproduce:
- Create a DV/CF rule
- Add a new range (valid or invalid)
- Click on the trash icon next to the range
- Nothing happens

Task: [4589387](https://www.odoo.com/odoo/2328/tasks/4589387)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/selection_input/selection_input.ts
+++ b/src/components/selection_input/selection_input.ts
@@ -224,8 +224,11 @@ export class SelectionInput extends Component<Props, SpreadsheetChildEnv> {
 
   removeInput(rangeId: number) {
     const index = this.store.selectionInputs.findIndex((range) => range.id === rangeId);
-    this.props.onSelectionRemoved?.(index);
-    this.props.onSelectionConfirmed?.();
+    if (this.ranges.find((range) => range.id === rangeId)?.xc) {
+      this.props.onSelectionRemoved?.(index);
+      this.props.onSelectionConfirmed?.();
+    }
+    this.store.removeRange(rangeId);
   }
 
   onInputChanged(rangeId: number, ev: InputEvent) {
--- a/tests/conditional_formatting/conditional_formatting_panel_component.test.ts
+++ b/tests/conditional_formatting/conditional_formatting_panel_component.test.ts
@@ -440,6 +440,54 @@ describe("UI of conditional formats", () => {
       );
     });
 
+    test("Can remove a valid, invalid or empty range", async () => {
+      await click(fixture, selectors.buttonAdd);
+      await nextTick();
+
+      await setInputValueAndTrigger(selectors.ruleEditor.editor.operatorInput, "BeginsWith");
+      setInputValueAndTrigger(".o-selection-input input", "A1:A4");
+      expect(document.querySelectorAll(".o-selection-input input")).toHaveLength(1);
+
+      // Add & remove a valid range
+      await simulateClick(".o-add-selection");
+      const range1 = document.querySelectorAll(".o-selection-input input")[1];
+      await setInputValueAndTrigger(range1, "B1:B4");
+      const remove1 = document.querySelectorAll(".o-remove-selection")[1];
+      await simulateClick(remove1);
+      expect(document.querySelectorAll(".o-selection-input input")).toHaveLength(1);
+
+      // Add & remove an invalid range
+      await simulateClick(".o-add-selection");
+      const range2 = document.querySelectorAll(".o-selection-input input")[1];
+      await setInputValueAndTrigger(range2, "MERA");
+      const remove2 = document.querySelectorAll(".o-remove-selection")[1];
+      await simulateClick(remove2);
+      expect(document.querySelectorAll(".o-selection-input input")).toHaveLength(1);
+
+      // Add & remove an empty range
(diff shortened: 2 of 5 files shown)


rrahir:
the commit hash you refer to doesn't belong to any active branch. Could you update it to the hash present in branch saas-18.2?

tests/conditional_formatting/conditional_formatting_panel_component.test.ts
```diff
@@ -440,6 +440,54 @@ describe("UI of conditional formats", () => {
...
+      const range1 = document.querySelectorAll(".o-selection-input input")[1];
+      await setInputValueAndTrigger(range1, "B1:B4");
+      const remove1 = document.querySelectorAll(".o-remove-selection")[1];
+      await simulateClick(remove1);
+      expect(document.querySelectorAll(".o-selection-input input")).toHaveLength(1);
+
+      // Add & remove an invalid range
+      await simulateClick(".o-add-selection");
+      const range2 = document.querySelectorAll(".o-selection-input input")[1];
+      await setInputValueAndTrigger(range2, "B1:MERA");
```
rrahir, line 462: a bit of nitpick but I'd rather have a stirng that explicitely says that the range is invalid like `invalid-range`  or `1:2:3` or even just `MERA` because the range you wrote here is invalid because it goes out out the bounds of the spreadsheet and not because it had a faulty syntax

tests/conditional_formatting/conditional_formatting_panel_component.test.ts
```diff
@@ -440,6 +440,54 @@ describe("UI of conditional formats", () => {
       );
     });
 
+    test("Can remove a valid, invalid or empty range", async () => {
+      await click(fixture, selectors.buttonAdd);
```
rrahir, line 444: Could you add a similar global test for all selection inputs in `selection_input_component.test.ts`?

src/components/selection_input/selection_input.ts
```diff
@@ -220,10 +220,14 @@ export class SelectionInput extends Component<Props, SpreadsheetChildEnv> {
 
   addEmptyInput() {
     this.store.addEmptyRange();
+    const rangeId = this.ranges[this.ranges.length - 1].id;
+    this.store.changeRange(rangeId, "");
+    this.triggerChange();
```
rrahir, line 225: no sure this is necesarry? the test pass without it

rrahir (changes requested):

src/components/selection_input/selection_input_store.ts
```diff
@@ -152,6 +152,8 @@ export class SelectionInputStore extends SpreadsheetStore {
     }
     this.insertNewRange(this.ranges.length, [""]);
     this.focusLast();
+    const rangeId = this.ranges[this.ranges.length - 1].id;
```
rrahir, line 155:
```suggestion
    const rangeId = this.ranges.at(-1).id;
```

src/components/selection_input/selection_input.xml
```diff
@@ -52,13 +52,13 @@
         <button class="o-button o-add-selection" t-if="canAddRange" t-on-click="addEmptyInput">
           Add range
         </button>
-        <div class="ms-auto" t-if="store.hasFocus">
+        <div class="ms-auto" t-if="store.hasMainFocus">
```
rrahir, line 55: is this change really necessary? it seems like an unrelated optimisation because `getFocus` *might* not be useful but not really realted to the issue at hand. If I missed something, of course I apologize but there is still a slight issue in that `hasMainFocus` is a private method of the store. it just happens not to be an issue when called in an xml template file but it's incorrect

rrahir (changes requested):

src/components/selection_input/selection_input_store.ts
Rachico: Why this change?

src/components/selection_input/selection_input.xml
Rachico: We had a bug with the focus of selection inputs on the grid; At some point, when you add many ranges, all the added ranges are focused.

Rachico: @rrahir Do you have any updates on this please 🙏 ?

rrahir: well, if it was the source of the issue, I would have expected to see it in previous version since it's been there since 18.0. If I rollback the change of store.hasFocus => sotre.hasMainFocus, I don't see the mentioned bug.

Rachico: After my latest updates, the issue doesn't appear to happening anymore. 
Therefore, I now removed the `hasMainFocus` fix. 
Can you please check? @rrahir

rrahir:
@robodoo r+

PR #5954, opened 2025-03-18
[FIX] xlsx: export reversed icon sets

LucasLefevre:
## Description:

With this commit, we export reversed icon sets to excel file.

Note: if icons from different types are in the set (arrows and smileys), we export the default icon set (based on the upper point) as excel doesn't support mixing icons from different types.

opw: 4237776
opw: 4241567

Task: [4655950](https://www.odoo.com/odoo/2328/tasks/4655950)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/xlsx/constants.ts
+++ b/src/xlsx/constants.ts
@@ -1,12 +1,13 @@
+import { IconSetType } from "../components/icons/icons";
 import { ExcelIconSet } from "../types/xlsx";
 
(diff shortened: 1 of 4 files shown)


hokolomopo:
@robodoo r+

PR #5973, opened 2025-03-19
[IMP] data: align data versions to release versions

LucasLefevre:
## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


tests/pivots/pivot_data.ts
```diff
@@ -2,555 +2,188 @@ import { toZone } from "../../src/helpers/index";
 
 export const pivotModelData = function (xc: string) {
   return {
-    version: 19,
+    version: 25,
```
LucasLefevre, line 5: not strictly necessary, but I realized it after it was done so... :man_shrugging:

src/migrations/data.ts
```diff
@@ -12,16 +12,23 @@ import {
...
- * function should be defined
- */
-export const CURRENT_VERSION = 25;
 const INITIAL_SHEET_ID = "Sheet1";
 
+export interface MigrationStep {
+  migrate: (data: any) => any;
+}
+
+export function getCurrentVersion() {
```
LucasLefevre, line 24: technically, it's not the "current" version if there was no changes in one saas. Should it be rather named `getLatestVersion` ? or is it more confusing and `getCurrentVersion` is clearer even if it's not strictly speaking 100% correct in all cases?

pro-odoo: Should we add an empty migration step if there is no changes in one saas version to keep consistency ?

pro-odoo (approved):
Very nice !
Just a little suggestion, not sure about it but IMO it would be good to have the `getCurrentVersion` matching the version of the branch

src/migrations/data.ts
LucasLefevre: In practice, I think we won't do it

pro-odoo: Except if we add it to the freeze checklist :D

pro-odoo:
Type in the first commit => `upgrade`

PR #5976, opened 2025-03-20
[FIX] color: `toHex` don't work with alpha = 0

hokolomopo:
## Description

When converting `rgba(0, 0, 0, 0)` to hex, the result should be `#00000000` but it was `#000000FF`. The code contained `alpha || 1`, which is wrong if the alpha is 0.

Task: [4660317](https://www.odoo.com/odoo/2328/tasks/4660317)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/color.ts
+++ b/src/helpers/color.ts
@@ -130,7 +130,10 @@ function rgbaStringToHex(color: Color): Color {
     throw new Error("invalid color");
   } else if (stringVals.length === 4) {
     const alpha = parseFloat(stringVals.pop() || "1");
-    alphaHex = Math.round((alpha || 1) * 255);
+    if (isNaN(alpha)) {
+      throw new Error("invalid alpha value");
+    }
+    alphaHex = Math.round(alpha * 255);
   }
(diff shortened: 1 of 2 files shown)


src/helpers/color.ts
```diff
@@ -130,7 +130,10 @@ function rgbaStringToHex(color: Color): Color {
     throw new Error("invalid color");
   } else if (stringVals.length === 4) {
     const alpha = parseFloat(stringVals.pop() || "1");
-    alphaHex = Math.round((alpha || 1) * 255);
+    if (isNaN(alpha)) {
```
rrahir, line 133: could we also add a check (or automatically clipping) that the value of alpha is between 0 and 1 ?

hokolomopo: `toHex` (that call this function) already checks that the resulting color is valid. I added some tests to make sure it works :)

rrahir: thanks :) I didn't check that far :'(

PR #5977, opened 2025-03-20
[FIX] filter: fix horrible performances with huge data filters

hokolomopo:
## Description

On a spreadsheet with 10,000 rows, opening the filter menu and filtering all the values is really slow. This commit improves the performance a bit.

On a data filter with 10,000 rows:
- `getFilterValues` (when opening filter menu): ~1000ms => ~70ms
- `updateHiddenRows` (when filtering all values): ~543ms => ~8ms

Note that the filter menu still takes a while to open and is laggy. It's because we create a HTML list with 10,000 elements, which is very slow. Fixing this would require big changes in the filter menu component, so it won't be done in stable.

Task: [4658998](https://www.odoo.com/odoo/2328/tasks/4658998)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/filters/filter_menu/filter_menu.ts
+++ b/src/components/filters/filter_menu/filter_menu.ts
@@ -174,26 +174,30 @@ export class FilterMenu extends Component<Props, SpreadsheetChildEnv> {
       position.col,
       position.row
     );
+    const normalizedFilteredValues = new Set(filterValues.map(toLowerCase));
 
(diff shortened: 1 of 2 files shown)


hokolomopo:
@robodoo help

hokolomopo:
@robodoo up to saas-18.3

PR #5991, opened 2025-03-21
[IMP] dashboard: tables are alive!

LucasLefevre:
## Description:

Currently, dashboards feel like static HTML pages from the early 2000s. Make dashboard feeling alive by adding hover effects on table lines.

Task: [4630549](https://www.odoo.com/odoo/2328/tasks/4630549)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/grid/hovered_cell_store.ts
+++ b/src/components/grid/hovered_cell_store.ts
@@ -13,7 +13,7 @@ export class HoveredCellStore extends SpreadsheetStore {
     }
   }
 
-  hover(position: Position) {
+  hover(position: Partial<Position>) {
     this.col = position.col;
     this.row = position.row;
   }
--- a/src/components/dashboard/dashboard.ts
+++ b/src/components/dashboard/dashboard.ts
@@ -56,10 +56,6 @@ export class SpreadsheetDashboard extends Component<Props, SpreadsheetChildEnv> 
     this.cellPopovers = useStore(CellPopoverStore);
   }
 
-  onCellHovered({ col, row }) {
-    this.hoveredCell.hover({ col, row });
-  }
-
   get gridContainer() {
     const sheetId = this.env.model.getters.getActiveSheetId();
     const { right } = this.env.model.getters.getSheetZone(sheetId);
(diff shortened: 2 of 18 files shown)


src/stores/cell_overlay_store.ts
pro-odoo, line 1: Remove empty file

src/components/grid/hovered_cell_store.ts
```diff
@@ -1,25 +1,59 @@
+import { setColorAlpha } from "../../helpers";
+import { PositionMap } from "../../plugins/ui_core_views/cell_evaluation/position_map";
 import { SpreadsheetStore } from "../../stores";
-import { Command, Position } from "../../types";
+import { Color, Command, Position } from "../../types";
 
 export class HoveredCellStore extends SpreadsheetStore {
-  mutators = ["clear", "hover"] as const;
+  mutators = ["clear", "hover", "debouncedHover"] as const;
```
pro-odoo, line 7: I'm not confortable with the distinction between `debouncedHover` and `hover `, what do you think of introducing a second store ?
- The first could be named something like `ImmediateHoveredCellStore`, and should be responsible to keep track of hovered cell when the mouse enter the cell.
- The second (could be named `DelayedHoveredCellStore`) would be responsible to track the hovered cell when the mouse stays in the cell ?

The two would expose `hover` mutator and `col` and `row` as accessor. It should be the role of the caller to know which store we would like to use.

What do you think ?

pro-odoo: Also, it's strange to have the computation of the overlays based on a table in such a generic store. Also, `col` and `row` seems not used outside the store. We could make a store `HoveredTableStore` instead of `ImmediateHoveredCellStore` which only expose `hoverCell` mutator and `overlayColors` accessor.

pro-odoo (changes requested):
Runbot is not happy :)

pro-odoo (approved):
robodoo rebase-ff r+

PR #6019, opened 2025-03-25
[REF] components: replace bootstrap collapse with custom collapse com…

pro-odoo:
…ponent

We currently use bootstrap.js to handle collapse elements. For now, we do not clean correctly the bootstrap instances, which leads to memory leaks.

Two solutions:
- Correctly clean bootstrap instances (by calling `.dispose()` on onWillUnmount)
- Implement our own Collapse component

We chose the option 2 to not rely on bootstrap.js anymore, one dependency less is better.

Task: 4671825

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -7,7 +7,6 @@
     <link rel="stylesheet" href="../build/o_spreadsheet.css" />
     <link rel="stylesheet" href="../node_modules/bootstrap/dist/css/bootstrap.min.css" />
     <link rel="stylesheet" href="../node_modules/font-awesome/css/font-awesome.min.css" />
-    <script src="../node_modules/bootstrap/dist/js/bootstrap.min.js"></script>
     <title>O-spreadsheet Dev</title>
   </head>
   <body>
--- a/src/components/composer/formula_assistant/formula_assistant.ts
+++ b/src/components/composer/formula_assistant/formula_assistant.ts
@@ -1,7 +1,8 @@
-import { Component } from "@odoo/owl";
+import { Component, useState } from "@odoo/owl";
 import { COMPOSER_ASSISTANT_COLOR } from "../../../constants";
 import { FunctionDescription, SpreadsheetChildEnv } from "../../../types";
 import { css } from "../../helpers/css";
+import { Collapse } from "../../side_panel/components/collapse/collapse";
 
 // -----------------------------------------------------------------------------
 // Formula Assistant component
@@ -50,6 +51,15 @@ export class FunctionDescriptionProvider extends Component<Props, SpreadsheetChi
     functionDescription: Object,
     argsToFocus: Array,
   };
+  static components = { Collapse };
+
+  private state: { isCollapsed: boolean } = useState({
+    isCollapsed: true,
+  });
+
+  toggle() {
+    this.state.isCollapsed = !this.state.isCollapsed;
+  }
 
   getContext(): Props {
     return this.props;
--- a/src/components/side_panel/components/collapse/collapse.ts
+++ b/src/components/side_panel/components/collapse/collapse.ts
@@ -1,0 +1,50 @@
+import { Component, onMounted, onWillUpdateProps, useRef } from "@odoo/owl";
+import { SpreadsheetChildEnv } from "../../../../types";
+
+interface Props {
+  isCollapsed: boolean;
+  slots: any;
+}
+
+export class Collapse extends Component<Props, SpreadsheetChildEnv> {
+  static template = "o-spreadsheet-Collapse";
+  static props = {
+    isCollapsed: Boolean,
+    slots: Object,
+  };
+
+  private contentRef = useRef("content");
+
+  setup() {
+    onMounted(() => {
+      if (this.props.isCollapsed) {
+        this.contentRef.el?.classList.add("d-none");
+      }
+    });
+    onWillUpdateProps((nextProps) => {
+      if (nextProps.isCollapsed !== this.props.isCollapsed) {
+        this.startTransition(nextProps.isCollapsed);
+      }
+    });
+  }
+
+  startTransition(isCollapsed: boolean) {
+    const el = this.contentRef.el;
+    if (!el) {
+      return;
+    }
+    el.classList.remove("d-none");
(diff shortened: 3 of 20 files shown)


pro-odoo:
image

pro-odoo:
*invoking css-wizard...*
@hokolomopo

src/components/side_panel/components/collapse/collapse.ts
```diff
@@ -0,0 +1,51 @@
...
+      }
+    });
+  }
+
+  setCollapsedState(open: boolean, skipTransition = false) {
+    const el = this.contentRef.el;
+    if (!el) return;
+
+    if (open) {
+      el.style.maxHeight = el.scrollHeight + "px";
```
hokolomopo, line 35: doesn't work at 100% :x Open the chart design panels and click on the button to add a trendline. `el.scrollHeight` is only valid at the moment of the transition, but this isn't updated when the content size is updated afterwards so it will overflow.

We probably could fix that by removing the max-height property on `transitionend` event ? Or a simple setTimeout.

src/components/side_panel/components/collapse/collapse.ts
```diff
@@ -0,0 +1,51 @@
...
+      el.style.overflow = "hidden";
+      el.style.transition = "max-height 0.3s ease";
+    } else {
+      el.style.maxHeight = "0px";
+      el.style.overflow = "hidden";
+      el.style.transition = "max-height 0.3s ease";
```
hokolomopo, line 41: nitpick: putting this in a class and in the css is a bit more elegant IMO

src/components/side_panel/components/collapse/collapse.ts
```diff
@@ -0,0 +1,51 @@
...
+    }
+    if (skipTransition) {
+      el.style.transition = "";
+    }
```
hokolomopo, line 45: I don't think this is useful ? Deleting it doesn't change anything that I can see. On mounted we go from a state of no max-height to a state with a max-height, there's no css transition to be run there.

src/components/side_panel/components/collapse/collapse.ts
hokolomopo: We should maybe keep bootstrap approach. A `d-none` when it's collapsed, nothing when it's not collapsed, when starting a collapse manually set the size to 0/scrollHeight, add a class/style with `transition`, set the target height, and at the end of the transition remove everything (I think that's what they do)

hokolomopo (commented):
you-shall-not-pass-lotr

Me stopping buggy CSS from reaching the codebase 🧙

src/components/composer/formula_assistant/formula_assistant.ts
```diff
@@ -50,6 +51,15 @@ export class FunctionDescriptionProvider extends Component<Props, SpreadsheetChi
     functionDescription: Object,
     argsToFocus: Array,
   };
+  static components = { Collapse };
+
+  private state: { opened: boolean } = useState({
```
VincentSchippefilt, line 56: I would rename the boolean to isOpen. It's always clearer in my mind when booleans start by is... :D

src/components/side_panel/components/collapse/collapse.ts
```diff
@@ -0,0 +1,51 @@
...
+
+interface Props {
+  open: boolean;
+  slots: any;
+}
+
+export class Collapse extends Component<Props, SpreadsheetChildEnv> {
+  static template = "o-spreadsheet-Collapse";
+  static props = {
+    open: Boolean,
```
VincentSchippefilt, line 12: same here, isOpen

src/components/side_panel/components/collapsible/side_panel_collapsible.ts
```diff
@@ -46,6 +39,13 @@ export class SidePanelCollapsible extends Component {
     collapsedAtInit: { type: Boolean, optional: true },
     class: { type: String, optional: true },
   };
+  static components = { Collapse };
+
+  private state: { opened: boolean } = useState({
+    opened: !this.props.collapsedAtInit,
```
VincentSchippefilt, line 45: isInitiallyCollapsed would be a better name

src/components/side_panel/components/collapse/collapse.ts
LucasLefevre, line 35: why not using `t-att-style` instead of fiddling with the DOM ?

src/components/side_panel/components/collapse/collapse.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  private contentRef = useRef("content");
+
+  setup() {
+    onMounted(() => {
+      if (this.props.isCollapsed) {
+        this.contentRef.el?.classList.add("d-none");
+      }
+    });
+    onWillUpdateProps((nextProps) => {
```
LucasLefevre, line 24: I'm not sure `onWillUpdateProps` is appropriate because it starts the transition *before* the component is rendered. The element could technically be changed (even it probably never happens in practice here)

demo/index.html
```diff
@@ -7,7 +7,6 @@
     <link rel="stylesheet" href="../build/o_spreadsheet.css" />
     <link rel="stylesheet" href="../node_modules/bootstrap/dist/css/bootstrap.min.css" />
     <link rel="stylesheet" href="../node_modules/font-awesome/css/font-awesome.min.css" />
-    <script src="../node_modules/bootstrap/dist/js/bootstrap.min.js"></script>
```
LucasLefevre, line 10: :partying_face:

LucasLefevre:
it's very red :red_circle: https://runbot206.odoo.com/runbot/static/build/77308820-master/logs/spreadsheet.txt

LucasLefevre:
It's orange :orange_circle: but I trust it'll be green :green_circle: 

robodoo r+

pro-odoo:
> it's very red 🔴 https://runbot206.odoo.com/runbot/static/build/77308820-master/logs/spreadsheet.txt

It's now very green 🟢

PR #6046, opened 2025-03-27
[IMP] chart: add zoom interactivity

anhe-odoo:
## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs.ts
```diff
@@ -0,0 +1,476 @@
...
+    const { left, right } = this.slider.chartArea;
+    if (x < left) {
+      return left;
+    } else if (x > right) {
+      return right;
+    }
+    return x;
+  }
+
+  private clipDeltaInsideChart(delta: number): number | undefined {
```
LucasLefevre, line 214: can you use the existing `clip` helper ?

src/types/chart/chart.ts
```diff
@@ -203,3 +203,8 @@ export type GenericDefinition<T extends ChartWithDataSetDefinition> = Partial<
 > & {
   dataSets?: Omit<T["dataSets"][number], "dataRange">[];
 };
+
+export interface ZoomConfiguration {
+  readonly enabled?: boolean;
+  readonly sliceable?: boolean;
```
LucasLefevre, line 209: I'm not sure to understand what `sliceable` is. In which situation is a chart zoomable but not sliceable ?

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs.ts
```diff
@@ -0,0 +1,476 @@
...
+import { isCtrlKey } from "../../../../helpers/dom_helpers";
+import { chartJsExtensionRegistry, getChartJSConstructor } from "../chart_js_extension";
+import { ChartJsComponent } from "../chartjs";
+import { currentlyShownArea, hoveredPosition } from "./zoomable_chartjs_plugins";
+
+chartJsExtensionRegistry
+  .add("currentlyShownArea", currentlyShownArea)
+  .add("hoveredPosition", hoveredPosition);
+
+export class ZoomableChartJsComponent extends ChartJsComponent {
```
LucasLefevre, line 17: would it be possible to move as much as possible the business logic to a store ? It would be infinitely easier to write tests, and also force us to decouple as much as possible from chartjs specifics.

src/components/side_panel/chart/index.ts
```diff
@@ -50,7 +51,7 @@ chartSidePanelComponentRegistry
     design: ChartWithAxisDesignPanel,
   })
   .add("combo", {
-    configuration: GenericChartConfigPanel,
+    configuration: ComboConfigPanel,
```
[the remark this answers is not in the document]
anhe-odoo: Not sure, as the only "new" config panel is for the ComboBox, so it will be introducing this one only for one chart type. What do you think ?

src/types/chart/chart.ts
anhe-odoo: `enabled` means the button to show/hide the slicer is visible. This is something like "allow the user to interact with the chart".
`sliceable` means the slicer is visible.
`wheelable` means we can CTRL+scroll to zoom directly in the chart (only when enabled is true and the axis is a linear axis)

src/components/figures/figure/figure.xml
```diff
@@ -17,6 +17,19 @@
           onFigureDeleted="props.onFigureDeleted"
           figureUI="props.figureUI"
         />
+        <div
+          t-if="zoomEnabled"
+          class="o-figure-menu position-absolute m-2"
+          style="margin-right:2.5rem!important">
+          <div class="d-flex">
+            <div class="o-figure-menu-item mx-1" t-on-click="resetZoom">
+              <t t-call="o-spreadsheet-Icon.HOME"/>
```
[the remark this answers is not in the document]
anhe-odoo: I totally agree, but (at the moment) the "is zoomed" state is something that is only known in the ZoomableChartJs component, and there is no ways to get the information in the figure as it's not stored in any definition or runtime :/ 
I can add something to get the information, but I'm not sure it will be clean :/

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chart_store.ts
```diff
@@ -0,0 +1,99 @@
+import { Command, UID } from "../../../../..";
+import {
+  MOVING_AVERAGE_TREND_LINE_XAXIS_ID,
+  TREND_LINE_XAXIS_ID,
+} from "../../../../../helpers/figures/charts/chart_common";
+import { SpreadsheetStore } from "../../../../../stores";
+
+export const __ZOOMABLE_AXIS_IDS__ = [
```
LucasLefevre:
```suggestion
const ZOOMABLE_AXIS_IDS = [
```

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chart_store.ts
```diff
@@ -0,0 +1,99 @@
...
+      const newXMin = this.currentAxisLimits[chartId]?.min;
+      const newXMax = this.currentAxisLimits[chartId]?.max;
+      this.trendLineAxisLimits[chartId][axisId].min =
+        newXMin !== undefined ? (newXMin as number) * slope + intercept : undefined;
+      this.trendLineAxisLimits[chartId][axisId].max =
+        newXMax !== undefined ? (newXMax as number) * slope + intercept : undefined;
+    }
+  }
+
+  cancel() {}
```
LucasLefevre: :eyes:

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs.ts
```diff
@@ -0,0 +1,521 @@
...
+    `;
+  }
+
+  get sliceable(): boolean {
+    const runtime = this.chartRuntime;
+    const zoomPlugin = runtime.chartJsConfig.options?.plugins?.zoom;
+    if (!zoomPlugin?.enabled || !zoomPlugin.sliceable) {
+      return false;
+    }
+    return true;
```
LucasLefevre:
```suggestion
    return zoomPlugin?.enabled && zoomPlugin.sliceable;
```

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chart_store.ts
```diff
@@ -0,0 +1,97 @@
...
+export const __ZOOMABLE_AXIS_IDS__ = [
+  "x",
+  TREND_LINE_XAXIS_ID,
+  MOVING_AVERAGE_TREND_LINE_XAXIS_ID,
+] as const;
+export type AxisId = (typeof __ZOOMABLE_AXIS_IDS__)[number];
+export type AxesLimits = {
+  [chartId: UID]: { [axisId in AxisId]: { min: number | undefined; max: number | undefined } };
+};
+export type ScaleLimits = { [chartId: UID]: { min: number | undefined; max: number | undefined } };
```
hokolomopo: nitpick: those don't need to be exported

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chart_store.ts
```diff
@@ -0,0 +1,97 @@
+import { Command, UID } from "../../../../..";
+import {
+  MOVING_AVERAGE_TREND_LINE_XAXIS_ID,
+  TREND_LINE_XAXIS_ID,
+} from "../../../../../helpers/figures/charts/chart_common";
+import { SpreadsheetStore } from "../../../../../stores";
+
+export const __ZOOMABLE_AXIS_IDS__ = [
```
hokolomopo:
```suggestion
const ZOOMABLE_AXIS_IDS = [
```

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chart_store.ts
```diff
@@ -0,0 +1,97 @@
...
+        break;
+    }
+  }
+
+  resetAxisLimits(
+    chartId: UID,
+    limits: { [key: string]: { min: number | undefined; max: number | undefined } }
+  ) {
+    for (const axisId of __ZOOMABLE_AXIS_IDS__) {
+      if (limits[axisId]) {
```
hokolomopo:
```suggestion
    for (const axisId in limits) {
```
no ?

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chart_store.ts
```diff
@@ -0,0 +1,97 @@
...
+  updateTrendLineConfiguration(chartId: UID) {
+    if (!this.originalAxisLimits[chartId]) {
+      return;
+    }
+    const chartLimits = this.originalAxisLimits[chartId].x;
+    if (chartLimits.min === undefined || chartLimits.max === undefined) {
+      return;
+    }
+    for (const axisId of __ZOOMABLE_AXIS_IDS__) {
+      if (axisId === "x" || !this.originalAxisLimits[chartId][axisId]) {
```
hokolomopo: just loop on all trend axis, instead of all axis then skipping the normal x axis

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chart_store.ts
```diff
@@ -0,0 +1,97 @@
...
+  [chartId: UID]: { [axisId in AxisId]: { min: number | undefined; max: number | undefined } };
+};
+export type ScaleLimits = { [chartId: UID]: { min: number | undefined; max: number | undefined } };
+
+export class ZoomableChartStore extends SpreadsheetStore {
+  mutators = ["resetAxisLimits", "updateAxisLimits", "updateTrendLineConfiguration"] as const;
+
+  originalAxisLimits: AxesLimits = {};
+  currentAxisLimits: ScaleLimits = {};
+  trendLineAxisLimits: AxesLimits = {};
```
hokolomopo: can't this be the same object ? currentAxisLimit should be a Record<ChartId, AxisId, Limits> no ? The current axis is always "x"

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chart_store.ts
```diff
@@ -0,0 +1,97 @@
...
+      const trendingLimits = this.originalAxisLimits[chartId][axisId];
+      const trendingRange = trendingLimits.max! - trendingLimits.min!;
+      const slope = trendingRange / realRange;
+      const intercept = trendingLimits.min! - chartLimits.min * slope;
+      const newXMin = this.currentAxisLimits[chartId]?.min;
+      const newXMax = this.currentAxisLimits[chartId]?.max;
+      this.trendLineAxisLimits[chartId][axisId].min =
+        newXMin !== undefined ? (newXMin as number) * slope + intercept : undefined;
+      this.trendLineAxisLimits[chartId][axisId].max =
+        newXMax !== undefined ? (newXMax as number) * slope + intercept : undefined;
```
hokolomopo: a small comment explaining what this shmilblick does would be nice

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chart_store.ts
hokolomopo: are there situations where min is defined and not max ? if not, typing as `{min:number, max:number} | undefined` would probably make your life simpler

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs.ts
```diff
@@ -0,0 +1,556 @@
...
+        },
+        layout: {
+          padding: {
+            ...chartData.options.layout?.padding,
+            top: 5,
+            bottom: 10,
+          },
+        },
+        scales: {
+          ...chartData.options.scales,
```
hokolomopo: should we spread ? The only thing it does is copying the trend line axis, which we don't want anyway, no ?

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs.ts
```diff
@@ -0,0 +1,556 @@
...
+                size: 9,
+              },
+            },
+          },
+        },
+      },
+    };
+  }
+
+  private getDetailChartConfiguration(chartData: ChartConfiguration<any>): ChartConfiguration<any> {
```
hokolomopo, line 98: detail ? isn't it zoomedChartConfig or something like that ?

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs.ts
```diff
@@ -0,0 +1,556 @@
...
+    const xAxis = this.store.currentAxisLimits[this.props.figureUI.id];
+    let xScale = {
+      ...chartData.options.scales?.x,
+    };
+    if (xAxis?.min !== undefined) {
+      xScale = {
+        ...xScale,
+        min: this.hasLinearScale ? xAxis.min : Math.ceil(xAxis.min) - this.axisOffset,
+      };
+    }
```
hokolomopo:
```suggestion
    const xScale = {...chartData.options.scales?.x};
    if (xAxis?.min !== undefined) {
      xScale.min = this.hasLinearScale ? xAxis.min : Math.ceil(xAxis.min) - this.axisOffset; 
      }
```
simpler, no ?

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs.ts
```diff
@@ -0,0 +1,556 @@
...
+    };
+  }
+
+  protected createChart(chartData: ChartConfiguration<any>) {
+    const allowAnimation =
+      this.env.model.getters.isDashboard() &&
+      this.animationStore &&
+      chartData.options?.plugins?.zoom?.enabled;
+    this.isBarChart = chartData.type === "bar";
+    this.store.updateAxisLimits(this.props.figureUI.id, { min: undefined, max: undefined });
```
hokolomopo: it it needed ? this will reset the zoom each time the chart is scrolled out/in of the screen

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs.ts
```diff
@@ -0,0 +1,556 @@
...
+    let updatedData = this.getDetailChartConfiguration(chartData);
+    if (allowAnimation) {
+      updatedData = this.enableAnimationInChartData(updatedData);
+      this.animationStore!.enableAnimationForChart(this.animationFigureId);
```
hokolomopo:
```suggestion
      this.animationStore?.enableAnimationForChart(this.animationFigureId);
```
avoid `!` when possible

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs.ts
```diff
@@ -0,0 +1,556 @@
...
+    }
+    super.createChart(updatedData);
+    this.hasLinearScale = this.chart!.scales?.x.type === "linear";
```
hokolomopo:
```suggestion
    this.hasLinearScale = this.chart?.scales?.x.type === "linear";
```

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs.ts
```diff
@@ -0,0 +1,556 @@
...
+      this.animationStore &&
+      chartData.options?.plugins?.zoom?.enabled;
+    this.isBarChart = chartData?.type === "bar";
+    let updatedData = this.getDetailChartConfiguration(chartData);
+    if (allowAnimation) {
+      updatedData = this.enableAnimationInChartData(updatedData);
+      this.animationStore!.enableAnimationForChart(this.animationFigureId);
+    }
+    super.updateChartJs(updatedData);
+    this.hasLinearScale = this.chart!.scales?.x.type === "linear";
```
hokolomopo:
```suggestion
    this.hasLinearScale = this.chart?.scales?.x.type === "linear";
```

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs.ts
hokolomopo: won't that break original animation conditions ? the chart will be animated on each render, no matter if the data has changed or not now ... and every time the chart appear on the screen

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs.xml
```diff
@@ -0,0 +1,24 @@
...
+          t-on-pointermove="onPointerMoveInDetailChart"
+        />
+      </div>
+      <div t-if="sliceable" class="o-chart-zoomable-slicer">
+        <canvas
+          class="o-figure-canvas o-zoom-slicer w-100 h-100"
+          t-ref="graphSlider"
+          t-on-pointerdown="onPointerDownInMasterChart"
+          t-on-pointermove="onPointerMoveInMasterChart"
+          t-on-mouseleave="onMouseLeaveMasterChart"
```
hokolomopo: Names are a bit confusing. I'd rename to MainChart and SlicerChart or something like that

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs.ts
```diff
@@ -0,0 +1,556 @@
...
+    const target = ev.target;
+    if (!target) {
+      return;
+    }
+    target["style"].cursor = "default";
+    this.currentHoverX = undefined;
+    this.chart?.update();
+  }
+
+  onPointerDownInDetailChart(ev: PointerEvent) {
```
hokolomopo: ctrl+click doesn't feel very good to use and is hard to discover. I'd agree with lucas and remove it

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs.ts
```diff
@@ -0,0 +1,556 @@
...
+        ev.stopPropagation();
+        ev.preventDefault();
+        ev.target!["style"].cursor = "move";
+      } else {
+        ev.target!["style"].cursor = "default";
+      }
+    }
+  }
+
+  onWheelOnDetailChart(ev: WheelEvent) {
```
hokolomopo: wasn't the zoom plugin removed ? wheel doesn't seem to do anything

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs_plugins.ts
```diff
@@ -0,0 +1,101 @@
+import { ChartType, Plugin } from "chart.js";
+
+// Plugin to show the current visible area on master chart
```
hokolomopo: should be a docstring on the plugin. And TBH reading this doesn't really help me understand what the plugin does.

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs_plugins.ts
```diff
@@ -0,0 +1,101 @@
...
+    ctx.moveTo(upperBound + 3, (top + bottom) / 2 - 7);
+    ctx.lineTo(upperBound + 3, (top + bottom) / 2 + 7);
+    ctx.stroke();
+    ctx.lineWidth = originalLineWidth;
+  },
+};
+
+window.Chart?.register(currentlyShownArea);
+
+// Plugin to show the current hover position on detail chart
```
hokolomopo: same here, should be doctring. And I double down on my comment saying that"detail chart" and "master chart" aren't the best names

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs_plugins.ts
hokolomopo: isn't it `SlidingWindowPlugin` or something ?

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs_plugins.ts
```diff
@@ -0,0 +1,101 @@
+import { ChartType, Plugin } from "chart.js";
+
+// Plugin to show the current visible area on master chart
+interface CurrentAreaPluginOptions {
+  getLowerBound: () => number | undefined;
+  getUpperBound: () => number | undefined;
```
hokolomopo, line 11: these could be just numbers instead of functions, no ?

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs_plugins.ts
```diff
@@ -0,0 +1,101 @@
...
+  id: "currentlyShownArea",
+  afterDatasetsDraw: function (chart, args, options: CurrentAreaPluginOptions) {
+    if (!options.getLowerBound || !options.getUpperBound) {
+      return;
+    }
+    const {
+      ctx,
+      chartArea: { left, right, top, bottom },
+    } = chart;
+    const originalLineWidth = ctx.lineWidth;
```
hokolomopo: why are you only restoring the lienwidth ? and not the fillstyle/strokestyle ? use ctx.save()/ctx.restore()

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs_plugins.ts
```diff
@@ -0,0 +1,101 @@
...
+    ctx.stroke();
+    ctx.beginPath();
+    ctx.moveTo(upperBound + 3, (top + bottom) / 2 - 7);
+    ctx.lineTo(upperBound + 3, (top + bottom) / 2 + 7);
+    ctx.stroke();
+    ctx.lineWidth = originalLineWidth;
+  },
+};
+
+window.Chart?.register(currentlyShownArea);
```
hokolomopo: handled in `chartJsExtensionRegistry`

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs_plugins.ts
```diff
@@ -0,0 +1,101 @@
...
+        : left + ((right - left) * (hoverCoordinate - min + 0.5)) / (max + 1 - min);
+    ctx.beginPath();
+    ctx.strokeStyle = "black";
+    ctx.moveTo(hoverPosition, top);
+    ctx.lineTo(hoverPosition, bottom);
+    ctx.stroke();
+  },
+};
+
+window.Chart?.register(hoveredPosition);
```
hokolomopo: handled in chartJsExtensionRegistry

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs_plugins.ts
```diff
@@ -0,0 +1,101 @@
...
+    }
+    const { min, max, type } = chart.scales.x;
+    if (hoverCoordinate < min || hoverCoordinate > max) {
+      return;
+    }
+    const {
+      ctx,
+      chartArea: { left, right, top, bottom },
+    } = chart;
+    const hoverPosition =
```
hokolomopo: you should round the position, the bar changes size because it it sometime blurry, sometime not

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs_plugins.ts
```diff
@@ -0,0 +1,101 @@
...
+    ctx.strokeStyle = "black";
+    ctx.moveTo(hoverPosition, top);
+    ctx.lineTo(hoverPosition, bottom);
+    ctx.stroke();
+  },
+};
+
+window.Chart?.register(hoveredPosition);
+
+interface MainZoomPluginOptions {
```
hokolomopo: zoom not used anymore ?

src/components/figures/chart/chart_dashboard_menu/chart_dashboard_menu.ts
```diff
@@ -147,4 +154,51 @@ export class ChartDashboardMenu extends Component<Props, SpreadsheetChildEnv> {
...
+      items.push({
+        id: "resetZoom",
+        label: _t("Reset zoom"),
+        iconClass: "fa fa-refresh",
+        onClick: () => {
+          this.chartZoomStore.updateAxisLimits(
+            figureId,
+            this.chartZoomStore.originalAxisLimits[figureId].x
+          );
+          this.env.model.dispatch("EVALUATE_CHARTS");
```
hokolomopo: it it needed ? an update to the store will trigger a render, no ?

src/components/figures/figure/figure.ts
```diff
@@ -160,7 +170,14 @@ export class FigureComponent extends Component<Props, SpreadsheetChildEnv> {
...
   get wrapperStyle() {
-    const { x, y, width, height } = this.props.figureUI;
+    let { x, y, width, height, tag } = this.props.figureUI;
+    if (tag === "chart") {
+      const figureId = this.props.figureUI.id;
+      const definition = this.env.model.getters.getChartDefinition(figureId);
+      if ("zoom" in definition && definition.zoom?.enabled && definition.zoom?.sliceable) {
+        height += CHART_ZOOM_SLIDER_HEIGHT; // Add space for zoom controls
+      }
+    }
```
hokolomopo: figure shouldn't be aware of chart stuff >_< Can't the button that enable the zoom dispatch an UPDATE_FIGURE ?

src/components/figures/figure/figure.ts
```diff
@@ -332,6 +351,62 @@ export class FigureComponent extends Component<Props, SpreadsheetChildEnv> {
...
+    if ("zoom" in definition) {
+      zoom = {
+        ...definition.zoom,
+        sliceable: !definition.zoom?.sliceable,
+      };
+    }
+    this.env.model.dispatch("UPDATE_CHART", {
+      figureId,
+      // @ts-ignore
+      definition: { ...definition, zoom },
```
hokolomopo: bad sign when you have to ts-ignore :upside_down_face: 

yeah all of this stuff should not be in figure.ts . Probably the chart component should begin to handle it's own menu ? It already does in dashboard, maybe we could always use that menu ?

src/components/side_panel/chart/bar_chart/bar_chart_config_panel.ts
```diff
@@ -16,4 +16,18 @@ export class BarConfigPanel extends GenericChartConfigPanel {
...
+
+  onToggleZoom(enabled: boolean) {
+    const definition = this.props.definition as BarChartDefinition;
+    const zoom = {
+      ...definition.zoom,
+      enabled,
+    };
+    if (enabled) {
+      zoom.sliceable = true;
+    }
```
hokolomopo: ther's no need for both zoom.sliceable and zoom.enable no ? And if there's a need for both, why is there a single checkbox to handle two booleans ?

src/components/side_panel/chart/combo_chart/combo_chart_config_panel.xml
```diff
@@ -0,0 +1,32 @@
...
+        onSelectionConfirmed.bind="onDataSeriesConfirmed"
+        onSelectionReordered.bind="onDataSeriesReordered"
+        onSelectionRemoved.bind="onDataSeriesRemoved"
+      />
+      <ChartLabelRange
+        range="this.getLabelRange()"
+        isInvalid="isLabelInvalid"
+        onSelectionChanged.bind="onLabelRangeChanged"
+        onSelectionConfirmed.bind="onLabelRangeConfirmed"
+        options="this.getLabelRangeOptions()"
```
hokolomopo: shouldn't the checkbox be with the other label range options ? Or in the design panel ?

src/helpers/figures/charts/bar_chart.ts
```diff
@@ -243,6 +249,7 @@ export function createBarChartRuntime(chart: BarChart, getters: Getters): BarCha
         legend: getBarChartLegend(definition, chartData),
         tooltip: getBarChartTooltip(definition, chartData),
         chartShowValuesPlugin: getChartShowValues(definition, chartData),
+        zoom: getChartZoom(definition, chartData),
```
hokolomopo: is it useful to put something in chartData ?

src/helpers/figures/charts/runtime/index.ts
```diff
@@ -1,6 +1,5 @@
 export * from "./chart_data_extractor";
 export * from "./chartjs_dataset";
-export * from "./chartjs_layout";
 export * from "./chartjs_legend";
```
hokolomopo: why ? This breaks odoo if the helpers are not exported ...

src/types/chart/chart.ts
hokolomopo: Oh that what it was. Is there really a point to have both tho ? Do we have a need for a chart where we _could_ zoom but the slider is not visible until we click a button ? IMO the there should be a single boolean that show/hide the slider

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs.xml
anhe-odoo: Master and Detail chart is an official term for this kind of chart, so I think it's better to use it. But I could try to make it more logical and avoid having "slicer" at some places then

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs_plugins.ts
anhe-odoo: No, the configuration is set when creating the chart, but the value can be changed with every manipulations of the slicer and the chart is then updated with chart.update, so the config will be unchanged

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chart_store.ts
anhe-odoo: Updated with an else clause no that limits can be undefined too :)

src/components/figures/chart/chart_dashboard_menu/chart_dashboard_menu.ts
anhe-odoo: No, the rendering isn't handled by the store, as it doesn't have access to the chart object, so cannot call chart.update()

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs_plugins.ts
anhe-odoo: It is. I only declared it here to be able to add the zoom configuration to the chart config. Maybe not a good idea ?

src/components/side_panel/chart/bar_chart/bar_chart_config_panel.ts
anhe-odoo: Because the idea of the zoom.enable if that the dashboard maker decide that this chart should be zoomable or not in the dashboard. The zoom.sliceable is there when the dashboard/spreadsheet decide to show the slicer. So there is one checkbox to handle zoom.enable, and one button on the chart to show the slicer

src/components/side_panel/chart/combo_chart/combo_chart_config_panel.xml
anhe-odoo: TBC but this has been discussed with FRGI

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chart_store.ts
```diff
@@ -0,0 +1,124 @@
...
+  mutators = [
+    "resetAxisLimits",
+    "updateAxisLimits",
+    "updateTrendLineConfiguration",
+    "clearAxisLimits",
+  ] as const;
+
+  originalAxisLimits: AxesLimits = {};
+  currentAxesLimits: AxesLimits = {};
+  idConversion: Record<UID, UID> = {};
```
LucasLefevre: I don't understand the purpose of `idConversion`

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chart_store.ts
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  /* Update the trend line configuration based on the current axis limits.
+   * This function calculates the new limits for the trend line axes based on the current x-axis
+   * limits and the original limits of the trend line axes.
+   * It assumes that the origininal trend line axes are linear transformations of the original x-axis
+   * limits and applies the same transformation to the current x-axis limits to get the new limits
+   * for the current trend line axes.
+   */
+  updateTrendLineConfiguration(chartId: UID) {
```
LucasLefevre, line 99: looks like the trend line isn't updated correctly
image

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs.ts
```diff
@@ -0,0 +1,494 @@
...
+          this.updateAxisLimits(xMin, xMax);
+        }
+      }
+      this.mode = undefined;
+      window.removeEventListener("pointermove", onDragFromMasterChart, true);
+      window.removeEventListener("pointerup", onPointerUpInMasterChart, true);
+    };
+
+    window.addEventListener("pointermove", onDragFromMasterChart, true);
+    window.addEventListener("pointerup", onPointerUpInMasterChart, true);
```
LucasLefevre, line 412: this can lead to memory leaks if the component is unmounted before `removeEventListener` is called

src/components/figures/figure/figure.ts
```diff
@@ -160,7 +161,14 @@ export class FigureComponent extends Component<Props, SpreadsheetChildEnv> {
   }
 
   get wrapperStyle() {
-    const { x, y, width, height } = this.props.figureUI;
+    let { x, y, width, height, tag } = this.props.figureUI;
+    if (tag === "chart") {
```
LucasLefevre: the figure component should be agnostic of the type of figure (chart, image, whatever).

The figure size  should be changed in the plugins when the user checks the "Show slicer" checkbox.
Another thing that doesn't work with the current approach: figure snapping doesn't take the slicer into account even though it should

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chart_store.ts
anhe-odoo: In fact, the store use a "custom" figure id (chartType-figureID), and the idea of thie Record was to know which custom Id was associated to the figureId, so that, when removing a figure, we can delete the corresponding data. But the handle of the commands is wrong, I made a mistake, and the idConversion type should be Record<UID, UID[]> to keep all the custom IDs associated to a figureID.

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chart_store.ts
anhe-odoo: Yes, this is kind of a "known" issue, but ADRM proposed to let it as is to merge the task and do some follow-up and small fix in another tasks right after.

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs.ts
```diff
@@ -0,0 +1,484 @@
...
+        target["style"].cursor = "default";
+        return;
+      }
+      const { left, right, top, bottom } = this.slider.chartArea;
+      const start = this.lowerBound ?? left;
+      const end = this.upperBound ?? right;
+      if (y < top || y > bottom) {
+        target["style"].cursor = "default";
+      } else if (Math.abs(start - x) < 5 || Math.abs(end - x) < 5) {
+        target["style"].cursor = "e-resize";
```
rrahir:
```suggestion
        target["style"].cursor = "ew-resize";
```

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs_plugins.ts
```diff
@@ -0,0 +1,64 @@
...
+    }
+    ctx.save();
+    ctx.fillStyle = "rgba(255,255,255,0.5)";
+    ctx.beginPath();
+    ctx.rect(left, bottom, lowerBound - left, top - bottom);
+    ctx.rect(upperBound, bottom, right - upperBound, top - bottom);
+    ctx.fill();
+    ctx.beginPath();
+    ctx.strokeStyle = "#bbb";
+    ctx.rect(lowerBound, bottom, upperBound - lowerBound, top - bottom);
```
rrahir, line 45: I now realise that there have been several style iterations, I just want to leave the suggestion so I don't forget about it later on but feel free to ignore completely :)
https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/roundRect
this could follow the generic design

src/plugins/core/chart.ts
```diff
@@ -83,6 +88,28 @@ export class ChartPlugin extends CorePlugin<ChartState> implements ChartState {
         this.addChart(cmd.figureId, cmd.definition);
         break;
       case "UPDATE_CHART": {
+        const definition = this.getters.getChartDefinition(cmd.figureId);
+        const previousZoomable = "zoomable" in definition && definition.zoomable;
+        const futureZoomable = "zoomable" in cmd.definition && cmd.definition.zoomable;
+        if (previousZoomable && !futureZoomable) {
+          const sheetId = this.getters.getFigureSheetId(cmd.figureId)!;
+          const figureDefinition = this.getters.getFigure(sheetId, cmd.figureId)!;
+          this.dispatch("UPDATE_FIGURE", {
```
rrahir: this will not suit well with the new carousel feature which allows several charts upon the same figure, so ticking the "zoom" of each chart would make the wrapper figure grow everytime. Of course we can filter out the figures of type "carousel" but then carousel and "simple" charts would not have the same behaviour :'(

Are we fucntionally attached tothis particular feature? I can see that it was raised by Francois on the 1rst of august but I don't know what was the conclusion on that point. Personally, I agree that it's weird/bad that it breaks the layout between the edition and the dashboard view. That being said, we can probably improve that later on if we don't find the "best" idea.

rrahir (commented):
Non -negligible rebase coming up :/

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs.xml
```diff
@@ -0,0 +1,19 @@
+<templates>
+  <t t-name="o-spreadsheet-ZoomableChartJsComponent">
+    <div class="w-100 h-100" t-att-style="canvasStyle">
+      <div t-att-style="containerStyle">
+        <canvas class="o-figure-canvas w-100 h-100" t-ref="graphContainer"/>
+      </div>
+      <div t-if="sliceable" class="o-master-chart-container">
+        <canvas
+          class="o-figure-canvas o-master-chart-canvas w-100 h-100"
+          t-ref="graphSlider"
```
rrahir:
```suggestion
          t-ref="chartSlider"
```

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chart_store.ts
```diff
@@ -0,0 +1,128 @@
...
+      return;
+    }
+    let { min, max } = limits;
+    if (min > max) {
+      [min, max] = [max, min];
+    }
+    this.currentAxesLimits[chartId] = { x: { min, max } };
+  }
+
+  /* Update the trend line configuration based on the current axis limits.
```
rrahir:
```suggestion
  /* Update the trend line axis configuration based on the current axis limits.
```

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs.ts
```diff
@@ -0,0 +1,492 @@
...
+import { chartJsExtensionRegistry } from "../chart_js_extension";
+import { ChartJsComponent } from "../chartjs";
+import { ZoomableChartStore } from "./zoomable_chart_store";
+import { slidingWindowPlugin } from "./zoomable_chartjs_plugins";
+
+css/* scss */ `
+  .o-spreadsheet {
+    .o-master-chart-container {
+      height: ${CHART_ZOOM_SLIDER_HEIGHT}px;
+      margin-left: 0px;
```
rrahir: remove those 3 rules and use the class `m-0` instead

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs.ts
```diff
@@ -0,0 +1,492 @@
...
+
+export class ZoomableChartJsComponent extends ChartJsComponent {
+  static template = "o-spreadsheet-ZoomableChartJsComponent";
+
+  private store!: Store<ZoomableChartStore>;
+  private fullScreenChartStore!: Store<FullScreenChartStore>;
+
+  private sliderCanvas = useRef("graphSlider");
+  private slider?: Chart;
+  private mode?: "selectInMaster" | "moveInMaster" | "moveInDetail";
```
rrahir: "moveInDetail" is not used anymore

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs.ts
rrahir: Perhaps, we could try to be consistent in the naming and stick to master/detail and not talk about sliders? 
Or go with slider/ zoomed,  master/zoomed.

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs.ts
```diff
@@ -0,0 +1,492 @@
...
+  }
+
+  protected updateChartJs(chartRuntime: ChartJSRuntime) {
+    const chartData = chartRuntime.chartJsConfig as ChartConfiguration<any>;
+    const newDatasetBoundaries = this.getAxisLimitsFromDataset(chartData);
+    if (
+      this.datasetBoundaries.xMin !== newDatasetBoundaries.xMin ||
+      this.datasetBoundaries.xMax !== newDatasetBoundaries.xMax
+    ) {
+      this.store.clearAxisLimits(this.chartId);
```
rrahir, line 180: This line is not tested, I'm not 100% sure why we have it

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chart_store.ts
```diff
@@ -0,0 +1,128 @@
...
+import { SpreadsheetStore } from "../../../../../stores";
+
+const TREND_LINE_AXES_IDS = [TREND_LINE_XAXIS_ID, MOVING_AVERAGE_TREND_LINE_XAXIS_ID] as const;
+const ZOOMABLE_AXIS_IDS = ["x", ...TREND_LINE_AXES_IDS] as const;
+export type AxisId = (typeof ZOOMABLE_AXIS_IDS)[number];
+export type AxesLimits = {
+  [chartId: UID]: { [axisId in AxisId]?: { min: number; max: number } | undefined };
+};
+export class ZoomableChartStore extends SpreadsheetStore {
+  mutators = [
```
rrahir, line 15: I think preatty much every mutator could return "noStateChange" to avoid triggering a full rendering of Spreadsheet. it should not break the flow since we rely on the internal rendering of chartjs

tests/figures/chart/charts_zoom.test.ts
```diff
@@ -0,0 +1,454 @@
...
+      expect(newMax).toEqual(3);
+    });
+
+    test("Can select from the right bound on the master chart", async () => {
+      const element = fixture.querySelector(".o-master-chart-canvas") as HTMLCanvasElement;
+      const { left, top, width, height } = element.getBoundingClientRect();
+      const startX = left + width;
+      const startY = top + height / 2;
+      const offsetX = -width / 2;
+      triggerMouseEvent(element, "pointerdown", startX, startY);
```
rrahir: these steps can be replaced by the async helper `dragElements`
we could rename it as `MouseDrag` or `ClickAndDrag` as we are not explicitely dragging the element, it's just that's it's always been used like this up until now

rrahir (changes requested):

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs.ts
anhe-odoo: We have it to clear the clear the current zoom status if the boundaries of the datasets have changed (for example, if we zoomed between x = 10 and x = 20 but after an update xmax = 15, the previous zoom status is useless). I can had a test to chck this :)

rrahir:
robodoo r+

rrahir:
robodoo rebase-ff

PR #6047, opened 2025-03-27
[REF] range: Range is a POJO

LucasLefevre:
These commits are part of a broader refactoring effort to simplify Range by
removing the `RangeImpl` class, making `Range` a plain old javascipt object
(POJO).

There are currently 2 concepts: `Range` and `RangeImp`. The original intent
was to abstract away the details of unbounded ranges from business logic.
However, in practice, various parts of the code increasingly require access
to these details as new features are introduced. And both have converged
more and more over time. While the abstraction was well-intentioned, it has
proven to be an abstraction that do not really fit all needs.

We now use a pojo because:

- (almost) all our other data structures are POJOs except `RangeImpl`
- POJOs are simpler and easy to work with
- keeping data separate from business logic is what we do everywhere else
- business logic is currently spread across 3 different locations:
  Range plugin, helper functions, Range class

Task: 4676075

### DIFF
--- a/src/clipboard_handlers/tables_clipboard.ts
+++ b/src/clipboard_handlers/tables_clipboard.ts
@@ -70,7 +70,7 @@ export class TableClipboardHandler extends AbstractCellClipboardHandler<
         ) {
           copiedTablesIds.add(table.id);
           copiedTable = {
-            range: coreTable.range.rangeData,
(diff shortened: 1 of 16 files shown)


pro-odoo:
robodoo rebase-merge

pro-odoo:
robodoo r+

PR #6053, opened 2025-03-31
[PERF] tokenize: faster space tokenize

LucasLefevre:
## Description:

Matching the white space regexp is expensive, especially compared to a
strict equality (`=== " "`).

A key observation is that special white space characters are much much less
frequent than regular " " spaces.

With this commit, we run the regexp only once first to check if the
formula contains special white characters. If it does, we keep the
current behavior. But if it doesn't, we use another faster function
to tokenize spaces.

Time spent by `tokenize` when loading the `BE timesheet` dashboard:
(average of 8)

before: ~578ms
after:  ~503ms (87%)

Task: [4684215](https://www.odoo.com/odoo/2328/tasks/4684215)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/formulas/tokenizer.ts
+++ b/src/formulas/tokenizer.ts
@@ -53,6 +53,7 @@ export function tokenize(str: string, locale = DEFAULT_LOCALE): Token[] {
 
   while (!chars.isOver()) {
     let token =
+      tokenizeNewLine(chars) ||
       tokenizeSpace(chars) ||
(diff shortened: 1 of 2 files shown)


pro-odoo (approved):
robodoo r+

pro-odoo:
robodoo rebase-ff

PR #6057, opened 2025-04-03
[PERF] table_conversion: avoid reading all cells multiple times

laa-odoo:
Previously, the content of each cell was inspected multiple times, once per table, leading to significant performance overhead. In this commit, the inspection process is optimized to traverse the cells only once.

A data structure now stores only the cells that need to be updated, improving both efficiency and calculation time.

This optimization results in a 10% reduction in total import time.

Task: [4699860](https://www.odoo.com/odoo/2328/tasks/4699860)

### DIFF
--- a/src/xlsx/conversion/table_conversion.ts
+++ b/src/xlsx/conversion/table_conversion.ts
@@ -162,41 +162,126 @@ function applyStyleToZone(appliedStyle: Style, zone: Zone, cells: CellMap, style
  * In all the sheets, replace the table-only references in the formula cells with standard references.
  */
 function convertTableFormulaReferences(convertedSheets: SheetData[], xlsxSheets: XLSXWorksheet[]) {
+  const deconstructedSheets = deconstructSheets(convertedSheets);
+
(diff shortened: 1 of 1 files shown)


rrahir:
the commit title could point towards the global context in which it is used, which is the Xlsx import if i'm not. mistaken?

laa-odoo:
> the commit title could point towards the global context in which it is used, which is the Xlsx import if i'm not. mistaken?

in fact this concerns the xlsx import, the name of the commit has been changed

LucasLefevre (commented):
Is it actually slower ? (for the common case where there are no table references)
I exported the large formula dataset to re-import it.
With this branch, `convertTableFormulaReferences` is 64ms, in 16.0 it's less than 1ms

laa-odoo:
> Is it actually slower ? (for the common case where there are no table references) I exported the large formula dataset to re-import it. With this branch, `convertTableFormulaReferences` is 64ms, in 16.0 it's less than 1ms

should be ok

rrahir:
robodoo r+

PR #6063, opened 2025-04-04
[FIX] popover: rename props to kebab-case

LucasLefevre:
There's a linting test in odoo flagging string props that should probably be translated:

`<Greeting a="'Hello'"/>`

In this case the props should be modified with `.translate`.

`<Greeting a.translate="Hello"/>`

But of course, not all string props are user-facing strings that should be translated. Some are just technical parameters.

The current heuristic implemeted in the linting test is:

if the string starts with an uppercase letter (and it's not all uppercase), then it's probably a user-facing string.

It means the props `"BottomLeft"` etc. are flagged as such even though they are technical parameter.

This commit renames those props to avoid being wrongly flagged.

Task: 0

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


pro-odoo (commented):
robodoo r+

PR #6073, opened 2025-04-07
[FIX] spreadsheet: add name to DUPLICATE_SHEET

fdamhaut:
When duplicating sheet, the name was based on the
translation of "Copy of", which would lead to divergent sheet name if multiple users had different locale.

Task: 4640070

## Description:

Task: [4640070](https://www.odoo.com/odoo/2328/tasks/4640070)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/migrations/data.ts
+++ b/src/migrations/data.ts
@@ -13,6 +13,7 @@ import {
   UuidGenerator,
   zoneToXc,
 } from "../helpers/index";
+import { _lt } from "../translation";
 import { StateUpdateMessage } from "../types/collaborative/transport_service";
 import {
   CoreCommand,
@@ -406,6 +407,7 @@ export function repairInitialMessages(
   initialMessages = dropCommands(initialMessages, "SORT_CELLS");
   initialMessages = dropCommands(initialMessages, "SET_DECIMAL");
   initialMessages = fixChartDefinitions(data, initialMessages);
+  initialMessages = fixTranslatedDuplicateSheetName(data, initialMessages);
   return initialMessages;
 }
 
@@ -508,6 +510,44 @@ function fixChartDefinitions(data: Partial<WorkbookData>, initialMessages: State
   return messages;
 }
 
+function fixTranslatedDuplicateSheetName(
+  data: Partial<WorkbookData>,
+  initialMessages: StateUpdateMessage[]
+): StateUpdateMessage[] {
+  const sheetNames = {};
+  for (const sheet of data.sheets || []) {
+    sheetNames[sheet.id] = sheet.name;
+  }
+  const messages: StateUpdateMessage[] = [];
+  for (const message of initialMessages) {
+    if (message.type === "REMOTE_REVISION") {
+      const commands: CoreCommand[] = [];
+      for (const cmd of message.commands) {
+        switch (cmd.type) {
+          case "DUPLICATE_SHEET":
+            cmd.sheetNameTo = getDuplicateSheetName(
+              sheetNames[cmd.sheetId],
+              Object.values(sheetNames)
+            );
+            break;
+          case "CREATE_SHEET":
+          case "RENAME_SHEET":
+            sheetNames[cmd.sheetId] = cmd.name || getNextSheetName(Object.values(sheetNames));
+            break;
+        }
+        commands.push(cmd);
+      }
+      messages.push({
+        ...message,
+        commands,
+      });
+    } else {
+      messages.push(message);
+    }
+  }
+  return initialMessages;
+}
+
 // -----------------------------------------------------------------------------
 // Helpers
 // -----------------------------------------------------------------------------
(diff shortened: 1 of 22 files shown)


src/migrations/data.ts
```diff
@@ -508,6 +510,44 @@ function fixChartDefinitions(data: Partial<WorkbookData>, initialMessages: State
...
+  for (const message of initialMessages) {
+    if (message.type === "REMOTE_REVISION") {
+      const commands: CoreCommand[] = [];
+      for (const cmd of message.commands) {
+        switch (cmd.type) {
+          case "DUPLICATE_SHEET":
+            cmd.sheetNameTo = getDuplicateSheetName(
+              sheetNames[cmd.sheetId],
+              Object.values(sheetNames)
+            );
```
hokolomopo, line 531: getDuplicateSheetName uses translations, so we'll end up with the exact same issue with different users having different sheet names after the repair :see_no_evil:

src/plugins/core/sheet.ts
```diff
@@ -766,7 +766,7 @@ export class SheetPlugin extends CorePlugin<SheetState> implements SheetState {
     this.history.update("sheetIdsMapName", sheetIdsMapName);
   }
 
-  private getDuplicateSheetName(sheetName: string) {
```
hokolomopo, line 769: mostly duplicated with helpers of migration, is it really useful to have both ? same for getNextSheetName

src/types/commands.ts
```diff
@@ -343,6 +343,7 @@ export interface DeleteSheetCommand extends SheetDependentCommand {
 export interface DuplicateSheetCommand extends SheetDependentCommand {
   type: "DUPLICATE_SHEET";
   sheetIdTo: UID;
+  sheetNameTo: string;
```
hokolomopo, line 346: should also have an allowDispatch to make sure we don't have a duplicate sheet name

hokolomopo (commented):
:wave:

src/plugins/core/sheet.ts
```diff
@@ -103,7 +105,10 @@ export class SheetPlugin extends CorePlugin<SheetState> implements SheetState {
         return this.checkValidations(cmd, this.checkSheetName, this.checkSheetPosition);
       }
       case "DUPLICATE_SHEET": {
-        return this.sheets[cmd.sheetIdTo] ? CommandResult.DuplicatedSheetId : CommandResult.Success;
+        return this.sheets[cmd.sheetIdTo] ||
+          this.orderedSheetIds.map(this.getSheetName.bind(this)).includes(cmd.sheetNameTo)
+          ? CommandResult.DuplicatedSheetId
```
hokolomopo, line 110:
```suggestion
          ? CommandResult.DuplicatedSheetName
```

hokolomopo (commented):
:+1:

src/migrations/data.ts
```diff
@@ -508,6 +510,44 @@ function fixChartDefinitions(data: Partial<WorkbookData>, initialMessages: State
...
+    sheetNames[sheet.id] = sheet.name;
+  }
+  const messages: StateUpdateMessage[] = [];
+  for (const message of initialMessages) {
+    if (message.type === "REMOTE_REVISION") {
+      const commands: CoreCommand[] = [];
+      for (const cmd of message.commands) {
+        switch (cmd.type) {
+          case "DUPLICATE_SHEET":
+            cmd.sheetNameTo = getDuplicateSheetName(
```
LucasLefevre, line 528: when this PR will be merged, some DUPLICATE_SHEET will have a `sheetNameTo` property. And in this case, it should not be overwritten

src/migrations/data.ts
```diff
@@ -555,3 +595,14 @@ export function createEmptyExcelWorkbookData(): ExcelWorkbookData {
     sheets: [createEmptyExcelSheet(INITIAL_SHEET_ID, "Sheet1")],
   };
 }
+
+function getDuplicateSheetName(nameToDuplicate: string, names: string[]): string {
+  let i = 1;
+  const baseName = `Copy of ${nameToDuplicate}`;
```
LucasLefevre, line 601: To avoid breaking existing spreadsheets, we should preserve the previous behavior (even if it's not correct in a collaborative context).
Let's say a company has all users in FR, they used to have a sheet named "Copie de [...]" but now it's "Copy of [...]" and they did based on the sheet name is broken (formulas)
```suggestion
  const baseName = _lt("Copy of %s", nameToDuplicate);
```

tests/plugins/sheets.test.ts
```diff
@@ -787,10 +815,32 @@ describe("sheets", () => {
...
+    });
     const result = model.dispatch("DUPLICATE_SHEET", {
       sheetId: firstSheetId,
       sheetIdTo: duplicatedSheetId,
+      sheetNameTo: "Copy of Sheet1",
+    });
+    expect(result).toBeCancelledBecause(CommandResult.DuplicatedSheetId);
+  });
+
+  test("cannot duplicate a sheet twice with the same new id", () => {
```
LucasLefevre, line 831:
```suggestion
  test("cannot duplicate a sheet twice with the same name", () => {
```

src/helpers/sheet.ts
```diff
@@ -10,3 +10,13 @@ export function createDefaultRows(rowNumber: number): Row[] {
   }
   return rows;
 }
+
+export function getNextSheetName(names: string[], baseName: string = "Sheet"): string {
```
LucasLefevre, line 14:
```suggestion
export function getNextSheetName(existingNames: string[], baseName: string = "Sheet"): string {
```

src/migrations/data.ts
hokolomopo: Yeah but that means that after repairing the revisions, the commands will be different for each user based on it's language. Isn't it way more of a problem ?

LucasLefevre: I think it's better to keep the current problem as it is because:

- we didn't get any ticket or feedback
- If this bug caused an issue for some customer, it is very likely that they renamed the sheet to make it work again.

so in practice I don't think it's a real issue in existing spreadsheets (or they have been solved already by renaming)

On the other hand, if we change the behavior, there's a risk of breaking spreadsheets that are working right now.

rrahir:
@robodoo r+

PR #6074, opened 2025-04-08
[IMP] registry: add key unicity on add

fdamhaut:
Throw an error if we try to add a key to
the registry that is already present.

`.replace()` allow to add/change an already
present key.

Task: 4712321

## Description:

Task: [4712321](https://www.odoo.com/odoo/2328/tasks/4712321)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/functions/index.ts
+++ b/src/functions/index.ts
@@ -67,6 +67,14 @@ export class FunctionRegistry extends Registry<FunctionDescription> {
 
   add(name: string, addDescr: AddFunctionDescription) {
     name = name.toUpperCase();
+    if (name in this.content) {
+      throw new Error(`${name} is already present in this registry!`);
+    }
+    return this.replace(name, addDescr);
+  }
+
+  replace(name: string, addDescr: AddFunctionDescription) {
+    name = name.toUpperCase();
     if (!functionNameRegex.test(name)) {
       throw new Error(
         _t(
@@ -78,7 +86,7 @@ export class FunctionRegistry extends Registry<FunctionDescription> {
     const descr = addMetaInfoFromArg(name, addDescr);
     validateArguments(descr);
     this.mapping[name] = createComputeFunction(descr);
-    super.add(name, descr);
+    super.replace(name, descr);
     return this;
   }
 }
--- a/src/registries/menu_items_registry.ts
+++ b/src/registries/menu_items_registry.ts
@@ -21,12 +21,20 @@ export class MenuItemRegistry extends Registry<ActionSpec> {
    * @param path Path of items to add this subitem
    * @param value Subitem to add
    */
-  addChild(
+  addChild(key: string, path: string[], value: ActionSpec | ActionBuilder): this {
+    return this._replaceChild(key, path, value, { force: false });
+  }
+
+  replaceChild(key: string, path: string[], value: ActionSpec | ActionBuilder): this {
+    return this._replaceChild(key, path, value, { force: true });
+  }
+
+  private _replaceChild(
     key: string,
     path: string[],
     value: ActionSpec | ActionBuilder,
-    options: { force: boolean } = { force: false }
-  ): MenuItemRegistry {
+    options: { force: boolean } = { force: true }
+  ): this {
     if (typeof value !== "function" && value.id === undefined) {
       value.id = key;
     }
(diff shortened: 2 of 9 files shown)


src/registries/menu_items_registry.ts
```diff
@@ -9,7 +9,7 @@ export class MenuItemRegistry extends Registry<ActionSpec> {
   /**
    * @override
    */
-  add(key: string, value: ActionSpec): MenuItemRegistry {
+  replace(key: string, value: ActionSpec): this {
     if (value.id === undefined) {
       value.id = key;
     }
```
hokolomopo, line 15: you should also do the same checks for `addChild` no ? Or do we not care about them ?

src/functions/index.ts
```diff
@@ -74,6 +75,12 @@ export class FunctionRegistry extends Registry<FunctionDescription> {
   } = {};
 
   add(name: string, addDescr: AddFunctionDescription) {
+    name = name.toUpperCase();
+    // @ts-ignore
+    return super.add(name, addDescr);
```
hokolomopo, line 80: here also, why does the `add` has all the logic removed ? and a brand new ts-ignore ?

src/registries/srt_registry.ts
```diff
@@ -24,6 +24,11 @@ export class SpecificRangeTransformRegistry extends Registry<
     return this;
   }
 
+  replace<C extends CoreCommand>(cmdType: C["type"], fn: CommandAdaptRangeFunction<C>): this {
+    super.replace(cmdType, fn);
+    return this;
+  }
```
hokolomopo, line 30: I don't think we need/want to allow replace for this registry

src/registries/repeat_commands_registry.ts
```diff
@@ -67,7 +67,6 @@ repeatCommandTransformRegistry.add("CREATE_CHART", repeatCreateChartCommand);
 repeatCommandTransformRegistry.add("CREATE_IMAGE", repeatCreateImageCommand);
 repeatCommandTransformRegistry.add("GROUP_HEADERS", repeatGroupHeadersCommand);
 repeatCommandTransformRegistry.add("UNGROUP_HEADERS", repeatGroupHeadersCommand);
-repeatCommandTransformRegistry.add("UNGROUP_HEADERS", repeatGroupHeadersCommand);
```
hokolomopo, line 70: oops :see_no_evil:

hokolomopo (commented):
bip bop :robot:

src/functions/index.ts
fdamhaut: But super.add() will call replace, so there is no need for duplicating the logic ?

@ts-ignore is because this registry add take as argument something different from the object that will be stored, it allows to use the generic registry as a base which doesn't support this usecase (type-wise only)

src/registries/topbar_component_registry.ts
```diff
@@ -17,9 +17,9 @@ class TopBarComponentRegistry extends Registry<TopbarComponent> {
   mapping: { [key: string]: Function } = {};
   uuidGenerator = new UuidGenerator();
 
-  add(name: string, value: Omit<TopbarComponent, "id">) {
+  replace(name: string, value: Omit<TopbarComponent, "id">) {
```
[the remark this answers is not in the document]
fdamhaut: same as above

src/registries/srt_registry.ts
fdamhaut: I feel like between having it work (and be explicit) or having it throw an error, it might as well be there :shrug:.
If you feel strongly about it, I can make it err instead.

src/functions/index.ts
hokolomopo: Oh I didn't catch that, indeed. 

nitpick: I think I'd prefer to not have a ts-ignore, do the check of key duplication here, and call `this.replace` instead of `this.add`.

src/registries/srt_registry.ts
hokolomopo: I don't feel that strongly about it 🤷 

But actually do we even need to override `replace` and `add` in this registry ? All we do is call `super`

fdamhaut: The override are only present for typing purposes (as it already was)

src/registries/menu_items_registry.ts
LucasLefevre: for consistency, I think we could remove `options.force` and introduce `replaceChild`

PR #6076, opened 2025-04-09
Saas 18.1 chart legend onclick dashboard v2 adrm

hokolomopo:
### [IMP] chart: add `mouseup` event to listened events

To detect a middle mouse click on a chart, we have to use the
`mouseup` event, the `click` event is not triggered with middle
mouse button.

### [FIX] charts: unregister ChartJS extension on unmount

The ChartJS extension were not unregistered when the leaving the
spreadsheet. Some of them only make sense in the context of the
spreadsheet, so keeping them when leaving could lead to crashes
when using ChartJS elsewhere.

Task: [4636147](https://www.odoo.com/odoo/2328/tasks/4636147)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/figures/chart/chartJs/chart_js_extension.ts
+++ b/src/components/figures/chart/chartJs/chart_js_extension.ts
@@ -1,12 +1,28 @@
-import { Plugin as ChartJSPlugin } from "chart.js";
 import { Registry } from "../../../../registries/registry";
 
-export const chartJsExtensionRegistry = new Registry<ChartJSPlugin>();
+export const chartJsExtensionRegistry = new Registry<{
+  register: (chart: typeof window.Chart) => void;
+  unregister: (chart: typeof window.Chart) => void;
+}>();
 
-/** Return window.Chart, making sure all our extensions are loaded in ChartJS */
-export function getChartJSConstructor() {
-  if (window.Chart && !window.Chart?.registry.plugins.get("chartShowValuesPlugin")) {
-    window.Chart.register(...chartJsExtensionRegistry.getAll());
+export function areChartJSExtensionsLoaded() {
+  return !!window.Chart.registry.plugins.get("chartShowValuesPlugin");
+}
+
+export function registerChartJSExtensions() {
+  if (!window.Chart || areChartJSExtensionsLoaded()) {
+    return;
   }
-  return window.Chart;
+  for (const registryItem of chartJsExtensionRegistry.getAll()) {
+    registryItem.register(window.Chart);
+  }
+}
+
+export function unregisterChartJsExtensions() {
+  if (!window.Chart) {
+    return;
+  }
+  for (const registryItem of chartJsExtensionRegistry.getAll()) {
+    registryItem.unregister(window.Chart);
+  }
 }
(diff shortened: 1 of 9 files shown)


src/components/figures/chart/chartJs/chart_js_extension.ts
```diff
@@ -1,12 +1,24 @@
...
-/** Return window.Chart, making sure all our extensions are loaded in ChartJS */
-export function getChartJSConstructor() {
-  if (window.Chart && !window.Chart?.registry.plugins.get("chartShowValuesPlugin")) {
-    window.Chart.register(...chartJsExtensionRegistry.getAll());
+export function registerChartJSExtensions() {
+  if (!window.Chart || window.Chart.registry.plugins.get("chartShowValuesPlugin")) {
+    return;
+  }
+  for (const registryItem of chartJsExtensionRegistry.getAll()) {
+    registryItem.register(window.Chart);
```
rrahir, line 13:
```suggestion
    window.Chart.register(registryItem);
```
maybe?
and the registry would only be a list of plugins

hokolomopo: In 18.2 (or 18.3?) we need to register things that aren't plugins 🙃

src/components/figures/chart/chartJs/chartjs.ts
```diff
@@ -3,17 +3,32 @@ import { Chart, ChartConfiguration } from "chart.js/auto";
...
 
 interface Props {
   figure: Figure;
 }
 
-chartJsExtensionRegistry.add("chartShowValuesPlugin", chartShowValuesPlugin);
-chartJsExtensionRegistry.add("waterfallLinesPlugin", waterfallLinesPlugin);
+css/* scss */ `
+  .o-spreadsheet {
+    .o-chart-custom-tooltip {
```
LucasLefevre, line 17: I'm not sure what this css is doing in this commit

hokolomopo: neither do I ...

src/components/figures/chart/chartJs/chart_js_extension.ts
```diff
@@ -1,12 +1,24 @@
...
+  register: (chart: typeof window.Chart) => void;
+  unregister: (chart: typeof window.Chart) => void;
+}>();
 
-/** Return window.Chart, making sure all our extensions are loaded in ChartJS */
-export function getChartJSConstructor() {
-  if (window.Chart && !window.Chart?.registry.plugins.get("chartShowValuesPlugin")) {
-    window.Chart.register(...chartJsExtensionRegistry.getAll());
+export function registerChartJSExtensions() {
+  if (!window.Chart || window.Chart.registry.plugins.get("chartShowValuesPlugin")) {
```
rrahir, line 9: I just have one issue with this strategy - we now HAVE to spawn a full spreadsheet to have the extensions available but we also use chartJs to export figures in XLSX  -  this means that we could theoretically have chartJs running alongside `Model` without any component. 

Could we imagine having this register return true if it had to actually do some work so we know that we need to unregister afterwards?  Then we could use it (in the future) for standalone models

hokolomopo: Rather than an opaque true/false return  value, I'd extract the `if` inside a `areChartJSExtensionLoaded()` I think. That would work for your use case no ?

rrahir:
@robodoo r+

rrahir:
@robodoo rebase-ff

PR #6077, opened 2025-04-10
[FIX] sheet: make sheetName comparison case incensitive

fdamhaut:
As sheetName are case incensitive, the comparaison should too.

Task: 4707398

## Description:

Task: [4707398](https://www.odoo.com/odoo/2328/tasks/4707398)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/sheet.ts
+++ b/src/helpers/sheet.ts
@@ -1,4 +1,5 @@
 import { Row } from "../types";
+import { getUnquotedSheetName } from "./misc";
 
 export function createDefaultRows(rowNumber: number): Row[] {
   const rows: Row[] = [];
@@ -9,4 +10,14 @@ export function createDefaultRows(rowNumber: number): Row[] {
     rows.push(row);
   }
   return rows;
+}
+
+export function isSheetNameEqual(name1: string | undefined, name2: string | undefined): boolean {
+  if (name1 === undefined || name2 === undefined) {
+    return false;
+  }
+  return (
+    getUnquotedSheetName(name1.trim().toUpperCase()) ===
+    getUnquotedSheetName(name2.trim().toUpperCase())
+  );
 }
--- a/src/plugins/core/range.ts
+++ b/src/plugins/core/range.ts
@@ -13,6 +13,7 @@ import {
   splitReference,
   toUnboundedZone,
 } from "../../helpers/index";
+import { isSheetNameEqual } from "../../helpers/sheet";
 import {
   ApplyRangeChange,
   ApplyRangeChangeResult,
@@ -167,7 +168,7 @@ export class RangeAdapter implements CommandHandler<CoreCommand> {
           if (range.sheetId === cmd.sheetId) {
             return { changeType: "CHANGE", range };
           }
-          if (cmd.name && range.invalidSheetName === cmd.name) {
+          if (isSheetNameEqual(range.invalidSheetName, cmd.name)) {
             const invalidSheetName = undefined;
             const sheetId = cmd.sheetId;
             const newRange = range.clone({ sheetId, invalidSheetName });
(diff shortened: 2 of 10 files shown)


src/helpers/sheet.ts
```diff
@@ -10,3 +10,10 @@ export function createDefaultRows(rowNumber: number): Row[] {
   }
   return rows;
 }
+
+export function isSheetNameEqual(name1: string | undefined, name2: string | undefined): boolean {
+  if (name1 === undefined || name2 === undefined) {
+    return false;
+  }
+  return name1.trim().toUpperCase() === name2.trim().toUpperCase();
+}
```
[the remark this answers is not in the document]
pro-odoo: Also, I think it could be easy and worth to add 3-4 tests for this helper

src/plugins/core/range.ts
```diff
@@ -167,7 +168,7 @@ export class RangeAdapter implements CommandHandler<CoreCommand> {
           if (range.sheetId === cmd.sheetId) {
             return { changeType: "CHANGE", range };
           }
-          if (cmd.name && range.invalidSheetName === cmd.name) {
+          if (isSheetNameEqual(range.invalidSheetName, cmd.name)) {
```
rrahir, line 171: this bit of code was not tested at all so far so it seems like it could be the time ^^
Best test case scenario I came up with so far: 
From an empty model
- Add Sheet2
- Add Sheet3
- UpdateCell in Sheet1 that references Sheet2 (=Sheet2!A1)
- delete sheet2
- rename Sheet3 to sheet2 (no capital s)

Then you can check the internal structure of the cell and check that the range in its dependencies has the correct sheet id and not invalidSheetName.

General remark about that bit of code specifically,:

Now the issue is that when we have an invalid range, we set  an invalidXc as well and we corrupt its internal zone on purpose ( https://github.com/odoo/o-spreadsheet/blob/16.0/src/plugins/core/range.ts#L157 ), which means that even if we fix the invalid sheetName, we cannot turn it back to a valid range at any point.
- Does the handler make sense?
- If so, I feel like we lack some functional behaviours that leverage it

@VincentSchippefilt

src/xlsx/conversion/formula_conversion.ts
```diff
@@ -66,8 +66,8 @@ function convertFormula(formula: string, data: XLSXImportData): string {
     externalRefId = Number(externalRefId) - 1;
     cellRef = cellRef.replace(/\$/g, "");
 
-    const sheetIndex = data.externalBooks[externalRefId].sheetNames.findIndex(
-      (name) => name === sheetName
+    const sheetIndex = data.externalBooks[externalRefId].sheetNames.findIndex((name) =>
```
rrahir, line 69: I don't really know how the external references work, but it feels like it requires a test

rrahir (changes requested):

rrahir:
comparation & comparaison >>>> comparison ;-)

src/plugins/core/range.ts
rrahir: We discussed this with Vincent and we will fix the behaviour of the invalid ranges in master

src/helpers/sheet.ts
```diff
@@ -10,3 +11,13 @@ export function createDefaultRows(rowNumber: number): Row[] {
   }
   return rows;
 }
+
+export function isSheetNameEqual(name1: string | undefined, name2: string | undefined): boolean {
+  if (name1 === undefined || name2 === undefined) {
+    return false;
```
rrahir, line 17: is it possible to have both names undefined at the same time? what does it mean then? is it really different?

LucasLefevre (commented):
robodoo r+

LucasLefevre:
robodoo r+

PR #6078, opened 2025-04-10
[FIX] OT: Fix chart commands OT

rrahir:
Task: 4717724

## Description:

Task: [4717724](https://www.odoo.com/odoo/2328/tasks/4717724)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/charts/gauge_chart.ts
+++ b/src/helpers/charts/gauge_chart.ts
@@ -31,7 +31,7 @@ import {
 import { Validator } from "../../types/validator";
 import { clip } from "../index";
 import { createValidRange } from "../range";
-import { rangeReference } from "../references";
+import { mergeReference, rangeReference, splitReference } from "../references";
 import { toUnboundedZone, zoneToXc } from "../zones";
 import { AbstractChart } from "./abstract_chart";
 import { adaptChartRange, chartFontColor, copyLabelRangeWithNewSheetId } from "./chart_common";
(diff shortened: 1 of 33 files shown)


src/helpers/charts/gauge_chart.ts
```diff
@@ -179,16 +181,23 @@ export class GaugeChart extends AbstractChart {
 
   static transformDefinition(
     definition: GaugeChartDefinition,
+    chartSheetId: UID,
+    sheetMap: Record<string, UID>,
     executed: AddColumnsRowsCommand | RemoveColumnsRowsCommand
   ): GaugeChartDefinition {
     let dataRangeZone: UnboundedZone | undefined;
+    let sheetName: string | undefined;
```
hokolomopo, line 189:
```suggestion
    let newDataRange: string | undefined;
```
nitpick, that way we don't have to have two `let` variables hanging around

same in scorecard

src/plugins/core/chart.ts
```diff
@@ -100,6 +100,7 @@ export class ChartPlugin extends CorePlugin<ChartState> implements ChartState {
                 size: { width: fig.width, height: fig.height },
                 definition: chart.getDefinition(),
                 sheetId: cmd.sheetIdTo,
+                sheetMap: this.getters.getSheetMap(),
```
hokolomopo, line 103: We talked about filtering the map with only relevant sheet name to reduce the size. Is it annoying/not necessary ?

hokolomopo (commented):
😔 sad day to be a command

PR #6081, opened 2025-04-11
[PERF] autofill: improve command dispatching

pro-odoo:
Before this commit, the command dispatching for autofill was inefficient, leading to large command sizes and performance issues. This commit optimizes the command dispatching process, reducing the size of the commands and improving performance.

The main changes are about borders, conditional formats and data validation rules. These three commands were previously dispatched separately, with a change of one cell in the range/zone. Now, the new range/zone is computed and dispatched in one command (one command by CF, by border, by DV rule).

This commit introduces a new command for the border (`SET_BORDER_TARGET`) as we cannot use the `SET_ZONE_BORDER` in this case as it does not take the same parameters. In master, we could change `SET_BORDER` command to take a target instead of a position and then remove the `SET_BORDER_TARGET` command.

Command size for autofilling a row (26 cells) with two CFs and a border to the bottom of the sheet (99 rows) was reduced from 2.9MB to 244KB:

Task: 4718202

### DIFF
--- a/src/plugins/ui_feature/autofill.ts
+++ b/src/plugins/ui_feature/autofill.ts
@@ -1,9 +1,12 @@
-import { clip, deepCopy, isInside, recomputeZones, toCartesian, toXC } from "../../helpers/index";
+import { clip, isInside, positionToZone, toCartesian, toXC } from "../../helpers/index";
+import { futureRecomputeZones } from "../../helpers/recompute_zones";
 import { autofillModifiersRegistry, autofillRulesRegistry } from "../../registries/index";
 import {
+  AutoFillCellCommand,
   AutofillData,
   AutofillModifier,
   AutofillResult,
+  Border,
   Cell,
   Command,
   CommandResult,
@@ -15,9 +18,12 @@ import {
   LAYERS,
   LocalCommand,
   Tooltip,
+  UID,
   Zone,
 } from "../../types/index";
 import { UIPlugin } from "../ui_plugin";
+
+type AutofillCellData = Omit<AutoFillCellCommand, "type">;
 
 /**
  * This plugin manage the autofill.
@@ -130,25 +136,6 @@ export class AutofillPlugin extends UIPlugin {
       case "AUTOFILL_AUTO":
         this.autofillAuto();
         break;
-      case "AUTOFILL_CELL":
-        this.autoFillMerge(cmd.originCol, cmd.originRow, cmd.col, cmd.row);
-        const sheetId = this.getters.getActiveSheetId();
-        this.dispatch("UPDATE_CELL", {
-          sheetId,
-          col: cmd.col,
-          row: cmd.row,
-          style: cmd.style || null,
-          content: cmd.content || "",
-          format: cmd.format || "",
-        });
-        this.dispatch("SET_BORDER", {
-          sheetId,
-          col: cmd.col,
-          row: cmd.row,
-          border: cmd.border,
-        });
-        this.autofillCF(cmd.originCol, cmd.originRow, cmd.col, cmd.row);
-        this.autofillDV(cmd.originCol, cmd.originRow, cmd.col, cmd.row);
     }
   }
 
@@ -176,6 +163,7 @@ export class AutofillPlugin extends UIPlugin {
     }
     const source = this.getters.getSelectedZone();
     const target = this.autofillZone;
+    const autofillCellsData: AutofillCellData[] = [];
 
     switch (this.direction) {
       case DIRECTION.DOWN:
@@ -186,7 +174,7 @@ export class AutofillPlugin extends UIPlugin {
           }
           const generator = this.createGenerator(xcs);
           for (let row = target.top; row <= target.bottom; row++) {
-            this.computeNewCell(generator, col, row, apply);
+            autofillCellsData.push(this.computeNewCell(generator, col, row));
           }
         }
         break;
@@ -198,7 +186,7 @@ export class AutofillPlugin extends UIPlugin {
           }
           const generator = this.createGenerator(xcs);
           for (let row = target.bottom; row >= target.top; row--) {
-            this.computeNewCell(generator, col, row, apply);
+            autofillCellsData.push(this.computeNewCell(generator, col, row));
           }
         }
         break;
@@ -210,7 +198,7 @@ export class AutofillPlugin extends UIPlugin {
           }
           const generator = this.createGenerator(xcs);
           for (let col = target.right; col >= target.left; col--) {
-            this.computeNewCell(generator, col, row, apply);
+            autofillCellsData.push(this.computeNewCell(generator, col, row));
           }
         }
         break;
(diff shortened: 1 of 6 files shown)


src/plugins/ui_feature/autofill.ts
```diff
@@ -238,6 +246,108 @@ export class AutofillPlugin extends UIPlugin {
...
+  private autofillCell(data: Omit<AutoFillCellCommand, "type">) {
+    this.dispatch("UPDATE_CELL", {
+      sheetId: this.getters.getActiveSheetId(),
+      col: data.col,
+      row: data.row,
+      content: data.content || "",
+      style: data.style || null,
+      format: data.format || "",
+    });
+    this.dispatch("AUTOFILL_CELL", data);
```
pro-odoo, line 303: Still needed for Odoo in master, and it's not an issue it's a UI command

src/plugins/ui_feature/autofill.ts
```diff
@@ -222,13 +216,27 @@ export class AutofillPlugin extends UIPlugin {
...
+      for (const data of autofillCellsData) {
+        this.collectBordersData(data, borders, bordersPositions);
+        this.autofillMerge(data);
+        this.autofillCell(data);
+        this.collectConditionalFormatsData(data, cfNewRanges);
+        this.collectDataValidationsData(data, dvNewZones);
+      }
+      this.autofillBorders(borders, bordersPositions);
+      this.autofillConditionalFormats(cfNewRanges);
+      this.autofillDataValidations(dvNewZones);
```
VincentSchippefilt, line 239: all these methods could receive the `activeSheetId`, thus calling  `getActiveSheetId` only once instead of once per cell per data type

src/plugins/ui_feature/autofill.ts
VincentSchippefilt: maybe add this as a comment in the code, so we don't forget ?

src/plugins/ui_feature/autofill.ts
```diff
@@ -238,6 +246,108 @@ export class AutofillPlugin extends UIPlugin {
...
+    for (const [cfId, changes] of Object.entries(cfNewRanges)) {
+      const sheetId = this.getters.getActiveSheetId();
+      const cf = this.getters.getConditionalFormats(sheetId).find((cf) => cf.id === cfId);
+      if (!cf) {
+        continue;
+      }
+      const newCfRanges = this.getters.getAdaptedCfRanges(sheetId, cf, changes, []);
+      if (newCfRanges) {
+        this.dispatch("ADD_CONDITIONAL_FORMAT", {
+          cf: deepCopy(cf),
```
VincentSchippefilt, line 326: the property "cf.ranges" (string)  is not correct, it looks like it contains the previous version of the ranges (zones)

pro-odoo: `ranges` is ignored, the `cf` property is typed ad `cf: Omit<ConditionalFormat, "ranges">;`
So I simply removed it ;-)

src/plugins/ui_feature/autofill.ts
```diff
@@ -238,6 +247,115 @@ export class AutofillPlugin extends UIPlugin {
...
+  private collectBordersData(
+    data: Omit<AutoFillCellCommand, "type">,
+    borders: Record<UID, Border>,
+    bordersPositions: Record<number, Zone[]>
+  ) {
+    const id = getItemId(data.border, borders);
+    if (!(id in bordersPositions)) {
+      bordersPositions[id] = [];
+    }
+    bordersPositions[id].push(positionToZone({ col: data.col, row: data.row }));
```
VincentSchippefilt, line 259: we could use unbound zones here, I don't think the borders support (or that it is possible to have) unbound zones. it would remove one property from each zone of the border: `hasHeader: false`

VincentSchippefilt (approved):

src/plugins/ui_feature/autofill.ts
```diff
@@ -222,13 +216,28 @@ export class AutofillPlugin extends UIPlugin {
...
-            this.computeNewCell(generator, col, row, apply);
+            autofillCellsData.push(this.computeNewCell(generator, col, row));
           }
         }
         break;
     }
 
     if (apply) {
+      const borders: Record<UID, Border> = {};
+      const bordersPositions: Record<number, Zone[]> = {};
```
hokolomopo, line 227:
```suggestion
      const bordersZones: Record<number, Zone[]> = {};
```

src/plugins/ui_feature/autofill.ts
```diff
@@ -222,13 +216,28 @@ export class AutofillPlugin extends UIPlugin {
...
           for (let col = target.left; col <= target.right; col++) {
-            this.computeNewCell(generator, col, row, apply);
+            autofillCellsData.push(this.computeNewCell(generator, col, row));
           }
         }
         break;
     }
 
     if (apply) {
+      const borders: Record<UID, Border> = {};
```
hokolomopo, line 226:
```suggestion
      const borders: Record<number, Border> = {};
```
if i'm not mistaken

hokolomopo: but can't we have a single `Record<StringifiedBorder, Zone[]>` instead of two different mappings for borders ? Or even just an array of `{zones: Zone[], border: Border}` and use this for the deepEquals ? I feel like it would be simpler than assigning a number ID to every different border

src/plugins/ui_feature/autofill.ts
```diff
@@ -238,6 +247,115 @@ export class AutofillPlugin extends UIPlugin {
...
+    bordersPositions[id].push(positionToZone({ col: data.col, row: data.row }));
+  }
+
+  private collectConditionalFormatsData(
+    sheetId: UID,
+    data: Omit<AutoFillCellCommand, "type">,
+    cfNewRanges: Record<UID, string[]>
+  ) {
+    const { originCol, originRow, col, row } = data;
+    const cfOrigin = this.getters.getRulesByCell(sheetId, originCol, originRow);
```
hokolomopo, line 268:
```suggestion
    const cfsAtOrigin = this.getters.getRulesByCell(sheetId, originCol, originRow);
```
nitpick

src/plugins/ui_feature/autofill.ts
```diff
@@ -238,6 +247,115 @@ export class AutofillPlugin extends UIPlugin {
...
+  }
+
+  private collectDataValidationsData(
+    sheetId: UID,
+    data: Omit<AutoFillCellCommand, "type">,
+    dvNewZones: Record<UID, Zone[]>
+  ) {
+    const { originCol, originRow, col, row } = data;
+    const cellPosition = { sheetId, col: originCol, row: originRow };
+    const dvOrigin = this.getters.getValidationRuleForCell(cellPosition);
```
hokolomopo, line 285:
```suggestion
    const dvsAtOrigin = this.getters.getValidationRuleForCell(cellPosition);
```
nitpick

src/plugins/ui_feature/autofill.ts
```diff
@@ -238,6 +247,115 @@ export class AutofillPlugin extends UIPlugin {
...
+    // Still usefull in odoo ATM to autofill field sync
+    this.dispatch("AUTOFILL_CELL", data);
+  }
+
+  private autofillBorders(
+    sheetId: UID,
+    borders: Record<UID, Border>,
+    bordersPositions: Record<number, Zone[]>
+  ) {
+    for (const [borderId, border] of Object.entries(borders)) {
```
hokolomopo, line 313: nitpick: Object.entries isn't great performance-wise (create a new array for each element). Same in the other methods

src/plugins/ui_feature/autofill.ts
```diff
@@ -238,6 +247,115 @@ export class AutofillPlugin extends UIPlugin {
...
+      const cf = this.getters.getConditionalFormats(sheetId).find((cf) => cf.id === cfId);
+      if (!cf) {
+        continue;
+      }
+      const newCfRanges = this.getters.getAdaptedCfRanges(sheetId, cf, changes, []);
+      if (newCfRanges) {
+        this.dispatch("ADD_CONDITIONAL_FORMAT", {
+          cf: {
+            id: cf.id,
+            rule: deepCopy(cf.rule),
```
hokolomopo, line 333: deepCopy probably unnecessary ?

src/plugins/ui_feature/autofill.ts
```diff
@@ -407,13 +522,8 @@ export class AutofillPlugin extends UIPlugin {
...
   }
 
-  private autoFillMerge(
-    originCol: HeaderIndex,
-    originRow: HeaderIndex,
-    col: HeaderIndex,
-    row: HeaderIndex
-  ) {
-    const sheetId = this.getters.getActiveSheetId();
+  private autofillMerge(sheetId: UID, data: Omit<AutoFillCellCommand, "type">) {
```
hokolomopo, line 525: nitpick: I'd create a type `AutofillCellData` at this point, `Omit<AutoFillCellCommand, "type">` is used like 7 times

tests/test_helpers/commands_helpers.ts
```diff
@@ -546,6 +546,19 @@ export function setBorders(
   });
 }
 
+export function setBorderTarget(
```
hokolomopo, line 549: rename this if you end up renaming the command

hokolomopo (commented):
A bunch of neetpicks (feel free to ignore them).

But I feel like we could do something a bit simpler for Borders, it seems a bit convoluted to create a `Border => Id` mapping just so we can create a `Id => Zone[]` mapping. Either use the stringified border as Id, or just use an array

src/plugins/ui_feature/autofill.ts
hokolomopo: That's funny I preferred the original code, calls to `getActiveSheetId` are free and I'd rather not clutter method arguments if not needed. Either way that's a nitpick, I don't care that much 🤷

hokolomopo (approved):

VincentSchippefilt:
robodoo r+

PR #6083, opened 2025-04-11
[IMP] pivot: add collapse/expand icons to pivot

hokolomopo:
### [IMP] pivot: add collapse/expand icons to pivot

This commit adds icons to collapse/expand dimensions of a pivot.

The computation of the collapsed state of the pivot is done in the
`SpreadsheetPivotTable`. But since some features need the full,
non-collapsed pivot (eg. calculated measures), we now need two
different `SpreadsheetPivotTable` per pivot, `pivot.getTableStructure()`
and `pivot.getNonCollapsedTableStructure()`.

### [REF] pivot: use empty icon instead of format form pivot indent

The indentation of sub-groups in the pivot row headers is currently
done by applying a format to the cell. This has two drawbacks:
1. referencing a a row header cell would give a formatted value with
    the indentation, which is not something we want outside of the pivot
2. it's not compatible with having expand/collapse icons inside the
    pivot. We want the indent to be before the icon, not between the
    icon and the text.

This commit changes the way the indentation is done. Now we add
empty icons in the `iconsOnCellRegistry` with a variable margin to
get the desired indentation.

### [REF] icons: use same registry for all icons

At the moment the implementation of the handling of icons in the grid
is different for every type of icon. CF icons are inside the
`iconsOnCellRegistry`. The filter/data validation icons are both
in the `doesCellHaveGridIcon` getter, but they are displayed in two
different overlays, and are handled separately from the CFs in the
renderer.

This commit put all icons in the `iconsOnCellRegistry`, handle them the
same way in the renderer and display them in the same overlay.

### [FIX] pivot: get column sub-total of calculated measure

On a pivot with columns grouped by, for example, `year` and `customer`,
the formula `=PIVOT(1, "calculated measure", "year", 2024)` would
always return 0. It worked if there were either no or a full column
domain in the formula, but a partial column domain returned 0 instead
of returning the sub-total of the calculated measure for the partial
column domain.

Task: [4674387](https://www.odoo.com/odoo/2328/tasks/4674387)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/types/pivot_runtime.ts
```diff
@@ -20,6 +20,7 @@ export interface Pivot<T = PivotRuntimeDefinition> {
   onDefinitionChange(nextDefinition: PivotCoreDefinition): void;
 
   getTableStructure(): SpreadsheetPivotTable;
+  getNonCollapsedTableStructure(): SpreadsheetPivotTable;
```
hokolomopo, line 23: maybe it should be the other way around and `getTableStructure` should return the non-collapsed table, it seems to be a bit more used. Or change `getTableStructure` to `getCollapsedTableStructure` so there's no confusion ?

hokolomopo: Good idea, updated :)

pro-odoo:
push-f to rebase, there was a conflict with the getters

pro-odoo (approved):
robodoo rebase-ff r+

PR #6089, opened 2025-04-11
18.0 function module ref assert and errors laa

laa-odoo:
Task: [4671645](https://www.odoo.com/odoo/2328/tasks/4671645)

### DIFF
--- a/src/functions/helper_assert.ts
+++ b/src/functions/helper_assert.ts
@@ -1,5 +1,5 @@
 import { Arg, FunctionResultNumber, FunctionResultObject, Matrix, isMatrix } from "../types";
-import { EvaluationError } from "../types/errors";
+import { CellErrorType } from "../types/errors";
 import { assert } from "./helpers";
 
 export function assertSingleColOrRow(errorStr: string, arg: Matrix) {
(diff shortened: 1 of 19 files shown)


LucasLefevre (commented):
mouais, a bit weird to throw something else than an error. But I must admit `assert` is easy and removing it would be a huge diff

Alternative with a super simple diff doing pretty much the same thing: just remove `extends Error`

```patch
diff --git a/src/types/errors.ts b/src/types/errors.ts
index 519a9b91e..39fc0627c 100644
--- a/src/types/errors.ts
+++ b/src/types/errors.ts
@@ -14,10 +14,8 @@ export const CellErrorType = {
 
 export const errorTypes: Set<string> = new Set(Object.values(CellErrorType));
 
-export class EvaluationError extends Error {
-  constructor(message = _t("Error"), readonly value: string = CellErrorType.GenericError) {
-    super(message);
-  }
+export class EvaluationError  {
+  constructor(readonly message = _t("Error"), readonly value: string = CellErrorType.GenericError) {}
 }
 
 export class BadExpressionError extends EvaluationError {
```

LucasLefevre:
also, we have a bunch of

```ts
try {
catch (e) {
      if (e instanceof EvaluationError) {
        // ...
      }
    }
```

laa-odoo:
done here https://github.com/odoo/o-spreadsheet/pull/6176

PR #6090, opened 2025-04-14
[FIX] clipboard: huge revision on pasting borders

hokolomopo:
## Description

When using the clipboard to copy/paste borders, we would do a `SET_BORDER` command for each cell in the selection. This can create quite a large revision.

This commit batches the similar `SET_BORDER` commands into a single `SET_BORDER_ON_TARGET` command. This reduces the size of the revisions.

Benchmark:
------------
Copy/paste a row with 100 empty cells with borders: Before: 32.5kB revision
After: 8.2kB revision, ~4

Task: [4718522](https://www.odoo.com/odoo/2328/tasks/4718522)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


VincentSchippefilt:
robodoo r+

PR #6092, opened 2025-04-14
[IMP] chart: detect hierarchical chart at chart creation

hokolomopo:
## Description

With this commit, at the chart creation we will check if the data in the selected zone looks like a hierarchical dataset (multiple columns with string values, and one column with numeric values). If so, we will create a sunburst chart instead of a car chart.

Task: [4724420](https://www.odoo.com/odoo/2328/tasks/4724420)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/figures/charts/chart_factory.ts
+++ b/src/helpers/figures/charts/chart_factory.ts
@@ -13,7 +13,7 @@ import {
   UID,
   Zone,
 } from "../../../types";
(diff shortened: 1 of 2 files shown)


pro-odoo (approved):
robodoo r+

PR #6111, opened 2025-04-15
Master dv icons perfs canvas adrm

hokolomopo:
## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/stores/grid_renderer_store.ts
```diff
@@ -338,7 +345,15 @@ export class GridRenderer {
     const { ctx } = renderingContext;
     for (const box of boxes) {
       for (const icon of Object.values(box.icons)) {
-        if (!icon || !icon.svg) {
+        if (!icon) {
+          continue;
```
LucasLefevre, line 349: the icon can be falsy ?

hokolomopo: image

box.icons could be `{left: undefined, right: anIcon}` as per the typing, thus the icon would be undefined. Not sure if it really happens in practice, but if I added the condition I guess it did at some point

src/stores/grid_renderer_store.ts
```diff
@@ -347,48 +362,18 @@ export class GridRenderer {
...
-        }
-        ctx.translate(x, iconY);
+        const { x, y } = this.getters.getCellIconRect(icon);
+        ctx.translate(x, y);
         ctx.scale(iconSize / svg.width, iconSize / svg.height);
-        ctx.fillStyle = svg.fillColor;
-        ctx.fill(new Path2D(svg.path));
+        for (const path of svg.paths) {
+          ctx.fillStyle = path.fillColor;
+          ctx.fill(new Path2D(path.path));
```
LucasLefevre: could be cached or moved to each svg path. It looks like it's the first CPU consumer :deciduous_tree: 
image
(profile when I was scrolling on a sheet full of checkboxes)

src/components/helpers/drag_and_drop.ts
```diff
@@ -7,7 +7,7 @@ type EventFn = (ev: PointerEvent) => void;
  */
 export function startDnd(onPointerMove: EventFn, onPointerUp: EventFn) {
   const removeListeners = () => {
-    window.removeEventListener("pointerup", _onPointerUp);
+    window.removeEventListener("pointerup", _onPointerUp, { capture: true });
```
hokolomopo, line 10: @anhe-odoo might interest you for your timeline PR. We could also use capture for pointerMove

PR #6122, opened 2025-04-15
[FIX] pivot: get column sub-total of calculated measure

hokolomopo:
## Description

On a pivot with columns grouped by, for example, `year` and `customer`, the formula `=PIVOT(1, "calculated measure", "year", 2024)` would always return 0. It worked if there were either no or a full column domain in the formula, but a partial column domain returned 0 instead of returning the sub-total of the calculated measure for the partial column domain.

Task: [4728690](https://www.odoo.com/odoo/2328/tasks/4728690)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


LucasLefevre (approved):
robodoo r+

PR #6127, opened 2025-04-17
Standalone composer fixes

hokolomopo:
### [FIX] pivot: add `invalid` indicator for wrong calculated measure

This commit adds an `invalid` indicator in the standalone composer of
the calculated measure if the formula is invalid.

Note that it's only in error if the compilation is invalid, we cannot
evaluate there as it will be different for each row/col of the pivot.

### [FIX] standalone composer: localize content

The content of the standalone composer was not localized.

### [IMP] tests: add additional jest matchers

This commit backports the `toHaveClass` and `toHaveAttribute` matchers.

Task: [4724069](https://www.odoo.com/odoo/2328/tasks/4724069)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/pivot/pivot_layout_configurator/pivot_measure/pivot_measure.xml
+++ b/src/components/side_panel/pivot/pivot_layout_configurator/pivot_measure/pivot_measure.xml
@@ -24,6 +24,7 @@
             composerContent="measure.computedBy.formula"
             defaultRangeSheetId="measure.computedBy.sheetId"
             contextualAutocomplete="getMeasureAutocomplete()"
+            invalid="isCalculatedMeasureInvalid"
           />
         </div>
       </div>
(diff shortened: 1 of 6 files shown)


src/components/side_panel/pivot/pivot_layout_configurator/pivot_measure/pivot_measure.xml
```diff
@@ -24,6 +24,7 @@
             composerContent="measure.computedBy.formula"
             defaultRangeSheetId="measure.computedBy.sheetId"
             contextualAutocomplete="getMeasureAutocomplete()"
+            invalid="isCalculatedMeasureInvalid"
```
rrahir, line 27: I'll let francois decide but it feels like the simple "o-invalid" doesn't draw enough attention

LucasLefevre (approved):
robodoo r+ rebase-ff

PR #6133, opened 2025-04-18
[FIX] collaborative: check all overlapping zones

LucasLefevre:
## Description:

Steps to reproduce:
- merge A2:A3 and F1:F2

Then, concurrently:
- Alice merges A1:A3 (increase the current merge) and C1:C2
- Bob removes merges in A2:A3 and F1:F2

Assuming Alice's revision arrives first to the server (`executed` in the code).
When Bob's revision is transformed against Alice's, both zones in Bob's
command are kept, which makes the command invalid since A2:A3 is no longer
a merge (it is now A1:A3). Because it's invalid, the entire command is
rejected, even though F1:F2 should still be un-merged to preserve Bob's
intention.

The merge transformation is wrong.
The current code keeps a zone if at least one of the zones in the executed
command doesn't overlap.

But it should check all zones and should be kept if none are overlapping.
Said differently: the zone should be dropped if it overlaps any of the
`executed.target` zone.

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/collaborative/ot/ot_helpers.ts
+++ b/src/collaborative/ot/ot_helpers.ts
@@ -21,5 +21,5 @@ export function transformZone<Z extends Zone | UnboundedZone>(
       executed.quantity
     );
   }
-  return { ...zone };
+  return zone;
 }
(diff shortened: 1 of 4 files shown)


pro-odoo:
robodoo rebase-ff r+

PR #6145, opened 2025-04-18
[FIX] spreadsheet: parasitic render at every click

hokolomopo:
## Description

Since the topbar refactor, a new render is triggered at every single click event, no matter where we click. This breaks every input with `t-att-value`, their content is reset at each click.

Task: [4737827](https://www.odoo.com/odoo/2328/tasks/4737827)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/top_bar/top_bar.ts
+++ b/src/components/top_bar/top_bar.ts
@@ -250,7 +250,9 @@ export class TopBar extends Component<Props, SpreadsheetChildEnv> {
   }
 
   private openMenu(menu: Action, ev: MouseEvent) {
-    this.topBarToolStore.closeDropdowns();
+    if (this.topBarToolStore.currentDropdown) {
+      this.topBarToolStore.closeDropdowns();
+    }
     this.state.toolsPopoverState.isOpen = false;
     this.state.menuState.isOpen = true;
     this.state.menuState.anchorRect = getBoundingRectAsPOJO(ev.currentTarget as HTMLElement);
@@ -264,7 +266,9 @@ export class TopBar extends Component<Props, SpreadsheetChildEnv> {
   }
 
   closeMenus() {
-    this.topBarToolStore.closeDropdowns();
+    if (this.topBarToolStore.currentDropdown) {
+      this.topBarToolStore.closeDropdowns();
+    }
     this.state.toolsPopoverState.isOpen = false;
     this.state.menuState.isOpen = false;
     this.state.menuState.parentMenu = undefined;
@@ -290,7 +294,9 @@ export class TopBar extends Component<Props, SpreadsheetChildEnv> {
   }
 
   toggleMoreTools() {
-    this.topBarToolStore.closeDropdowns();
+    if (this.topBarToolStore.currentDropdown) {
+      this.topBarToolStore.closeDropdowns();
+    }
     this.state.toolsPopoverState.isOpen = !this.state.toolsPopoverState.isOpen;
   }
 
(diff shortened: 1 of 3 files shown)


src/components/top_bar/top_bar.ts
```diff
@@ -250,7 +250,9 @@ export class TopBar extends Component<Props, SpreadsheetChildEnv> {
   }
 
   private openMenu(menu: Action, ev: MouseEvent) {
-    this.topBarToolStore.closeDropdowns();
+    if (this.topBarToolStore.currentDropdown) {
+      this.topBarToolStore.closeDropdowns();
```
rrahir, line 254: I think the inputs should be somewhat more resilient. Not sayning that we should not fix the calls here but any full render is a source of problem, which means any message received by the network is as well.

hokolomopo: Beside never using `t-att-value` I'm not sure what else we can do ...

tests/figures/chart/charts_component.test.ts
```diff
@@ -157,7 +157,10 @@ describe("charts", () => {
         },
       ],
     };
+    jest.useFakeTimers();
     model = new Model(data, { external: { geoJsonService: mockGeoJsonService } });
+    jest.advanceTimersByTime(DEBOUNCE_TIME + 10); // wait for the debounce of session.move
+    jest.useRealTimers();
```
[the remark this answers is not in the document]
hokolomopo: Actually scratch that, this *is* a problem. I'm 80% sure that's why the composer component tests keep randomly failing on my machine ....

src/components/top_bar/top_bar.ts
rrahir: well it should write to a state and react at each onchange I gues, or we could imagine some sort of partial rendering / reorganise the component structure to have the sidepanel not subject to the hard render eveytime something occurs. @VincentSchippefilt  Maybe a good discussion for next week

hokolomopo: It's also a problem for the topbar font selection input, so not only the side panel. But yeah idk

tests/top_bar_component.test.ts
```diff
@@ -975,10 +975,10 @@ test("Clicking on a topbar button triggers three renders", async () => {
 
   await click(fixture, ".o-spreadsheet-topbar [title='Bold (Ctrl+B)']");
 
-  // two renders from the collaboration session
+  // one renders from the collaboration session
```
rrahir, line 978: the comment does not match the actual situation (while I agree it's a nightmare here)

tests/figures/chart/charts_component.test.ts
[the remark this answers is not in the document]
rrahir: I don't recall if we found a idea to solve this? I only recall the ideas:
- skipping the first debounce
- having an overall mock of the session in the tests but it lead to the import nightmare...

rrahir (commented):
🚚  50/50 (I hope) on the comments relevance

tests/figures/chart/charts_component.test.ts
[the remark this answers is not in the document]
hokolomopo: Fixed in another PR :) (In master, because it needed some tweaks in the file structure)

https://github.com/odoo/o-spreadsheet/pull/6162

src/components/top_bar/top_bar.ts
hokolomopo: https://www.odoo.com/odoo/2328/tasks/4756754

tests/top_bar_component.test.ts
hokolomopo: Indeed, I fixed the comment :)

rrahir:
@robodoo r+

PR #6162, opened 2025-04-22
Session & debounce

hokolomopo:
### [FIX] tests: indeterministic tests with `debounce`

When creating a model, we trigger a `session.move` that is debounced,
and that trigger a render when the debounce is over. This can lead to
indeterministic tests, with random renders happening in the middle of
a test.

This commit removes the debounce of `session.move` for all the tests.

### [IMP] registries: remove registries index

We had a `index.ts` file for some registries. The file was a bit of a
problem because:

- It mixes apples and oranges: some registries are purely UI, some not,
    so importing everything at once don't really make sense.
- It is an easy source of circular dependencies. For example, we had
    to split `side_panel_registry` in two files to avoid circular dependencies

This commit removes the `index.ts` file and imports the registries
individually where needed.

### [IMP] session: better handling of `session.move`

Currently, `session.move` do nothing if `session.join` is not called
beforehand. But as it turns out, we do actually call `session.move`
before `session.join`. It just always work in practice, because
`session.move` is debounced.

With this commit, we store the client position in `session.move`, and
send it to the server when `session.join` is called.

Task: [4737879](https://www.odoo.com/odoo/2328/tasks/4737879)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


rrahir:
@robodoo rebase-ff

LucasLefevre (approved):
robodoo r+

PR #6173, opened 2025-04-24
[FIX] grid_overlay: stop interval on error

hokolomopo:
## Description

If there is a bug somewhere and that an error is raised in the model, every subsequent call to a model getter might raise an error since the model is not in a valid state.

This is a bit of a problem in the grid_overlay where we have an interval running every 200ms. If the model is in error and the interval is running, we will raise a new error every 200ms. This is less than ideal.

This commit stops the interval when an error is raised in its callback.

Task: [4746478](https://www.odoo.com/odoo/2328/tasks/4746478)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/helpers/time_hooks.ts
+++ b/src/components/helpers/time_hooks.ts
@@ -11,21 +11,30 @@ interface IntervalTimer {
 export function useInterval(callback: () => void, delay: number): IntervalTimer {
   let intervalId: number | undefined;
   const { setInterval, clearInterval } = window;
+  const pause = () => {
+    clearInterval(intervalId);
+    intervalId = undefined;
+  };
+  const safeCallback = () => {
+    try {
+      callback();
+    } catch (e) {
+      pause();
+      throw e;
+    }
+  };
   useEffect(
     () => {
-      intervalId = setInterval(callback, delay);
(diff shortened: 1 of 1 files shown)


src/components/grid_overlay/grid_overlay.ts
```diff
@@ -38,16 +38,21 @@ function useCellHovered(
...
   function checkTiming() {
-    const { col, row } = getPosition();
-    const delta = Date.now() - lastMoved;
-    if (delta > 300 && (col !== hoveredPosition.col || row !== hoveredPosition.row)) {
-      setPosition(undefined, undefined);
-    }
-    if (delta > 300) {
-      if (col < 0 || row < 0) {
-        return;
+    try {
```
LucasLefevre, line 41: the try/catch should be moved to `useInterval` IMO

hokolomopo: Was thinking what if we want to have the useInterval callback to handle thrown errors ? But I was stupid, we can still do a try/catch in the callback 🙈

LucasLefevre (approved):
robodoo r+

PR #6176, opened 2025-04-24
[PERF] formula functions: avoid to use errors derived from native error.

laa-odoo:
Task: [4671645](https://www.odoo.com/odoo/2328/tasks/4671645)


LucasLefevre (commented):
robodoo rebase-ff r+

PR #6209, opened 2025-04-25
[FIX] MOVE_COLUMNS_ROWS: adapt range on col/row move

fdamhaut:
Steps to reproduce:

    Enter the formula =SUM(C1, C2) in any cell.
    Drag column C and drop it after column H.
    The formula now results in a #REF! error, indicating broken references.

Task: 4743039

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_stateful/clipboard.ts
+++ b/src/plugins/ui_stateful/clipboard.ts
@@ -316,8 +316,6 @@ export class ClipboardPlugin extends UIPlugin {
       "figureId" in data
         ? clipboardHandlersRegistries.figureHandlers
         : clipboardHandlersRegistries.cellHandlers;
-    if ("figureId" in data) {
-    }
     return handlersRegistry.getKeys().map((handlerName) => {
       const Handler = handlersRegistry.get(handlerName);
       return { handlerName, handler: new Handler(this.getters, this.dispatch) };


tests/sheet/sheet_manipulation_plugin.test.ts
```diff
@@ -1188,6 +1233,45 @@ describe("Rows", () => {
...
+              D4: { content: "=B3+C3" },
+            },
+          },
+        ],
+      });
+      moveRows(model, 2, [0], "after");
+      expect(getCellsObject(model, "sheet1")).toMatchSnapshot();
+    });
+
+    test("On move", () => {
```
hokolomopo, line 1269: Why two tests in the same describe with the exact same name ? What do they do differently ? whats the point of this test compared to the previous one ? I'm confused, be more explicit with test names

tests/sheet/sheet_manipulation_plugin.test.ts
```diff
@@ -200,6 +202,49 @@ describe("Columns", () => {
...
+            cells: {
+              A1: { content: "A Col" },
+              A2: { content: "1" },
+              A3: { content: "=A2" },
+              A4: { content: "=A2+A3" },
+              B1: { content: "B Col" },
+              B4: { content: "=A2+A3" },
+              C1: { content: "C Col" },
+              C4: { content: "2" },
+              D4: { content: "=A2+A3" },
```
hokolomopo, line 222: isn't 80% of this content useless for your testing purpose ? You only want to test that references are updated right? there should already be tests for moving random stuff in columns

tests/sheet/sheet_manipulation_plugin.test.ts
```diff
@@ -1188,6 +1233,45 @@ describe("Rows", () => {
...
+              A4: { content: "R4" },
+              B4: { content: "=B1+C1" },
+              C4: { content: "=B2+C2" },
+              D4: { content: "=B3+C3" },
+            },
+          },
+        ],
+      });
+      moveRows(model, 2, [0], "after");
+      expect(getCellsObject(model, "sheet1")).toMatchSnapshot();
```
hokolomopo, line 1266: why is this test using snapshot while the previous one is not ? Just don't use snapshot tests when not needed
- it makes it harder to read the test. I have no clue if the references were correctly moved reading this
- it makes us 10x more likely to ignore a bug, and just update the snapshots

tests/sheet/sheet_manipulation_plugin.test.ts
```diff
@@ -200,6 +202,49 @@ describe("Columns", () => {
       expect(model.getters.getColSize(sheetId, 1)).toBe(DEFAULT_CELL_WIDTH);
       expect(model.getters.getActiveSheet().numberOfCols).toBe(2);
     });
+
```
hokolomopo, line 205: all the tests should be moved with the other tests testing moving columns/rows, I don't think they belong here at all. And the test should be a bit more concise, I'm pretty sure you don't need 250 lines of tests to test a 3 lines fix.

src/plugins/ui_stateful/selection.ts
```diff
@@ -582,6 +582,14 @@ export class GridSelectionPlugin extends UIPlugin {
       currentIndex += 1;
     }
 
+    this.dispatch("MOVE_RANGES", {
```
hokolomopo, line 585: Sure but ... isn't MOVE_RANGES already dispatched inside the cut/paste ? Do you know why it isn't working ? Is cut/paste also broken in the grid ?

hokolomopo (commented):
bip bop 🤖

src/plugins/ui_stateful/selection.ts
fdamhaut: > isn't MOVE_RANGES already dispatched inside the cut/paste ?
Not since the 17.4 refactor

Ty for the review, I indeed focused on solving the problem at hand and didn't see that the clipboard only copied the content of the cells and not the rest :/

tests/sheet/sheet_manipulation_plugin.test.ts
fdamhaut: > all the tests should be moved with the other tests testing moving columns/rows,

There are only 2 test file with moveRow (or dispatch to move_colrow) not related to OT and I don't think the selection plugin test is a better place.

The test are in the suite "Rows/Cols > Correctly update references > On move", next to "Col/Row addition/deletion". It's an appropriate place

hokolomopo: The MOVE_COLUMNS_ROWS feature and your fix are in the selection plugin, why woudln't you want to test it in the selection plugin test file 😅 

The other tests in this file are related to core sheet manipulation add/remove columns). On the outside moveColumns sounds like the same, but it really isn't. It isn't a core command, it's just the selection UI plugin doing copy/paste things and make it look like a column moved.

PR #6210, opened 2025-04-25
[FIX] ts: fix compiling of `.d.ts` files

hokolomopo:
It wasn't possible to compile the lib into `.d.ts` since 8a02a6b, error `... exported class expression may not be private or protected`.[1]

It seems that we cannot have an anonymous class with private/protected methods. We had a similar issue with the `PivotPresentationLayer`.

This commit adds explicit typing to the functions generating the funnel chart classes, so typescript doesn't try to infer types for them.

[1]: see https://github.com/microsoft/TypeScript/issues/30355 or https://github.com/microsoft/TypeScript/issues/36060

Task: 0

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/figures/chart/chartJs/chartjs_funnel_chart.ts
+++ b/src/components/figures/chart/chartJs/chartjs_funnel_chart.ts
@@ -1,13 +1,20 @@
 import {
+  BarController,
   BarControllerChartOptions,
   BarControllerDatasetOptions,
   BarElement,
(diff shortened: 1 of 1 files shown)


pro-odoo (commented):
robodoo r+

PR #6211, opened 2025-04-25
[FIX] pivot: spreadsheet pivot with date fields

hokolomopo:
### [FIX] pivot: date dimension detection

A pivot field was detected with type `date` if all of the cells had a
date format, but we didn't check if the cell values were actually
dates.

Task: [4742696](https://www.odoo.com/odoo/2328/tasks/4742696)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/pivot/pivot_side_panel/pivot_side_panel_store.ts
+++ b/src/components/side_panel/pivot/pivot_side_panel/pivot_side_panel_store.ts
@@ -243,7 +243,7 @@ export class PivotSidePanelStore extends SpreadsheetStore {
       granularitiesPerFields[field.name] = new Set(granularities);
     }
     for (const field of dateFields) {
-      granularitiesPerFields[field.name].delete(field.granularity);
+      granularitiesPerFields[field.name].delete(field.granularity || "month");
     }
     return granularitiesPerFields;
   }
--- a/src/helpers/pivot/pivot_helpers.ts
+++ b/src/helpers/pivot/pivot_helpers.ts
@@ -229,10 +229,7 @@ export function toNormalizedPivotValue(
 }
 
 function normalizeDateTime(value: CellValue, granularity: Granularity) {
-  if (!granularity) {
-    throw new Error("Missing granularity");
-  }
-  return pivotTimeAdapter(granularity).normalizeFunctionValue(value);
+  return pivotTimeAdapter(granularity ?? "month").normalizeFunctionValue(value);
 }
 
 export function toFunctionPivotValue(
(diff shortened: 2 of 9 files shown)


src/helpers/pivot/spreadsheet_pivot/date_spreadsheet_pivot.ts
```diff
@@ -6,7 +6,7 @@ import { toNormalizedPivotValue } from "../pivot_helpers";
 const NULL_SYMBOL = Symbol("NULL");
 
 export function createDate(dimension: PivotDimension, value: CellValue, locale: Locale): CellValue {
-  const granularity = dimension.granularity;
+  const granularity = dimension.granularity || "month";
   if (!granularity || !(granularity in MAP_VALUE_DIMENSION_DATE)) {
```
LucasLefevre, line 10: `!granularity || ` can be removed

src/helpers/pivot/pivot_helpers.ts
```diff
@@ -230,7 +230,7 @@ export function toNormalizedPivotValue(
 
 function normalizeDateTime(value: CellValue, granularity: Granularity) {
   if (!granularity) {
-    throw new Error("Missing granularity");
+    granularity = "month";
```
LucasLefevre, line 233: slightly shorter and easier with `pivotTimeAdapter(granularity ?? "month")` instead of reassigning the argument

src/components/side_panel/pivot/pivot_side_panel/pivot_side_panel_store.ts
```diff
@@ -215,13 +217,18 @@ export class PivotSidePanelStore extends SpreadsheetStore {
     const { columns, rows } = definition;
     const columnsWithGranularity = deepCopy(columns);
     const rowsWithGranularity = deepCopy(rows);
-    const unusedGranularities = this.getUnusedDateTimeGranularities(fields, definition);
+    const unusedGranularities = this.getUnusedDateTimeGranularities(fields, definition, undefined);
     for (const dimension of columnsWithGranularity.concat(rowsWithGranularity)) {
       const fieldType = fields[dimension.name]?.type;
       if ((fieldType === "date" || fieldType === "datetime") && !dimension.granularity) {
-        const granularity = unusedGranularities[dimension.name]?.values().next().value || "year";
```
LucasLefevre, line 222: I'd keep the previous simpler code. We don't care if month or year is the first default value through the side panel UI. No need for a special case

LucasLefevre, line 222: after thinking about it a second time and discussing with @pro-odoo . I'd rather live with the simpler code even if we have the strange behavior where "month" becomes "year". It's ok IMO, the user is updating and breaking the dimensions anyway.
Not that the code is really difficult, but the flow for which it matters is tordu (the "why" is this code like that).

hokolomopo: Sure. I added a comment in the test to say it's foireux

PR #6220, opened 2025-04-28
[FIX] pivot: copy from referenced id

LucasLefevre:
Steps to reproduce

- insert a pivot into a spreadsheet
- type the formula =PIVOT(A1) where A1 contains the pivot id
- copy any value from the pivot
- paste it somewhere else => the "fixed" formula should be pasted, but only the value is pasted

The bug was introduced by 816977d7b8a5

`...token` doesn't work as expected. For reference tokens coming stored on the cells, the objects are not POJO, but instances of `RangeReferenceToken`. The `value` property is a getter `get value()` which is ignored when spreading the object (=> `.value` is undefined)

Task: 4756878

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/formulas/parser.ts
+++ b/src/formulas/parser.ts
@@ -323,7 +323,11 @@ export function parse(str: string): AST {
 }
 
 export function parseTokens(tokens: Token[]): AST {
-  const richTokens = tokens.map((token, index) => ({ ...token, tokenIndex: index }));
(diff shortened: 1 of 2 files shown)


pro-odoo (commented):
robodoo r+

PR #6231, opened 2025-04-29
[IMP] merge: allow merging multiple selected zones

dhrp-odoo:
## Description:

Previously, merging was restricted to a single contiguous selection.

Now, multiple non-overlapping zones can be merged at once.

If the first selected zone contains merged cells among the selected cells, all selected zones will be unmerged instead.

Task: [4179256](https://www.odoo.com/odoo/2328/tasks/4179256)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/actions/edit_actions.ts
+++ b/src/actions/edit_actions.ts
@@ -1,4 +1,4 @@
-import { isEqual, positionToZone } from "../helpers";
+import { doesAnyZoneCrossFrozenPane, getZoneArea, hasOverlappingZones } from "../helpers";
 import { interactiveCut } from "../helpers/ui/cut_interactive";
 import { interactiveAddMerge } from "../helpers/ui/merge_interactive";
 import { handlePasteResult } from "../helpers/ui/paste_interactive";
@@ -147,7 +147,7 @@ export const deleteCellShiftLeft: ActionSpec = {
 export const mergeCells: ActionSpec = {
   name: _t("Merge cells"),
   isEnabled: (env) => !cannotMerge(env),
-  isActive: (env) => isInMerge(env),
+  isActive: (env) => hasMergeInAnySelectedZone(env),
   execute: (env) => toggleMerge(env),
   icon: "o-spreadsheet-Icon.MERGE_CELL",
 };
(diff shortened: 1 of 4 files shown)


src/actions/edit_actions.ts
```diff
@@ -170,31 +170,31 @@ function cannotMerge(env: SpreadsheetChildEnv): boolean {
...
+function isUnmergeAction(env: SpreadsheetChildEnv): boolean {
+  if (cannotMerge(env)) {
+    return false;
   }
-  return false;
+
+  const firstZone = env.model.getters.getSelectedZones()[0];
+  const sheetId = env.model.getters.getActiveSheetId();
+  const mergesInFirstZone = env.model.getters.getMergesInZone(sheetId, firstZone);
+  return mergesInFirstZone.length > 0;
```
hokolomopo, line 188: you change the current behavior with a single selected zone that contains a merge. Before we would expand the merge to contain all of the the selection, now you'll just remove the merge.

That's a bit strange considering that we do the opposite when there's multiple selections. Not sure if that's what you discussed with Francois

hokolomopo: Also just checking only if the first zone has merges seems a bit arbitrary... But idk how we can be consistent if we keep the "expand merge" feature...

Maybe if any zone in the selection has a merge we could remove all of the merges, and if there's no merge in the selection we'd merge every zone. That'd be consistent at least, and I think that's what excel does ? And we'd drop the "expand merge" feature" To discuss with Francois.

tests/cells/merges_plugin.test.ts
```diff
@@ -554,12 +554,6 @@ describe("merges", () => {
     expect(unMerge(model, "A1:A2")).toBeCancelledBecause(CommandResult.InvalidTarget);
   });
 
-  test("un-merge zone overlapping another merge is refused", () => {
```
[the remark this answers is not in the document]
hokolomopo: You could also in the action get all the merges overlapping the selection, and dispatch a REMOVE_MERGE with all of the merges you found. That way you don't have to change at all how the core command behaves. Would that work ? 

Changing how a core command behave is a bit dangerous, it could possibly break some spreadsheets :x

hokolomopo (commented):
👋

src/actions/edit_actions.ts
Francois-frgi: Hi @dhrp-odoo, let's keep what you did. It is seems consistent. Thanks!

hokolomopo (commented):
hello hello 👋

src/actions/edit_actions.ts
```diff
@@ -170,32 +170,36 @@ function cannotMerge(env: SpreadsheetChildEnv): boolean {
...
```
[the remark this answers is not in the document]
hokolomopo: Maybe he didn't test everything ? I talked with him again and he agreed that we sould do the logic "Any merge inside any zone -> remove merges, else add merges"

hokolomopo (commented):
Seems good ! Thanks

@robodoo r+

hokolomopo:
@robodoo rebase-ff

PR #6248, opened 2025-04-29
Master formula function ref assert and errors laa

laa-odoo:
Task: [4760596](https://www.odoo.com/odoo/2328/tasks/4760596)

### DIFF
--- a/src/functions/helper_assert.ts
+++ b/src/functions/helper_assert.ts
@@ -1,31 +1,42 @@
+import { _t } from "../translation";
 import { Arg, FunctionResultNumber, FunctionResultObject, Matrix, isMatrix } from "../types";
-import { EvaluationError } from "../types/errors";
-import { assert } from "./helpers";
+import { DivisionByZeroError, EvaluationError } from "../types/errors";
 
-export function assertSingleColOrRow(errorStr: string, arg: Matrix) {
-  assert(() => arg.length === 1 || arg[0].length === 1, errorStr);
+export function assert(condition: boolean, message: string): asserts condition {
+  if (!condition) {
+    throw new EvaluationError(message);
+  }
 }
 
-export function assertSameDimensions(errorStr: string, ...args: Arg[]) {
+export function assertNotZero(
+  value: number,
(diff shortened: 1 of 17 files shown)


laa-odoo:
> I apologize if I missed something obvious, I justquickly went over the diff. Is there a reason why we keep some asserts in given situations? If so, it could be worth mentioning it somewhere (commit message and/or codebase if relevant?)
> 
> unrelated to the content but still worth mentioning:
> 
> * there are typos in the commit message
> * @anhe-odoo is not properly credited as a co-author, the correct syntax is `{username} <{githubAccountemail}>`
>   in this case, `Anthony Hendrickx (anhe) <anhe@odoo.com>`

**Translating French text**

The user wants me to translate this French text into basic English. It looks like an easy task, no need for additional searches or checking details. I’ll keep it straightforward and use simple language for the translation.

Here’s the English translation:

"Initially, the time spent on calculating error management for formulas was unusually high."


Thought for 5 seconds


Initially, the compute time spent handling errors in formulas was unusually high.
This happened mainly for two reasons:

* Creating errors from native JavaScript errors is very costly.
* Throwing errors is also costly (but still less than the first point).

At the same time, we had a unification problem: in all the code that implements formulas, we had several ways to handle errors:

* Using `throw` with the error we want
* Using `return` with the error we want
* Using `assert`, which throws specific errors (more elegant than a simple `throw` after a condition)

Note that there is also a technical constraint: it is not possible to use `return error` in the helpers that formulas use. Helpers only can use `throw` and `assert`.

To try to fix both issues, a first proposal was made in this PR:

* We fixed the heavy performance cost by avoiding creating errors based on native JS errors.
* We unified error handling into a single method: using `assert`.

After that proposal, it was decided that we want to maximize performance even if it means giving up unification. So we chose to return errors instead of throwing/asserting where possible (in the compute functions), and we keep throwing/asserting in the helper functions.

Now, the question you raise is interesting. Since there are only about thirty `assert` calls left in the helpers, maybe we could replace them with `throw`. I can do that in another commit  but this is becoming nitpicking (IMO)

laa-odoo:
@rrahir  Thanks for comments on the commit message :)

LucasLefevre:
robodoo r+

PR #6284, opened 2025-05-02
tocartesian

LucasLefevre:
## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/helpers/coordinates.ts
```diff
@@ -43,6 +48,41 @@ function isCharADigit(char: string) {
...
+  return letterPart;
+}
+
+export function consumeDigits(chars: TokenizingChars) {
+  if (chars.current === "$") chars.advanceBy(1);
+  if (!chars.current || !isCharADigit(chars.current)) {
+    return -1;
+  }
+  let num = 0;
+  while (chars.current && isCharADigit(chars.current)) {
```
LucasLefevre, line 80: interestingly, this `chars.current` is really important for performance (even though it could be removed and the code would behave exactly the same way).

Without `chars.current`, the input of `isCharADigit` can be a char or `undefined`. With `chars.current` the inputs are always chars and it looks like the JS engine is able to inline or heavily optimize the function and it's ~3x faster

src/helpers/coordinates.ts
```diff
@@ -28,13 +29,17 @@ export function lettersToNumber(letters: string): number {
...
-    const colIndex = charCode >= 65 && charCode <= 90 ? charCode - 64 : charCode - 96;
+    const colIndex = charToNumber(letters[i]);
     result = result * 26 + colIndex;
   }
   return result - 1;
 }
 
+function charToNumber(char: string) {
+  const charCode = char.charCodeAt(0);
+  return charCode >= 65 && charCode <= 90 ? charCode - 64 : charCode - 96;
```
hokolomopo, line 40: maybe throw if we receiver a non-letter character ?

src/helpers/coordinates.ts
```diff
@@ -43,6 +48,41 @@ function isCharADigit(char: string) {
...
+    chars.advanceBy(1);
+  }
+}
+
+export function consumeLetters(chars: TokenizingChars) {
+  if (chars.current === "$") chars.advanceBy(1);
+  if (!chars.current || !isCharALetter(chars.current)) {
+    return -1;
+  }
+  let letterPart = 0;
```
hokolomopo, line 66: nitpick: letterPart sounds like it would be the substring with the letters. Maybe just `colIndex` ?`currentcol` ?

src/helpers/coordinates.ts
```diff
@@ -53,37 +93,21 @@ function isCharADigit(char: string) {
...
-  if (xc[i] === "$") i++;
-  while (i < xc.length && isCharADigit(xc[i])) {
-    numberPart += xc[i++];
-  }
-
-  if (i !== xc.length || numberPart.length === 0 || numberPart.length > 7) {
-    // limit to max 7 numbers for performance reasons
-    throw new Error(`Invalid cell description: ${xc}`);
-  }
+  const num = consumeDigits(chars);
```
hokolomopo, line 105: case where consumeDigits returns -1 is not handled

src/helpers/zones.ts
```diff
@@ -111,7 +108,18 @@ export function toZoneWithoutBoundaryChanges(xc: string): UnboundedZone {
  */
 export function toUnboundedZone(xc: string): UnboundedZone {
   const zone = toZoneWithoutBoundaryChanges(xc);
-  return reorderZone(zone);
+  const orderedZone = reorderZone(zone);
+  const bottom = orderedZone.bottom;
+  const right = orderedZone.right;
+  if ((bottom !== undefined && bottom > MAX_ROW) || (right !== undefined && right > MAX_COL)) {
+    throw new Error(`Invalid range string: ${xc}`);
```
hokolomopo, line 115: could have a more precise error message

src/helpers/coordinates.ts
LucasLefevre: well, it turns out `toCartesian("A0")` was "valid" before this commit. At least it's valid from the regexp point of view. And there's bloody test failing if I handle -1 and rejects. I didn't bother changing this behavior, especially in stable

src/helpers/coordinates.ts
LucasLefevre: I wouldn't do it. We have typescript for most cases, and this is a hot path were perf matters.

src/helpers/coordinates.ts
```diff
@@ -43,6 +48,41 @@ function isCharADigit(char: string) {
...
+  }
+}
+
+export function consumeLetters(chars: TokenizingChars) {
+  if (chars.current === "$") chars.advanceBy(1);
+  if (!chars.current || !isCharALetter(chars.current)) {
+    return -1;
+  }
+  let letterPart = 0;
+  while (chars.current && isCharALetter(chars.current)) {
```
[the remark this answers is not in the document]
LucasLefevre: no because of `letterPart = letterPart * 26 + colIndex;` that would have a different result if we start at -1 instead of 0

hokolomopo:
@robodoo r+

hokolomopo:
@robodoo rebase-ff

PR #6292, opened 2025-05-05
[IMP] clipboard: copy whole pivot as dynamic pivot

hokolomopo:
## Description

Backport of a747d46

When we copy part of a dynamic pivot, we don't copy the values but pivot formulas instead.

But when copying (or cutting) the whole pivot, we should copy the dynamic pivot and not the pivot formulas. The intent of the user in this case is most likely to copy the pivot as a whole and not to replace it with pivot formulas (particularly when cutting).

Task: [4771494](https://www.odoo.com/odoo/2328/tasks/4771494)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


hokolomopo:
@robodoo fw=no

pro-odoo (approved):
robodoo r+

PR #6293, opened 2025-05-06
[FIX] Composer: keep focus with interacting with formula assistant

rrahir:
How to reproduce:
- start editing a cell content in the *topbar* composer with "=sum"
- click on the "x" icon of the formula assistant to close it

==> the composer was closed and the cell contains only "=sum" which is evaluated as a #BAD_EXPR formula

This seemed to work on the grid composer because the 'onBlur' handler would find a related target that has the `composerFocusableElement` attribute where it would not find an equivalent in the top bar composer.

Task: 4771772

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/composer/composer/composer.xml
+++ b/src/components/composer/composer/composer.xml
@@ -44,7 +44,11 @@
       <div
         class="o-composer-assistant-container shadow position-absolute z-1"
         t-att-style="assistantContainerStyle"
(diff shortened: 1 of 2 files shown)


pro-odoo (commented):
robodoo r+

PR #6294, opened 2025-05-06
[FIX] pivot: chart field with mix of text and numbers

hokolomopo:
## Description

If a field of a pivot contains a mix of text and numbers, the field is treated as a char field. But when using the field as a row/col of the pivot, the number values wouldn't be displayed at all.

This commit fixes the issue by putting string values in the data entries rather than the cell value for char fields.

Note: this means that we don't support mixed text/number values for measures anymore. This was always a niche case, and probably indicate that the dataset is not well designed.

Task: [4771482](https://www.odoo.com/odoo/2328/tasks/4771482)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/tests/pivots/spreadsheet_pivot/spreadsheet_pivot.test.ts
+++ b/tests/pivots/spreadsheet_pivot/spreadsheet_pivot.test.ts
@@ -513,21 +513,6 @@ describe("Spreadsheet Pivot", () => {
     );
   });
 
-  test("Sum with a field that contains a string should work", () => {
-    const model = createModelWithPivot("A1:I5");
-    updatePivot(model, "1", {
-      columns: [],
-      rows: [],
-      measures: [{ name: "Expected Revenue", aggregator: "sum" }],
-    });
-    setCellContent(model, "A26", `=pivot(1)`);
-    expect(getCellContent(model, "B28")).toBe("$17,500.00");
-
-    expect(getCellContent(model, "F2")).toBe("$2,000.00");
-    setCellContent(model, "F2", "Hello");
-    expect(getCellContent(model, "B28")).toBe("$15,500.00");
-  });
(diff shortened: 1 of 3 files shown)


tests/pivots/spreadsheet_pivot/spreadsheet_pivot.test.ts
```diff
@@ -477,21 +478,6 @@ describe("Spreadsheet Pivot", () => {
     );
   });
 
-  test("Sum with a field that contains a string should work", () => {
```
pro-odoo, line 480: Maybe we could add a test which explicitly ensure that the sum should not work ? With `git blame` we will find this commit :)

hokolomopo: Yeah ... but TBH I'm not too happy with the result. The aggregation result is set to "0", which isn't great...

1) It's fine
2) we should explicitly have the pivot cells be in error if the aggregator isn't supported by the field type
3) having sum work on a text field (by having rawValue or something in the data entry probably)

wdyt ?

pro-odoo: I would go for 2 if it's not too big. If it's too big, go for 1

hokolomopo: @pro-odoo added some commits. What do you think ?

pro-odoo: Nice, good for me :)

pro-odoo (approved):
robodoo r+

PR #6305, opened 2025-05-06
[FIX] pivot: pivot panel with multiple of same dimensions

hokolomopo:
## Description

It's possible to end up in a situation where multiple dimensions of the pivot are the same. This happens if we have multiple date dimensions with different granularities, and that the field change type to become a char/number field. This made the pivot panel crash.

Note: it's a really niche case. The fix only ensure the panel does not crash, however there is some weirdness in the pivot panel (drag & drop drags both dimensions, delete remove both dimensions, etc.). That's not ideal, but since the use case is very niche and the user can manually remove the duplicates, it's probably ok to leave it like that.

Task: [4775660](https://www.odoo.com/odoo/2328/tasks/4775660)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/pivot/pivot_layout_configurator/pivot_layout_configurator.xml
+++ b/src/components/side_panel/pivot/pivot_layout_configurator/pivot_layout_configurator.xml
@@ -9,7 +9,7 @@
           fields="props.unusedGroupableFields"
         />
       </div>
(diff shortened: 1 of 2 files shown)


pro-odoo (commented):
robodoo r+

PR #6327, opened 2025-05-08
[FIX] Stores: skip unnecessary renderings

rrahir:
Currently, calling the mutator of a Store (through `useStore`) will automatically trigger a full-on rendering of the spreadsheet. The rationale being that the sotre state **might** have been updated but this also mean that we will trigger a full render even if the call to the mutatore had 0 effect.

This revision adds the possibility for a store to specifcally notify that its state was not altered and that the render can therefore be skipped.

This leaves more room for granularity than a reactivity observer would and we avoid its overhead as well.

Task: 4781429

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/helpers/draw_grid_hook.ts
+++ b/src/components/helpers/draw_grid_hook.ts
@@ -42,10 +42,6 @@ export function useGridDrawing(refName: string, model: Model, canvasSize: () => 
 
     for (const layer of OrderedLayers()) {
       model.drawLayer(renderingContext, layer);
-      // @ts-ignore 'drawLayer' is not declated as a mutator because:
-      // it does not mutate anything. Most importantly it's used
-      // during rendering. Invoking a mutator during rendering would
-      // trigger another rendering, ultimately resulting in an infinite loop.
       rendererStore.drawLayer(renderingContext, layer);
     }
   }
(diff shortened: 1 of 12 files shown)


src/stores/DOM_focus_store.ts
```diff
@@ -4,5 +4,6 @@ export class DOMFocusableElementStore {
 
   setFocusableElement(element: HTMLElement) {
```
hokolomopo, line 5: didn't you want to make the `focusableElement` private again ?

src/components/helpers/draw_grid_hook.ts
```diff
@@ -42,10 +42,6 @@ export function useGridDrawing(refName: string, model: Model, canvasSize: () =>
 
     for (const layer of OrderedLayers()) {
       model.drawLayer(renderingContext, layer);
-      // @ts-ignore 'drawLayer' is not declated as a mutator because:
-      // it does not mutate anything. Most importantly it's used
-      // during rendering. Invoking a mutator during rendering would
-      // trigger another rendering, ultimately resulting in an infinite loop.
```
hokolomopo, line 48: :partying_face:

src/stores/DOM_focus_store.ts
rrahir: indeed ><

LucasLefevre (approved):
robodoo r+

PR #6334, opened 2025-05-09
[FIX] function: IF should not mutate its input

LucasLefevre:
## Description:

Steps to reproduce:

- add a formula =IF(TRUE,B1,B2) where B1 is empty
- add a formula below =ISBLANK(C1) where C1 is empty => the result of ISBLANK is false but it should be true.

The first IF changed the value of the shared EMPTY_CELL object from `null` to `""`.

A future step to prevent other similar issues would be to leverage Typescript readonly properties to mark all evaluated cells and intermediary results as readonly. But it's for an another task, probably in master.

Task: [4783504](https://www.odoo.com/odoo/2328/tasks/4783504)
opw: [4778681](https://www.odoo.com/odoo/2328/tasks/4778681)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/functions/module_logical.ts
+++ b/src/functions/module_logical.ts
@@ -70,7 +70,7 @@ export const IF = {
       return { value: "" };
     }
     if (result.value === null) {
-      result.value = "";
+      return { ...result, value: "" };
     }
     return result;
   },
@@ -98,7 +98,7 @@ export const IFERROR = {
       return { value: "" };
     }
     if (result.value === null) {
-      result.value = "";
+      return { ...result, value: "" };
     }
     return result;
   },
@@ -126,7 +126,7 @@ export const IFNA = {
       return { value: "" };
     }
     if (result.value === null) {
-      result.value = "";
+      return { ...result, value: "" };
     }
     return result;
   },
@@ -167,7 +167,7 @@ export const IFS = {
           return { value: "" };
         }
         if (result.value === null) {
-          result.value = "";
+          return { ...result, value: "" };
         }
         return result;
       }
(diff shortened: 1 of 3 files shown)


src/functions/module_logical.ts
```diff
@@ -70,7 +70,7 @@ export const IF = {
       return { value: "" };
     }
     if (result.value === null) {
-      result.value = "";
```
LucasLefevre, line 73: @laa-odoo 
I don't know why it was not simply `return { value: "" }`, included in the if above.
Is it important to preserve the format in this case?

rrahir: the format of the arguments is passed (in GSheet but not in Excel Desktop) while going through IF. 
I think it make sense to display the same representation that you point towards to

src/plugins/ui_core_views/cell_evaluation/evaluator.ts
```diff
@@ -36,7 +36,7 @@ import { SpreadingRelation } from "./spreading_relation";
 
 const MAX_ITERATION = 30;
 const ERROR_CYCLE_CELL = createEvaluatedCell(new CircularDependencyError());
-const EMPTY_CELL = createEvaluatedCell({ value: null });
+const EMPTY_CELL = Object.freeze(createEvaluatedCell({ value: null }));
```
hokolomopo, line 39: might as well freeze both

rrahir:
@robodoo r+

PR #6350, opened 2025-05-12
[FIX] OT: Fix chart transformations

rrahir:
Currently, we transform chart ranges upon the addition/deletion of
headers in a sheet regardless if the range belongs to the said sheet.

e.g.
- given a chart with a range  A1:A3 on sheet1 and a second A1:A3  on sheet2
- delete a the second row of sheet1

Both ranges are transformed and set to A1:A2. There is a double issue
with the range on sheet2.
1. it's range zone is updated
2. the sheet prefix is removed, meaning that it no longer points towards
   the same sheet!

Task: 4717724

Task: [4717724](https://www.odoo.com/odoo/2328/tasks/4717724)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/collaborative/ot/srt_specific.ts
+++ b/src/collaborative/ot/srt_specific.ts
@@ -1,11 +1,14 @@
 import { deepCopy } from "../../helpers";
+import { transformDefinition } from "../../helpers/figures/charts";
 import { adaptFormulaStringRanges, adaptStringRange } from "../../helpers/formulas";
 import { specificRangeTransformRegistry } from "../../registries/srt_registry";
 import {
   AddConditionalFormatCommand,
   AddDataValidationCommand,
   AddPivotCommand,
+  CreateChartCommand,
   UpdateCellCommand,
+  UpdateChartCommand,
   UpdatePivotCommand,
 } from "../../types/commands";
 import { RangeAdapter } from "../../types/misc";
@@ -101,3 +104,16 @@ function addPivotCommandAdaptRange<Cmd extends AddPivotCommand | UpdatePivotComm
 }
 specificRangeTransformRegistry.add("ADD_PIVOT", addPivotCommandAdaptRange);
 specificRangeTransformRegistry.add("UPDATE_PIVOT", addPivotCommandAdaptRange);
+
+specificRangeTransformRegistry.add("CREATE_CHART", updateChartRangesTransformation);
+specificRangeTransformRegistry.add("UPDATE_CHART", updateChartRangesTransformation);
+
+function updateChartRangesTransformation<Cmd extends UpdateChartCommand | CreateChartCommand>(
+  cmd: Cmd,
+  applyChange: RangeAdapter
+): Cmd {
+  return {
+    ...cmd,
+    definition: transformDefinition(cmd.sheetId, cmd.definition, applyChange),
+  };
+}
--- a/src/helpers/figures/charts/funnel_chart.ts
+++ b/src/helpers/figures/charts/funnel_chart.ts
@@ -1,14 +1,13 @@
 import { ChartConfiguration } from "chart.js";
 import { BACKGROUND_CHART_COLOR } from "../../../constants";
 import {
-  AddColumnsRowsCommand,
   ApplyRangeChange,
   Color,
   CommandResult,
   CoreGetters,
   Getters,
   Range,
-  RemoveColumnsRowsCommand,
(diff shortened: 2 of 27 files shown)


src/collaborative/ot/srt_specific.ts
```diff
@@ -101,3 +105,92 @@ function addPivotCommandAdaptRange<Cmd extends AddPivotCommand | UpdatePivotComm
...
+    case "line":
+    case "combo":
+    case "funnel":
+    case "pie":
+    case "waterfall":
+    case "pyramid":
+    case "radar":
+    case "sunburst":
+    case "treemap":
+    case "geo":
```
hokolomopo, line 127: I predict a 94.87% of chance that we will forget about this the next time we add a new chart type. Can't we do some `if "dataset" in definition` ?

src/helpers/figures/charts/funnel_chart.ts
```diff
@@ -84,13 +81,6 @@ export class FunnelChart extends AbstractChart {
     this.cumulative = definition.cumulative;
   }
 
-  static transformDefinition(
-    definition: FunnelChartDefinition,
-    executed: AddColumnsRowsCommand | RemoveColumnsRowsCommand
-  ): FunnelChartDefinition {
-    return transformChartDefinitionWithDataSetsWithZone(definition, executed);
```
hokolomopo, line 91: can also delete helper transformChartDefinitionWithDataSetsWithZone

src/collaborative/ot/srt_specific.ts
```diff
@@ -101,3 +105,92 @@ function addPivotCommandAdaptRange<Cmd extends AddPivotCommand | UpdatePivotComm
 }
 specificRangeTransformRegistry.add("ADD_PIVOT", addPivotCommandAdaptRange);
 specificRangeTransformRegistry.add("UPDATE_PIVOT", addPivotCommandAdaptRange);
+
+function adaptChartRange<Cmd extends CreateChartCommand | UpdateChartCommand>(
```
hokolomopo, line 109: A drawback to have the transform be defined here rather than in the charts is that it cannot be extended if we add a new type of chart outside of o-spreadhseet.

src/helpers/range.ts
```diff
@@ -457,7 +458,7 @@ function getApplyRangeChangeAddColRow(cmd: AddColumnsRowsCommand): ApplyRangeCha
 
 function getApplyRangeChangeDeleteSheet(cmd: DeleteSheetCommand): ApplyRangeChange {
   return (range: Range) => {
-    if (range.sheetId !== cmd.sheetId && range.invalidSheetName !== cmd.sheetName) {
+    if (range.sheetId !== cmd.sheetId && !isSheetNameEqual(range.invalidSheetName, cmd.sheetName)) {
```
hokolomopo, line 461: probably deserves a commit/a test ?

hokolomopo (commented):
vwoup vwoup :wave:

src/collaborative/ot/srt_specific.ts
rrahir: yup :o i'll just update `transformDefinition` and where it's called then :)

tests/collaborative/collaborative_sheet_manipulations.test.ts
```diff
@@ -640,7 +640,7 @@ describe("Collaborative Sheet manipulation", () => {
         (user) => user.getters.getChartDefinition(chartId),
         {
           ...chartDef,
-          dataSets: [{ dataRange: "A1:A3" }, { dataRange: "H1:H3" }],
+          dataSets: [{ dataRange: "A1:A3", yAxisId: "y" }, { dataRange: "H1:H3" }],
```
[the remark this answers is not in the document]
rrahir: the previous implementation seemingly was only recreating the dadaset with a datarange and nothing else:
```typescript
const dataSets: CustomizedDataSet[] = definition.dataSets
    .map((ds) => toUnboundedZone(ds.dataRange))
    .map((zone) => transformZone(zone, executed))
    .filter(isDefined)
    .map((xc) => ({ dataRange: zoneToXc(xc) })); // << there is the issue
    ```
whereas with this fix we preserve the dataset structure and just modify dataRange.

hokolomopo:
@robodoo r+ rebase-ff

PR #6378, opened 2025-05-12
[FIX] module logical: vectorize only if valid expression

laa-odoo:
In logical formulas, avoid vectorizing arguments not affected by the logical expression.

Task: [4775222](https://www.odoo.com/odoo/2328/tasks/4775222)

### DIFF
--- a/src/functions/helper_matrices.ts
+++ b/src/functions/helper_matrices.ts
@@ -139,12 +139,20 @@ export function multiplyMatrices(matrix1: Matrix<number>, matrix2: Matrix<number
 /**
  * Return the input if it's a scalar or the first element of the input if it's a matrix.
  */
-export function toScalar<T>(matrix: Matrix<T> | T): T {
-  if (!isMatrix(matrix)) {
-    return matrix;
+export function toScalar<T>(arg: Matrix<T> | T): T {
+  if (!isMatrix(arg)) {
+    return arg;
   }
-  if (matrix.length !== 1 || matrix[0].length !== 1) {
+  if (!isSingleElementMatrix(arg)) {
     throw new EvaluationError(_t("The value should be a scalar or a 1x1 matrix"));
   }
-  return matrix[0][0];
+  return arg[0][0];
+}
+
+function isSingleElementMatrix<T>(matrix: Matrix<T>) {
+  return matrix.length === 1 && matrix[0].length === 1;
+}
+
+export function isMultipleElementMatrix(arg: any) {
+  return isMatrix(arg) && !isSingleElementMatrix(arg);
 }
--- a/src/functions/index.ts
+++ b/src/functions/index.ts
@@ -11,9 +11,9 @@ import {
   Matrix,
   isMatrix,
 } from "../types";
-import { BadExpressionError, EvaluationError, NotAvailableError } from "../types/errors";
+import { BadExpressionError, EvaluationError } from "../types/errors";
 import { addMetaInfoFromArg, validateArguments } from "./arguments";
-import { generateMatrix, isEvaluationError, matrixForEach, matrixMap } from "./helpers";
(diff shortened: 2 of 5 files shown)


src/functions/index.ts
```diff
@@ -277,6 +207,86 @@ export function handleError(e: unknown, functionName: string): FunctionResultObj
   );
 }
 
+export function applyVectorization<T extends Arg[]>(
```
VincentSchippefilt, line 210: Document me, sempaï...
Who should call me, why and when ?

src/functions/helper_matrices.ts
```diff
@@ -139,12 +139,26 @@ export function multiplyMatrices(matrix1: Matrix<number>, matrix2: Matrix<number
...
+}
+
+function isOneOneMatrix<T>(matrix: Matrix<T>) {
+  if (matrix.length !== 1) {
+    return false;
+  }
+  if (matrix[0].length !== 1) {
+    return false;
+  }
+  return true;
```
hokolomopo, line 159:
```suggestion
    return matrix.length === 1 && matrix[0].length === 1;
```

hokolomopo: And rename to `isOneByOneMatrix` matrix ? `isSingleCellMatrix` ?

src/functions/module_logical.ts
```diff
@@ -49,27 +51,33 @@ export const IF = {
...
-      "logical_expression (boolean)",
+      "logical_expression (boolean, range<boolean>)",
       _t(
         "An expression or reference to a cell containing an expression that represents some logical value, i.e. TRUE or FALSE."
       )
     ),
-    arg("value_if_true (any)", _t("The value the function returns if logical_expression is TRUE.")),
     arg(
-      "value_if_false (any, default=FALSE)",
+      "value_if_true (any, range)",
```
hokolomopo, line 60:
```suggestion
      "value_if_true (any, range<any>)",
```
probably a bit more clear no ? Otherwise I'm wondering why `range` is not included in `any`

hokolomopo (commented):
I didn't follow the discussion at all, so it was just a fast scroll scroll and I'll let someone that know what you did reveiw it correctly :)

src/functions/helper_matrices.ts
laa-odoo: thx

src/functions/module_logical.ts
laa-odoo: both notations are possible, we should clean everything up if we want only one representation

rrahir: if both are possible (atm) i think it doesn't hurt to add <any> to avoid misunderstandings

src/functions/module_logical.ts
```diff
@@ -49,27 +56,33 @@ export const IF = {
...
-    valueIfTrue: Maybe<FunctionResultObject>,
-    valueIfFalse: Maybe<FunctionResultObject>
-  ): FunctionResultObject {
-    const result = toBoolean(logicalExpression?.value) ? valueIfTrue : valueIfFalse;
+  compute: function (logicalExpression: Arg, valueIfTrue: Arg, valueIfFalse: Arg) {
+    if (isMultipleElementMatrix(logicalExpression)) {
+      return applyVectorization(IF.compute, [logicalExpression, valueIfTrue, valueIfFalse]);
+    }
+    let result = toBoolean(toScalar(logicalExpression)) ? valueIfTrue : valueIfFalse;
+    // useful for interpreting empty cell references as empty strings. But must be removed to make empty cell references equal to zero
```
rrahir, line 78: Could you add a dedicated test to this case so we can state clearly "we have decided to return empty stringsin this specific case" ?

rrahir:
@robodoo r+

laa-odoo:
rebase-ff

laa-odoo:
robodoo rebase-ff

PR #6414, opened 2025-05-14
Master enhance pivot formula pro

pro-odoo:
## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/functions/module_lookup.ts
+++ b/src/functions/module_lookup.ts
@@ -856,13 +856,18 @@ export const PIVOT = {
       _t("Whether to include the column titles or not.")
     ),
     arg("column_count (number, optional)", _t("number of columns")),
+    arg(
+      "include_measure_titles (boolean, default=TRUE)",
+      _t("Whether to include the measure titles row or not.")
+    ),
   ],
   compute: function (
     pivotFormulaId: Maybe<FunctionResultObject>,
     rowCount: Maybe<FunctionResultObject> = { value: 10000 },
     includeTotal: Maybe<FunctionResultObject> = { value: true },
     includeColumnHeaders: Maybe<FunctionResultObject> = { value: true },
-    columnCount: Maybe<FunctionResultObject> = { value: Number.MAX_VALUE }
+    columnCount: Maybe<FunctionResultObject> = { value: Number.MAX_VALUE },
+    includeMeasureTitles: Maybe<FunctionResultObject> = { value: true }
   ) {
     const _pivotFormulaId = toString(pivotFormulaId);
     const _rowCount = toNumber(rowCount, this.locale);
@@ -876,6 +881,7 @@ export const PIVOT = {
     const visibilityOptions: PivotVisibilityOptions = {
       displayColumnHeaders: toBoolean(includeColumnHeaders),
       displayTotals: toBoolean(includeTotal),
+      displayMeasuresRow: toBoolean(includeMeasureTitles),
     };
 
     const pivotId = getPivotId(_pivotFormulaId, this.getters);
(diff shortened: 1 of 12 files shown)


tests/pivots/pivot_menu_items.test.ts
```diff
@@ -368,7 +368,7 @@ describe("Pivot fix formula menu item", () => {
   test("Fixing the formula take into account the arguments of PIVOT()", () => {
     // prettier-ignore
     const grid = {
-     A1: "Customer", B1: "Price", C1: "Date",     E1: "=PIVOT(1, 1, false, false, 1)",
+     A1: "Customer", B1: "Price", C1: "Date",     E1: "=PIVOT(1, 1, false, false, 1, false)",
```
hokolomopo, line 371: Why this change ? You change the result of the pivot function in the test (now the measure isn't shown), but the test didn't have to change ?

hokolomopo: Nvm I understood, before hiding the headers also hid the measures. The change might break some spreadsheets then ? Are we fine with that ?

pro-odoo: Yes it might break some spreadsheets, and from what we discussed yesterday with @LucasLefevre it was fine.
Another solution is to make this new attribute false by default and did not take it into account if `include_columns_title` is true to keep the current behavior, but meh...

As soon as it's not a traceback, I think I will be able to sleep tonight

src/functions/module_lookup.ts
```diff
@@ -848,13 +856,18 @@ export const PIVOT = {
       _t("Whether to include the column titles or not.")
     ),
     arg("column_count (number, optional)", _t("number of columns")),
+    arg(
+      "include_measures (boolean, default=TRUE)",
+      _t("Whether to include the measures row or not.")
```
LucasLefevre: Just "measure" could be understood as measure values ?
```suggestion
      "include_measure_titles (boolean, default=TRUE)",
      _t("Whether to include the measure titles row or not.")
```

src/helpers/pivot/table_spreadsheet_pivot.ts
```diff
@@ -157,12 +157,29 @@ export class SpreadsheetPivotTable {
     return this.columns.at(-1)?.length || 0;
   }
 
+  getSkippedRows(visibilityOptions: PivotVisibilityOptions) {
```
LucasLefevre: private

src/helpers/pivot/table_spreadsheet_pivot.ts
```diff
@@ -174,10 +191,13 @@ export class SpreadsheetPivotTable {
...
-      const startRow = displayColumnHeaders ? 0 : this.columns.length;
+      const skippedRows = this.getSkippedRows(visibilityOptions);
       for (let col = 0; col < pivotWidth; col++) {
         domainArray.push([]);
-        for (let row = startRow; row < pivotHeight; row++) {
+        for (let row = 0; row < pivotHeight; row++) {
+          if (skippedRows.has(row)) {
+            continue;
+          }
           if (!displayTotals && row === pivotHeight) {
```
LucasLefevre: this `if` should be moved to `getSkippedRows`

pro-odoo: Actually it should be removed, the condition will never be true (for loop stopped at `row = pivotHeight - 1`

LucasLefevre (commented):
robodoo rebase-ff r+

PR #6415, opened 2025-05-14
[REF] charts: smaller templates

LucasLefevre:
## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/components/side_panel/chart/building_blocks/general_design/general_design_editor.ts
```diff
@@ -22,6 +22,7 @@ interface Props {
   figureId: UID;
   definition: ChartDefinition;
   updateChart: (figureId: UID, definition: Partial<ChartDefinition>) => DispatchResult;
+  canUpdateChart: (figureId: UID, definition: Partial<ChartDefinition>) => DispatchResult;
```
hokolomopo, line 25: could also create an interface DefaultChartProps or something, and have everything using/extending it. It's now used at like 20 places

src/components/side_panel/chart/combo_chart/combo_chart_design_panel.xml
```diff
@@ -3,7 +3,16 @@
     <GeneralDesignEditor t-props="props">
       <t t-set-slot="general-extension">
         <ChartLegend t-props="props"/>
-        <ChartShowValues t-props="props"/>
+        <ChartShowValues t-props="props">
+          <t t-set-slot="values-extension">
+            <Checkbox
+              name="'showDataMarkers'"
```
hokolomopo, line 9: could add the option to radar charts also

hokolomopo: And if you add it to radar chart, that's 3 places with the same checkbox. Maybe create a dedicated component at this point ?

src/components/side_panel/chart/combo_chart/combo_chart_design_panel.xml
```diff
@@ -3,7 +3,16 @@
     <GeneralDesignEditor t-props="props">
       <t t-set-slot="general-extension">
         <ChartLegend t-props="props"/>
-        <ChartShowValues t-props="props"/>
+        <ChartShowValues t-props="props">
+          <t t-set-slot="values-extension">
```
hokolomopo, line 7: I get that we want to it next to "show value", but does it really make sense to have them in the same component ? As t-slot ? 

IMO it would have more sense to have two components `ChartShowValueCheckbox` and `ChartShowMarkers` checkbox, and in the templates put them both in the same `Section`

What do you think ?

src/components/side_panel/chart/line_chart/line_chart_design_panel.xml
```diff
@@ -0,0 +1,30 @@
+<templates>
+  <t t-name="o-spreadsheet-LineChartDesignPanel">
+    <GeneralDesignEditor t-props="props">
+      <t t-set-slot="general-extension">
+        <ChartLegend t-props="props"/>
+        <ChartShowValues t-props="props">
+          <t t-set-slot="values-extension">
```
hokolomopo, line 7: If we end up keeping the t-slot, use the default slot to make the tempaltes easier to read/write

hokolomopo (commented):
bip bop :robot:

hokolomopo (commented):
@robodoo r+

hokolomopo:
@robodoo rebase-ff

PR #6417, opened 2025-05-14
[ADD] formulas: introduce `TEXTSPLIT` function

Rachico:
The current `SPLIT` function (inspired by Google Sheets) allows splitting a text using delimiters, but it does not exist in Excel. This causes errors when importing sheets containing the `SPLIT` function into Excel. This commit introduces the `TEXTSPLIT`function to enrich the library and ensure better compatibility with Excel.

Task: [4766568](https://www.odoo.com/odoo/2328/tasks/4766568)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/functions/module_text.ts
+++ b/src/functions/module_text.ts
@@ -1,9 +1,9 @@
 import { escapeRegExp, formatValue, trimContent } from "../helpers";
 import { _t } from "../translation";
 import { AddFunctionDescription, Arg, FunctionResultObject, Maybe } from "../types";
-import { CellErrorType, EvaluationError } from "../types/errors";
+import { CellErrorType, EvaluationError, NotAvailableError } from "../types/errors";
 import { arg } from "./arguments";
-import { reduceAny, toBoolean, toNumber, toString, transposeMatrix } from "./helpers";
+import { reduceAny, toBoolean, toMatrix, toNumber, toString, transposeMatrix } from "./helpers";
 
 const DEFAULT_STARTING_AT = 1;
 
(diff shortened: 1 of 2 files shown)


src/functions/module_text.ts
```diff
@@ -436,6 +436,112 @@ export const SPLIT = {
   isExported: true,
 } satisfies AddFunctionDescription;
 
+// -----------------------------------------------------------------------------
+// TEXTSPLIT
+// -----------------------------------------------------------------------------
+
+const TEXTSPLIT_DEFAULT_IGNORE_EMPTY = false;
+const TEXTSPLIT_DEFAULT_MATCH_MODE = 0;
+const TEXTSPLIT_DEFAULT_PAD_WITH = "N/A";
```
hokolomopo, line 445:
```suggestion
const TEXTSPLIT_DEFAULT_PAD_WITH = "#N/A";
```

hokolomopo: And we should return actual #N/A errors, not cells with  #N/A string

src/functions/module_text.ts
```diff
@@ -436,6 +436,112 @@ export const SPLIT = {
...
+
+export const TEXTSPLIT = {
+  description: _t("Splits text into rows or columns using specified column and row delimiters."),
+  args: [
+    arg("text (string)", _t("The text to split.")),
+    arg(
+      "col_delimiter (string, range<string>, optional)",
```
hokolomopo, line 452: arg is marked as mandatory in excel

src/functions/module_text.ts
```diff
@@ -436,6 +436,112 @@ export const SPLIT = {
...
+// -----------------------------------------------------------------------------
+
+const TEXTSPLIT_DEFAULT_IGNORE_EMPTY = false;
+const TEXTSPLIT_DEFAULT_MATCH_MODE = 0;
+const TEXTSPLIT_DEFAULT_PAD_WITH = "N/A";
+
+export const TEXTSPLIT = {
+  description: _t("Splits text into rows or columns using specified column and row delimiters."),
+  args: [
+    arg("text (string)", _t("The text to split.")),
```
hokolomopo, line 450: excel return #VALUE if the text is empty. we return an empty string

src/functions/module_text.ts
```diff
@@ -436,6 +436,112 @@ export const SPLIT = {
...
+    colDelimiter: Arg,
+    rowDelimiter: Arg,
+    ignoreEmpty: Maybe<FunctionResultObject> = { value: TEXTSPLIT_DEFAULT_IGNORE_EMPTY },
+    matchMode: Maybe<FunctionResultObject> = { value: TEXTSPLIT_DEFAULT_MATCH_MODE },
+    padWith: Maybe<FunctionResultObject> = { value: TEXTSPLIT_DEFAULT_PAD_WITH }
+  ) {
+    const _text = toString(text);
+    const _colDelimiter = toMatrix(colDelimiter)
+      .flat()
+      .map((v) => escapeRegExp(toString(v)));
```
hokolomopo, line 483:
```suggestion
      .flatMap((v) => escapeRegExp(toString(v)));
```

hokolomopo: same for rows

src/functions/module_text.ts
```diff
@@ -436,6 +436,112 @@ export const SPLIT = {
...
+    const splitByEachColDelimiter = _colDelimiter.length > 1;
+    const _rowDelimiter = toMatrix(rowDelimiter)
+      .flat()
+      .map((v) => escapeRegExp(toString(v)));
+    const splitByEachRowDelimiter = _rowDelimiter.length > 1;
```
hokolomopo, line 488: nitpick: could inline the condition where we need it. I don't think the separate variable helps the comprehension

src/functions/module_text.ts
```diff
@@ -436,6 +436,112 @@ export const SPLIT = {
...
+        ? new RegExp(_colDelimiter.join("|"), regexpFlags)
+        : new RegExp(_colDelimiter[0], regexpFlags);
+      const columns = _text.split(colRegexp);
+      cells[0] = [];
+      for (const col of columns) {
+        if (col === "") {
+          if (_ignoreEmpty) {
+            continue;
+          }
+        }
```
hokolomopo, line 530:
```suggestion
        if (col === "" && _ignoreEmpty) {
            continue;
          
        }
```

src/functions/module_text.ts
```diff
@@ -436,6 +436,112 @@ export const SPLIT = {
...
+    }
+
+    let cells: string[][] = [];
+    const regexpFlags = _matchMode === 1 ? "gi" : "g";
+
+    if (_rowDelimiter.length > 0) {
+      const rowRegexp = splitByEachRowDelimiter
+        ? new RegExp(_rowDelimiter.join("|"), regexpFlags)
+        : new RegExp(_rowDelimiter[0], regexpFlags);
+      const rowParts = _text.split(rowRegexp);
```
hokolomopo, line 504: you're not applying _ignoreEmpty if there is a row delimiter but no col delimiter

src/functions/module_text.ts
```diff
@@ -436,6 +436,112 @@ export const SPLIT = {
...
+      if (_colDelimiter.length > 0) {
+        for (const rowText of rowParts) {
+          const colRegexp = splitByEachColDelimiter
+            ? new RegExp(_colDelimiter.join("|"), regexpFlags)
+            : new RegExp(_colDelimiter[0], regexpFlags);
```
hokolomopo, line 509: no need to create a new regex at each loop iteration is there ?

src/functions/module_text.ts
hokolomopo: or use `.filter` ?

src/functions/module_text.ts
hokolomopo: and the _ignoreEmpty is wrong if there is both col & row delimiter, I can end up with a row full of #N/A which should not happens

src/functions/module_text.ts
```diff
@@ -436,6 +436,112 @@ export const SPLIT = {
...
+        cells[0].push(col);
+      }
+    }
+    const maxLength = Math.max(...cells.map((row) => row.length));
+    for (const row of cells) {
+      while (row.length < maxLength) {
+        row.push(toString(_padWith));
+      }
+    }
+    return transposeMatrix(cells);
```
hokolomopo, line 540: nitpick: performance-wise, is there a way to apply the split in a different order so we don't have to transpose at the end? If it's too annoying/make the code more complex, it's fine to keep the transpose

tests/functions/module_text.test.ts
```diff
@@ -571,6 +571,94 @@ describe("SPLIT function", () => {
...
+    const model = createModelFromGrid(grid);
+    setCellContent(model, "F1", ", ");
+    setCellContent(model, "F2", ";");
+    setCellContent(model, "A2", "=TEXTSPLIT(A1, F1:F2)");
+    expect(getRangeValuesAsMatrix(model, "A2:E2")).toEqual([
+      ["Apple", "Banana", "Orange", "Grapes", "Tomato"],
+    ]);
+  });
+
+  test("Ignore empty values", () => {
```
hokolomopo, line 629: missing test for "Ignore empty values" with row delimiter, and row & col delimiter since they do not work :)

tests/functions/module_text.test.ts
```diff
@@ -571,6 +571,94 @@ describe("SPLIT function", () => {
...
+    const grid = { A1: "Dog,,Cat,,Bird" };
+    const model = createModelFromGrid(grid);
+    setCellContent(model, "A2", '=TEXTSPLIT(A1, ",", , TRUE)');
+    expect(getRangeValuesAsMatrix(model, "A2:C2")).toEqual([["Dog", "Cat", "Bird"]]);
+  });
+
+  test("Case-insensitive split", () => {
+    const grid = { A1: "AppleDELIMbanaNADelimcHerry" };
+    const model = createModelFromGrid(grid);
+    setCellContent(model, "A2", '=TEXTSPLIT(A1, "delim",,,1)');
```
hokolomopo, line 639: also test what happens if the argument is set to 0

tests/functions/module_text.test.ts
```diff
@@ -571,6 +571,94 @@ describe("SPLIT function", () => {
...
+      ["Name", "Age"],
+      ["City", "Country"],
+    ]);
+  });
+
+  test("Multiple delimiters", () => {
+    const grid = { A1: "Apple, Banana;Orange, Grapes;Tomato" };
+    const model = createModelFromGrid(grid);
+    setCellContent(model, "F1", ", ");
+    setCellContent(model, "F2", ";");
```
hokolomopo, line 622: nitpick: you could set those in the grid just above

hokolomopo (commented):
👋 

ignore_empty parameters seems a bit bugged unfortunately :x

src/functions/module_text.ts
Rachico: Yes that's what I did in the beginning but I think it's more readable this way :)

src/functions/module_text.ts
```diff
@@ -436,6 +436,120 @@ export const SPLIT = {
...
+    if (_rowDelimiter.length > 0) {
+      const rowRegexp = splitByEachRowDelimiter
+        ? new RegExp(_rowDelimiter.join("|"), regexpFlags)
+        : new RegExp(_rowDelimiter[0], regexpFlags);
+      let rowParts = _text.split(rowRegexp);
+      if (_ignoreEmpty) {
+        rowParts = rowParts.filter((v) => v !== "");
+      }
+      if (_colDelimiter.length > 0) {
+        const colRegexp = splitByEachColDelimiter
```
laa-odoo, line 514: does `const colRegexp =  new RegExp(_colDelimiter.join("|"), regexpFlags)` naturally handles the default case where `_colDelimiter.length` is equal to 1 ? same with `splitByEachRowDelimiter` ?

in this case, we can remove `splitByEachRowDelimiter` and `splitByEachColDelimiter`

rrahir (commented):
just some fine tuning for the error messages, otherwise good 👍

src/functions/module_text.ts
```diff
@@ -526,6 +526,105 @@ export const TEXTJOIN = {
...
+    }
+
+    const maxLength = Math.max(...cells.map((row) => row.length));
+    for (const row of cells) {
+      while (row.length < maxLength) {
+        row.push(padWith);
+      }
+    }
+
+    return transposeMatrix(cells);
```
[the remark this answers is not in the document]
laa-odoo: I'm not a fan either, but it's already used a large number of times in all the formulas. A performance improvement task is planned where we need to reduce, if possible, calls to `transposeMatrix` and, more generally, to `generateMatrix`, which is behind it.

Here, we've already significantly reduced the size of the compute to make it easier to understand (because there are many subtleties with this formula). But we can try to work on a version without the transpose.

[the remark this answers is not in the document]
rrahir: I don't think it's necessary, it was more a generic comment for the future, but if we are aware of  this, it's fine :)

PR #6420, opened 2025-05-14
[IMP] xlsx: prevent exporting the SPLIT function

Rachico:
The `SPLIT` function does not exist in Excel yet. This function
causes an error when exporting and importing an .xlsx file in excel.
The `TEXTSPLIT` function will be therefore introduced here: https://github.com/odoo/o-spreadsheet/pull/6417

Task: [4766568](https://www.odoo.com/odoo/2328/tasks/4766568)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/functions/module_text.ts
+++ b/src/functions/module_text.ts
@@ -430,7 +430,7 @@ export const SPLIT = {
 
     return transposeMatrix([result]);
   },
-  isExported: true,
+  isExported: false,
 } satisfies AddFunctionDescription;
 
 // -----------------------------------------------------------------------------


src/functions/module_text.ts
```diff
@@ -430,7 +430,7 @@ export const SPLIT = {
 
     return transposeMatrix([result]);
   },
-  isExported: true,
+  isExported: false,
```
hokolomopo, line 433: In commit message:

> causes an error at import

It causes an error at export not import tight ?

rrahir: it fails when importing in Excel ;) but the wording of the message could be improved, I agree.

Rachico: Should be good now :)

PR #6448, opened 2025-05-21
[IMP] renderer: add animations on cell change

hokolomopo:
## Description:

This commit adds animations in the grid's canvas when a cell is changed.
The animations are defined in the `cell_animation_registry.ts` and
includes:

- fade in the cell when it gets a new content.
- fate out the cell when its content is removed.
- text sliding animation when the text is changed.
- color gradient animation when the cell colors is changed
- and more

The feature is designed to not impact the current grid drawing logic.
We still have a first step yo create `Box` from the model content,
and a second step to draw the `Box` in the canvas. The animations
are a step in between, where we change the `Box` depending on the
animation progress, and possibly add some additional `Box` to be drawn.

Task: [4805149](https://www.odoo.com/odoo/2328/tasks/4805149)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/registries/cell_animation_registry.ts
```diff
@@ -0,0 +1,371 @@
...
+
+interface CellAnimationItem {
+  id: string;
+  easingFn: keyof typeof EASING_FN;
+  /**
+   * Checks if the changes between the oldBox and newBox require an animation.
+   */
+  hasAnimation: (oldBox?: RenderingBox, newBox?: RenderingBox) => boolean;
+  /**
+   * Updates in place the given animatedBox with the animation progress.
```
LucasLefevre, line 14: could you add the range of values for `progress`. Is it a number between 0 and 1 ?

src/components/grid_overlay/grid_overlay.ts
```diff
@@ -1,5 +1,5 @@
 import { Component, onMounted, onWillUnmount, useExternalListener, useRef } from "@odoo/owl";
-import { deepEquals } from "../../helpers";
+import { deepEquals, positionToZone } from "../../helpers";
```
rrahir, line 2: typo in the commit msg:
"We still have a first step ~~yo~~ to create `Box`"

src/types/commands.ts
```diff
@@ -651,6 +652,11 @@ export interface UpdateLocaleCommand {
   locale: Locale;
 }
 
+export interface ToggleCellAnimationsCommand {
+  type: "TOGGLE_CELL_ANIMATIONS";
```
rrahir: if the command is called "toggle", it should just jump from state active to inactive, and therefore does not need any payload. 
If we want a payload, I 'd suggest changing the type to `SET_CELL_ANIMATIONS`, and with a payload `enableCellAnimations`

src/types/workbook_data.ts
```diff
@@ -64,6 +64,7 @@ export interface SheetData {
 
 interface WorkbookSettings {
   locale: Locale;
+  disableCellAnimations?: boolean;
```
rrahir, line 67: I think it's worth having a proper migration, it won't cost much

rrahir (commented):
Hell of a feature, I just have some small nitpicks ^^

src/types/workbook_data.ts
hokolomopo: I think it's 40 lines (w/ test) of useless code TBH

rrahir:
robodoo rebase-ff r+

PR #6457, opened 2025-05-22
[FIX] waterfall: wrong show value for waterfall chart

hokolomopo:
## Description

The shot value chartJs plugin would show the same value for a waterfall chart than abar chart (the value of the top of the bar), instead of the correct value for a waterfall chart (the difference between the top and the bottom of the bar).

Task: [4812692](https://www.odoo.com/odoo/2328/tasks/4812692)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/components/figures/chart/chartJs/chartjs_show_values_plugin.ts
```diff
@@ -103,7 +104,13 @@ function drawLineOrBarChartValues(
...
       ctx.fillStyle = point.options.backgroundColor;
       ctx.strokeStyle = options.background || "#ffffff";
-      drawTextWithBackground(options.callback(value - 0), xPosition, yPosition, ctx);
+
+      let valueToDisplay = value;
+      if (options.isWaterfall) {
+        const raw = dataset._dataset.data[i];
+        valueToDisplay = raw[1] - raw[0];
+      }
+      drawTextWithBackground(options.callback(Number(valueToDisplay)), xPosition, yPosition, ctx);
```
rrahir, line 113: magic language ><

src/helpers/figures/charts/waterfall_chart.ts
```diff
@@ -330,6 +330,7 @@ function getWaterfallConfiguration(
   config.options.plugins!.chartShowValuesPlugin = {
     showValues: chart.showValues,
     background: chart.background,
+    isWaterfall: true,
```
LucasLefevre: nitpick. At least it's not coupled strictly speaking to waterfall ? We could imagine also showing the delta on line charts
```suggestion
    showDelta: true,
```

src/components/figures/chart/chartJs/chartjs_show_values_plugin.ts
```diff
@@ -103,7 +104,13 @@ function drawLineOrBarChartValues(
 
       ctx.fillStyle = point.options.backgroundColor;
       ctx.strokeStyle = options.background || "#ffffff";
-      drawTextWithBackground(options.callback(value - 0), xPosition, yPosition, ctx);
```
LucasLefevre, line 106: `- 0` ? :thinking:

rrahir: it casts as a number :D

src/helpers/figures/charts/waterfall_chart.ts
hokolomopo: I ended up changing approche, I use the `callback` given to the plugin rather than a new option. I think it make a bit more sense, and it's possible to test.

src/components/figures/chart/chartJs/chartjs_show_values_plugin.ts
hokolomopo: The person, whose name I won't cite, that first wrote this plugin decided that it was the best way to convert a string to a number in javascript 🙃

LucasLefevre: very very strange. Because this person could have chosen many other ways

```
const str = "123.45";

console.log(Number(str));
console.log(+str);
console.log(parseInt(str, 10));
console.log(parseFloat(str));
console.log(str * 1);
console.log(str / 1);
console.log(str - 0);
console.log(str | 0);       // truncates to 123
console.log(str >> 0);      // truncates to 123
console.log(str >>> 0);     // truncates to 123
console.log(~~str);         // truncates to 123
console.log(eval(str));
console.log(JSON.parse(str));
console.log(new Number(str).valueOf());
console.log(Object(str) - 0);
console.log(Function('"use strict"; return "' + str + '"')() * 1);
console.log(["123.45"][0] * 1);
console.log(`${+str}` - 0);
```
(thanks chatGPT)

src/helpers/figures/charts/waterfall_chart.ts
hokolomopo: Also subtotals should probably be not displayed as deltas

src/helpers/figures/charts/pyramid_chart.ts
```diff
@@ -217,8 +217,8 @@ export function createPyramidChartRuntime(
     return tooltipLabelCallback(tooltipItem);
   };
   const callback = config.options!.plugins!.chartShowValuesPlugin!.callback;
-  config.options!.plugins!.chartShowValuesPlugin!.callback = (x) =>
-    callback!(Math.abs(x as number));
+  config.options!.plugins!.chartShowValuesPlugin!.callback = (x, dataset, index) =>
```
rrahir:
```suggestion
  config.options!.plugins!.chartShowValuesPlugin!.callback = (value, dataset, index) =>
```

rrahir:
robodoo r+

PR #6480, opened 2025-05-26
Master chips lul

LucasLefevre:
## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/composer/grid_composer/grid_composer.ts
+++ b/src/components/composer/grid_composer/grid_composer.ts
@@ -9,7 +9,13 @@ import {
   toXC,
 } from "../../../helpers";
 import { Store, useStore } from "../../../store_engine";
-import { ComposerFocusType, DOMDimension, Rect, SpreadsheetChildEnv } from "../../../types/index";
+import {
+  CellPosition,
+  ComposerFocusType,
+  DOMDimension,
+  Rect,
+  SpreadsheetChildEnv,
+} from "../../../types/index";
 import { getTextDecoration } from "../../helpers";
 import { css, cssPropertiesToCss } from "../../helpers/css";
 import { CellComposerStore } from "../composer/cell_composer_store";
@@ -62,6 +68,11 @@ export class GridComposer extends Component<Props, SpreadsheetChildEnv> {
   private rect: Rect = this.defaultRect;
   private isEditing: boolean = false;
   private isCellReferenceVisible: boolean = false;
+  private currentEditedCell: CellPosition = {
+    col: 0,
+    row: 0,
+    sheetId: this.env.model.getters.getActiveSheetId(),
+  };
 
   private composerStore!: Store<CellComposerStore>;
   composerFocusStore!: Store<ComposerFocusStore>;
@@ -212,12 +223,22 @@ export class GridComposer extends Component<Props, SpreadsheetChildEnv> {
       this.composerFocusStore.focusComposer(this.composerInterface, { focusMode: "inactive" });
     }
 
+    let shouldRecomputeRect = !deepEquals(
+      this.currentEditedCell,
+      this.composerStore.currentEditedCell
+    );
+
     if (this.isEditing !== isEditing) {
       this.isEditing = isEditing;
       if (!isEditing) {
         this.rect = this.defaultRect;
         return;
       }
+      this.currentEditedCell = this.composerStore.currentEditedCell;
+      shouldRecomputeRect = true;
+    }
+
+    if (shouldRecomputeRect) {
       const position = this.env.model.getters.getActivePosition();
       const zone = this.env.model.getters.expandZone(position.sheetId, positionToZone(position));
       this.rect = this.env.model.getters.getVisibleRect(zone);
--- a/tests/composer/composer_integration_component.test.ts
+++ b/tests/composer/composer_integration_component.test.ts
@@ -1,4 +1,4 @@
-import { Model } from "../../src";
+import { HeaderIndex, Model } from "../../src";
 import { CellComposerStore } from "../../src/components/composer/composer/cell_composer_store";
 import {
   DEFAULT_CELL_HEIGHT,
@@ -669,6 +669,22 @@ describe("Grid composer", () => {
 
       expect(getElComputedStyle(composerContainerSelector, "top")).toBe(expectedTop + "px");
       expect(getElComputedStyle(composerContainerSelector, "left")).toBe(expectedLeft + "px");
+    });
+
+    test("Grid Composer Position is recomputed if we change the edited cell", async () => {
+      const expectedTop = (index: HeaderIndex) => HEADER_HEIGHT + index * DEFAULT_CELL_HEIGHT;
+      const expectedLeft = (index: HeaderIndex) => HEADER_WIDTH + index * DEFAULT_CELL_WIDTH - 1; //-1 to include cell border
+      env.model.selection.selectCell(2, 4);
+      await typeInComposerGrid("coucou", true);
+      expect(getElComputedStyle(composerContainerSelector, "top")).toBe(expectedTop(4) + "px");
+      expect(getElComputedStyle(composerContainerSelector, "left")).toBe(expectedLeft(2) + "px");
+
+      env.model.selection.getBackToDefault();
+      env.model.selection.selectCell(1, 1);
+      env.startCellEdition();
+      await nextTick();
+      expect(getElComputedStyle(composerContainerSelector, "top")).toBe(expectedTop(1) + "px");
+      expect(getElComputedStyle(composerContainerSelector, "left")).toBe(expectedLeft(1) + "px");
     });
 
     test("Grid composer container have a min-height / min-width to have the same size as the edited cell ", async () => {
--- a/tests/grid/__snapshots__/grid_component.test.ts.snap
+++ b/tests/grid/__snapshots__/grid_component.test.ts.snap
@@ -90,7 +90,7 @@ exports[`Grid component simple rendering snapshot 1`] = `
           contenteditable="true"
           inputmode="none"
           spellcheck="false"
-          style="max-height: 1008.6px;"
+          style="max-height: 982.6px;"
           tabindex="1"
         />
       </div>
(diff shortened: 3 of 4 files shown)


demo/data.js
```diff
@@ -2951,7 +2951,7 @@ export const demoData = {
         C6: "-445248",
```
hokolomopo, line 2951: Some functional notes: 

- hover style of simple arrow (without chip) is worst than before IMO ? particularly on white background
- hover style of arrow/chip is set when hovering on the cell, not on the icon. And the rest of the cell isn't clickable
- whole chip should probably be clickable, not just the arrow ?

hokolomopo: - cannot use chip in "Values in range" ?

hokolomopo: - font size looks really small in the autocomplete dropdown

src/components/data_validation_overlay/dv_list_icon/dv_list_icon.scss
```diff
@@ -0,0 +1,5 @@
+.o-spreadsheet {
```
hokolomopo, line 1: css won't work with my PR to put all the icons in the canvas. Actually we should discuss how both task will interact, because I don't think it'll be easy

src/components/data_validation_overlay/dv_list_icon/dv_list_icon.ts
```diff
@@ -33,10 +35,27 @@ export class DataValidationListIcon extends Component<Props, SpreadsheetChildEnv
...
   }
+
+  get chipStyle() {
+    const style = this.env.model.getters.getDataValidationCellStyle(this.props.cellPosition);
+    const isHovered =
+      this.hoveredCellStore.col === this.props.cellPosition.col &&
+      this.hoveredCellStore.row === this.props.cellPosition.row;
+    const shadowColor = darkenColor(style?.textColor || TEXT_BODY_MUTED, 0.2);
+    return cssPropertiesToCss({
+      color: style?.textColor,
```
hokolomopo, line 57: should probably darken the color rather than adding a shadow on hover

src/plugins/ui_core_views/evaluation_data_validation.ts
```diff
@@ -124,6 +148,37 @@ export class EvaluationDataValidationPlugin extends CoreViewPlugin {
...
 
+  private hasChip(position: CellPosition) {
+    const rule = this.getters.getValidationRuleForCell(position);
+    if (!rule) {
+      return false;
+    }
+    return rule.criterion.type === "isValueInList" && rule.criterion.displayStyle === "chip";
+  }
+
+  private getDataValidationStyle(
```
hokolomopo, line 159: something is wrong in style computation, number cells are not styled

src/stores/grid_renderer_store.ts
```diff
@@ -607,7 +598,10 @@ export class GridRenderer {
     const showFormula = this.getters.shouldShowFormulas();
     const { x, y, width, height } = this.getters.getRect(zone);
     const { verticalAlign } = this.getters.getCellStyle(position);
-    let style = this.getters.getCellComputedStyle(position);
+    let style = {
+      ...this.getters.getCellComputedStyle(position),
+      ...this.getters.getDataValidationCellStyle(position),
+    };
```
hokolomopo, line 604: I don't like this. Every call to `getCellComputedStyle` is now wrong, except when we add `getDataValidationCellStyle` manually. Data validation style should be included in cell compute style.

demo/data.js
hokolomopo: - cannot change text color of chip item with dark background. Chip text color should not take priority over user-defined color. Not cf colors.

hokolomopo: - when chip arrow is white on dark background, hover style is barely noticeable

hokolomopo (commented):
Bip bop. Fast read :robot:

src/constants.ts
```diff
@@ -60,6 +60,7 @@ export const ALERT_INFO_BG = "#CDEDF1";
 export const ALERT_INFO_BORDER = "#98DBE2";
 export const ALERT_INFO_TEXT_COLOR = "#09414A";
 export const BADGE_SELECTED_COLOR = "#E6F2F3";
+export const CHIP_DEFAULT_COLOR = "#E8EAED";
```
hokolomopo: amost the same as GRAY_200 if we want to limit our color palette

src/constants.ts
```diff
@@ -179,6 +180,7 @@ export const GRID_ICON_EDGE_LENGTH = 17;
 export const FILTER_ICON_MARGIN = 2;
 export const FILTER_ICON_EDGE_LENGTH = 17;
 export const FOOTER_HEIGHT = 2 * DEFAULT_CELL_HEIGHT;
+export const CHIP_MARGIN = 5;
```
hokolomopo:
```suggestion
export const DATA_VALIDATION_CHIP_MARGIN = 5;
```
maybe ? we probably need to be specific if we put it along other constants

src/plugins/ui_core_views/evaluation_data_validation.ts
```diff
@@ -97,6 +117,15 @@ export class EvaluationDataValidationPlugin extends CoreViewPlugin {
     return evaluator.isCriterionValueValid(value) ? undefined : evaluator.criterionValueErrorString;
   }
 
+  getDataValidationRangeValues(sheetId: UID, criterion: EvaluatedCriterion) {
+    const range = this.getters.getRangeFromSheetXC(sheetId, String(criterion.values[0]));
+    const criterionValues = this.getters.getRangeValues(range);
+    return criterionValues
+      .filter(isNotNull)
+      .map((value) => value.toString())
+      .filter((val) => val !== "");
```
hokolomopo: do we need to filter twice ? Can't we filter everything the first time ?

src/plugins/ui_core_views/evaluation_data_validation.ts
```diff
@@ -124,6 +153,38 @@ export class EvaluationDataValidationPlugin extends CoreViewPlugin {
...
+  ): Pick<Style, "fillColor" | "textColor"> | undefined {
+    const rule = this.getters.getValidationRuleForCell(position);
+    if (rule?.criterion.type !== "isValueInList" && rule?.criterion.type !== "isValueInRange") {
+      return undefined;
+    }
+    if (this.isDataValidationInvalid(position)) {
+      return undefined;
+    }
+    const evaluatedCell = this.getters.getEvaluatedCell(position);
+    const color = Object.entries(rule.criterion.colors ?? {}).find(
```
hokolomopo: nitpick: with Object.entries we'll create bazillions of arrays. And the colors being a Record<value, color> seems a bit dumb if we cannot use the keys to lookup in the record because we need to lowercase them. I'd either lowercase the keys before entering them in the records, or replace the record by a `{color, value}[]`

src/registries/auto_completes/data_validation_auto_complete.ts
```diff
@@ -18,23 +19,31 @@ autoCompleteProviders.add("dataValidation", {
...
+    return values.map((value) => {
+      const color = colors?.[value];
+      return {
+        text: value,
+        htmlContent: [
+          {
+            value,
+            color: color ? chipTextColor(color) : undefined,
+            backgroundColor: color || CHIP_DEFAULT_COLOR,
+            classes: ["badge rounded-pill fs-6 fw-normal"],
```
rrahir:
```suggestion
            classes: ["badge rounded-pill fs-6 fw-normal w-100 mt-1 text-start"],
```
i woulld look like this
image

src/constants.ts
rrahir: bump ,think it's not necessary tohave 50 shades of grey

demo/data.js
rrahir: bump?

src/plugins/ui_core_views/evaluation_data_validation.ts
rrahir: bump

src/registries/icons_on_cell_registry.ts
```diff
@@ -70,15 +72,39 @@ iconsOnCellRegistry.add("data_validation_checkbox", (getters, position) => {
...
+      svg: getChipSvg(chipStyle),
+      hoverSvg: getHoveredChipSvg(chipStyle),
+      priority: 10,
+      horizontalAlign: "right",
+      size: computeTextFontSizeInPixels(cellStyle),
+      margin: 4,
+      position,
+      onClick: (position, env) => {
+        const { col, row } = position;
+        env.model.selection.selectCell(col, row);
```
rrahir, line 89: If I was editing a selectionInput (which does not do magic on blur but maybe it should) when I click ion the icon, it will first write the DV reference in the selectionInput and then open a composer where the previous selection was :'(. 

quick fix, force the selection back to the grid selection and we can thiink about how to tackle generally afterwards

src/plugins/ui_core_views/evaluation_data_validation.ts
```diff
@@ -70,6 +76,20 @@ export class EvaluationDataValidationPlugin extends CoreViewPlugin {
     return !this.getValidationResultForCell(cellPosition).isValid;
   }
 
+  getDataValidationCellStyle(position: CellPosition): Style | undefined {
+    if (this.hasChip(position)) {
+      return undefined; // chips are not applied to the cell style, but to the chip itself
```
rrahir: I understand because I read the code but tthe sentence is not clear, the only thing I get from it is "this is an exception"  in the rendering process and I'm not 100% certain it's a good idea. wouldn't it be possible to consider adding the "chip" to the cell evaluated style?

LucasLefevre: > wouldn't it be possible to consider adding the "chip" to the cell evaluated style?

I tried it originally but it was a(nother) mess.
It can only be added partially (only the text color, the background color cannot because it would overwrite the cell background color), which means other special case in the renderer and in the composer (the composer takes the cell style).

src/plugins/ui_core_views/evaluation_data_validation.ts
LucasLefevre:
```ts
    return criterionValues
      .filter((value) => isNotNull(value) && value !== "")
      .map((value) => value.toString());
```
`'value' is possibly 'null'.`

rrahir:
robodoo r+

LucasLefevre:
robodoo rebase-ff

rrahir:
classic rémi

rrahir:
robodoo r+

rrahir:
robodoo +

rrahir:
robodoo rebase-ff

rrahir:
robodoo r+

PR #6490, opened 2025-05-26
[IMP/FIX] autocomplete_dropdown: Improve dropdown readability

dhrp-odoo:
## Description:

Previously, the auto-complete dropdown in the composer was difficult to read when the cell background was dark and the text color was light. This commit fixes the issue by setting the auto-complete dropdown text color to black by default.

Task: [4813433](https://www.odoo.com/odoo/2328/tasks/4813433)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/composer/autocomplete_dropdown/autocomplete_dropdown.xml
+++ b/src/components/composer/autocomplete_dropdown/autocomplete_dropdown.xml
@@ -17,7 +17,7 @@
               t-as="content"
               t-key="content_index"
               t-att-class="content.class"
-              t-attf-style="color: {{content.color || 'inherit'}};"
+              t-attf-style="color: {{content.color || '#000000'}};"
               t-esc="content.value"
             />
           </div>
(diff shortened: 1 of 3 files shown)


tests/composer/autocomplete_dropdown_component.test.ts
```diff
@@ -113,6 +113,15 @@ describe("Functions autocomplete", () => {
       expect(fixture.querySelectorAll(".o-autocomplete-value")[1].textContent).toBe("SZZ");
     });
 
+    test("Autocomplete dropdown text should be black by default", async () => {
```
hokolomopo, line 116: doesn't really test the original bug:

- the problem was only on data validation drop down, not on function proposal
- we don't check if the cell style affects the proposals anymore

hokolomopo (commented):
small issue in the test :)

PR #6531, opened 2025-06-03
[IMP] evaluatedCell: add origin

fdamhaut:
## Description:

Task: [3810138](https://www.odoo.com/odoo/2328/tasks/3810138)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/cells/cell_evaluation.ts
+++ b/src/helpers/cells/cell_evaluation.ts
@@ -2,6 +2,7 @@ import { isEvaluationError, toString } from "../../functions/helpers";
 import {
   BooleanCell,
   Cell,
+  CellPosition,
   CellValue,
   CellValueType,
   DEFAULT_LOCALE,
@@ -28,13 +29,14 @@ import { isNumber, parseNumber } from "../numbers";
 
 export function evaluateLiteral(
   literalCell: LiteralCell,
-  localeFormat: LocaleFormat
+  localeFormat: LocaleFormat,
+  position?: CellPosition
 ): EvaluatedCell {
   const value =
     isTextFormat(localeFormat.format) && literalCell.parsedValue !== null
       ? literalCell.content
       : literalCell.parsedValue;
-  const functionResult = { value, format: localeFormat.format };
+  const functionResult = { value, format: localeFormat.format, origin: position };
   return createEvaluatedCell(functionResult, localeFormat.locale);
 }
 
@@ -61,11 +63,13 @@ export function parseLiteral(content: string, locale: Locale): CellValue {
 export function createEvaluatedCell(
   functionResult: FunctionResultObject,
   locale: Locale = DEFAULT_LOCALE,
-  cell?: Cell
+  cell?: Cell,
+  origin?: CellPosition
 ): EvaluatedCell {
   const link = detectLink(functionResult.value);
   if (!link) {
-    return _createEvaluatedCell(functionResult, locale, cell);
+    const evaluateCell = _createEvaluatedCell(functionResult, locale, cell);
+    return addOrigin(evaluateCell, functionResult.origin ?? origin);
   }
   const value = parseLiteral(link.label, locale);
   const format =
@@ -77,10 +81,13 @@ export function createEvaluatedCell(
     value,
     format,
   };
-  return {
-    ..._createEvaluatedCell(linkPayload, locale, cell),
-    link,
-  };
+  return addOrigin(
+    {
+      ..._createEvaluatedCell(linkPayload, locale, cell),
+      link,
+    },
+    functionResult.origin ?? origin
+  );
 }
 
 function _createEvaluatedCell(
@@ -194,4 +201,16 @@ function errorCell(value: string, message?: string): ErrorCell {
     isAutoSummable: false,
     defaultAlign: "center",
   };
+}
+
+function addOrigin(cell: EvaluatedCell, origin: CellPosition | undefined): EvaluatedCell {
+  if (cell.value === null) {
+    // ignore empty cells to allow sharing the same object instance
+    return cell;
+  }
+  if ("origin" in cell) {
+    return cell;
+  }
+  cell.origin = origin;
+  return cell;
 }
--- a/src/plugins/ui_core_views/cell_evaluation/evaluator.ts
+++ b/src/plugins/ui_core_views/cell_evaluation/evaluator.ts
@@ -42,7 +42,9 @@ import { RTreeBoundingBox } from "./r_tree";
 import { SpreadingRelation } from "./spreading_relation";
 
 const MAX_ITERATION = 30;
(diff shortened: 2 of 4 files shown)


src/helpers/cells/cell_evaluation.ts
```diff
@@ -62,7 +64,9 @@ export function createEvaluatedCell(
 ): EvaluatedCell {
   const link = detectLink(functionResult.value);
   if (!link) {
-    return _createEvaluatedCell(functionResult, locale, cell);
+    const evaluateCell = _createEvaluatedCell(functionResult, locale, cell);
+    if (!evaluateCell.origin) evaluateCell.origin = functionResult.origin;
```
LucasLefevre: I think this `if` is useless. What comes out of `_createEvaluatedCell` never has an origin
```suggestion
    evaluateCell.origin = functionResult.origin;
```

src/helpers/cells/cell_evaluation.ts
```diff
@@ -148,6 +153,7 @@ const emptyCell = memoize(function emptyCell(format: string | undefined): EmptyC
     type: CellValueType.empty,
     isAutoSummable: true,
     defaultAlign: "left",
+    origin: { sheetId: "", col: 0, row: 0 },
```
LucasLefevre: remove this since `origin` is optional.
It's better to have a missing origin rather than a wrong value

src/plugins/ui_core_views/cell_evaluation/evaluator.ts
```diff
@@ -42,8 +42,12 @@ import { RTreeBoundingBox } from "./r_tree";
 import { SpreadingRelation } from "./spreading_relation";
 
 const MAX_ITERATION = 30;
-const ERROR_CYCLE_CELL = Object.freeze(createEvaluatedCell(new CircularDependencyError()));
-const EMPTY_CELL = Object.freeze(createEvaluatedCell({ value: null }));
+const ERROR_CYCLE_CELL = Object.freeze(
+  createEvaluatedCell({ ...new CircularDependencyError(), origin: { sheetId: "", col: 0, row: 0 } })
+);
+const EMPTY_CELL = Object.freeze(
+  createEvaluatedCell({ value: null, origin: { sheetId: "", col: 0, row: 0 } })
```
LucasLefevre: same here, you can remove origin

src/plugins/ui_core_views/cell_evaluation/evaluator.ts
```diff
@@ -327,7 +331,9 @@ export class Evaluator {
 
     const cell = this.getters.getCell(position);
     if (cell === undefined) {
+      // Todo discuss about correctness vs perf
       return EMPTY_CELL;
+      // return { ...EMPTY_CELL, origin: position };
```
LucasLefevre: Let's keep it without the origin for now

src/plugins/ui_core_views/cell_evaluation/evaluator.ts
```diff
@@ -500,6 +509,9 @@ export class Evaluator {
         this.getters.getLocale(),
         cell
       );
+      if (!evaluatedCell.origin) {
+        evaluatedCell.origin = position;
```
LucasLefevre: an alternative approach to all this would be to add the origin when accessing the reference value in `refFn` and `range`.
It means a cell with a formula which is not returning a reference wouldn't have a `.origin` (nor literal cells), but that makes (also/more?) sense.

I would only be 2 places to change instead of the 5 or more here.

src/helpers/cells/cell_evaluation.ts
```diff
@@ -62,7 +64,8 @@ export function createEvaluatedCell(
 ): EvaluatedCell {
   const link = detectLink(functionResult.value);
   if (!link) {
-    return _createEvaluatedCell(functionResult, locale, cell);
+    const evaluateCell = _createEvaluatedCell(functionResult, locale, cell);
+    return addOrigin(evaluateCell, functionResult.origin);
```
LucasLefevre: if `addOrigin` mutates the input, it's better if it doesn't return it. At least it's very clear that it mutates in place

```suggestion
    const evaluateCell = _createEvaluatedCell(functionResult, locale, cell);
    addOrigin(evaluateCell, functionResult.origin);
    return evaluateCell;
```

tests/evaluation/evaluation.test.ts
```diff
@@ -1139,6 +1139,121 @@ describe("evaluateCells", () => {
       toCellPosition(sheetId, "A2")
     );
   });
+
+  describe("origin", () => {
+    let model, sheetId;
```
LucasLefevre: add types. It cannot be inferred and leads to `any`

tests/evaluation/evaluation.test.ts
```diff
@@ -1139,6 +1139,121 @@ describe("evaluateCells", () => {
...
+      model = new Model();
+      sheetId = model.getters.getActiveSheetId();
+    });
+
+    test("on literal", () => {
+      setCellContent(model, "A3", "5");
+      expect(getEvaluatedCell(model, "A3").origin).toMatchObject({
+        sheetId,
+        col: 0,
+        row: 2,
```
LucasLefevre: easier to read
```suggestion
        ...toCartesian("A3")
```

tests/evaluation/evaluation.test.ts
```diff
@@ -1139,6 +1139,121 @@ describe("evaluateCells", () => {
...
+  describe("origin", () => {
+    let model, sheetId;
+    beforeEach(() => {
+      model = new Model();
+      sheetId = model.getters.getActiveSheetId();
+    });
+
+    test("on literal", () => {
+      setCellContent(model, "A3", "5");
+      expect(getEvaluatedCell(model, "A3").origin).toMatchObject({
```
LucasLefevre:
```suggestion
      expect(getEvaluatedCell(model, "A3").origin).toEqual({
```

src/plugins/ui_core_views/cell_evaluation/evaluator.ts
```diff
@@ -369,7 +371,7 @@ export class Evaluator {
 
     if (!isMatrix(formulaReturn)) {
       const evaluatedCell = createEvaluatedCell(
-        nullValueToZeroValue(formulaReturn),
+        addOrigin(nullValueToZeroValue(formulaReturn), formulaPosition),
```
LucasLefevre: add position to createEvaluatedCell ?

src/helpers/cells/cell_evaluation.ts
```diff
@@ -191,3 +197,13 @@ function errorCell(value: string, message?: string): ErrorCell {
     defaultAlign: "center",
   };
 }
+
+export function addOrigin<T extends FunctionResultObject | EvaluatedCell>(
```
LucasLefevre: not needed to be generic anymore ?

PR #6532, opened 2025-06-03
[FIX] core: prevent full re-render in dashboards

Rachico:
Prior to this commit, when we type a value in an input on the side panel and hover back to the grid, the value is lost as we perform a full re-render of the model. This commit mitigates this behaviour by preventing the re-render when we are editing a spreadsheet of type `dashboard`.

Task: [4772026](https://www.odoo.com/odoo/2328/tasks/4772026)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/tables/hovered_table_store.ts
+++ b/src/components/tables/hovered_table_store.ts
@@ -19,7 +19,7 @@ export class HoveredTableStore extends SpreadsheetStore {
   }
 
   hover(position: Partial<Position>) {
-    if (position.col === this.col && position.row === this.row) {
+    if (!this.getters.isDashboard() || (position.col === this.col && position.row === this.row)) {
       return "noStateChange";
     }
     this.col = position.col;


src/components/tables/hovered_table_store.ts
```diff
@@ -19,6 +19,9 @@ export class HoveredTableStore extends SpreadsheetStore {
   }
 
   hover(position: Partial<Position>) {
+    if (!this.getters.isDashboard()) {
+      return "noStateChange";
+    }
     if (position.col === this.col && position.row === this.row) {
       return "noStateChange";
     }
```
hokolomopo:
```suggestion
    if (!this.getters.isDashboard() || position.col === this.col && position.row === this.row) {
      return "noStateChange";
    }
```
nitpick

hokolomopo:
@robodoo r+

PR #6549, opened 2025-06-06
[WIP] base_extractor: handle integer '0' correctly 

suju-odoo:
## Description:

Task: [4730308](https://www.odoo.com/odoo/my-tasks/4730308)
[Freeze Planning Numbers (1).xlsx](https://github.com/user-attachments/files/20633188/Freeze.Planning.Numbers.1.xlsx)

Tested Versions:
16.0 and later (unable to test earlier versions).

Steps to Reproduce:
1. Open the Documents app.
2. Import the attached Excel file created/exported from Numbers.
3. Attempt to open the file.
4. The operation will fail.

Background:
macOS Numbers omits the border_id attribute for a cell when applyBorder is 0. (same for fillId and other missing attributes)
Microsoft Excel, by contrast, always writes a border_id attribute, no matter what applyBorder is set to.

Now, odoo reads the cell

https://github.com/odoo/o-spreadsheet/blob/2ff9f1ab0ea3380c3897c24fb15e66d560b5bf57/src/xlsx/extraction/style_extractor.ts#L198-L213

`getStyles()` call `extractAttr()` with `optionalArgs.default = 0` 

https://github.com/odoo/o-spreadsheet/blob/2ff9f1ab0ea3380c3897c24fb15e66d560b5bf57/src/xlsx/extraction/base_extractor.ts#L167-L178

`attribute` is falsy, calling `handleMissingValue()`  

https://github.com/odoo/o-spreadsheet/blob/2ff9f1ab0ea3380c3897c24fb15e66d560b5bf57/src/xlsx/extraction/base_extractor.ts#L282-L298

Since `optionalArgs.default = 0`, `optionalArgs?.default` is falsy.
This lead to the system to throw error, instead of `addParsingWarning()` and assign the default value zero to the cell. 

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/xlsx/extraction/base_extractor.ts
+++ b/src/xlsx/extraction/base_extractor.ts
@@ -285,7 +285,7 @@ export class XlsxBaseExtractor {
     optionalArgs?: ExtractArg
   ) {
     if (optionalArgs?.required) {
-      if (optionalArgs?.default) {
+      if (optionalArgs?.default != null) {
         this.warningManager.addParsingWarning(
           `Missing required ${missingElementName} in element <${parentElement.tagName}> of ${this.currentFile}, replacing it by the default value ${optionalArgs.default}`
         );


src/xlsx/extraction/base_extractor.ts
```diff
@@ -285,7 +285,7 @@ export class XlsxBaseExtractor {
     optionalArgs?: ExtractArg
   ) {
     if (optionalArgs?.required) {
-      if (optionalArgs?.default) {
+      if (optionalArgs?.default != null) {
```
rrahir:
```suggestion
      if (optionalArgs?.default !== undefined) {
```
the typing (if correctly defined) does not allow null so the confition would always be true
```typescript
interface ExtractArg {
  required?: boolean;
  default?: string | boolean | number;
}
```

Otherwise, the fix looks good, thanks for investigating and for the extensive report :)

sunhyeokjung: @rrahir This makes sense! Thanks for the explanation

sunhyeokjung:
@rrahir  thank you for the review! I made requested changes 👋

rrahir:
Wonderful, thanks a lot!

robodoo r+

PR #6550, opened 2025-06-08
[FIX] Formats: text format should not mark the cell as non-empty

rrahir:
Currently, a cell without content but with a text format is interpreted as a non-empty cell which is incorrect. Regardless of the format applied, if the cells has no content, it should be marked as an empty evaluated cell.

Task: 4856923

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


pro-odoo:
typo in commit message: "bu" -> "but"

PR #6564, opened 2025-06-10
[FIX] menu: do not execute click callback on disabled menus

rrahir:
Following the split of the menu component in two parts, the logic that prevent an `Action` callback to be called when the menu is disabled was not correctly split and just existed in the `MenuPopover` component.

Task: 4859723

## Description:

Task: [4859723](https://www.odoo.com/odoo/2328/tasks/4859723)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/menu/menu.ts
+++ b/src/components/menu/menu.ts
@@ -192,4 +192,11 @@ export class Menu extends Component<MenuProps, SpreadsheetChildEnv> {
     menu.onStopHover?.(this.env);
     this.props.onMouseLeave?.(menu, ev);
   }
+
+  onClickMenu(menu: Action, ev: CustomEvent) {
+    if (!this.isEnabled(menu)) {
+      return;
+    }
+    this.props.onClickMenu?.(menu, ev);
+  }
 }
--- a/tests/menus/menu_component.test.ts
+++ b/tests/menus/menu_component.test.ts
@@ -1,0 +1,26 @@
+import { createActions } from "../../src/actions/action";
+import { Menu } from "../../src/components/menu/menu";
+import { simulateClick } from "../test_helpers/dom_helper";
+import { mountComponent } from "../test_helpers/helpers";
+
+describe("Menu component", () => {
+  test("Execute is not called when menu item is disabled", async () => {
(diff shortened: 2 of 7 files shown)


src/components/menu/menu.ts
```diff
@@ -192,4 +192,11 @@ export class Menu extends Component<MenuProps, SpreadsheetChildEnv> {
     menu.onStopHover?.(this.env);
     this.props.onMouseLeave?.(menu, ev);
   }
+
+  onClickMenu(menu: Action, ev: CustomEvent) {
```
hokolomopo, line 196: alternatively, this could be handled inside the generated `execute` of `createAction`. It would also work for the few places where we use an `Action` w/o a `Menu` component

tests/menus/menu_component.test.ts
```diff
@@ -0,0 +1,26 @@
+import { createActions } from "../../src/actions/action";
+import { Menu } from "../../src/components/menu/menu";
+import { simulateClick } from "../test_helpers/dom_helper";
+import { mountComponent } from "../test_helpers/helpers";
+
+describe("Menu component", () => {
```
hokolomopo, line 6: could be tested inside `context_menu_component.test.ts` no ? Or `menu_items_registry.test.ts` if we decide to implement the logic inside `createAction`

src/components/menu/menu.ts
rrahir: I added the logic in the action as well ,the only issue is that it's not sufficient by itself because the Menu components (like `RibbonMenu`) have some callbacks related to whether or not we execute the action. This means that we'd need to determine whether or not the execute callback of the actionSpec was actually called or rejected.
-> I kept the check in the  `Menu` component as well

tests/menus/menu_component.test.ts
rrahir: The context_menu (which is a `MenuPopover` actually passes this test since the check was only done in that Component.

hokolomopo (commented):
@robodoo r+ rebase-ff

PR #6566, opened 2025-06-11
[FIX] dashboard: remove unwanted hover effects

rmbh-odoo:
## Description:

This commit resolves a minor UI bug by disabling hover effects on non-interactive table areas.

Before this commit:
- A hover effect was incorrectly shown on empty table rows.
- The hover effect was also visible on table headers.

After this commit:
- The hover effect is only shown on data rows that contain content.
- The hover effect is now correctly disabled for empty rows and headers.

Technical Changes:
- Added 'TABLE_HOVER_BACKGROUND_COLOR' constant.
- Render logic now checks for content before applying hover effect.

Task: [4854451](https://www.odoo.com/odoo/2328/tasks/4854451)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/components/tables/hovered_table_store.ts
```diff
@@ -34,12 +34,8 @@ export class HoveredTableStore extends SpreadsheetStore {
   }
 
   private computeOverlay() {
-    if (!this.getters.isDashboard()) {
-      return;
-    }
```
dhrp-odoo, line 39: Why was this change made? I added it back and ran the tests, and everything still passes without any issues!

rmbh-odoo: Morning! 👋 I removed a dead condition from computeOverlay. It was redundant after we added a similar dashboard check to the hover function in pr #6532. Just tidying things up! ✨

Let me know if this condition is still needed! 😊

src/components/tables/hovered_table_store.ts
```diff
@@ -48,9 +44,25 @@ export class HoveredTableStore extends SpreadsheetStore {
     if (!table) {
       return;
     }
-    const { left, right } = table.range.zone;
+    const { left, right, top } = table.range.zone;
+    const isTableHeader = row < top + table.config.numberOfHeaders;
+    const isTableRowHasContent = (): boolean => {
```
hokolomopo:
```suggestion
    const doesTableRowHaveContent = (): boolean => {
```
nitpick. Or change to `isTableRowEmpty`

src/components/tables/hovered_table_store.ts
```diff
@@ -48,9 +44,25 @@ export class HoveredTableStore extends SpreadsheetStore {
...
+      for (let c = left; c <= right; c++) {
+        if (
+          !this.getters.isColHidden(sheetId, c) &&
+          this.getters.getEvaluatedCell({ sheetId, col: c, row })?.formattedValue.length > 0
```
hokolomopo:
```suggestion
          this.getters.getEvaluatedCell({ sheetId, col: c, row })?.formattedValue
```
nitpick

src/components/tables/hovered_table_store.ts
```diff
@@ -48,9 +44,25 @@ export class HoveredTableStore extends SpreadsheetStore {
     if (!table) {
       return;
     }
-    const { left, right } = table.range.zone;
+    const { left, right, top } = table.range.zone;
+    const isTableHeader = row < top + table.config.numberOfHeaders;
+    const isTableRowHasContent = (): boolean => {
+      for (let c = left; c <= right; c++) {
```
hokolomopo: nitpick: could be shortened with `range(left, right).some((col) => !isHidden && cell.evaluatedValue)`

hokolomopo (commented):
Some nitpicks, otherwise LGTM :)

tests/renderer_store.test.ts
```diff
@@ -471,45 +472,78 @@ describe("renderer", () => {
...
-    );
+  describe("Table cell hover effect", () => {
+    let model: Model;
+    let ctx: MockGridRenderingContext;
+    let container: DependencyContainer;
+    let drawGridRenderer: (ctx: GridRenderingContext) => void;
+    let fillStyles: any[];
+
     const background = "#DC6CDF";
     const hoverColor = "#017E8414";
```
LucasLefevre: you can use the new constant here 
```suggestion
    const hoverColor = TABLE_HOVER_BACKGROUND_COLOR;
```

tests/renderer_store.test.ts
```diff
@@ -471,45 +472,78 @@ describe("renderer", () => {
...
+      container.get(HoveredTableStore).hover({ col: 0, row: 1 });
+      drawGridRenderer(ctx);
+
+      expect(fillStyles).toEqual([
+        { color: background, h: 23, w: 96, x: 0, y: 23 },
+        { color: hoverColor, h: 23, w: 96, x: 0, y: 23 },
+      ]);
+    });
+
+    test("should not render a hover overlay on a header row with data", () => {
```
LucasLefevre: you could maybe add a dedicated test file to test the store directly with `makeStore` (without any rendering at all). It'd be much easier I think than those rendering tests.
Thanks!

src/components/tables/hovered_table_store.ts
```diff
@@ -48,9 +45,19 @@ export class HoveredTableStore extends SpreadsheetStore {
...
+    const doesTableRowHaveContent = range(left, right + 1).some((c) => {
+      return (
+        !this.getters.isColHidden(sheetId, c) &&
+        this.getters.getEvaluatedCell({ sheetId, col: c, row }).formattedValue
+      );
+    });
+
+    if (!isTableHeader && doesTableRowHaveContent) {
+      for (let c = left; c <= right; c++) {
+        this.overlayColors.set({ sheetId, col: c, row }, TABLE_HOVER_BACKGROUND_COLOR);
```
LucasLefevre: nitpicking
```suggestion
      for (let col = left; col <= right; col++) {
        this.overlayColors.set({ sheetId, col, row }, TABLE_HOVER_BACKGROUND_COLOR);
```

LucasLefevre (commented):
LGTM :+1:  except the test that could be changed to a store test :wink:

tests/table/hovered_table_store.test.ts
```diff
@@ -0,0 +1,86 @@
...
+import { createTable, setCellContent } from "../test_helpers/commands_helpers";
+import { makeStore } from "../test_helpers/stores";
+
+describe("Hovered Table Store", () => {
+  let hovered_table_store: HoveredTableStore;
+  let model: Model;
+  let sheetId: UID;
+
+  beforeEach(() => {
+    ({ model, store: hovered_table_store } = makeStore(HoveredTableStore));
```
LucasLefevre: :wink: 
```suggestion
    ({ model, store: hoveredTableStore } = makeStore(HoveredTableStore));
```

tests/table/hovered_table_store.test.ts
```diff
@@ -0,0 +1,86 @@
...
+    hovered_table_store.clear();
+    sheetId = model.getters.getActiveSheetId();
+    createTable(model, "A1:B2");
```
LucasLefevre: as a general rule, I don't like `beforeEach` with business stuff (updating the spreadsheet):

- you can't see and understand the full picture when looking at a test. You have to scroll up to see there's a table on A1:B2
- one day, someone will want to add a test but with another table config. So that person will add the table in the test, but then there are two tables while only one would be enough. The test won't be a minimal test

It's ok (better) to have the setup in each test, even if it's copy-pasted. I'd move `createTable(model, "A1:B2");` in each test

tests/table/hovered_table_store.test.ts
```diff
@@ -0,0 +1,86 @@
...
+    const headerPositionWithData = { col: 0, row: 0 };
+    setCellContent(model, "A1", "Header");
+    hovered_table_store.hover(headerPositionWithData);
+    expect(hovered_table_store.overlayColors.has({ sheetId, ...headerPositionWithData })).toBe(
+      false
+    );
+  });
+
+  test("Should store overlay for full data rows with content in dashboard mode", () => {
+    const position = { col: 0, row: 1 };
```
LucasLefevre: you can even name the position :) (same below)
```suggestion
    const A2 = { col: 0, row: 1 };
```

LucasLefevre (commented):
Nice, thanks!
Much easier to read and write tests that way!

tests/table/hovered_table_store.test.ts
rmbh-odoo: Noted your suggestion for future contributions 😊
Thank you! Let me know if there’s anything else I should improve 🌟

LucasLefevre (approved):
Thanks!
robodoo r+

PR #6575, opened 2025-06-11
[FIX] Table: avoid triggering an unecessary evaluation

rrahir:
When inserting a table, we first clear the merges taht could overlap. However, the command is dispatched regardless of the presence of an overlapping merge and the simple action of dispatching it will trigger a full reevaluation of the spreadsheet because `REMOVE_MERGE` is flagged as a command that invalidates the full evaluation.

Task: 4862862

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/tests/table/tables_plugin.test.ts
+++ b/tests/table/tables_plugin.test.ts
@@ -31,10 +31,16 @@ import {
   getFilter,
   getTable,
 } from "../test_helpers/getters_helpers";
-import { getFilterHiddenValues, toRangeData, toRangesData } from "../test_helpers/helpers";
+import {
+  getFilterHiddenValues,
+  getPlugin,
+  toRangeData,
+  toRangesData,
+} from "../test_helpers/helpers";
 
 import { DEFAULT_BORDER_DESC } from "../../src/constants";
 import { TABLE_PRESETS } from "../../src/helpers/table_presets";
+import { EvaluationPlugin } from "../../src/plugins/ui_core_views";
 import { TABLE_STYLE_ALL_RED } from "../test_helpers/constants";
 import { DEFAULT_TABLE_CONFIG } from "./../../src/helpers/table_presets";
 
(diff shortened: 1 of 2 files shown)


tests/table/tables_plugin.test.ts
```diff
@@ -116,6 +122,20 @@ describe("Table plugin", () => {
         createTable(model, "A1:A5");
         expect(merge(model, "A1:A2")).toBeCancelledBecause(CommandResult.MergeInTable);
       });
+
+      test("Inserting a table only invalidates the evaluation if it overlaps a merge", () => {
+        merge(model, "A1:B1");
+
+        // @ts-ignore
+        const evaluator = getPlugin(model, EvaluationPlugin).evaluator;
```
hokolomopo:
```suggestion
        const evaluator = getPlugin(model, EvaluationPlugin)["evaluator"];
```
nitpick. I don't like ts-ignore >_<

src/plugins/core/merge.ts
```diff
@@ -367,6 +369,13 @@ export class MergePlugin extends CorePlugin<MergeState> implements MergeState {
     return CommandResult.Success;
   }
 
+  private checkTargetValidity({ target }: AddMergeCommand | RemoveMergeCommand): CommandResult {
+    if (target.length === 0) {
```
hokolomopo: Do we have any commands that would be valid with an empty target ? If no, this could be bumped in the `allowDispatch` of the sheet plugin

src/plugins/core/merge.ts
```diff
@@ -24,6 +24,7 @@ import {
   HeaderIndex,
```
hokolomopo: typo in commit message `the merges taht could overlap`

src/plugins/core/merge.ts
rrahir: I don't think so, I haven't really thought it through but imagine a chart-like command with a target and you explicitely empty the list of range for whatever reason. I think it could vary from one use to the other

rrahir: I quickly implemented the check in sheet and it breaks on the CF commadns for insance (yes they are not. target but range dependent but the logic is pretty much the same. I don't think it'd be a good idea to make a generic change. Anywway I can just remove the commit from here and create a followup task

rrahir: > https://www.odoo.com/odoo/2328/tasks/4863945

PR #6602, opened 2025-06-14
[FIX] Model: change load order of rangeAdapter

rrahir:
RangeAdapter is an esential component of the core plugin operation as it allows to change the internal ranges of all plugins with a generic transformation.
However, the current situation is not ideal as it will handle all the commands before any other plugin, including the SheetPlugin.

This is an issue because we have a 2-side approach for treating ranges:
1. the Range object which are basically a zone and a sheetId, which seems enough information to work with
2. the xc, which do not hold the sheetId information but rather the sheetName

The RangeAdapter only handles ranges which means it is not designed to react to, let's say, a sheet name change.
the opposite is true as well, XC does not care about change of sheetId.

We could think that the range needs to know the sheet name and react to such a change but the design decisions up until now were not in that favor, we designed several flows, notably commands, that rely on the information of a Range (sheetId + Zone) the migration willbe a nightmare, we'll have to transform lots of stuff but feasible.

This fix is trying to keep the current situation "a range only needs to know the sheetId and zone;, the sheetName can be handled outbound"

With this moto  in mind, the problem we face is that RangeAdapter handles the 'rename_sheet' command which is sooooo specific to sheetPlugin before sheetPlugin handles it, ahd the range transformation (since the range don't know about the sheetname) cannot react properly to that change.

-> solution: move rangeAdapter elsewhere in the stack of plugins, more specifically, it be loaded just before another plugin that needs range adaptation.

## NOTE
the fix suggested here does not work because we implemented some tuff in SpreadsheetPivotCorePlugin that actually relies on the fact that the transformation did not occur yet. 
So on one hand, we want it to occur, for the string transformation, 
on the other hand, we don't want it for the _rangeData transformation.

The real solution should probably be to force every internal structure representation to ranges? which probably means keepng some sort of compiled formula instead of formula strings

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/core_plugin.ts
+++ b/src/plugins/core_plugin.ts
@@ -10,12 +10,10 @@ import {
 } from "../types";
 import { CoreGetters } from "../types/getters";
 import { BasePlugin } from "./base_plugin";
-import { RangeAdapter } from "./core/range";
 
 export interface CorePluginConfig {
   readonly getters: CoreGetters;
   readonly stateObserver: StateObserver;
-  readonly range: RangeAdapter;
   readonly dispatch: CoreCommandDispatcher["dispatch"];
   readonly canDispatch: CoreCommandDispatcher["dispatch"];
   readonly custom: ModelConfig["custom"];
@@ -41,9 +39,10 @@ export class CorePlugin<State = any>
   protected dispatch: CoreCommandDispatcher["dispatch"];
   protected canDispatch: CoreCommandDispatcher["dispatch"];
 
-  constructor({ getters, stateObserver, range, dispatch, canDispatch }: CorePluginConfig) {
+  adaptRanges?(applyChange: ApplyRangeChange, sheetId?: UID, sheetName?: string): void;
+
+  constructor({ getters, stateObserver, dispatch, canDispatch }: CorePluginConfig) {
     super(stateObserver);
-    range.addRangeProvider(this.adaptRanges.bind(this));
     this.getters = getters;
     this.dispatch = dispatch;
     this.canDispatch = canDispatch;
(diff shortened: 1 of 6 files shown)


src/plugins/core_plugin.ts
```diff
@@ -66,7 +67,7 @@ export class CorePlugin<State = any>
    * @param applyChange a function that, when called, will adapt the range according to the change on the grid
    * @param sheetId an optional sheetId to adapt either range of that sheet specifically, or ranges pointing to that sheet
    */
-  adaptRanges(applyChange: ApplyRangeChange, sheetId?: UID, sheetName?: string): void {}
+  // adaptRanges?(applyChange: ApplyRangeChange, sheetId?: UID, sheetName?: string): void {}
```
LucasLefevre, line 68: :eyes:

src/plugins/core_plugin.ts
```diff
@@ -41,9 +41,10 @@ export class CorePlugin<State = any>
   protected dispatch: CoreCommandDispatcher["dispatch"];
   protected canDispatch: CoreCommandDispatcher["dispatch"];
 
-  constructor({ getters, stateObserver, range, dispatch, canDispatch }: CorePluginConfig) {
+  adaptRanges?(applyChange: ApplyRangeChange, sheetId?: UID, sheetName?: string): void;
+
+  constructor({ getters, stateObserver, dispatch, canDispatch }: CorePluginConfig) {
```
LucasLefevre, line 44: we can remove `range` from `CorePluginConfig` then :wink:

LucasLefevre (commented):
could you also add a test for the issue about the gauge chart described in the task ?

src/plugins/core_plugin.ts
rrahir: eh ben ,,doesn't work ><

PR #6615, opened 2025-06-17
[PERF] collaborative: avoid useless transformations

LucasLefevre:
before: 5+minutes
after:  <2 seconds

dashboard_id 184 on odoo.com is very slow to load: 5+minutes https://www.odoo.com/odoo/dashboards?dashboard_id=184

It's caused by a huge number of transformations performed while replaying revisions upon loading the model.

Specifically, there are 2 revisions with 20k commands each. All commands are UPDATE_CELL or SET_BORDERS. These revisions most probably originates from a huge copy & paste.

Because of history changes (UNDO/REDO) in a collaborative context, those 2 revisions are transformed one against the other. With the current code, it leads to 20k*20k=400M transformations.

One key observation to fix the problem is that an executed UPDATE_CELL or SET_BORDERS won't ever actually transform any command. All transformations performed are actually doing nothing and returns the original command untouched.
Note that all executed commands which are part of the generic transformations are also in the registry.

With this commit, we only call the transformation if the executed command is part of the transformation registry (=it can transform other commands). I had to change the order in `OTRegistry` to be able to use `.getKeys()`

Also, I took the opportunity to replace `.map().filter()` with a single `reduce` to save one array allocation.

Task: [4873461](https://www.odoo.com/odoo/2328/tasks/4873461)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/collaborative/ot/ot.ts
+++ b/src/collaborative/ot/ot.ts
@@ -67,10 +67,20 @@ export function transformAll(
   executed: readonly CoreCommand[]
 ): CoreCommand[] {
   let transformedCommands = [...toTransform];
+  const possibleTransformations = new Set(otRegistry.getKeys());
   for (const executedCommand of executed) {
-    transformedCommands = transformedCommands
-      .map((cmd) => transform(cmd, executedCommand))
-      .filter(isDefined);
+    // If the executed command is not in the registry, we skip it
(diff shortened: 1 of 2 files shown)


rrahir:
robodoo r+

PR #6616, opened 2025-06-17
[FIX] Selection: Fix navigation through merges

rrahir:
How to reproduce:
- Create a merge in C3:D4
- click on C5
- Press Arrow-Up (this will select C3:D4)
- press Shift+ArrowDown several times

-> your selection gets stuck to C3:D5

This occurs because the method that extends the selection does not behave correctly when the anchor is set on a Merge, it assumes that the anchor is always a single cell.

Task: 4873718

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/selection_stream/selection_stream_processor.ts
+++ b/src/selection_stream/selection_stream_processor.ts
@@ -225,26 +225,28 @@ export class SelectionStreamProcessor
       };
     };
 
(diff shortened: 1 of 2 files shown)


pro-odoo (commented):
robodoo r+

PR #6624, opened 2025-06-18
[FIX] data_validation: preserve rule order when updating a rule

dhrp-odoo:
## Description:

Previously, when multiple data validation rules existed, updating a rule caused it to move to the end of the list, disrupting the original order.

This PR ensures that the rule order is preserved when a rule is edited and saved.

Task: [4863726](https://www.odoo.com/odoo/2328/tasks/4863726)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/core/data_validation.ts
+++ b/src/plugins/core/data_validation.ts
@@ -172,7 +172,7 @@ export class DataValidationPlugin
       newRule.criterion.values = Array.from(new Set(newRule.criterion.values));
     }
 
-    const adaptedRules = this.removeRangesFromRules(sheetId, newRule.ranges, rules);
+    const adaptedRules = this.removeRangesFromRules(sheetId, newRule.ranges, rules, newRule.id);
     const ruleIndex = adaptedRules.findIndex((rule) => rule.id === newRule.id);
 
     if (ruleIndex !== -1) {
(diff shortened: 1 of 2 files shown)


src/plugins/core/data_validation.ts
```diff
@@ -179,7 +180,9 @@ export class DataValidationPlugin
       adaptedRules[ruleIndex] = newRule;
       this.history.update("rules", sheetId, adaptedRules);
     } else {
-      this.history.update("rules", sheetId, [...adaptedRules, newRule]);
+      const insertIndex = indexInOriginalRules !== -1 ? indexInOriginalRules : adaptedRules.length;
+      adaptedRules.splice(insertIndex, 0, newRule);
+      this.history.update("rules", sheetId, adaptedRules);
```
hokolomopo: removeRangesFromRules could remove a bunch of data validation rules, and in this case your indexInOriginalRules could be totally wrong and out of bounds, no ? 

We could just add an additional argument to removeRangesFromRules so it doesn't delete the rule we're editing (that's why the ranges were moving around in the first place, right ?)

dhrp-odoo: If multiple rules were removed and the index went out of bounds, splice would insert the rule at the end?

That said, suggested solution passing additional argument is definitely simpler and cleaner — I’ve updated the code accordingly. Thanks!

hokolomopo: Oh i didn't know splice did that. That would still have been the wrong index, but not a breaking issue

rrahir:
robodoo r+

PR #6625, opened 2025-06-19
[FIX] GridComposer: Recompute composer position if we change edited cell

rrahir:
Currently, the gridComposer position is not recomputed if we start editing another cell, that is, if we decide to manipulate the composerStore outside of a standard flow where the edition state is reset when changing editing cells.

Task: 4879186

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


tests/grid/__snapshots__/grid_component.test.ts.snap
```diff
@@ -88,7 +88,7 @@ exports[`Grid component simple rendering snapshot 1`] = `
           class="o-composer w-100 text-start"
           contenteditable="true"
           spellcheck="false"
-          style="max-height: 1008.6px;"
+          style="max-height: 982.6px;"
```
hokolomopo: not sure why the max-height of the inactive grid composer changed with this fix ?

rrahir: Indeed, the condition should have taken in account the fact that we edited. I suspect that the first placement is always wrong and only becomes correct once we start editing.

LucasLefevre (commented):
robodoo r+

PR #6671, opened 2025-06-24
[FIX] header_size: resize row based on link label instead of full link

rmbh-odoo:
## Description:

Before this commit:
- Row height was calculated using the full Markdown link (label + URL), 
which could lead to incorrect row resizing when word wrap was enabled.

After this commit:
- Only the link label is used to compute row height,
resulting in accurate sizing.

Task: [4886598](https://www.odoo.com/odoo/2328/tasks/4886598)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


hokolomopo (commented):
@robodoo r+

PR #6673, opened 2025-06-24
[FIX] animations: fix pivot collapse animation

hokolomopo:
## Description

The cells animations on pivot collapse were not working correctly. The headers of the pivots would appear suddenly without any animation.

This is because the `textChange` animation isn't working when text is appearing or disappearing. This animation should not even be used in that case, we should use `fadeIn` and `fadeOut` animations instead.

Task: [4894210](https://www.odoo.com/odoo/2328/tasks/4894210)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/registries/cell_animation_registry.ts
+++ b/src/registries/cell_animation_registry.ts
@@ -107,6 +107,57 @@ cellAnimationRegistry.add("textFadeOut", {
   },
 });
 
+cellAnimationRegistry.add("iconFadeIn", {
+  id: "iconFadeIn",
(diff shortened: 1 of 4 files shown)


pro-odoo:
@hokolomopo Should target saas-18.4, right ?

hokolomopo:
> @hokolomopo Should target saas-18.4, right ?
 
indeed it should. I re-targeted it :)

PR #6677, opened 2025-06-26
[FIX] composer: zone selection with arrow keys

dhrp-odoo:
## Description:

The range selection with arrow keys in the composer was previously broken for the first row and column.

This happened because captureSelection used `col || zone.left` and `row || zone.top`, which incorrectly treated 0 as falsy and fell back to the zone boundaries.

This PR replaces `||` with the nullish coalescing operator (??) to correctly preserve 0 values.

Task: [4890605](https://www.odoo.com/odoo/2328/tasks/4890605)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


pro-odoo (approved):
robodoo r+

PR #6689, opened 2025-06-26
[FIX] spreadsheet: unwrap 1x1 matrix results across rules (CF, DV, filters, charts)

dhrp-odoo:
## Description:

Previously, formulas returning a 1x1 matrix (e.g. `=IF(TRUE, A1, "else")`) were not properly handled across multiple features like **Conditional Formatting**, **Data Validation**, **Filters**, and **Gauge Chart configuration**.

These formulas were ignored because the evaluated result was a matrix, and we did not unwrap it to a scalar value. As a result:

* Conditional formatting styles were not applied
* Data validation rules silently failed
* Filter conditions were skipped
* Gauge chart configuration failed validation
* Users could enter invalid 1x1 formulas that violated blocking rules

### This PR fixes all of the above:

* Unwraps 1x1 matrices into scalar values before evaluating rules
* Applies this logic consistently across CF, DV, filters, and gauge charts
* Prevents 1x1 formulas from being accepted when they violate blocking DV rules
* Adds tests to cover these behaviors

Task: [4876691](https://www.odoo.com/odoo/2328/tasks/4876691)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_core_views/evaluation_conditional_format.ts
+++ b/src/plugins/ui_core_views/evaluation_conditional_format.ts
@@ -1,4 +1,5 @@
 import { compile } from "../../formulas";
+import { isMultipleElementMatrix, toScalar } from "../../functions/helper_matrices";
 import { parseLiteral } from "../../helpers/cells";
 import { colorNumberString, getColorScale, isInside, percentile } from "../../helpers/index";
 import { clip, largeMax, largeMin, lazy } from "../../helpers/misc";
@@ -6,7 +7,6 @@ import { criterionEvaluatorRegistry } from "../../registries/criterion_registry"
 import {
   CellIsRule,
   CellPosition,
-  CellValue,
   CellValueType,
   ColorScaleMidPointThreshold,
   ColorScaleRule,
@@ -24,7 +24,6 @@ import {
   UID,
   Zone,
   invalidateCFEvaluationCommands,
-  isMatrix,
 } from "../../types/index";
 import { CoreViewPlugin } from "../core_view_plugin";
 import { CoreViewCommand, invalidateEvaluationCommands } from "./../../types/commands";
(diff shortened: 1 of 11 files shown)


src/plugins/ui_core_views/evaluation_conditional_format.ts
```diff
@@ -366,13 +367,24 @@ export class EvaluationConditionalFormatPlugin extends CoreViewPlugin {
     const { sheetId } = target;
     const evaluator = criterionEvaluatorRegistry.get(rule.operator);
 
+    let hasMultipleElementMatrix = false;
     const evaluatedCriterionValues = rule.values.map((value) => {
       if (!value.startsWith("=")) {
         return parseLiteral(value, DEFAULT_LOCALE);
       }
-      return this.getters.evaluateFormula(sheetId, value) ?? "";
+      const result = this.getters.evaluateFormula(sheetId, value);
```
LucasLefevre: It looks like something like this would also work and is shorter.
It also allows to remove the type casting

```diff
@@ -372,13 +372,13 @@ export class EvaluationConditionalFormatPlugin extends CoreViewPlugin {
       }
       return this.getters.evaluateFormula(sheetId, value) ?? "";
     });
-    if (evaluatedCriterionValues.some(isMatrix)) {
+    if (evaluatedCriterionValues.some(isMultipleElementMatrix)) {
       return false;
     }
 
     const evaluatedCriterion = {
       type: rule.operator,
-      values: evaluatedCriterionValues as CellValue[],
+      values: evaluatedCriterionValues.map(toScalar),
     };
     return evaluator.isValueValid(cell.value ?? "", evaluatedCriterion, this.getters, sheetId);
   }
```

LucasLefevre (commented):
Hi :wave: 
When I grep ` (evaluatedCriterionValues.some(isMatr` in the code, the same pattern appears for filter and data validation.
Could you check if those are also affected by the same bug ?

LucasLefevre (approved):
Thanks for all the fixes!
robodoo r+ rebase-ff

PR #6716, opened 2025-06-30
FIX] Cells: content with new lines is always a string

rrahir:
## Description:
How to reproduce:

- in a cell, write 5 and then a bunch of new lines (Alt+Enter)
- validate the input

=> we only keep the value 5 and trim the rest

Task: [4907670](https://www.odoo.com/odoo/2328/tasks/4907670)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/numbers.ts
+++ b/src/helpers/numbers.ts
@@ -1,5 +1,5 @@
 import { Locale } from "../types";
-import { escapeRegExp, memoize } from "./misc";
+import { escapeRegExp, memoize, whiteSpaceCharacters } from "./misc";
 
 /**
  * This function returns a regexp that is supposed to be as close as possible as the numberRegexp,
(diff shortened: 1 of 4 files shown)


src/helpers/numbers.ts
```diff
@@ -50,7 +60,7 @@ const getNumberRegex = memoize(function getNumberRegex(locale: Locale) {
 export function isNumber(value: string | undefined, locale: Locale): boolean {
   if (!value) return false;
   // TO DO: add regexp for DATE string format (ex match: "28 02 2020")
-  return getNumberRegex(locale).test(value.trim());
+  return getNumberRegex(locale).test(value.replaceAll(spaceTrimRegex, ""));
```
hokolomopo, line 63: Maybe a small comment explaining that trim() also removes newLines

src/helpers/numbers.ts
```diff
@@ -50,7 +60,7 @@ const getNumberRegex = memoize(function getNumberRegex(locale: Locale) {
 export function isNumber(value: string | undefined, locale: Locale): boolean {
```
hokolomopo, line 60: `\n\n a/random/date` is still detected as a date, while `\n\n aRandomNumber` is detected as a string. We should be consistent

hokolomopo (commented):
Not 100% sure if this doesn't belong in master as a change in behavior, rather than a fix in 17.0 🤷

rrahir:
this one probably makes more sense and covers more cases https://github.com/odoo/o-spreadsheet/pull/6735

PR #6723, opened 2025-07-01
[FIX] auto_complete: remove toggle button for data validation

rmbh-odoo:
## Description:

Before this pr:
- The autocomplete assistant in data validation
included an open/close toggle button.

After this pr:
- The toggle button is removed in the data validation context to simplify the UI
and prevent accidental closure.

Task: [4854464](https://www.odoo.com/odoo/2328/tasks/4854464)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/composer/composer/composer.ts
+++ b/src/components/composer/composer/composer.ts
@@ -608,10 +608,12 @@ export class Composer extends Component<CellComposerProps, SpreadsheetChildEnv> 
   }
 
   closeAssistant() {
+    if (!this.canBeToggled) return;
     this.assistant.forcedClosed = true;
   }
 
   openAssistant() {
+    if (!this.canBeToggled) return;
     this.assistant.forcedClosed = false;
   }
 
@@ -623,6 +625,10 @@ export class Composer extends Component<CellComposerProps, SpreadsheetChildEnv> 
     ) {
       event.stopPropagation();
     }
+  }
+
+  get canBeToggled() {
+    return this.autoCompleteState.provider?.canBeToggled ?? true;
   }
 
   // ---------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


tests/composer/autocomplete_dropdown_component.test.ts
```diff
@@ -382,6 +382,22 @@ describe("Functions autocomplete", () => {
   });
 });
 
+describe("Data validation autocomplete", () => {
```
rrahir, line 385: Could you add a test about how hiding another assistant type (typically the formula assistant) has no impact on the visibility of this type of autocomplete?

src/registries/auto_completes/auto_complete_registry.ts
```diff
@@ -23,6 +23,7 @@ export interface AutoCompleteProvider {
   proposals: AutoCompleteProposal[];
   selectProposal(text: string): void;
   autoSelectFirstProposal: boolean;
+  hideAutoCompleteBtn: boolean;
```
rrahir: I don't think it's a good name to mention the autocomplete button as it's 100% UI related while this object is merely a data structure. Also, I find the conditions it appears in are not clear. Maybe something like `canBeToggled`/`isHid(e)able`/`isFoldable`  ?

src/components/composer/composer/composer.ts
```diff
@@ -822,7 +822,9 @@ export class Composer extends Component<CellComposerProps, SpreadsheetChildEnv>
   }
 
   private autoComplete(value: string) {
-    if (!value || this.assistant.forcedClosed) {
+    const isAutoCompleteOpen =
```
rrahir: both `closeAssistant` and `openAssistant` could return early if we cannot actually impact the visibility

hokolomopo (commented):
@robodoo r+

Thanks for the PR !

PR #6725, opened 2025-07-01
[FIX] Cells: newline is not a valid date separator

rrahir:
How to reproduce:
- Write "5\n5" in a cell

-> the cell is in error

It is detected as a date but the format parser crashes as it does not expect a newline character and rightfully so. If the user purposely adds a newline character, it is clear they are not trying to write something that looks like a number.

Task: 4910327

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/dates.ts
+++ b/src/helpers/dates.ts
@@ -3,7 +3,7 @@
 // -----------------------------------------------------------------------------
 
 import { CellValue, Format, Locale } from "../types";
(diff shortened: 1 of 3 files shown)


LucasLefevre (commented):
robodoo r+

PR #6737, opened 2025-07-01
[IMP] pivot: allow grouping of dimension values

hokolomopo:
## Description

With this commit, it is now possible to group dimension values of a pivot. The grouping is done using right-click context menu of the pivot cells.

Grouping pivot headers together creates a new field in the pivot, that defines which values are grouped together. Those new `customFields` are defined in the pivot definition.

Task: [4266072](https://www.odoo.com/odoo/2328/tasks/4266072)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/pivot/pivot_helpers.ts
+++ b/src/helpers/pivot/pivot_helpers.ts
@@ -7,6 +7,7 @@ import {
   toBoolean,
   toNumber,
   toString,
+  toValue,
 } from "../../functions/helpers";
 import { Registry } from "../../registries/registry";
 import { _t } from "../../translation";
@@ -22,13 +23,17 @@ import {
 import { EvaluationError } from "../../types/errors";
 import {
   Granularity,
+  PivotCoreDefinition,
   PivotCoreDimension,
+  PivotCustomGroupedField,
   PivotDimension,
   PivotDomain,
   PivotField,
+  PivotFields,
   PivotSortedColumn,
   PivotTableCell,
 } from "../../types/pivot";
+import { getUniqueText, isDefined } from "../misc";
 import { PivotRuntimeDefinition } from "./pivot_runtime_definition";
 import { pivotTimeAdapter } from "./pivot_time_adapter";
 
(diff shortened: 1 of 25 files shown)


src/helpers/pivot/pivot_helpers.ts
```diff
@@ -308,3 +317,95 @@ export function isSortedColumnValid(sortedColumn: PivotSortedColumn, pivot: Pivo
...
+  for (const customField of Object.values(definition.customFields || {})) {
+    const parentField = fields[customField.parentField];
+    if (!parentField) {
+      continue;
+    }
+    newFields[customField.name] = {
+      type: "custom",
+      isCustomField: true,
+      name: customField.name,
+      string: customField.name,
```
hokolomopo, line 367: With current implementation, for custom fields the `fieldName` is the same as the `field.string` (field label). Is it worth differenciating the two ? And maybe add a prefix to the `fieldName` ? Do we care ?

src/helpers/pivot/pivot_helpers.ts
```diff
@@ -266,7 +275,8 @@ pivotNormalizationValueRegistry
   .add("datetime", normalizeDateTime)
   .add("integer", (value) => toNumber(value, DEFAULT_LOCALE))
   .add("boolean", (value) => toBoolean(value))
-  .add("char", (value) => toString(value));
+  .add("char", (value) => toString(value))
+  .add("custom", (value) => value);
```
pro-odoo, line 279: `toNormalizedPivotValue` early returns if dimension.type === `custom`, so this one seems useless

pro-odoo: Or remove the early return to keep consistency

src/helpers/pivot/pivot_helpers.ts
pro-odoo: It's a bit "capilotracté", but if you create a custom group for the field Stage, it creates `Stage2`. After that, if you rename another field in the source to `Stage2`, the custom group replaces the `Stage2` field.

src/helpers/pivot/pivot_helpers.ts
hokolomopo: I cannot remove the early return, because `toNormalizedPivotValue` does `toString` to the group value, which I explicitly don't want to do. But yeah, it's useless to add it in the registry then.

pro-odoo:
robodoo r+

PR #6747, opened 2025-07-02
[FIX] Composer: Remove formatting when pasting external content

rrahir:
Currently, if we copy html content from outside and paste it in the composer, we let the full formatting take place and then override the content with our own. However, we did not consider this possibility when we introduced a "diff" strategy to override the composer content and in some cases, if the html has a color style attribute that is not a color (inherit, initial, ...) the code crashes.

That being said, Good news everyone, the attribute removed in #3922 is finally available on all browsers!! `plaintext-only` will automatically clear the formatting of a content pasted inside a contentEditable div.

Task: 4910559

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


LucasLefevre (commented):
robodoo r+

PR #6755, opened 2025-07-03
[REF] borders: refactor border for infinite spreadsheet

fdamhaut:
This commit refactor the border plugin to save border style
based on zones with similar borders style instead of saving cell-wise borders.

The goal of this commit was to allow the writing and editing of style on large
zones without the current performance dip.

Groundwork for the infinite viewport task

Task: 4852413

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/clipboard_handlers/borders_clipboard.ts
+++ b/src/clipboard_handlers/borders_clipboard.ts
@@ -1,43 +1,61 @@
-import { positionToZone, recomputeZones } from "../helpers";
+import { splitZoneForPaste } from "../helpers/clipboard/clipboard_helpers";
+import { deepEquals, groupConsecutive } from "../helpers/misc";
+import { ZoneBorder, ZoneBorderData } from "../plugins/core";
 import {
-  Border,
-  CellPosition,
+  BorderDescr,
+  BorderPosition,
   ClipboardCellData,
   ClipboardOptions,
   ClipboardPasteTarget,
   HeaderIndex,
   UID,
-  Zone,
 } from "../types";
 import { AbstractCellClipboardHandler } from "./abstract_cell_clipboard_handler";
 
 type ClipboardContent = {
-  borders: (Border | null)[][];
+  borders: ZoneBorder[];
+  width: number;
+  height: number;
 };
 
 export class BorderClipboardHandler extends AbstractCellClipboardHandler<
   ClipboardContent,
-  Border | null
+  ZoneBorder
 > {
-  private queuedBordersToAdd: Record<string, Zone[]> = {};
-
   copy(data: ClipboardCellData): ClipboardContent | undefined {
     const sheetId = data.sheetId;
     if (data.zones.length === 0) {
       return;
     }
-    const { rowsIndexes, columnsIndexes } = data;
-    const borders: (Border | null)[][] = [];
-
-    for (const row of rowsIndexes) {
-      const bordersInRow: (Border | null)[] = [];
-      for (const col of columnsIndexes) {
-        const position = { col, row, sheetId };
-        bordersInRow.push(this.getters.getCellBorder(position));
+    const borders: ZoneBorder[] = [];
+    let colsBefore = 0;
+    for (const cols of groupConsecutive(data.columnsIndexes)) {
+      let rowsBefore = 0;
+      for (const rows of groupConsecutive(data.rowsIndexes)) {
+        const zone = {
+          left: cols[0],
+          right: cols[cols.length - 1],
+          top: rows[0],
+          bottom: rows[rows.length - 1],
+        };
+        borders.push(
+          ...this.getters.getBorders(sheetId, zone).map(({ zone: borderZone, style }) => {
+            return {
+              zone: {
+                left: borderZone.left - zone.left + colsBefore,
+                right: borderZone.right && borderZone.right - zone.left + colsBefore,
+                top: borderZone.top - zone.top + rowsBefore,
+                bottom: borderZone.bottom && borderZone.bottom - zone.top + rowsBefore,
+              },
+              style,
+            };
+          })
+        );
+        rowsBefore += rows.length;
       }
-      borders.push(bordersInRow);
+      colsBefore += cols.length;
     }
-    return { borders };
+    return { borders, width: data.columnsIndexes.length, height: data.rowsIndexes.length };
   }
 
   paste(target: ClipboardPasteTarget, content: ClipboardContent, options: ClipboardOptions) {
@@ -47,47 +65,69 @@ export class BorderClipboardHandler extends AbstractCellClipboardHandler<
     }
     const zones = target.zones;
     if (!options.isCutOperation) {
-      this.pasteFromCopy(sheetId, zones, content.borders);
+      for (const zone of zones) {
+        for (const pasteZone of splitZoneForPaste(zone, content.width, content.height)) {
+          this.pasteBorderZone(sheetId, pasteZone.left, pasteZone.top, content.borders);
+        }
+      }
     } else {
       const { left, top } = zones[0];
-      this.pasteZone(sheetId, left, top, content.borders);
+      this.pasteBorderZone(sheetId, left, top, content.borders);
     }
-
-    this.executeQueuedChanges(sheetId);
   }
 
-  pasteZone(sheetId: UID, col: HeaderIndex, row: HeaderIndex, borders: (Border | null)[][]) {
-    for (const [r, rowBorders] of borders.entries()) {
-      for (const [c, originBorders] of rowBorders.entries()) {
-        const position = { col: col + c, row: row + r, sheetId };
-        this.pasteBorder(originBorders, position);
+  pasteBorderZone(sheetId: UID, col: HeaderIndex, row: HeaderIndex, borders: ZoneBorder[]) {
+    for (const border of borders) {
+      const zone = {
+        left: border.zone.left + col,
+        right: (border.zone.right && border.zone.right + col) || border.zone.left + col,
+        top: border.zone.top + row,
+        bottom: (border.zone.bottom && border.zone.bottom + row) || border.zone.top + row,
+      };
+      for (const [position, style] of this.getOptimalBorderCommands(border.style)) {
+        if (style)
+          this.dispatch("SET_ZONE_BORDERS", {
+            sheetId,
+            target: [zone],
+            border: { position, ...style },
+          });
       }
     }
   }
 
-  /**
-   * Paste the border at the given position to the target position
-   */
-  private pasteBorder(originBorders: Border | null, target: CellPosition) {
-    const targetBorders = this.getters.getCellBorder(target);
-    const border = {
-      ...targetBorders,
-      ...originBorders,
-    };
-    const borderKey = JSON.stringify(border);
-    if (!this.queuedBordersToAdd[borderKey]) {
-      this.queuedBordersToAdd[borderKey] = [];
+  getOptimalBorderCommands(border: ZoneBorderData): [BorderPosition, BorderDescr | undefined][] {
+    const hv = deepEquals(border.horizontal, border.vertical);
+    const external = deepEquals(border.left, border.right, border.top, border.bottom);
+    if (hv && external && deepEquals(border.horizontal, border.left)) {
+      return [["all", border.top]];
+    } else if (hv && external) {
+      return [
+        ["hv", border.horizontal],
+        ["external", border.top],
+      ];
+    } else if (external) {
+      return [
+        ["h", border.horizontal],
+        ["v", border.vertical],
+        ["external", border.top],
+      ];
+    } else if (hv) {
+      return [
+        ["hv", border.horizontal],
+        ["top", border.top],
+        ["bottom", border.bottom],
+        ["left", border.left],
+        ["right", border.right],
+      ];
+    } else {
+      return [
+        ["h", border.horizontal],
+        ["v", border.vertical],
+        ["top", border.top],
+        ["bottom", border.bottom],
+        ["left", border.left],
+        ["right", border.right],
+      ];
     }
-    this.queuedBordersToAdd[borderKey].push(positionToZone(target));
-  }
-
-  private executeQueuedChanges(pasteSheetTarget: UID) {
-    for (const borderKey in this.queuedBordersToAdd) {
-      const zones = this.queuedBordersToAdd[borderKey];
-      const border = JSON.parse(borderKey) as Border;
-      const target = recomputeZones(zones, []);
-      this.dispatch("SET_BORDERS_ON_TARGET", { sheetId: pasteSheetTarget, target, border });
-    }
-    this.queuedBordersToAdd = {};
   }
 }
--- a/src/clipboard_handlers/index.ts
+++ b/src/clipboard_handlers/index.ts
@@ -28,7 +28,7 @@ clipboardHandlersRegistries.cellHandlers
   .add("cell", CellClipboardHandler)
   .add("sheet", SheetClipboardHandler)
   .add("merge", MergeClipboardHandler)
-  .add("border", BorderClipboardHandler)
   .add("table", TableClipboardHandler)
   .add("conditionalFormat", ConditionalFormatClipboardHandler)
-  .add("references", ReferenceClipboardHandler);
+  .add("references", ReferenceClipboardHandler)
+  .add("border", BorderClipboardHandler);
--- a/src/helpers/cells/position_map.ts
+++ b/src/helpers/cells/position_map.ts
@@ -34,6 +34,10 @@ export class PositionMap<T> {
     return this.map[sheetId];
   }
 
+  clearSheet(sheetId: UID) {
+    delete this.map[sheetId];
+  }
+
   has({ sheetId, col, row }: CellPosition): boolean {
     return this.map[sheetId]?.[col]?.[row] !== undefined;
   }
--- a/src/helpers/cells/position_map.ts
+++ b/src/helpers/cells/position_map.ts
@@ -1,4 +1,4 @@
-import { CellPosition, UID } from "../../../types";
+import { CellPosition, UID } from "../..";
 
 export class PositionMap<T> {
   private map: Record<UID, Record<number, Record<number, T>>> = {};
@@ -18,6 +18,12 @@ export class PositionMap<T> {
       map[sheetId][col] = {};
     }
     map[sheetId][col][row] = value;
+  }
+
+  setMany(values: Iterable<[CellPosition, T]>) {
+    for (const [position, value] of values) {
+      this.set(position, value);
+    }
   }
 
   get({ sheetId, col, row }: CellPosition): T | undefined {
--- a/src/helpers/zones.ts
+++ b/src/helpers/zones.ts
@@ -1,4 +1,12 @@
-import { CellPosition, Position, UID, UnboundedZone, Zone, ZoneDimension } from "../types";
+import {
+  AdjacentEdge,
+  CellPosition,
+  Position,
+  UID,
+  UnboundedZone,
+  Zone,
+  ZoneDimension,
+} from "../types";
 import {
   MAX_COL,
   MAX_ROW,
@@ -347,15 +355,15 @@ export function unionUnboundedZones(...zones: UnboundedZone[]): UnboundedZone {
 /**
  * Compute the intersection of two zones. Returns nothing if the two zones don't overlap
  */
-export function intersection(z1: Zone, z2: Zone): Zone | undefined {
+export function intersection(z1: UnboundedZone, z2: Zone): Zone | undefined {
   if (!overlap(z1, z2)) {
     return undefined;
   }
   return {
     top: Math.max(z1.top, z2.top),
     left: Math.max(z1.left, z2.left),
-    bottom: Math.min(z1.bottom, z2.bottom),
-    right: Math.min(z1.right, z2.right),
+    bottom: z1.bottom !== undefined ? Math.min(z1.bottom, z2.bottom) : z1.bottom ?? z2.bottom,
+    right: z1.right !== undefined ? Math.min(z1.right, z2.right) : z1.right ?? z2.right,
   };
 }
 
@@ -363,20 +371,68 @@ export function intersection(z1: Zone, z2: Zone): Zone | undefined {
  * Two zones are equal if they represent the same area, so we clearly cannot use
  * reference equality.
  */
-export function isEqual(z1: Zone, z2: Zone): boolean {
+export function isEqual(z1: UnboundedZone, z2: UnboundedZone): boolean {
   return (
-    z1.left === z2.left && z1.right === z2.right && z1.top === z2.top && z1.bottom === z2.bottom
+    z1.left === z2.left &&
+    z1.right === z2.right &&
+    z1.top === z2.top &&
+    z1.bottom === z2.bottom &&
+    z1.hasHeader === z2.hasHeader
   );
+}
+
+/**
+ * Two zones are adjacent if they -partially- share an edge.
+ * Returns the adjacent size of z1 as well as the indexes of the header by which they are adjacent.
+ */
+export function adjacent(z1: UnboundedZone, z2: Zone): AdjacentEdge | undefined {
+  if (intersection(z1, z2)) return undefined;
+  let adjacentEdge: AdjacentEdge | undefined = undefined;
+  if (z1.left === z2.right + 1) {
+    adjacentEdge = {
+      position: "left",
+      start: Math.max(z1.top, z2.top),
+      stop: z1.bottom !== undefined ? Math.min(z1.bottom, z2.bottom) : z2.bottom,
+    };
+  }
+  if (z1.right !== undefined && z1.right + 1 === z2.left) {
+    adjacentEdge = {
+      position: "right",
+      start: Math.max(z1.top, z2.top),
+      stop: z1.bottom !== undefined ? Math.min(z1.bottom, z2.bottom) : z2.bottom,
+    };
+  }
+  if (z1.top === z2.bottom + 1) {
+    adjacentEdge = {
+      position: "top",
+      start: Math.max(z1.left, z2.left),
+      stop: z1.right !== undefined ? Math.min(z1.right, z2.right) : z2.right,
+    };
+  }
+  if (z1.bottom !== undefined && z1.bottom + 1 === z2.top) {
+    adjacentEdge = {
+      position: "bottom",
+      start: Math.max(z1.left, z2.left),
+      stop: z1.right !== undefined ? Math.min(z1.right, z2.right) : z2.right,
+    };
+  }
+  return adjacentEdge && adjacentEdge.start <= adjacentEdge.stop ? adjacentEdge : undefined;
 }
 
 /**
  * Return true if two zones overlap, false otherwise.
  */
-export function overlap(z1: Zone, z2: Zone): boolean {
-  if (z1.bottom < z2.top || z2.bottom < z1.top) {
+export function overlap(z1: UnboundedZone, z2: UnboundedZone): boolean {
+  if (
+    (z1.bottom !== undefined && z1.bottom < z2.top) ||
+    (z2.bottom !== undefined && z2.bottom < z1.top)
+  ) {
     return false;
   }
-  if (z1.right < z2.left || z2.right < z1.left) {
+  if (
+    (z1.right !== undefined && z1.right < z2.left) ||
+    (z2.right !== undefined && z2.right < z1.left)
+  ) {
     return false;
   }
   return true;
@@ -385,7 +441,7 @@ export function overlap(z1: Zone, z2: Zone): boolean {
 /**
  * Returns true if any two zones in the given list overlap.
  */
-export function hasOverlappingZones(zones: Zone[]): boolean {
+export function hasOverlappingZones(zones: UnboundedZone[]): boolean {
   for (let i = 0; i < zones.length - 1; i++) {
     for (let j = i + 1; j < zones.length; j++) {
       if (overlap(zones[i], zones[j])) {
@@ -776,4 +832,62 @@ export function mergeContiguousZones(zones: Zone[]) {
     }
   }
   return mergedZones;
+}
+
+export function splitIfAdjacent(zone: UnboundedZone, zoneToRemove: Zone): UnboundedZone[] {
+  const adjacentEdge = adjacent(zone, zoneToRemove);
+  if (!adjacentEdge) return [zone];
+  const newZones: UnboundedZone[] = [];
+  switch (adjacentEdge.position) {
+    case "bottom":
+    case "top":
+      newZones.push({
+        top: zone.top,
+        bottom: zone.bottom,
+        left: adjacentEdge.start,
+        right: adjacentEdge.stop,
+      });
+      if (adjacentEdge.start > zone.left) {
+        newZones.push({
+          top: zone.top,
+          bottom: zone.bottom,
+          left: zone.left,
+          right: adjacentEdge.start - 1,
+        });
+      }
+      if (zone.right === undefined || adjacentEdge.stop < zone.right) {
+        newZones.push({
+          top: zone.top,
+          bottom: zone.bottom,
+          left: adjacentEdge.stop + 1,
+          right: zone.right,
+        });
+      }
+      return newZones;
+    case "left":
+    case "right":
+      newZones.push({
+        top: adjacentEdge.start,
+        bottom: adjacentEdge.stop,
+        left: zone.left,
+        right: zone.right,
+      });
+      if (adjacentEdge.start > zone.top) {
+        newZones.push({
+          top: zone.top,
+          bottom: adjacentEdge.start - 1,
+          left: zone.left,
+          right: zone.right,
+        });
+      }
+      if (zone.bottom === undefined || adjacentEdge.stop < zone.bottom) {
+        newZones.push({
+          top: adjacentEdge.stop + 1,
+          bottom: zone.bottom,
+          left: zone.left,
+          right: zone.right,
+        });
+      }
+      return newZones;
+  }
 }
--- a/src/plugins/core/borders.ts
+++ b/src/plugins/core/borders.ts
@@ -1,21 +1,20 @@
-import { DEFAULT_BORDER_DESC } from "../../constants";
+import { PositionMap } from "../../helpers/cells/position_map";
 import {
   deepCopy,
   deepEquals,
   getItemId,
-  groupConsecutive,
-  groupItemIdsByZones,
-  isDefined,
-  iterateItemIdsPositions,
-  range,
+  intersection,
+  positionToZone,
   recomputeZones,
+  removeFalsyAttributes,
   toZone,
 } from "../../helpers/index";
+import { adjacent, overlap, splitIfAdjacent, zoneToXc } from "../../helpers/zones";
 import {
-  AddColumnsRowsCommand,
+  ApplyRangeChange,
   Border,
+  BorderData,
   BorderDescr,
-  BorderPosition,
   CellPosition,
   Color,
   CommandResult,
@@ -23,23 +22,39 @@ import {
   ExcelWorkbookData,
   HeaderIndex,
   SetBorderCommand,
+  SetZoneBordersCommand,
   UID,
+  UnboundedZone,
   WorkbookData,
   Zone,
 } from "../../types/index";
 import { CorePlugin } from "../core_plugin";
 
+export type ZoneBorderData = {
+  top?: BorderDescr;
+  bottom?: BorderDescr;
+  left?: BorderDescr;
+  right?: BorderDescr;
+  vertical?: BorderDescr;
+  horizontal?: BorderDescr;
+};
+
+export type ZoneBorder = {
+  zone: UnboundedZone;
+  style: ZoneBorderData;
+};
+
 interface BordersPluginState {
-  readonly borders: Record<UID, ((Border | undefined)[] | undefined)[] | undefined>;
+  readonly borders: Record<UID, ZoneBorder[] | undefined>;
 }
-/**
- * Formatting plugin.
- *
- * This plugin manages all things related to a cell look:
- * - borders
- */
+
 export class BordersPlugin extends CorePlugin<BordersPluginState> implements BordersPluginState {
-  static getters = ["getCellBorder", "getBordersColors"] as const;
+  static getters = [
+    "getCellBorder",
+    "getBorders",
+    "getBordersColors",
+    "getCellBordersInZone",
+  ] as const;
 
   public readonly borders: BordersPluginState["borders"] = {};
 
@@ -50,7 +65,7 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
   allowDispatch(cmd: CoreCommand) {
     switch (cmd.type) {
       case "SET_BORDER":
-        return this.checkBordersUnchanged(cmd);
+        return this.checkValidations(cmd, this.checkBordersUnchanged, this.ensureHasBorder);
       default:
         return CommandResult.Success;
     }
@@ -64,15 +79,7 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
         }
         break;
       case "DUPLICATE_SHEET":
-        const borders = this.borders[cmd.sheetId];
-        if (borders) {
-          // borders is a sparse 2D array.
-          // map and slice preserve empty values and do not set `undefined` instead
-          const bordersCopy = borders
-            .slice()
-            .map((col) => col?.slice().map((border) => deepCopy(border)));
-          this.history.update("borders", cmd.sheetIdTo, bordersCopy);
-        }
+        this.history.update("borders", cmd.sheetIdTo, deepCopy(this.borders[cmd.sheetId]));
         break;
       case "DELETE_SHEET":
         const allBorders = { ...this.borders };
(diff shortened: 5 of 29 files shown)


src/helpers/zones.ts
```diff
@@ -347,15 +347,15 @@ export function unionUnboundedZones(...zones: UnboundedZone[]): UnboundedZone {
 /**
  * Compute the intersection of two zones. Returns nothing if the two zones don't overlap
  */
-export function intersection(z1: Zone, z2: Zone): Zone | undefined {
+export function intersection(z1: UnboundedZone, z2: Zone): Zone | undefined {
```
pro-odoo, line 358: IMO the two arguments should have the same type, so z2 should also be `UnboundedZone`

src/helpers/zones.ts
```diff
@@ -369,14 +369,58 @@ export function isEqual(z1: Zone, z2: Zone): boolean {
   );
 }
 
+/**
+ * Two zones are adjacent if they -partially- share a border.
+ * Returns the adjacent size of z1 as well as the indexes of the header by which they are adjacent.
+ */
+export function adjacent(z1: UnboundedZone, z2: Zone): Adjacence | undefined {
```
pro-odoo: Same as above

src/helpers/zones.ts
```diff
@@ -369,14 +369,58 @@ export function isEqual(z1: Zone, z2: Zone): boolean {
   );
 }
 
+/**
+ * Two zones are adjacent if they -partially- share a border.
```
pro-odoo: share a edge ?
In the context of spreadsheet, border can be confusing

src/helpers/zones.ts
pro-odoo: Not sure `Adjacence` means something :p
What about something like `AdjacentZoneEdge` ?

src/plugins/core/borders.ts
```diff
@@ -39,23 +38,19 @@ interface BordersPluginState {
...
   // ---------------------------------------------------------------------------
 
-  allowDispatch(cmd: CoreCommand) {
-    switch (cmd.type) {
-      case "SET_BORDER":
-        return this.checkBordersUnchanged(cmd);
-      default:
-        return CommandResult.Success;
-    }
-  }
```
pro-odoo: Why is it removed ?

src/plugins/core/borders.ts
```diff
@@ -64,575 +59,427 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
...
+        this.history.update("borders", cmd.sheetIdTo, deepCopy(this.borders[cmd.sheetId]));
         break;
       case "DELETE_SHEET":
         const allBorders = { ...this.borders };
         delete allBorders[cmd.sheetId];
         this.history.update("borders", allBorders);
         break;
       case "SET_BORDER":
-        this.setBorder(cmd.sheetId, cmd.col, cmd.row, cmd.border);
+        if (!cmd.border) break;
```
pro-odoo: This kind of check should be done in `allowDispatch`

src/plugins/core/borders.ts
```diff
@@ -1,36 +1,35 @@
...
   WorkbookData,
   Zone,
 } from "../../types/index";
 import { CorePlugin } from "../core_plugin";
 
 interface BordersPluginState {
-  readonly borders: Record<UID, ((Border | undefined)[] | undefined)[] | undefined>;
+  readonly borders: Record<UID, NewBorder[] | undefined>;
 }
 /**
```
pro-odoo: Take the opportunity to clean this message

src/plugins/core/borders.ts
```diff
@@ -39,23 +38,19 @@ interface BordersPluginState {
  * - borders
  */
 export class BordersPlugin extends CorePlugin<BordersPluginState> implements BordersPluginState {
-  static getters = ["getCellBorder", "getBordersColors"] as const;
+  static getters = [
+    "getCellBorder",
+    "getBordersColors",
+    "getBorders",
```
pro-odoo, line 54: Seems not used ?

src/plugins/core/borders.ts
```diff
@@ -39,23 +38,19 @@ interface BordersPluginState {
...
+    "getZoneCellBorders",
```
pro-odoo: `getBordersInZone` ?

src/types/border.ts
```diff
@@ -0,0 +1,18 @@
...
+  left?: BorderDescr;
+  right?: BorderDescr;
+  vertical?: BorderDescr;
+  horizontal?: BorderDescr;
+};
+
+export type NewBorder = {
+  zone: UnboundedZone;
+  border: NewBorderData;
+};
```
pro-odoo: Why `New` ? It's confusing with `BorderData` and `Border`

pro-odoo: Seems it's a copy/paste from misc.ts types.
Same for `borderStyles` and `BorderStyle`, ensure to keep only one occurence

src/types/workbook_data.ts
```diff
@@ -74,7 +74,7 @@ export interface WorkbookData {
   sheets: SheetData[];
   styles: { [key: number]: Style };
   formats: { [key: number]: Format };
-  borders: { [key: number]: Border };
+  borders: { [key: number]: NewBorderData };
```
pro-odoo: Double check at the end of the dev if it does not require a migration (for now, no)

tests/borders/border_plugin.test.ts
```diff
@@ -108,21 +108,6 @@ describe("borders", () => {
...
-    expect(setBorders(model, "A1", border)).toBeCancelledBecause(CommandResult.NoChanges);
-  });
-
-  test("reset border when there is no border is cancelled", () => {
-    const model = new Model();
-    expect(setBorders(model, "A1", undefined)).toBeCancelledBecause(CommandResult.NoChanges);
-    expect(setBorders(model, "A1", { top: undefined })).toBeCancelledBecause(
-      CommandResult.NoChanges
-    );
-  });
```
pro-odoo: Could be kept, right ?

tests/borders/border_plugin.test.ts
```diff
@@ -755,37 +740,31 @@ describe("Grid manipulation", () => {
...
+     * ----------------     ----------------
+     * | A1 | B1 | C1 |     | A1 | B1 | C1 |
+     * ----------------     ------    ------
+     * | A2 | B2 | C2 |  => | A2   B2   C2 |
+     * ----------------     ------    ------
+     * | A3 | B3 | C3 |     | A3 | B3 | C3 |
+     * ----------------     ----------------
+     */
+
+    expect(getBorder(model, "A1")).toEqual({ top: b, left: b, right: b });
```
pro-odoo: As discussed, should be top, left, right and bottom. This use case should create 9 zones as they are all differents

pro-odoo (commented):
scroll scroll

src/helpers/zones.ts
```diff
@@ -347,36 +347,80 @@ export function unionUnboundedZones(...zones: UnboundedZone[]): UnboundedZone {
...
+    right: z1.right !== undefined ? Math.min(z1.right, z2.right) : z1.right ?? z2.right,
   };
 }
 
 /**
  * Two zones are equal if they represent the same area, so we clearly cannot use
  * reference equality.
  */
-export function isEqual(z1: Zone, z2: Zone): boolean {
+export function isEqual(z1: UnboundedZone, z2: UnboundedZone): boolean {
```
LucasLefevre, line 374: the implementation is not correct for unbounded zones. Because one could have `{ top: 0, hasHeader: true, ...}` but the other `{ top: 0, hasHeader: false, ...}`

src/helpers/zones.ts
```diff
@@ -347,36 +347,80 @@ export function unionUnboundedZones(...zones: UnboundedZone[]): UnboundedZone {
...
   return (
     z1.left === z2.left && z1.right === z2.right && z1.top === z2.top && z1.bottom === z2.bottom
   );
 }
 
+/**
+ * Two zones are adjacent if they -partially- share an edge.
+ * Returns the adjacent size of z1 as well as the indexes of the header by which they are adjacent.
+ */
+export function adjacent(z1: UnboundedZone, z2: Zone): AdjacentEdge | undefined {
```
LucasLefevre, line 388: worth some unit tests in `zones_helpers.test.ts`

src/helpers/cells/position_map.ts
```diff
@@ -20,6 +20,12 @@ export class PositionMap<T> {
     map[sheetId][col][row] = value;
   }
 
+  setMultiple(values: Iterable<[CellPosition, T]>) {
```
LucasLefevre: to be consistent with `addMany` from `PositionSet`
```suggestion
  setMany(values: Iterable<[CellPosition, T]>) {
```

src/clipboard_handlers/borders_clipboard.ts
```diff
@@ -28,12 +29,16 @@ export class BorderClipboardHandler extends AbstractCellClipboardHandler<
     }
     const { rowsIndexes, columnsIndexes } = data;
     const borders: (Border | null)[][] = [];
+    const bordersPosition = new PositionMap<Border>();
+    data.zones.map((zone) =>
+      bordersPosition.setMultiple(this.getters.getCellBordersInZone(sheetId, zone).entries())
```
LucasLefevre: Don't use a `.map` if you don't need the resulting mapped array. It allocates memory for nothing.
There's `.forEach`, but a good old for-loop is usually more readable IMHO

src/helpers/cells/position_map.ts
LucasLefevre: Actually, it's used only once and used to fill an empty `PositionMap`.
We could remove this and add an optional `iterable` parameter to the constructor to be consistent with the native `Map`
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/Map

src/helpers/zones.ts
```diff
@@ -347,36 +347,80 @@ export function unionUnboundedZones(...zones: UnboundedZone[]): UnboundedZone {
...
+  }
+  return adjacence && adjacence.start <= adjacence.stop ? adjacence : undefined;
+}
+
 /**
  * Return true if two zones overlap, false otherwise.
  */
-export function overlap(z1: Zone, z2: Zone): boolean {
-  if (z1.bottom < z2.top || z2.bottom < z1.top) {
+export function overlap(z1: UnboundedZone, z2: UnboundedZone): boolean {
```
LucasLefevre, line 425: same here, you can add a few unit tests to `describe("overlap", ...`

src/helpers/zones.ts
```diff
@@ -742,3 +786,61 @@ export function mergeContiguousZones(zones: Zone[]) {
   }
   return mergedZones;
 }
+
+export function splitIfAdjacent(zone: UnboundedZone, zoneToRemove: Zone): UnboundedZone[] {
```
LucasLefevre, line 837: unit tests.
They also help understanding what it does

src/helpers/zones.ts
```diff
@@ -347,36 +347,80 @@ export function unionUnboundedZones(...zones: UnboundedZone[]): UnboundedZone {
...
   );
 }
 
+/**
+ * Two zones are adjacent if they -partially- share an edge.
+ * Returns the adjacent size of z1 as well as the indexes of the header by which they are adjacent.
+ */
+export function adjacent(z1: UnboundedZone, z2: Zone): AdjacentEdge | undefined {
+  if (intersection(z1, z2)) return undefined;
+  let adjacence: AdjacentEdge | undefined = undefined;
```
LucasLefevre: adjacence isn't standard English
```suggestion
  let adjacency: AdjacentEdge | undefined = undefined;
```

src/helpers/zones.ts
```diff
@@ -742,3 +786,61 @@ export function mergeContiguousZones(zones: Zone[]) {
...
+      });
+      if (adjacentEdge.start > zone.left) {
+        newZones.push({
+          top: zone.top,
+          bottom: zone.bottom,
+          left: zone.left,
+          right: adjacentEdge.start - 1,
+        });
+      }
+      if (!zone.right || adjacentEdge.stop < zone.right) {
```
LucasLefevre: is `!zone.right` here for when `right` is `0` or `undefined` or both ?
It's not super clear to me what this function is supposed to do TBH

src/plugins/core/borders.ts
```diff
@@ -1,45 +1,62 @@
...
+export type ZoneBorderData = {
+  top?: BorderDescr;
+  bottom?: BorderDescr;
+  left?: BorderDescr;
+  right?: BorderDescr;
+  vertical?: BorderDescr;
+  horizontal?: BorderDescr;
+};
+
+export type ZoneBorder = {
```
LucasLefevre, line 42:
```suggestion
type ZoneBorder = {
```

src/plugins/core/borders.ts
```diff
@@ -50,7 +67,12 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
   allowDispatch(cmd: CoreCommand) {
     switch (cmd.type) {
       case "SET_BORDER":
-        return this.checkBordersUnchanged(cmd);
+        return this.checkValidations(
+          cmd,
+          this.chainValidations(this.checkBordersUnchanged, this.ensureHasBorder)
```
LucasLefevre:
```suggestion
          this.checkBordersUnchanged, this.ensureHasBorder
```

src/plugins/core/borders.ts
```diff
@@ -50,7 +67,12 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
...
+        );
+      case "SET_ZONE_BORDERS":
+        return this.ensureHasBorder(cmd);
```
LucasLefevre: According to the type of `SetZoneBordersCommand`, `cmd.border` cannot be falsy. Is this needed ? :thinking:

src/plugins/core/borders.ts
```diff
@@ -64,551 +86,323 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
...
-    const bordersTopLeft = this.getCellBorder({ sheetId, col: left, row: top });
-    const bordersBottomRight = this.getCellBorder({ sheetId, col: right, row: bottom });
-    this.clearBorders(sheetId, [zone]);
-    if (bordersTopLeft?.top) {
-      this.setBorders(sheetId, [{ ...zone, bottom: top }], "top", bordersTopLeft.top);
+  private borderDataToNewBorderData(border: BorderData): ZoneBorderData {
+    const borderPosition = {};
+    const borderStyle = { color: border.color ?? "#000000", style: border.style ?? "thin" };
+    if (["all", "external", "top"].includes(border.position)) {
+      borderPosition["top"] = { ...borderStyle };
```
LucasLefevre: The `["top"]` syntax bypasses type checking
```suggestion
    const borderPosition: ZoneBorderData = {};
    const borderStyle = { color: border.color ?? "#000000", style: border.style ?? "thin" };
    if (["all", "external", "top"].includes(border.position)) {
      borderPosition.top = { ...borderStyle };
```

src/plugins/core/borders.ts
```diff
@@ -621,18 +415,54 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
...
+  }
+
+  private minimalZone(zone: Zone, border: ZoneBorderData): Zone {
+    if (border.horizontal || border.vertical) {
+      return zone;
+    }
+
+    if (!border.left && !border.right) {
+      if (!border.bottom) {
+        zone.bottom = zone.top;
```
LucasLefevre: do not mutate the input AND return the same object
(if you return something, someone reading `const zone = this.minimalZone(myZone, border)` would probably expect `zone` to be another object, and the original one to be unchanged.

src/plugins/core/borders.ts
```diff
@@ -621,18 +415,54 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
     return CommandResult.Success;
   }
 
+  private ensureHasBorder(cmd: SetBorderCommand | SetZoneBordersCommand) {
+    if (!cmd.border) return CommandResult.NoChanges;
+    return CommandResult.Success;
+  }
+
+  private minimalZone(zone: Zone, border: ZoneBorderData): Zone {
```
LucasLefevre: is this actually useful to minimize the zone ? I removed it and all tests pass :heavy_check_mark:

src/plugins/core/borders.ts
```diff
@@ -1,45 +1,62 @@
...
+  bottom?: BorderDescr;
+  left?: BorderDescr;
+  right?: BorderDescr;
+  vertical?: BorderDescr;
+  horizontal?: BorderDescr;
+};
+
+export type ZoneBorder = {
+  zone: UnboundedZone;
+  style: ZoneBorderData;
```
LucasLefevre, line 44: I don't think `style` is a good word to use here because it can be confused with the style from `cell.style`.
Maybe everything can be flatten ?
```ts
export type ZoneBorder = {
  zone: UnboundedZone;
  top?: BorderDescr;
  bottom?: BorderDescr;
  left?: BorderDescr;
  ...
};
```

src/plugins/core/borders.ts
```diff
@@ -64,551 +86,323 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
...
-          continue;
-        }
-        for (let row = zone.top; row <= zone.bottom; row++) {
-          this.addBorder(sheetId, zone.left, row, { right: border });
-          for (let col = zone.left + 1; col < zone.right; col++) {
-            this.addBorder(sheetId, col, row, { left: border, right: border });
-          }
-          this.addBorder(sheetId, zone.right, row, { left: border });
+      if (plannedBorder) {
+        let border = this.getNewBorderFromZone(inter, plannedBorder).style;
```
LucasLefevre: I'm a bit confused because `plannedBorder` is the new border, but the second parameter of `getNewBorderFromZone` is called `oldBorder`. So you're saying the new border is old ? :thinking:

src/plugins/core/borders.ts
```diff
@@ -64,551 +86,323 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
...
-  // Getters
-  // ---------------------------------------------------------------------------
+  private adaptBorderRange(applyChange: ApplyRangeChange, sheetId: UID) {
+    const newBorders: ZoneBorder[] = [];
+    for (const border of this.borders[sheetId] ?? []) {
+      const change = applyChange(
+        createRange(
+          { zone: border.zone, sheetId, parts: [], prefixSheet: false },
+          this.getters.getSheetSize
+        )
```
LucasLefevre:
```suggestion
        this.getters.getRangeFromZone(sheetId, border.zone)
```

src/plugins/core/borders.ts
```diff
@@ -64,551 +86,323 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
...
-      colLeftOfInsertion = cmd.base - 1;
-      colRightOfInsertion = cmd.base + cmd.quantity;
-    } else {
-      this.shiftBordersHorizontally(cmd.sheetId, cmd.base + 1, cmd.quantity);
-      colLeftOfInsertion = cmd.base;
-      colRightOfInsertion = cmd.base + cmd.quantity + 1;
+  beforeHandle(cmd: CoreCommand): void {
+    if (cmd.type === "REMOVE_COLUMNS_ROWS") {
+      if (cmd.dimension === "ROW") this.onRowRemove(cmd.sheetId, cmd.elements);
+      else this.onColRemove(cmd.sheetId, cmd.elements);
```
LucasLefevre: not very easy to read IMO
```suggestion
      if (cmd.dimension === "ROW") {
        this.onRowRemove(cmd.sheetId, cmd.elements);
      } else {
        this.onColRemove(cmd.sheetId, cmd.elements);
      }
```

src/plugins/core/borders.ts
```diff
@@ -64,551 +86,323 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
...
+      "borders",
+      sheetId,
+      newBorders.filter((border) => !this.borderIsClear(border))
+    );
+  }
 
-  getCellBorder({ sheetId, col, row }: CellPosition): Border | null {
-    const border = this.borders[sheetId]?.[col]?.[row];
-    return border?.top || border?.bottom || border?.left || border?.right ? deepCopy(border) : null;
+  private onRowRemove(sheetId: UID, rowsIndex: HeaderIndex[]) {
```
LucasLefevre, line 151: `onRowRemove` is not tested. Removing the entire method body doesn't break any test.
Same for `onColRemove`

src/plugins/core/borders.ts
```diff
@@ -64,551 +86,323 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
...
+      if (border.zone.bottom !== undefined && rows.has(border.zone.bottom)) {
+        newBorder = newBorder === border ? deepCopy(border) : newBorder;
+        newBorder.style.bottom = border.style.horizontal;
+      }
+      newBorders.push(newBorder);
+    }
+    this.history.update(
+      "borders",
+      sheetId,
+      newBorders.filter((border) => !this.borderIsClear(border))
```
LucasLefevre: is the filter/borderIsClear useful here ?

src/plugins/core/borders.ts
```diff
@@ -64,551 +86,323 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
...
-        const commonSides = this.getCommonSides(aboveBorder, belowBorder);
-        for (const row of targetRows) {
-          this.addBorder(sheetId, col, row, commonSides);
-        }
+  // This is available as a getter for testing purposes
+  getZoneExternalBorders(sheetId: UID, zone: Zone) {
+    const externalBorders = {};
+    for (const border of this.borders[sheetId] ?? []) {
+      if (overlap(border.zone, zone)) {
+        externalBorders["right"] =
```
LucasLefevre:
```suggestion
    const externalBorders: ZoneBorderData = {};
    for (const border of this.borders[sheetId] ?? []) {
      if (overlap(border.zone, zone)) {
        externalBorders.right =
```

src/plugins/core/borders.ts
```diff
@@ -64,551 +86,323 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
...
+  getCellBordersInZone(sheetId: UID, zone: Zone): PositionMap<Border> {
+    const borders = new PositionMap<Border>();
+    for (const border of this.borders[sheetId] ?? []) {
+      const { zone: bzone, style: bborder } = border;
+      const inter = intersection(bzone, zone);
+      if (!inter) continue;
+      for (let col = inter.left; col <= inter.right; col++) {
+        for (let row = inter.top; row <= inter.bottom; row++) {
+          const cell = borders.get({ sheetId, col, row }) ?? {};
+          cell["right"] = (col === bzone.right ? bborder.right : bborder.vertical) ?? cell["right"];
```
LucasLefevre:
```suggestion
          cell.right = (col === bzone.right ? bborder.right : bborder.vertical) ?? cell.right;
```

LucasLefevre: small shorted :wink: 
```suggestion
          cell["right"] ??= (col === bzone.right ? bborder.right : bborder.vertical);
```

src/clipboard_handlers/index.ts
```diff
@@ -26,7 +26,7 @@ clipboardHandlersRegistries.cellHandlers
   .add("cell", CellClipboardHandler)
   .add("sheet", SheetClipboardHandler)
   .add("merge", MergeClipboardHandler)
-  .add("border", BorderClipboardHandler)
   .add("table", TableClipboardHandler)
   .add("conditionalFormat", ConditionalFormatClipboardHandler)
-  .add("references", ReferenceClipboardHandler);
+  .add("references", ReferenceClipboardHandler)
+  .add("border", BorderClipboardHandler);
```
LucasLefevre, line 34: There's a performance issue.
On the large dataset, I selected the entire sheet to set borders everywhere (internal).
Then I copy/paste the first columns (10k cells). I can't copy/paste more because it crashes.
Pasting is super slow and `addBorder` and `borderIsClear` are on top of the chart
image

tests/sheet/sheet_manipulation_plugin.test.ts
```diff
@@ -315,133 +314,14 @@ describe("Columns", () => {
     });
   });
 
-  describe("Correctly update borders", () => {
```
LucasLefevre, line 318: can't we keep those tests ?
They are super annoying and hard to read and understand, but it makes it quite risky to delete them because we don't know what they test (a lot, at least indirectly)

src/helpers/cells/position_map.ts
fdamhaut: There is already an iterable(`entries`) parameters for the constructor but the only usage of `setMultiple` add an array of iterators to the map

src/helpers/zones.ts
fdamhaut: :nerd_face: Hum, achtually it's technically a very old form of adjacency (which i guess come from the french) :nerd_face:.
I'll change it

src/plugins/core/borders.ts
fdamhaut: It's an optimization;
The goal is to have the smallest possible zone to avoid collision when possible

src/plugins/core/borders.ts
fdamhaut: If you have a zone that has top and bottom and you remove both top and bottom row, then it might become clear with leftover zone (the middle).
But it would also be removed later in the adaptBorderRange, so it might be indeed too much

src/plugins/core/borders.ts
fdamhaut: In the very plausible event where we would merge the borders with the style, keeping it encapsulated would make that change much easier. As for the name, border would be good in that future but here we end up with border.border everywhere.
I feel like style if good enough for now but I can change if you feel strongly

src/clipboard_handlers/borders_clipboard.ts
```diff
@@ -1,43 +1,62 @@
...
-    const { rowsIndexes, columnsIndexes } = data;
-    const borders: (Border | null)[][] = [];
-
-    for (const row of rowsIndexes) {
-      const bordersInRow: (Border | null)[] = [];
-      for (const col of columnsIndexes) {
-        const position = { col, row, sheetId };
-        bordersInRow.push(this.getters.getCellBorder(position));
+    const borders: ZoneBorder[] = [];
+    // TODO see if sorting is already done
```
rrahir: 👀

src/helpers/zones.ts
rrahir: a bit of docstring, would be helpful in this case ^^a small description and/or an explicit example

rrahir:
robodoo rebase-ff

src/helpers/zones.ts
```diff
@@ -347,36 +347,84 @@ export function unionUnboundedZones(...zones: UnboundedZone[]): UnboundedZone {
...
+    z1.bottom === z2.bottom &&
+    z1.hasHeader === z2.hasHeader
   );
 }
 
+/**
+ * Two zones are adjacent if they -partially- share an edge.
+ * Returns the adjacent size of z1 as well as the indexes of the header by which they are adjacent.
+ */
+export function adjacent(z1: UnboundedZone, z2: Zone): AdjacentEdge | undefined {
```
rrahir, line 388: can this support both unbounded zones?
I guess it will have to once we have the full suport of infinite viewport

src/plugins/core/borders.ts
rrahir: this ^^ we generally try not to write if statements as 1liners

src/plugins/core/borders.ts
rrahir: Right now, we end up with a call this.borders[sheet][index].style.top.style so its already confusing (n a different way).
even if we were to merge both features (cell style & border style) in the same pluginm i'd assune you would have different internal mappings ?
```typescript
{
  cells: [{
    zone: any,
    style: any,
  }],
  zones: [{
    zone: any,
    style: any,
  }]
}
```

src/plugins/core/borders.ts
```diff
@@ -1,45 +1,62 @@
...
- */
+
 export class BordersPlugin extends CorePlugin<BordersPluginState> implements BordersPluginState {
-  static getters = ["getCellBorder", "getBordersColors"] as const;
+  static getters = [
+    "getCellBorder",
+    "getBorders",
+    "getBordersColors",
+    "getCellBordersInZone",
+    "getZoneExternalBorders",
```
rrahir: if you only need this for the tests ,you can cheat 
```typescript
const borderPlugin = getPlugin(model, BorderPlugin);
borderPlugin["getZoneExternalBorders"]()
 ```

File: src/plugins/core/borders.ts (hunk 31 above)
Comment (rrahir) [reply src/plugins/core/borders.ts:151]: could be worth adding some tests indeed

File: src/plugins/core/borders.ts [hunk 40]
```diff
@@ -64,551 +81,330 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
...
+    const oldZone = border.zone;
+    const equalSide = {
+      top: newZone.top === oldZone.top,
+      bottom: newZone.bottom === oldZone.bottom,
+      left: newZone.left === oldZone.left,
+      right: newZone.right === oldZone.right,
+    };
+    return {
+      zone: newZone,
+      style: removeFalsyAttributes({
```
rrahir: removeFalsyAttributes does not seem necessary and is quite costywhen executed on a lot of zones

src/plugins/core/borders.ts
```diff
@@ -64,551 +81,330 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
...
-            if (zone.top > 0) {
-              this.history.update("borders", sheetId, col, zone.top - 1, "bottom", undefined);
-            }
-            if (zone.bottom < maxRow) {
-              this.history.update("borders", sheetId, col, zone.bottom + 1, "top", undefined);
-            }
+  private removeAndClearAdjacent(sheetId: UID, zone: Zone) {
+    const borders: ZoneBorder[] = [];
+    for (const existingBorder of this.borders[sheetId] ?? []) {
+      for (const updatedBorderZone of recomputeZones([existingBorder.zone], [zone])) {
```
rrahir, line 306: In corner cases,  this becomes tremendously slow comapred to the actual situation.
Test setup to add to demo data:
``` typescript
Class Demo extends Component {
    ...

    topbarMenuRegistry.addChild("boom", ["file"], {
      name: "load borders",
      sequence: 1e6,
      execute: () => {
        const sheetId = this.model.getters.getActiveSheetId();
        const rows = this.model.getters.getNumberRows(sheetId);
        const columns = this.model.getters.getNumberCols(sheetId);
        const zones = []
        for (let row = 0; row < rows; row+=2) {
          for (let col = row%2; col < columns; col+=2) {
            zones.push({ left: col, right: col, top: row, bottom: row });
          }
        }
        this.model.dispatch("SET_ZONE_BORDERS", {
          sheetId,
          target:zones,
          border:{position:"all" }
        })
      },
      icon: "o-spreadsheet-Icon.CLEAR_AND_RELOAD",
    });

    topbarMenuRegistry.addChild("remove", ["file"], {
      name: "remove borders",
      sequence: 1e6,
      execute: () => {
        const sheetId = this.model.getters.getActiveSheetId();
        const rows = this.model.getters.getNumberRows(sheetId);
        const columns = this.model.getters.getNumberCols(sheetId);
        const zones = []
        for (let row = 0; row < rows; row+=2) {
          for (let col = row%2; col < columns; col+=2) {
            zones.push({ left: col, right: col, top: row, bottom: row });
          }
        }
        this.model.dispatch("SET_ZONE_BORDERS", {
          sheetId,
          target:zones,
          border:{position:"clear" }
        })
      },
      icon: "o-spreadsheet-Icon.CLEAR_AND_RELOAD",
    });
```
in a 26 * 100 cells
Call File > Load Borders then Files > Remove borders

In master
load: 14 ms
remove: 9ms

in this branch:
load: 20 ms.   // ok
remove: 462 ms   // not ok :p

rrahir: The loop call to recompute zones is problematic,maybe there is a way to call it only once?

src/plugins/core/borders.ts
```diff
@@ -621,18 +417,54 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
...
+    return zone;
+  }
+
+  /**
+   * Compute the borders to add to the given zone merged.
+   */
+  private addBordersToMerge(sheetId: UID, zone: Zone) {
+    const border = {
+      ...this.getZoneExternalBorders(sheetId, zone),
+      ...removeFalsyAttributes(this.getCellBorder({ sheetId, col: zone.left, row: zone.top })),
```
rrahir, line 453: same herem I don't think we need `removeFalsyAttributes`

tests/borders/border_plugin.test.ts
```diff
@@ -755,37 +778,77 @@ describe("Grid manipulation", () => {
     expect(model.exportData().borders).toEqual({});
   });
 
-  test("Adding a border on a cell removes it on the adjacent cells if it differs", () => {
```
rrahir: After some consideration, I think this feature should stay. If I draw different colors on adjactent edges with the same thickness, the second one drawn will hide the first one, regardless on the order of insertion.

image
 believe it or not, but the bottom border of B2 is actually blue  and was added after the `all` border of B3 ^^
 


 There is also another issue, the behaviour is not consistent:
If I add a new border to an adjacent one, it does not clear the border of the neighbour, but if  i clear the borders then it clears both the border of the selection and the borders of the adjacent cells.

tests/sheet/sheet_manipulation_plugin.test.ts
rrahir: +1

src/helpers/zones.ts
fdamhaut: The bottleneck is intersection, We looked with LUL and could not create a correct typing for the method with two unbounded zone :/ (And it should not be necessary for infinite viewport)

src/plugins/core/borders.ts
fdamhaut: It was only needed in `addBorder` after the (if !force) so i removed it from here and added it there

src/plugins/core/borders.ts
fdamhaut: Oral discussion resulted in =>
 The remove is not really "human-like" as it would require clicking on all the cell to remove the borders instead of removing them in one large zone.
When removing in 1 (or even 20) large zone, the remove is faster than the load method presented here.

src/plugins/core/borders.ts
fdamhaut: `getCellBorder()` returns undefined if there is no border, so if there is no removeFalsyAttributes it would just always override the external borders

src/plugins/core/borders.ts
fdamhaut: Oral discussion: TLDR will move when updating style and no better suggestion yet

tests/sheet/sheet_manipulation_plugin.test.ts
fdamhaut: Exactly why I want to remove them, As far as I can see: these test the border continuity (ex adding a row after a cell with border, between 2 cells with borders,...) but they are so hard to read that you cannot understand without writing the test on a sheet of paper.

This behaviour was already thoroughly tested in border_plugin.test, and the test from here that had no equivalent were moved to there to have everything in the same place

tests/borders/border_plugin.test.ts
fdamhaut: This wasn't changed because i disagree & i believe we should have a conversation about it with more ppl.

src/plugins/ui_stateful/selection.ts
```diff
@@ -589,13 +590,22 @@ export class GridSelectionPlugin extends UIPlugin {
       },
     ];
 
-    for (const Handler of clipboardHandlersRegistries.cellHandlers.getAll()) {
-      const handler = new Handler(this.getters, this.dispatch);
-      const data = handler.copy(getClipboardDataPositions(sheetId, target));
-      if (!data) {
+    const handlers: [string, AbstractCellClipboardHandler<any, any>][] =
```
LucasLefevre: could you write a commit message to explain ?

LucasLefevre: `<any, any>` => `<unknown, unknown>`
With `any`, TS lets you do whatever you want that could crash, with `unknown`, it prevents to do anything without explicit checks.
In this case, the code doesn't need to know what the data actually is, so `unknown` is a perfect fit

src/plugins/ui_stateful/selection.ts
```diff
@@ -589,13 +590,22 @@ export class GridSelectionPlugin extends UIPlugin {
...
+      clipboardHandlersRegistries.cellHandlers.getKeys().map((name) => {
+        const Handler = clipboardHandlersRegistries.cellHandlers.get(name);
+        return [name, new Handler(this.getters, this.dispatch)];
+      });
+    const data: Record<string, any> = {};
```
LucasLefevre: unknown

tests/helpers/zones_helpers.test.ts
```diff
@@ -21,6 +23,272 @@ describe("overlap", () => {
     expect(overlap(z1, z2)).toBe(false);
     expect(overlap(z2, z1)).toBe(false);
   });
+
+  test("1 vertically unbounded zone", () => {
+    const z1 = { top: 0, right: 0, bottom: undefined, left: 0 };
+    const z2 = { top: 3, right: 0, bottom: 5, left: 0 };
```
LucasLefevre: Could you use `const z1 = toUnboundedZone(...);` instead ? It makes the test much more readable

src/plugins/core/borders.ts
```diff
@@ -64,551 +80,366 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
...
+        this.history.update("borders", cmd.sheetIdTo, deepCopy(this.borders[cmd.sheetId]));
         break;
       case "DELETE_SHEET":
         const allBorders = { ...this.borders };
         delete allBorders[cmd.sheetId];
         this.history.update("borders", allBorders);
         break;
       case "SET_BORDER":
-        this.setBorder(cmd.sheetId, cmd.col, cmd.row, cmd.border);
+        this.addBorders(cmd.sheetId, [positionToZone(cmd)], cmd.border!);
```
LucasLefevre: do not use `!`, sooner or later it'll lead to a bug, it always does.
```suggestion
        this.addBorders(cmd.sheetId, [positionToZone(cmd)], cmd.border);
```

src/plugins/core/borders.ts
```diff
@@ -64,551 +80,366 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
...
-      rowAboveInsertion = cmd.base - 1;
-      rowBelowInsertion = cmd.base + cmd.quantity;
-    } else {
-      this.shiftBordersVertically(cmd.sheetId, cmd.base + 1, cmd.quantity);
-      rowAboveInsertion = cmd.base;
-      rowBelowInsertion = cmd.base + cmd.quantity + 1;
+  adaptRanges(applyChange: ApplyRangeChange, sheetId?: UID, sheetName?: string) {
+    const sheetIds = sheetId ? [sheetId] : Object.keys(this.borders);
+    for (const sheetId of sheetIds) {
+      this.adaptBorderRange(applyChange, sheetId);
```
LucasLefevre: this is not needed anymore since https://github.com/odoo/o-spreadsheet/commit/781f1c78b21a6274db7aacb5769357cebbab0ad8

src/plugins/core/borders.ts
```diff
@@ -64,551 +80,366 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
...
-      const aboveBorder = this.getCellBorder({ sheetId, col, row: topRow });
-      const belowBorder = this.getCellBorder({ sheetId, col, row: bottomRow });
-      if (aboveBorder && belowBorder) {
-        const commonSides = this.getCommonSides(aboveBorder, belowBorder);
-        for (const row of targetRows) {
-          this.addBorder(sheetId, col, row, commonSides);
+  getCellBordersInZone(sheetId: UID, zone: Zone): PositionMap<Border> {
+    const borders = new PositionMap<Border>();
+    for (const border of this.borders[sheetId] ?? []) {
+      const { zone: bzone, style: bborder } = border;
```
LucasLefevre: bborder =>  bstyle ?

src/plugins/core/borders.ts
```diff
@@ -64,551 +80,366 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
...
-        const commonSides = this.getCommonSides(leftBorder, rightBorder);
-        for (const col of targetCols) {
-          this.addBorder(sheetId, col, row, commonSides);
-        }
+  getCellBorder(position: CellPosition): Border {
+    return this.getZoneExternalBorders(position.sheetId, positionToZone(position));
+  }
+
+  // This is available as a getter for testing purposes
+  private getZoneExternalBorders(sheetId: UID, zone: Zone): ZoneBorderData {
```
LucasLefevre:
```suggestion
  private getZoneExternalBorders(sheetId: UID, zone: Zone): ZoneBorderData {
```

tests/borders/border_plugin.test.ts
```diff
@@ -772,20 +799,230 @@ describe("Grid manipulation", () => {
...
+    expect(getZoneBorder("E3:F4")).toEqual({ top: b, bottom: b, right: b });
+    expect(getZoneBorder("A5:B6")).toEqual(allSides);
+    expect(getZoneBorder("C5:D6")).toEqual({ bottom: b, left: b, right: b });
+    expect(getZoneBorder("E5:F6")).toEqual(allSides);
+  });
+
+  test("deleting row at the edge of the border", () => {
+    const sheetId = model.getters.getActiveSheetId();
+    const borders = getPlugin(model, BordersPlugin);
+    const getZoneBorder = (zone) => {
```
LucasLefevre:
```suggestion
    const getZoneBorder = (zone: string) => {
```
same below

src/stores/grid_renderer_store.ts
```diff
@@ -848,23 +855,24 @@ export class GridRenderer extends SpreadsheetStore {
     const bottom = visibleRows[visibleRows.length - 1];
     const viewport = { left, right, top, bottom };
     const sheetId = this.getters.getActiveSheetId();
+    const borders = this.getters.getCellBordersInZone(sheetId, zone);
```
LucasLefevre: It no longer takes table borders into account. Why do you stop using `getCellComputedBorder` like it was before ? is it too costly ? `getters.getCellBordersInZone` is used just here

src/types/commands.ts
```diff
@@ -682,6 +685,11 @@ export interface SetZoneBordersCommand extends TargetDependentCommand {
   border: BorderData;
 }
 
+export interface SetZoneBorderDataCommand extends TargetDependentCommand {
+  type: "SET_ZONE_BORDERDATA";
+  border: ZoneBorderData;
```
LucasLefevre: It's not obvious at first sight was this command is and how it's different from `SET_ZONE_BORDERS`. And it's used only once. Can't we re-use `SET_ZONE_BORDERS` ?

Introducing new core commands should be weighted carefully as we'll have to support it ad vitam aeternam (or find a way to upgrade it, which is not always easy or even possible)

src/clipboard_handlers/borders_clipboard.ts
```diff
@@ -1,43 +1,59 @@
...
+      let rowsBefore = 0;
+      for (const rows of groupConsecutive(data.rowsIndexes)) {
+        const zone = {
+          left: cols[0],
+          right: cols[cols.length - 1],
+          top: rows[0],
+          bottom: rows[rows.length - 1],
+        };
+        borders.push(
+          ...this.getters.getBorders(sheetId, zone).map((zb) => {
```
LucasLefevre: I don't know what `zb` is when reading
```suggestion
          ...this.getters.getBorders(sheetId, zone).map(({ style, zone: borderZone}) => {
```

src/plugins/core/borders.ts
```diff
@@ -621,18 +452,33 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
...
   import(data: WorkbookData) {
-    // Borders
     if (Object.keys(data.borders || {}).length) {
       for (const sheet of data.sheets) {
         for (const [position, borderId] of iterateItemIdsPositions(sheet.id, sheet.borders)) {
-          const { sheetId, col, row } = position;
+          const { sheetId } = position;
           const border = data.borders[borderId];
-          this.setBorder(sheetId, col, row, border, false);
+          this.addBorder(sheetId, positionToZone(position), border);
```
LucasLefevre: at import, we don't need to import position by position but by zones. Let's get faster :rocket:

src/types/commands.ts
fdamhaut: "SET_ZONE_BORDERDATA" allow for more granularity as each border can have it's own style but now that you mention it, I could split it into (one best case, 6 worst case) "SET_ZONE_BORDER" to avoid adding another command

rrahir: or extend SET_ZONE_BORDERS to have a wider granularity?

src/plugins/core/borders.ts
```diff
@@ -621,18 +440,31 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
...
   import(data: WorkbookData) {
-    // Borders
     if (Object.keys(data.borders || {}).length) {
       for (const sheet of data.sheets) {
-        for (const [position, borderId] of iterateItemIdsPositions(sheet.id, sheet.borders)) {
-          const { sheetId, col, row } = position;
-          const border = data.borders[borderId];
-          this.setBorder(sheetId, col, row, border, false);
+        for (const [zone, borderId] of iterateItemIdsZone(sheet.borders)) {
+          this.addBorder(sheet.id, zone, data.borders[borderId]);
```
LucasLefevre: I think the helper `iterateItemIdsZone` is useful here. It's a bazooka pour tuer une mouche.
It's easier to understand if the 2 lines are inlined.
```suggestion
        for (const zoneXc in sheet.borders) {
          const borderId = sheet.borders[zoneXc];
          this.addBorder(sheet.id, toZone(zoneXc), data.borders[borderId]);
```

LucasLefevre (approved):
robodoo rebase-ff r+

PR #6760, opened 2025-07-04
[FIX] selection: prevent partial move of table rows with headers

dhrp-odoo:
#### Steps to reproduce

* Create a table with headers
* Select some rows including the header row, but not the whole table
* Try to move the selected rows

#### Current behavior

The move is allowed even when only part of the table (including headers) is selected.
This results in broken table structure or orphaned headers.

#### Desired behavior after PR is merged

Row moves are allowed only if:

* The headers are not included in the selection, **or**
* The entire table (including headers) is selected.

Task: [4862731](https://www.odoo.com/odoo/2328/tasks/4862731)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_stateful/selection.ts
+++ b/src/plugins/ui_stateful/selection.ts
@@ -6,6 +6,7 @@ import {
   deepCopy,
   isEqual,
   positionToZone,
+  range,
   uniqueZones,
   updateSelectionOnDeletion,
   updateSelectionOnInsertion,
@@ -620,7 +621,38 @@ export class GridSelectionPlugin extends UIPlugin {
     if (headers.some((h) => h < 0 || h >= maxHeaderValue)) {
       return CommandResult.InvalidHeaderIndex;
     }
+    if (!isCol && !this.isTableRowMoveAllowed(id, cmd.elements)) {
+      return CommandResult.CannotMoveTableHeader;
+    }
     return CommandResult.Success;
+  }
+
+  private isTableRowMoveAllowed(sheetId: UID, selectedRows: HeaderIndex[]): boolean {
+    const tables = this.getters.getCoreTables(sheetId);
+    if (tables.length === 0) {
+      return true;
+    }
+
+    const selectedRowSet = new Set(selectedRows);
+    return tables.every(({ range: { zone }, config }) => {
+      const { top, bottom } = zone;
+
+      if (config.numberOfHeaders === 0) {
+        return true;
+      }
+
+      const headerRowEnd = top + config.numberOfHeaders - 1;
+
+      // Moving the table is allowed if table header rows are not part of the selection
+      // Or if the entire table (including header) is selected
+      const isHeaderSelected = selectedRows.some((row) => row >= top && row <= headerRowEnd);
+      if (!isHeaderSelected) {
+        return true;
+      }
+
+      const isWholeTableSelected = range(top, bottom + 1).every((r) => selectedRowSet.has(r));
+      return isWholeTableSelected;
+    });
   }
 
   private fallbackToVisibleSheet() {
--- a/src/components/headers_overlay/headers_overlay.ts
+++ b/src/components/headers_overlay/headers_overlay.ts
@@ -23,7 +23,7 @@ import { ContextMenuType } from "../grid/grid";
 import { css, cssPropertiesToCss } from "../helpers/css";
 import { isCtrlKey } from "../helpers/dom_helpers";
 import { dragAndDropBeyondTheViewport, startDnd } from "../helpers/drag_and_drop";
-import { MergeErrorMessage } from "../translations_terms";
+import { MergeErrorMessage, TableHeaderMoveErrorMessage } from "../translations_terms";
 import { ComposerFocusStore } from "./../composer/composer_focus_store";
 
 // -----------------------------------------------------------------------------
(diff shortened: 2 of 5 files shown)


src/plugins/ui_stateful/selection.ts
```diff
@@ -612,9 +612,41 @@ export class GridSelectionPlugin extends UIPlugin {
     if (headers.some((h) => h < 0 || h >= maxHeaderValue)) {
       return CommandResult.InvalidHeaderIndex;
     }
+    if (!isCol && this.isPartiallyMovingTableWithHeader(id, cmd.elements)) {
+      return CommandResult.CannotMovePartialTableWithHeader;
+    }
     return CommandResult.Success;
   }
 
+  private isPartiallyMovingTableWithHeader(sheetId: UID, selectedRows: number[]): boolean {
```
hokolomopo: Not sure about the name. Maybe just `isMovingTableHeader` ? Same for the command result.

src/plugins/ui_stateful/selection.ts
```diff
@@ -612,9 +612,41 @@ export class GridSelectionPlugin extends UIPlugin {
...
+        return false;
+      }
+
+      const headerSelected = selectedRows.some(
+        (row) => row >= headerRowStart && row < headerRowEnd
+      );
+
+      if (!headerSelected) {
+        return false;
+      }
```
hokolomopo: probably not needed ? the loop below is enough no ?

src/plugins/ui_stateful/selection.ts
```diff
@@ -612,9 +612,41 @@ export class GridSelectionPlugin extends UIPlugin {
...
+
+      for (let r = top; r <= bottom; r++) {
+        if (!selectedRowSet.has(r)) {
+          return true;
+        }
+      }
+
+      return false;
```
hokolomopo:
```suggestion
     return range(top, bottom).some(r => !selectedRowSet.has(r));
```
nitpick

src/plugins/ui_stateful/selection.ts
dhrp-odoo: I kept the `headerSelected` check because we want to allow moving rows when the header is not selected.
The loop below only checks for completeness but doesn't distinguish whether the header is selected.
So this check is necessary to enforce: "Header selected but partial rows - not allowed."

src/plugins/ui_stateful/selection.ts
```diff
@@ -612,9 +613,32 @@ export class GridSelectionPlugin extends UIPlugin {
...
+    const selectedRowSet = new Set(selectedRows);
+    return this.getters.getCoreTables(sheetId).some(({ range: { zone }, config }) => {
+      const { top, bottom } = zone;
+
+      if (config.numberOfHeaders === 0) {
+        return false;
+      }
+
+      const headerRowEnd = top + config.numberOfHeaders;
+      const headerSelected = selectedRows.some((row) => row >= top && row < headerRowEnd);
```
LucasLefevre: I don't understand why you can't just return this. All tests pass if I change it like this
```suggestion
      return selectedRows.some((row) => row >= top && row < headerRowEnd);
```

dhrp-odoo: With the code you suggested, we will always reject the command if the table header is in the selected rows.

But with that, it won’t be possible to move a table within the sheet.

IMO, We should allow to move the table if (table header is selected && all the rows of the table are also selected).

LucasLefevre: Thanks for the explanation :+1: 
Could you write a test for this behavior then ?
and also maybe name the condition by assigning it to a well-named variable.
```ts
      const isTablePartiallySelected = range(top, bottom + 1).some((r) => !selectedRowSet.has(r));
      return isTablePartiallySelected;
```
Something like that, although `isTablePartiallySelected` isn't maybe the best name but I don't have better ideas right now

dhrp-odoo: variable name looks correct since we're checking if table rows are partially or fully selected. I’ve added a test for it, please take a look. :)

src/plugins/ui_stateful/selection.ts
```diff
@@ -620,9 +621,32 @@ export class GridSelectionPlugin extends UIPlugin {
     if (headers.some((h) => h < 0 || h >= maxHeaderValue)) {
       return CommandResult.InvalidHeaderIndex;
     }
+    if (!isCol && this.isMovingTableHeader(id, cmd.elements)) {
+      return CommandResult.CannotMoveTableHeader;
+    }
     return CommandResult.Success;
   }
 
+  private isMovingTableHeader(sheetId: UID, selectedRows: number[]): boolean {
```
rrahir:
```suggestion
  private isMovingTableHeader(sheetId: UID, movingRows: HeaderIndex[]): boolean {
```

src/plugins/ui_stateful/selection.ts
```diff
@@ -620,9 +621,32 @@ export class GridSelectionPlugin extends UIPlugin {
...
+    const selectedRowSet = new Set(selectedRows);
+    return this.getters.getCoreTables(sheetId).some(({ range: { zone }, config }) => {
+      const { top, bottom } = zone;
+
+      if (config.numberOfHeaders === 0) {
+        return false;
+      }
+
+      const headerRowEnd = top + config.numberOfHeaders;
```
rrahir:
```suggestion
      const headerRowEnd = top + config.numberOfHeaders - 1;
```
then check <=

src/plugins/ui_stateful/selection.ts
```diff
@@ -620,9 +621,32 @@ export class GridSelectionPlugin extends UIPlugin {
...
+      const isHeaderSelected = selectedRows.some((row) => row >= top && row < headerRowEnd);
+      if (!isHeaderSelected) {
+        return false;
+      }
+
+      const isTablePartiallySelected = range(top, bottom + 1).some((r) => !selectedRowSet.has(r));
```
rrahir: since the spec and commit message hint that we only allow the move *either* if:
- we don't touch the headers 
- we move the whole table as a whole

i would consider writing the code to reflect this formulation and not the opposite (as it is now)o
i think it'd be clearer and easier to name the steps/variables

dhrp-odoo: I’ve updated the code to reflect that logic, please have a look :)

rrahir:
Nice, thanks for the update :) I think it's clearer this way


robodoo r+

PR #6773, opened 2025-07-08
[IMP] Charts: add calendar chart

anhe-odoo:
## Task Description
This task aims to add the possibility to analyze temporal records by grouping the value associated to each date into time periods (e.g. year, date, month, hour of day, ...). This analysis is added as a new chart which could be further adapted if we want to make heat map based on a matrix data.

## Related Task
- Task: 4816317

### DIFF
--- a/src/components/side_panel/chart/building_blocks/general_design/general_design_editor.ts
+++ b/src/components/side_panel/chart/building_blocks/general_design/general_design_editor.ts
@@ -1,23 +1,21 @@
 import { CHART_TITLE_FONT_SIZE } from "@odoo/o-spreadsheet-engine/constants";
 import { SpreadsheetChildEnv } from "@odoo/o-spreadsheet-engine/types/spreadsheet_env";
 import { Component, useState } from "@odoo/owl";
-import { ChartDefinition, Color, DispatchResult, TitleDesign, UID } from "../../../../../types";
+import { ChartDefinition, Color, TitleDesign } from "../../../../../types";
 import { SidePanelCollapsible } from "../../../components/collapsible/side_panel_collapsible";
 import { RadioSelection } from "../../../components/radio_selection/radio_selection";
 import { RoundColorPicker } from "../../../components/round_color_picker/round_color_picker";
 import { Section } from "../../../components/section/section";
+import { ChartSidePanelProps, ChartSidePanelPropsObject } from "../../common";
 import { ChartTitle } from "../chart_title/chart_title";
 
 interface GeneralDesignEditorState {
   activeTool: string;
 }
 
-interface Props {
-  chartId: UID;
-  definition: ChartDefinition;
-  updateChart: (chartId: UID, definition: Partial<ChartDefinition>) => DispatchResult;
-  canUpdateChart: (chartId: UID, definition: Partial<ChartDefinition>) => DispatchResult;
+interface Props extends ChartSidePanelProps<ChartDefinition> {
   defaultChartTitleFontSize?: number;
+  slots?: object;
 }
 
 export class GeneralDesignEditor extends Component<Props, SpreadsheetChildEnv> {
@@ -30,10 +28,7 @@ export class GeneralDesignEditor extends Component<Props, SpreadsheetChildEnv> {
     RadioSelection,
   };
   static props = {
-    chartId: String,
-    definition: Object,
-    updateChart: Function,
-    canUpdateChart: Function,
+    ...ChartSidePanelPropsObject,
     defaultChartTitleFontSize: { type: Number, optional: true },
     slots: { type: Object, optional: true },
   };
--- a/src/components/side_panel/chart/building_blocks/series_design/series_with_axis_design_editor.ts
+++ b/src/components/side_panel/chart/building_blocks/series_design/series_with_axis_design_editor.ts
@@ -7,22 +7,18 @@ import {
   ChartJSRuntime,
   ChartWithDataSetDefinition,
   Color,
-  DispatchResult,
   TrendConfiguration,
-  UID,
 } from "../../../../../types";
 import { NumberInput } from "../../../../number_input/number_input";
 import { Checkbox } from "../../../components/checkbox/checkbox";
 import { RadioSelection } from "../../../components/radio_selection/radio_selection";
 import { RoundColorPicker } from "../../../components/round_color_picker/round_color_picker";
 import { Section } from "../../../components/section/section";
+import { ChartSidePanelProps, ChartSidePanelPropsObject } from "../../common";
 import { SeriesDesignEditor } from "./series_design_editor";
 
-interface Props {
-  chartId: UID;
-  definition: ChartWithDataSetDefinition;
-  canUpdateChart: (chartId: UID, definition: Partial<ChartWithDataSetDefinition>) => DispatchResult;
-  updateChart: (chartId: UID, definition: Partial<ChartWithDataSetDefinition>) => DispatchResult;
+interface Props extends ChartSidePanelProps<ChartWithDataSetDefinition> {
+  slots?: object;
 }
 
 export class SeriesWithAxisDesignEditor extends Component<Props, SpreadsheetChildEnv> {
@@ -36,10 +32,7 @@ export class SeriesWithAxisDesignEditor extends Component<Props, SpreadsheetChil
     NumberInput,
   };
   static props = {
-    chartId: String,
-    definition: Object,
-    canUpdateChart: Function,
-    updateChart: Function,
+    ...ChartSidePanelPropsObject,
     slots: { type: Object, optional: true },
   };
 
--- a/src/components/side_panel/chart/calendar_chart/calendar_chart_config_panel.ts
+++ b/src/components/side_panel/chart/calendar_chart/calendar_chart_config_panel.ts
@@ -8,17 +8,13 @@ import {
 import { createValidRange, isDateTime } from "../../../../helpers";
 import { createDataSets } from "../../../../helpers/figures/charts";
 import { getBarChartData } from "../../../../helpers/figures/charts/runtime";
-import { DEFAULT_LOCALE, DispatchResult, UID } from "../../../../types";
+import { DEFAULT_LOCALE } from "../../../../types";
 import { GenericChartConfigPanel } from "../building_blocks/generic_side_panel/config_panel";
+import { ChartSidePanelProps } from "../common";
 
-interface Props {
-  chartId: UID;
-  definition: CalendarChartDefinition;
-  canUpdateChart: (chartId: UID, definition: Partial<CalendarChartDefinition>) => DispatchResult;
-  updateChart: (chartId: UID, definition: Partial<CalendarChartDefinition>) => DispatchResult;
-}
-
-export class CalendarChartConfigPanel extends GenericChartConfigPanel<Props> {
+export class CalendarChartConfigPanel extends GenericChartConfigPanel<
+  ChartSidePanelProps<CalendarChartDefinition>
+> {
   static template = "o-spreadsheet-CalendarChartConfigPanel";
 
   groupByChoices = CALENDAR_CHART_GRANULARITIES.map((value) => ({
--- a/src/components/side_panel/chart/calendar_chart/calendar_chart_design_panel.ts
+++ b/src/components/side_panel/chart/calendar_chart/calendar_chart_design_panel.ts
@@ -3,7 +3,7 @@ import { LegendPosition } from "@odoo/o-spreadsheet-engine/types/chart";
 import { CalendarChartDefinition } from "@odoo/o-spreadsheet-engine/types/chart/calendar_chart";
 import { SpreadsheetChildEnv } from "@odoo/o-spreadsheet-engine/types/spreadsheet_env";
 import { Component } from "@odoo/owl";
-import { Color, DispatchResult, UID } from "../../../../types/index";
+import { Color } from "../../../../types/index";
 import { SidePanelCollapsible } from "../../components/collapsible/side_panel_collapsible";
 import { RoundColorPicker } from "../../components/round_color_picker/round_color_picker";
 import { Section } from "../../components/section/section";
@@ -14,15 +14,12 @@ import {
 import { ColorScalePicker } from "../building_blocks/color_scale/color_scale_picker";
 import { GeneralDesignEditor } from "../building_blocks/general_design/general_design_editor";
 import { ChartShowValues } from "../building_blocks/show_values/show_values";
+import { ChartSidePanelProps, ChartSidePanelPropsObject } from "../common";
 
-interface Props {
-  chartId: UID;
-  definition: CalendarChartDefinition;
-  canUpdateChart: (chartId: UID, definition: Partial<CalendarChartDefinition>) => DispatchResult;
-  updateChart: (chartId: UID, definition: Partial<CalendarChartDefinition>) => DispatchResult;
-}
-
-export class CalendarChartDesignPanel extends Component<Props, SpreadsheetChildEnv> {
+export class CalendarChartDesignPanel extends Component<
+  ChartSidePanelProps<CalendarChartDefinition>,
+  SpreadsheetChildEnv
+> {
   static template = "o-spreadsheet-CalendarChartDesignPanel";
   static components = {
     GeneralDesignEditor,
@@ -33,12 +30,7 @@ export class CalendarChartDesignPanel extends Component<Props, SpreadsheetChildE
     ColorScalePicker,
     RoundColorPicker,
   };
-  static props = {
-    chartId: String,
-    definition: Object,
-    canUpdateChart: Function,
-    updateChart: Function,
-  };
+  static props = ChartSidePanelPropsObject;
 
   get axesList(): AxisDefinition[] {
     return [
--- a/src/components/side_panel/chart/geo_chart_panel/geo_chart_design_panel.ts
+++ b/src/components/side_panel/chart/geo_chart_panel/geo_chart_design_panel.ts
@@ -1,19 +1,14 @@
-import { Color, UID } from "@odoo/o-spreadsheet-engine";
+import { Color } from "@odoo/o-spreadsheet-engine";
 import { ChartColorScale, LegendPosition } from "@odoo/o-spreadsheet-engine/types/chart";
 import { GeoChartDefinition } from "@odoo/o-spreadsheet-engine/types/chart/geo_chart";
-import { DispatchResult } from "@odoo/o-spreadsheet-engine/types/commands";
 import { RoundColorPicker } from "../../components/round_color_picker/round_color_picker";
 import { ColorScalePicker } from "../building_blocks/color_scale/color_scale_picker";
 import { ChartWithAxisDesignPanel } from "../chart_with_axis/design_panel";
+import { ChartSidePanelProps } from "../common";
 
-interface Props {
-  chartId: UID;
-  definition: GeoChartDefinition;
-  canUpdateChart: (chartId: UID, definition: Partial<GeoChartDefinition>) => DispatchResult;
-  updateChart: (chartId: UID, definition: Partial<GeoChartDefinition>) => DispatchResult;
-}
-
-export class GeoChartDesignPanel extends ChartWithAxisDesignPanel<Props> {
+export class GeoChartDesignPanel extends ChartWithAxisDesignPanel<
+  ChartSidePanelProps<GeoChartDefinition>
+> {
   static template = "o-spreadsheet-GeoChartDesignPanel";
   static components = {
     ...ChartWithAxisDesignPanel.components,
--- a/src/components/side_panel/chart/sunburst_chart/sunburst_chart_design_panel.ts
+++ b/src/components/side_panel/chart/sunburst_chart/sunburst_chart_design_panel.ts
@@ -7,7 +7,6 @@ import {
 import { SpreadsheetChildEnv } from "@odoo/o-spreadsheet-engine/types/spreadsheet_env";
 import { Component } from "@odoo/owl";
 import { deepCopy } from "../../../../helpers";
-import { DispatchResult, UID } from "../../../../types/index";
 import { Checkbox } from "../../components/checkbox/checkbox";
 import { SidePanelCollapsible } from "../../components/collapsible/side_panel_collapsible";
 import { RoundColorPicker } from "../../components/round_color_picker/round_color_picker";
@@ -18,15 +17,12 @@ import { ChartLegend } from "../building_blocks/legend/legend";
 import { PieHoleSize } from "../building_blocks/pie_hole_size/pie_hole_size";
 import { ChartShowValues } from "../building_blocks/show_values/show_values";
 import { TextStyler } from "../building_blocks/text_styler/text_styler";
+import { ChartSidePanelProps, ChartSidePanelPropsObject } from "../common";
 
-interface Props {
-  chartId: UID;
-  definition: SunburstChartDefinition;
-  canUpdateChart: (chartId: UID, definition: Partial<SunburstChartDefinition>) => DispatchResult;
-  updateChart: (chartId: UID, definition: Partial<SunburstChartDefinition>) => DispatchResult;
-}
-
-export class SunburstChartDesignPanel extends Component<Props, SpreadsheetChildEnv> {
+export class SunburstChartDesignPanel extends Component<
+  ChartSidePanelProps<SunburstChartDefinition>,
+  SpreadsheetChildEnv
+> {
   static template = "o-spreadsheet-SunburstChartDesignPanel";
   static components = {
     GeneralDesignEditor,
@@ -40,12 +36,7 @@ export class SunburstChartDesignPanel extends Component<Props, SpreadsheetChildE
     PieHoleSize,
     ChartHumanizeNumbers,
   };
-  static props = {
-    chartId: String,
-    definition: Object,
-    updateChart: Function,
-    canUpdateChart: { type: Function, optional: true },
-  };
+  static props = ChartSidePanelPropsObject;
 
   defaults = SunburstChartDefaults;
 
--- a/src/components/side_panel/chart/building_blocks/generic_side_panel/config_panel.ts
+++ b/src/components/side_panel/chart/building_blocks/generic_side_panel/config_panel.ts
@@ -22,31 +22,23 @@ import {
   CommandResult,
   CustomizedDataSet,
   DispatchResult,
-  UID,
   Zone,
 } from "../../../../../types";
 import { Checkbox } from "../../../components/checkbox/checkbox";
 import { Section } from "../../../components/section/section";
+import { ChartSidePanelProps, ChartSidePanelPropsObject } from "../../common";
 import { ChartDataSeries } from "../data_series/data_series";
 import { ChartErrorSection } from "../error_section/error_section";
 import { ChartLabelRange } from "../label_range/label_range";
-
-interface Props {
-  chartId: UID;
-  definition: ChartWithDataSetDefinition;
-  canUpdateChart: (chartId: UID, definition: Partial<ChartWithDataSetDefinition>) => DispatchResult;
-  updateChart: (chartId: UID, definition: Partial<ChartWithDataSetDefinition>) => DispatchResult;
-}
 
 interface ChartPanelState {
   datasetDispatchResult?: DispatchResult;
   labelsDispatchResult?: DispatchResult;
 }
 
-export class GenericChartConfigPanel<P extends Props = Props> extends Component<
-  P,
-  SpreadsheetChildEnv
-> {
+export class GenericChartConfigPanel<
+  P extends ChartSidePanelProps<ChartWithDataSetDefinition> = ChartSidePanelProps<ChartWithDataSetDefinition>
+> extends Component<P, SpreadsheetChildEnv> {
   static template = "o-spreadsheet-GenericChartConfigPanel";
   static components = {
     ChartDataSeries,
@@ -55,12 +47,7 @@ export class GenericChartConfigPanel<P extends Props = Props> extends Component<
     Checkbox,
     ChartErrorSection,
   };
-  static props = {
-    chartId: String,
-    definition: Object,
-    updateChart: Function,
-    canUpdateChart: Function,
-  };
+  static props = ChartSidePanelPropsObject;
 
   protected state: ChartPanelState = useState({
     datasetDispatchResult: undefined,
--- a/src/components/side_panel/chart/building_blocks/humanize_numbers/humanize_numbers.ts
+++ b/src/components/side_panel/chart/building_blocks/humanize_numbers/humanize_numbers.ts
@@ -2,27 +2,19 @@ import { _t } from "@odoo/o-spreadsheet-engine";
 import { formatLargeNumber, formatValue } from "@odoo/o-spreadsheet-engine/helpers/format/format";
 import { SpreadsheetChildEnv } from "@odoo/o-spreadsheet-engine/types/spreadsheet_env";
 import { Component } from "@odoo/owl";
-import { ChartWithDataSetDefinition, DispatchResult, UID } from "../../../../../types";
+import { ChartWithDataSetDefinition } from "../../../../../types";
 import { Checkbox } from "../../../components/checkbox/checkbox";
+import { ChartSidePanelProps, ChartSidePanelPropsObject } from "../../common";
 
-interface Props {
-  chartId: UID;
-  definition: ChartWithDataSetDefinition;
-  updateChart: (chartId: UID, definition: Partial<ChartWithDataSetDefinition>) => DispatchResult;
-  canUpdateChart: (chartId: UID, definition: Partial<ChartWithDataSetDefinition>) => DispatchResult;
-}
-
-export class ChartHumanizeNumbers extends Component<Props, SpreadsheetChildEnv> {
+export class ChartHumanizeNumbers extends Component<
+  ChartSidePanelProps<ChartWithDataSetDefinition>,
+  SpreadsheetChildEnv
+> {
   static template = "o-spreadsheet-ChartHumanizeNumbers";
   static components = {
     Checkbox,
   };
-  static props = {
-    chartId: String,
-    definition: Object,
-    updateChart: Function,
-    canUpdateChart: Function,
-  };
+  static props = ChartSidePanelPropsObject;
 
   get title() {
     const locale = this.env.model.getters.getLocale();
--- a/src/components/side_panel/chart/building_blocks/legend/legend.ts
+++ b/src/components/side_panel/chart/building_blocks/legend/legend.ts
@@ -1,26 +1,18 @@
 import { SpreadsheetChildEnv } from "@odoo/o-spreadsheet-engine/types/spreadsheet_env";
 import { Component } from "@odoo/owl";
-import { ChartWithDataSetDefinition, DispatchResult, UID } from "../../../../../types";
+import { ChartWithDataSetDefinition } from "../../../../../types";
 import { Section } from "../../../components/section/section";
+import { ChartSidePanelProps, ChartSidePanelPropsObject } from "../../common";
 
-interface Props {
-  chartId: UID;
-  definition: ChartWithDataSetDefinition;
-  updateChart: (chartId: UID, definition: Partial<ChartWithDataSetDefinition>) => DispatchResult;
-  canUpdateChart: (chartId: UID, definition: Partial<ChartWithDataSetDefinition>) => DispatchResult;
-}
-
-export class ChartLegend extends Component<Props, SpreadsheetChildEnv> {
+export class ChartLegend extends Component<
+  ChartSidePanelProps<ChartWithDataSetDefinition>,
+  SpreadsheetChildEnv
+> {
   static template = "o-spreadsheet-ChartLegend";
   static components = {
     Section,
   };
-  static props = {
-    chartId: String,
-    definition: Object,
-    updateChart: Function,
-    canUpdateChart: Function,
-  };
+  static props = ChartSidePanelPropsObject;
 
   updateLegendPosition(ev) {
     this.props.updateChart(this.props.chartId, {
--- a/src/components/side_panel/chart/building_blocks/series_design/series_design_editor.ts
+++ b/src/components/side_panel/chart/building_blocks/series_design/series_design_editor.ts
@@ -2,16 +2,14 @@ import { SpreadsheetChildEnv } from "@odoo/o-spreadsheet-engine/types/spreadshee
 import { Component, useState } from "@odoo/owl";
 import { getColorsPalette, getNthColor, toHex } from "../../../../../helpers";
 import { isTrendLineAxis } from "../../../../../helpers/figures/charts";
-import { ChartWithDataSetDefinition, DispatchResult, UID } from "../../../../../types";
+import { ChartWithDataSetDefinition } from "../../../../../types";
 import { SidePanelCollapsible } from "../../../components/collapsible/side_panel_collapsible";
 import { RoundColorPicker } from "../../../components/round_color_picker/round_color_picker";
 import { Section } from "../../../components/section/section";
+import { ChartSidePanelProps, ChartSidePanelPropsObject } from "../../common";
 
-interface Props {
-  chartId: UID;
-  definition: ChartWithDataSetDefinition;
-  canUpdateChart: (chartId: UID, definition: Partial<ChartWithDataSetDefinition>) => DispatchResult;
-  updateChart: (chartId: UID, definition: Partial<ChartWithDataSetDefinition>) => DispatchResult;
+interface Props extends ChartSidePanelProps<ChartWithDataSetDefinition> {
+  slots?: object;
 }
 
 export class SeriesDesignEditor extends Component<Props, SpreadsheetChildEnv> {
@@ -22,10 +20,7 @@ export class SeriesDesignEditor extends Component<Props, SpreadsheetChildEnv> {
     RoundColorPicker,
   };
   static props = {
-    chartId: String,
-    definition: Object,
-    updateChart: Function,
-    canUpdateChart: Function,
+    ...ChartSidePanelPropsObject,
     slots: { type: Object, optional: true },
   };
 
--- a/src/components/side_panel/chart/building_blocks/show_data_markers/show_data_markers.ts
+++ b/src/components/side_panel/chart/building_blocks/show_data_markers/show_data_markers.ts
@@ -1,24 +1,16 @@
 import { SpreadsheetChildEnv } from "@odoo/o-spreadsheet-engine/types/spreadsheet_env";
 import { Component } from "@odoo/owl";
-import { ChartWithDataSetDefinition, DispatchResult, UID } from "../../../../../types";
+import { ChartWithDataSetDefinition } from "../../../../../types";
 import { Checkbox } from "../../../components/checkbox/checkbox";
+import { ChartSidePanelProps, ChartSidePanelPropsObject } from "../../common";
 
-interface Props {
-  chartId: UID;
-  definition: ChartWithDataSetDefinition;
-  updateChart: (chartId: UID, definition: Partial<ChartWithDataSetDefinition>) => DispatchResult;
-  canUpdateChart: (chartId: UID, definition: Partial<ChartWithDataSetDefinition>) => DispatchResult;
-}
-
-export class ChartShowDataMarkers extends Component<Props, SpreadsheetChildEnv> {
+export class ChartShowDataMarkers extends Component<
+  ChartSidePanelProps<ChartWithDataSetDefinition>,
+  SpreadsheetChildEnv
+> {
   static template = "o-spreadsheet-ChartShowDataMarkers";
   static components = {
     Checkbox,
   };
-  static props = {
-    chartId: String,
-    definition: Object,
-    updateChart: Function,
-    canUpdateChart: Function,
-  };
+  static props = ChartSidePanelPropsObject;
 }
--- a/src/components/side_panel/chart/building_blocks/show_values/show_values.ts
+++ b/src/components/side_panel/chart/building_blocks/show_values/show_values.ts
@@ -1,13 +1,10 @@
 import { SpreadsheetChildEnv } from "@odoo/o-spreadsheet-engine/types/spreadsheet_env";
 import { Component } from "@odoo/owl";
-import { ChartWithDataSetDefinition, DispatchResult, UID } from "../../../../../types";
+import { ChartWithDataSetDefinition } from "../../../../../types";
 import { Checkbox } from "../../../components/checkbox/checkbox";
+import { ChartSidePanelProps, ChartSidePanelPropsObject } from "../../common";
 
-interface Props {
-  chartId: UID;
-  definition: ChartWithDataSetDefinition;
-  updateChart: (chartId: UID, definition: Partial<ChartWithDataSetDefinition>) => DispatchResult;
-  canUpdateChart: (chartId: UID, definition: Partial<ChartWithDataSetDefinition>) => DispatchResult;
+interface Props extends ChartSidePanelProps<ChartWithDataSetDefinition> {
   defaultValue?: boolean;
 }
 
@@ -17,10 +14,7 @@ export class ChartShowValues extends Component<Props, SpreadsheetChildEnv> {
     Checkbox,
   };
   static props = {
-    chartId: String,
-    definition: Object,
-    updateChart: Function,
-    canUpdateChart: Function,
+    ...ChartSidePanelPropsObject,
     defaultValue: { type: Boolean, optional: true },
   };
 }
--- a/src/components/side_panel/chart/chart_with_axis/design_panel.ts
+++ b/src/components/side_panel/chart/chart_with_axis/design_panel.ts
@@ -2,12 +2,7 @@ import { _t } from "@odoo/o-spreadsheet-engine/translation";
 import { SpreadsheetChildEnv } from "@odoo/o-spreadsheet-engine/types/spreadsheet_env";
 import { Component } from "@odoo/owl";
 import { getDefinedAxis } from "../../../../helpers/figures/charts";
-import {
-  ChartWithDataSetDefinition,
-  DispatchResult,
-  GenericDefinition,
-  UID,
-} from "../../../../types/index";
+import { ChartWithDataSetDefinition } from "../../../../types/index";
 import { SidePanelCollapsible } from "../../components/collapsible/side_panel_collapsible";
 import { Section } from "../../components/section/section";
 import {
@@ -19,24 +14,11 @@ import { ChartHumanizeNumbers } from "../building_blocks/humanize_numbers/humani
 import { ChartLegend } from "../building_blocks/legend/legend";
 import { SeriesWithAxisDesignEditor } from "../building_blocks/series_design/series_with_axis_design_editor";
 import { ChartShowValues } from "../building_blocks/show_values/show_values";
+import { ChartSidePanelProps, ChartSidePanelPropsObject } from "../common";
 
-interface Props {
-  chartId: UID;
-  definition: ChartWithDataSetDefinition;
-  canUpdateChart: (
-    chartId: UID,
-    definition: GenericDefinition<ChartWithDataSetDefinition>
-  ) => DispatchResult;
-  updateChart: (
-    chartId: UID,
-    definition: GenericDefinition<ChartWithDataSetDefinition>
-  ) => DispatchResult;
-}
-
-export class ChartWithAxisDesignPanel<P extends Props = Props> extends Component<
-  P,
-  SpreadsheetChildEnv
-> {
+export class ChartWithAxisDesignPanel<
+  P extends ChartSidePanelProps<ChartWithDataSetDefinition>
+> extends Component<P, SpreadsheetChildEnv> {
   static template = "o-spreadsheet-ChartWithAxisDesignPanel";
   static components = {
     GeneralDesignEditor,
@@ -48,12 +30,7 @@ export class ChartWithAxisDesignPanel<P extends Props = Props> extends Component
     ChartShowValues,
     ChartHumanizeNumbers,
   };
-  static props = {
-    chartId: String,
-    definition: Object,
-    canUpdateChart: Function,
-    updateChart: Function,
-  };
+  static props = ChartSidePanelPropsObject;
 
   get axesList(): AxisDefinition[] {
     const { useLeftAxis, useRightAxis } = getDefinedAxis(this.props.definition);
--- a/src/components/side_panel/chart/combo_chart/combo_chart_design_panel.ts
+++ b/src/components/side_panel/chart/combo_chart/combo_chart_design_panel.ts
@@ -1,24 +1,13 @@
 import { _t } from "@odoo/o-spreadsheet-engine/translation";
 import { ComboChartDefinition } from "@odoo/o-spreadsheet-engine/types/chart/combo_chart";
-import { DispatchResult, GenericDefinition, UID } from "../../../../types/index";
 import { RadioSelection } from "../../components/radio_selection/radio_selection";
 import { ChartShowDataMarkers } from "../building_blocks/show_data_markers/show_data_markers";
+import { ChartSidePanelProps } from "../common";
 import { GenericZoomableChartDesignPanel } from "../zoomable_chart/design_panel";
 
-interface Props {
-  chartId: UID;
-  definition: ComboChartDefinition;
-  canUpdateChart: (
-    chartId: UID,
-    definition: GenericDefinition<ComboChartDefinition>
-  ) => DispatchResult;
-  updateChart: (
-    chartId: UID,
-    definition: GenericDefinition<ComboChartDefinition>
-  ) => DispatchResult;
-}
-
-export class ComboChartDesignPanel extends GenericZoomableChartDesignPanel<Props> {
+export class ComboChartDesignPanel extends GenericZoomableChartDesignPanel<
+  ChartSidePanelProps<ComboChartDefinition>
+> {
   static template = "o-spreadsheet-ComboChartDesignPanel";
   static components = {
     ...GenericZoomableChartDesignPanel.components,
--- a/src/components/side_panel/chart/common.ts
+++ b/src/components/side_panel/chart/common.ts
@@ -1,0 +1,15 @@
+import { ChartDefinition, DispatchResult, UID } from "../../..";
+
+export interface ChartSidePanelProps<T extends ChartDefinition> {
+  chartId: UID;
+  definition: T;
+  canUpdateChart: (chartId: UID, definition: Partial<T>) => DispatchResult;
+  updateChart: (chartId: UID, definition: Partial<T>) => DispatchResult;
+}
+
+export const ChartSidePanelPropsObject = {
+  chartId: String,
+  definition: Object,
+  canUpdateChart: Function,
+  updateChart: Function,
+};
--- a/src/components/side_panel/chart/funnel_chart_panel/funnel_chart_design_panel.ts
+++ b/src/components/side_panel/chart/funnel_chart_panel/funnel_chart_design_panel.ts
@@ -4,22 +4,18 @@ import { SpreadsheetChildEnv } from "@odoo/o-spreadsheet-engine/types/spreadshee
 import { Component } from "@odoo/owl";
 import { replaceItemAtIndex } from "../../../../helpers";
 import { getFunnelLabelColors } from "../../../../helpers/figures/charts/runtime";
-import { DispatchResult, UID } from "../../../../types/index";
 import { SidePanelCollapsible } from "../../components/collapsible/side_panel_collapsible";
 import { RoundColorPicker } from "../../components/round_color_picker/round_color_picker";
 import { Section } from "../../components/section/section";
 import { GeneralDesignEditor } from "../building_blocks/general_design/general_design_editor";
 import { ChartHumanizeNumbers } from "../building_blocks/humanize_numbers/humanize_numbers";
 import { ChartShowValues } from "../building_blocks/show_values/show_values";
+import { ChartSidePanelProps } from "../common";
 
-interface Props {
-  chartId: UID;
-  definition: FunnelChartDefinition;
-  canUpdateChart: (chartId: UID, definition: Partial<FunnelChartDefinition>) => DispatchResult;
-  updateChart: (chartId: UID, definition: Partial<FunnelChartDefinition>) => DispatchResult;
-}
-
-export class FunnelChartDesignPanel extends Component<Props, SpreadsheetChildEnv> {
+export class FunnelChartDesignPanel extends Component<
+  ChartSidePanelProps<FunnelChartDefinition>,
+  SpreadsheetChildEnv
+> {
   static template = "o-spreadsheet-FunnelChartDesignPanel";
   static components = {
     ChartShowValues,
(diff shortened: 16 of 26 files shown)


src/helpers/figures/charts/colormap.ts
```diff
@@ -0,0 +1,1013 @@
...
+export type ColorScale = (typeof COLORMAPS)[number];
+
+export abstract class ColorMap {
+  constructor(protected minVal: number = 0, protected maxVal: number = 1) {}
+  abstract getColor(value: number): string;
+}
+
+type ColorsList = [number, number, number][];
+
+export class RainbowColorMap extends ColorMap {
```
LucasLefevre: look at `getColorScale`

tests/test_helpers/commands_helpers.ts
```diff
@@ -318,6 +319,44 @@ export function createRadarChart(
...
+  model: Model,
+  data: Partial<CalendarChartDefinition>,
+  chartId?: UID,
+  sheetId?: UID,
+  figureData: Partial<CreateFigureCommand> = {}
+) {
+  createSheet(model, { sheetId: "calendar", activate: true, rows: 365, cols: 2 });
+  setCellContent(model, "A1", "=DATE(1,1,1) + SEQUENCE(365,1,1,1) + SEQUENCE(365,1, 0, 1/366)");
+  setFormat(model, "A1:A365", "mm/dd/yyyy hh:mm:ss");
+  setCellContent(model, "B1", "=RANDARRAY(365,1)");
```
LucasLefevre: Similarly to data setup in `beforeEach`, tests should not share data, specifically very specific data like this one. Each test should setup its own minimal data to test a behavior.

src/components/figures/chart/chartJs/chart_with_color_scale.ts
```diff
@@ -0,0 +1,75 @@
...
+    const formattedMinValue = humanizeNumber(
+      { value: minValue, format: undefined },
+      this.env.model.getters.getLocale()
+    );
+    const formattedMaxValue = humanizeNumber(
+      { value: maxValue, format: undefined },
+      this.env.model.getters.getLocale()
+    );
+    this.minValue.el!.innerHTML = formattedMinValue;
+    this.maxValue.el!.innerHTML = formattedMaxValue;
```
LucasLefevre: Hello [XSS](https://en.wikipedia.org/wiki/Cross-site_scripting) !
-> `t-esc`

src/components/figures/chart/chartJs/chart_with_color_scale.ts
```diff
@@ -0,0 +1,75 @@
...
+    }
+    return "width: 100%";
+  }
+
+  get colorScaleStyle() {
+    const definition = this.env.model.getters.getChartDefinition(
+      this.props.chartId
+    ) as ChartWithColorScaleDefinition;
+    if (definition.showColorBar) {
+      return "display:flex; margin: 5px; width: 70px; margin-top:15px";
```
LucasLefevre: use `cssPropertiesToCss`

src/components/figures/chart/chartJs/chart_with_color_scale.ts
```diff
@@ -0,0 +1,75 @@
...
+      this.env.model.getters.getLocale()
+    );
+    this.minValue.el!.innerHTML = formattedMinValue;
+    this.maxValue.el!.innerHTML = formattedMaxValue;
+    const definition = this.env.model.getters.getChartDefinition(
+      this.props.chartId
+    ) as ChartWithColorScaleDefinition;
+    const colorScale = definition.colorScale;
+    if (typeof colorScale === "object" && "minColor" in colorScale && "maxColor" in colorScale) {
+      this.colorScale.el!.style.background = `linear-gradient(to top, ${colorScale.minColor}, ${
```
LucasLefevre: don't change DOM elements yourself ! Owl should do it. Otherwise owl could overwrite your changes at anytime since you do not control when owl re-renders and how it updates the DOM.
`t-att-style`

src/components/figures/chart/chartJs/chart_with_color_scale.xml
```diff
@@ -0,0 +1,23 @@
...
+      <div t-att-style="colorScaleStyle">
+        <div
+          style="
+            width: 10px;
+            height: 90%;
+            border: 1px solid;
+          "
+          t-ref="colorScale"
+        />
+        <div style="margin-left: 4px; position: relative; height: 90%">
```
LucasLefevre: don't set style directly on elements -> do it in .scss files or even better: with bootsrap classes

src/components/figures/chart/chartJs/chart_with_color_scale.ts
```diff
@@ -0,0 +1,75 @@
...
+      { value: maxValue, format: undefined },
+      this.env.model.getters.getLocale()
+    );
+    this.minValue.el!.innerHTML = formattedMinValue;
+    this.maxValue.el!.innerHTML = formattedMaxValue;
+    const definition = this.env.model.getters.getChartDefinition(
+      this.props.chartId
+    ) as ChartWithColorScaleDefinition;
+    const colorScale = definition.colorScale;
+    if (typeof colorScale === "object" && "minColor" in colorScale && "maxColor" in colorScale) {
```
LucasLefevre: do it the other way around. It doesn't require to check every key in the object
```suggestion
    if (typeof colorScale === "string" || !colorScale) {
```

src/components/figures/chart/chartJs/chart_with_color_scale.ts
```diff
@@ -0,0 +1,75 @@
...
+
+export class ChartJsComponentWithColorScale extends ChartJsComponent {
+  static template = "o-spreadsheet-ChartJsComponentWithColorScale";
+  private minValue = useRef("minValue");
+  private maxValue = useRef("maxValue");
+  private colorScale = useRef("colorScale");
+
+  protected createChart(chartRuntime: ChartJSRuntime) {
+    super.createChart(chartRuntime);
+    this.updateColorScaleBoundaries(chartRuntime.chartJsConfig);
```
LucasLefevre: this is a code smell. The rendering should be state less as much as possible. Owl is doing the work of syncing the UI with the application state for you.
UI is a function of the state, here `UI = f(chartRuntime)`. Nothing should be stored in this component, everything should be computed.

```js
get boundaries() {
  ...
  return { minValue, maxValue };
}
```

src/components/figures/chart/chartJs/chart_with_color_scale.ts
```diff
@@ -0,0 +1,75 @@
...
+import { humanizeNumber } from "../../../../helpers";
+import { colorScalesCSSBackground } from "../../../../helpers/figures/charts/colormap";
+import { ChartJSRuntime, ChartWithColorScaleDefinition } from "../../../../types";
+import { ChartJsComponent } from "./chartjs";
+
+export class ChartJsComponentWithColorScale extends ChartJsComponent {
+  static template = "o-spreadsheet-ChartJsComponentWithColorScale";
+  private minValue = useRef("minValue");
+  private maxValue = useRef("maxValue");
+  private colorScale = useRef("colorScale");
```
LucasLefevre: `useRef` is also a code smell.
It should be used as the very last solution if owl really cannot do want you want to do, because it's a sign you're modifying the DOM manually, which is owl's responsibility to do it in its own way.
Here, just using basic `t-esc` is enough

src/components/figures/chart/chartJs/chart_with_color_scale.ts
```diff
@@ -0,0 +1,75 @@
+import { useRef } from "@odoo/owl";
+import { ChartConfiguration } from "chart.js/auto";
+import { humanizeNumber } from "../../../../helpers";
+import { colorScalesCSSBackground } from "../../../../helpers/figures/charts/colormap";
+import { ChartJSRuntime, ChartWithColorScaleDefinition } from "../../../../types";
+import { ChartJsComponent } from "./chartjs";
+
+export class ChartJsComponentWithColorScale extends ChartJsComponent {
```
LucasLefevre:
```suggestion
export class ChartJsCalendarComponent extends ChartJsComponent {
```

src/helpers/figures/charts/time_matrix_chart.ts
LucasLefevre, line 1: is it called "time matrix" or "calendar" ? Choose one and use it everywhere

src/helpers/figures/charts/runtime/chartjs_tooltip.ts
```diff
@@ -57,6 +58,32 @@ export function getBarChartTooltip(
...
+    external: customTooltipHandler,
+    callbacks: {
+      title: function (tooltipItems) {
+        return tooltipItems.some((item) => !isTrendLineAxis(item.dataset.xAxisID)) ? undefined : "";
+      },
+      beforeLabel: (tooltipItem) => tooltipItem.dataset?.label || tooltipItem.label,
+      label: function (tooltipItem) {
+        const yLabel = tooltipItem.dataset.values[tooltipItem.dataIndex];
+        if (yLabel === undefined) {
+          return "N./A.";
```
LucasLefevre: or `#N/A` ?
```suggestion
          return new NotAvailableError();
```

src/components/figures/chart/chartJs/chart_with_color_scale.ts
```diff
@@ -0,0 +1,75 @@
...
+    super.createChart(chartRuntime);
+    this.updateColorScaleBoundaries(chartRuntime.chartJsConfig);
+  }
+
+  get heatMapStyle() {
+    const definition = this.env.model.getters.getChartDefinition(
+      this.props.chartId
+    ) as ChartWithColorScaleDefinition;
+    if (definition.showColorBar) {
+      return "width: calc(100% - 70px);";
```
LucasLefevre: you probably don't need that. If the color bar takes 70px, the other parts will naturally have to be pushed away

src/components/figures/chart/chartJs/chart_with_color_scale.ts
LucasLefevre: use bootstrap classes for margins
https://getbootstrap.com/docs/5.3/utilities/spacing/#margin-and-padding

src/components/figures/chart/chartJs/chart_with_color_scale.ts
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  get colorScaleStyle() {
+    const definition = this.env.model.getters.getChartDefinition(
+      this.props.chartId
+    ) as ChartWithColorScaleDefinition;
+    if (definition.showColorBar) {
+      return "display:flex; margin: 5px; width: 70px; margin-top:15px";
+    }
+    return "display: none";
```
LucasLefevre: use `t-if` in the template instead.
With `display: none`, it renders the DOM, computes the style, just to hide the DOM.
It's better if the DOM isn't there in the first place =)

src/components/side_panel/chart/building_blocks/color_scale/color_scale_picker.ts
```diff
@@ -0,0 +1,910 @@
...
+import { isChildEvent } from "../../../../helpers/dom_helpers";
+import { Popover, PopoverProps } from "../../../../popover";
+import { Checkbox } from "../../../components/checkbox/checkbox";
+import { RoundColorPicker } from "../../../components/round_color_picker/round_color_picker";
+import { Section } from "../../../components/section/section";
+
+css/* scss */ `
+  .colormap-container {
+    display: flex;
+    justify-content: right;
```
LucasLefevre: use bootstrap classes as much as possible

src/components/side_panel/chart/building_blocks/color_scale/color_scale_picker.ts
```diff
@@ -0,0 +1,910 @@
...
+  }
+  .blues-colormap {
+    background: linear-gradient(
+      90deg,
+      #f7fbff 0%,
+      #f5fafe 1%,
+      #f3f8fe 2%,
+      #f2f7fd 3%,
+      #eff6fc 4%,
+      #eef5fc 5%,
```
LucasLefevre: reduce all these lines as much as possible and re-use what's in `colormap.ts`

src/components/side_panel/chart/building_blocks/color_scale/color_scale_picker.ts
```diff
@@ -0,0 +1,910 @@
...
+  };
+  static props = {
+    definition: Object,
+    onUpdateColorScale: Function,
+    onShowColorBarChange: { type: Function, optional: true },
+  };
+
+  COLORSCALES = COLORSCALES.map((colormap) => ({
+    value: colormap,
+    label: _t(colormap.charAt(0).toUpperCase() + colormap.slice(1)),
```
LucasLefevre: `_t` only works on static strings!
Translators can't guess what value will be in `colormap` and they can't possibly translate every single string

src/components/side_panel/chart/building_blocks/color_scale/color_scale_picker.xml
```diff
@@ -0,0 +1,73 @@
...
+          <span class="colormap-label">
+            <t t-esc="currentColormapLabel"/>
+          </span>
+          <div class="colormap-preview w-100" t-att-class="currentColormapPreview"/>
+          <span class="ms-1">
+            <t t-call="o-spreadsheet-Icon.CARET_DOWN"/>
+          </span>
+        </div>
+      </div>
+      <Popover t-if="state.popoverProps" t-props="state.popoverProps">
```
LucasLefevre: I believe it doesn't need to be a popover. Maybe it could be `position-absolute` relative to a parent.
Or maybe even better, use the new-ish `popover` Web API https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/popover

src/components/side_panel/chart/building_blocks/color_scale/color_scale_picker.ts
```diff
@@ -0,0 +1,910 @@
...
+    className: `${colormap}-colormap`,
+  }));
+
+  state = useState<ColorScalePickerState>({ popoverProps: undefined, popoverStyle: "" });
```
LucasLefevre, line 40: same here, `state` should be avoided as much as possible. Here the props and style can probably be computing at rendering time.
All you need is a simple boolean to know if the popover is open or not.

src/components/side_panel/chart/building_blocks/color_scale/color_scale_picker.ts
```diff
@@ -0,0 +1,910 @@
...
+    }
+    return currentColormap + "-colormap";
+  }
+
+  get currentColormapLabel(): string {
+    if (typeof this.currentColormap === "object") {
+      return _t("Custom");
+    }
+    const currentColormap = this.currentColormap;
+    return _t(currentColormap.charAt(0).toUpperCase() + currentColormap.slice(1));
```
LucasLefevre: `_t` on dynamic string

src/components/side_panel/chart/building_blocks/color_scale/color_scale_picker.xml
```diff
@@ -0,0 +1,73 @@
...
+        <div t-ref="popoverRef" class="o-chart-select-popover" t-att-style="state.popoverStyle">
+          <div class="w-100">
+            <div
+              class="colormap-container custom-colormap-container"
+              t-on-click="() => this.onColormapChange('custom')">
+              <span class="colormap-label">
+                <t t-esc="'Custom'"/>
```
LucasLefevre: otherwise it's not translated (it's just a javascript string)
```suggestion
                Custom
```

src/components/side_panel/chart/calendar_chart/calendar_chart_config_panel.xml
```diff
@@ -0,0 +1,59 @@
+<templates>
+  <t t-name="o-spreadsheet-CalendarChartConfigPanel">
+    <div>
+      <ChartDataSeries
+        ranges="this.getDataSeriesRanges()"
+        onSelectionChanged.bind="onDataSeriesRangesChanged"
+        onSelectionConfirmed.bind="onDataSeriesConfirmed"
+        onSelectionReordered.bind="onDataSeriesReordered"
+        onSelectionRemoved.bind="onDataSeriesRemoved"
+        maxNumberOfUsedRanges="maxNumberOfUsedRanges"
```
LucasLefevre: You don't need the `maxNumberOfUsedRanges` get function. It's a bazooka pour tuer une mouche and adds more levels or indirection when none is needed
```suggestion
        maxNumberOfUsedRanges="1"
```

src/components/side_panel/chart/calendar_chart/calendar_chart_config_panel.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  updateGroupBy(currentAxis: "horizontal" | "vertical", value: CalendarChartGroupBy) {
+    this.props.updateChart(this.props.chartId, {
+      [`${currentAxis}GroupBy`]: value,
+    });
+  }
+
+  get maxNumberOfUsedRanges(): number {
+    return 1;
+  }
```
LucasLefevre: lol

src/components/side_panel/chart/calendar_chart/calendar_chart_design_panel.ts
```diff
@@ -0,0 +1,76 @@
...
+
+interface Props {
+  chartId: UID;
+  definition: CalendarChartDefinition;
+  canUpdateChart: (chartId: UID, definition: Partial<CalendarChartDefinition>) => DispatchResult;
+  updateChart: (chartId: UID, definition: Partial<CalendarChartDefinition>) => DispatchResult;
+}
+
+export class CalendarChartDesignPanel<P extends Props = Props> extends Component<
+  P,
```
LucasLefevre:
```suggestion
export class CalendarChartDesignPanel extends Component<
  Props,
```

src/components/side_panel/chart/calendar_chart/calendar_chart_design_panel.xml
```diff
@@ -0,0 +1,46 @@
...
+            <option value="right">Right</option>
+            <option value="left">Left</option>
+          </select>
+        </Section>
+        <Section class="'pt-0'" title.translate="Values">
+          <ChartShowValues t-props="props"/>
+        </Section>
+      </t>
+    </GeneralDesignEditor>
+    <SidePanelCollapsible isInitiallyCollapsed="false" title.translate="Calendar chart options">
```
LucasLefevre, line 20: The title `Calendar chart options` isn't very useful for the user. He already known he's configuring options for a calendar chars.
Maybe promote `Color scale` here ?

src/helpers/dates.ts
```diff
@@ -714,3 +714,30 @@ export function isDateStrictlyAfter(date: number, dateAfter: number) {
 export function isDateAfter(date: number, dateAfter: number) {
   return Math.trunc(date) >= Math.trunc(dateAfter);
 }
+
+export function weekNumber(date: DateTime, type: number): number {
```
LucasLefevre: could you change `type` to something else than a number ? It's a horrible API: when you see `weekNumber(date, 12)`, do you know what it does ?
Keep those number things in WEEKNUMBER function. This monstrosity doesn't need to leak else where

src/helpers/figures/charts/colormap.ts
LucasLefevre, line 1: As discussed, reduce the number of points and re-use functions from geo chart when it can be a simple function

src/types/chart/calendar_chart.ts
```diff
@@ -0,0 +1,33 @@
+import { ChartConfiguration } from "chart.js";
+import { ChartColorScale, CommonChartDefinition } from ".";
+import { Color } from "../misc";
+
+export const TIME_MATRIX_GROUP_BY_CHOICES = [
```
LucasLefevre: could you re-use the same granularities as for the pivot ?
Look at `CalendarChartGroupBy`
Also, there's most probably already some code to aggregate the values

src/helpers/figures/charts/runtime/chartjs_dataset.ts
```diff
@@ -101,6 +120,220 @@ export function getBarChartDatasets(
...
+  _t("Sunday"),
+  _t("Monday"),
+  _t("Tuesday"),
+  _t("Wednesday"),
+  _t("Thursday"),
+  _t("Friday"),
+  _t("Saturday"),
+];
+const MONTHS = [
+  _t("January"),
```
LucasLefevre: already defined in `format.js`. Use existing formatting functions :wink:

src/helpers/figures/charts/runtime/chartjs_dataset.ts
```diff
@@ -101,6 +120,220 @@ export function getBarChartDatasets(
   return dataSets;
 }
 
+function getPosition(
```
LucasLefevre: I believe this should be part of `chart_data_extractor.js` which is supposed to prepare the data.
In odoo, for odoo charts, the data is already aggregated.

src/helpers/figures/charts/runtime/chartjs_dataset.ts
```diff
@@ -101,6 +120,220 @@ export function getBarChartDatasets(
...
+  _t("December"),
+];
+
+function getHoursLabel(hour: number): string {
+  return hour < 12 ? `${hour} ${_t("AM")}` : `${hour - 12} ${_t("PM")}`;
+}
+
+function computeValuesAndLabels(
+  timeValues,
+  values,
```
LucasLefevre: types

src/helpers/figures/charts/runtime/chartjs_dataset.ts
```diff
@@ -101,6 +120,220 @@ export function getBarChartDatasets(
...
+  horizontalGroupBy: CalendarChartGroupBy,
+  verticalGroupBy: CalendarChartGroupBy
+) {
+  const grouping: any = {};
```
LucasLefevre: any :eyes:

src/helpers/figures/charts/colormap.ts
```diff
@@ -1,1775 +1,128 @@
...
+      ...acc,
+      [key]: COLORSCHEMES[key].join(", "),
+    }),
+    {}
+  ),
+  rainbow: Array.apply(0, Array(100)).map((_, i) => new RainbowColorMap(0, 99).getColor(i)),
+};
+
+export const COLORSCALES = Object.keys(colorScalesCSSBackground);
+export type ColorScale = (typeof COLORSCALES)[number];
```
LucasLefevre: `export type ColorScale = (keyof typeof COLORSCHEMES) | "greys" | "rainbow";`
+ as const

src/components/figures/chart/chartJs/chart_with_color_scale.ts
anhe-odoo: The idea of this name was to reuse it for general heatmap after :)

src/components/side_panel/chart/calendar_chart/calendar_chart_design_panel.xml
anhe-odoo: This has been chosen to stay consistent with the geo chart ("Geo Chart Options"). But maybe we should use "Design" instead of "Options", as we also have "Waterfall Design" and "Gauge Design" ?

src/helpers/dates.ts
anhe-odoo: Would you find it okay if the second parameter became `startDayOfWeek` and the computation of this from `type` is done in the WEEKNUMBER function ?

LucasLefevre: yes

tests/figures/chart/charts_component.test.ts
```diff
@@ -168,10 +168,10 @@ describe("charts", () => {
...
 
   test.each(CHART_TYPES)("Can open a chart sidePanel", async (chartType) => {
-    await mountSpreadsheet();
-    createTestChart(chartType);
-    await openChartConfigSidePanel(model, env, chartId);
-    expect(fixture.querySelector(".o-figure")).toBeTruthy();
+      await mountSpreadsheet();
+      createTestChart(chartType);
+      await openChartConfigSidePanel(model, env, chartId);
+      expect(fixture.querySelector(".o-figure")).toBeTruthy();
```
pro-odoo: Remove spaces

src/components/side_panel/chart/building_blocks/color_scale/color_scale_picker.ts
```diff
@@ -0,0 +1,188 @@
+import { Component, useExternalListener, useRef, useState } from "@odoo/owl";
```
pro-odoo, line 4: There are not so much tests for this component 😅

src/components/side_panel/chart/building_blocks/color_scale/color_scale_picker.ts
```diff
@@ -0,0 +1,188 @@
...
+    }
+    return undefined;
+  }
+
+  get currentColorScaleLabel(): string {
+    if (typeof this.currentColorScale === "object") {
+      return _t("Custom");
+    }
+    const currentColorScale = this.currentColorScale;
+    return _t(currentColorScale.charAt(0).toUpperCase() + currentColorScale.slice(1));
```
pro-odoo: These terms  are not extracted. Use ChartTerms

src/components/side_panel/chart/building_blocks/color_scale/color_scale_picker.ts
```diff
@@ -86,11 +86,12 @@ export class ColorScalePicker extends Component<Props, SpreadsheetChildEnv> {
...
     definition: Object,
     onUpdateColorScale: Function,
+    onShowColorBarChange: { type: Function, optional: true },
   };
 
-  colorScales = COLORSCALES.map((colorScale) => ({
+  COLORSCALES = COLORSCALES.map((colorScale) => ({
     value: colorScale,
-    label: _t(colorScale.charAt(0).toUpperCase() + colorScale.slice(1)),
+    label: colorScale.charAt(0).toUpperCase() + colorScale.slice(1),
```
pro-odoo: Should use a translated version (ChartTerms)

src/components/side_panel/chart/building_blocks/color_scale/color_scale_picker.ts
```diff
@@ -86,11 +86,12 @@ export class ColorScalePicker extends Component<Props, SpreadsheetChildEnv> {
   static props = {
     definition: Object,
     onUpdateColorScale: Function,
+    onShowColorBarChange: { type: Function, optional: true },
```
pro-odoo: Seems unused

src/components/side_panel/chart/calendar_chart/calendar_chart_config_panel.ts
```diff
@@ -0,0 +1,38 @@
...
+import { Granularity } from "../../../../types";
+import { GenericChartConfigPanel } from "../building_blocks/generic_side_panel/config_panel";
+
+export class CalendarChartConfigPanel extends GenericChartConfigPanel {
+  static template = "o-spreadsheet-CalendarChartConfigPanel";
+
+  groupByChoices = [
+    { value: "year", label: _t("Year") },
+    { value: "month", label: _t("Month") },
+    { value: "day_of_week", label: _t("Weekday") },
```
pro-odoo: `Day of Week`

src/components/side_panel/chart/calendar_chart/calendar_chart_config_panel.ts
```diff
@@ -0,0 +1,38 @@
...
+    { value: "hour_number", label: _t("Hour") },
+    { value: "day_of_month", label: _t("Day of month") },
```
pro-odoo: Day of Month

src/components/side_panel/chart/calendar_chart/calendar_chart_config_panel.ts
```diff
@@ -0,0 +1,38 @@
+import { _t } from "../../../../translation";
+import { Granularity } from "../../../../types";
+import { GenericChartConfigPanel } from "../building_blocks/generic_side_panel/config_panel";
+
+export class CalendarChartConfigPanel extends GenericChartConfigPanel {
+  static template = "o-spreadsheet-CalendarChartConfigPanel";
+
+  groupByChoices = [
+    { value: "year", label: _t("Year") },
+    { value: "month", label: _t("Month") },
```
pro-odoo: Could we use `month_number` instead of `month` ? 
With that, we could have a consistency with pivots, and you could use `ALL_PERIODS` filtered.
WDYT ?

src/components/side_panel/chart/calendar_chart/calendar_chart_design_panel.ts
```diff
@@ -0,0 +1,67 @@
...
+import { ColorScalePicker } from "../building_blocks/color_scale/color_scale_picker";
+import { GeneralDesignEditor } from "../building_blocks/general_design/general_design_editor";
+import { ChartShowValues } from "../building_blocks/show_values/show_values";
+
+interface Props {
+  chartId: UID;
+  definition: CalendarChartDefinition;
+  canUpdateChart: (chartId: UID, definition: Partial<CalendarChartDefinition>) => DispatchResult;
+  updateChart: (chartId: UID, definition: Partial<CalendarChartDefinition>) => DispatchResult;
+}
```
pro-odoo: This block is repeated many many times.
We should create something like this:
```typescript
export interface ChartSidePanelProps<T extends ChartDefinition> {
  chartId: UID;
  definition: T;
  canUpdateChart: (chartId: UID, definition: Partial<T>) => DispatchResult;
  updateChart: (chartId: UID, definition: Partial<T>) => DispatchResult;
}

export const ChartSidePanelPropsObject = {
  chartId: String,
  definition: Object,
  canUpdateChart: Function,
  updateChart: Function,
};
```

And use them everywhere.
I'm not sure of the naming

src/components/side_panel/chart/calendar_chart/calendar_chart_design_panel.ts
```diff
@@ -0,0 +1,67 @@
...
+    ChartShowValues,
+    ColorScalePicker,
+    RoundColorPicker,
+  };
+  static props = {
+    chartId: String,
+    definition: Object,
+    canUpdateChart: Function,
+    updateChart: Function,
+  };
```
pro-odoo: See comment above

src/helpers/dates.ts
```diff
@@ -714,3 +714,22 @@ export function isDateStrictlyAfter(date: number, dateAfter: number) {
 export function isDateAfter(date: number, dateAfter: number) {
   return Math.trunc(date) >= Math.trunc(dateAfter);
 }
+
+export function weekNumber(date: DateTime, startDayOfWeek: number): number {
```
pro-odoo: Is there a reason to extract it ? It seems to only be used in `module_date`...
If there is a good reason, take the opportunity to write some unit test on this function :D

src/helpers/figures/charts/runtime/chart_data_extractor.ts
```diff
@@ -94,6 +97,124 @@ export function getBarChartData(
...
+      return `Q${value}`;
+    }
+    default: {
+      return value.toString();
+    }
+  }
+}
+
+function getHoursLabel(hour: number): string {
+  return hour < 12 ? `${hour} AM` : `${hour - 12} PM`;
```
pro-odoo: Should be translated

src/helpers/figures/charts/runtime/chart_data_extractor.ts
```diff
@@ -94,6 +97,124 @@ export function getBarChartData(
...
+      return DAYS[value - 1].toString();
+    }
+    case "hour_number": {
+      return getHoursLabel(value);
+    }
+    case "month_number": {
+      return MONTHS[value - 1].toString();
+    }
+    case "iso_week_number": {
+      return `W${value < 10 ? "0" : ""}${value}`;
```
pro-odoo: Translate

src/helpers/figures/charts/runtime/chart_data_extractor.ts
```diff
@@ -94,6 +97,124 @@ export function getBarChartData(
...
+    }
+    case "quarter_number": {
+      return `Q${value}`;
```
pro-odoo: Translate

src/helpers/figures/charts/runtime/chart_data_extractor.ts
pro-odoo: And use this form to help translators (btw, :
```typescript
const weekNumber = String(value).padStart(2, "0");
return _t("W%(week_number)s, { week_number: weekNumber});
```

packages/o-spreadsheet-engine/src/types/pivot.ts
```diff
@@ -17,9 +17,7 @@ export type Aggregator =
 
 export type Granularity =
   | "day"
-  | "week"
```
pro-odoo, line 19: Could be worth to make it in another commit, to explain why. I remember we discussed it, but in two weeks I will not remember anymore :D

packages/o-spreadsheet-engine/src/migrations/migration_steps.ts
```diff
@@ -548,6 +548,20 @@ migrationStepRegistry
       }
       return data;
     },
+  })
+  .add("18.5.2", {
```
pro-odoo: Will probably be merged after https://github.com/odoo/o-spreadsheet/pull/6914, so the version should probably be increased.

pro-odoo: And an upgrade script will be required as well

src/components/side_panel/chart/building_blocks/color_scale/color_scale_picker.ts
anhe-odoo: It's tested in `tests/figures/chart/calendar/calendar_chart_panel_component.test.ts` and `tests/figures/geochart/geo_chart_panel_component.test.ts`

src/components/side_panel/chart/building_blocks/color_scale/color_scale_picker.ts
anhe-odoo: I'm not sure "Rainbow" should be translated to "Arc-en-ciel" for example, as it's the name of the colormap. We should keep the names as is (and I have to remove the _t at line 143

anhe-odoo: Nvm, as strange as it is, they are already translated for the geo chart. I'll use them then

src/helpers/figures/charts/runtime/chart_data_extractor.ts
anhe-odoo: 'AM' and 'PM' ?

pro-odoo: Not exactly AM and PM, but the date formatting. Would be worth to check what Luxon does

src/helpers/dates.ts
anhe-odoo: There were a good reason but as I now use the date computation from the pivot, there isn't anymore. I forgot to reput it in the original place

src/components/side_panel/chart/building_blocks/color_scale/color_scale_picker.xml
hokolomopo: But why ? Popover works great. Using position absolute create scrollbars when used at the bottom of the panel, and won't put the menu at the top of the input if there is not enough space.

Idk about HTML popover, maybe it works great ? But shouldn't we stay consistent with what we use literally everywhere else ? Why are you now against Popover ?

src/components/side_panel/chart/geo_chart_panel/geo_chart_design_panel.ts
```diff
@@ -1,9 +1,5 @@
-import { LegendPosition } from "../../../../types/chart";
-import {
-  GeoChartColorScale,
-  GeoChartCustomColorScale,
-  GeoChartDefinition,
-} from "../../../../types/chart/geo_chart";
+import { ChartColorScale, ChartCustomColorScale, LegendPosition } from "../../../../types/chart";
```
hokolomopo: some commits are missing description/Task id

hokolomopo: And the first commit talks about `colormaps`, but is about `colorScales`

src/components/side_panel/chart/building_blocks/color_scale/color_scale_picker.ts
```diff
@@ -0,0 +1,188 @@
...
+  Color,
+  SpreadsheetChildEnv,
+} from "../../../../../types";
+import { css } from "../../../../helpers";
+import { isChildEvent } from "../../../../helpers/dom_helpers";
+import { ChartTerms } from "../../../../translations_terms";
+import { RoundColorPicker } from "../../../components/round_color_picker/round_color_picker";
+import { Section } from "../../../components/section/section";
+
+css/* scss */ `
```
hokolomopo: Can you try making it into a scss file ? We're trying to remove the `css` helper

src/components/figures/chart/chartJs/chart_with_color_scale.ts
```diff
@@ -0,0 +1,82 @@
+import { ChartConfiguration } from "chart.js/auto";
+import { CHART_PADDING } from "../../../../constants";
+import { COLORSCHEMES, humanizeNumber } from "../../../../helpers";
+import { ChartJSRuntime, ChartWithColorScaleDefinition } from "../../../../types";
+import { css, cssPropertiesToCss } from "../../../helpers";
+import { ChartJsComponent } from "./chartjs";
+
+css/* scss */ `
```
hokolomopo: same here, put in scss file

src/components/figures/chart/chartJs/chart_with_color_scale.xml
```diff
@@ -0,0 +1,23 @@
...
+      <div class="w-100 h-100">
+        <canvas class="o-figure-canvas w-100 h-100" t-ref="graphContainer"/>
+      </div>
+      <div t-if="colorScalePosition === 'right'" class="chart-color-scale d-flex h-50">
+        <div class="chart-color-scale-preview h-100" t-att-style="colorScalePreviewStyle"/>
+        <div class="d-flex flex-column chart-color-scale-labels d-flex flex-column h-100">
+          <div class="mb-auto" t-esc="boundaries.max"/>
+          <div t-esc="boundaries.min"/>
+        </div>
+      </div>
```
hokolomopo: I'm sure that you can just put the legend left/right with some css, and that you don't have to repeat everything twice

src/components/figures/chart/chartJs/chart_with_color_scale.ts
```diff
@@ -0,0 +1,82 @@
...
+    margin-left: 4px;
+  }
+  .chart-color-scale-preview {
+    width: 10px;
+    height: 90%;
+    border: 1px solid;
+  }
+`;
+
+export class ChartJsComponentWithColorScale extends ChartJsComponent {
```
hokolomopo: have you tried using a chartJS plugin to display the legend instead of another component ?

src/components/side_panel/chart/building_blocks/color_scale/color_scale_picker.ts
```diff
@@ -89,7 +89,7 @@ export class ColorScalePicker extends Component<Props, SpreadsheetChildEnv> {
     onUpdateColorScale: Function,
   };
 
-  colorScales = COLORSCALES.map((colorScale) => ({
+  COLORSCALES = COLORSCALES.map((colorScale) => ({
```
hokolomopo: image

src/components/side_panel/chart/calendar_chart/calendar_chart_config_panel.ts
```diff
@@ -0,0 +1,41 @@
...
+        name: "dataSetsHaveTitle",
+        label: this.dataSetsHaveTitleLabel,
+        value: this.props.definition.dataSetsHaveTitle,
+        onChange: this.onUpdateDataSetsHaveTitle.bind(this),
+      },
+    ];
+  }
+
+  getGroupByType(currentAxis: "horizontal" | "vertical"): Granularity {
+    return this.props.definition[`${currentAxis}GroupBy`] || "year";
```
hokolomopo: avoid using brackets to get a property if you can, it ignores typing

src/components/side_panel/chart/calendar_chart/calendar_chart_config_panel.ts
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  updateGroupBy(currentAxis: "horizontal" | "vertical", value: Granularity) {
+    this.props.updateChart(this.props.chartId, {
+      [`${currentAxis}GroupBy`]: value,
```
hokolomopo: same here

src/helpers/figures/charts/calendar_chart.ts
```diff
@@ -0,0 +1,220 @@
...
+      sheetId,
+      definition.dataSetsHaveTitle
+    );
+    this.labelRange = createValidRange(getters, sheetId, definition.labelRange);
+    this.background = definition.background;
+    this.showValues = definition.showValues;
+    this.colorScale = definition.colorScale;
+    this.axesDesign = definition.axesDesign;
+    this.horizontalGroupBy = definition.horizontalGroupBy ?? "day_of_week";
+    this.verticalGroupBy = definition.verticalGroupBy ?? "hour_number";
```
hokolomopo: either remove the default, or type this.horizontalGroupBy  as not optional

src/helpers/figures/charts/calendar_chart.ts
```diff
@@ -0,0 +1,220 @@
...
+    return {
+      background: context.background,
+      dataSets: context.range ?? [],
+      dataSetsHaveTitle: context.dataSetsHaveTitle ?? false,
+      title: context.title || { text: "" },
+      type: "calendar",
+      labelRange: context.auxiliaryRange || undefined,
+      showValues: context.showValues,
+      axesDesign: context.axesDesign,
+      legendPosition: context.legendPosition || "top",
```
hokolomopo: top doesn't exist for this chart. And you need to handle the context having top/bottom legend, which you don't support

src/helpers/figures/charts/runtime/chart_data_extractor.ts
```diff
@@ -94,6 +98,122 @@ export function getBarChartData(
...
+      previousYLabels.push(yLabel);
+    }
+    if (!(yLabel in grouping[xLabel])) {
+      grouping[xLabel][yLabel] = 0;
+    }
+    grouping[xLabel][yLabel] += values[i];
+  }
+
+  xLabels = xLabels.sort((a, b) => a - b);
+  yLabels = yLabels.sort((a, b) => a - b);
```
hokolomopo: `sort` transforms the array in place, the assignment is useless

src/helpers/figures/charts/runtime/chart_data_extractor.ts
```diff
@@ -94,6 +98,122 @@ export function getBarChartData(
...
+  let xLabels: number[] = [];
+  let yLabels: number[] = [];
+  const previousYLabels: number[] = [];
+  for (let i = 0; i < timeValues?.length; i++) {
+    const xValue = createDate(
+      { granularity: horizontalGroupBy, type: "date", displayName: "date" },
+      timeValues[i],
+      locale
+    );
+    const xLabel = toNumber(xValue, locale);
```
hokolomopo: The variables are badly named, those aren't labels at all if they are numbers ...

src/helpers/figures/charts/runtime/chartjs_dataset.ts
```diff
@@ -101,6 +103,47 @@ export function getBarChartDatasets(
...
+      label: dataSetValues.label,
+      data: dataSetValues.data.map((v) => 1),
+      backgroundColor: dataSetValues.data.map((v) =>
+        v !== undefined ? colorMap(v) : definition.missingValueColor || COLOR_TRANSPARENT
+      ),
+      borderColor: definition.background || BACKGROUND_CHART_COLOR,
+      borderSkipped: false,
+      borderWidth: 1,
+      barPercentage: 1.0,
+      categoryPercentage: 1.0,
```
hokolomopo: 1.0 ? Why not 1 ?

src/helpers/figures/charts/runtime/chartjs_scales.ts
```diff
@@ -84,6 +84,37 @@ export function getBarChartScales(
...
+      title: getChartAxisTitleRuntime(definition.axesDesign?.y),
+      stacked: true,
+      min: 0,
+      max: yLabels.length,
+      ticks: {
+        stepSize: 0.5,
+        callback: function (label, index, labels) {
+          if (index % 2 === 0) {
+            return undefined;
+          }
```
hokolomopo, line 114: could you add a comment explaining that ? why a stepsize  0.5? And why ignore the tick if the index isn't an even number ? Can't we do a stepSize of 2 then ?

src/helpers/figures/charts/runtime/chartjs_show_values.ts
```diff
@@ -26,6 +28,22 @@ export function getChartShowValues(
...
+  args: ChartRuntimeGenerationArgs
+): ChartShowValuesPluginOptions {
+  const { locale } = args;
+  return {
+    horizontal: false,
+    showValues: "showValues" in definition ? !!definition.showValues : false,
+    background: definition.background,
+    callback: (_value: number | string, dataset: ChartMeta<any>, index) => {
+      const value = dataset._dataset.values[index];
+      return value === undefined ? "" : humanizeNumber({ value }, locale);
```
hokolomopo: missing dataset format

src/helpers/figures/charts/runtime/chartjs_title.ts
```diff
@@ -25,7 +25,12 @@ export function getChartTitle(
     padding: {
       // Disable title top/left/right padding to use the chart padding instead.
       // The legend already has a top padding, so bottom padding is useless for the title there.
-      bottom: definition.legendPosition === "top" ? 0 : CHART_PADDING,
+      bottom:
+        "legendPosition" in definition
+          ? definition.legendPosition === "top"
+            ? 0
+            : CHART_PADDING
+          : 0,
```
hokolomopo: isn't the change useless ? legendPosition is also in the calendar charts right ?

src/helpers/figures/charts/runtime/chartjs_tooltip.ts
```diff
@@ -57,6 +58,32 @@ export function getBarChartTooltip(
...
+    external: customTooltipHandler,
+    callbacks: {
+      title: function (tooltipItems) {
+        return tooltipItems.some((item) => !isTrendLineAxis(item.dataset.xAxisID)) ? undefined : "";
+      },
+      beforeLabel: (tooltipItem) => tooltipItem.dataset?.label || tooltipItem.label,
+      label: function (tooltipItem) {
+        const yLabel = tooltipItem.dataset.values[tooltipItem.dataIndex];
+        if (yLabel === undefined) {
+          return "#N/A";
```
hokolomopo: if we keep N/A, use the constant value instead of hard coding it

anhe-odoo: Which constant value ? `CellErrorType.NotAvailable` ?

hokolomopo: yes

src/components/side_panel/chart/building_blocks/color_scale/color_scale_picker.ts
anhe-odoo: image

(more seriously, it was a REPLACE-ALL mistake)

src/components/side_panel/chart/building_blocks/general_design/general_design_editor.ts
```diff
@@ -1,28 +1,18 @@
...
 }
 
-interface Props {
-  chartId: UID;
-  definition: ChartDefinition;
-  updateChart: (chartId: UID, definition: Partial<ChartDefinition>) => DispatchResult;
-  canUpdateChart: (chartId: UID, definition: Partial<ChartDefinition>) => DispatchResult;
+interface Props extends ChartSidePanelProps<ChartDefinition> {
   defaultChartTitleFontSize?: number;
 }
```
rrahir, line 19: `slots` is missing from the interface

src/components/side_panel/chart/sunburst_chart/sunburst_chart_design_panel.ts
```diff
@@ -37,12 +34,7 @@ export class SunburstChartDesignPanel extends Component<Props, SpreadsheetChildE
     ChartLegend,
     PieHoleSize,
   };
-  static props = {
-    chartId: String,
-    definition: Object,
-    updateChart: Function,
-    canUpdateChart: { type: Function, optional: true },
```
rrahir, line 47: pure curiositym do you know why it was set as optional in the first place? does not seem to match the TS interface

src/components/side_panel/chart/building_blocks/series_design/series_with_axis_design_editor.ts
```diff
@@ -6,25 +6,20 @@ import {
...
-interface Props {
-  chartId: UID;
-  definition: ChartWithDataSetDefinition;
-  canUpdateChart: (chartId: UID, definition: Partial<ChartWithDataSetDefinition>) => DispatchResult;
-  updateChart: (chartId: UID, definition: Partial<ChartWithDataSetDefinition>) => DispatchResult;
-}
-
-export class SeriesWithAxisDesignEditor extends Component<Props, SpreadsheetChildEnv> {
+export class SeriesWithAxisDesignEditor extends Component<
+  ChartSidePanelProps<ChartWithDataSetDefinition>,
```
rrahir: same here, slots is missing

src/components/side_panel/chart/sunburst_chart/sunburst_chart_design_panel.ts
anhe-odoo: Probably because it was not used here but should be added in the props to be able to use `t-props=props` ?

src/components/figures/chart/chartJs/chartjs_show_values_plugin.ts
```diff
@@ -2,10 +2,12 @@ import { ChartMeta, ChartType, Plugin } from "chart.js";
 import { computeTextWidth } from "../../../../helpers";
 import { chartFontColor, isTrendLineAxis } from "../../../../helpers/figures/charts/chart_common";
 import { Color } from "../../../../types";
+import { ChartType as ChartTypeEnum } from "../../../../types/chart";
```
hokolomopo:
```suggestion
import { ChartType as SpreadsheetChartType} from "../../../../types/chart";
```

nitpick 
or rename the other ChartJSChartType or something. There's no enum anywhere 🙃

src/components/figures/chart/chartJs/chartjs.ts
```diff
@@ -6,6 +6,7 @@ import { SpreadsheetChildEnv, UID } from "../../../../types";
 import { ChartJSRuntime } from "../../../../types/chart/chart";
 import { chartJsExtensionRegistry, registerChartJSExtensions } from "./chart_js_extension";
 import { ChartAnimationStore } from "./chartjs_animation_store";
+import { chartColorScalePlugin } from "./chartjs_colorscale_plugin";
```
rrahir: I would add at least some demo data, I picked up the task randomly and ... wel don't know what kind of data I should put in it

src/helpers/figures/charts/runtime/chartjs_scales.ts
```diff
@@ -406,15 +407,20 @@ function getChartAxis(
...
 
-function getRuntimeColorScale(definition: GeoChartDefinition) {
-  if (!definition.colorScale || typeof definition.colorScale === "string") {
-    return definition.colorScale || "oranges";
+function getRuntimeColorScale(colorScale: ChartColorScale, minValue = 0, maxValue = 1) {
+  if (!colorScale || typeof colorScale === "string") {
+    const colorScheme = COLORSCHEMES[colorScale || "oranges"];
+    return getColorScale([
+      { value: minValue, color: colorScheme[0] },
+      { value: (minValue + maxValue) / 2, color: colorScheme[Math.floor(1)] },
```
rrahir: Math.floor(1) ?

src/components/side_panel/chart/building_blocks/color_scale/color_scale_picker.xml
```diff
@@ -0,0 +1,73 @@
...
+            </div>
+            <t t-foreach="colorScales" t-as="colorScale" t-key="colorScale.value">
+              <div
+                class="color-scale-container d-flex justify-content-end"
+                t-on-click="() => this.onColorScaleChange(colorScale.value)"
+                t-att-title="colorScale.label">
+                <span class="color-scale-label">
+                  <t t-esc="colorScale.label"/>
+                </span>
+                <div
```
rrahir: the layout seems a bit weird with a large sidepanel but I gues it's not an real issue
image

src/components/side_panel/chart/building_blocks/color_scale/color_scale_picker.css
```diff
@@ -0,0 +1,29 @@
...
+    height: 20px;
+    width: 70%;
+    border: 1px solid #d8dadd;
+  }
+  .custom-color-scale-container {
+    padding-bottom: 5px;
+    border-bottom: 1px solid #d8dadd;
+  }
+  .custom-color-scale {
+    background: url(data:image/svg+xml,%0A%3Csvg%20xmlns%3D%22http%3A%2F%2Fwww.w3.org%2F2000%2Fsvg%22%20width%3D%2210%22%20height%3D%2210%22%3E%0A%20%20%3Cpath%20fill%3D%22%23d9d9d9%22%20d%3D%22M5%205h5v5H5zH0V0h5%22%2F%3E%0A%3C%2Fsvg%3E%0A);
```
rrahir: this looks kind of exactly like the background image of the round button, actually their encoded versions only differ by their formatting.
image

I think it would make sense to have a global class "checkers-background" that handles this.
Let's be as precise as possible and target the background-image property

src/components/side_panel/chart/building_blocks/color_scale/color_scale_picker.ts
```diff
@@ -0,0 +1,140 @@
...
+
+  setup(): void {
+    useExternalListener(window, "click", this.closePopover);
+  }
+
+  get currentColorScale(): ChartColorScale {
+    return this.props.definition.colorScale || "oranges";
+  }
+
+  get currentColorScalePreview(): string {
```
rrahir: I recently read a post (can't find it after 20 minutes so f-it) that explained that for holding information that are not directly related to the style, it's better to rely on the data-{customProp} attribute. In a nutshell, adding the class semantically implies that you want to style your node a certain way, which is not the case here, evenmore so when we know that this is only meaningful for the tests.

I would either:
- rewrite the component with a store so you can test against the store
- use a t-data-colorscale="custom/color" in the template and look for it inside the test -> expect(target.dataset.colorscale).toBe('something')

we could even go a step further and mark the atribute as "test-colorscale" so we know that it can be ignored outside the tests but that it has a purpose.

I think it's not worth adding a full store just for this as I don't see any other added value

src/components/figures/chart/chartJs/chartjs_colorscale_plugin.ts
```diff
@@ -0,0 +1,79 @@
...
+  afterDatasetsDraw(chart: any, args, options: ChartColorScalePluginOptions) {
+    if (!options.position || options.position === "none" || !options.colorScale.length) {
+      return;
+    }
+    const ctx = chart.ctx as CanvasRenderingContext2D;
+    ctx.save();
+
+    ctx.textAlign = "center";
+    ctx.textBaseline = "middle";
+    ctx.miterLimit = 1; // Avoid sharp artifacts on strokeText
```
rrahir, line 37: 👀 I find this quite interesting. do you think it could be useful outside of the chart drawings?  like for the borders maybe?

src/components/side_panel/chart/building_blocks/color_scale/color_scale_picker.ts
```diff
@@ -0,0 +1,140 @@
...
+  colorScales = COLORSCALES.map((colorScale) => ({
+    value: colorScale,
+    label: ChartTerms.ColorScales[colorScale],
+    className: `${colorScale}-color-scale`,
+  }));
+
+  state = useState<ColorScalePickerState>({ popoverProps: undefined, popoverStyle: "" });
+  popoverRef = useRef("popoverRef");
+
+  setup(): void {
```
rrahir:
```suggestion
  setup() {
```

src/components/side_panel/chart/building_blocks/color_scale/color_scale_picker.ts
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  get currentColorScaleLabel(): string {
+    if (typeof this.currentColorScale === "object") {
+      return _t("Custom");
+    }
+    return ChartTerms.ColorScales[this.currentColorScale];
+  }
+
+  onColorScaleChange(value): void {
```
rrahir:
```suggestion
onColorScaleChange(value:string) {

File: packages/o-spreadsheet-engine/src/types/pivot.ts (hunk 46 above)
Comment (rrahir) [reply packages/o-spreadsheet-engine/src/types/pivot.ts:19]: I discussed it with @pro-odoo , the commit message could be more precise, if im' not mistaken, those granularities are not used for spreadsheet pivots specifically

Maybe the variable ALL_PERIODS should be updated (and typed) accordingly

File: packages/o-spreadsheet-engine/src/constants.ts [hunk 77]
```diff
@@ -40,6 +40,7 @@ export const CHART_PADDING_TOP = 15;
 export const CHART_TITLE_FONT_SIZE = 16;
 export const CHART_AXIS_TITLE_FONT_SIZE = 12;
 export const MASTER_CHART_HEIGHT = 60;
+export const CHART_COLORSCALE_WIDTH = 70;
```
rrahir, line 43: commit msg : This analyzis > This analysis

src/components/figures/chart/chartJs/chartjs_colorscale_plugin.ts
```diff
@@ -0,0 +1,79 @@
...
+  locale: Locale;
+}
+
+declare module "chart.js" {
+  interface PluginOptionsByType<TType extends ChartType> {
+    chartColorScalePlugin?: ChartColorScalePluginOptions;
+  }
+}
+
+/** This is a chartJS plugin that will draw the values of each data next to the point/bar/pie slice */
```
rrahir: > draws the heatmap legend / colorscale?

src/components/figures/chart/chartJs/chartjs_show_values_plugin.ts
```diff
@@ -124,13 +134,64 @@ function drawLineOrBarOrRadarChartValues(
...
+        continue;
+      }
+
+      const point = dataset.data[i];
+      const xPosition = point.x;
+
+      let yPosition = 0;
+      const yZeroLine = yAxisScale.getPixelForValue(0);
+      const distanceFromAxisOrigin = Math.abs(yZeroLine - point.y);
+      const textHeight = 12; // ChartJS default text height
```
rrahir: Would it make sense to use Chart.defaults.font.size then?

src/helpers/figures/charts/calendar_chart.ts
```diff
@@ -0,0 +1,228 @@
...
+    applyChange: RangeAdapter
+  ): BarChartDefinition {
+    return transformChartDefinitionWithDataSetsWithZone(chartSheetId, definition, applyChange);
+  }
+
+  static validateChartDefinition(
+    validator: Validator,
+    definition: BarChartDefinition
+  ): CommandResult | CommandResult[] {
+    return validator.checkValidations(definition, checkDataset, checkLabelRange);
```
rrahir: I would add a check on the types of granularities that are allowed, currently, the check only exists in the sidepanel

packages/o-spreadsheet-engine/src/migrations/migration_steps.ts
rrahir: I was so late to review this taht indeed, the step version needs to be updated to something else (keep in mind taht we could already have migration steps in odoo for versiojn 19.1

src/helpers/figures/charts/runtime/chartjs_dataset.ts
```diff
@@ -104,6 +106,47 @@ export function getBarChartDatasets(
   return dataSets;
 }
 
+export function getCalendarChartDatasetAndLabels(
+  definition: CalendarChartDefinition,
+  args: ChartRuntimeGenerationArgs
+): {
+  datasets: (ChartDataset<"bar" | "line"> & { values: number[] })[];
```
rrahir: why do you need to specify bar or line? it does not seem to be required and in the only place where you use it, it's to generate a barchart, which means you had to add  a ts-ignore to cover for the line type. I don't really see the need

src/components/figures/chart/chartJs/chartjs_show_values_plugin.ts
```diff
@@ -2,13 +2,15 @@ import {
...
+import { ChartType as AllChartType } from "@odoo/o-spreadsheet-engine/types/chart";
 import { ChartMeta, ChartType, Plugin } from "chart.js";
 import { computeTextWidth } from "../../../../helpers";
 import { Color } from "../../../../types";
 
 export interface ChartShowValuesPluginOptions {
+  type: AllChartType;
   showValues: boolean;
-  background?: Color;
+  background?: (value: number | string, dataset: ChartMeta, index: number) => Color | undefined;
```
rrahir:
```suggestion
  background: (value: number | string, dataset: ChartMeta, index: number) => Color | undefined;
```
Since you now provide a default callback in getChartShowValues
the optional chaining should also be removed everywhere in the code

src/components/side_panel/chart/building_blocks/color_scale/color_scale_picker.css
anhe-odoo: Yes, it's the same, I took it from the round color picker. I'll do a global class :)

src/helpers/figures/charts/runtime/chartjs_scales.ts
anhe-odoo: Just to be sure ... -_-'

src/components/side_panel/chart/building_blocks/color_scale/color_scale_picker.ts
anhe-odoo: I have also seen a post about it a few weeks ago, and started to use things like `data-test-id='...'` in another task when only searching for it in the test but wasn't sure it was a good idea. I will change that here then :)

anhe-odoo: In fact, here, we need to have the correct `currentColorScale + "-color-scale"` as a class to have the right color background for the preview :)

src/components/figures/chart/chartJs/chartjs_colorscale_plugin.ts
anhe-odoo: TBH this was suggested by copilot when I was writing these lines, and after a few checks, it seemed to be a good idea so I kept the line as is, but I'm not sure if it can be useful somewhere else. I'll have a look :)

src/components/figures/chart/chartJs/chartjs_show_values_plugin.ts
anhe-odoo: Yes, totally

rrahir:
robodoo r+

rrahir:
robodoo rebase-ff

PR #6780, opened 2025-07-09
[FIX] pivot: various evaluation bugs

LucasLefevre:
## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/helpers/pivot/pivot_presentation.ts
```diff
@@ -66,13 +65,13 @@ export default function (PivotClass: PivotUIConstructor) {
       this.getters = params.getters;
     }
 
-    init(params?: InitPivotParams | undefined): void {
+    markAsDirtyForEvaluation(): void {
```
VincentSchippefilt, line 68: changing init to something else breaks the API, this should only be done in master (and it might need to be reflected on odoo pivots)

src/components/side_panel/pivot/pivot_side_panel/pivot_side_panel_store.ts
```diff
@@ -28,15 +28,6 @@ export class PivotSidePanelStore extends SpreadsheetStore {
...
   }
 
-  handle(cmd: Command) {
-    switch (cmd.type) {
-      case "UPDATE_PIVOT":
-        if (cmd.pivotId === this.pivotId) {
-          this.getters.getPivot(this.pivotId).init();
-        }
-    }
-  }
```
hokolomopo: Was that always useless ? I thought it would be useful if the pivot had no pivot formula in any sheet, but it seems buggy as hell .... If you rename the fields in the grid, the side panel won't update and mark them as invalid if you do'nt have a pivot formula somewhere. Not really related to your task tho

src/helpers/pivot/pivot_presentation.ts
LucasLefevre: `init` is still part of the API, I'm just adding a new method to the interface and override it here instead of `init`

In the interface: https://github.com/odoo/o-spreadsheet/blob/c8b32d7831ca3dd7e18976ff2963d21efd08a628/src/types/pivot_runtime.ts#L18
used here: https://github.com/odoo/o-spreadsheet/blob/c8b32d7831ca3dd7e18976ff2963d21efd08a628/src/helpers/pivot/spreadsheet_pivot/spreadsheet_pivot.ts#L108

src/components/side_panel/pivot/pivot_side_panel/pivot_side_panel_store.ts
LucasLefevre: it wasn't useless before https://github.com/odoo/o-spreadsheet/commit/425d5506b3ab8a17b33c476e2aab406073ddc75e

LucasLefevre: But I'll revert the commit just for the sake of being extra safe in stable

src/helpers/pivot/pivot_presentation.ts
LucasLefevre: also `markAsDirtyForEvaluation` already existed specifically for spreadsheet pivot. I'm just moving it to the common interface

hokolomopo:
@robodoo r+

PR #6781, opened 2025-07-09
[FIX] pivot: don't aggregate calculated missing value

LucasLefevre:
Steps to reproduce:

- create a pivot with two levels of groups in the rows (Let's say group by Country > Customer)
- add a calculated measure
- add the formula =PIVOT.VALUE(1, "calculated", "Country", "IN") where the value of Country is not found anywhere in the data set

=> `Maximum call stack size exceeded`

Task: [4933818](https://www.odoo.com/odoo/2328/tasks/4933818)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/tests/pivots/pivot_calculated_measure.test.ts
+++ b/tests/pivots/pivot_calculated_measure.test.ts
@@ -1033,4 +1033,36 @@ describe("Pivot calculated measure", () => {
     expect(getEvaluatedCell(model, "E2").value).toEqual(40);
     expect(getEvaluatedCell(model, "E3").value).toEqual(60);
   });
+
+  test("formula with a domain not matching any data in the pivot", () => {
+    const grid = {
+      A1: "Customer",
+      B1: "Country",
+      C1: "Price",
+      A2: "Alice",
+      B2: "BE",
+      C2: "10",
+      A3: '=PIVOT.VALUE(1, "calculated", "Country", "BE", "Customer", "Bob")', // Missing Bob in BE
+      A4: '=PIVOT.VALUE(1, "calculated", "Country", "IN", "Customer", "Alice")', // Missing IN
+      A5: '=PIVOT.VALUE(1, "calculated", "Country", "IN", "Customer", "Bob")', // All missing
+      A6: '=PIVOT.VALUE(1, "calculated", "Country", "IN")', // aggregated value
+    };
+    const model = createModelFromGrid(grid);
+    const sheetId = model.getters.getActiveSheetId();
+    addPivot(model, "A1:C2", {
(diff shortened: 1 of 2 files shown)


VincentSchippefilt:
robodoo r+

tests/pivots/pivot_calculated_measure.test.ts
```diff
@@ -1033,4 +1033,36 @@ describe("Pivot calculated measure", () => {
...
+      measures: [
+        {
+          id: "calculated",
+          fieldName: "calculated",
+          aggregator: "sum",
+          computedBy: { formula: "='Country'&'Customer'", sheetId },
+        },
+      ],
+    });
+    expect(getEvaluatedCell(model, "A3").value).toEqual("BE");
```
hokolomopo, line 1063: I don't think this result make much sense, no ? Shouldn't it be empty since BE+Bob does not exist in the pivot ?

VincentSchippefilt:
robodoo r-

VincentSchippefilt:
I killed the staging so we can see about the comment of Adrien :)

tests/pivots/pivot_calculated_measure.test.ts
LucasLefevre: I agree it half-makes sense.
However, we'd need to check wether the domain matches an entry for every calculated measure formula; for example with `getPivotHeaderValueAndFormat(domain)`, which implies a full scan of entries (for spreadsheet pivots). Do you think it's worth it for such a niche case ?

hokolomopo: I'd say if it takes < 5 line of code it can be worth it. Otherwise, whatever

LucasLefevre: Actually, fixing it "the easy way" (with `getPivotHeaderValueAndFormat(domain)` in 2 LoC) breaks a (tested) behavior that might be a valid use case

Here, there's no entry for "Customer=Bob, Category=Drink", even though Bob is part of the dataset. There's a world where you still want to get something from "Bob" or "Drink" with a calculated measure. (and it would be a breaking change anyway)
image

Fixing it differently would require a new method for both spreadsheet and odoo pivot to check if values exists anywhere, no matter the combination in the domain. And that's not worth it IMO

I was also more concerned about the perf overhead it would be for every calculated measure formula

hokolomopo: We can leave it as it then 🤷

hokolomopo:
@robodoo r+

PR #6783, opened 2025-07-10
[FIX] side_panel: replace side panel instead of stacking it

rmbh-odoo:
## Description:

Steps to reproduce:
- Open the global filter side panel (list of all filters)
- Pin the panel
- Click on a specific filter to view/edit it

Current behavior before PR:
- A second side panel was opened instead of replacing the current one
- Clicking 'Cancel' on the new panel had no effect
- Clicking 'Remove' caused a traceback

- The new side panel replaces the current one (main or secondary)
- The previous panel is restored correctly on 'Cancel' or 'Remove'
- If the new panel is already open, we simply close the current one

Task: [4911603](https://www.odoo.com/odoo/2328/tasks/4911603)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/pivot/pivot_measure_display_panel/pivot_measure_display_panel.ts
+++ b/src/components/side_panel/pivot/pivot_measure_display_panel/pivot_measure_display_panel.ts
@@ -54,12 +54,24 @@ export class PivotMeasureDisplayPanel extends Component<Props, SpreadsheetChildE
   }
 
   onSave() {
-    this.env.openSidePanel("PivotSidePanel", { pivotId: this.props.pivotId });
+    this.env.replaceSidePanel(
+      "PivotSidePanel",
+      `pivot_measure_display_${this.props.pivotId}_${this.props.measure.id}`,
+      {
+        pivotId: this.props.pivotId,
+      }
+    );
   }
 
   onCancel() {
     this.store.cancelMeasureDisplayEdition();
-    this.env.openSidePanel("PivotSidePanel", { pivotId: this.props.pivotId });
+    this.env.replaceSidePanel(
+      "PivotSidePanel",
+      `pivot_measure_display_${this.props.pivotId}_${this.props.measure.id}`,
+      {
+        pivotId: this.props.pivotId,
+      }
+    );
   }
 
   get fieldChoices() {
--- a/src/components/side_panel/side_panel/side_panel_store.ts
+++ b/src/components/side_panel/side_panel/side_panel_store.ts
@@ -35,6 +35,7 @@ interface PanelInfo {
 export class SidePanelStore extends SpreadsheetStore {
   mutators = [
     "open",
+    "replace",
     "toggle",
     "close",
     "changePanelSize",
@@ -110,8 +111,7 @@ export class SidePanelStore extends SpreadsheetStore {
       return;
     }
 
-    const mainPanelKey = this.mainPanel ? this.getPanelKey(this.mainPanel) : undefined;
-    if (!this.mainPanel || !this.mainPanel.isPinned || mainPanelKey === state.key) {
+    if (!this.mainPanel || !this.mainPanel.isPinned || this.mainPanelKey === state.key) {
       this._openPanel("mainPanel", newPanelInfo, state);
       return;
     }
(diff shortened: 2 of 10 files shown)


src/components/side_panel/side_panel/side_panel_store.ts
```diff
@@ -99,7 +99,11 @@ export class SidePanelStore extends SpreadsheetStore {
     return undefined;
   }
 
-  open(componentTag: string, initialPanelProps: SidePanelComponentProps = {}) {
+  open(
+    componentTag: string,
+    initialPanelProps: SidePanelComponentProps = {},
+    isCallingFromSubPanel?: boolean
```
dhrp-odoo: nitpick: maybe `isTriggeredBySecondaryPanel` for consistency, since we’ve been using “secondary panel” in the code.

src/components/side_panel/side_panel/side_panel_store.ts
```diff
@@ -110,8 +114,12 @@ export class SidePanelStore extends SpreadsheetStore {
...
     }
 
-    const mainPanelKey = this.mainPanel ? this.getPanelKey(this.mainPanel) : undefined;
-    if (!this.mainPanel || !this.mainPanel.isPinned || mainPanelKey === state.key) {
+    const isSameKey = this.mainPanelKey === state.key || this.secondaryPanelKey === state.key;
+    if (!this.mainPanel || !this.mainPanel.isPinned || isSameKey || isCallingFromSubPanel) {
+      if (isCallingFromSubPanel && this.secondaryPanel && isSameKey) {
+        this.secondaryPanel.initialPanelProps.onCloseSidePanel?.();
+        this.secondaryPanel = undefined;
+      }
```
dhrp-odoo: Maybe we can `return;` the function here? Otherwise it will try to open the main sidepanel, which is already opened?

src/components/side_panel/side_panel/side_panel_store.ts
```diff
@@ -135,6 +143,26 @@ export class SidePanelStore extends SpreadsheetStore {
...
+    const isSecondaryPanel = this.secondaryPanelKey === state.key;
+    if (isMainPanel && this.secondaryPanel) {
+      this.close();
+      return;
+    }
+    if (isSecondaryPanel && this.secondaryPanel) {
+      this.closeMainPanel();
+      this.togglePinPanel();
+      return;
+    }
```
dhrp-odoo: Can you explain when these conditions will be true?

(Also, don’t forget to update the PR message for enterprise since you changed the logic.)

rmbh-odoo: Case 1: if (isMainPanel && this.secondaryPanel)

1. Open Global Filter → Open Date Filter Editor (main panel) → Pin it
3. Open Global Filter again (secondary panel)
4. From the secondary panel, open Text Filter Editor
5. In the main panel, click Cancel/Save -> opens Global Filter panel in main panel
6. Now in the secondary panel, click Cancel/Save → hits this condition → close secondary panel because it's already present in the main panel

Case 2: if (isSecondaryPanel && this.secondaryPanel)

1. Open Global Filter → Open Date Filter Editor (main panel) → Pin it
2. Open Global Filter again (secondary panel)
3. In the main panel, click Cancel/Save → target is already open in the secondary panel (hits this condition)
4. So, we close the main panel and promote the secondary panel as the new main panel

Let me know if we can improve this further 🙂

src/components/side_panel/pivot/pivot_measure_display_panel/pivot_measure_display_panel.ts
```diff
@@ -54,12 +54,20 @@ export class PivotMeasureDisplayPanel extends Component<Props, SpreadsheetChildE
   }
 
   onSave() {
-    this.env.openSidePanel("PivotSidePanel", { pivotId: this.props.pivotId });
+    this.env.openSidePanel(
+      "PivotSidePanel",
+      { pivotId: this.props.pivotId },
+      `pivot_measure_display_${this.props.pivotId}`
+    );
```
hokolomopo: I would create a separate "replaceSidePanel" method, it isn't really clear what the additional argument means when reading this

src/components/side_panel/side_panel/side_panel_store.ts
```diff
@@ -135,6 +143,26 @@ export class SidePanelStore extends SpreadsheetStore {
     this._openPanel("secondaryPanel", newPanelInfo, state);
   }
 
+  private _handleSourcePanel(
```
hokolomopo: might want a docstring/better function name, it's not clear what it does w/o reading the whole code

src/components/side_panel/side_panel/side_panel_store.ts
```diff
@@ -135,6 +136,41 @@ export class SidePanelStore extends SpreadsheetStore {
...
+    currentPanelKey: string,
+    initialPanelProps: SidePanelComponentProps = {}
+  ) {
+    const newPanelInfo = { initialPanelProps, componentTag, size: DEFAULT_SIDE_PANEL_SIZE };
+    const state = this.computeState(newPanelInfo);
+    if (!state.isOpen) {
+      return;
+    }
+    const expandMainPanelIfCollapsed = () => {
+      if (this.mainPanel?.isCollapsed) {
```
rrahir, line 149: why not expending it whatever happens?

rmbh-odoo: I considered the same, but found a case where just expanding isn’t enough.

If the user has resized the panel (without collapsing), we shouldn't override it.
So we check if the panel was resized; if not, apply the default size.
I think the current check is better than checking the panel size directly.
REF : [sourceLink](https://github.com/odoo/o-spreadsheet/blob/saas-18.4/src/components/side_panel/side_panel/side_panel_store.ts#L155)

What do you think 🤔

rrahir: I think i's not that easy to follow what goes on in that panel :p 

Indeed it looks necessary :)

rrahir:
robodoo r+

PR #6807, opened 2025-07-16
[IMP] dashboard: add full screen button to pivot

hokolomopo:
## Description:

Task: [4497373](https://www.odoo.com/odoo/2328/tasks/4497373)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/demo/main.js
+++ b/demo/main.js
@@ -84,7 +84,7 @@ class Demo extends Component {
         stores.resetStores();
         this.leaveCollaborativeSession();
         await fetch(`http://${window.location.hostname}:9090/clear`);
-        await this.initiateConnection({});
+        await this.initiateConnection({ data: {}, enableTransportService: true });
         this.state.key = this.state.key + 1;
       },
       icon: "o-spreadsheet-Icon.CLEAR_AND_RELOAD",
@@ -103,7 +103,14 @@ class Demo extends Component {
       sequence: 12,
       isReadonlyAllowed: true,
       isVisible: () => this.model.config.mode !== "dashboard",
-      execute: () => this.model.updateMode("dashboard"),
+      execute: async () => {
+        const data = this.model.exportData();
+        stores.resetStores();
+        this.leaveCollaborativeSession();
+        await this.initiateConnection({ data, enableTransportService: false });
+        this.model.updateMode("dashboard");
+        this.state.key = this.state.key + 1;
+      },
       icon: "o-spreadsheet-Icon.OPEN_DASHBOARD",
     });
 
@@ -197,7 +204,7 @@ class Demo extends Component {
             const imageSrc = await this.fileStore.upload(file);
             inputFiles[images[i]] = { imageSrc };
           }
-          await this.initiateConnection(inputFiles);
+          await this.initiateConnection({ data: inputFiles, enableTransportService: false });
           stores.resetStores();
           this.state.key = this.state.key + 1;
 
@@ -221,7 +228,7 @@ class Demo extends Component {
       });
     });
 
-    onWillStart(() => this.initiateConnection());
+    onWillStart(async () => this.initiateConnection({ enableTransportService: true }));
 
     onMounted(() => console.log("Mounted: ", Date.now() - start));
     onWillUnmount(this.leaveCollaborativeSession.bind(this));
@@ -235,7 +242,7 @@ class Demo extends Component {
     });
   }
 
-  async initiateConnection(data = undefined) {
+  async createTransportService() {
     this.transportService = new WebsocketTransport();
     try {
       const [history, _] = await Promise.all([
@@ -251,7 +258,22 @@ class Demo extends Component {
       this.transportService = undefined;
       this.stateUpdateMessages = [];
     }
-    this.createModel(data || demoData);
+  }
+
+  /**
+   *
+   * @param {Object} [args.data] - The initial dataset to load into the model. If omitted, a default demo dataset is used.
+   * @param {boolean} [args.enableTransportService] - Whether to enable the collaborative WebSocket transport service.
+   * @returns {Promise<void>}
+   */
+  async initiateConnection(args) {
+    if (args.enableTransportService) {
+      await this.createTransportService();
+    } else {
+      this.transportService = undefined;
+      this.stateUpdateMessages = [];
+    }
+    this.createModel(args.data || demoData);
     // this.createModel(makePivotDataset(10_000));
     // this.createModel(makeLargeDataset(26, 10_000, ["numbers"]));
     // this.createModel(makeLargeDataset(26, 10_000, ["formulas"]));
--- a/src/components/dashboard/dashboard.ts
+++ b/src/components/dashboard/dashboard.ts
@@ -1,4 +1,5 @@
 import { Component, toRaw, useChildSubEnv, useRef } from "@odoo/owl";
+import { SCROLLBAR_WIDTH } from "../../constants";
 import { Store, useStore } from "../../store_engine";
 import {
   DOMCoordinates,
@@ -32,7 +33,7 @@ css/* scss */ `
 
 export class SpreadsheetDashboard extends Component<Props, SpreadsheetChildEnv> {
   static template = "o-spreadsheet-SpreadsheetDashboard";
-  static props = { getGridSize: Function };
+  static props = {};
   static components = {
     GridOverlay,
     GridPopover,
@@ -49,6 +50,7 @@ export class SpreadsheetDashboard extends Component<Props, SpreadsheetChildEnv> 
   clickableCellsStore!: Store<ClickableCellsStore>;
 
   private gridRef!: Ref<HTMLElement>;
+  private dashboardRef = useRef("dashboard");
 
   setup() {
     this.gridRef = useRef("grid");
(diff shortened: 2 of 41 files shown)


src/components/full_screen_sheet/full_screen_sheet_store.ts
```diff
@@ -0,0 +1,109 @@
...
+      this.originalSheetId = undefined;
+    }
+  }
+
+  private makePivotFullScreen(pivotId: UID) {
+    const sheetId = new UuidGenerator().smallUuid();
+
+    this.model.dispatch("CREATE_SHEET", {
+      position: this.getters.getSheetIds().length,
+      name: this.getters.getNextSheetName(),
```
hokolomopo, line 89: So ... we want to discard all the changes done after in the playground. This is probably the less worse option we found, and it's not great (changes in model/sheetview were necessary to make multiple undo in a single command work when we don't have a transport service).

But the alternatives are somehow worse:
- creating  new model to display the playground. We'd need a way to not reload every pivot (using value from the other model), because that can take a while in dahsboard
- similar to the first version of version history, do a `session.revision.goTo(revisionId)`. It doesn't really bring much over spamming UNDO, and we'd need to remember the changed datasources and refresh them by hand. (the plugins handle that in UNDO)

src/components/helpers/dom_helpers.ts
```diff
@@ -259,3 +259,50 @@ function isOtherMobileOS() {
 export function isMobileOS() {
   return isAndroid() || isIOS() || isOtherMobileOS();
 }
+
+/** Return a base64 representation of the current spreadsheet, with the grid canvas and all the figures */
+export function getSpreadsheetAsBase64() {
```
hokolomopo, line 264: So ... when we open a pivot in fullscreen we use a dialog-like style with a backdrop. But we cannot have two separate viewports, so we convert the spreadsheet to a png, and use that as background image 🙃

src/components/full_screen_sheet/full_screen_sheet.ts
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  exitFullScreen() {
+    this.fullScreenSheetStore.exitFullScreen();
+    // ADRM TODO DISCUSS WITH RAR: Manually resize the sheet here avoid having a flicker when exiting the full screen, because the
+    // first render has the wrong size, then the resize observer is triggered and the size is fixed for the next render.
+    // Kinda ugly. Not sure if we want to keep it. Not sure what alternative we have.
+    const { width, height } = getRefBoundingRect(this.spreadsheetRef);
+    this.env.model.dispatch("RESIZE_SHEETVIEW", { width, height, gridOffsetX: 0, gridOffsetY: 0 });
+  }
```
hokolomopo, line 56: 🙈

src/model.ts
```diff
@@ -690,9 +701,16 @@ export class Model extends EventBus<any> implements CommandDispatcher {
...
+      throw new Error(
+        "Cannot switch to dashboard mode in a collaborative session. The model shouldn't have a transport service."
+      );
+    }
     // @ts-ignore For testing purposes only
     this.config.mode = mode;
-    this.trigger("update");
+    if (this.status === Status.Ready) {
+      this.trigger("update");
+    }
```
hokolomopo, line 713: Probably need to change so we can change the mode of the model in a clean way.

And aren't all those `if (this.status === Status.Ready) {` pretty and cool ? 🙈 🙉

src/components/full_screen_sheet/full_screen_sheet.scss
```diff
@@ -0,0 +1,50 @@
...
+    @keyframes growIn {
+      /* Hack to have the first keyFrame be the final size of the content, because resizeObservers won't react to animation afterwards */
+      0% {
+        transform: scale(1);
+        visibility: hidden;
+      }
+      1% {
+        transform: scale(0.8);
+        visibility: visible;
+      }
```
hokolomopo, line 45: yeah ... 🙈

hokolomopo: Image

hokolomopo (commented):
So there's a bunch of weird/strange/magic code in there. But for most of that, I don't think we have much of a choice to reach the specs 😔

demo/main.js
```diff
@@ -251,7 +258,22 @@ class Demo extends Component {
...
+  }
+
+  /**
+   *
+   * @param {Object} [args.data] - The initial dataset to load into the model. If omitted, a default demo dataset is used.
+   * @param {boolean} [args.enableTransportService] - Whether to enable the collaborative WebSocket transport service.
+   * @returns {Promise<void>}
+   */
+  async initiateConnection(args) {
+    if (args.enableTransportService) {
```
pro-odoo, line 270: `withTransportService` ? 
Nitpick

src/components/spreadsheet editor/spreadsheet editor.ts
```diff
@@ -0,0 +1,152 @@
+import {
```
pro-odoo, line 1: typo in the file name => `_` missing
And in the folder name

pro-odoo: Same for the xml one

src/components/full_screen_sheet/full_screen_sheet_store.ts
pro-odoo: I'm still unsure of the chosen strategy, IMO the creation of a new model is a cleaner way to do:
- No changes in the model/sheetview
- No magic undo needed
- No restriction to have a transport service in dashboard (useless for now, but it could be useful if your dashboard is in an external screen and you do not want to refresh everything when you change something in the dashboard.)
- No need to make a screenshot of the dashboard
- No scroll position loosing
- Almost all your comments are outdated :D 

The only downside of this solution is the need to reload all the data. For that, we have multiple solutions:
- Make a cache of the read_group/search_read. I wonder if we could re-use the cache version of framework js...  In that case, it's probably easy. The only thing we have to make sure is to invalidate this cache when the commands `REFRESH_ALL_DATA_SOURCE`, `REFRESH_PIVOT` are triggered. 
- Do nothing => Ok bad solution, but still
- Find a way to share PivotModel and ListDataSource to the newly created model

What do you think ?

pro-odoo (changes requested):
Not confortable with the chosen solution, see my answer to your comment

src/components/grid_overlay/grid_overlay.ts
```diff
@@ -68,15 +68,15 @@ function useCellHovered(env: SpreadsheetChildEnv, gridRef: Ref<HTMLElement>): Pa
     };
   }
 
-  const { pause, resume } = useInterval(checkTiming, 200);
+  const { pause, resume } = useInterval(checkTiming, 100);
```
rrahir, line 71: feels kewl

tests/setup/jest.setup.ts
```diff
@@ -85,6 +85,12 @@ beforeEach(() => {
...
       height: 1000 + HEADER_HEIGHT,
     }),
+    "o-spreadsheet-editor": () => ({
+      x: 0,
+      y: 0,
+      width: 1000 + HEADER_WIDTH,
+      height: 1000 + HEADER_HEIGHT,
+    }),
     "o-grid": () => ({ x: 0, y: 0, width: 1000 + HEADER_WIDTH, height: 1000 + HEADER_HEIGHT }),
     // "o-grid-overlay": () => ({ x: HEADER_WIDTH, y: HEADER_HEIGHT, width: 1000 + HEADER_WIDTH, height: 1000 + HEADER_HEIGHT }),
```
rrahir, line 95: my fault, ofc but I guess we can remove it in this PR

src/components/dashboard/dashboard.ts
```diff
@@ -130,11 +122,8 @@ export class SpreadsheetDashboard extends Component<Props, SpreadsheetChildEnv>
     });
   }
 
-  private getGridRect(): Rect {
-    return {
-      ...getRefBoundingRect(this.gridRef),
-      ...this.env.model.getters.getSheetViewDimensionWithHeaders(),
-    };
+  getGridRect(): Rect {
+    return getRefBoundingRect(this.gridRef);
```
rrahir, line 126: could we set the t-ref to the actual ".o-grid.o-two-columns" element of the DOM instead of its child for consistency ?

src/components/grid_overlay/grid_overlay.ts
```diff
@@ -84,11 +84,19 @@ function useCellHovered(env: SpreadsheetChildEnv, gridRef: Ref<HTMLElement>): Pa
     }
   }
   function updateMousePosition(e: MouseEvent) {
-    if (isChildEvent(gridRef.el, e)) {
+    // moving on the grid overlay or the grid (which is a parent and does not necessarily have the same size)
```
rrahir, line 87: this is super fishy, the hook should then be defined on the parent... Except it's just a parent class, not a component. If we decided to remove the grid class of  `Dashboard` I wonder how many features would crumble :/

rrahir: shouldn't we define useCellHovered on the parent component and pass it as a props to the child? Might trigger renders of the parent instead of just  gridOverlay though

src/types/commands.ts
```diff
@@ -1028,6 +1029,15 @@ export interface SplitPivotFormulaCommand extends PositionDependentCommand {
   pivotId: UID;
 }
 
+export interface MakePivotFullScreen {
```
rrahir, line 1032:
```suggestion
export interface MakePivotFullScreenCommand {
```

src/components/full_screen_sheet/full_screen_sheet_store.ts
```diff
@@ -0,0 +1,123 @@
...
+      sheetIdTo: sheetId,
+    });
+  }
+
+  makePivotFullScreen(pivotId: UID) {
+    this.model.dispatch("MAKE_PIVOT_FULL_SCREEN", { pivotId });
+    // AUTORESIZE in separate command because we need to wait for a finalize() to have an evaluation
+    if (this.fullScreenSheetId) {
+      const cols = this.getters.getPivot(pivotId).getExpandedTableStructure().numberOfColumns;
+      this.model.dispatch("AUTORESIZE_COLUMNS", {
```
rrahir, line 47: :puke:

rrahir (changes requested):

PR #6809, opened 2025-07-16
[FIX] pivot: support SUM and AVG aggregators for duration fields

rmbh-odoo:
## Description:
Steps to reproduce:
- Create a pivot with a duration-formatted column.
- Go to the side panel and add the duration field to the Measures section.

Current behavior before PR:
- Datetime-formatted fields (such as durations) could not
be aggregated using SUM or AVG in pivot measures.
- This blocked users from summing or averaging durations

- SUM and AVG aggregators are supported for datetime fields,
- Allowing meaningful aggregation of duration values in pivot tables.

Task: [4945217](https://www.odoo.com/odoo/2328/tasks/4945217)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/pivot/pivot_helpers.ts
+++ b/src/helpers/pivot/pivot_helpers.ts
@@ -44,13 +44,13 @@ const AGGREGATOR_NAMES = {
   sum: _t("Sum"),
 };
 
-const NUMBER_CHAR_AGGREGATORS = ["max", "min", "avg", "sum", "count_distinct", "count"];
+const DEFAULT_AGGREGATORS = ["max", "min", "avg", "sum", "count_distinct", "count"];
 
 const AGGREGATORS_BY_FIELD_TYPE = {
-  integer: NUMBER_CHAR_AGGREGATORS,
-  char: NUMBER_CHAR_AGGREGATORS,
+  integer: DEFAULT_AGGREGATORS,
+  char: DEFAULT_AGGREGATORS,
+  datetime: DEFAULT_AGGREGATORS,
   boolean: ["count_distinct", "count", "bool_and", "bool_or"],
-  datetime: ["max", "min", "count_distinct", "count"],
 };
 
 export const AGGREGATORS = {};
--- a/tests/pivots/spreadsheet_pivot/spreadsheet_pivot.test.ts
+++ b/tests/pivots/spreadsheet_pivot/spreadsheet_pivot.test.ts
@@ -1808,6 +1808,30 @@ describe("Spreadsheet Pivot", () => {
     expect(getEvaluatedCell(model, "B27").formattedValue).toBe("2022/04/14 01:02:03");
   });
 
+  test("Pivot aggregates duration fields with correct SUM and AVG values", () => {
+    const model = createModelFromGrid({
+      A1: "Duration",
+      A2: "01:30:00",
+      A3: "02:00:00",
+      A4: "00:45:00",
+    });
+    addPivot(model, "A1:A4", {
+      columns: [],
+      rows: [],
+      measures: [
+        { id: "Duration:avg", fieldName: "Duration", aggregator: "avg" },
+        { id: "Duration:sum", fieldName: "Duration", aggregator: "sum" },
+      ],
+    });
+    setCellContent(model, "A6", '=PIVOT.VALUE(1,"Duration:avg")');
+    setCellContent(model, "B6", '=PIVOT.VALUE(1,"Duration:sum")');
+
+    const pivot = model.getters.getPivot("1");
+    expect(pivot.getFields().Duration?.type).toBe("datetime");
+    expect(getEvaluatedCell(model, "A6").formattedValue).toBe("01:25:00");
+    expect(getEvaluatedCell(model, "B6").formattedValue).toBe("04:15:00");
+  });
+
   test("Pivot with measure AVG on text values does not crash", () => {
     const model = createModelFromGrid({ A1: "Customer", A2: "Jean", A3: "Marc" });
     addPivot(model, "A1:A3", {
--- a/tests/pivots/spreadsheet_pivot/spreadsheet_pivot_side_panel.test.ts
+++ b/tests/pivots/spreadsheet_pivot/spreadsheet_pivot_side_panel.test.ts
@@ -440,6 +440,33 @@ describe("Spreadsheet pivot side panel", () => {
     ]);
   });
 
+  test("Supports SUM and AVG aggregators for duration fields", async () => {
+    setCellContent(model, "A1", "name");
+    setCellContent(model, "A2", "Alice");
+    setCellContent(model, "B1", "duration");
+    setCellContent(model, "B2", "01:30:00");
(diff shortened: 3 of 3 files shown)


src/helpers/pivot/pivot_helpers.ts
```diff
@@ -44,13 +44,13 @@ const AGGREGATOR_NAMES = {
...
-const NUMBER_CHAR_AGGREGATORS = ["max", "min", "avg", "sum", "count_distinct", "count"];
+const DEFAULT_AGGREGATORS = ["max", "min", "avg", "sum", "count_distinct", "count"];
 
 const AGGREGATORS_BY_FIELD_TYPE = {
-  integer: NUMBER_CHAR_AGGREGATORS,
-  char: NUMBER_CHAR_AGGREGATORS,
+  integer: DEFAULT_AGGREGATORS,
+  char: DEFAULT_AGGREGATORS,
+  datetime: DEFAULT_AGGREGATORS,
   boolean: ["count_distinct", "count", "bool_and", "bool_or"],
```
hokolomopo, line 53: I don't think this is the correct fix. Date values didn't have SUM aggregator by design, because it doesn't make sense to sum `12/01/2024` and `11/01/2024` to end up with something like `1/1/4048`.

I think if the field has a duration format, it should be detected as a number field instead of a date field. Unless i'm missing some other behaviour where we would want durations to act as a date field.

rmbh-odoo: Excel and Google Sheets store durations as datetime values and allow SUM/AVG on all datetime-formatted fields.

We can’t add our check in isDateTimeFormat, since it's used globally — and durations should still be treated as datetime for formatting or calculations elsewhere.

In pivot, we specifically detect durationFormat and handle it separately, instead of treating it as full datetime.

Let me know if this works or if I’m missing any edge case! 🙂

src/helpers/format/format.ts
```diff
@@ -442,6 +442,16 @@ export function numberToString(number: number, decimalSeparator: string): string
...
+/**
+ * Check if the given date format represents a duration,
+ * such as "hh:mm", "hh:mm:ss", or "hhhh:mm:ss".
+ * It must include hours ("h") and minutes ("m"),
+ * and must not contain date parts ("d", "y") or time-of-day indicators like "a" (AM/PM).
+ */
+export function isDurationFormat(format: Format): boolean {
+  return format.includes("h") && format.includes("m") && !/[ady]/.test(format);
+}
+
```
hokolomopo: I think duration is only `.*hhhh.*`. See `formatJSDatePart`

`hh:mm:ss` is a time, and it doesn't make much sense to SUM them. `12:00:00` + `20:00:00` would give `08:00:00` which doesn't make that much sense. I guess you could argue that AVG make some sense in that case, but it's probably not worth a separate case.

With `hhhh`, we get `32:00:00` which make sense.

src/helpers/pivot/spreadsheet_pivot/spreadsheet_pivot.ts
```diff
@@ -396,7 +396,11 @@ export class SpreadsheetPivot implements Pivot<SpreadsheetPivotRuntimeDefinition
     }
     if (
       nonEmptyCells.every(
-        (cell) => cell.type === CellValueType.number && cell.format && isDateTimeFormat(cell.format)
+        (cell) =>
+          cell.type === CellValueType.number &&
+          cell.format &&
+          isDateTimeFormat(cell.format) &&
+          !isDurationFormat(cell.format)
```
hokolomopo: nitpick: add a check that return `integer` if `isDurationFormat` is true, rather tha relying on the default case

tests/pivots/spreadsheet_pivot/spreadsheet_pivot.test.ts
```diff
@@ -1808,6 +1808,27 @@ describe("Spreadsheet Pivot", () => {
...
+      rows: [],
+      measures: [
+        { id: "Duration:avg", fieldName: "Duration", aggregator: "avg" },
+        { id: "Duration:sum", fieldName: "Duration", aggregator: "sum" },
+      ],
+    });
+    setCellContent(model, "A6", '=PIVOT.VALUE(1,"Duration:avg")');
+    setCellContent(model, "B6", '=PIVOT.VALUE(1,"Duration:sum")');
+    expect(getEvaluatedCell(model, "A6").formattedValue).toBe("01:25:00");
+    expect(getEvaluatedCell(model, "B6").formattedValue).toBe("04:15:00");
```
hokolomopo, line 1832: nitpick: you could also add a expect to check that the field in pivot.getFields() has the correct type

hokolomopo (commented):
👋

src/helpers/pivot/pivot_helpers.ts
hokolomopo: Oh you're right, Excel allows a SUM of a date field. Well the result doesn't make any sense, but it give the freedom for the user to do whatever he wants. You can run the idea by Francois if you want, it would make the code simpler, but don't think we should let user do things that don't make sense.

tests/pivots/spreadsheet_pivot/spreadsheet_pivot_side_panel.test.ts
```diff
@@ -440,6 +440,39 @@ describe("Spreadsheet pivot side panel", () => {
...
+    await click(fixture.querySelectorAll(".o-autocomplete-value")[1]);
+
+    // Default aggregator for duration fields is "count"
+    expect(model.getters.getPivotCoreDefinition("3").measures).toEqual([
+      { id: "duration:count", fieldName: "duration", aggregator: "count" },
+    ]);
+
+    let selectEl = fixture.querySelector(".pivot-measure select") as HTMLSelectElement;
+    selectEl.value = "sum";
+    selectEl.dispatchEvent(new Event("change"));
```
LucasLefevre: nitpicking: You can use the `setInputValueAndTrigger` helper function :wink: 
```suggestion
    await setInputValueAndTrigger(".pivot-measure select", "sum");
```

tests/pivots/spreadsheet_pivot/spreadsheet_pivot_side_panel.test.ts
```diff
@@ -440,6 +440,39 @@ describe("Spreadsheet pivot side panel", () => {
...
+
+    expect(model.getters.getPivotCoreDefinition("3").measures).toEqual([
+      { id: "duration:sum", fieldName: "duration", aggregator: "sum" },
+    ]);
+
+    selectEl = fixture.querySelector(".pivot-measure select") as HTMLSelectElement;
+    selectEl.value = "avg";
+    selectEl.dispatchEvent(new Event("change"));
```
LucasLefevre:
```suggestion
    await setInputValueAndTrigger(".pivot-measure select", "avg");
```

LucasLefevre (commented):
LGTM except the small comment to simplify the component test :+1:

PR #6816, opened 2025-07-18
[FIX] chart: chartJS extensions loaded too late

hokolomopo:
## Description

The `onMounted` hook in child components runs before the one in their parent. So loading the ChartJS extensions in the Spreadsheet's `onMounted` doesn't work, since the `ChartJsComponent` needs them in its own `onMounted`, which runs first.

It actually worked by purr chance before. At the first render of the model, the sheetview size isn't set yet, so `getters.getVisibleFigures()` doesn't return anything. And only at the second render the charts are displayed. However if we tried to load a `Spreadsheet` with a model that has an initialized sheetview size, the charts would not be loaded correctly.

Task: [4954034](https://www.odoo.com/odoo/2328/tasks/4954034)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/spreadsheet/spreadsheet.ts
+++ b/src/components/spreadsheet/spreadsheet.ts
@@ -2,6 +2,7 @@ import {
   Component,
   onMounted,
   onPatched,
+  onWillStart,
   onWillUnmount,
   onWillUpdateProps,
   useEffect,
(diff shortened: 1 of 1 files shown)


pro-odoo (commented):
robodoo r+

PR #6827, opened 2025-07-22
[FIX] Evaluation: Optimize invalidation of cell depending of spreads

rrahir:
Currently, we invalidate all the cells that depend on a spilling formula during its evaluation. However, some edgecases (like the one described below) can make the amount of dependant cells explode with a noticeable amount of redundancy in the cells invalidated.

study case:
- in A1 write  `=SPLIT("1 2", " ")`
- in C1, write `=B2`
- autofill these over 1000 columns
- in another sheet, write a VLOOKUP that search in column C of sheet 1 and autofill around 7000 times

-> the evaluation becomes really slow

The example above is problematic because for each cell evaluated in A1-B1, we will the corresponding C formula along with EVERY cell with a VLOOKUP, which means the cells with VLOOKUP will be invalidated 500 (# or rows) times.
It so happens that the invalidation step requires to regroup all the cells in zones and this step can be costy, especially if repeated hundreds of times.

This revision captures the cells that need to be invalidated during a pass of evaluation and invalidates all of them at once to limit the overhead of regrouping them.

In the aforementioned example, the time to evaluate drops from 6.7 seconds to 2.1 seconds on average.

Task: 4954710

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_core_views/cell_evaluation/evaluator.ts
+++ b/src/plugins/ui_core_views/cell_evaluation/evaluator.ts
@@ -55,6 +55,7 @@ export class Evaluator {
   );
   private blockedArrayFormulas = new PositionSet({});
   private spreadingRelations = new SpreadingRelation();
+  private positionsToInvalidate: RTreeBoundingBox[] = [];
 
   constructor(private readonly context: ModelConfig["custom"], getters: Getters) {
     this.getters = getters;
@@ -304,6 +305,8 @@ export class Evaluator {
           this.evaluatedCells.set(position, evaluatedCell);
         }
       }
+
+      this.invalidatePendingPositionsStack();
       onIterationEndEvaluationRegistry.getAll().forEach((callback) => callback(this.getters));
     }
     if (currentIteration >= MAX_ITERATION) {
@@ -395,7 +398,7 @@ export class Evaluator {
       // thanks to the isMatrix check above, we know that formulaReturn is MatrixFunctionReturn
       this.spreadValues(formulaPosition, formulaReturn)
     );
-    this.invalidatePositionsDependingOnSpread(formulaPosition.sheetId, resultZone);
+    this.positionsToInvalidate.push({ zone: resultZone, sheetId: formulaPosition.sheetId });
     return createEvaluatedCell(
       nullValueToZeroValue(formulaReturn[0][0]),
       this.getters.getLocale(),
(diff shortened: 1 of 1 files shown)


src/plugins/ui_core_views/cell_evaluation/evaluator.ts
```diff
@@ -412,6 +434,32 @@ export class Evaluator {
...
+      const boundingboxes = [...this.positionsToInvalidate]
+        .map((range) =>
+          excludeTopLeft(range.zone).map((zone) => ({
+            sheetId: range.sheetId,
+            zone,
+          }))
+        )
+        .flat();
+      const invalidatedPositions = this.formulaDependencies().getCellsDependingOn(boundingboxes);
+      // the result matrix is split in 2 zones to exclude the array formula position
```
rrahir: multiple zones, > plural form

src/plugins/ui_core_views/cell_evaluation/evaluator.ts
```diff
@@ -412,6 +415,32 @@ export class Evaluator {
     this.nextPositionsToUpdate.addMany(invalidatedPositions);
   }
 
+  private invalidatePendingPositionsStack() {
+    if (this.positionsToInvalidate.length > 0) {
+      // the result matrices are split in subzones to exclude the array formula position.
+      // This avoids to invalidate the topleft position of an array formula
+      // and create an infinite loop of self-invalidation
+      const boundingboxes = this.positionsToInvalidate
```
hokolomopo, line 423:
```suggestion
      const boundingBoxes = this.positionsToInvalidate
```

src/plugins/ui_core_views/cell_evaluation/evaluator.ts
```diff
@@ -412,6 +415,32 @@ export class Evaluator {
...
+        .map((range) =>
+          excludeTopLeft(range.zone).map((zone) => ({
+            sheetId: range.sheetId,
+            zone,
+          }))
+        )
+        .flat();
+      const invalidatedPositions = this.formulaDependencies().getCellsDependingOn(boundingboxes);
+      for (const range of this.positionsToInvalidate) {
```
hokolomopo, line 432: So this removes the top-left cells from the positions to invalidate ... wasn't this already done in the `excludeTopLeft` above ? I'm a bit confused

hokolomopo, line 432: So are they ranges or positions ? I feel like we could improve the naming here

src/plugins/ui_core_views/cell_evaluation/evaluator.ts
```diff
@@ -412,6 +415,32 @@ export class Evaluator {
...
+        invalidatedPositions.delete({
```
LucasLefevre, line 433: this is causing the issue @hokolomopo posted on the task chatter.
A1 should be recomputed (because A3 spilled on the zone it depends on), but this line removes it from `this.nextPositionsToUpdate`

PR #6838, opened 2025-07-23
[FIX] cf: do not override changes on cancel

hokolomopo:
## Description

- Open the side panel to edit a conditional format.
- Another user changes the CF
- The first user presses cancel
- The changes from the second user will be overriden. They should not, the first user changed nothing and pressed cancel.

This commit fixes that, if the user did not change anything, cancel will do nothing. Note that as soon the user changes anything, pressing cancel will override the changes from another user. We cannot have a side panel that both make the changes in the sheet instantly, and also preserve changes from other users.

Task: [4609695](https://www.odoo.com/odoo/2328/tasks/4609695)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/conditional_formatting/cf_editor/cf_editor.ts
+++ b/src/components/side_panel/conditional_formatting/cf_editor/cf_editor.ts
@@ -137,8 +137,9 @@ css/* scss */ `
 `;
 interface Props {
   editedCf: ConditionalFormat;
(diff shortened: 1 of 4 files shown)


rrahir:
robodoo r+

PR #6841, opened 2025-07-24
Broken behavior with "Delete cell" in a array formula

hokolomopo:
## Description:

Task: [4938311](https://www.odoo.com/odoo/2328/tasks/4938311)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/clipboard_handlers/cell_clipboard.ts
+++ b/src/clipboard_handlers/cell_clipboard.ts
@@ -6,6 +6,7 @@ import {
   CellPosition,
   ClipboardCell,
   ClipboardCellData,
+  ClipboardCopyOptions,
   ClipboardOptions,
   ClipboardPasteTarget,
   CommandResult,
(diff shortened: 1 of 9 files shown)


src/clipboard_handlers/cell_clipboard.ts
```diff
@@ -32,9 +33,11 @@ export class CellClipboardHandler extends AbstractCellClipboardHandler<
     return CommandResult.Success;
   }
 
-  copy(data: ClipboardCellData): ClipboardContent | undefined {
+  copy(
+    data: ClipboardCellData,
+    mode: ClipboardCopyOptions = "copyPaste"
+  ): ClipboardContent | undefined {
```
hokolomopo, line 39: Alternatively, we could have created a new separate clipboard handler that handle copying part of the array formula, one for part or tables, and not loaded tests clipboards on INSERT_CELL. Would have made the code more complex tho, so not sure it's worth.

rrahir: How about a new boolean option argument (similar to the `isCutOperation` on the paste method) that we call `skipSpilledCells` or something similar

achieves the same but having a method 'copy' that specifically has a mode different than 'copy' looks weird.

It would be nice if the commit message also mentionned the "move_header" command

hokolomopo: Table style are not spilled cells tho :/ I don't have better ideas for a name

rrahir: yup ...

hokolomopo: And the commit message already mention move header no ? `The insert/delete cell features, as well as the drag & drop of rows/columns is implemented with the clipboards`

rrahir:
robodoo r+

PR #6849, opened 2025-07-25
[FIX] clipboard: insert cells bugged with array formulas

hokolomopo:
## Description

The insert/delete cell features, as well as the drag & drop of rows/columns is implemented with the clipboards. Under the hood, they do cut/paste operations.

But some behaviour is specific to copy/paste and should not be applied to those features: here we should not copy part of array formulas. We don't want an "insert cell" inside an array formula to copy a part of the values of the array formula.

Task: [4938311](https://www.odoo.com/odoo/2328/tasks/4938311)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/clipboard/clipboard_cells_state.ts
+++ b/src/helpers/clipboard/clipboard_cells_state.ts
@@ -16,7 +16,12 @@ import {
   UID,
   Zone,
 } from "../../types";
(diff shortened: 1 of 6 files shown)


hokolomopo:
@robodoo fw=no

rrahir:
robodoo r+

PR #6854, opened 2025-07-25
[FIX] clipboard: don't paste both image and text content

hokolomopo:
## Description

If we have a clipboard content that has bot an image and some text data, we would paste both inside the spreadsheet. There's a good chance that this is not what the user expects, as the iamge and the text are likely alternatives for the same content.

That's what happened when pasting from Excel (desktop). The user would end up with both the copied data, and an image of the copied data.

With this commit, we'll ignore the image data if we have text data in the clipboard.

Task: [4876682](https://www.odoo.com/odoo/2328/tasks/4876682)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


rrahir:
robodoo r+

PR #6869, opened 2025-07-29
[FIX] helpers: preserve sparse(empty) elements in removeIndexesFromArray

rmbh-odoo:
## Description:
Steps to reproduce:
- Enter multiline text in row 1, 2, and (optionally) row 3.
- Resize one of the rows (e.g. row 1 or 2).
- Move the other row above the resized one.
- One of the rows either resets to the default height or wrongly keeps a user-defined height.

Current behavior before PR:
- removeIndexesFromArray() used filter(), which skips sparse (empty) elements.
- This caused user-defined row sizes (stored sparsely) to be lost or misapplied during row moves.

- Rewrote the function to use a for loop with Set to track removed indexes.
- This preserves all array elements, including sparse (empty) ones.

Task: [4977932](https://www.odoo.com/odoo/2328/tasks/4977932)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/misc.ts
+++ b/src/helpers/misc.ts
@@ -506,8 +506,17 @@ export function memoize<T extends any[], U>(func: (...args: T) => U): (...args: 
   }[funcName];
 }
 
+/**
+ * removeIndexesFromArray removes only the specified indexes from the array.
+ * Sparse (empty) elements are preserved unless their index is explicitly removed.
+ */
 export function removeIndexesFromArray<T>(array: readonly T[], indexes: number[]): T[] {
-  return array.filter((_, index) => !indexes.includes(index));
+  const toRemove = new Set(indexes);
+  const newArray: T[] = [];
+  for (let i = 0; i < array.length; i++) {
+    if (!toRemove.has(i)) newArray.push(array[i]);
+  }
+  return newArray;
 }
 
 export function insertItemsAtIndex<T>(array: readonly T[], items: T[], index: number): T[] {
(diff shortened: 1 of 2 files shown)


src/helpers/misc.ts
```diff
@@ -506,8 +506,17 @@ export function memoize<T extends any[], U>(func: (...args: T) => U): (...args:
   }[funcName];
 }
 
+/**
+ * removeIndexesFromArray removes only the specified indexes from the array.
+ * Sparse (empty) elements are preserved unless their index is explicitly removed.
```
LucasLefevre: > Sparse (empty) elements are preserved

well, not correct strictly speaking as empty elements are transformed to `undefined`

src/helpers/misc.ts
```diff
@@ -506,8 +506,17 @@ export function memoize<T extends any[], U>(func: (...args: T) => U): (...args:
...
+ */
 export function removeIndexesFromArray<T>(array: readonly T[], indexes: number[]): T[] {
-  return array.filter((_, index) => !indexes.includes(index));
+  const toRemove = new Set(indexes);
+  const newArray: T[] = [];
+  for (let i = 0; i < array.length; i++) {
+    if (!toRemove.has(i)) newArray.push(array[i]);
```
LucasLefevre: nitpicking: we usually avoid one-line ifs, it's less readable
```suggestion
    if (!toRemove.has(i)) {
      newArray.push(array[i]);
    }
```

PR #6904, opened 2025-08-04
[IMP] chart: generate smart chart definition based on all selected zones

rmbh-odoo:
## Description:

Current behavior before PR:
- Only the last selected zone was used when inserting a chart.
- The chart type and structure were determined solely from that zone.

- All selected zones are considered to determine the most suitable chart.
- The chart type is inferred based on cell types, counts, and structure.
- Labels and dataset titles are detected automatically.
- This improves UX when inserting charts
- Examples:
  - One number column then bar, doughnut, pie or scorecard based on total/count.
  - Text + Number then bar, pie or treemap depending on unique values.
  - Date + Number then line chart.
  - Multiple text columns + number column then treemap or sunburst.
  - Cyclic labels + numbers then radar chart.

Task: [4893903](https://www.odoo.com/odoo/2328/tasks/4893903)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/helpers/figures/charts/smart_chart_engine.ts
```diff
@@ -0,0 +1,463 @@
...
+    }
+    if (uniqueCount <= 6) {
+      return {
+        type: "pie",
+        title: {},
+        dataSets,
+        labelRange: dataRange,
+        dataSetsHaveTitle,
+        aggregated: true,
+        legendPosition: "none",
```
dhrp-odoo: Why hide the legend? Wouldn't it be more useful to display it, especially for a pie chart with categories?

src/helpers/figures/charts/smart_chart_engine.ts
```diff
@@ -0,0 +1,463 @@
...
+      title: {},
+      dataSets,
+      labelRange: dataRange,
+      dataSetsHaveTitle,
+      aggregated: true,
+      stacked: false,
+      legendPosition: "none",
+    };
+  }
+  return {
```
dhrp-odoo, line 120: For a single column of dates, it makes sense to use a line chart. But shouldn’t the dates be on the Y-axis instead of the X-axis? Otherwise it will be just flat line.

src/helpers/figures/charts/smart_chart_engine.ts
```diff
@@ -0,0 +1,463 @@
...
+      legendPosition: "top",
+    };
+  }
+
+  // Case: Radar Chart (cyclic categories)
+  const labelValues = textCols[0]?.values;
+  if (
+    textCols.length === 1 &&
+    numericCols.length >= 2 &&
+    isCyclicData(dataSetsHaveTitle ? labelValues.slice(1) : labelValues)
```
dhrp-odoo: The definition of isCyclicData is quite specific and works on the assumption that labels are in English.

src/helpers/figures/charts/smart_chart_engine.ts
```diff
@@ -0,0 +1,463 @@
...
+  const titleCell = analyzer.getters.getEvaluatedCell({
+    sheetId: analyzer["sheetId"],
+    col: zone.left,
+    row: zone.top,
+  });
+  const dataSetsHaveTitle =
+    titleCell.type !== CellValueType.number && titleCell.type !== CellValueType.empty;
+
+  if (type === "number") {
+    if (sum <= 100) {
```
dhrp-odoo: Would it make sense to also check for values formatted with a `%`?

src/helpers/figures/charts/smart_chart_engine.ts
```diff
@@ -0,0 +1,463 @@
...
+    return cells.map((c) => c.formattedValue ?? "").filter((v) => v !== "");
+  }
+
+  private detectType(cells: EvaluatedCell[]): ColumnType {
+    const counts = { number: 0, text: 0, date: 0 };
+    for (const cell of cells) {
+      if (cell.type === CellValueType.number) {
+        if (cell.format && isDateTimeFormat(cell.format)) counts.date++;
+        else counts.number++;
+      } else if (cell.type === CellValueType.text) counts.text++;
```
dhrp-odoo: Expand the if blocks with braces {} to make it easier to read.

src/helpers/figures/charts/smart_chart_engine.ts
rmbh-odoo: Maybe a line chart for a single column of date values doesn’t make much sense unless there's another variable to plot.

Right now, there are two issues:

* [`beginAtZero`](https://github.com/odoo/o-spreadsheet/blob/master/src/helpers/figures/charts/runtime/chartjs_scales.ts#L375) is always true, so we should disable it.
* Dates are shown on the X-axis, but in this case, they should ideally be on the Y-axis, which we don’t currently support.

I’ll check with the PO if there’s any real use case for this. If needed, we can create an improvement task to handle this properly.

Update: once the time matrix chart is merged, we could improve this :)

src/helpers/figures/charts/smart_chart_engine.ts
rmbh-odoo: For simplicity, I’ve removed the isCyclic logic since it was too specific (English-only). Instead, we now rely on column labels to cover a more general and flexible case.

src/helpers/figures/charts/smart_chart_engine.ts
```diff
@@ -0,0 +1,429 @@
...
+
+  if (type === "number") {
+    const sum = analyzer.getSum(values);
+    if (sum <= 100 && sum > 0) {
+      return {
+        type: "pie",
+        title: {},
+        dataSets,
+        dataSetsHaveTitle,
+        isDoughnut: sum < 100,
```
LucasLefevre: what's the rational for choosing a doughnut instead of pie if sum < 100?

src/helpers/figures/charts/smart_chart_engine.ts
```diff
@@ -0,0 +1,429 @@
...
+    }
+  }
+
+  if (type === "text") {
+    return {
+      type: "pie",
+      title: {},
+      dataSets,
+      labelRange: dataRange,
+      dataSetsHaveTitle,
```
LucasLefevre: I believe `dataSetsHaveTitle` doesn't do anything for pie charts and `titleCell` shoudl be used for `title`

src/helpers/figures/charts/smart_chart_engine.ts
```diff
@@ -0,0 +1,429 @@
...
+        labelRange,
+        dataSetsHaveTitle,
+        isDoughnut: sum < 100,
+        aggregated: true,
+        legendPosition: "top",
+      };
+    }
+
+    // Non-unique labels suggest hierarchical data
+    if (uniqueCount !== labelCol.values.length) {
```
LucasLefevre: How can it be hierarchical if there's only 1 column for the categories?
The comment is misleading

src/helpers/figures/charts/smart_chart_engine.ts
```diff
@@ -0,0 +1,429 @@
...
+      sheetId,
+      col: colInfo.zone.left,
+      row: colInfo.zone.top,
+    });
+    return cell.type !== CellValueType.empty && cell.type !== CellValueType.number;
+  });
+
+  // Hierarchical data: Multiple text columns + one numeric
+  if (textCols.length >= 2 && numericCols.length === 1) {
+    const sortedTextCols = [...textCols].sort((a, b) => {
```
LucasLefevre:
```suggestion
    const sortedTextCols = [...textCols].sort((colA, colB) => {
```

src/helpers/figures/charts/smart_chart_engine.ts
```diff
@@ -0,0 +1,429 @@
...
+      const uniqA = analyzer.getUniqueLabelCount(a.values);
+      const uniqB = analyzer.getUniqueLabelCount(b.values);
+      return uniqA - uniqB;
+    });
+    const dataSets = sortedTextCols
+      .map((z) => z.zone)
+      .map((zone) => ({
+        dataRange: analyzer.getUnboundRange(zone),
```
LucasLefevre: you can avoid the double `.map` which allocates a new array for nothing
```suggestion
    const dataSets = sortedTextCols
      .map(({ zone }) => ({
        dataRange: analyzer.getUnboundRange(zone),
```

src/helpers/figures/charts/smart_chart_engine.ts
```diff
@@ -0,0 +1,429 @@
...
+  public getSum(values: string[]): number {
+    return values.reduce((sum, value) => {
+      if (!value) return sum;
+      const num = parseFloat(value.replace("%", ""));
+      return sum + (!isNaN(num) && isFinite(num) ? num : 0);
+    }, 0);
+  }
+
+  public getUniqueLabelCount(values: string[]): number {
+    return new Set(values).size;
```
LucasLefevre: those 3 methods do not belong to the class. They do not use anything from the object instance. They should be simple functions

src/helpers/figures/charts/smart_chart_engine.ts
```diff
@@ -0,0 +1,429 @@
...
+import { getZoneArea, getZonesByColumns, zoneToXc } from "../../zones";
+
+type ColumnType = "number" | "text" | "date" | "empty";
+type ColumnInfo = {
+  zone: Zone;
+  type: ColumnType;
+  values: string[];
+};
+
+class SmartChartAnalyzer {
```
LucasLefevre: I don't think a class is useful here.
None of the public methods use `this` (except `getColumnsByType` which could be inlined or a little helper function).

Let's keep it simple with POJOs and functions :wink:

src/helpers/figures/charts/smart_chart_engine.ts
```diff
@@ -0,0 +1,429 @@
...
+        const type = this.detectType(cells);
+        return type === "empty" ? null : { zone, type, values };
+      })
+      .filter(Boolean) as ColumnInfo[];
+  }
+
+  private getValues(cells: EvaluatedCell[]): string[] {
+    const values: string[] = [];
+    for (const cell of cells) {
+      const value = cell.formattedValue ?? "";
```
LucasLefevre: why are you using `cell.formattedValue` instead of `cell.value` ? It's error prone
I'm sure there's a reason but is it a good reason :smile: ? It looks like a shortcut to make something else easier (handling the percentages maybe?) but it's error prone

src/helpers/figures/charts/smart_chart_engine.ts
```diff
@@ -0,0 +1,429 @@
...
+      labelRange: dataRange,
+      dataSetsHaveTitle,
+      aggregated: true,
+      legendPosition: "top",
+    };
+  }
+
+  if (type === "date") {
+    return {
+      type: "line",
```
LucasLefevre: once the time matrix chart is merged, we could improve this :) This chart makes sense for one temporal column
https://www.odoo.com/odoo/project/2328/tasks/4816317

src/helpers/figures/charts/smart_chart_engine.ts
```diff
@@ -0,0 +1,429 @@
...
+ * 1. No valid data -> fallback bar chart.
+ *
+ * 2. One column:
+ *    - If only one non-empty cell -> Scorecard.
+ *    - If number column:
+ *        - Sum < 100 && Sum > 0 -> Doughnut chart.
+ *        - Sum = 100 -> Pie chart.
+ *        - Else -> Bar chart.
+ *    - If text column -> Pie chart.
+ *    - If date column -> Line chart.
```
LucasLefevre: This part of the comment explaining the rules is useless: I read the code if I want to know the rules, it's easy to read. And if (when) the rules change, there's a good chance this comment won't be updated accordingly

src/helpers/figures/charts/smart_chart_engine.ts
rmbh-odoo: Yeah, the same for the percentage use case 😅, but now I’m using 'value' instead of 'formattedValue'.

src/helpers/figures/charts/smart_chart_engine.ts
rmbh-odoo: Its mentioned in the task, but I’ve now changed the code to show a doughnut chart only for percentages. Thanks for pointing it out.😌

src/helpers/figures/charts/runtime/chart_data_extractor.ts
```diff
@@ -898,8 +898,8 @@ function getChartDatasetValues(getters: Getters, dataSets: DataSet[]): DatasetVa
     }
     const data = ds.dataRange ? getData(getters, ds) : [];
     if (
-      data.every((e) => typeof e === "string" && !isEvaluationError(e)) &&
-      data.some((e) => e !== "")
+      data.every((e) => e === null || (typeof e === "string" && !isEvaluationError(e))) &&
+      data.some((e) => typeof e === "string" && e !== "")
```
LucasLefevre: It counts all empty cells as ones
image
I believe the functional behavior we want is that the dataset range shouldn't be `E:E` but `E1:E22` instead when the chart is created

image

src/helpers/figures/charts/smart_chart_engine.ts
```diff
@@ -0,0 +1,368 @@
...
+const CHART_LIMITS = {
+  MAX_PIE_CATEGORIES: 7,
+  MAX_PIE_CATEGORIES_NO_TITLE: 6,
+  MAX_RADAR_LABELS: 12,
+  PERCENTAGE_THRESHOLD: 100,
+  MIN_RADAR_CATEGORIES: 3,
+} as const;
+const CHART_DEFAULTS = {
+  LEGEND_POSITION: "none" as const,
+  LEGEND_POSITION_TOP: "top" as const,
```
LucasLefevre:
```suggestion
  LEGEND_POSITION: "none",
  LEGEND_POSITION_TOP: "top",
```

src/helpers/figures/charts/smart_chart_engine.ts
```diff
@@ -0,0 +1,368 @@
...
+import { getZoneArea, getZonesByColumns, zoneToXc } from "../../zones";
+
+type ColumnType = "number" | "text" | "date" | "percentage" | "empty";
+
+interface ColumnInfo {
+  zone: Zone;
+  type: ColumnType;
+}
+interface AnalysisCtx {
+  sheetId: UID;
```
LucasLefevre: `sheetId` could also be part of `ColumnInfo`, what do you think ? It makes sense that the column is linked  to a given sheet

src/helpers/figures/charts/smart_chart_engine.ts
```diff
@@ -0,0 +1,368 @@
...
+        labelRange: dataRange,
+        dataSetsHaveTitle,
+        cumulative: CHART_DEFAULTS.CUMULATIVE,
+        labelsAsText: CHART_DEFAULTS.LABELS_AS_TEXT,
+      });
+  }
+  return createBaseChart("bar", [{ dataRange }], { dataSetsHaveTitle });
+}
+
+function buildTwoColumnChart(ctx: AnalysisCtx): ChartDefinition {
```
LucasLefevre: I think it would be more direct to just give the arguments just like that and remove the `ctx` concept.
What do you think ? (if `sheetId` is moved to `ColumnInfo`, it's also one less param to move around.)

Another idea would be to move everything in a dedicated plugin, so you'd have `this.getters` readily available everywhere
```suggestion
function buildTwoColumnChart(getters: Getters, columns: ...): ChartDefinition {
```

src/helpers/figures/charts/smart_chart_engine.ts
```diff
@@ -0,0 +1,368 @@
...
+        typeCounts.number++;
+      }
+    } else if (cell.type === CellValueType.text) {
+      typeCounts.text++;
+    }
+  }
+  return (Object.entries(typeCounts) as [ColumnType, number][]).reduce(
+    (max, [type, count]) => (count > max[1] ? [type, count] : max),
+    ["empty", 0]
+  )[0];
```
LucasLefevre: `reduce` is most of the time not easy to read
```suggestion
  let max = 0;
  let detectedType = "empty";
  for (const type in typeCounts) {
    const count = typeCounts[type];
    if (count > max) {
      max = count;
      detectedType = type;
    }
  }
  return detectedType as ColumnType;
```

src/helpers/figures/charts/smart_chart_engine.ts
```diff
@@ -0,0 +1,368 @@
...
+    }
+  }
+  return { sheetId, getters, columns };
+}
+
+function getCellStats(getters: Getters, sheetId: UID, zone: Zone) {
+  const cells = getters.getEvaluatedCellsInZone(sheetId, zone);
+  const uniqueValues = new Set<string>();
+  let totalCount = 0,
+    percentageSum = 0;
```
LucasLefevre:
```suggestion
  let totalCount = 0;
  let percentageSum = 0;
```

src/helpers/figures/charts/smart_chart_engine.ts
```diff
@@ -0,0 +1,368 @@
...
+  cells.forEach(({ value }) => {
```
LucasLefevre: could you use a good old for loop instead of `forEach` ? (usually more readable, consistent with the codebase, faster)

src/helpers/figures/charts/smart_chart_engine.ts
```diff
@@ -0,0 +1,368 @@
...
+    const str = value?.toString().trim();
+    if (!str) return;
+
+    uniqueValues.add(str);
+    totalCount++;
+
+    const num = Number(value);
+    if (!isNaN(num)) percentageSum += Math.abs(num) * 100;
```
LucasLefevre:
```suggestion
    if (!isNaN(num)) {
      percentageSum += Math.abs(num) * 100;
    }
```

src/helpers/figures/charts/smart_chart_engine.ts
rmbh-odoo: I don’t think that’s necessary, since we only create charts from the active sheet. Storing the sheetId at the column level doesn’t add much value in this case. What do you think?

EDIT: Now `sheetId` is fetched from getters, so there’s no need to pass it everywhere.

src/helpers/figures/charts/smart_chart_engine.ts
rmbh-odoo: I tried the first approach you mentioned, and while it works well for single-column cases, it gets messy for two- or multi-column scenarios; the arguments just pile up and create unnecessary code.

The plugin idea could work, but if this logic is pretty isolated and won’t be reused elsewhere, it might be overkill. I’d lean toward keeping it simple for now, unless we see repeated patterns that justify a plugin.

What do you think?

EDIT: Simplified and streamlined by directly passing columns and getters
removed the complex ctx.

src/helpers/figures/charts/runtime/chart_data_extractor.ts
rmbh-odoo: Yup, that makes sense. I’ll update it accordingly. 😉

src/helpers/figures/charts/smart_chart_engine.ts
```diff
@@ -0,0 +1,391 @@
...
+        if (val !== null && val !== "") {
+          if (firstIndex === -1) {
+            firstIndex = i;
+            titleCell = cells[i];
+          }
+          if (titleCell && cells[i].value === titleCell.value) {
+            titleCount++;
+          }
+          lastIndex = i;
+        } else if (firstIndex !== -1) {
```
LucasLefevre: I'm sorry, I mislead you with this comment https://github.com/odoo/o-spreadsheet/pull/6904#discussion_r2313354886

I checked with François and the data set should not be changed, it should still be E:E. (the pie chart runtime generation should be fixed to ignore empty cells though, but it can be another task :wink: )
Could you revert this part ? then we are good to go I think.
Sorry again

src/helpers/figures/charts/runtime/chart_data_extractor.ts
```diff
@@ -896,20 +896,21 @@ function getChartDatasetValues(getters: Getters, dataSets: DataSet[]): DatasetVa
...
-        (cell) => cell === undefined || cell === null || !isNumber(cell.toString(), DEFAULT_LOCALE)
-      )
-    ) {
+    const rawData = ds.dataRange ? getData(getters, ds) : [];
+    let hasValidString = false;
+    let hasValidNumber = false;
+    const data = rawData.map((e) => {
+      if (typeof e === "string" && e !== "" && !isEvaluationError(e)) {
+        hasValidString = true;
+        return 1;
```
rmbh-odoo: Thanks for the suggestion 🙂. Handled the pie chart case completely here, 
No separate future task is needed.

tests/figures/chart/menu_item_insert_chart.test.ts
```diff
@@ -524,41 +531,127 @@ describe("Insert chart menu item", () => {
...
+    payload.definition.isDoughnut = false;
+    expect(dispatchSpy).toHaveBeenCalledWith("CREATE_CHART", payload);
+  });
+
+  test("Chart with text cells (including empty cells) is a pie chart", () => {
+    setCellContent(model, "K1", "Country");
+    setCellContent(model, "K2", "India");
+    setCellContent(model, "K3", "Pakistan");
+    setCellContent(model, "K4", "India");
+    setCellContent(model, "K6", "USA");
```
rmbh-odoo, line 576: This test case covers the conditions for rendering a pie chart

src/helpers/figures/charts/runtime/chart_data_extractor.ts
LucasLefevre: Hmm with the current changes, the behavior is slightly weird when strings are mixed with numbers. I'm not entirely sure it's the behavior that we want. Could you check with FRGI ?

left: this PR; right: Gsheet
image

rmbh-odoo: Discussed with our PO, and the conclusion is that we will treat numbers as text when mixed with Strings. 
Keeping the behavior consistent with GSheets. 
The updates have already been applied. Please take a look. 😌

src/helpers/figures/charts/runtime/chart_data_extractor.ts
```diff
@@ -878,33 +878,23 @@ function getChartDatasetFormat(
...
-      data.some((e) => e !== "")
+      data.every((e) => !e || (typeof e === "string" && !isEvaluationError(e))) &&
+      data.filter((e) => typeof e === "string").length > 1
     ) {
-      // In this case, we want a chart based on the string occurrences count
-      // This will be done by associating each string with a value of 1 and
-      // then using the classical aggregation method to sum the values.
-      data.fill(1);
+      // Convert categorical data into counts
+      data = data.map(e => e && !isEvaluationError(e) ? 1 : null);
```
rmbh-odoo: Here in test https://github.com/odoo/o-spreadsheet/blob/master/tests/figures/chart/chart_plugin.test.ts#L3238
When numbers are used in a pie chart, we should ignore non-numeric values.  
This behavior works very well. 😄

PR #6907, opened 2025-08-05
[IMP] charts: add carousel figure

hokolomopo:
### [IMP] figure: add carousel figure

This commits introduces a new type of figure: the carousel figure.
A carousel figure allows to display multiple charts in a single figure,
with a tabbed interface to switch between them.

### [REF] charts: de-correlate chartId and figureId

This commit make it so that the `chartId` and `figureId` can be
different for a chart.

Migrations are needed:
- UPDATE/CREATE_CHART command now need both `chartId` and `figureId`.
- The `chartId` is put inside the exported figure data

The tests helpers were updated to create charts with a different `chartId`
and `figureId`, to properly test the new behavior.

Task: [4981828](https://www.odoo.com/odoo/2328/tasks/4981828)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/components/figures/chart/chartJs/chartjs.ts
```diff
@@ -173,8 +173,6 @@ export class ChartJsComponent extends Component<Props, SpreadsheetChildEnv> {
   }
 
   get animationFigureId() {
```
rrahir, line 175:
```suggestion
  get animationChartId() {
```

src/components/figures/figure/figure.xml
```diff
@@ -1,8 +1,9 @@
 <templates>
   <t t-name="o-spreadsheet-FigureComponent">
-    <div class="o-figure-wrapper pe-auto" t-att-style="wrapperStyle">
+    <div class="o-figure-wrapper" t-att-style="wrapperStyle" t-ref="figureWrapper">
```
rrahir, line 3: this removes the pe-auto for standard charts, wasn't there a reason why it was present in the first place?

src/plugins/ui_stateful/carousel_ui.ts
```diff
@@ -0,0 +1,133 @@
...
+      definition: CAROUSEL_DEFAULT_CHART_DEFINITION,
+    });
+
+    const definition: Carousel = { items: [...carousel.items, { type: "chart", chartId }] };
+    this.dispatch("UPDATE_CAROUSEL", { sheetId, figureId, definition });
+  }
+
+  private addFigureChartToCarousel(figureId: UID, chartFigureId: UID, sheetId: string) {
+    const chartId = this.getChartIdFromFigureId(chartFigureId);
+    if (!chartId) {
```
rrahir, line 118: I think this check should be in. a beforeHandle, otherwise we can have an history step taht does nothing

src/plugins/core/carousel.ts
```diff
@@ -0,0 +1,142 @@
...
+        this.dispatch("DELETE_CHART", { chartId: item.chartId, sheetId: cmd.sheetId });
+      }
+    }
+  }
+
+  import(data: WorkbookData) {
+    for (const sheet of data.sheets) {
+      const carousels = (sheet.figures || []).filter((figure) => figure.tag === "carousel");
+      for (const carousel of carousels) {
+        this.history.update("carousels", sheet.id, carousel.id, { items: carousel.data.items });
```
rrahir, line 127: we should not use histdory update at import but nitpick for later, maybe add a note

src/plugins/core/chart.ts
```diff
@@ -232,28 +301,45 @@ export class ChartPlugin extends CorePlugin<ChartState> implements ChartState {
...
-    return deepEquals(this.getChartDefinition(cmd.figureId), cmd.definition)
+    if (cmd.figureId !== this.charts[cmd.chartId]?.figureId) {
+      return CommandResult.Success;
+    }
+    return deepEquals(this.getChartDefinition(cmd.chartId), cmd.definition)
       ? CommandResult.NoChanges
       : CommandResult.Success;
   }
+
+  /** If the command would create a new figure, there need to be a position & offset defined in the command */
```
rrahir, line 335:
```suggestion
  /** If the command is meant a new figure, the position & offset argument need to be defined in the command */
```

src/plugins/ui_stateful/carousel_ui.ts
```diff
@@ -0,0 +1,133 @@
...
+      .find((chartId) => this.getters.getFigureIdFromChartId(chartId) === figureId);
+  }
+
+  private fixWrongCarouselState(figureId: UID) {
+    if (!this.getters.doesCarouselExist(figureId)) {
+      delete this.carouselStates[figureId];
+      return;
+    }
+    const carousel = this.getters.getCarousel(figureId);
+    if (!carousel) {
```
rrahir, line 89: don't think this can happen

src/plugins/core/carousel.ts
```diff
@@ -0,0 +1,142 @@
...
+                sheetId: cmd.sheetIdTo,
+                figureId: duplicatedFigureId,
+                offset: fig.offset,
+                col: fig.col,
+                row: fig.row,
+                size,
+                definition: {
+                  items: carousel.items.map((item): CarouselItem => {
+                    if (item.type === "carouselDataView") {
+                      return { type: "carouselDataView" };
```
rrahir, line 71: don't forget the title

src/plugins/core/carousel.ts
```diff
@@ -0,0 +1,142 @@
...
+                    }
+                    const chartIdBase = item.chartId.split(FIGURE_ID_SPLITTER).pop();
+                    const newChartId = `${cmd.sheetIdTo}${FIGURE_ID_SPLITTER}${chartIdBase}`;
+                    return { type: "chart", chartId: newChartId };
```
rrahir, line 75: same here

src/plugins/core/carousel.ts
hokolomopo: why ? we use `this.history.update` for cell/charts/probably some other plugins

rrahir:
robodoo r+

hokolomopo:
@robodoo rebase-ff

src/components/figures/figure/figure.xml
hokolomopo: It's still there, it was moved to the CSS, so it can be overridden without `!important`

src/plugins/ui_stateful/carousel_ui.ts
hokolomopo: I don't think we create a history step for a UI command that does nothing ? 🤔 

But yeah we could have an `allowDispatch`

src/plugins/core/carousel.ts
rrahir: it does exactly nothing ^^ and you'd think that you are registering an history related stage which is not true. It's just a mistake that we never really bothered to fix because it's has no functional sideeffect but at some point it added some overhead at import time (pretty sure it still does to some extent). I think we just use it be convenience since it'll generate multi depth objects automatically

src/plugins/ui_stateful/carousel_ui.ts
rrahir: allowDispatch, correct ><

src/plugins/core/carousel.ts
hokolomopo: I 100% used it for the convenience of multi-depth object creation 🙃 We can make a PR to remove all of these `this.history.update` if we really want to remove them

src/actions/menu_items_actions.ts
```diff
@@ -404,6 +404,7 @@ export const CREATE_CHART = (env: SpreadsheetChildEnv) => {
   const result = env.model.dispatch("CREATE_CHART", {
     sheetId,
     figureId,
+    chartId: figureId,
```
LucasLefevre, line 407: Shouldn't we create charts with different ids ? It would help discover bugs (places where we messed up) when we dev and play around

src/plugins/core/carousel.ts
LucasLefevre: > I think we just use it be convenience since it'll generate multi depth objects automatically

We also use it for read-only-ness .The plugin states should be typed as read-only to prevent updating it without using the history

src/actions/menu_items_actions.ts
hokolomopo: The test helpers now create charts with different ids, so it should be tested, but I guess it doesn't hurt to use a different id here too

PR #6908, opened 2025-08-05
[IMP] Infinite Viewport

fdamhaut:
Multiple improvement to allow for large sheet size

Task: [5008540](https://www.odoo.com/odoo/2328/tasks/5008540)

### DIFF
--- a/src/plugins/ui_core_views/cell_evaluation/binary_grid.ts
+++ b/src/plugins/ui_core_views/cell_evaluation/binary_grid.ts
@@ -1,6 +1,15 @@
+import { MAX_COL } from "../../../helpers";
 import { CellPosition } from "../../../types";
 
 type Bit = 0 | 1;
+
+const BLOCK_SIZE = 32;
+const BLOCK_HEIGHT = 8;
+const BLOCK_WIDTH = Math.floor(BLOCK_SIZE / BLOCK_HEIGHT);
+const BLOCK_COLUMN_OFFSET = Math.ceil(Math.log2(BLOCK_WIDTH));
+
+const HORIZONTAL_BLOCK_NUMBER = Math.ceil(MAX_COL / BLOCK_WIDTH);
+const HORIZONTAL_BLOCK_COLUMN_OFFSET = Math.ceil(Math.log2(HORIZONTAL_BLOCK_NUMBER));
 
 /**
  * Implements a fixed-sized grid or 2D matrix of bits.
@@ -14,22 +23,15 @@ type Bit = 0 | 1;
  * - bucket: Index of an item in the Uint32Array, a 32-bit integer.
  * - bitPosition: The position of a bit within the bucket 32-bit integer.
  */
-export class BinaryGrid extends Uint32Array {
-  private columnOffset = 0;
-  cols = 0;
-  rows = 0;
-
+export class SparseBinaryGrid {
+  private default: Bit = 0;
+  private grid: Map<number, number> = new Map();
   /**
    * Creates a binary grid of specified dimensions.
    */
-  static create(rows: number, columns: number): BinaryGrid {
-    const columnOffset = log2Ceil(columns);
-    const length = (rows << columnOffset) >> 5;
-    const grid = new this(length + 1);
-    grid.columnOffset = columnOffset;
-    grid.cols = columns;
-    grid.rows = rows;
-    return grid;
+  constructor(prefillRows: number = 1, prefillColumns: number = 1) {
+    // TODO
+    return this;
   }
 
   /**
@@ -37,7 +39,8 @@ export class BinaryGrid extends Uint32Array {
    */
   getValue(position: CellPosition): Bit {
     const [bucket, bitPosition] = this.getCoordinates(position);
-    return ((this[bucket] >> bitPosition) & 1) as Bit;
+    if (this.grid[bucket] === undefined) return this.default;
+    return ((this.grid[bucket] >> bitPosition) & 1) as Bit;
   }
 
   /**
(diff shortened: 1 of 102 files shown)


src/plugins/ui_core_views/cell_evaluation/binary_grid.ts
```diff
@@ -66,33 +70,25 @@ export class BinaryGrid extends Uint32Array {
...
-function log2Ceil(value: number) {
-  // A faster version of Math.ceil(Math.log2(value)).
-  if (value === 0) {
-    return -Infinity;
-  } else if (value < 0) {
-    return NaN;
+    const bucket =
+      (Math.floor(row / BLOCK_HEIGHT) << HORIZONTAL_BLOCK_COLUMN_OFFSET) +
+      Math.floor(col / BLOCK_WIDTH);
+    const index = (row % BLOCK_HEIGHT << BLOCK_COLUMN_OFFSET) + (col % BLOCK_WIDTH);
```
LucasLefevre, line 91: divisions and modulo are a lot slower. They take in the order of ~10-100 CPU cycles compared to addition/bit shifting (1 CPU cycle).
I think it could be done without / and % since everything is aligned on powers of 2

src/plugins/ui_core_views/cell_evaluation/binary_grid.ts
```diff
@@ -1,7 +1,16 @@
...
+const BLOCK_HEIGHT = 8;
+const BLOCK_WIDTH = Math.floor(BLOCK_SIZE / BLOCK_HEIGHT);
+const BLOCK_COLUMN_OFFSET = Math.ceil(Math.log2(BLOCK_WIDTH));
+
+const HORIZONTAL_BLOCK_NUMBER = Math.ceil(MAX_COL / BLOCK_WIDTH);
+const HORIZONTAL_BLOCK_COLUMN_OFFSET = Math.ceil(Math.log2(HORIZONTAL_BLOCK_NUMBER));
+
 /**
  * Implements a fixed-sized grid or 2D matrix of bits.
  * based on https://github.com/zandaqo/structurae
```
LucasLefevre, line 16: not based on https://github.com/zandaqo/structurae anymore

src/plugins/ui_core_views/cell_evaluation/binary_grid.ts
```diff
@@ -14,40 +23,35 @@ type Bit = 0 | 1;
...
  * - bitPosition: The position of a bit within the bucket 32-bit integer.
  */
-export class BinaryGrid extends Uint32Array {
-  private columnOffset = 0;
-  cols = 0;
-  rows = 0;
-
+export class SparseBinaryGrid {
+  private default: Bit = 0;
+  private grid: Map<number, number> = new Map();
```
LucasLefevre, line 28: I'd use a typed array (uint32 or maybe bigint64). It's most probably (a lot) faster than Map (hash table) and since we are manipulating bits, it's also more explicit (and more robust) rather than relying on implementation details of `number`

I think it's ok to have a mapping array to map blocks to their location in the typed array if the size of blocks isn't too small

src/plugins/ui_core_views/cell_evaluation/binary_grid.ts
```diff
@@ -14,40 +23,35 @@ type Bit = 0 | 1;
  * - bucket: Index of an item in the Uint32Array, a 32-bit integer.
  * - bitPosition: The position of a bit within the bucket 32-bit integer.
  */
-export class BinaryGrid extends Uint32Array {
-  private columnOffset = 0;
-  cols = 0;
-  rows = 0;
-
+export class SparseBinaryGrid {
```
LucasLefevre, line 26: we can keep the previous name IMO; sparse is an implementation detail

src/plugins/ui_core_views/cell_evaluation/position_set.ts
```diff
@@ -68,20 +62,6 @@ export class PositionSet {
     return true;
   }
 
-  fillAllPositions() {
-    this.insertions = new Array<CellPosition>(this.maxSize);
```
LucasLefevre, line 72: iterator is broken since `this.insertions` is empty

src/plugins/ui_core_views/cell_evaluation/binary_grid.ts
```diff
@@ -66,33 +70,25 @@ export class BinaryGrid extends Uint32Array {
...
   isEmpty() {
-    return !this.some((bucket) => bucket !== 0);
+    return !this.default && !Object.values(this.grid).some((v) => v !== 0);
   }
 
   fillAllPositions() {
-    const thirtyTwoOnes = -1 >>> 0; // same as 2 ** 32 - 1, a 32-bit number with all bits set to 1
-    this.fill(thirtyTwoOnes);
+    this.clear();
+    this.default = 1;
```
LucasLefevre, line 78: `fillAllPositions` isn't used anymore

src/plugins/ui_core_views/cell_evaluation/binary_grid.ts
```diff
@@ -1,7 +1,16 @@
+import { MAX_COL } from "../../../helpers";
 import { CellPosition } from "../../../types";
 
 type Bit = 0 | 1;
 
+const BLOCK_SIZE = 32;
```
LucasLefevre, line 6: Is 32 the number of positions or the number of integers (each holding 32 bits -> 32*32 positions)?
32 is really really small. I think we can go way up.

src/plugins/ui_core_views/cell_evaluation/binary_grid.ts
```diff
@@ -1,7 +1,16 @@
...
+const BLOCK_HEIGHT = 8;
+const BLOCK_WIDTH = Math.floor(BLOCK_SIZE / BLOCK_HEIGHT);
```
LucasLefevre, line 8: Later, I'd also want to investigate the shape of blocks and the order in which we iterate over those positions. The idea would be to always iterate sequentially on the typed array (to make it fast because the data would already be loaded in the CPU cache)

src/plugins/ui_core_views/cell_evaluation/binary_grid.ts
LucasLefevre, line 28: one interesting thing with `Map` though: it preserves insertion order, which means we no longer need `this.insertions`.
We need some performance measures :)

LucasLefevre (commented):
Preliminary comments on the new `PositionSet` implementation.
I really like the idea of using blocks.
We need to measure performance. It's ok if it's not super optimal as long as it doesn't hurt the current performance too much

PR #6914, opened 2025-08-11
[IMP] model: improve figure export data

hokolomopo:
## Description:

Currently, charts/images/carousels are exported inside `figure.data`
in the JSON. That's not ideal, `figure.data` then needs to be typed as
`any`, and the charts plugin need to be aware of carousels for its
import.

This commit changes the export, now each plugin has its own key in the
data, so it can be typed correctly and they are not interlinked.

Task: 5002886

Task: [5002886](https://www.odoo.com/odoo/2328/tasks/5002886)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/migrations/data.ts
+++ b/src/migrations/data.ts
@@ -275,13 +275,13 @@ function fixChartDefinitions(data: Partial<WorkbookData>, initialMessages: State
       if (
         figure.tag === "chart" &&
         data.version &&
-        compareVersions(String(data.version), "18.5.2") < 0
+        compareVersions(String(data.version), "18.5.11") < 0
       ) {
         map[figure.id] = (figure as any).data;
       }
     });
     for (const chartId in sheet.charts || {}) {
-      if (data.version && compareVersions(String(data.version), "18.5.2") >= 0) {
+      if (data.version && compareVersions(String(data.version), "18.5.11") >= 0) {
         map[chartId] = sheet.charts[chartId].chart;
       }
     }
--- a/src/migrations/data.ts
+++ b/src/migrations/data.ts
@@ -272,15 +272,19 @@ function fixChartDefinitions(data: Partial<WorkbookData>, initialMessages: State
   const map = {};
   for (const sheet of data.sheets || []) {
     sheet.figures?.forEach((figure) => {
-      if (figure.tag === "chart") {
-        // chart definition
-        if (data.version && compareVersions(String(data.version), "18.5.1") <= 0) {
-          map[figure.data.chartId] = figure.data;
-        } else {
-          map[figure.id] = figure.data;
-        }
+      if (
+        figure.tag === "chart" &&
+        data.version &&
+        compareVersions(String(data.version), "18.5.2") < 0
+      ) {
+        map[figure.id] = (figure as any).data;
       }
     });
+    for (const chartId in sheet.charts || {}) {
+      if (data.version && compareVersions(String(data.version), "18.5.2") >= 0) {
+        map[chartId] = sheet.charts[chartId].chart;
+      }
+    }
   }
   for (const message of initialMessages) {
     if (message.type === "REMOTE_REVISION") {
@@ -415,6 +419,9 @@ export function createEmptySheet(sheetId: UID, name: string): SheetData {
     conditionalFormats: [],
     dataValidationRules: [],
     figures: [],
+    charts: {},
+    carousels: {},
+    images: {},
     tables: [],
     isVisible: true,
   };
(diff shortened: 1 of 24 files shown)


src/migrations/data.ts
```diff
@@ -272,15 +272,19 @@ function fixChartDefinitions(data: Partial<WorkbookData>, initialMessages: State
...
-        // chart definition
-        if (data.version && compareVersions(String(data.version), "18.5.1") <= 0) {
-          map[figure.data.chartId] = figure.data;
-        } else {
-          map[figure.id] = figure.data;
-        }
+      if (
+        figure.tag === "chart" &&
+        data.version &&
+        compareVersions(String(data.version), "18.5.1") > 0
```
LucasLefevre: wouldn't it be better to call `repairInitialMessages` after the data is upgraded to the current version ?

src/xlsx/conversion/figure_conversion.ts
```diff
@@ -12,31 +12,50 @@ import {
...
 import { XLSXFigureAnchor } from "./../../types/xlsx";
 import { convertColor } from "./color_conversion";
 import { EXCEL_TO_SPREADSHEET_TRENDLINE_TYPE_MAPPING } from "./conversion_maps";
 
-export function convertFigures(sheetData: XLSXWorksheet): FigureData<any>[] {
+export function convertFigures(sheetData: XLSXWorksheet): {
+  figures: SheetData["figures"];
+  images: SheetData["images"];
+  charts: SheetData["charts"];
+} {
```
LucasLefevre: :socks: 
`: Pick<SheetData, "figures" | "images" | "charts">`

src/plugins/index.ts
```diff
@@ -77,8 +77,8 @@ export const corePluginRegistry = new Registry<CorePluginConstructor>()
   .add("borders", BordersPlugin)
   .add("conditional formatting", ConditionalFormatPlugin)
   .add("figures", FigurePlugin)
-  .add("chart", ChartPlugin)
   .add("carousel", CarouselPlugin)
+  .add("chart", ChartPlugin)
```
rrahir, line 81: I think this is wrong,  carousel should know about charts and not the other way around.

src/plugins/core/carousel.ts
```diff
@@ -122,19 +122,19 @@ export class CarouselPlugin extends CorePlugin<CarouselState> implements Carouse
...
     }
   }
 
   export(data: WorkbookData) {
     for (const sheet of data.sheets) {
-      const carousels = sheet.figures.filter((figure) => figure.tag === "carousel");
-      for (const carousel of carousels) {
-        if (this.carousels[sheet.id]?.[carousel.id]) {
-          carousel.data = { ...carousel.data, ...this.carousels[sheet.id]?.[carousel.id] };
+      for (const carouselId in this.carousels[sheet.id] || {}) {
```
LucasLefevre, line 134: Is it `carouselId` or `figureId`. `import` and `export` are not consistent. And I don't know which is which ?
```suggestion
      for (const figureId in this.carousels[sheet.id] || {}) {
```

src/xlsx/conversion/figure_conversion.ts
```diff
@@ -12,31 +12,50 @@ import {
...
-  return sheetData.figures
+  const figs = sheetData.figures
     .map((figure) => convertFigure(figure, (id++).toString(), sheetData))
     .filter(isDefined);
+
+  for (const { figure, chart, image } of figs) {
+    figures.push(figure);
+    if (chart) {
+      charts[figure.id] = { figureId: figure.id, chart };
+    } else if (image) {
```
LucasLefevre: There's unnecessary complexity that could be improved here. There are 2 *"if chart else if image ..."* (here and in `convertFigure`)  while the code could be reorganized to have only one.

src/migrations/data.ts
hokolomopo: No clue why it's implemented that way, there's probably a reason ? :shrug: @rrahir

src/types/workbook_data.ts
```diff
@@ -87,11 +68,11 @@ export interface ExcelWorkbookData extends WorkbookData {
...
 }
 
-export interface ExcelSheetData extends Omit<SheetData, "figureTables" | "cols" | "rows"> {
+export interface ExcelSheetData
+  extends Omit<SheetData, "figureTables" | "cols" | "rows" | "charts"> {
   cellValues: { [xc: string]: CellValue | undefined };
   formulaSpillRanges: { [xc: string]: string };
-  charts: FigureData<ExcelChartDefinition>[];
-  images: FigureData<Image>[];
+  charts: { [chartId: string]: { figureId: UID; chart: any } };
```
LucasLefevre, line 83: `any` ? :eyes:

src/migrations/migration_steps.ts
```diff
@@ -538,11 +538,24 @@ migrationStepRegistry
     },
   })
   .add("18.5.1", {
-    migrate(data: WorkbookData): any {
+    migrate(data: any): WorkbookData {
```
LucasLefevre: Could you add another step instead of changing this one ?
because dashboards in odoo need to go through this upgrade moulinette

LucasLefevre (commented):
There's also a small conflict

LucasLefevre (approved):
robodoo r+

PR #6923, opened 2025-08-12
[FIX] tests: Fix migration test

rrahir:
The test written for migration 7 (which removed forbidden characters from the sheet names to keep some compatibility with other spreadsheets) was poorly written. There were two types of mistakes:
- we defined `figure.type` instead of `figure.tag` in both the test and the script
- we defined formulas instead of ranges in both the chart and CF definition

The problem is two-fold: import data are not typed and we never checked the validity of the given data inside a working `Model` instance.

Task: 5008468

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/migrations/migration_steps.ts
+++ b/src/migrations/migration_steps.ts
@@ -102,7 +102,6 @@ migrationStepRegistry
           continue;
         }
         const oldName = sheet.name;
-        sanitizeSheetName;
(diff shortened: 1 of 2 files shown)


pro-odoo (commented):
robodoo r+

PR #6930, opened 2025-08-13
[FIX] evaluation: fix operation with empty matrices

anhe-odoo:
## Task Description

When using empty matrices for math operation (multiplication or inversion), as in the FORECAST formula, we got a traceback and the error message was not clear for the user. This commits aims to fix it by checking that the matrices used in theses operations are not empty and returning an understandable error if it's not the case.

## Related Task

- Task: [5001658](https://www.odoo.com/odoo/2328/tasks/5001658)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/functions/helper_matrices.ts
+++ b/src/functions/helper_matrices.ts
@@ -2,6 +2,9 @@ import { _t } from "../translation";
 import { Matrix, isMatrix } from "../types";
 
 export function getUnitMatrix(n: number): Matrix<number> {
+  if (n < 1) {
+    throw new Error(_t("Function [[FUNCTION_NAME]] unit matrix error, n should be >= 1"));
+  }
   const matrix: Matrix<number> = Array(n);
   for (let i = 0; i < n; i++) {
     matrix[i] = Array(n).fill(0);
@@ -29,6 +32,10 @@ export function invertMatrix(M: Matrix<number>): {
   // (a) Swap 2 rows. This multiply the determinant by -1.
   // (b) Multiply a row by a scalar. This multiply the determinant by that scalar.
   // (c) Add to a row a multiple of another row. This does not change the determinant.
+
+  if (M.length < 1 || M[0].length < 1) {
+    throw new Error(_t("Function [[FUNCTION_NAME]] invert matrix error, empty matrix"));
+  }
 
   if (M.length !== M[0].length) {
     throw new Error(
@@ -113,8 +120,11 @@ function swapMatrixRows(matrix: number[][], row1: number, row2: number) {
  * Note: we use indexing [col][row] instead of the standard mathematical notation [row][col]
  */
 export function multiplyMatrices(matrix1: Matrix<number>, matrix2: Matrix<number>): Matrix<number> {
+  if (matrix1.length < 1 || matrix2.length < 1) {
+    throw new Error(_t("Cannot multiply matrices: empty matrices cannot be multiplied."));
+  }
   if (matrix1.length !== matrix2[0].length) {
-    throw new Error(_t("Cannot multiply matrices : incompatible matrices size."));
+    throw new Error(_t("Cannot multiply matrices: incompatible matrices size."));
   }
 
   const rowsM1 = matrix1[0].length;
--- a/src/functions/module_array.ts
+++ b/src/functions/module_array.ts
@@ -382,6 +382,11 @@ export const MMULT = {
     const _matrix2 = toMatrix(matrix2);
 
     assert(
+      () => _matrix1.length > 0 && _matrix2.length > 0,
+      _t("In [[FUNCTION_NAME]], the first and second arguments must be non-empty matrices.")
+    );
+
+    assert(
       () => _matrix1.length === _matrix2[0].length,
       _t(
         "In [[FUNCTION_NAME]], the number of columns of the first matrix (%s) must be equal to the \
--- a/tests/functions/module_statistical.test.ts
+++ b/tests/functions/module_statistical.test.ts
@@ -1,4 +1,6 @@
 import { toXC } from "../../src/helpers";
+import { ErrorCell } from "../../src/types";
+import { setCellContent } from "../test_helpers/commands_helpers";
 import { getEvaluatedCell } from "../test_helpers/getters_helpers";
 import {
   createModelFromGrid,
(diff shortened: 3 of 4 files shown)


src/functions/helper_matrices.ts
```diff
@@ -30,6 +33,10 @@ export function invertMatrix(M: Matrix<number>): {
   // (b) Multiply a row by a scalar. This multiply the determinant by that scalar.
   // (c) Add to a row a multiple of another row. This does not change the determinant.
 
+  if (M.length < 1 || M[0].length < 1) {
+    throw new Error(_t("Function [[FUNCTION_NAME]] invert matrix error, empty matrix"));
```
LucasLefevre: Not super user-friendly
```suggestion
    throw new Error("An empty matrix cannot be inverted"));
```

src/functions/helper_matrices.ts
```diff
@@ -113,8 +120,11 @@ function swapMatrixRows(matrix: number[][], row1: number, row2: number) {
  * Note: we use indexing [col][row] instead of the standard mathematical notation [row][col]
  */
 export function multiplyMatrices(matrix1: Matrix<number>, matrix2: Matrix<number>): Matrix<number> {
+  if (matrix1.length < 1 || matrix2.length < 1) {
+    throw new Error(_t("Cannot multiply matrices: empty matrices cannot be multiplied."));
```
LucasLefevre:
```suggestion
    throw new Error(_t("Empty matrices cannot be multiplied."));
```

src/functions/helper_matrices.ts
```diff
@@ -113,8 +120,11 @@ function swapMatrixRows(matrix: number[][], row1: number, row2: number) {
...
+  }
   if (matrix1.length !== matrix2[0].length) {
-    throw new Error(_t("Cannot multiply matrices : incompatible matrices size."));
```
LucasLefevre, line 117: you cannot change a translated string is stable. It breaks all existing translations

src/functions/module_array.ts
```diff
@@ -381,6 +381,11 @@ export const MMULT = {
     const _matrix1 = toMatrix(matrix1);
     const _matrix2 = toMatrix(matrix2);
 
+    assert(
+      () => _matrix1.length > 0 && _matrix2.length > 0,
+      _t("In [[FUNCTION_NAME]], the first and second arguments must be non-empty matrices.")
```
LucasLefevre:
```suggestion
      _t("The first and second arguments of [[FUNCTION_NAME]] must be non-empty matrices.")
```

src/functions/helper_matrices.ts
```diff
@@ -2,6 +2,9 @@ import { _t } from "../translation";
 import { Matrix, isMatrix } from "../types";
 
 export function getUnitMatrix(n: number): Matrix<number> {
+  if (n < 1) {
+    throw new Error(_t("Function [[FUNCTION_NAME]] unit matrix error, n should be >= 1"));
+  }
```
LucasLefevre: if n < 1, is this a user error message or a programming error ?
It's already validated in `MUNIT`

src/functions/helper_matrices.ts
LucasLefevre: It is impossible for a user to provide an empty matrix.

tests/functions/module_statistical.test.ts
```diff
@@ -2853,6 +2855,23 @@ describe("MATTHEWS formula", () => {
...
+      "Cannot multiply matrices: empty matrices cannot be multiplied."
+    );
+  });
+
+  test("Slope with an empty matrix for the x values", () => {
+    const model = createModelFromGrid({ B2: "", B3: "", C2: "1", C3: "2" });
+    setCellContent(model, "A1", "=SLOPE(C2:C3, B2:B3)");
+    expect(getEvaluatedCell(model, "A1").value).toBe("#ERROR");
+    expect((getEvaluatedCell(model, "A1") as ErrorCell).error.message).toBe(
+      "Cannot multiply matrices: empty matrices cannot be multiplied."
```
LucasLefevre: I think the arguments should be validated in the `compute` with a proper error message telling which argument is faulty. Here the message is quite useless: it doesn't tell you which arg is wrong, and at first glance you're trying to compute the slope, but the error is mentioning matrix multiplication.

src/functions/helper_matrices.ts
LucasLefevre: user-facing errors should be `EvaluationError`
```suggestion
    throw new EvaluationError(_t("Function [[FUNCTION_NAME]] invert matrix error, empty matrix"));
```

anhe-odoo: It was, at first, as the `invertMatrix` function is called in other part of the code than the INVERT_MATRIX formula (it's also used in the getLMS method). But as we first try to multiply two matrices before trying to invert the result, indeed it may be not needed anymore.
Shouldn't we let this part "in case of" one day we use this method in other part of the code, or keeping thins as a safety guard is not a good idea ?

LucasLefevre: we can keep the check, but remove the translation (it's useless work for translators if the string is not user-facing)

src/functions/helper_statistical.ts
```diff
@@ -15,3 +16,16 @@ export function assertSameNumberOfElements(...args: any[][]) {
...
+export function assertNonEmptyMatrix(matrix: any[][], argName: string) {
+  assert(
+    () => matrix.length > 0 && matrix[0].length > 0,
+    _t("[[FUNCTION_NAME]] expects the provided values of %s to be a non-empty matrix.", argName)
+  );
+}
+
+export function assertNonEmpty(...data: any[][]) {
+  if (data.length === 0 || data.some((arg) => arg.length === 0)) {
+    throw new NotAvailableError(_t("[[FUNCTION_NAME]] has no valid input."));
```
rrahir: probably will need to be adapted in 18.4 following https://github.com/odoo/o-spreadsheet/pull/6248

rrahir:
```suggestion
    throw new NotAvailableError(_t("[[FUNCTION_NAME]] contains empty input data."));
```
only a part of the input is invalid in this case. 

That being said, I like the idea of the assert above that provides the name of the faulty argument to help out.
maybe something like 
```typescript
    throw new NotAvailableError(_t("[[FUNCTION_NAME]] expects the provided values of %(argName)s to contain values.", {argName}));
```

src/functions/helper_statistical.ts
```diff
@@ -15,3 +16,16 @@ export function assertSameNumberOfElements(...args: any[][]) {
     )
   );
 }
+
+export function assertNonEmptyMatrix(matrix: any[][], argName: string) {
+  assert(
+    () => matrix.length > 0 && matrix[0].length > 0,
+    _t("[[FUNCTION_NAME]] expects the provided values of %s to be a non-empty matrix.", argName)
```
rrahir:
```suggestion
    _t("[[FUNCTION_NAME]] expects the provided values of %(argName)s to be a non-empty matrix.", {argName})
```

rrahir (changes requested):
jus a bit of nitpick ^^ what do you think?

src/functions/helper_statistical.ts
anhe-odoo: Yes, I was aware of that :)

anhe-odoo: It won't be possible as the `filterAndFlatData` used before will return both `data_x` and `data_y` filtered so that we ignore every value of an array if the value at the same position of the other array is undefined or null. So if we have `data_X` OR `data_Y` empty, both will be filtered to empty array

rrahir:
robodoo r+

PR #6946, opened 2025-08-19
[IMP] PivotRegistry: Add `isPivotUnused` to extend the condition in Odoo

rrahir:
## Description:

Task: [4079971](https://www.odoo.com/odoo/2328/tasks/4079971)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/components/radio_selection/radio_selection.css
+++ b/src/components/side_panel/components/radio_selection/radio_selection.css
@@ -9,6 +9,7 @@
       border: 1px solid var(--os-gray-300);
       outline: none;
       border-radius: 8px;
+      flex-shrink: 0;
 
       &:checked {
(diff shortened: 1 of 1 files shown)


hokolomopo:
small rebase needed 🙂

packages/o-spreadsheet-engine/src/helpers/pivot/pivot_registry.ts
```diff
@@ -59,4 +62,6 @@ pivotRegistry.add("SPREADSHEET", {
   isMeasureCandidate: (field: PivotField) => field.type !== "boolean",
   isGroupable: () => true,
   canHaveCustomGroup: (field: PivotField) => field.type === "char" && !field.isCustomField,
+  isPivotUnused: (env: SpreadsheetChildEnv, pivotId: UID) =>
+    env.model.getters.isPivotUnused(pivotId),
```
LucasLefevre: If I want to develop something which requires `isPivotUnused`, it's not super clear which one should I use.
Wouldn't it be easier if `getters.isPivotUnused` was the main thing to use for business code (instead of using the registry). and `getters.isPivotUnused` would used the registry itself.

pro-odoo (approved):
robodoo rebase-ff r+

PR #6948, opened 2025-08-19
Master popout carousel vsc

VincentSchippefilt:
## Description:

Task: [5026408](https://www.odoo.com/odoo/2328/tasks/5026408)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/actions/figure_menu_actions.ts
+++ b/src/actions/figure_menu_actions.ts
@@ -143,6 +143,23 @@ export function getCarouselMenuActions(
       name: _t("Download chart"),
     },
     {
+      id: "popout_chart",
+      name: _t("Pop out chart"),
+      icon: "o-spreadsheet-Icon.EXTERNAL",
+      execute: () => {
+        const selectedItem = env.model.getters.getSelectedCarouselItem(figureId);
+        if (!selectedItem || selectedItem.type !== "chart") {
+          return;
+        }
+        env.model.dispatch("POPOUT_CHART_FROM_CAROUSEL", {
+          carouselId: figureId,
+          chartId: selectedItem.chartId,
+          sheetId: env.model.getters.getActiveSheetId(),
+        });
+      },
+      isVisible: isChartSelected,
+    },
+    {
       id: "delete_carousel_item",
       name: (env) => {
         const item = env.model.getters.getSelectedCarouselItem(figureId);
--- a/src/components/side_panel/carousel_panel/carousel_panel.ts
+++ b/src/components/side_panel/carousel_panel/carousel_panel.ts
@@ -1,4 +1,5 @@
 import { Component, onWillUpdateProps, useRef } from "@odoo/owl";
+import { ActionSpec } from "../../../actions/action";
 import { DEFAULT_CAROUSEL_TITLE_STYLE } from "../../../constants";
 import { deepEquals } from "../../../helpers";
 import { getCarouselItemPreview, getCarouselItemTitle } from "../../../helpers/carousel_helpers";
@@ -8,6 +9,7 @@ import { getBoundingRectAsPOJO } from "../../helpers/dom_helpers";
 import { useDragAndDropListItems } from "../../helpers/drag_and_drop_dom_items_hook";
 import { TextInput } from "../../text_input/text_input";
 import { TextStyler } from "../chart/building_blocks/text_styler/text_styler";
+import { CogWheelMenu } from "../components/cog_wheel_menu/cog_wheel_menu";
 import { Section } from "../components/section/section";
 
 interface Props {
@@ -18,7 +20,7 @@ interface Props {
 export class CarouselPanel extends Component<Props, SpreadsheetChildEnv> {
   static template = "o-spreadsheet-CarouselPanel";
   static props = { onCloseSidePanel: Function, figureId: String };
-  static components = { Section, TextInput, TextStyler };
+  static components = { Section, TextInput, TextStyler, CogWheelMenu };
 
   DEFAULT_CAROUSEL_TITLE_STYLE = DEFAULT_CAROUSEL_TITLE_STYLE;
 
@@ -58,7 +60,7 @@ export class CarouselPanel extends Component<Props, SpreadsheetChildEnv> {
   addNewChartToCarousel() {
     this.env.model.dispatch("ADD_NEW_CHART_TO_CAROUSEL", {
       figureId: this.props.figureId,
-      sheetId: this.env.model.getters.getActiveSheetId(),
+      sheetId: this.carouselSheetId,
     });
   }
 
@@ -74,7 +76,7 @@ export class CarouselPanel extends Component<Props, SpreadsheetChildEnv> {
   activateCarouselItem(item: CarouselItem) {
     this.env.model.dispatch("UPDATE_CAROUSEL_ACTIVE_ITEM", {
       figureId: this.props.figureId,
-      sheetId: this.env.model.getters.getActiveSheetId(),
+      sheetId: this.carouselSheetId,
       item,
     });
   }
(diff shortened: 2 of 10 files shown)


VincentSchippefilt:
robodoo rebase-ff

src/plugins/ui_stateful/carousel_ui.ts
```diff
@@ -142,6 +142,6 @@ export class CarouselUIPlugin extends UIPlugin {
   }
 
   private getCarouselItemId(item: CarouselItem): UID {
-    return item.type === "chart" ? item.chartId : "carouselDataView";
+    return item?.type === "chart" ? item?.chartId : "carouselDataView";
```
pro-odoo: I think it's the wrong fix, here you will have a `carouselDataView` when a carousel is empty (and so should be removed).
I think the right fix is to change the order of the first and second if starting L97:
```typescript
if (carousel.items.length === 0) {
  delete this.carouselStates[figureId];
} else if (!this.carouselStates[figureId]) {
  this.carouselStates[figureId] = this.getCarouselItemId(carousel.items[0]);
} else if (
  !carousel.items.some((item) => this.getCarouselItemId(item) === this.carouselStates[figureId])
) {
  this.carouselStates[figureId] = this.getCarouselItemId(carousel.items[0]);
}
```

src/components/side_panel/carousel_panel/carousel_panel.ts
```diff
@@ -118,6 +118,15 @@ export class CarouselPanel extends Component<Props, SpreadsheetChildEnv> {
     });
   }
 
+  popOutCarouselItem(item: CarouselItem, ev: MouseEvent) {
```
pro-odoo: `popOutCarouselItem(item: CarouselItem) {`

src/components/side_panel/carousel_panel/carousel_panel.xml
```diff
@@ -33,6 +33,12 @@
                 t-on-click.stop="(ev) => this.editCarouselItem(item, ev)"
                 title="Edit chart"
               />
+              <div
+                t-if="item.type === 'chart'"
+                class="o-popout-button o-button-icon pe-2 ps-1 flex-shrink-0 fa fa-external-link"
+                t-on-click.stop="(ev) => this.popOutCarouselItem(item, ev)"
```
pro-odoo: `t-on-click.stop="() => this.popOutCarouselItem(item)"`

src/plugins/ui_stateful/carousel_ui.ts
pro-odoo: You can remove this commit, it has been included in https://github.com/odoo/o-spreadsheet/pull/6934/commits

src/components/side_panel/carousel_panel/carousel_panel.ts
```diff
@@ -118,6 +118,15 @@ export class CarouselPanel extends Component<Props, SpreadsheetChildEnv> {
...
+    if (item.type !== "chart") return;
+    this.env.model.dispatch("POPOUT_CHART_FROM_CAROUSEL", {
+      sheetId: this.env.model.getters.getActiveSheetId(),
```
rrahir: All calls to activeSheetId here are wrong as we can change the active sheet while the sidepanel is open.

We could either:
- close the side panel if we change of activeSheet (update the `computeState` in the sidePanelRegistry)
- fetch the sheetId of the carousel and store it locally in the component

src/plugins/ui_stateful/carousel_ui.ts
```diff
@@ -41,6 +44,33 @@ export class CarouselUIPlugin extends UIPlugin {
...
+    }
+    this.dispatch("CREATE_CHART", {
+      chartId: this.uuidGenerator.smallUuid(),
+      figureId: this.uuidGenerator.smallUuid(),
+      sheetId,
+      col: 0,
+      row: figure.row,
+      offset: { x: 0, y: 0 },
+      size: { width: figure.width, height: figure.height },
+      definition: { ...chartDefinition },
```
rrahir, line 101: we could maybe add the carousel item title as the chart title

rrahir (changes requested):

src/actions/figure_menu_actions.ts
```diff
@@ -129,7 +129,33 @@ export function getCarouselMenuActions(
...
+        });
+      },
+      isEnabled: isChartSelected,
+    },
+    {
+      ...getDeleteMenuItem(figureId, onFigureDeleted, env),
+      isVisible: (env) => {
+        // only visible if thre are no charts in the carousel
+        const carouselItems = env.model.getters.getCarousel(figureId)?.items;
+        return carouselItems?.length === 0 || !carouselItems?.some((item) => item.type === "chart");
```
anhe-odoo: Shouldn't we have the "delete button" available when there is one or less item ? This would simplify the condition (even if it's a nitpick). And I don't know if it's an issue, but we can still delete the carousel with the "Delete" key. But I understand the hiding of this delete button is just an aesthetic thing.

anhe-odoo (commented):
Only a small comment but LGTM. Don't forget to address PRO and RAR's comments :)

src/plugins/ui_stateful/carousel_ui.ts
VincentSchippefilt: the title is still in flux so I don't want to do anything about it

src/actions/figure_menu_actions.ts
VincentSchippefilt: IMHO it's easier to understand that you don't have the delete button while there is a graph, but we can change that easily later if we want :)

src/actions/figure_menu_actions.ts
```diff
@@ -129,7 +129,33 @@ export function getCarouselMenuActions(
...
+        if (!selectedItem || selectedItem.type !== "chart") {
+          return;
+        }
+        env.model.dispatch("POPOUT_CHART_FROM_CAROUSEL", {
+          carouselId: figureId,
+          chartId: selectedItem.chartId,
+          sheetId: env.model.getters.getActiveSheetId(),
+        });
+      },
+      isEnabled: isChartSelected,
```
hokolomopo: Other menu items use `isVisible`. I think we should stay consistent :)

src/components/side_panel/carousel_panel/carousel_panel.ts
```diff
@@ -46,7 +47,7 @@ export class CarouselPanel extends Component<Props, SpreadsheetChildEnv> {
   addNewChartToCarousel() {
     this.env.model.dispatch("ADD_NEW_CHART_TO_CAROUSEL", {
       figureId: this.props.figureId,
-      sheetId: this.env.model.getters.getActiveSheetId(),
+      sheetId: this.props.activeSheetId,
```
hokolomopo: we could just use `getters.getFigureSheetId()` instead of adding an additional props, no ? And if we want to keep the props, `activeSheetId` is probably not the correct name, as it is not always the active sheet Id :upside_down_face:

src/components/side_panel/carousel_panel/carousel_panel.xml
```diff
@@ -34,11 +34,11 @@
                 title="Edit chart"
               />
               <div
-                class="o-delete-button o-button-icon pe-2 ps-1 flex-shrink-0"
-                t-on-click.stop="(ev) => this.deleteCarouselItem(item, ev)"
-                title="Remove chart">
-                <t t-call="o-spreadsheet-Icon.TRASH_FILLED"/>
-              </div>
+                t-if="item.type === 'chart'"
```
hokolomopo: See my comment in the Odoo pad. What about data view ?

hokolomopo (commented):
:wave:

pro-odoo:
robodoo r+

PR #6993, opened 2025-08-26
[FIX] menu_items: fix sequence calculation for pivot data sources

pro-odoo:
Depending on the number of pivot tables, the pivot items could be displayed after other items (e.g. Re-insert dynamic pivot) in the topbar menu.

Practically, the number of pivots to observe the issue is not realistic, as it requires more than 970 pivots.
But the issue can be observed easily with 25 lists in Odoo.

To fix the issue, we need to compute the sequence with decimal numbers, to ensure that the pivot items are always between the first sequence (50) and sequence + 1 (51).

Task: 5025230

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/actions/data_actions.ts
+++ b/src/actions/data_actions.ts
@@ -72,7 +72,7 @@ export const splitToColumns: ActionSpec = {
 export const reinsertDynamicPivotMenu: ActionSpec = {
   id: "reinsert_dynamic_pivot",
   name: _t("Re-insert dynamic pivot"),
-  sequence: 1020,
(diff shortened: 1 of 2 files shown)


pro-odoo:
Don't know how to add a relevant test, I think testing the sequence number is not useful at all.

rrahir:
robodoo r?

rrahir:
robodoo r+

PR #7008, opened 2025-08-28
[FIX] xlsx: `=undefined` when importing array formula

dhrp-odoo:
## Description:

Steps to reproduce:

* Write an array formula =RANDARRAY(20) in Excel.
* Import the file into `o-spreadsheet`.

Current behavior:

* The anchor cell shows #SPILL!.
* Spill cells are populated with =undefined, causing #BAD_EXPR and spill errors.

Desired behavior:

* Only the anchor retains the formula.
* Spill cells are empty (no phantom content), so the array spills correctly without #SPILL!.

Task: [4812508](https://www.odoo.com/odoo/2328/tasks/4812508)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/xlsx/extraction/sheet_extractor.ts
+++ b/src/xlsx/extraction/sheet_extractor.ts
@@ -1,3 +1,4 @@
+import { positions, toXC, toZone } from "../../helpers";
 import {
   XLSXCell,
   XLSXColumn,
@@ -245,12 +246,13 @@ export class XlsxSheetExtractor extends XlsxBaseExtractor {
   }
 
   private extractRows(worksheet: Element): XLSXRow[] {
+    const spilledCells = new Set<string>();
     return this.mapOnElements(
       { parent: worksheet, query: "sheetData row" },
       (rowElement): XLSXRow => {
         return {
           index: this.extractAttr(rowElement, "r", { required: true })?.asNum()!,
-          cells: this.extractCells(rowElement),
+          cells: this.extractCells(rowElement, spilledCells),
           height: this.extractAttr(rowElement, "ht")?.asNum(),
           customHeight: this.extractAttr(rowElement, "customHeight")?.asBool(),
           hidden: this.extractAttr(rowElement, "hidden")?.asBool(),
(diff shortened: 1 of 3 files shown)


src/xlsx/extraction/sheet_extractor.ts
```diff
@@ -44,6 +48,7 @@ export class XlsxSheetExtractor extends XlsxBaseExtractor {
       { query: "worksheet", parent: this.rootFile.file.xml },
       (sheetElement): XLSXWorksheet => {
         const sheetWorkbookInfo = this.getSheetWorkbookInfo();
+        this.arrayFormulaSpillZones = [];
```
hokolomopo: this.arrayFormulaSpillZones could be a local variable inside `extractCells` (and be passed as argument to `isCellInArraySpillZones`). It would avoid having to worry about resetting the variable for each sheet.

src/xlsx/extraction/sheet_extractor.ts
```diff
@@ -262,26 +267,49 @@ export class XlsxSheetExtractor extends XlsxBaseExtractor {
...
+    return { content, sharedIndex, ref };
+  }
+
+  private isCellInArraySpillZones(col: number, row: number, xc: string): boolean {
+    return this.arrayFormulaSpillZones.some((zone) => {
+      if (xc === zone.anchorXc) {
+        return false;
+      }
+      return isInside(col, row, zone);
+    });
```
hokolomopo: performance-wise it could cost a lot if there are a lot of cells & array formulas. Maybe when you find an array formula, you should put all of the spreaded cells XC's inside a set ? Would probably be faster

src/xlsx/extraction/sheet_extractor.ts
```diff
@@ -262,26 +267,49 @@ export class XlsxSheetExtractor extends XlsxBaseExtractor {
...
-      content: this.extractTextContent(formulaElement),
-      sharedIndex: this.extractAttr(formulaElement, "si")?.asNum(),
-      ref: this.extractAttr(formulaElement, "ref")?.asString(),
-    };
+    const content = this.extractTextContent(formulaElement);
+    const sharedIndex = this.extractAttr(formulaElement, "si")?.asNum();
+    const ref = this.extractAttr(formulaElement, "ref")?.asString();
+
+    if ((content === undefined || content.trim() === "") && sharedIndex === undefined) {
+      return undefined;
```
hokolomopo, line 302: Mayeb a comment/docstring that says this happen for spilled cells

src/xlsx/extraction/sheet_extractor.ts
dhrp-odoo: Agreed, Changes made :)

src/xlsx/extraction/sheet_extractor.ts
```diff
@@ -245,12 +246,13 @@ export class XlsxSheetExtractor extends XlsxBaseExtractor {
   }
 
   private extractRows(worksheet: Element): XLSXRow[] {
+    const spillFollowers = new Set<string>();
```
hokolomopo: nitpick: not sure about `follower`. `spilledCells` maybe ?

src/xlsx/extraction/sheet_extractor.ts
```diff
@@ -260,28 +262,46 @@ export class XlsxSheetExtractor extends XlsxBaseExtractor {
...
-      content: this.extractTextContent(formulaElement),
-      sharedIndex: this.extractAttr(formulaElement, "si")?.asNum(),
-      ref: this.extractAttr(formulaElement, "ref")?.asString(),
-    };
+    const content = this.extractTextContent(formulaElement);
+    const sharedIndex = this.extractAttr(formulaElement, "si")?.asNum();
+    const ref = this.extractAttr(formulaElement, "ref")?.asString();
+
+    // This is the case of spilled cells of array formulas where <f> is empty
+    if ((content === undefined || content.trim() === "") && sharedIndex === undefined) {
```
rrahir, line 301: in which case can this occur?

dhrp-odoo: It happens for **spill dependents** of dynamic/array formulas. Excel writes followers like this:

image

`<f>` has no text - content is undefined. The anchor is the only one with `<f t="array" ref="…">RANDARRAY(…)</f>`.

So this check identifies these follower cells and treats them as non-formulas, preventing
`=undefined` and letting the spill come from the anchor.

(Is the comment/docstring above the if statement still unclear?)

dhrp-odoo: And it may appear to work even without this condition.
But I think, `extractCellFormula` should return `undefined` when the cell is not itself a real formula. What do you think?

rrahir: my bad, I was trying to understand for what reason we'd have and empty `f` tag as I was testing the function `MUNIT` which does not propagate the same way as `RANDARRAY` for some reason

rrahir:
thanks for the fix - I agree that `extractCellFormula` should plainly ignore those empty formulas, they hold no relevant meaning.

robodoo r+

PR #7009, opened 2025-08-28
[IMP] demo: add new data generator

VincentSchippefilt:
add a generator that creates a sheet with problematic split and vlookup as mentionned in this task https://www.odoo.com/odoo/project/2328/tasks/4954710

Task: 0


VincentSchippefilt:
robodoo r+

PR #7027, opened 2025-09-02
[IMP] formulas: fit repeating args

laa-odoo:
Allow to format repeating args by group

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


packages/o-spreadsheet-engine/src/formulas/formula_formatter.ts
```diff
@@ -240,12 +242,26 @@ function astToDoc(ast: AST): Doc {
       return ast.value;
 
     case "FUNCALL":
-      const docs = ast.args.map(astToDoc);
-      return wrapInParentheses(
-        concat(docs.map((doc, i) => (i < 1 ? doc : concat([", ", line(), doc])))),
-        ast.value
-      );
-
+      const functionDescription = functionRegistry.get(ast.value.toUpperCase());
```
LucasLefevre, line 246: où sont les tests ?

tests/functions/module_math.test.ts
```diff
@@ -2660,7 +2660,7 @@ describe("SUM formula", () => {
     expect(evaluateCell("A1", { A1: "=SUM()" })).toBe("#BAD_EXPR"); // @compatibility: on google sheets, return #N/A
     expect(evaluateCell("A1", { A1: "=SUM(,)" })).toBe(0);
     expect(evaluateCell("A1", { A1: "=SUM(0)" })).toBe(0);
-    expect(evaluateCell("A1", { A1: "=SUM(1, 2, 3, 1, 2)" })).toBe(9);
+    expect(evaluateCell("A1", { A1: "=?SUM(1, 2, 3, 1, 2)" })).toBe(9);
```
LucasLefevre: remove `?`

src/functions/module_math.ts
```diff
@@ -420,12 +420,8 @@ export const COUNTBLANK = {
...
   args: [
     arg(
-      "value1 (any, range)",
-      _t("The first value or range in which to count the number of blanks.")
-    ),
-    arg(
-      "value2 (any, range, repeating)",
-      _t("Additional values or ranges in which to count the number of blanks.")
+      "value (any, range, repeating)",
+      _t("Values or ranges in which to count the number of blanks.")
```
hokolomopo: Strange to have plural, when in the tooltip it show only one value (value). Either it should be in singular form, or we should show something like `[optional, repeating]`, in which case we can just display `value` instead of `valueX`. I'd say the second option make a bit more sense

image

src/components/composer/formula_assistant/formula_assistant.xml
```diff
@@ -4,22 +4,77 @@
...
+                    </span>
+                  </span>
+                  <span t-if="displayBrackets and (!nextArg or !nextArg.repeating)">]</span>
                 </span>
               </span>
+
+              <span t-if="arg.repeating" style="order: 1;">
+                <span
+                  t-if="arg_index > '0' or firstRepeatableGroupNotOptional"
+                  t-esc="formulaArgSeparator"
```
hokolomopo: Is there a way to force the comma to be on the same line as the first argument ? Here in my example, comma should be before the newline

image

src/formulas/formula_formatter.ts
```diff
@@ -240,12 +242,26 @@ function astToDoc(ast: AST): Doc {
...
+          const repeatingArgSeries = ast.args.slice(i, i + functionDescription.nbrArgRepeating);
+          const docsSeries = repeatingArgSeries.map((arg) => astToDoc(arg));
+          docs.push(
+            group(concat(docsSeries.flatMap((doc, j) => (j === 0 ? [doc] : [", ", line(), doc]))))
+          );
+          i += functionDescription.nbrArgRepeating;
+          continue;
+        }
+        docs.push(astToDoc(ast.args[i]));
+        i++;
```
LucasLefevre: IMO it's easier to read with a `else` rather than a `continue` in this case
```suggestion
        } else {
          docs.push(astToDoc(ast.args[i]));
          i++;
        }
```

src/formulas/formula_formatter.ts
```diff
@@ -240,12 +242,26 @@ function astToDoc(ast: AST): Doc {
...
-        ast.value
-      );
-
+      const functionDescription = functionRegistry.get(ast.value.toUpperCase());
+      const getArgToFocus = argTargeting(functionDescription, ast.args.length);
+      const docs: Doc[] = ast.args.length ? [astToDoc(ast.args[0])] : [];
+      let i = 1;
+      while (i < ast.args.length) {
+        docs.push(", ", line());
+        const isRepeating = functionDescription.args[getArgToFocus(i).index || -1]?.repeating;
```
LucasLefevre: `||` or `??` ? is it expected that it falls back to -1 if the index is zero ?

src/types/functions.ts
```diff
@@ -53,7 +53,7 @@ export type FunctionDescription = AddFunctionDescription & {
   minArgRequired: number;
   maxArgPossible: number;
   nbrArgRepeating: number;
-  nbrArgOptional: number;
+  nbrArgOptionalThatIsNotRepeating: number;
```
LucasLefevre: `optionalNonRepeatingArgCount` or `optionalFixedArgCount` or `nbrOptionalNonRepeatingArgs` or `nbrOptionalFixedArgs` ?
(and maybe rename `nbrArgRepeating` to `repeatingArgCount` if we go with the "count" alternative)

src/components/composer/formula_assistant/formula_assistant.xml
```diff
@@ -4,22 +4,77 @@
       <t t-set="context" t-value="getContext()"/>
       <div class="o-formula-assistant" t-if="context.functionDescription.name">
         <div class="o-formula-assistant-head d-flex flex-row justify-content-between">
-          <div>
-            <span t-esc="context.functionDescription.name"/>
-            (
+          <div
+            style="display: flex; flex-direction: row; flex-wrap: wrap; align-items: center; word-break: break-word;">
```
LucasLefevre: all this style should be bootstrap classes; and if there are no corresponding class, then it should be in a `scss` file, not in-lined

src/components/composer/formula_assistant/formula_assistant.xml
```diff
@@ -4,22 +4,77 @@
...
+              (
+            </span>
+
             <t t-foreach="context.functionDescription.args" t-as="arg" t-key="arg.name">
-              <span t-if="arg_index > '0'" t-esc="formulaArgSeparator"/>
-              <span
-                t-att-class="{ 'o-formula-assistant-focus': context.argsToFocus.includes(arg_index) }">
+              <t t-set="prevArg" t-value="context.functionDescription.args[arg_index - 1]"/>
+              <t t-set="nextArg" t-value="context.functionDescription.args[arg_index + 1]"/>
+              <t t-if="arg.repeating">
```
LucasLefevre: all this looks like a lot of business logic for a xml template. Could you extract all in the js part ?
You can keep the `order` strategy, but computed in the js part.

src/components/composer/formula_assistant/formula_assistant.xml
```diff
@@ -4,22 +4,77 @@
...
+                      <span t-esc="arg.name"/>
+                    </span>
+                  </span>
+                  <span t-if="displayBrackets and (!nextArg or !nextArg.repeating)">]</span>
                 </span>
               </span>
+
+              <span t-if="arg.repeating" style="order: 1;">
+                <span
+                  t-if="arg_index > '0' or firstRepeatableGroupNotOptional"
```
LucasLefevre: `'0'` => `0` ?

src/formulas/formula_formatter.ts
laa-odoo: It's an error,
but that never happens before this commits because we never had a repeating argument at index 0

src/components/composer/composer/composer.ts
```diff
@@ -776,7 +776,8 @@ export class Composer extends Component<CellComposerProps, SpreadsheetChildEnv>
     nbrArgSupplied: number,
     argPosition: number
   ): number[] {
-    const { nbrArgRepeating, minArgRequired, nbrArgOptional, maxArgPossible } = description;
+    const { nbrArgRepeating, minArgRequired, nbrOptionalNonRepeatingArgs, maxArgPossible } =
```
rrahir, line 814: there is a typo in the commit msg ^^ it's pattern

src/components/composer/formula_assistant/formula_assistant.xml
```diff
@@ -37,22 +28,29 @@
...
-                }">
-              <div>
-                <span t-esc="arg.name"/>
+              t-att-class="{'o-formula-assistant-gray': context.argsToFocus.length > 0}">
+              <div
+                t-att-class="{'o-formula-assistant-focus': context.argsToFocus.includes(arg_index)}">
+                <span t-if="arg.repeating">
+                  <span t-esc="arg.name + (context.repeatingArgGroupIndex + 1)"/>
+                </span>
+                <span t-if="!arg.repeating">
```
rrahir: t-else

src/formulas/formula_formatter.ts
```diff
@@ -240,12 +242,26 @@ function astToDoc(ast: AST): Doc {
...
-        concat(docs.map((doc, i) => (i < 1 ? doc : concat([", ", line(), doc])))),
-        ast.value
-      );
-
+      const functionDescription = functionRegistry.get(ast.value.toUpperCase());
+      const getArgToFocus = argTargeting(functionDescription, ast.args.length);
+      const docs: Doc[] = ast.args.length ? [astToDoc(ast.args[0])] : [];
+      let i = 1;
+      while (i < ast.args.length) {
+        docs.push(", ", line());
```
LucasLefevre: does this implies that the first arg is never part of a repeating group ?

laa-odoo: Well seen! This has been corrected.

PR #7032, opened 2025-09-02
[IMP] carousel: improve  drag & drop of chart over carousel

hokolomopo:
## Description

We can drag & drop a chart over a carousel to add it to the carousel.

The current behaviour is not the best:
- even if the chart barely overlaps the carousel, it will still be added to it
- we merge the chart with the first overlapping carousel we find, not the one with the most overlap

This commit fixes the issues by using the distance between the centers of the two figures to check if they should be merged, and which carousel to choose.

Task: [5059979](https://www.odoo.com/odoo/2328/tasks/5059979)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/figures/figure_container/figure_container.ts
+++ b/src/components/figures/figure_container/figure_container.ts
@@ -559,21 +559,34 @@ export class FiguresContainer extends Component<Props, SpreadsheetChildEnv> {
     if (figureUI.tag !== "chart") {
       return undefined;
     }
-    const minimumOverlap = 20; // Minimum overlap in pixels to consider a carousel overlapping
-    const carousels = otherFigures.filter((f) => f.tag === "carousel");
 
-    return carousels.find((carousel) => {
-      const xOverlap = Math.max(
-        0,
-        Math.min(figureUI.x + figureUI.width, carousel.x + carousel.width) -
-          Math.max(figureUI.x, carousel.x)
-      );
-      const yOverlap = Math.max(
-        0,
-        Math.min(figureUI.y + figureUI.height, carousel.y + carousel.height) -
-          Math.max(figureUI.y, carousel.y)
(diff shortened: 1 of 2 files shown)


src/components/figures/figure_container/figure_container.ts
```diff
@@ -559,21 +559,33 @@ export class FiguresContainer extends Component<Props, SpreadsheetChildEnv> {
...
-    });
+
+    let bestMatch: FigureUI | undefined;
+    let smallestDistance = Infinity;
+
+    for (const figure of otherFigures) {
+      if (figure.tag !== "carousel") {
+        continue;
+      }
+      const figureCenterX = figureUI.x + figureUI.width / 2;
```
anhe-odoo: These two lines can be put outside of the for loop, can't they ?

src/components/figures/figure_container/figure_container.ts
```diff
@@ -559,21 +559,33 @@ export class FiguresContainer extends Component<Props, SpreadsheetChildEnv> {
...
+      const figureCenterY = figureUI.y + figureUI.height / 2;
+      const carouselCenterX = figure.x + figure.width / 2;
+      const carouselCenterY = figure.y + figure.height / 2;
+
+      const distanceX = Math.abs(figureCenterX - carouselCenterX);
+      const distanceY = Math.abs(figureCenterY - carouselCenterY);
+      const distance = Math.sqrt(distanceX ** 2 + distanceY ** 2);
```
anhe-odoo: Nitpick : We don't need to sqrt the distance as we only want to compare positive numbers :)

pro-odoo (approved):
robodoo r+

PR #7052, opened 2025-09-03
[FIX] charts: prevent unnecessary treemap animations

hokolomopo:
## Description

Charts animations are played every time the chart data changes in dashboards. But the treemap data contains callback, which mess up the `deepEqual` we do to check if the data changed.

This adds an argument to `deepEqual` to ignore functions.

Task: [5003595](https://www.odoo.com/odoo/2328/tasks/5003595)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


hokolomopo:
Technically the fix should be in 18.3, but the feature was in odoo then. I can create a PR spreadsheet 18.3 + a PR odoo 18.3 if people think it's worth

rrahir (approved):

src/helpers/misc.ts
```diff
@@ -389,7 +389,7 @@ export function getAddHeaderStartIndex(position: "before" | "after", base: numbe
 /**
  * Compares two objects.
  */
-export function deepEquals(o1: any, o2: any): boolean {
+export function deepEquals(o1: any, o2: any, ignoreFunctions?: "ignoreFunctions"): boolean {
```
LucasLefevre: I must admit I don't like the fix.
Ignoring function is wrong and error prone. It might be correct in this very specific use-case, but as a general helper, it can't guarantee two objects are equal if we ignore functions, and this hack looks like something to workaround another design problem

hokolomopo: I mean it's as you said, it's a parameter that must only be enabled if we need it for the specific use case we're working with. It's the dev responsibility to know if he needs it for his use case.

What alternative do you propose ?

rrahir:
robodoo r+

PR #7069, opened 2025-09-05
[FIX] charts: small zoom is disabled

anhe-odoo:
## Task Description:

This task aims to reduce issue coming from the zoom feature in two cases :
1. Make the slider looks "disabled" when there is only one data point
2. Don't zoom when the new min and max value are equals

## Related task

- Task: [5058567](https://www.odoo.com/odoo/2328/tasks/5058567)

### DIFF
--- a/src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs.ts
+++ b/src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs.ts
@@ -11,7 +11,7 @@ import { ChartJSRuntime } from "../../../../../types";
 import { css } from "../../../../helpers";
 import { chartJsExtensionRegistry } from "../chart_js_extension";
 import { ChartJsComponent } from "../chartjs";
-import { ZoomableChartStore } from "./zoomable_chart_store";
+import { Boundaries, ZoomableChartStore } from "./zoomable_chart_store";
 import { zoomWindowPlugin } from "./zoomable_chartjs_plugins";
 
 css/* scss */ `
@@ -38,8 +38,9 @@ export class ZoomableChartJsComponent extends ChartJsComponent {
   private hasLinearScale?: boolean;
   private isBarChart?: boolean;
   private chartId: string = "";
-  private datasetBoundaries: { xMin: number; xMax: number } = { xMin: 0, xMax: 0 };
+  private datasetBoundaries: Boundaries = { min: 0, max: 0 };
   private removeEventListeners = () => {};
+  private isMasterChartAllowed: boolean = false;
 
   setup() {
     this.store = useStore(ZoomableChartStore);
@@ -59,12 +60,20 @@ export class ZoomableChartJsComponent extends ChartJsComponent {
     `;
   }
 
+  get masterChartContainerStyle() {
+    const runtime = this.env.model.getters.getChartRuntime(this.props.chartId) as ChartJSRuntime;
+    if (runtime && !runtime.chartJsConfig.data.datasets.some((ds) => ds.data.length > 1)) {
+      return "opacity: 0.3;";
+    }
+    return "";
+  }
+
   get sliceable(): boolean {
     if (this.props.isFullScreen) {
       return true;
     }
     const definition = this.env.model.getters.getChartDefinition(this.props.chartId);
-    return ("zoomable" in definition && definition?.zoomable) ?? false;
+    return ("zoomable" in definition && definition.zoomable) ?? false;
   }
 
   get axisOffset(): number {
@@ -92,15 +101,13 @@ export class ZoomableChartJsComponent extends ChartJsComponent {
     if (!this.sliceable) {
       return chartData;
     }
-    const xAxis = this.store.currentAxesLimits[this.chartId]?.x;
-    const xScale = {
-      ...chartData.options.scales?.x,
-    };
-    if (xAxis?.min !== undefined) {
-      xScale.min = this.hasLinearScale ? xAxis.min : Math.ceil(xAxis.min) - this.axisOffset;
-    }
-    if (xAxis?.max !== undefined) {
-      xScale.max = this.hasLinearScale ? xAxis.max : Math.floor(xAxis.max) - this.axisOffset;
+    let x = chartData.options.scales.x;
+    const limits = this.store.currentAxesLimits[this.chartId]?.x;
+    if (limits) {
+      x = {
+        ...x,
+        ...this.getStoredBoundaries(),
+      };
     }
     return {
       ...chartData,
@@ -108,7 +115,7 @@ export class ZoomableChartJsComponent extends ChartJsComponent {
         ...chartData.options,
         scales: {
           ...chartData.options.scales,
-          x: xScale,
+          x,
         },
         layout: {
           ...chartData.options.layout,
@@ -121,15 +128,23 @@ export class ZoomableChartJsComponent extends ChartJsComponent {
     };
   }
 
-  private getAxisLimitsFromDataset(chartData: ChartConfiguration<any>): {
-    xMin: number;
-    xMax: number;
-  } {
+  private getAxisLimitsFromDataset(chartData: ChartConfiguration<any>): Boundaries {
     const data = chartData.data.datasets.map((ds) => ds.data).flat();
     const xValues = data.map((d, i) => (typeof d === "object" && d !== null ? d.x : i));
-    const xMin = Math.min(...xValues);
-    const xMax = Math.max(...xValues);
-    return { xMin, xMax };
+    const min = Math.min(...xValues);
+    const max = Math.max(...xValues);
+    return { min, max };
+  }
+
+  private setMasterChartCursor(runtime: ChartJSRuntime) {
+    const masterElement = this.masterChartCanvas?.el as HTMLCanvasElement;
+    if (runtime && !runtime.chartJsConfig.data.datasets.some((ds) => ds.data.length > 1)) {
+      masterElement.style.cursor = "not-allowed";
+      this.isMasterChartAllowed = false;
+      return;
+    }
+    masterElement.style.cursor = "default";
+    this.isMasterChartAllowed = true;
   }
 
   protected createChart(chartRuntime: ChartJSRuntime) {
@@ -144,14 +159,16 @@ export class ZoomableChartJsComponent extends ChartJsComponent {
     }
 
     super.createChart(chartRuntime);
-    this.hasLinearScale = this.chart?.scales?.x.type === "linear";
+    this.hasLinearScale = this.chart?.scales?.x?.type === "linear";
     if (!this.sliceable || !("masterChartConfig" in chartRuntime)) {
+      this.isMasterChartAllowed = false;
       return;
     }
 
     this.masterChart?.destroy();
-    const masterChartCtx = (this.masterChartCanvas!.el as HTMLCanvasElement).getContext("2d")!;
+    const masterChartCtx = (this.masterChartCanvas?.el as HTMLCanvasElement).getContext("2d")!;
 
+    this.setMasterChartCursor(chartRuntime);
     this.masterChart = new window.Chart(
       masterChartCtx,
       this.getMasterChartConfiguration(chartRuntime["masterChartConfig"] as ChartConfiguration<any>)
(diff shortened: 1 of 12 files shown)


src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs.ts
```diff
@@ -295,18 +292,36 @@ export class ZoomableChartJsComponent extends ChartJsComponent {
...
+  private computeMinMaxFromStore(): { min: number | undefined; max: number | undefined } {
+    let { min, max } = this.store.currentAxesLimits[this.chartId]?.x ?? {};
+    if (min !== undefined && max !== undefined && !this.hasLinearScale) {
+      min = Math.ceil(min) - this.axisOffset;
+      max = Math.floor(max) - this.axisOffset;
+    }
+    return { min, max };
+  }
+
+  private computeMinMaxFromScale({ min, max }: Boundaries): Boundaries {
```
hokolomopo: maybe some docstring explaining what the difference beween the two are. Why is one adding axisOffset to the max and the other substracting it ?

tests/figures/chart/zoomable_charts/zoomable_charts_component.test.ts
```diff
@@ -268,19 +268,6 @@ describe("zoom", () => {
       expect(newMin).toEqual(1);
       expect(newMax).toEqual(2);
     });
-
-    test("Boundaries are rounded for categorical axis", async () => {
```
hokolomopo, line 269: removed tests ? if wanted, explain the reasoning in the commit msg

tests/figures/chart/charts_zoom.test.ts
```diff
@@ -436,4 +410,19 @@ describe("zoom", () => {
       expect(newMax).toBeUndefined();
     });
   });
+
+  test("Chart with one point is not shown as disabled", async () => {
```
hokolomopo:
```suggestion
  test("Chart with one point shows timeline as disabled", async () => {
```
no ?

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs.xml
rrahir: for consistency, we should use pointerX events everywhere and not sometimes mouse,sometimes pointer

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs.ts
```diff
@@ -295,18 +292,53 @@ export class ZoomableChartJsComponent extends ChartJsComponent {
...
+  private computeMinMaxFromStore(): { min: number | undefined; max: number | undefined } {
+    let { min, max } = this.store.currentAxesLimits[this.chartId]?.x ?? {};
+    if (min !== undefined && max !== undefined && !this.hasLinearScale) {
+      min = Math.ceil(min);
+      max = Math.floor(max);
+    }
+    return { min, max };
+  }
+
+  private computeMinMaxFromScale({ min, max }: Boundaries): Boundaries {
```
rrahir:
```suggestion
  private adjustBoundariesForMasterChart({ min, max }: Boundaries): Boundaries {
```
Could you place the docstrings above each concerned method?

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs.ts
```diff
@@ -295,18 +292,53 @@ export class ZoomableChartJsComponent extends ChartJsComponent {
...
+   * add it to the max, because when computing from the scale, chartJs use integer values as
+   * the limits for non linear scales. If we have a min value of 1, it means we want to start
+   * displaying from 0.5, and if we have a max value of 4, it means we want to display until 4.5.
+   */
+
+  private computeMinMaxFromStore(): { min: number | undefined; max: number | undefined } {
+    let { min, max } = this.store.currentAxesLimits[this.chartId]?.x ?? {};
+    if (min !== undefined && max !== undefined && !this.hasLinearScale) {
+      min = Math.ceil(min);
+      max = Math.floor(max);
```
rrahir, line 325: you can end up in situations where max<min, how does it work then?
```suggestion
      max = Math.max(Math.floor(max),min);
```

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs.ts
```diff
@@ -295,18 +292,53 @@ export class ZoomableChartJsComponent extends ChartJsComponent {
     );
   }
 
-  private updateAxisLimits(xMin: number, xMax: number) {
+  /* Compute min and max from the store, adjusting them if needed for non linear scales.
+   * Getting the value from the store, we have to ensure that the values are integers for
+   * non linear scales (bar and category), otherwise the zooming will be off. After rounding,
+   * we have to adjust the values:
+   *
+   * When computing from the store, we adjust both min and max by substracting the axis offset,
```
rrahir: This part is not clear, at no point in the computation below do you remove any offset.

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs.ts
```diff
@@ -295,18 +292,53 @@ export class ZoomableChartJsComponent extends ChartJsComponent {
...
+  }
+
+  private updateAxisLimits(xMin: number, xMax: number) {
+    if (xMin === xMax) {
+      return;
     }
     this.store.updateAxisLimits(this.chartId, { min: xMin, max: xMax });
-    this.updateTrendingLineAxes();
+    const { min, max } = this.computeMinMaxFromStore();
+    if (max! > min! || (this.isBarChart! && max === min)) {
```
rrahir: the typing is really suspicious, from what I see we si;ply cannot trust it. can it be improved a bit, for instance `this.store.currentAxesLimits` could always have a value ?

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs.ts
anhe-odoo: No, you can't, because the value stored in the store for a non linear axis has been adjusted, and the only case where `ceil(min) > floor(max)` would be if `min = max`, which should be rejected before when storing the value. But i can add a test here to be sure.

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs.ts
anhe-odoo: Yes, you're totally right. At first, bot min and max were adjusted with a substraction but then I realised that it wasn't necessary and forgot to remove this part of the comment :/

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs.xml
anhe-odoo: Is there an equivalent to onMouseLeave for pointerX ?

rrahir: yes  :) https://developer.mozilla.org/en-US/docs/Web/API/Element/pointerleave_event

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs.ts
rrahir: image
min = 0.2
max = 0.9
math.ceil(0.2) = 1
math.floor(0.9) = 0
the screenshot attached with the debug point shows otherwise. This can be reproduced by mousedown/dragging the cursor on the first bar of any group or data.

Now perhaps this has no sense from a functional POV, just trying to select 1 data bar without hitting the center of the group, but it's very feasible

anhe-odoo: Yes, indeed, I mismatched my comment. This is possible but should be rejected after, not before (see https://github.com/odoo/o-spreadsheet/pull/7069/files#diff-d87b713bf99acb712149b81f4d05d23635d00780aced07db7b31d52f248709b5R335). Anyway, I'll add something to ensure we return undefined if it happens here :)

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs.ts
```diff
@@ -290,18 +287,63 @@ export class ZoomableChartJsComponent extends ChartJsComponent {
     );
   }
 
-  private updateAxisLimits(xMin: number, xMax: number) {
+  /* Compute min and max from the store, adjusting them if needed for non linear scales.
```
rrahir:
```suggestion
  /** Computes min and max from the store, adjusting them if needed for non linear scales.
```

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs.ts
```diff
@@ -290,18 +287,63 @@ export class ZoomableChartJsComponent extends ChartJsComponent {
...
-    } else {
-      this.chart!.config.options!.scales!.x!.min = xMin;
-      this.chart!.config.options!.scales!.x!.max = xMax;
+      min = Math.ceil(min);
+      max = Math.floor(max);
+    }
+    return { min, max };
+  }
+
+  /* Compute min and max from the scale, adjusting them if needed for non linear scales.
```
rrahir:
```suggestion
  /** Computes min and max from the scale, adjusting them if needed for non linear scales.
```
seems necessary to be recognized as the method docstring

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs.ts
rrahir: Still not convinced by the naming; but its quite the nitpick. "FromScale" means nothing to me. It implies that we take info from the scale and compute from it, but aren't we *adapting* the boundaries FOR the scale?

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs.ts
```diff
@@ -290,18 +287,63 @@ export class ZoomableChartJsComponent extends ChartJsComponent {
...
+   * the right way to do it.
+   * Sometimes, we can get a minimal value > the maximal value, which arise when the user
+   * select a very small area in the master chart, and hasn't selected the middle of a bar
+   * or a group of bars (in case of more than one data series).
+   * Assuming we have to select the middle of a bar/a groupe of bars, we will reject the
+   * coming value afterward. In this case, we do not update the chart because it would lead
+   * to an empty chart.
+   */
+
+  private computeMinMaxFromStore(): Boundaries {
```
rrahir: same here for the naming, I don't need it to describe where it gets the info from but rather describe what is does or why it does it or how it will be used.

computeBoundaryForDetailChart (+ the docstring) maybe, wdyt?

anhe-odoo: I went for adjustBoundaries and getStoredBoundaries :)

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs.ts
```diff
@@ -408,9 +450,16 @@ export class ZoomableChartJsComponent extends ChartJsComponent {
...
-    const { offsetX: x, offsetY: y } = ev;
+    const { offsetX: x, offsetY: y, target } = ev;
+    if (!target) {
+      return;
+    }
+    const runtime = this.env.model.getters.getChartRuntime(this.props.chartId) as ChartJSRuntime;
+    if (runtime && !runtime.chartJsConfig.data.datasets.some((ds) => ds.data.length > 1)) {
+      target["style"].cursor = "not-allowed";
+      return;
+    }
```
VincentSchippefilt: why is this style computed on mouse move ? It doesn't change depending on the mouse being in or out of the zoom chart, it changes based on the content of the data of the chart. 
This style should be set and reset when the chart changes, not on mouse events.

src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs.ts
```diff
@@ -290,18 +287,65 @@ export class ZoomableChartJsComponent extends ChartJsComponent {
...
+      this.chart.config.options!.scales!.x!.min = min;
+      this.chart.config.options!.scales!.x!.max = max;
+      this.updateTrendingLineAxes();
+      this.chart.update();
+    }
     this.masterChart?.update();
-    this.chart?.update();
   }
 
   onPointerDownInMasterChart(ev: PointerEvent) {
```
VincentSchippefilt:
```suggestion
  onMasterChartPointerDown(ev: PointerEvent) {
```
consider renaming. the "in" might be understood as "entering"

PR #7088, opened 2025-09-09
[FIX] charts: prevent unnecessary treemap animations

hokolomopo:
## Description

Charts animations are played every time the chart data changes in dashboards. But the treemap data contains callbacks, which mess up the `deepEqual` we use to check if the data changed, since new callbacks are created each time.

This adds an argument to `deepEqual` to ignore functions.

Task: [5003595](https://www.odoo.com/odoo/2328/tasks/5003595)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/misc.ts
+++ b/src/helpers/misc.ts
@@ -389,7 +389,7 @@ export function getAddHeaderStartIndex(position: "before" | "after", base: numbe
 /**
  * Compares two objects.
  */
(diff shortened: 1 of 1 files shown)


hokolomopo:
@robodoo fw=no

PR #7090, opened 2025-09-09
[FIX] xlsx: remove useless warnings on import

hokolomopo:
## Description

We generate some useless warnings when importing an xlsx:

- "FillStyle 'none' is not supported". We obviously support "none" fill style, as it is just the absence of fill style.

- "Number format 'General' is not supported". "General" is the default number format of a spreadsheet application, we support it. The conversion of the format was slightly wrong, as we only checked for "General" format for excel defaults format (numFmtId === 0), but not for user defined format with "General" as formatCode.

Task: [5075112](https://www.odoo.com/odoo/2328/tasks/5075112)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


pro-odoo:
robodoo r+

PR #7091, opened 2025-09-09
[FIX] Sort: Allow to sort array formula that do not spread

rrahir:
Some array formulas can lead to a single cell result. This can occur quite easily when manipulating array functions that end up aggregated in a single result. An easy example,

```javascript
=min(if(A1:B4>1, 2))
```
leads to a single cell result but should be sortable.

Task: 5033192

## Description:

Task: [5033192](https://www.odoo.com/odoo/2328/tasks/5033192)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/plugins/ui_feature/sort.ts
```diff
@@ -79,9 +86,11 @@ export class SortPlugin extends UIPlugin {
   }
 
   private checkArrayFormulaInSortZone({ sheetId, zone }: SortCommand): CommandResult {
-    const arrayFormulaInZone = positions(zone).some(({ col, row }) =>
-      this.getters.getArrayFormulaSpreadingOn({ sheetId, col, row })
-    );
+    const arrayFormulaInZone = positions(zone).some(({ col, row }) => {
+      const originPosition = this.getters.getArrayFormulaSpreadingOn({ sheetId, col, row });
+      // return originPosition
```
pro-odoo: 👀

src/plugins/ui_feature/sort.ts
```diff
@@ -1,4 +1,11 @@
-import { isInside, overlap, positions, range, zoneToDimension } from "../../helpers/index";
```
hokolomopo, line 1: Probably for another task, but there's a bug when sorting an array formula:

- enter =MUNIT(1) in A1, and a number literal in A2 (eg. 9)
- sort descending
- the sort is correct, 9 will be in A1
- but the top-bar composer of A1 is grayed out, like it would be if 9 was a value spreaded from an array formula

PR #7106, opened 2025-09-10
[FIX] charts: truncate radar chart labels correctly

rmbh-odoo:
## Description:
Current behavior before PR:
- Radar chart point labels were truncated incorrectly because Chart.js passes
extra arguments (label, index, labels) to the callback.
- As a result, truncateLabel received the index as maxLen and applied
unnecessary truncation.

- The callback now wraps truncateLabel to pass only the label string.
- This ensures long labels are truncated with ellipsis and short labels
remain unchanged.

Task: [5078858](https://www.odoo.com/odoo/2328/tasks/5078858)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/figures/charts/runtime/chartjs_scales.ts
+++ b/src/helpers/figures/charts/runtime/chartjs_scales.ts
@@ -226,7 +226,7 @@ export function getRadarChartScales(
       },
       pointLabels: {
         color: chartFontColor(definition.background),
-        callback: truncateLabel,
+        callback: (label: string) => truncateLabel(label),
       },
       suggestedMin: minValue < 0 ? minValue - 1 : 0,
     },
--- a/tests/figures/chart/radar_chart_plugin.test.ts
+++ b/tests/figures/chart/radar_chart_plugin.test.ts
@@ -148,6 +148,17 @@ describe("radar chart", () => {
     const runtime = model.getters.getChartRuntime("chartId") as RadarChartRuntime;
     expect(runtime.chartJsConfig.options?.scales?.r?.suggestedMin).toBe(-8);
   });
+
+  test("Radar chart point labels are truncated properly", () => {
+    const model = new Model();
+    createRadarChart(model, { dataSets: [{ dataRange: "A1:A2" }] }, "chartId");
+    const runtime = model.getters.getChartRuntime("chartId") as RadarChartRuntime;
+    const callback = (runtime.chartJsConfig.options?.scales?.r as any)?.pointLabels
+      ?.callback as Function;
+
+    expect(callback("short", 0)).toBe("short");
+    expect(callback("very very long label of radar", 1)).toBe("very very long label…");
+  });
 });
 
 test("Humanization is taken into account for the axis ticks of a radar chart", async () => {


src/helpers/figures/charts/runtime/chartjs_scales.ts
```diff
@@ -226,7 +228,17 @@ export function getRadarChartScales(
...
         color: chartFontColor(definition.background),
-        callback: truncateLabel,
+        font: { size: CHART_LABEL_FONT_SIZE },
+        display: "auto",
+        callback: function (label: string) {
+          let maxLen = MAX_CHAR_LABEL;
+          if (this as any) {
+            const chart = (this as any).chart;
+            const availableArc = (Math.PI * chart.chartArea.width) / chart.data.labels.length;
+            maxLen = Math.max(3, availableArc / CHART_LABEL_FONT_SIZE);
```
rmbh-odoo: Now, based on the width of the radar chart, we find a maxLen.
I can't find another way to get the width of the chart here, so I used this.
I can't test this because in the test case I get this as undefined, so I tested a simple callback function with maxLen.
Let me know if there is a better way to handle this. 🤔

hokolomopo: I'm not sure about these computations ... I don't think dividing by the number of the data points is correct, if you have a radar chart with 5 or 10 points the available width is not divided by 2, the radar "circle" stays about the same size. The available width is the space between the edge of the radar "circle" and the edge of the chart area. And why multiply by PI ? If you have a single data point, your available width is three time the width of the chart ? This doesn't seems right ...

Anyway I would only keep the previous version of your PR (removing index from truncateLabel arguments) in a FIX in 19.0, what Francois asked in the notes sounds more like an IMP for master to me.

hokolomopo: You can use [function.call](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/call) (or function.apply) test a function with `this` 🙂 

Something like `callback.call(mockedChart, other_arguments)`

hokolomopo:
And shouldn't the fix be in 18.2 ?

rmbh-odoo:
> And shouldn't the fix be in 18.2 ?

From 18.2 -> 18.4, truncateLabel doesn’t accept maxLen, so no fix is needed. 😌
From 19.0, maxLen was added (Chart.js callback index is passed as maxLen)
[See here: 6046](https://github.com/odoo/o-spreadsheet/pull/6046/files#diff-a91722d1796f17931e24fedf5e911ec77d0212123b0ecbc8631ca56a159c1a37)

src/helpers/figures/charts/runtime/chartjs_scales.ts
rmbh-odoo: https://drive.google.com/file/d/11eGDs-5rXq7qfdk22KHfMEaJTewJzLso/view?usp=sharing

I did a simple calculation here. 😉
it’s not fully accurate since I don’t think we can precisely calculate the available width for the label. 
If this feature is really needed, we’d have to do more detailed calculations and also apply it to other charts. 
Thanks for pointing it out!

hokolomopo:
@robodoo r+

Thanks for the PR !

PR #7109, opened 2025-09-11
[FIX] functions: fix LINEST error massage

hokolomopo:
## Description

The functions using the helper `tryCastAsNumberMatrix` had a wrong error message:

- the two arguments in the translated string were inverted
- an argument of `_t` was `typeof cell`, which isn't translated
- the other argument of the `_t` was something like `the first argument (data_y)`, which wasn't translated either

Task: [5059375](https://www.odoo.com/odoo/2328/tasks/5059375)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/functions/helpers.ts
+++ b/src/functions/helpers.ts
@@ -78,13 +78,24 @@ export function tryCastAsNumberMatrix(data: Matrix<any>, argName: string): Matri
   data.forEach((row) => {
     row.forEach((cell) => {
       if (typeof cell !== "number") {
-        throw new Error(
-          _t(
-            "Function [[FUNCTION_NAME]] expects number values for %s, but got a %s.",
-            typeof cell,
+        let message = "";
+        if (typeof cell === "object") {
+          message = _t(
+            "Function [[FUNCTION_NAME]] expects number values for %s, but got an empty value.",
             argName
-          )
-        );
+          );
+        } else if (typeof cell === "string") {
+          message = _t(
+            "Function [[FUNCTION_NAME]] expects number values for %s, but got a string.",
+            argName
+          );
+        } else if (typeof cell === "boolean") {
+          message = _t(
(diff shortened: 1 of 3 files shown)


src/functions/helpers.ts
```diff
@@ -74,15 +74,28 @@ export function tryToNumber(
   }
 }
 
+function cellTypeToTranslatedString(cellValue: CellValue): string {
+  switch (typeof cellValue) {
+    case "number":
+      return _t("a number");
+    case "string":
+      return _t("a string");
```
LucasLefevre: it will lead to better translations if the sentence isn't cut in half.
I'd "duplicate" the full error messages
```suggestion
function cellTypeToTranslatedString(cellValue: CellValue): string {
  switch (typeof cellValue) {
    case "number":
      return _t("Function [[FUNCTION_NAME]] expects number values for %s, but got a number.");
    case "string":
      return _t("Function [[FUNCTION_NAME]] expects number values for %s, but got a string.");
```

src/functions/helpers.ts
```diff
@@ -78,13 +78,24 @@ export function tryCastAsNumberMatrix(data: Matrix<any>, argName: string): Matri
...
+        if (typeof cell === "object") {
+          message = _t(
+            "Function [[FUNCTION_NAME]] expects number values for %s, but got an empty value.",
             argName
-          )
-        );
+          );
+        } else if (typeof cell === "string") {
+          message = _t(
+            "Function [[FUNCTION_NAME]] expects number values for %s, but got an empty string.",
```
rrahir:
```suggestion
            "Function [[FUNCTION_NAME]] expects number values for %s, but got a string.",
```

rrahir:
robodoo r+

PR #7116, opened 2025-09-11
[FIX] package: fix the version of swc

VincentSchippefilt:
when switching from version to version of node, it might be that the SWC/Jest package updates from 0.2.36 to 0.2.38 on newer branches. This brakes the ability to run tests on lower branches with "cannot find bindings..." error.


pro-odoo:
https://github.com/odoo/o-spreadsheet/pull/7147

rrahir:
solved in #7147

PR #7119, opened 2025-09-11
[IMP] carousel: add context menu button

hokolomopo:
## Description

This commit adds a button to open the context menu of a carousel, in addition to the right-click which is not discoverable.

Task: [5081771](https://www.odoo.com/odoo/2328/tasks/5081771)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/figures/figure/figure.xml
+++ b/src/components/figures/figure/figure.xml
@@ -19,6 +19,7 @@
           onFigureDeleted="props.onFigureDeleted"
           figureUI="props.figureUI"
           editFigureStyle.bind="editWrapperStyle"
+          openContextMenu.bind="openContextMenu"
         />
         <div class="o-figure-menu position-absolute m-2" t-if="!env.isDashboard()">
           <div
--- a/src/components/figures/figure_carousel/figure_carousel.ts
+++ b/src/components/figures/figure_carousel/figure_carousel.ts
@@ -12,6 +12,7 @@ import {
   CarouselItem,
   FigureUI,
   MenuMouseEvent,
+  Rect,
   SpreadsheetChildEnv,
 } from "../../../types";
 import { FullScreenFigureStore } from "../../full_screen_figure/full_screen_figure_store";
@@ -26,6 +27,7 @@ interface Props {
   onFigureDeleted: () => void;
   editFigureStyle?: (properties: CSSProperties) => void;
   isFullScreen?: boolean;
+  openContextMenu?: (anchorRect: Rect, onClose?: () => void) => void;
 }
 
 export class CarouselFigure extends Component<Props, SpreadsheetChildEnv> {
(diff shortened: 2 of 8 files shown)


rrahir:
mass conflict :/

Do you know how this'll work with the fullscreen ?

hokolomopo:
@rrahir rebased :) 

Well the fullscreeen button is only present in dashboard, and this button is only present when not in dashboard. So there shouldn't be any problem there :D

src/components/figures/figure/figure.xml
```diff
@@ -33,7 +34,7 @@
             t-if="menuState.isOpen"
             anchorRect="menuState.anchorRect"
             menuItems="menuState.menuItems"
-            onClose="() => this.menuState.isOpen=false"
```
rrahir, line 36: i dont understand that change, is it necessary? I couldnt find a test that fails with this change

hokolomopo: Yeah, it's to have an `active` class on the button when the menu is open, and remove it when it closes. 

But you're right it's untested, I added 3 lines to test it

src/components/figures/figure_carousel/figure_carousel.ts
```diff
@@ -41,6 +44,7 @@ export class CarouselFigure extends Component<Props, SpreadsheetChildEnv> {
   private hiddenItems: CarouselItem[] = [];
 
   protected animationStore: Store<ChartAnimationStore> | undefined;
+  private state = useState({ isMenuActive: false });
```
LucasLefevre: It's not obvious what's the difference between `this.state.isMenuActive` and `this.menuState.isOpen`.

Maybe a bit or renaming might help to differentiate both menus

rrahir:
rebase required :/

src/components/figures/figure/figure.xml
LucasLefevre: The same button (ellipsis) in dashboard mode doesn't have any `active` class or any different style when the menu is open.
I'd happily live without the `active` class here, especially if it removes complexity and `state`

grep `o-chart-dashboard-item btn text-muted lh-1 p-1 fa fa-ellipsis-v`

image

LucasLefevre (commented):
conflicts are real

src/components/figures/figure/figure.xml
hokolomopo: You know I was of the exact opposite mind, let's add a bit of complexity everywhere so it's prettier everywhere. I'll ask francois if he cares at all, then we can discuss it :)

(note: where the hell is the menu opening in the screenshot you send ... we should fix that)

hokolomopo:
@LucasLefevre I forgot to run the tests 🙈 The props validation was wrong

LucasLefevre:
robodoo r+

PR #7123, opened 2025-09-12
Master convert scss pro

pro-odoo:
## Description:

Task: [5086011](https://www.odoo.com/odoo/2328/tasks/5086011)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


.husky/scss_check.js
```diff
@@ -9,7 +9,8 @@ const firstLevelSelectorPattern = /^(?!(\s|\$|\})).+/gm;
 const files = execSync("git diff --name-only --cached")
   .toString()
   .split("\n")
-  .filter((file) => file.endsWith(".scss"));
+  .filter((file) => file.endsWith(".scss"))
+  .filter((file) => !file.endsWith("variables.scss")); // variables.scss is allowed to have global styles
```
rrahir: I think we should keep it bounded, we will end up colliding with some too generic variable. why do you want to ignore it?

src/components/top_bar/top_bar.scss
```diff
@@ -17,4 +17,88 @@
...
+
+  .o-topbar-divider {
+    border-right: 1px solid $os-separator-color;
+    width: 0;
+    margin: 0 6px;
+    height: 30px;
+  }
+
+  .o-toolbar-button {
+    height: 30px;
```
rrahir, line 29: this specific line was overriding the default height of the class `o-font-size-editor` to force it to 30px. it probably zworked by pure luck because the files were loaded in a different order depending on their cross dependencies. SHould be fixed in an ulterior work

src/components/side_panel/side_panel/side_panel.scss
```diff
@@ -0,0 +1,102 @@
...
+        color: $os-text-body;
+      }
+    }
+
+    .o-sidePanelButtons {
+      display: flex;
+      gap: 8px;
+    }
+
+    .o-invalid {
```
rrahir: i am 99% sure this can be removed. it was added to have a generic support of invalid values in the standalone composer and it has a dedicated rule since then
https://github.com/odoo/o-spreadsheet/pull/7123/files#diff-8381fb056fa8538e95f048b8e24c4258c6e61157e7653d3d2fd445fa6e57b47bR12

edit: this address the points of Lucas in the task

.husky/scss_check.js
pro-odoo: Because it does not work if it's bounded

src/components/top_bar/top_bar.scss
pro-odoo: Noted

.husky/scss_check.js
rrahir: I just checked and apparently we don't have bound them but it was able to compile without the bounding so I don't see why you need this. `scss_check` has a specific rule that allows  saas variables `const firstLevelSelectorPattern = /^(?!(\s|\$|\})).+/gm;`
So it should work and the scss_check should be fixed if it doesn't

rrahir: If we decid to introduce *css* variables, those should probably be scoped though

.husky/scss_check.js
```diff
@@ -3,7 +3,7 @@ import { existsSync, readFileSync } from "fs";
 import process from "process";
 
 const commentPattern = /\/\/.*|\/\*[\s\S]*?\*\//g;
-const firstLevelSelectorPattern = /^(?!(\s|\$|\})).+/gm;
+const firstLevelSelectorPattern = /^(?!(\s|\$|\}|\))).+/gm;
```
rrahir, line 6: the selector pattern was not supporting some saas  operator

src/variables.scss
```diff
@@ -4,13 +4,119 @@ $os-gray-300: #d8dadd;
...
+$os-alert-info-border: #98dbe2;
+$os-alert-info-text-color: #09414a;
+
+$os-default-font: "Roboto", "Arial";
+$os-default-font-size: 10px;
+$os-header-height: 26px;
+$os-header-width: 48px;
+$os-desktop-topbar-toolbar-height: 34px;
+$os-mobile-topbar-toolbar-height: 44px;
+$os-desktop-bottombar-height: 36px;
```
hokolomopo, line 54: a bunch of variables in this file are never used

src/variables.scss
```diff
@@ -4,13 +4,119 @@ $os-gray-300: #d8dadd;
...
+/** Color Picker **/
+$os-line-vertical-padding: 1px;
+$os-picker-padding: 8px;
+$os-item-border-width: 1px;
+$os-item-edge-length: 18px;
+$os-items-per-line: 10;
+$os-magnifier-edge: 16px;
+$os-item-gap: 2px;
+
+$os-content-width: (
```
hokolomopo: nitpick: rename to os-color-picker-width

hokolomopo (commented):
:wave:

rrahir:
robodoo rebase-ff r+

PR #7141, opened 2025-09-16
[FIX] headers: can add lots of headers

LucasLefevre:
Javascript engines have a limit to the number of function arguments [1].

calling `splice(...items)` where `items` is a huge array, crashes with a `Maximum call stack size exceeded` error

[1]: Experimentally, that limit is 117k on Google Chrome and 501k on Firefox.

Task: [5092626](https://www.odoo.com/odoo/2328/tasks/5092626)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/helpers/misc.ts
```diff
@@ -511,9 +511,7 @@ export function removeIndexesFromArray<T>(array: readonly T[], indexes: number[]
 }
 
 export function insertItemsAtIndex<T>(array: readonly T[], items: T[], index: number): T[] {
-  const newArray = [...array];
-  newArray.splice(index, 0, ...items);
-  return newArray;
+  return array.slice(0, index).concat(items).concat(array.slice(index));
```
hokolomopo, line 514: nitpick: isn't `[...array.slice(0, index), ...items, ...array.slice(index)]` theoretically better since we create on less (possibly huge) array ? Or is concat faster ?

rrahir: I just discovered https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSpliced  

This benchmark seems quite promissing
```js
const bench = (fn, iterations=5000) => {
  const start = (typeof performance !== 'undefined') ? performance.now() : Date.now();
  for (let i=0;i<iterations;i++) fn();
  const end = (typeof performance !== 'undefined') ? performance.now() : Date.now();
  return end - start;
};

const big = Array.from({length:100_000}, (_,i)=>i);
const items = Array.from({length:1000}, (_,i)=>i);

// setup functions
const f1 = () => {
  const index = 50000;
  return big.slice(0,index).concat(items).concat(big.slice(index));
};

const f2 = () => {
  const index = 50000;
  return [...big.slice(0,index), ...items, ...big.slice(index)];
};

const f3 = () => {
  const index = 50000;
  return big.toSpliced(index, 0, ...items);;
};

console.log('chained concat (ms):', bench(f1, 500));
console.log('spread literal (ms):', bench(f2, 500));
console.log('toSpliced is optimized (ms):', bench(f3, 500));
```

rrahir: toSpliced is equivalent to "chained concat" on FF but it destroys the 2 other solutions on V8 (chrome)

LucasLefevre: Cool benchmark, thanks!
at least we know `concat` is better in this case (`toSpliced` has the same problem as `splice` with the max number of arguments obviously)

rrahir: just realized that toSpliced suffers from the very same problem we try to solve -_- 

but the benchmark is still relevant:
chained concat is faster than destructuring by a factor 8 on V8 and a factor 2 on FF

from a memory POV,  destructuring is more efficient since we copy less stuff.

hokolomopo: toSpliced is so much faster than the others on firefox 😮 

But yeah it looks like spreading sucks 🙃 Let's keep concat then (or we could use toSpliced for arrays < 100k elements, but that seems overkill)

hokolomopo:
@robodoo r+

PR #7146, opened 2025-09-16
[FIX] data_validation: selecting range from another sheet

dhrp-odoo:
## Description:

Steps to reproduce:
- Create a new sheet (Sheet2).
- On Sheet1, open the Data Validation side panel.
- Add a DV with type "isValueInRange".
- Click on the selection input to select the range.
- Navigate to Sheet2 and select the range.
- Click on Save without confirming the ranges.

Before this PR:
- The DV was created on Sheet2 instead of remaining on Sheet1.

After this PR:
- The DV is correctly created on the original sheet (Sheet1).

Task: [4948201](https://www.odoo.com/odoo/2328/tasks/4948201)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/data_validation/dv_editor/dv_editor.ts
+++ b/src/components/side_panel/data_validation/dv_editor/dv_editor.ts
@@ -10,6 +10,7 @@ import {
   DataValidationRule,
   DataValidationRuleData,
   SpreadsheetChildEnv,
+  UID,
 } from "../../../../types";
 import { css } from "../../../helpers";
 import { SelectionInput } from "../../../selection_input/selection_input";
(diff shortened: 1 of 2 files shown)


src/components/side_panel/data_validation/dv_editor/dv_editor.ts
```diff
@@ -99,7 +101,7 @@ export class DataValidationEditor extends Component<Props, SpreadsheetChildEnv>
       .map((value) => canonicalizeContent(value, locale));
     rule.criterion = { ...criterion, values };
     return {
-      sheetId,
+      sheetId: this.editingSheetId,
       ranges: this.state.rule.ranges.map((xc) =>
         this.env.model.getters.getRangeDataFromXc(sheetId, xc)
```
hokolomopo: should also be `this.editingSheetId` there, no ?

dhrp-odoo: Yes, we need to change it to `this.editingSheetId`, otherwise it will use the wrong sheet reference in apply range when selecting a range from another sheet. Thanks for catching this!

PR #7179, opened 2025-09-22
[FIX] Pivot: Adapt computed measures on sheet rename

rrahir:
Currently, the sheet rename command does not properly adapt the references in  the computed measures.

Task: 5095364

## Description:

Task: [5095364](https://www.odoo.com/odoo/2328/tasks/5095364)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/formulas.ts
+++ b/src/helpers/formulas.ts
@@ -1,5 +1,6 @@
 import { rangeTokenize } from "../formulas";
 import { Range, RangeAdapter, UID } from "../types";
+import { CellErrorType } from "../types/errors";
 import { concat } from "./misc";
 import { createInvalidRange, createRangeFromXc, getRangeString } from "./range";
 import { rangeReference, splitReference } from "./references";
@@ -8,7 +9,8 @@ import { isSheetNameEqual } from "./sheet";
 export function adaptFormulaStringRanges(
   defaultSheetId: string,
   formula: string,
-  applyChange: RangeAdapter
+  applyChange: RangeAdapter,
+  { preserveInvalid }: { preserveInvalid?: boolean } = { preserveInvalid: true }
 ): string {
   if (!formula.startsWith("=")) {
     return formula;
@@ -19,7 +21,7 @@ export function adaptFormulaStringRanges(
       continue;
     }
     const sheetXC = tokens[tokenIdx].value;
-    const newSheetXC = adaptStringRange(defaultSheetId, sheetXC, applyChange);
+    const newSheetXC = adaptStringRange(defaultSheetId, sheetXC, applyChange, { preserveInvalid });
 
     if (sheetXC !== newSheetXC) {
       tokens[tokenIdx] = {
@@ -34,7 +36,8 @@ export function adaptFormulaStringRanges(
 export function adaptStringRange(
   defaultSheetId: UID,
   sheetXC: string,
-  applyChange: RangeAdapter
+  applyChange: RangeAdapter,
+  { preserveInvalid }: { preserveInvalid?: boolean } = { preserveInvalid: true }
 ): string {
   const sheetName = splitReference(sheetXC).sheetName;
   if (
@@ -52,8 +55,11 @@ export function adaptStringRange(
   }
 
   const change = applyChange.applyChange(range);
-  if (change.changeType === "NONE" || change.changeType === "REMOVE") {
+  if (change.changeType === "NONE") {
     return sheetXC;
+  }
+  if (change.changeType === "REMOVE") {
+    return preserveInvalid ? sheetXC : CellErrorType.InvalidReference;
   }
 
   return getRangeString(change.range, defaultSheetId, getSheetNameGetter(applyChange));
--- a/src/plugins/core/pivot.ts
+++ b/src/plugins/core/pivot.ts
@@ -1,9 +1,11 @@
 import { compile } from "../../formulas";
 import { deepCopy, deepEquals } from "../../helpers";
+import { adaptFormulaStringRanges } from "../../helpers/formulas";
 import { createPivotFormula, getMaxObjectId } from "../../helpers/pivot/pivot_helpers";
 import { pivotRegistry } from "../../helpers/pivot/pivot_registry";
 import { SpreadsheetPivotTable } from "../../helpers/pivot/table_spreadsheet_pivot";
 import {
+  AdaptSheetName,
   ApplyRangeChange,
   CellPosition,
   CommandResult,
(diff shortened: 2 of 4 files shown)


src/plugins/core/pivot.ts
```diff
@@ -157,13 +159,19 @@ export class PivotCorePlugin extends CorePlugin<CoreState> implements CoreState
...
-          sheetId,
-          compiledFormula.tokens,
-          newDependencies
-        );
+        const newFormulaString =
+          // this.getters.getFormulaString(
+          //   formulaSheetId,
+          //   compiledFormula.tokens,
+          //   newDependencies
+          // );
```
hokolomopo: 🙈

src/helpers/formulas.ts
```diff
@@ -52,9 +52,12 @@ export function adaptStringRange(
   }
 
   const change = applyChange.applyChange(range);
-  if (change.changeType === "NONE" || change.changeType === "REMOVE") {
+  if (change.changeType === "NONE") {
     return sheetXC;
   }
+  if (change.changeType === "REMOVE") {
+    return "#REF";
+  }
```
hokolomopo, line 63: no test for this change ?

rrahir: I just looked into it and there is a catch - if you could have a look at the second commit -  the test shows that in case of a concurrent deleted sheet, the reference to the sheet is definitely lost, whereas before, we would fallback to the old sheetXC which was invalid but at least persistent.

I applied this change because the new transformation would break the test here <https://github.com/odoo/o-spreadsheet/blob/saas-18.3/tests/pivots/pivot_calculated_measure.test.ts#L1078> which makes sense from a functional POV but the test added in the second commit shows that we using the same code in OT, we then lose the  sheetXC upon OT conflicts.

I can add an option to `adaptStringRange`  and `adaptFormulaStringRanges` to try to keep the reference on OT conflict but then we have two different behaviours when adapting ranges through a local change and command conversion with OT, Which I don't really like... because we don't follow the same flow between adapranges and the OT transformation.

The thing is, we theoretically have more possibilities locally because of the history, the adaptation of commands with OT require us to make a choice and I don't know which one---

A. an XC being invalidated with OT becomes #REF and is lost for ever
B. We preserve the XC and then we have a difference of behaviour between OT and adaptRange

not 100% sure but it feels like its again the issue that we sometimes manipulate ranges (or RangeData) and sometimes XC in the commands, its hard to factorize behaviours having to juggle with both since they dont contain the same information

@pro-odoo @LucasLefevre @VincentSchippefilt @anhe-odoo @laa-odoo @fdamhaut  if any of you guys have an input...

tests/collaborative/collaborative_sheet_manipulations.test.ts
```diff
@@ -1209,3 +1252,64 @@ test("concurrent pivot computed measure and rename sheet", () => {
...
+    (user) => user.getters.getPivotCoreDefinition("pivot1").measures,
+    [
+      {
+        id: "measure",
+        fieldName: "mm",
+        aggregator: "sum",
+        computedBy: { formula: `=${CellErrorType.InvalidReference}*2`, sheetId: secondSheetId },
+      },
+    ]
+  );
```
VincentSchippefilt, line 1314: cannot be true, if alice didn't delete the sheet, we have to have the correct reference

src/plugins/core/spreadsheet_pivot.ts
```diff
@@ -45,6 +45,10 @@ export class SpreadsheetPivotCorePlugin extends CorePlugin {
       }
       if (definition.dataSet) {
         const { sheetId, zone } = definition.dataSet;
+        if (!this.getters.tryGetSheet(sheetId) || !zone || !isZoneValid(zone)) {
+          this.dispatch("UPDATE_PIVOT", { pivotId, pivot: { ...definition, dataSet: undefined } });
+          return;
+        }
```
LucasLefevre: is this supposed to be in the second commit ?

rrahir:
rebased

src/helpers/formulas.ts
```diff
@@ -8,7 +9,8 @@ import { isSheetNameEqual } from "./sheet";
 export function adaptFormulaStringRanges(
   defaultSheetId: string,
   formula: string,
-  applyChange: RangeAdapter
+  applyChange: RangeAdapter,
+  preserveInvalid: boolean = true
```
LucasLefevre: name the boolean parameter
```suggestion
  { preserveInvalid }: { preserveInvalid?: boolean } = { preserveInvalid: true }
```

src/plugins/core/spreadsheet_pivot.ts
```diff
@@ -45,6 +45,10 @@ export class SpreadsheetPivotCorePlugin extends CorePlugin {
       }
       if (definition.dataSet) {
         const { sheetId, zone } = definition.dataSet;
+        if (!this.getters.tryGetSheet(sheetId) || !zone || !isZoneValid(zone)) {
+          this.dispatch("UPDATE_PIVOT", { pivotId, pivot: { ...definition, dataSet: undefined } });
+          return;
```
LucasLefevre: It skips other iterations with other pivots
```suggestion
          continue;
```

src/plugins/core/spreadsheet_pivot.ts
LucasLefevre: oupss, it also skips all other pivots

src/helpers/formulas.ts
LucasLefevre: Sorry, I missed that comment at the time.
I'd preserve the XC I think (and adapt the failing test)

src/plugins/core/pivot.ts
LucasLefevre, line 170: the `newDependencies` stuff is not needed anymore

src/plugins/core/spreadsheet_pivot.ts
rrahir: https://www.odoo.com/odoo/2328/tasks/5379776

src/plugins/core/spreadsheet_pivot.ts
rrahir: nope, in that case, we have to account for the fact that the sheet related to the range might already have disappeared (it was processed in the Sheet core plugin already) so the range might have become stale. In that case, the pivot definition needs to be adapted accordingly, the dataset is simply invalid.

rrahir:
we went with another implementation
https://github.com/odoo/o-spreadsheet/pull/7751

PR #7183, opened 2025-09-22
[REF] side_panel: split CF/DV panel into main and editor panel

rmbh-odoo:
## Description:

Current behavior before PR:
- CF and DV side panels combined the list and editor into a single panel.
- The editor could not be opened directly.
- Logic for the main panel and editor was mixed, reducing modularity.

- Split CF and DV into separate main and editor panels.
- Panels can be opened or replaced directly with openSidePanel / replaceSidePanel.
- Use a store to manage the CF editor state.
- Refactor related test cases.

Task: [4982596](https://www.odoo.com/odoo/2328/tasks/4982596)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/conditional_formatting/cf_editor/cf_editor.xml
+++ b/src/components/side_panel/conditional_formatting/cf_editor/cf_editor.xml
@@ -1,6 +1,6 @@
 <templates>
   <t t-name="o-spreadsheet-ConditionalFormattingEditor">
-    <div class="o-cf-ruleEditor">
+    <div class="o-cf-editor">
       <t t-set="state" t-value="store.state"/>
       <Section class="'o-cf-range pb-0'" title.translate="Apply to range">
         <div class="o-selection-cf">
@@ -23,7 +23,7 @@
           />
         </div>
       </Section>
-      <Section class="'o-cf-editor'">
+      <Section>
         <t t-if="state.currentCFType === 'CellIsRule'" t-call="o-spreadsheet-CellIsRuleEditor">
           <t t-set="rule" t-value="state.rules.cellIs"/>
         </t>
--- a/tests/conditional_formatting/conditional_formatting_panel_component.test.ts
+++ b/tests/conditional_formatting/conditional_formatting_panel_component.test.ts
@@ -55,20 +55,18 @@ function errorMessages(): string[] {
 const selectors = {
   listPreviewPanel: ".o-cf-preview-list",
   listPreview: ".o-cf-preview-list .o-cf-preview",
-  editorPanel: ".o-cf-ruleEditor",
+  editorPanel: ".o-cf-editor",
   ruleEditor: {
-    range: ".o-cf-ruleEditor .o-cf-range .o-range input",
+    range: ".o-cf-editor .o-cf-range .o-range input",
     editor: {
-      operatorInput: ".o-cf-ruleEditor .o-cf-editor .o-cell-is-operator",
-      valueInput: ".o-cf-ruleEditor .o-cf-editor .o-cf-cell-is-rule .o-composer",
-      bold: ".o-cf-ruleEditor .o-cf-editor .o-sidePanel-tools div.o-menu-item-button[title='Bold']",
-      italic:
-        ".o-cf-ruleEditor .o-cf-editor .o-sidePanel-tools div.o-menu-item-button[title='Italic']",
-      underline:
-        ".o-cf-ruleEditor .o-cf-editor .o-sidePanel-tools div.o-menu-item-button[title='Underline']",
+      operatorInput: ".o-cf-editor .o-cell-is-operator",
+      valueInput: ".o-cf-editor .o-cf-cell-is-rule .o-composer",
+      bold: ".o-cf-editor .o-sidePanel-tools div.o-menu-item-button[title='Bold']",
+      italic: ".o-cf-editor .o-sidePanel-tools div.o-menu-item-button[title='Italic']",
+      underline: ".o-cf-editor .o-sidePanel-tools div.o-menu-item-button[title='Underline']",
       strikethrough:
-        ".o-cf-ruleEditor .o-cf-editor .o-sidePanel-tools div.o-menu-item-button[title='Strikethrough']",
-      colorDropdown: ".o-cf-ruleEditor .o-cf-editor .o-color-picker-widget .o-color-picker-button",
+        ".o-cf-editor .o-sidePanel-tools div.o-menu-item-button[title='Strikethrough']",
+      colorDropdown: ".o-cf-editor .o-color-picker-widget .o-color-picker-button",
       iconSetRule: {
         container: ".o-cf-iconset-rule",
         iconsets: ".o-cf-iconset-rule .o-cf-iconsets .o-cf-iconset",
--- a/src/components/side_panel/conditional_formatting/cf_editor/cf_editor.css
+++ b/src/components/side_panel/conditional_formatting/cf_editor/cf_editor.css
@@ -1,5 +1,5 @@
 .o-spreadsheet {
-  .o-cf-ruleEditor {
+  .o-cf-editor {
     .o-cf-preview-display {
       padding: 10px;
     }
--- a/tests/top_bar_component.test.ts
+++ b/tests/top_bar_component.test.ts
@@ -840,7 +840,7 @@ describe("TopBar - CF", () => {
     await click(fixture, ".o-topbar-menu[data-id='format']");
     await click(fixture, ".o-menu-item[data-name='format_cf']");
     expect(fixture.querySelector(".o-sidePanel .o-sidePanelBody .o-cf-preview-list")).toBeTruthy();
-    expect(fixture.querySelector(".o-sidePanel .o-sidePanelBody .o-cf-ruleEditor")).toBeFalsy();
+    expect(fixture.querySelector(".o-sidePanel .o-sidePanelBody .o-cf-editor")).toBeFalsy();
   });
 
   test("open sidepanel with one CF in selected zone", async () => {
@@ -867,7 +867,7 @@ describe("TopBar - CF", () => {
     await click(fixture, ".o-topbar-menu[data-id='format']");
     await click(fixture, ".o-menu-item[data-name='format_cf']");
     expect(fixture.querySelector(".o-sidePanel .o-sidePanelBody .o-cf-preview-list")).toBeFalsy();
-    expect(fixture.querySelector(".o-sidePanel .o-sidePanelBody .o-cf-ruleEditor")).toBeTruthy();
+    expect(fixture.querySelector(".o-sidePanel .o-sidePanelBody .o-cf-editor")).toBeTruthy();
   });
 
   test("open sidepanel with with more then one CF in selected zone", async () => {
@@ -909,9 +909,7 @@ describe("TopBar - CF", () => {
     await click(fixture, ".o-topbar-menu[data-id='format']");
     await click(fixture, ".o-menu-item[data-name='format_cf']");
     expect(fixture.querySelector(".o-sidePanel .o-sidePanelBody .o-cf-preview-list")).toBeTruthy();
-    expect(
-      fixture.querySelector(".o-sidePanel .o-sidePanelBody .o-cf-ruleEditor")
-    ).toBeFalsy();
+    expect(fixture.querySelector(".o-sidePanel .o-sidePanelBody .o-cf-editor")).toBeFalsy();
   });
 
   test("will update sidepanel if we reopen it from other cell", async () => {
@@ -942,13 +940,13 @@ describe("TopBar - CF", () => {
     await click(fixture, ".o-topbar-menu[data-id='format']");
     await click(fixture, ".o-menu-item[data-name='format_cf']");
     expect(fixture.querySelector(".o-sidePanel .o-sidePanelBody .o-cf-preview-list")).toBeFalsy();
-    expect(fixture.querySelector(".o-sidePanel .o-sidePanelBody .o-cf-ruleEditor")).toBeTruthy();
+    expect(fixture.querySelector(".o-sidePanel .o-sidePanelBody .o-cf-editor")).toBeTruthy();
 
     setSelection(model, ["A1:F1"]);
     await click(fixture, ".o-topbar-menu[data-id='format']");
     await click(fixture, ".o-menu-item[data-name='format_cf']");
     expect(fixture.querySelector(".o-sidePanel .o-sidePanelBody .o-cf-preview-list")).toBeTruthy();
-    expect(fixture.querySelector(".o-sidePanel .o-sidePanelBody .o-cf-ruleEditor")).toBeFalsy();
+    expect(fixture.querySelector(".o-sidePanel .o-sidePanelBody .o-cf-editor")).toBeFalsy();
   });
 });
 


tests/conditional_formatting/conditional_formatting_panel_component.test.ts
```diff
@@ -138,31 +137,25 @@ describe("UI of conditional formats", () => {
...
-  beforeEach(async () => {
-    ({ model, fixture, env } = await mountComponentWithPortalTarget(ConditionalFormattingPanel, {
-      props: { onCloseSidePanel: () => {} },
-    }));
-    sheetId = model.getters.getActiveSheetId();
-  });
-
-  describe("Conditional format list", () => {
-    let sheetId: UID;
+  describe("Conditional formatting list panel", () => {
```
rmbh-odoo, line 141: I didn’t remove any test cases; I only reordered them based on the preferred environment and titles. 😑

src/components/side_panel/conditional_formatting/cf_editor/cf_editor_store.ts
rmbh-odoo, line 1: Here, I used a store to separate business logic from UI logic, and it was also needed to handle the undo operation.

src/actions/menu_items_actions.ts
```diff
@@ -554,7 +555,44 @@ export const FORMAT_PERCENT_ACTION = (env: SpreadsheetChildEnv) => setFormatter(
...
+    if (result.isSuccessful) {
+      const ranges = zones.map((zone) =>
+        zoneToXc(env.model.getters.getUnboundedZone(sheetId, zone))
+      );
+      env.openSidePanel("ConditionalFormattingEditor", { cf: { ...cf, ranges }, isNewCf: true });
+    }
+  };
+
+  // TODO: add manage rule in (sleection/sheet) feature like excel does
+  const rules = env.model.getters.getConditionalFormats(sheetId);
```
rmbh-odoo: image

We can improve this by adding an option to manage rules at the selection or sheet level,
as shown in the attached image.

rmbh-odoo (commented):
Hi 👋
Here are some comments to help speed up the review process.

src/components/side_panel/conditional_formatting/conditional_formatting.xml
```diff
@@ -1,21 +1,12 @@
...
-          editedCf="editedCF"
-          onExit.bind="switchToList"
-          onCancel.bind="cancelEdition"
-          isNewCf="originalEditedCf === undefined"
-        />
-      </t>
+      <ConditionalFormattingPreviewList conditionalFormats="conditionalFormats"/>
+      <div
+        class="o-button-link p-4 o-cf-add float-end"
+        t-on-click.prevent.stop="this.addConditionalFormat">
```
dhrp-odoo:
```suggestion
        t-on-click.prevent.stop="addConditionalFormat">
```

rmbh-odoo:
> maybe the `ConditionalFormattingPanel` and `DataValidationPanel` might no longer be needed, since you’re already switching between panels directly through the side panel.

I removed the list panel because it has become useless. 
However, we cannot remove the main panels, since doing so would add more complexity to the preview panel. 😐

tests/data_validation/data_validation_list_component.test.ts
```diff
@@ -54,8 +53,9 @@ describe("Edit criterion in side panel", () => {
         values: ["ok", "hello", "okay"],
         displayStyle: "arrow",
       });
-      ({ fixture } = await mountDataValidationPanel(model));
-      await click(fixture, ".o-dv-preview");
+      ({ fixture, env } = await mountSpreadsheet({ model }));
```
hokolomopo: you can mount the component SidePanels instead, it's lighter than mounting the whols=e spreadsheet

src/components/side_panel/data_validation/dv_editor/dv_editor.ts
```diff
@@ -77,15 +75,17 @@ export class DataValidationEditor extends Component<Props, SpreadsheetChildEnv>
     this.state.rule.isBlocking = isBlocking === "true";
   }
 
+  onCancel() {
+    this.env.replaceSidePanel("DataValidation", `DataValidationEditor_${this.state.rule.id}`);
```
hokolomopo: isn't it this.props.id instead of this.state.rule.id ?

src/actions/insert_actions.ts
```diff
@@ -1,6 +1,5 @@
 import { functionRegistry } from "../functions";
```
hokolomopo: The first commit message seems somewhat wrong, there were already 2 side panels before the commit

src/components/side_panel/conditional_formatting/cf_preview_list/cf_preview_list.ts
```diff
@@ -1,71 +0,0 @@
-import { Component, onWillUpdateProps, useRef } from "@odoo/owl";
```
hokolomopo, line 3: I'd keep this file, and make it the main CF side panel instead of `conditional_formatting.ts`. The name is better, and this won't explode git history

src/actions/menu_items_actions.ts
```diff
@@ -554,7 +555,44 @@ export const FORMAT_PERCENT_ACTION = (env: SpreadsheetChildEnv) => setFormatter(
 // Side panel
 //------------------------------------------------------------------------------
 export const OPEN_CF_SIDEPANEL_ACTION = (env: SpreadsheetChildEnv) => {
-  env.openSidePanel("ConditionalFormatting", { selection: env.model.getters.getSelectedZones() });
+  const sheetId = env.model.getters.getActiveSheetId();
+  const zones = env.model.getters.getSelectedZones();
+  const createNewCf = () => {
```
hokolomopo: 1) if we keep this here, it' s probably easier to read without creating a callback, and just inline the code

2) it's kinda strange to have this piece of code here, and to handle the cancel that remove the CF in the side panel. IMO either everything should be handled on the side panel, or the onCancel that deletes this new CF should be a props of the panel

src/components/side_panel/conditional_formatting/cf_editor/cf_editor.xml
```diff
@@ -23,7 +23,7 @@
           />
         </div>
       </Section>
-      <Section class="'o-cf-editor'">
```
hokolomopo, line 25: while you're changing this, also change the class name to not have camelCase (o-cf-rule-editor)

src/actions/menu_items_actions.ts
hokolomopo: I personally find it kinda useless. You already have the highlight on hover to check which CF you're working with, and I don't think normal sheets have enough CFs to justify the additional code to handle the feature.

Excel only showing me the rules in my selection by default also tends to piss me off more than anything, I'm always wondering why my CFs don't show.

Just my 2 cents, you can run the idea by Francois if you want :)

hokolomopo (commented):
Hello 👋 

I'm not against the changes, having the code in the store + in different panels makes it a bit better (it will break forward ports tho, but w/e)

But functionally, opening two CFs side panels (which was the goal of the task AFAIK) is still kinda wonky. We cannot have 2 panels CF list side by side, so if we have two CF panels side by side clicking save/cancel on both will combine them into a single list panel. Maybe we should allow to open 2 of the same panel ? but then we probably need real ids for the panels

src/actions/menu_items_actions.ts
rmbh-odoo: That makes more sense. :+1: 
For now, we won’t add this, as the use case is limited and it could confuse users. :confused:

src/actions/menu_items_actions.ts
rmbh-odoo: You're right - I actually used `createNewCf` multiple times and forgot to remove it at the end.

For the second point, I attempted to create the CF in setup when no CF is passed; however, the dispatch inside setup triggers an infinite re-render. If you know a clean way to handle that, let me know. :slightly_smiling_face: 
Also, I don’t think passing onCancel as a prop and handling it everywhere makes sense here.

src/components/side_panel/conditional_formatting/cf_editor/cf_editor.xml
```diff
@@ -23,7 +23,7 @@
...
+      <Section class="'o-cf-rule-editor'">
```
rrahir: this class can be removed entirely, it's just used for tests and the 7 selectors that use it can make without

src/components/side_panel/data_validation/dv_editor/dv_editor.ts
```diff
@@ -25,9 +27,8 @@ import { Section } from "../../components/section/section";
 import { SelectMenu } from "../../select_menu/select_menu";
 
 interface Props {
-  rule: DataValidationRule | undefined;
-  onExit: () => void;
-  onCloseSidePanel?: () => void;
+  id: UID;
```
rrahir: nitpick - from an exterior point of view id is really not specific
```suggestion
  ruleId: UID;
```

src/actions/menu_items_actions.ts
```diff
@@ -549,7 +550,39 @@ export const FORMAT_PERCENT_ACTION = (env: SpreadsheetChildEnv) => setFormatter(
 // Side panel
 //------------------------------------------------------------------------------
 export const OPEN_CF_SIDEPANEL_ACTION = (env: SpreadsheetChildEnv) => {
-  env.openSidePanel("ConditionalFormatting", { selection: env.model.getters.getSelectedZones() });
+  const sheetId = env.model.getters.getActiveSheetId();
+  const zones = env.model.getters.getSelectedZones();
+  const rules = env.model.getters.getConditionalFormats(sheetId);
+  const ruleIds = Array.from(env.model.getters.getRulesSelection(sheetId, zones));
+  if (ruleIds.length === 0) {
```
rrahir: Just a thought but let's keep it as is for now,

I really don't like the feature, the same I dislike the already present automatic opening of a pre-existing rule. 
Opening a pre-exixting rule makes sense, if you have 20 of them on the same sheet but that' it. reste of the time, I might want to have an overview, which means I now have to create a CF in order to look at all the other existing ones. I don't think it makes much sense.

We should at least have two actions then. one that allows to see the cf and the other to insert them maybe?

src/actions/menu_items_actions.ts
```diff
@@ -549,7 +550,39 @@ export const FORMAT_PERCENT_ACTION = (env: SpreadsheetChildEnv) => setFormatter(
...
+    const cf: Omit<ConditionalFormat, "ranges"> = {
```
rrahir: this is a full on copy/paste of the action inside ConditionalFormatPreviewList - it should be refactored and probably use the deffault values defined in `ConditionalFormattingEditorStore.getDefaultRules`

src/components/side_panel/conditional_formatting/cf_preview_list/cf_preview_list.ts
```diff
@@ -58,8 +54,36 @@ export class ConditionalFormatPreviewList extends Component<Props, SpreadsheetCh
     });
   }
 
+  onAddConditionalFormat() {
+    const sheetId = this.env.model.getters.getActiveSheetId();
+    const zones = this.env.model.getters.getSelectedZones();
+    const cf: Omit<ConditionalFormat, "ranges"> = {
```
rrahir, line 60: nitpick, if you have a cool idea for a name, an alias could be nice since we use it at 3 different places already

src/components/side_panel/conditional_formatting/cf_editor/cf_editor_store.ts
rrahir: why do you need to handle undo operations? the command can come from pretty much anywhere and have nothing to do with your sidepanel. 
e.g.
- you create a cf
- take time to tune it, with formulas
- test the CF with a given input in a cell
- undo the change to the cell
-> your sidepanel closes
It seems quite unrelated

Why limit that feature to the CF ?

src/actions/menu_items_actions.ts
rrahir: I thnk the feature does make sense (though not by default). if you have 10 cf and 3 of them applying to a given cell and the result seems off on that particular cell, you have to go through your 10 cf to see which one applies to it, look at the condition, see if it applies  in order to find the problem. 
Limit the cells applying to the selection seems to be a nice debugging tool, but I agree it would be a pita by default, ... just like opening a CF by default on my selection when all I want to do is see the list of all CF's...

rmbh-odoo:
Latest push resolves merge conflicts. :smiling_face_with_tear:

src/components/side_panel/conditional_formatting/cf_preview/cf_preview.ts
```diff
@@ -11,23 +11,25 @@ import { useHighlightsOnHover } from "../../../helpers/highlight_hook";
...
   icons = ICONS;
-
   private ref = useRef("cfPreview");
 
   setup() {
     useHighlightsOnHover(this.ref, this);
   }
 
-  getPreviewImageStyle(): string {
+  get previewImageStyle(): string {
```
rrahir, line 32: 👍

src/components/side_panel/conditional_formatting/cf_editor/cf_editor_store.ts
rmbh-odoo: Hi @rrahir :wave: 
Previously, the CF list and the CF editor were tightly coupled. In the main CF panel, we used updateProps to look up the current CF, and if it was not found, we switched back to the list view.

This is no longer the case. The editor panel is no longer dependent on the full CF list, so there is no real reason to monitor all CFs or close the editor based on the current CF state.

So I tried to align with GSheets' behavior, when the user performs an undo, we switch back to the list panel. This is not a commonly occurring case in daily usage.

rrahir: We discussed this with Francois and Vincentand we think i's best to limit ther changes here to the minimum
- Remove the auto insertion of cf when clicking the "conditional formatting" button 
- remove the 'undo' handling and keep the current behaviour (if the cf is deleted, fallback to the cf list sidepanel)

these are the 2 functional changes I can think of which are not related to the strict split of sidepanels (the goal being to be able to open 2 of them side by side)

I am again sorry for this request; we should have caught it way earlier but we are not convinced about the functional improvement it brings and is quite out of scope

rmbh-odoo: > I am again sorry for this request; we should have caught it way earlier but we are not convinced about the functional improvement it brings and is quite out of scope

No worries, it’s my mistake. I added a small improvement during the refactor; we can handle it separately if needed. :slightly_smiling_face: 
Let me know if anything is wrong, I’m happy to make any required changes.

src/actions/menu_items_actions.ts
```diff
@@ -549,7 +549,17 @@ export const FORMAT_PERCENT_ACTION = (env: SpreadsheetChildEnv) => setFormatter(
 // Side panel
 //------------------------------------------------------------------------------
 export const OPEN_CF_SIDEPANEL_ACTION = (env: SpreadsheetChildEnv) => {
-  env.openSidePanel("ConditionalFormatting", { selection: env.model.getters.getSelectedZones() });
+  const sheetId = env.model.getters.getActiveSheetId();
+  const zones = env.model.getters.getSelectedZones();
+  const rules = env.model.getters.getConditionalFormats(sheetId);
+  const ruleIds = Array.from(env.model.getters.getRulesSelection(sheetId, zones));
```
rrahir:
```suggestion
  const ruleIds = env.model.getters.getRulesSelection(sheetId, zones);
```

tests/conditional_formatting/conditional_formatting_panel_component.test.ts
rrahir: tip for next time, try to make a separate commit for this kind of change. I must ay I don"t entirely follow the logic of the reorganisation; `Conditional formatting list panel` and `Conditional Formatting panel interactions` seem to do the same kind of stuff.

rrahir:
robodoo r+

rrahir:
robodoo rebase-ff

PR #7184, opened 2025-09-22
[FIX] SidePanel: reset initial props on model change

rrahir:
Currently, we store the initial props with which the sidepanel was open. This specifically occurs for charts sidepanel where we want the stat to update when we select another chart BUT we also want it to stay open when clicking the grid (hence - no active chart left).

However the moment we click on another chart than the first one, then it is supposed to become the new "default" chart.

How to reproduce:
- open the sidepanel of the first chart
- select the second chart
- delete the second chart

=> the sidepanel is still open and falls back on the first chart instead of closing.

Task: 5059484

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/side_panel/side_panel_store.ts
+++ b/src/components/side_panel/side_panel/side_panel_store.ts
@@ -10,6 +10,7 @@ interface OpenSidePanel {
   isOpen: true;
   props?: SidePanelProps;
   key?: string;
+  updatedProps?: SidePanelProps;
 }
 
 interface ClosedSidePanel {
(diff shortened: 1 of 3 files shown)


src/components/side_panel/side_panel/side_panel_store.ts
```diff
@@ -98,7 +99,11 @@ export class SidePanelStore extends SpreadsheetStore {
         props: panelProps,
       };
     } else {
-      return customComputeState(this.getters, panelProps);
+      const state = customComputeState(this.getters, panelProps);
+      if (state.isOpen) {
+        this.initialPanelProps = state.updatedProps ?? this.initialPanelProps;
```
LucasLefevre: it looks like without the new `updatedProps` concept, it works.
```suggestion
        this.initialPanelProps = state.props ?? this.initialPanelProps;
```
but maybe we should rename `this.initialPanelProps`

rrahir: I was afraid of breaking another sidepanel behaviour that would still rely on the actual initial panel props but it does not seem necessary indeed. I will renamte it currentPanel props

VincentSchippefilt:
robodoo r+

PR #7194, opened 2025-09-23
[FIX] RibbonMenu: reset scroll value on menu exploration

rrahir:
How to reproduce:

- Open the ribbon menu
- select "insert"
- scroll the menu completely
- select "Functions"

The scroll is persistent from the previous menu and we arrive in the `Functions` menu already scrolled, issue that is highlighted by the fact that scrollbars are hidden by defaut on mobile browsers.

This revision resets the scroll when ;oving from one menu section to the other.

### [FIX] RibbonMenu: hint the scrollable menu

Currently, it is not quite clear that the ribbon menu can be scrolled as
the scrollbars are not displayed by default on a mobile browser (or on
some desktop browsers like firefox).

This revision adds some style that will hint the possibility to scroll
either to the bottom or top of the menu depending on the current
scrolling value.

Task: 5106796

## Description:

Task: [5106796](https://www.odoo.com/odoo/2328/tasks/5106796)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/small_bottom_bar/ribbon_menu/ribbon_menu.scss
+++ b/src/components/small_bottom_bar/ribbon_menu/ribbon_menu.scss
@@ -12,6 +12,25 @@ $item-height: 40px;
   }
   .o-ribbon-menu-wrapper {
     max-height: 100%;
+
+    &.scroll-top::before {
+      width: 100%;
+      height: 15px;
+      content: "";
+      background: linear-gradient(to bottom, rgba(0, 0, 0, 0.2), transparent);
+      position: absolute;
+      pointer-events: none;
+    }
+    &.scroll-bottom::after {
+      width: 100%;
+      height: 15px;
+      content: "";
+      background: linear-gradient(to top, rgba(0, 0, 0, 0.2), transparent);
+      position: absolute;
+      pointer-events: none;
+      z-index: 1;
+      bottom: 0;
+    }
   }
   .o-menu-item {
     height: $item-height;
--- a/src/components/small_bottom_bar/ribbon_menu/ribbon_menu.ts
+++ b/src/components/small_bottom_bar/ribbon_menu/ribbon_menu.ts
@@ -1,4 +1,4 @@
-import { Component, useExternalListener, useRef, useState } from "@odoo/owl";
+import { Component, onMounted, useExternalListener, useRef, useState } from "@odoo/owl";
 import { Action } from "../../../actions/action";
 import { topbarMenuRegistry } from "../../../registries/menus";
 import { _t } from "../../../translation";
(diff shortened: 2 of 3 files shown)


src/components/small_bottom_bar/ribbon_menu/ribbon_menu.ts
```diff
@@ -51,6 +52,7 @@ export class RibbonMenu extends Component<RibbonMenuProps, SpreadsheetChildEnv>
       this.state.parentState = { ...this.state };
       this.state.menuItems = children;
       this.state.title = menu.name(this.env);
+      this.containerRef.el?.scrollTo({ top: 0 });
```
hokolomopo, line 56: If we want to be really fancy, we could store a scroll for each menu, so we keep the scroll when going back to the parent menu. Probably not worth tho

tests/bottom_bar/bottom_bar_component.test.ts
```diff
@@ -587,12 +583,6 @@ describe("BottomBar component", () => {
     });
 
     test("Spam click on the arrow button scrolls a lot", async () => {
-      let scrollTo = 0;
-      //@ts-ignore - scrollTo is not defined in JSDOM
-      sheetListEl.scrollTo = (arg: ScrollToOptions) => {
```
hokolomopo, line 592: Is scrollTo in JSDOM now ? 😮 

Edit: NEVERMIND you added it in the jest.setup. For a second, I was hoping JSDOM was getting good 😢

src/components/small_bottom_bar/ribbon_menu/ribbon_menu.scss
```diff
@@ -12,6 +12,25 @@ $item-height: 40px;
   }
   .o-ribbon-menu-wrapper {
     max-height: 100%;
+
+    &.scroll-top::before {
```
hokolomopo, line 16: css seems a bit broken, the left/right side of the menu also have a shadow for me ... 

image

src/components/small_bottom_bar/ribbon_menu/ribbon_menu.ts
rrahir: i thought about it but you need to render the old menu first and apply the scroll afterwards. the 2-step flow looked a bit shady which is why i didnt keep it

src/components/small_bottom_bar/ribbon_menu/ribbon_menu.scss
rrahir: apparently this behaves dfferently on miobile devices (perharps per browser/os ) i will change the strategy

PR #7208, opened 2025-09-25
19.0 border migration flda

fdamhaut:
## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


demo/main.js
```diff
@@ -210,6 +210,35 @@ class Demo extends Component {
...
+          }
+          const file = await input.files[0].text();
+          const data = JSON.parse(file);
+          this.leaveCollaborativeSession();
+          await fetch("http://localhost:9090/clear");
+          await this.initiateConnection(data);
+          stores.resetStores();
+          this.state.key = this.state.key + 1;
+
+          // note : the onchange won't be called if we cancel the dialog w/o selecting a file, so this won't be called.
```
rrahir:
```suggestion
         input.oncancel = () => input.remove();
         // alternatively
         input.oncancel = input.remove;
```
;-)

rrahir: it could be nice to add the export as well :)

src/migrations/migration_steps.ts
```diff
@@ -548,6 +548,32 @@ migrationStepRegistry
...
+          if (!borderId) continue;
+          if (!(borderId in borderType)) borderType[borderId] = [];
+          borderType[borderId].push(toZone(zoneXc));
+        }
+        const borders = {};
+        for (const borderId in borderType) {
+          const border = data.borders[borderId];
+          if (!border) continue;
+          const zones = recomputeZones(borderType[borderId]);
+          if (border.left || border.right) border.vertical = border.left || border.right;
```
anhe-odoo: Can we have `border.left = 0` or `border.top = 0` ? If so, we should compare to `undefined`.

demo/main.js
```diff
@@ -210,6 +210,53 @@ class Demo extends Component {
       icon: "o-spreadsheet-Icon.IMPORT_XLSX",
     });
 
+    topbarMenuRegistry.addChild("osheetExport", ["file"], {
```
anhe-odoo, line 213: Is it really needed to be merged ? Or it was only something you added to test your code ?

src/migrations/migration_steps.ts
fdamhaut: border are not value so it shouldn't be a problem

demo/main.js
fdamhaut: Oh rar already merged that in another PR since then, I forgot to remove it oops

src/plugins/core/borders.ts
```diff
@@ -475,9 +475,17 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
...
+        const borderType: Record<UID, Zone[]> = {};
         for (const zoneXc in sheet.borders) {
           const borderId = sheet.borders[zoneXc];
-          this.addBorder(sheet.id, toZone(zoneXc), data.borders[borderId]);
+          if (!borderId) continue;
+          if (!(borderId in borderType)) borderType[borderId] = [];
+          borderType[borderId].push(toZone(zoneXc));
+        }
+        for (const borderId in borderType) {
+          const zones = recomputeZones(borderType[borderId]);
```
LucasLefevre: do we need both this and the upgrade ?

src/migrations/migration_steps.ts
LucasLefevre: what if the zone is not a single cell ? this adds vertical

src/plugins/core/borders.ts
```diff
@@ -472,12 +472,55 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
...
   import(data: WorkbookData) {
     if (Object.keys(data.borders || {}).length) {
       for (const sheet of data.sheets) {
+        const borderType: Record<UID, Record<string, Zone[]>> = {};
         for (const zoneXc in sheet.borders) {
           const borderId = sheet.borders[zoneXc];
-          this.addBorder(sheet.id, toZone(zoneXc), data.borders[borderId]);
+          if (!borderId) {
+            continue;
+          }
```
LucasLefevre: we choose to keep the import as simple as possible, and move this where it's potentially needed (upgrade and xlsx import)

PR #7209, opened 2025-09-26
[FIX] spreadsheet: prevent sheet name edit from losing focus

dhrp-odoo:
## Description:

Steps to reproduce:
1. Create a new sheet.
2. Try renaming it to `Spreadsheet/dashboard` - error because of "/".
3. Click OK on the error dialog.
4. Click inside the sheet name near the "/" - edition is validated instead of moving the cursor; focus jumps to the default composer.

Issue:
The click event propagated to the parent component, which then focused the default composer and caused the loss of focus from the editable sheet name.

Fix:
Stop propagation of the click events when in editing mode, ensuring that:
- clicking positions the cursor correctly,
- editing is not validated prematurely,
- focus remains on the editable sheet name.

Task: [5109129](https://www.odoo.com/odoo/2328/tasks/5109129)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/bottom_bar/bottom_bar_sheet/bottom_bar_sheet.xml
+++ b/src/components/bottom_bar/bottom_bar_sheet/bottom_bar_sheet.xml
@@ -18,6 +18,7 @@
           t-ref="sheetNameSpan"
           t-esc="sheetName"
           t-on-pointerdown="(ev) => this.onMouseEventSheetName(ev)"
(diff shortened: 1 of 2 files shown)


hokolomopo:
@robodoo r+

Thanks for the PR :wave:

PR #7210, opened 2025-09-29
[IMP] format: align cell left for repeated char format

hokolomopo:
## Description

We have a "repeated character" format (e.g. "dd* ") that is mostly used to make sure the pivot headers are all aligned to the left, even when they contain numbers or dates.

The format repeats a character to fill the cell width. But the fit isn't perfect, and depending on the cell content it can be misaligned by a few pixels. This is very visible in pivot tables.

This commit makes it so the text alignment is always "left" when such a format is used, to avoid the misalignment. The solution isn't perfect, as now there will be misalignment to the right of the cell. But it is less problematic, and invisible when repeating spaces.

Task: [5122526](https://www.odoo.com/odoo/2328/tasks/5122526)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/format/format.ts
+++ b/src/helpers/format/format.ts
@@ -24,6 +24,7 @@ import {
   DateInternalFormat,
   InternalFormat,
   MAX_DECIMAL_PLACES,
+  MultiPartInternalFormat,
   NumberInternalFormat,
   TextInternalFormat,
   convertInternalFormatToFormat,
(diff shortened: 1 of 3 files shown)


LucasLefevre (approved):
robodoo r+

PR #7212, opened 2025-09-29
[IMP] evaluation: implement curly brackets

laa-odoo:
Task: [4735250](https://www.odoo.com/odoo/2328/tasks/4735250)

### DIFF
--- a/packages/o-spreadsheet-engine/src/formulas/formula_formatter.ts
+++ b/packages/o-spreadsheet-engine/src/formulas/formula_formatter.ts
@@ -260,6 +260,21 @@ function astToDoc(ast: AST): Doc {
         }
       }
       return wrapInParentheses(concat(splitArgsWithCommas(docs)), ast.value);
+
+    case "ARRAY": {
+      const rowDocs = ast.value.map((row) =>
+        concat(
+          row.map((value, index) =>
+            index === 0 ? astToDoc(value) : concat([", ", line(), astToDoc(value)])
+          )
+        )
+      );
+      const body = concat(
+        rowDocs.map((doc, index) => (index === 0 ? group(doc) : concat(["; ", line(), group(doc)])))
+      );
+      return wrapInBraces(body);
+    }
+
     case "UNARY_OPERATION":
       const operandDoc = astToDoc(ast.operand);
       const needParenthesis = ast.postfix
@@ -312,6 +327,10 @@ function wrapInParentheses(doc: Doc, functionName: undefined | string = undefine
   return group(concat(docToConcat));
 }
 
+function wrapInBraces(doc: Doc): Doc {
+  return group(concat(["{", nest(1, concat([line(), doc])), line(), "}"]));
+}
+
 /**
  * Converts an ast formula to the corresponding string
  */
@@ -339,6 +358,12 @@ export function astToFormula(ast: AST): string {
         ? `(${astToFormula(ast.operand)})`
         : astToFormula(ast.operand);
       return ast.value + rightOperand;
+    case "ARRAY":
+      return (
+        "{" +
+        ast.value.map((row) => row.map((cell) => astToFormula(cell)).join(",")).join(";") +
+        "}"
+      );
     case "BIN_OPERATION":
       const leftOperation = leftOperandNeedsParenthesis(ast)
         ? `(${astToFormula(ast.left)})`
--- a/packages/o-spreadsheet-engine/src/formulas/parser.ts
+++ b/packages/o-spreadsheet-engine/src/formulas/parser.ts
@@ -91,6 +91,11 @@ export interface ASTSymbol extends ASTBase {
   value: string;
 }
 
+export interface ASTArray extends ASTBase {
+  type: "ARRAY";
+  value: AST[][];
+}
+
 interface ASTEmpty extends ASTBase {
   type: "EMPTY";
   value: "";
@@ -101,6 +106,7 @@ export type AST =
   | ASTUnaryOperation
   | ASTFuncall
   | ASTSymbol
+  | ASTArray
   | ASTNumber
   | ASTBoolean
   | ASTString
@@ -219,6 +225,8 @@ function parseOperand(tokens: TokenList): AST {
         tokenStartIndex: current.tokenIndex,
         tokenEndIndex: rightParen.tokenIndex,
       };
+    case "LEFT_BRACE":
+      return parseArrayLiteral(tokens, current);
     case "OPERATOR":
       const operator = current.value;
       if (UNARY_OPERATORS_PREFIX.includes(operator)) {
@@ -274,6 +282,33 @@ function consumeOrThrow(tokens: TokenList, type, message?: string) {
     throw new BadExpressionError(message);
   }
   return token;
+}
+
+function parseArrayLiteral(tokens: TokenList, leftBrace: RichToken): ASTArray {
+  const rows: AST[][] = [];
+  let currentRow: AST[] = [parseExpression(tokens)]; // there must be at least one element
+
+  while (tokens.current?.type !== "RIGHT_BRACE") {
+    const nextToken = tokens.shift();
+    if (!nextToken) {
+      throw new BadExpressionError(_t("Missing closing brace"));
+    } else if (nextToken.type === "ARG_SEPARATOR") {
+      currentRow.push(parseExpression(tokens));
+    } else if (nextToken.type === "ARRAY_ROW_SEPARATOR") {
+      rows.push(currentRow);
+      currentRow = [parseExpression(tokens)];
+    } else {
+      throw new BadExpressionError(_t("Unexpected token: %s", nextToken.value));
+    }
+  }
+  const rightBrace = consumeOrThrow(tokens, "RIGHT_BRACE", _t("Missing closing brace"));
+  rows.push(currentRow);
+  return {
+    type: "ARRAY",
+    value: rows,
+    tokenStartIndex: leftBrace.tokenIndex,
+    tokenEndIndex: rightBrace.tokenIndex,
+  };
 }
 
 function parseExpression(tokens: TokenList, parent_priority: number = 0): AST {
@@ -376,6 +411,13 @@ function* astIterator(ast: AST): Iterable<AST> {
         yield* astIterator(arg);
       }
       break;
+    case "ARRAY":
+      for (const row of ast.value) {
+        for (const cell of row) {
+          yield* astIterator(cell);
+        }
+      }
+      break;
     case "UNARY_OPERATION":
       yield* astIterator(ast.operand);
       break;
@@ -396,6 +438,11 @@ export function mapAst<T extends AST["type"]>(
       return {
         ...ast,
         args: ast.args.map((child) => mapAst(child, fn)),
+      };
+    case "ARRAY":
+      return {
+        ...ast,
+        value: ast.value.map((row) => row.map((cell) => mapAst(cell, fn))),
       };
     case "UNARY_OPERATION":
       return {
--- a/packages/o-spreadsheet-engine/src/functions/module_array.ts
+++ b/packages/o-spreadsheet-engine/src/functions/module_array.ts
@@ -17,8 +17,80 @@ import {
   transposeMatrix,
 } from "./helpers";
 
+function stackHorizontally(
+  ranges: Arg[],
+  options?: { requireSameRowCount?: boolean }
+): Matrix<FunctionResultObject> | EvaluationError {
+  const matrices = ranges.map(toMatrix);
+  const nbRowsArr = matrices.map((m) => m?.[0]?.length ?? 0);
+  const nbRows = Math.max(...nbRowsArr);
+
+  if (options?.requireSameRowCount) {
+    const firstLength = nbRowsArr[0];
+    if (nbRowsArr.some((len) => len !== firstLength)) {
+      return new EvaluationError(
+        _t(
+          "All ranges in [[FUNCTION_NAME]] must have the same number of columns (got %s).",
+          nbRowsArr.join(", ")
+        )
+      );
+    }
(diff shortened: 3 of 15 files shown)


packages/o-spreadsheet-engine/src/formulas/formula_formatter.ts
```diff
@@ -332,20 +354,21 @@ export function astToFormula(ast: AST): string {
 function leftOperandNeedsParenthesis(operationAST: ASTOperation | ASTUnaryOperation): boolean {
   const mainOperator = operationAST.value;
   const leftOperation = "left" in operationAST ? operationAST.left : operationAST.operand;
+  if (leftOperation.type !== "BIN_OPERATION") {
```
anhe-odoo: Why have you changed these lines ? It does the same thing (except the assignment of leftOperator), doesn't it ?

packages/o-spreadsheet-engine/src/formulas/formula_formatter.ts
```diff
@@ -332,20 +354,21 @@ export function astToFormula(ast: AST): string {
...
 
 function rightOperandNeedsParenthesis(operationAST: ASTOperation | ASTUnaryOperation): boolean {
   const mainOperator = operationAST.value;
   const rightOperation = "right" in operationAST ? operationAST.right : operationAST.operand;
-  const rightPriority = OP_PRIORITY[rightOperation.value];
-  const mainPriority = OP_PRIORITY[mainOperator];
   if (rightOperation.type !== "BIN_OPERATION") {
     return false;
   }
+  const rightPriority = OP_PRIORITY[rightOperation.value];
```
anhe-odoo: Same questions as above

packages/o-spreadsheet-engine/src/formulas/parser.ts
```diff
@@ -276,6 +284,56 @@ function consumeOrThrow(tokens: TokenList, type, message?: string) {
...
+      tokens.shift();
+      if (expectValue) {
+        throw new BadExpressionError(_t("Unexpected empty array element"));
+      }
+      expectValue = true;
+      continue;
+    }
+    if (nextToken.type === "ARRAY_ROW_SEPARATOR") {
+      tokens.shift();
+      if (expectValue) {
```
anhe-odoo: What if we want to have an empty cell inside the row of an array ? Isn't there any case where we sohuld want it ?

src/formulas/tokenizer.ts
```diff
@@ -91,6 +96,11 @@ const parenthesis = {
   ")": { type: "RIGHT_PAREN", value: ")" },
 } as const;
 
+const braces = {
```
anhe-odoo: Nitpick: Maybe we can put these just before tokenizeBraces ?

src/helpers/pivot/pivot_composer_helpers.ts
```diff
@@ -86,7 +86,7 @@ export function insertTokenAfterLeftParenthesis(
  */
 export function extractFormulaIdFromToken(tokenAtCursor: EnrichedToken) {
   const idAst = tokenAtCursor.functionContext?.args[0];
-  if (!idAst || !["STRING", "NUMBER"].includes(idAst.type)) {
+  if (!idAst || (idAst.type !== "STRING" && idAst.type !== "NUMBER")) {
```
anhe-odoo: Nitpick: Another useless diff. If you ally want to add them, maybe make another commit ?

src/registries/auto_completes/pivot_auto_complete.ts
```diff
@@ -60,7 +60,10 @@ autoCompleteProviders.add("pivot_measures", {
       return [];
     }
     const pivotFormulaId = extractFormulaIdFromToken(tokenAtCursor);
-    const pivotId = this.getters.getPivotId(pivotFormulaId);
+    if (pivotFormulaId === undefined) {
```
anhe-odoo: All these changes should go in another commit with all the others minor changes you made to improve the code

tests/evaluation/parser.test.ts
```diff
@@ -102,6 +102,35 @@ describe("parser", () => {
...
+      ],
+    });
+  });
+
+  test("array literal throw error when empty entries", () => {
+    expect(() => parse("={1,,}")).toThrow("Unexpected empty array element");
+    expect(() => parse("={1;}")).toThrow("Unexpected empty array element");
+  });
+
+  test("array literal with mismatched row length not throws", () => {
```
anhe-odoo: array literal with mismatched row length DOES not throw

packages/o-spreadsheet-engine/src/formulas/formula_formatter.ts
laa-odoo: These are some minor changes made when I re-dived into the code to better understand what it does. But in fact this is refactoring and does not belong here. I am removing it as well as the others you mentioned. ;)

packages/o-spreadsheet-engine/src/formulas/parser.ts
laa-odoo: Here, I've chosen to arbitrarily replicate Google's behavior. But we can ask ourselves why they return an error directly to the parser:
In their mindset, curly brackets are used to aggregate objects if and only if their dimensions are compatible. An empty value here would correspond to an object without dimensions, which would therefore not be compatible with the other objects.
--> So they allow themselves to return the error directly to the parser to avoid unnecessary evaluation. 
(I suppose)

To the question, how to introduce empty cells: using the value 0, or a subsection composed of several 0, e.g.: 
`={ A1:B2, {0,0;0,0} }`

tests/evaluation/parser.test.ts
laa-odoo: thx

packages/o-spreadsheet-engine/src/formulas/formula_formatter.ts
laa-odoo: I didn't say anything; the change is necessary since I created a new AST (ASTArray) that no longer has the `value` attribute.

So, should we avoid this by keeping `value` to store the matrix and thus losing comprehension? It's up to the team to decide.

packages/o-spreadsheet-engine/src/formulas/formula_formatter.ts
laa-odoo: Same answer :p

src/helpers/pivot/pivot_composer_helpers.ts
laa-odoo: Same answer :)

src/registries/auto_completes/pivot_auto_complete.ts
laa-odoo: Same answer :)

src/formulas/tokenizer.ts
laa-odoo: thx

packages/o-spreadsheet-engine/src/formulas/formula_formatter.ts
anhe-odoo: Then shoudn't we also do the same change for `mainOperator` 2 lines above ?

src/helpers/pivot/pivot_composer_helpers.ts
anhe-odoo: How can this be the same "maybe we don't have value" answer ? Here we have exactly the same behavior, don't we ?

laa-odoo: https://github.com/odoo/o-spreadsheet/pull/7212/files#diff-db78c118c484e53350e22ce04fcc29b472cf06f367dd3daedafa3b52f1ebe78eR89

typescript fails to infer that with `includes` the type can only be "STRING" or "NUMBER". So it assumes it can also be "ARRAY" and the code fails to compile.

packages/o-spreadsheet-engine/src/formulas/formula_formatter.ts
laa-odoo: This is just so TypeScript can compile on
https://github.com/odoo/o-spreadsheet/pull/7212/files#diff-88a81a3427c2adbad1c2ca7ed98a682c8145fc3487db8eb01e60155b8d204043R361

`mainOperator` is defined as `ASTOperation`, so we can be sure that `value` exists on it.

demo/data.js
```diff
@@ -3783,6 +3783,7 @@ export const demoData = {
       dateFormat: "m/d/yyyy",
       timeFormat: "hh:mm:ss a",
       formulaArgSeparator: ",",
+      arrayRowSeparator: ";",
```
LucasLefevre: what else could it be ? does it really depends on the language ? or does it depends  more on the `formulaArgSeparator` ?

I think it could/should be removed from the locale since it can be derived from `locale.formulaArgSeparator` (In my current understanding)

packages/o-spreadsheet-engine/src/formulas/parser.ts
```diff
@@ -276,6 +284,56 @@ function consumeOrThrow(tokens: TokenList, type, message?: string) {
...
+  }
+  // Handle the closing brace
+  if (expectValue) {
+    throw new BadExpressionError(_t("Unexpected empty array element"));
+  }
+  // we could check here that all rows have the same size, but the real
+  // dimensions depend on the values inside, which could be a formula
+  // whose dimensions are not known yet.
+
+  const rightBrace = consumeOrThrow(tokens, "RIGHT_BRACE");
```
LucasLefevre:
```suggestion
  const rightBrace = consumeOrThrow(tokens, "RIGHT_BRACE", _t("Missing closing brace"));
```

packages/o-spreadsheet-engine/src/functions/module_array.ts
```diff
@@ -17,6 +17,78 @@ import {
...
+  ranges: Arg[],
+  options?: { requireSameRowCount?: boolean }
+): Matrix<FunctionResultObject> | EvaluationError {
+  const matrices = ranges.map(toMatrix);
+  const nbRowsArr = matrices.map((m) => m?.[0]?.length ?? 0);
+  const nbRows = Math.max(...nbRowsArr);
+
+  if (options?.requireSameRowCount) {
+    const firstLength = nbRowsArr[0];
+    if (!nbRowsArr.every((len) => len === firstLength)) {
```
LucasLefevre: Exit at the first with a different length, instead of checking all of them everytime
```suggestion
    if (nbRowsArr.some((len) => len !== firstLength)) {
```

packages/o-spreadsheet-engine/src/functions/module_array.ts
```diff
@@ -17,6 +17,78 @@ import {
...
+          nbColsArr.join(", ")
+        )
+      );
+    }
+  }
+
+  const nbRows = matrices.reduce((acc, m) => acc + (m?.[0]?.length ?? 0), 0);
+  const result: Matrix<FunctionResultObject> = Array(nbCols)
+    .fill([])
+    .map(() => Array(nbRows).fill({ value: null }));
```
LucasLefevre:
```suggestion
  const result: Matrix<FunctionResultObject> = generateMatrix(nbCols, nbRows, () => ({ value: null }));
```

packages/o-spreadsheet-engine/src/formulas/parser.ts
```diff
@@ -91,6 +91,11 @@ export interface ASTSymbol extends ASTBase {
   value: string;
 }
 
+export interface ASTArray extends ASTBase {
+  type: "ARRAY";
+  rows: AST[][];
```
LucasLefevre: I would name it `value` to avoid all the mlsqjfqsf required because all AST no longer has a `value` property
```suggestion
  value: AST[][];
```

laa-odoo:
thx @LucasLefevre for the suggestions

Think should be okay :)

src/components/side_panel/pivot/pivot_side_panel/pivot_side_panel_store.ts
```diff
@@ -255,8 +255,16 @@ export class PivotSidePanelStore extends SpreadsheetStore {
       const cell = this.getters.getCell(position);
       if (cell?.isFormula) {
         const pivotFunction = getFirstPivotFunction(cell.compiledFormula.tokens);
-        const pivotFormulaId = pivotFunction?.args[0]?.value;
-        if (pivotFunction && updatedPivotFormulaId === pivotFormulaId.toString()) {
+        const pivotArg = pivotFunction?.args[0];
+        const pivotFormulaId =
```
LucasLefevre: diff not useful anymore ?

src/registries/auto_completes/pivot_auto_complete.ts
```diff
@@ -115,7 +115,17 @@ autoCompleteProviders.add("pivot_group_fields", {
     } else {
       args = args.filter((ast, index) => index % 2 === 1); // keep only the field names
     }
-    const argGroupBys = args.map((ast) => ast?.value).filter(isDefined);
+    const argGroupBys = args
+      .map((ast) => {
+        if (!ast) {
```
LucasLefevre: The diff in this entire file is not useful anymore ?

tests/composer/formula_assistant_component.test.ts
```diff
@@ -322,7 +322,10 @@ describe("formula assistant", () => {
       });
 
       test("arguments separator is localized", async () => {
-        updateLocale(parent.env.model, { ...DEFAULT_LOCALE, formulaArgSeparator: ";" });
+        updateLocale(parent.env.model, {
+          ...DEFAULT_LOCALE,
+          formulaArgSeparator: ";",
+        });
```
LucasLefevre: useless reformatting

laa-odoo:
thx @LucasLefevre, 
Your comments have been taken into account.

PR #7227, opened 2025-09-30
[FIX] chart: wrong label format for date chart

hokolomopo:
## Description

Date charts would always have a datetime format for the labels, no matter the format of the data. They would also be truncated, which we probably don't want for dates.

The issue is that the date charts had a callback for the ticks using `truncateLabel`, and `getLabelForValue` which seems bugged in date charts[1].

This commit removes the tick callback for date charts, letting the luxon time adapter format the dates.

[1] https://github.com/chartjs/Chart.js/issues/12128

Task: [4908471](https://www.odoo.com/odoo/2328/tasks/4908471)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


rrahir:
robodoo r+

PR #7251, opened 2025-10-03
[FIX] composer: display placeholder

LucasLefevre:
## Description:

Commit d8db270a68995 mistakenly remove the css for the composer placeholder. Standalone composer were all missing their placeholder

Task: [5139499](https://www.odoo.com/odoo/2328/tasks/5139499)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/components/composer/composer/composer.css
```diff
@@ -40,6 +40,14 @@
       }
     }
 
+    .o-composer[placeholder]:empty:not(:focus):not(.active)::before {
+      content: attr(placeholder);
+      color: #bdbdbd;
```
LucasLefevre: none of the existing --os-gray-xxx seem to be a good fit for this. I'm just keeping the value as it was before

hokolomopo: If we really want to use our current grays, we can do something like `color: color-mix(in hsl, var(--os-gray-300), var(--os-gray-500));` to get a similar gray result.

In any case, this should be in the css variables. And while you're modifying this, could you apply the same color to the Fx icon ? The color is hard coded inside the SVG ATM.

rrahir:
robodoo r+

rrahir:
robodoo r+

PR #7252, opened 2025-10-06
[IMP] dv: autofocus input

LucasLefevre:
## Description:

Task: [4862354](https://www.odoo.com/odoo/2328/tasks/4862354)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/filters/filter_menu_criterion/filter_menu_criterion.xml
+++ b/src/components/filters/filter_menu_criterion/filter_menu_criterion.xml
@@ -13,6 +13,7 @@
       criterion="state.criterion"
       onCriterionChanged.bind="onCriterionChanged"
       disableFormulas="true"
+      autofocus="true"
     />
   </t>
 </templates>
--- a/tests/table/filter_menu_component.test.ts
+++ b/tests/table/filter_menu_component.test.ts
@@ -444,6 +444,7 @@ describe("Filter menu component", () => {
 
       await simulateClick(".o-filter-criterion-type");
       await simulateClick(".o-menu-item[data-name='containsText']");
+      expect(document.activeElement).toBe(fixture.querySelector(".o-dv-input input"));
       await setInputValueAndTrigger(".o-dv-input input", "hello");
       await simulateClick(".o-filter-menu-confirm");
 


tests/setup/jest_extend.ts
```diff
@@ -268,16 +268,26 @@ CancelledReasons: ${this.utils.printReceived(dispatchResult.reasons)}
...
-        ? ""
-        : `expect(target).toHaveClass(expected);\n\n${this.utils.printDiffOrStringify(
-            expectedClass,
-            element.className,
-            "Expected class",
-            "Received class",
-            false
-          )}`;
+    const message = () => {
+      if (this.isNot && pass) {
```
hokolomopo, line 272: can we do something generic ? All of the other custom matchers are also wrong ...

src/components/side_panel/criterion_form/criterion_input/criterion_input.ts
```diff
@@ -40,8 +40,8 @@ export class CriterionInput extends Component<Props, SpreadsheetChildEnv> {
   setup() {
     useEffect(
       () => {
-        if (this.props.focused) {
-          this.inputRef.el!.focus();
+        if (this.props.focused && this.inputRef.el) {
```
hokolomopo, line 44: ... can the input ref really be undefiend at useEffect ? Or is it just to avoid `!` ?

src/components/side_panel/data_validation/dv_editor/dv_editor.ts
```diff
@@ -60,6 +65,7 @@ export class DataValidationEditor extends Component<Props, SpreadsheetChildEnv>
 
   onCriterionTypeChanged(type: DataValidationCriterionType) {
     this.state.rule.criterion.type = type;
+    this.state.isTypeUpdated = true;
```
hokolomopo, line 70: we should probably have the same behaviour for CFs and in the data filter menu, no ? To confirm with frgi, but I don't see why implement it in DVs but not CFs

hokolomopo (commented):
👋

src/components/side_panel/criterion_form/criterion_input/criterion_input.ts
LucasLefevre: yes, if the input isn't displayed because it's a composer instead

tests/setup/jest_extend.ts
rrahir: let's do this in another task

src/components/side_panel/data_validation/dv_editor/dv_editor.xml
```diff
@@ -23,6 +23,7 @@
             t-key="state.rule.criterion.type"
             criterion="state.rule.criterion"
             onCriterionChanged.bind="onCriterionChanged"
+            autofocus="state.isTypeUpdated"
```
rrahir, line 26: shouldn't the t-key already force the destroy/reload of the component? 
I think autofocus="true" could do the trick

src/components/side_panel/data_validation/dv_editor/dv_editor.ts
rrahir: +1

src/components/side_panel/data_validation/dv_editor/dv_editor.xml
LucasLefevre: The first time the component is mounted, when opening the side panel, you don't want the input to be focused

src/components/side_panel/data_validation/dv_editor/dv_editor.ts
LucasLefevre: It's done :wink:

rrahir: i'm blind

src/components/side_panel/conditional_formatting/cf_editor/cell_is_rule_editor.xml
```diff
@@ -29,6 +29,7 @@
         t-key="state.rules.cellIs.operator"
         criterion="genericCriterion"
         onCriterionChanged.bind="onRuleValuesChanged"
+        autofocus="this.state.hasEditedCf"
```
rrahir, line 32: Do you think it makes sense to add it for other CF as well ?
For instance ColorScale has a set iof inputs -
image
 and so does Icon set (but that one is a bit far fetched
 
image

LucasLefevre: from a functional POV, probably yes it makes sense. From a technical POV, is it worth it ? I don't know...

rrahir: let's keep it as it is for now then

rrahir:
robodoo r+ rebase-ff

PR #6990, opened 2025-08-22
[IMP] dashboard: sort pivot from header

LucasLefevre:
## Description:

description of this task, what is implemented and why it is implemented that way.

Task: [TASK_ID](https://www.odoo.com/odoo/2328/tasks/TASK_ID)

## review checklist

- [ ] feature is organized in plugin, or UI components
- [ ] support of duplicate sheet (deep copy)
- [ ] in model/core: ranges are Range object, and can be adapted (adaptRanges)
- [ ] in model/UI: ranges are strings (to show the user)
- [ ] undo-able commands (uses this.history.update)
- [ ] multiuser-able commands (has inverse commands and transformations where needed)
- [ ] new/updated/removed commands are documented
- [ ] exportable in excel
- [ ] translations (\_t("qmsdf %s", abc))
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes
- [ ] doc is rebuild (npm run doc)
- [ ] status is correct in Odoo

PR #7264, opened 2025-10-08
Master lock sheet rar

rrahir:
## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/actions/action.ts
+++ b/src/actions/action.ts
@@ -96,7 +96,7 @@ let nextItemId = 1;
 
 export function createAction(item: ActionSpec): Action {
   const name = item.name;
-  const children = item.children;
+  const children = item.children || [];
   const description = item.description;
   const icon = item.icon;
   const secondaryIcon = item.secondaryIcon;
@@ -116,14 +116,15 @@ export function createAction(item: ActionSpec): Action {
           return undefined;
         }
       : undefined,
-    children: children
-      ? (env) => {
-          return children
-            .map((child) => (typeof child === "function" ? child(env) : child))
-            .flat()
-            .map(createAction);
-        }
-      : () => [],
+    children:
+      children.length > 0
+        ? (env) => {
+            return children
+              .map((child) => (typeof child === "function" ? child(env) : child))
+              .flat()
+              .map(createAction);
+          }
+        : () => [],
     isReadonlyAllowed: item.isReadonlyAllowed || false,
     separator: item.separator || false,
     icon: typeof icon === "function" ? icon : () => icon || "",
--- a/src/actions/insert_actions.ts
+++ b/src/actions/insert_actions.ts
@@ -12,6 +12,7 @@ export const insertRow: ActionSpec = {
     return number === 1 ? _t("Insert row") : _t("Insert %s rows", number.toString());
   },
   isVisible: (env) => ACTIONS.CAN_INSERT_HEADER(env, "ROW"),
+  isEnabled: (env) => !env.model.getters.isCurrentSheetLocked(),
   icon: "o-spreadsheet-Icon.INSERT_ROW",
 };
 
@@ -22,6 +23,7 @@ export const rowInsertRowBefore: ActionSpec = {
   },
   execute: ACTIONS.INSERT_ROWS_BEFORE_ACTION,
   isVisible: (env) => ACTIONS.CAN_INSERT_HEADER(env, "ROW"),
+  isEnabled: (env) => !env.model.getters.isCurrentSheetLocked(),
   icon: "o-spreadsheet-Icon.INSERT_ROW_BEFORE",
 };
 
@@ -56,6 +58,7 @@ export const rowInsertRowsAfter: ActionSpec = {
     return number === 1 ? _t("Insert row below") : _t("Insert %s rows below", number.toString());
   },
   isVisible: (env) => ACTIONS.CAN_INSERT_HEADER(env, "ROW"),
+  isEnabled: (env) => !env.model.getters.isCurrentSheetLocked(),
   icon: "o-spreadsheet-Icon.INSERT_ROW_AFTER",
 };
 
@@ -76,6 +79,8 @@ export const insertCol: ActionSpec = {
     return number === 1 ? _t("Insert column") : _t("Insert %s columns", number.toString());
   },
   isVisible: (env) => ACTIONS.CAN_INSERT_HEADER(env, "COL"),
+  isEnabled: (env) => !env.model.getters.isCurrentSheetLocked(),
+
   icon: "o-spreadsheet-Icon.INSERT_COL",
 };
 
@@ -145,6 +150,8 @@ export const insertCell: ActionSpec = {
     ACTIONS.IS_ONLY_ONE_RANGE(env) &&
     env.model.getters.getActiveCols().size === 0 &&
     env.model.getters.getActiveRows().size === 0,
+  isEnabled: (env) => !env.model.getters.isCurrentSheetLocked(),
+
   icon: "o-spreadsheet-Icon.INSERT_CELL",
 };
 
@@ -175,14 +182,14 @@ export const insertCellShiftRight: ActionSpec = {
 export const insertChart: ActionSpec = {
   name: _t("Chart"),
   execute: ACTIONS.CREATE_CHART,
-  isEnabled: (env) => !env.isSmall,
+  isEnabled: (env) => !env.isSmall && !env.model.getters.isCurrentSheetLocked(),
   icon: "o-spreadsheet-Icon.INSERT_CHART",
 };
 
 export const insertCarousel: ActionSpec = {
   name: _t("Carousel"),
   execute: ACTIONS.CREATE_CAROUSEL,
-  isEnabled: (env) => !env.isSmall,
+  isEnabled: (env) => !env.isSmall && !env.model.getters.isCurrentSheetLocked(),
   icon: "o-spreadsheet-Icon.CAROUSEL",
 };
 
@@ -198,7 +205,7 @@ export const insertImage: ActionSpec = {
   description: "Ctrl+O",
   execute: ACTIONS.CREATE_IMAGE,
   isVisible: (env) => env.imageProvider !== undefined,
-  isEnabled: (env) => !env.isSmall,
+  isEnabled: (env) => !env.isSmall && !env.model.getters.isCurrentSheetLocked(),
   icon: "o-spreadsheet-Icon.INSERT_IMAGE",
 };
 
@@ -207,13 +214,14 @@ export const insertTable: ActionSpec = {
   execute: ACTIONS.INSERT_TABLE,
   isVisible: (env) =>
     ACTIONS.IS_SELECTION_CONTINUOUS(env) && !env.model.getters.getFirstTableInSelection(),
-  isEnabled: (env) => !env.isSmall,
+  isEnabled: (env) => !env.isSmall && !env.model.getters.isCurrentSheetLocked(),
   icon: "o-spreadsheet-Icon.PAINT_TABLE",
 };
 
 export const insertFunction: ActionSpec = {
   name: _t("Function"),
   icon: "o-spreadsheet-Icon.FORMULA",
+  isEnabled: (env) => !env.model.getters.isCurrentSheetLocked(),
 };
 
 export const insertFunctionSum: ActionSpec = {
@@ -277,6 +285,7 @@ export const insertLink: ActionSpec = {
   name: _t("Link"),
   execute: ACTIONS.INSERT_LINK,
   icon: "o-spreadsheet-Icon.INSERT_LINK",
+  isEnabled: (env) => !env.model.getters.isCurrentSheetLocked(),
 };
 
 export const insertCheckbox: ActionSpec = {
@@ -297,6 +306,7 @@ export const insertCheckbox: ActionSpec = {
       },
     });
   },
+  isEnabled: (env) => !env.model.getters.isCurrentSheetLocked(),
   icon: "o-spreadsheet-Icon.INSERT_CHECKBOX",
 };
 
@@ -330,7 +340,7 @@ export const insertDropdown: ActionSpec = {
       },
     });
   },
-  isEnabled: (env) => !env.isSmall,
+  isEnabled: (env) => !env.isSmall && !env.model.getters.isCurrentSheetLocked(),
   icon: "o-spreadsheet-Icon.INSERT_DROPDOWN",
 };
 
--- a/src/components/bottom_bar/bottom_bar_sheet/bottom_bar_sheet.ts
+++ b/src/components/bottom_bar/bottom_bar_sheet/bottom_bar_sheet.ts
@@ -220,4 +220,8 @@ export class BottomBarSheet extends Component<Props, SpreadsheetChildEnv> {
     const color = this.env.model.getters.getSheet(this.props.sheetId).color || "";
     return cssPropertiesToCss({ background: color });
   }
+
+  get isSheetLocked() {
+    return this.env.model.getters.isSheetLocked(this.props.sheetId);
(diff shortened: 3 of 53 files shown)


packages/o-spreadsheet-engine/src/model.ts
```diff
@@ -37,7 +37,6 @@ import {
   CoreCommand,
```
hokolomopo, line 37: todo in commit message

tests/model/model.test.ts
```diff
@@ -100,7 +100,7 @@ describe("Model", () => {
     expect(getCellText(model, "A2")).toBe("copy&paste me");
   });
 
-  test("UI plugins cannot refuse core commands", () => {
```
hokolomopo, line 103: maybe add a test that they cannot refuse remote commands

packages/o-spreadsheet-engine/src/plugins/ui_feature/lock_sheet.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export class LockSheetPlugin extends UIPlugin {
+  static getters = ["isCurrentSheetLocked"] as const;
+
+  allowDispatch(cmd: Command): CommandResult | CommandResult[] {
+    /**
+     * isDashboard() implies that the user is not connected
+     * to other users and can do any operation and can do core modifications that will only affect them
+     * It is acceptable to bypass the locked sheet restriction in this case
+     */
```
hokolomopo, line 17: didn't we way that we would add a commit to ensure that ? we can have a dashboard mode with still a collaborative session

packages/o-spreadsheet-engine/src/types/commands.ts
```diff
@@ -221,6 +219,36 @@ export const readonlyAllowedCommands = new Set<CommandTypes>([
...
+  "REDO",
+  "REQUEST_UNDO",
+  "REQUEST_REDO",
+  "REPLACE_SEARCH",
+  "HIDE_SHEET",
+  "SHOW_SHEET",
+  "UPDATE_PIVOT",
+  "INSERT_NEW_PIVOT",
+  "ADD_PIVOT",
+  "REMOVE_PIVOT",
```
hokolomopo: are those even sheet dependent ?

src/actions/insert_actions.ts
```diff
@@ -145,6 +151,8 @@ export const insertCell: ActionSpec = {
     ACTIONS.IS_ONLY_ONE_RANGE(env) &&
     env.model.getters.getActiveCols().size === 0 &&
     env.model.getters.getActiveRows().size === 0,
+  isEnabled: (env) => !env.model.getters.isCurrentSheetLocked(),
```
hokolomopo: I'm almost tempted to do the opposite, and add a `isMenuItemEnabledOnLockedSheet` :see_no_evil:

src/components/menu/menu.ts
```diff
@@ -118,7 +118,13 @@ export class Menu extends Component<MenuProps, SpreadsheetChildEnv> {
   }
 
   isEnabled(menu: Action) {
-    if (menu.isEnabled(this.env)) {
+    const children = menu.children?.(this.env);
+    // to discuss, implies a new a way to define isEnabled for a menu with children
+    // could either apply to the parent or to the children and both sould counter act
+    const areChildrenEnabled = children.length
```
hokolomopo: this seems fine to me

src/actions/action.ts
```diff
@@ -97,7 +97,7 @@ let nextItemId = 1;
```
hokolomopo, line 105: the "%" format button is still clickable on locked sheet

hokolomopo: the delete table context menu neither

src/components/side_panel/chart/line_chart/line_chart_design_panel.xml
```diff
@@ -22,7 +22,7 @@
       </t>
     </GeneralDesignEditor>
     <SeriesWithAxisDesignEditor t-props="props"/>
-    <SidePanelCollapsible isInitiallyCollapsed="true" title.translate="Axes">
+    <SidePanelCollapsible isInitiallyCollapsed="chartSheetIsLocked" title.translate="Axes">
```
hokolomopo: Shouldn't it be the opposite ? Always expanded when the sheet is locked and the collapsible isn't clickable ?

src/components/side_panel/chart/main_chart_panel/main_chart_panel.xml
```diff
@@ -19,7 +19,11 @@
       </div>
 
       <t t-set="definition" t-value="getChartDefinition(this.chartId)"/>
-      <div class="o-panel-content overflow-y-auto" t-ref="panelContent">
+      <div
+        class="o-panel-content overflow-y-auto"
+        t-ref="panelContent"
+        t-att="{'inert': chartSheetIsLocked }"
```
hokolomopo: The way the panels are handled is very inconsistent:

- the "edit chart" menu item is disabled, but we can still double click in a chart to open the panel. If we shouldn't be able to handled the panel, making it inert is useless. And if we should be able to, don't disable the button
- opening the edit chart panel give me a notification saying that the sheet is locked ?
- if we make this panel inert, we should really make every panel inert. Why is the edit carousel/edit table not inert too ? 

IMO we should just disable all buttons that open chart/table/carousels side panels and not bother with inert and `get chartSheetIsLocked` and whatever

packages/o-spreadsheet-engine/src/types/commands.ts
```diff
@@ -221,6 +219,36 @@ export const readonlyAllowedCommands = new Set<CommandTypes>([
...
+  "SET_VIEWPORT_OFFSET",
+  "SET_FORMULA_VISIBILITY",
+  "SELECT_FIGURE", // not  sure
+  "EVALUATE_CHARTS",
+  "EVALUATE_CELLS",
+  "UNDO",
+  "REDO",
+  "REQUEST_UNDO",
+  "REQUEST_REDO",
+  "REPLACE_SEARCH",
```
hokolomopo, line 242: allowing replace on a locked sheet is strange. Does the command contains the sheetId if we replace on all sheet?

src/components/bottom_bar/bottom_bar_sheet/bottom_bar_sheet.ts
```diff
@@ -23,6 +32,20 @@ interface State {
   pickerOpened: boolean;
 }
 
+const getSheetLockAnimation = (
```
hokolomopo, line 36: not a big fan of the animation, maybe use lighter colors ? OR maybe a  wiggle animation on the lock icon ?

src/components/composer/top_bar_composer/top_bar_composer.ts
```diff
@@ -55,6 +55,9 @@ export class TopBarComposer extends Component<any, SpreadsheetChildEnv> {
       "max-height": `${COMPOSER_MAX_HEIGHT}px`,
       "line-height": "24px",
     };
+    if (this.env.model.getters.isCurrentSheetLocked()) {
+      style["pointer-events"] = "none";
+    }
```
hokolomopo, line 60: We probably could want to still be able to select the text for copy/paste and such. 

Or if we want to prevent any event, the composer should also be grayed out.

src/actions/action.ts
hokolomopo: zoom action in toolbar is also not grayed out, but we cannot change the zoom on a locked sheet

src/components/side_panel/chart/main_chart_panel/main_chart_panel.xml
rrahir: this means that you cannot consult the chart definition in that case, i need to check what we did for other structures but a chart with datasources could be consulted. I need to see if this can ben an odoo datasource chart thing only

src/components/bottom_bar/bottom_bar_sheet/bottom_bar_sheet.ts
rrahir: I have 0 opinion, i'd let the PO decide
Anithing is possible I guess

packages/o-spreadsheet-engine/src/types/commands.ts
rrahir: the plugin computes all positions eligible for a replace and will skip the ones on a locked sheet, the same it skips the cells which contain a formula and for which we don't specify the `searchFormula` argument.

src/actions/insert_actions.ts
rrahir: could be worth indeed

packages/o-spreadsheet-engine/src/types/commands.ts
hokolomopo: Sorry I wasn't clear at all. I was wondering why the command was in the `lockedSheetAllowedCommands` if the command don't contain a sheetId. But you use the activeSheetId in that case, instead of `cmd.sheetId`. Not sure if I really like that. In a betrer world, our local command would have the sheetId inside the command, instead of implying (sometime) the activeSheetId.

packages/o-spreadsheet-engine/src/plugins/ui_feature/lock_sheet.ts
rrahir: https://github.com/odoo/o-spreadsheet/issues/6990  and specially https://github.com/odoo/o-spreadsheet/pull/6990/commits/b6e7bd66dfcaeeaf8af6a147c3097c751f10d6b5 ;) I also didn't know this was merged

packages/o-spreadsheet-engine/src/model.ts
LucasLefevre: todo in commit message

src/actions/action.ts
```diff
@@ -47,6 +47,12 @@ export interface ActionSpec {
    */
   isReadonlyAllowed?: boolean;
   /**
+   * is the action allowed when the active sheet is locked
+   */
+  isEnabledOnLockedSheet?: boolean;
+  /**
+  /**
```
LucasLefevre: double `/**`

src/actions/action.ts
```diff
@@ -117,15 +124,17 @@ export function createAction(item: ActionSpec): Action {
...
-    children: children
-      ? (env) => {
-          return children
-            .map((child) => (typeof child === "function" ? child(env) : child))
-            .flat()
-            .map(createAction);
-        }
-      : () => [],
+    children:
+      children.length > 0
```
LucasLefevre: Why this change again ? Didn't we find the root cause some time ago ?

packages/o-spreadsheet-engine/src/types/commands.ts
```diff
@@ -220,6 +218,35 @@ export const readonlyAllowedCommands = new Set<CommandTypes>([
...
 ]);
 
+export const lockedSheetAllowedCommands = new Set<Command["type"]>([
+  // core commands
+  "LOCK_SHEET",
+  "UNLOCK_SHEET",
+  "MOVE_SHEET",
+  "DUPLICATE_SHEET",
+  "CREATE_SHEET",
+  "HIDE_SHEET",
```
LucasLefevre, line 226: The menu to hide a sheet is disabled

src/components/figures/figure/figure.ts
```diff
@@ -154,7 +156,13 @@ export class FigureComponent extends Component<Props, SpreadsheetChildEnv> {
 
   onKeyDown(ev: KeyboardEvent) {
     const keyDownShortcut = keyboardEventToShortcutString(ev);
-
+    const sheetId = this.env.model.getters.getFigureSheetId(this.props.figureUI.id);
+    if (!ev.composed && sheetId && this.env.model.getters.isSheetLocked(sheetId)) {
+      this.env.model.trigger("command-rejected", {
```
LucasLefevre: why is it needed to manually trigger here ?

tests/lock_sheet/lock_sheet_plugin.test.ts
```diff
@@ -0,0 +1,93 @@
...
+    beforeAll(() => {
+      model = new Model();
+      createSheet(model, { name: "Another sheet", position: 0 });
+      createChart(model, { type: "bar" }, "chartId");
+      addPivot(model);
+      lockSheet(model);
+      model.updateMode("dashboard");
+    });
+
+    test.each<Command["type"]>(readonlyCommands)(
```
LucasLefevre: if there's a single test, `beforeAll` could be moved inside the test, and there's no need for a `describe`

src/components/bottom_bar/bottom_bar_sheet/bottom_bar_sheet.ts
```diff
@@ -64,6 +112,9 @@ export class BottomBarSheet extends Component<Props, SpreadsheetChildEnv> {
       },
       () => [this.env.model.getters.getActiveSheetId()]
     );
+    onWillUnmount(() => {
```
LucasLefevre, line 118: it could happen that owl starts the rendering (and executes `setup`), but the rendering is later cancelled.
It would be a memory leak since `onWillUnmount` would never be called.
=> the call to `this.env.model.on(...)` should be done in `onMounted`

src/components/bottom_bar/bottom_bar_sheet/bottom_bar_sheet.ts
LucasLefevre: we should ask an artist to do the animation

src/components/spreadsheet/spreadsheet.ts
```diff
@@ -208,6 +214,16 @@ export class Spreadsheet extends Component<SpreadsheetProps, SpreadsheetChildEnv
 
   private bindModelEvents() {
```
LucasLefevre, line 214: Tiens, same here. If `setup/bindModelEvents` is called but the rendering is called => memory leak

rrahir: probably deserves a dedicated commit in prior versions ?wdyt?
EDIT: i'll still move the registering in this PR ofc

src/components/figures/figure/figure.ts
rrahir: if you don't handle the shortcut here, then it might be handler somewhere else in the Element tree. for instance, hit any key that does nothing on the figure but is handled in grid (like `Escape`), you will try dispatching some command (in the case of grid, it defaults to `CLEAN_CLIPBOARD_HIGHLIGHT`) which will fail since the sheet is locked and you end up with a random keystroke that fails, but try hitting "Y" and nothing will happen. 
the keystrokes in Figure should probably not bubble (until I realize it's used in some way)

rrahir: at the time, I think I went with the idea "no keystroke should have any effect" and went the hard way to block everything.

src/components/spreadsheet/spreadsheet.ts
LucasLefevre: > probably deserves a dedicated commit in prior versions ?wdyt?

yup

src/components/figures/figure/figure.ts
rrahir: I guess the "wall"' is a bad idea as it prevents a proper handling of Ctrl Z/Y

rrahir: I removed the code and added a stop propagation inthe figure component, seems better

src/actions/action.ts
rrahir: relicate of the first implementation of this PR: https://github.com/odoo/o-spreadsheet/pull/7489
the code is indeed irrelevant and was removed

LucasLefevre:
robodoo delegate+

rrahir:
robodoo r+

PR #7286, opened 2025-10-09
[IMP] pivot: add pivot design panel

hokolomopo:
## Description

This commit adds a new tab "Design" in the pivot side panel, allowing users to customize the appearance of pivot tables.

At the moment the panel only allow for the same customization as the arguments of the `=PIVOT()` function. The user can easily edit those settings in the side panel, and can override them in the formula if needed.

Task: [5150320](https://www.odoo.com/odoo/2328/tasks/5150320)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


packages/o-spreadsheet-engine/src/types/pivot.ts
```diff
@@ -237,8 +238,10 @@ export interface DimensionTreeNode {
...
-  displayColumnHeaders: boolean;
-  displayTotals: boolean;
-  displayMeasuresRow: boolean;
+export interface PivotStyle {
+  numberOfRows?: number;
+  numberOfColumns?: number;
+  displayTotals?: boolean;
+  displayColumnHeaders?: boolean;
+  displayMeasuresRow?: boolean;
 }
```
hokolomopo, line 249: Having everything optional makes it a bit harder to use, we need to do `style.displayTotals ?? DEFAULT_PIVOT_STYLE.displayTotals`

1) It's fine
2) rename variables so they are falsy by default (eg. `hideTotals` instead)
3) make everything mandatory. Need migration + make json/revisions a bit more heavy

note that 2/3 don't solve everything, it only works for boolean values not really for number values (I don't think we want to add Number.MAX_VALUE or w/e in the JSON)

pro-odoo: I would go for 1, it's fine, especially because as you say 2 and 3 are not perfect :)

pro-odoo:
robodoo retry

PR #7287, opened 2025-10-09
[FIX] charts: correctly humanize decimal numbers

hokolomopo:
## Description

We have an `humanizeNumbers` option for the chart to make large numbers more readable. The option is enabled by default on every chart, but it was not correctly handling decimal numbers. The function was always rounding them to the nearest integer, which does not make sense for small numbers (eg. 0.15 should not be displayed as 0).

Task: [5155591](https://www.odoo.com/odoo/2328/tasks/5155591)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/chart/zoomable_chart/design_panel.ts
+++ b/src/components/side_panel/chart/zoomable_chart/design_panel.ts
@@ -5,7 +5,6 @@ import {
   ZoomableChartDefinition,
 } from "../../../../types/index";
 import { Checkbox } from "../../components/checkbox/checkbox";
-import { ChartHumanizeNumbers } from "../building_blocks/humanize_numbers/humanize_numbers";
 import { ChartWithAxisDesignPanel } from "../chart_with_axis/design_panel";
 
 interface Props {
@@ -28,7 +27,6 @@ export class GenericZoomableChartDesignPanel<
   static components = {
     ...ChartWithAxisDesignPanel.components,
     Checkbox,
-    ChartHumanizeNumbers,
   };
 
   onToggleZoom(zoomable: boolean) {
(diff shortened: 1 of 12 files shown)


src/components/side_panel/chart/zoomable_chart/design_panel.ts
```diff
@@ -27,6 +28,7 @@ export class GenericZoomableChartDesignPanel<
   static components = {
     ...ChartWithAxisDesignPanel.components,
     Checkbox,
+    ChartHumanizeNumbers,
```
anhe-odoo: Not needed here, it's already in the `ChartWithAxisDesignPanel.components` :)

src/helpers/format/format.ts
```diff
@@ -682,14 +682,18 @@ function _roundFormat<T extends InternalFormat>(internalFormat: T): T {
...
-    locale
-  );
+  const numberValue = tryToNumber(value, locale);
+  if (numberValue === undefined) {
+    return "";
+  }
+  let numberFormat: Format | undefined = format;
+  if (Math.abs(numberValue) < 1000) {
+    const hasDecimal = numberValue % 1 !== 0;
+    numberFormat = !format && hasDecimal ? "0.####" : format;
```
anhe-odoo, line 692: Not sure we need to have 4 decimals, I would have gone to 2, but why not :)

hokolomopo: Ideally it's dependant on the number of digit before the comma (123.4, 12.34, 123.4). And ideally, we should also probably handle the different formats (a formatLargeNumber function, but for small ones)

Not sure it's worth tho, I went with the fast & dumb approach.

src/components/side_panel/chart/zoomable_chart/design_panel.ts
rrahir: up

src/helpers/format/format.ts
rrahir: 4 digits seems quite weird tbh
image

rrahir: Honestly cannot reproduce. I bet I did something weird by setting a forced format somehow

rrahir:
robodoo r+ rebase-ff

PR #7288, opened 2025-10-09
[FIX] Composer: Fix assistant position

rrahir:
The composer assistant position was updated with the introduction of the mobile mode but there are two issues with it:
- the change of CSS rule did not account for cells with the content set on multiple lines
- the structure of the bottom bar composer (in mobile mode) still hides the formula assistant

This revision reverts the assistant position and simply hides it in mobile. A future work will try to reintroduce it with a modified aspect that suits the mobile interface.

Task: 5155838

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/composer/composer/composer.ts
+++ b/src/components/composer/composer/composer.ts
@@ -234,7 +234,8 @@ export class Composer extends Component<CellComposerProps, SpreadsheetChildEnv> 
         assistantStyle["max-height"] = `${availableSpaceAbove - CLOSE_ICON_RADIUS}px`;
         // render top
         // We compensate 2 px of margin on the assistant style + 1px for design reasons
-        assistantStyle.transform = `translate(0, calc(-100% - ${cellHeight + 3}px))`;
+        assistantStyle.top = `-3px`;
+        assistantStyle.transform = `translate(0, -100%)`;
       }
       if (cellX + ASSISTANT_WIDTH > this.props.delimitation.width) {
         // render left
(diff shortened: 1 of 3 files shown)


src/components/composer/composer/composer.ts
```diff
@@ -234,7 +234,8 @@ export class Composer extends Component<CellComposerProps, SpreadsheetChildEnv>
         assistantStyle["max-height"] = `${availableSpaceAbove - CLOSE_ICON_RADIUS}px`;
         // render top
         // We compensate 2 px of margin on the assistant style + 1px for design reasons
-        assistantStyle.transform = `translate(0, calc(-100% - ${cellHeight + 3}px))`;
+        assistantStyle.top = `-3px`;
+        assistantStyle.transform = `translate(0, -100%)`;
```
hokolomopo, line 238: nitpick: there is a space between the composer and the assistant. Maybe it's good enough 🤷 
image

src/components/small_bottom_bar/small_bottom_bar.ts
```diff
@@ -101,7 +101,7 @@ export class SmallBottomBar extends Component<Props, SpreadsheetChildEnv> {
         height: this.focus === "inactive" ? "26px" : "fit-content",
         "max-height": `130px`,
       }),
-      showAssistant: !isIOS(), // Hide assistant on iOS as it breaks visually
+      showAssistant: false, // Hide assistant in small composer as it gets cropped ATM
```
hokolomopo, line 108: I don't see a task to fix this, am I blind or we do we not care about this ?

rrahir: Its on the task description  - it's not properly visible due to the html structure because we have several overflow:hidden that we use for other purpose. since it's not usable, its beter to deactivate it and try to reintroduce it as a banner for bottom bar specifically

hokolomopo: Yeah I see why you disabled it, but I didn't find a task to fix it properly in the future

LucasLefevre (commented):
robodoo r+

PR #7289, opened 2025-10-10
[FIX] composer: set editionMode to inactive when composer is unmounted

rmbh-odoo:
## Description:

Steps to reproduce:
- Open the CF side panel and edit a StandaloneComposer inside a color scale.
- Switch the sheet. The panel closes, so the composer is unmounted.
- Previously, the composer remained in editing mode after unmount.

Current behavior before PR:
- Unmounting the composer did not set editionMode to inactive.
- This caused focus issues in other parts of the UI.

- Unmounting the composer sets editionMode to inactive.
- Focus and selection are now properly released.

Task: [5149215](https://www.odoo.com/odoo/2328/tasks/5149215)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/composer/composer/abstract_composer_store.ts
+++ b/src/components/composer/composer/abstract_composer_store.ts
@@ -93,6 +93,7 @@ export abstract class AbstractComposerStore extends SpreadsheetStore {
     this.highlightStore.register(this);
     this.onDispose(() => {
       this.highlightStore.unRegister(this);
+      this._cancelEdition();
     });
   }
   protected abstract confirmEdition(content: string): void;
--- a/tests/conditional_formatting/conditional_formatting_panel_component.test.ts
+++ b/tests/conditional_formatting/conditional_formatting_panel_component.test.ts
@@ -1,5 +1,6 @@
 import { Component } from "@odoo/owl";
 import { Model } from "../../src";
+import { ComposerFocusStore } from "../../src/components/composer/composer_focus_store";
 import { ConditionalFormattingPanel } from "../../src/components/side_panel/conditional_formatting/conditional_formatting";
 import { toHex, toZone } from "../../src/helpers";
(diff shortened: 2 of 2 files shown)


src/components/composer/composer/abstract_composer_store.ts
```diff
@@ -167,6 +168,15 @@ export abstract class AbstractComposerStore extends SpreadsheetStore {
     this.cancelEditionAndActivateSheet();
   }
 
+  deactivateComposer() {
```
LucasLefevre: The composer could/should clean itself.
Something like this seems to be working

```patch
diff --git a/src/components/composer/composer/abstract_composer_store.ts b/src/components/composer/composer/abstract_composer_store.ts
index edc1e0bce..83fdf0854 100644
--- a/src/components/composer/composer/abstract_composer_store.ts
+++ b/src/components/composer/composer/abstract_composer_store.ts
@@ -93,6 +93,7 @@ export abstract class AbstractComposerStore extends SpreadsheetStore {
     this.highlightStore.register(this);
     this.onDispose(() => {
       this.highlightStore.unRegister(this);
+      this._cancelEdition();
     });
   }
   protected abstract confirmEdition(content: string): void;

```

tests/composer/composer_component.test.ts
```diff
@@ -1627,3 +1627,11 @@ describe("Double click selection in composer", () => {
     });
   });
 });
+
+test("Composer becomes inactive when unmounted", async () => {
```
LucasLefevre: Could you write a test that's showing the actual behavior (with the CF and sheet change) ?

src/components/composer/composer/abstract_composer_store.ts
rmbh-odoo: Oops, I didn’t know about this
Thanks for pointing it out! Learned something new today 😄

PR #7291, opened 2025-10-10
[FIX] clipboard: paste as value with empty format string

hokolomopo:
## Description

Our paste as value keeps the format from the origin cell, removing only the formula. But if the origin cell had an empty format string, it would copy this empty string instead of the evaluated cell format.

Note: we get empty string format as result of, for example, auto-fill an `=TODAY()` formula. It can be argued that we should have `undefined` core format rather than an empty string, but every other place in the codebase handles all falsy format values the same way. The clipboard was the only place where we would make the distinction between `""` and `undefined` with `??`.

Task: [5156459](https://www.odoo.com/odoo/2328/tasks/5156459)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/tests/clipboard/clipboard_plugin.test.ts
+++ b/tests/clipboard/clipboard_plugin.test.ts
@@ -1283,6 +1283,22 @@ describe("clipboard", () => {
     expect(getCellContent(model, "C3")).toBe("45.10%");
   });
 
+  test("paste as value works with both no core format and empty string core format", () => {
+    const model = new Model();
+    setCellContent(model, "D4", "=DATE(2024,6,5)");
+
+    copy(model, "D4");
+    paste(model, "E4", "asValue");
+    expect(getCell(model, "E4")).toMatchObject({ content: "45448", format: "m/d/yyyy" });
+
+    setFormat(model, "D4", ""); // An empty string format is equivalent to no format
+    expect(getCellContent(model, "D4")).toBe("6/5/2024");
+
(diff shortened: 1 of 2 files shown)


tests/clipboard/clipboard_plugin.test.ts
```diff
@@ -1283,6 +1283,22 @@ describe("clipboard", () => {
...
+  test("paste as value works with both no core format and empty string core format", () => {
+    const model = new Model();
+    setCellContent(model, "D4", "=DATE(2024,6,5)");
+
+    copy(model, "D4");
+    paste(model, "E4", "asValue");
+    expect(getCell(model, "E4")).toMatchObject({ content: "45448", format: "m/d/yyyy" });
+
+    setFormat(model, "D4", ""); // An empty string format is equivalent to no format
+    expect(getCellContent(model, "D4")).toBe("6/5/2024");
```
anhe-odoo, line 1295: Mega-mini-Nitpick: This line is not really needed here, is it ?

hokolomopo: Not really, but it checks that an empty string format doesn't mess up the cell format, in which case the test doesn't make sense anymore.

It doesn't hurt so I would leave it :shrug:

rrahir:
robodoo r+

PR #7337, opened 2025-10-15
[FIX] range: invalid sheet name with special character

hokolomopo:
## Description

If we have a range with an invalid sheet name, the sheet name is not escaped with quotes when it contains special characters when converting the range back to a string.

Task: [5125762](https://www.odoo.com/odoo/2328/tasks/5125762)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/tests/model/model_import_export.test.ts
+++ b/tests/model/model_import_export.test.ts
@@ -287,7 +287,7 @@ describe("Migrations", () => {
 
     const cfs = data.sheets[1].conditionalFormats;
     const rule1 = cfs[0].rule as ColorScaleRule;
-    expect(cfs[0].ranges).toEqual(["=sheetName_!A1:A2"]);
+    expect(cfs[0].ranges).toEqual(["'=sheetName_'!A1:A2"]);
     expect(rule1.minimum.value).toEqual("=sheetName_!B1");
     expect(rule1.midpoint?.value).toEqual("=sheetName_!B1");
     expect(rule1.maximum.value).toEqual("=sheetName_!B1");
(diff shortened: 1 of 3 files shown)


tests/model/model_import_export.test.ts
```diff
@@ -287,7 +287,7 @@ describe("Migrations", () => {
 
     const cfs = data.sheets[1].conditionalFormats;
     const rule1 = cfs[0].rule as ColorScaleRule;
-    expect(cfs[0].ranges).toEqual(["=sheetName_!A1:A2"]);
+    expect(cfs[0].ranges).toEqual(["'=sheetName_'!A1:A2"]);
```
hokolomopo, line 290: This whole test is both totally broken, and it highlight some problematic behaviors: 

**Broken because:**
- in the spreadsheet data, it uses `figure.type` instead of `figure.tag`. So Charts are never actually created.
- it creates a sheet with invalid characters in its name such as `sheetName?`, but the ranges of the charts/cfs are `=sheetName!`, an invalid sheet name that does not exist (notice the `=`, those are ranges and not formulas)

**Strange behavior highlighted**
- cfs formulas are stored as string in the plugin, and exported as it in the data. So they never go though the createRange/toRangeString moulinette. So sheet name with wrong character are never escaped.

Not sure if we want/need to fix that.

hokolomopo: @rrahir didn't we talk about this test sometime ? I kinda remember something like that

rrahir: I do recall that as well, I think we agreed to fix it in later versions (specifically when it came to introduce the carousel but I might be wrong) I don't recall what we put in place though

rrahir: https://github.com/odoo/o-spreadsheet/pull/6923

PR #7358, opened 2025-10-21
[PERF] evaluation: zonify evaluation

LucasLefevre:
This commit improves evaluation time from 5% to 20% on regular spreadsheet, depending on the spreadsheet. And even from 99% on some pathological cases

Benchmark: https://www.odoo.com/odoo/documents/d9dGIrjZRTSa9nf153wEIwo9e4c5

Let’s say you have two formulas that depend on the same reference:

A1: =C1 
A2: =C1

In the FormulaDependencyGraph, both dependencies are added independently. Specifically, the R-tree data structure ends up with two separate entries: C1 -> A1 and C1 -> A2.

Ideally, the internal R-tree structure should group these dependencies into a single node in the tree:

C1 -> [A1, A2] or even C1 -> [A1:A2] because they can be grouped into a single zone.

When this pattern occurs frequently, the R-tree becomes huged (and unbalanced) therefore sub-optimal.
- Huge: lots and lots of individual positions
- Unbalanced: this is common when using individual pivot formulas (e.g., for spreadsheet pivots), where each PIVOT.VALUE formula depends on the same range—the pivot dataset.

With this commit, the dependencies R-Tree maps bounding boxes to zones, instead of bounding boxes to positions.
To take advantage of this grouping, the entire evaluation process now uses zones as much as possible.

Task: 4936229

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_core_views/cell_evaluation/dependencies_r_tree.ts
+++ b/src/plugins/ui_core_views/cell_evaluation/dependencies_r_tree.ts
@@ -1,0 +1,140 @@
+import { BoundedRange, UID } from "../../../types";
+import { RTreeBoundingBox, RTreeItem, SpreadsheetRTree } from "./r_tree";
+import { RangeSet } from "./range_set";
+
+interface RangeSetItem {
+  boundingBox: RTreeBoundingBox;
+  data: RangeSet;
+}
+
+type RTreeRangeItem = RTreeItem<BoundedRange>;
+
+/**
+ * R-Tree of ranges, mapping zones (r-tree bounding boxes) to ranges (data of the r-tree item).
+ * Ranges associated to the exact same bounding box are grouped together
+ * to reduce the number of nodes in the R-tree.
+ */
+export class DependenciesRTree {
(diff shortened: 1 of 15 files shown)


src/plugins/ui_core_views/cell_evaluation/dependencies_r_tree.ts
```diff
@@ -0,0 +1,141 @@
...
+    if (zone.bottom > maxRow) {
+      maxRow = zone.bottom;
+    }
+  }
+  maxCol += 1;
+  maxRow += 1;
+
+  // in most real-world cases, we can use a fast numeric key
+  // but if the zones are too far right or bottom, we fallback to a slower string key
+  const maxPossibleKey =
```
fdamhaut:
```suggestion
  const maxPossibleKey = (((maxRow  +1) * maxCol +1)* maxRow + 1 ) * maxCol 
```

rrahir:
robodoo r+

PR #7362, opened 2025-10-22
[IMP] spreadsheet: implement dark mode for UI

hokolomopo:
### [IMP] spreadsheet: implement dark mode for UI

This commit implements a dark mode for spreadsheet for the whole UI,
excluding the canvas and the figures.

It is mostly implemented using the `light-dark()` CSS function, which
allows to define colors for both light and dark modes. The style
`color-scheme: dark/light` is set on the main spreadsheet component
based on its props.

Some notes/limitations:
- the composer token colors stay the same for dark/light mode, the
colors were simply a bit tweaked to look okay in both modes. Having them
change based on the mode would be hard mostly because we need to sync
the token colors with the highlight colors, but the highlight are drawn
on the canvas, which does not support `light-dark()`.
- box-shadow colors were not changed for dark mode. Maybe they should
be white instead of dark-gray, but they are not in odoo. That means
that we have mostly invisible shadows in dark mode, and need to add
a border to the popover to separate it from the background.

### [REF] composer: use opacity rather than color alpha

For the dark mode, we will start using colors using `light-dark()` CSS
variables. Those colors do not work with the `toHex` method.

To prepare for this change:
- compare raw color string in content_editable_helper instead of hex value
- use opacity style rather then changing the color alpha value

### [FIX] demo: log something when catching global errors

We have a external listener listening on `window.error`, but we don't
log anything when catching an error.

Task: [5082659](https://www.odoo.com/odoo/2328/tasks/5082659)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/components/composer/composer/composer.css
```diff
@@ -74,5 +74,12 @@
         font-weight: 600 !important;
       }
     }
+
+    /* ADRM TODO: sucks in light mode. Change that */
```
pro-odoo: 👀

src/components/popover/popover.css
```diff
@@ -4,7 +4,14 @@
     z-index: var(--os-components-importance-popover);
     overflow: auto;
     box-shadow: 1px 2px 5px 2px rgba(51, 51, 51, 0.15);
+    /* ADRM TODO */
```
pro-odoo: 👀

src/components/spreadsheet/spreadsheet.ts
```diff
@@ -64,6 +65,7 @@ export class Spreadsheet extends Component<SpreadsheetProps, SpreadsheetChildEnv
     notifyUser: { type: Function, optional: true },
     raiseError: { type: Function, optional: true },
     askConfirmation: { type: Function, optional: true },
+    colorScheme: { type: String, optional: true },
```
pro-odoo, line 64: Nitpick and feel free to ignore:
I would have move it to the env, for me it's similar to `isSmall` (which is in the env).

hokolomopo: You mean that the parent component of `Spreadsheet` would have to add it to the env ? Isn't it just more work/less intuitive setup for nothing ? The children of `Spreadsheet` don't need to know the color scheme ATM, so having it in the env isn't really useful.

src/components/popover/popover.css
hokolomopo: Forgot to remove the TODO, I ended up not changing box-shadows colors in dark mode

src/components/composer/composer/composer.css
hokolomopo: forgot to remove the TODO, it looks fine now :)

pro-odoo (approved):
robodoo r+

pro-odoo:
robodoo rebase-ff

PR #7377, opened 2025-10-27
[FIX] recompute_zones: accept unboundedZone on profilesContainsZone

laa-odoo:
Task: [5213859](https://www.odoo.com/odoo/2328/tasks/5213859)

### DIFF
--- a/packages/o-spreadsheet-engine/src/helpers/recompute_zones.ts
+++ b/packages/o-spreadsheet-engine/src/helpers/recompute_zones.ts
@@ -181,25 +181,31 @@ export function modifyProfiles<T extends Zone | UnboundedZone>( // export for te
 export function profilesContainsZone(
   profilesStartingPosition: number[],
   profiles: Map<number, number[]>,
-  zone: Zone
+  zone: Zone | UnboundedZone
 ): boolean {
   const leftValue = zone.left;
   const rightValue = zone.right;
-  const topValue = zone.top;
-  const bottomValue = zone.bottom + 1;
   const leftIndex = binaryPredecessorSearch(profilesStartingPosition, leftValue, 0);
-  const rightIndex = binaryPredecessorSearch(profilesStartingPosition, rightValue, leftIndex);
-  if (leftIndex === -1 || rightIndex === -1) {
-    return false;
-  }
+  const rightIndex =
+    rightValue === undefined
+      ? profilesStartingPosition.length - 1
+      : binaryPredecessorSearch(profilesStartingPosition, rightValue, leftIndex);
+  /**
+   * The `profilesStartingPosition` array always contains at least the value `0` at its first position,
+   * ensuring that applying `binaryPredecessorSearch` will always return a valid index.
+   * Therefore, it is not necessary to check if the result of `binaryPredecessorSearch` equals `-1`.
+   */
+  const topValue = zone.top;
+  const bottomValue = zone.bottom === undefined ? undefined : zone.bottom + 1;
   for (let i = leftIndex; i <= rightIndex; i++) {
     const profile = profiles.get(profilesStartingPosition[i])!;
-    const topPredIndex = binaryPredecessorSearch(profile, topValue, 0, true);
-    const bottomSuccIndex = binarySuccessorSearch(profile, bottomValue, 0, true);
+    const topPredIndex = binaryPredecessorSearch(profile, topValue, 0);
     if (topPredIndex === -1 || topPredIndex % 2 !== 0) {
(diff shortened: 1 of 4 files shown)


packages/o-spreadsheet-engine/src/helpers/recompute_zones.ts
```diff
@@ -181,25 +181,31 @@ export function modifyProfiles<T extends Zone | UnboundedZone>( // export for te
...
-  }
+  const rightIndex =
+    rightValue === undefined
+      ? profilesStartingPosition.length - 1
+      : binaryPredecessorSearch(profilesStartingPosition, rightValue, leftIndex);
+  /**
+   * The `profilesStartingPosition` array always contains at least the value `0` at its first position,
+   * ensuring that applying `binaryPredecessorSearch` will always return a valid index.
+   * Therefore, it is not necessary to check if the result of `binaryPredecessorSearch` equals `-1`.
+   */
```
anhe-odoo, line 197: Even if you don't have to check, isn't it a change of behavior ? The check was to make sure the `binaryPredecessorSearch` has found something, not only that the value is != -1. If we make sure we always return, at least, 0, won't the 0 value be kind of wrong sometimes ?

packages/o-spreadsheet-engine/src/helpers/recompute_zones.ts
```diff
@@ -181,25 +181,31 @@ export function modifyProfiles<T extends Zone | UnboundedZone>( // export for te
...
-    const topPredIndex = binaryPredecessorSearch(profile, topValue, 0, true);
-    const bottomSuccIndex = binarySuccessorSearch(profile, bottomValue, 0, true);
+    const topPredIndex = binaryPredecessorSearch(profile, topValue, 0);
     if (topPredIndex === -1 || topPredIndex % 2 !== 0) {
       return false;
     }
-    if (topValue < profile[topPredIndex] || bottomValue > profile[bottomSuccIndex]) {
+    const bottomSuccIndex =
+      bottomValue === undefined ? profile.length : binarySuccessorSearch(profile, bottomValue, 0);
+    if (topPredIndex + 1 !== bottomSuccIndex) {
```
anhe-odoo, line 208: Why did you change this line ? (Probably a stupid question but I don't understand :/ )

tests/helpers/zone_set.test.ts
```diff
@@ -32,6 +32,70 @@ describe("ZoneSet", () => {
     expect(Array.from(set)).toEqual([]);
   });
 
+  test("add and remove single unbounded zone: B2:3", () => {
+    const set = new ZoneSet();
+    set.add(toUnboundedZone("B2:3"));
+    expect(set.has(toUnboundedZone("B2"))).toBe(true);
```
anhe-odoo: `expect(set.has(toUnboundedZone("B2"))).toBTruthy();` and same for the other line with Truthy or Falsy

packages/o-spreadsheet-engine/src/helpers/recompute_zones.ts
laa-odoo: Except for the case of `rightValue`, which is new here (and which we now handle with `profilesStartingPosition.length - 1`) --> No, this doesn't constitute a change in behavior because the removed part of the code was never executed (provided the input field is valid).

packages/o-spreadsheet-engine/src/helpers/recompute_zones.ts
laa-odoo: Here, the two previous conditions (`topValue < profile[topPredIndex]` and `bottomValue > profile[bottomSuccIndex]`) were not sufficient to ensure that the area fell within the current profile. Because a profile can consist of subsections, and these conditions failed to pass the case where a zone straddled two subsections:
https://github.com/odoo/o-spreadsheet/pull/7377/files#diff-6706ec8ed56f19533f131a8c007021c40fdba10d6d9d1780313f2b3300e1a358R291

The new condition resolves this by ensuring that `topPredIndex` and `bottomSuxxIndex` are indeed registered within the same profile subsection (in other words, that the indexes are consecutive here).

tests/helpers/zone_set.test.ts
laa-odoo: thx

packages/o-spreadsheet-engine/src/helpers/recompute_zones.ts
laa-odoo: will be the commit message :)

PR #7378, opened 2025-10-27
[FIX] Charts: Ensure Chart js extension are loaded on chart creation

rrahir:
When calling the method chartToImage, the chartJs extensions might not be loaded as we load them conditionally when mounting a chart. Hence, calling `ChartToImage` when there are no visible charts in the viewport or if we just instantiate a model without loading a component `Spreadsheet`, the extensions will be missing.

Right now, the plugins/extensions are not fundamental to convert a chart to an image but this becomes problematic with the arrival of future charts (for instance Funnel).

Task: 5214007

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/figures/charts/chart_ui_common.ts
+++ b/src/helpers/figures/charts/chart_ui_common.ts
@@ -1,4 +1,9 @@
 import type { ChartConfiguration, ChartOptions } from "chart.js";
+import {
+  areChartJSExtensionsLoaded,
+  registerChartJSExtensions,
+  unregisterChartJsExtensions,
+} from "../../../components/figures/chart/chartJs/chart_js_extension";
 import { Figure } from "../../../types";
 import { GaugeChartRuntime, ScorecardChartRuntime } from "../../../types/chart";
 import { ChartRuntime } from "../../../types/chart/chart";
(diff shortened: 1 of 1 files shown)


rrahir:
@robodoo help

rrahir:
robodoo fw=no

src/helpers/figures/charts/chart_ui_common.ts
```diff
@@ -39,27 +44,33 @@ export function chartToImage(
...
     drawGaugeChart(canvas, runtime as GaugeChartRuntime);
-    const imgContent = canvas.toDataURL();
+    imgContent = canvas.toDataURL();
     div.remove();
-    return imgContent;
   }
-  return undefined;
+  if (!extensionsLoaded) {
+    unregisterChartJsExtensions();
+  }
```
hokolomopo: this seems wrong no ? this will unload the chartJExtensions if they were loaded before calling this helper, and we only converted gauge/scorecards. The load/unload should probably be fully inside the `if("chartJsConfig" in runtime)`

rrahir: true

LucasLefevre (commented):
robodoo r+

PR #5287, opened 2024-12-02
[IMP] filters: modernize look

LucasLefevre:
## Description:

Convert the data filter popover to the new style.

Task: [4358098](https://www.odoo.com/odoo/2328/tasks/4358098)

## review checklist

- [ ] feature is organized in plugin, or UI components
- [ ] support of duplicate sheet (deep copy)
- [ ] in model/core: ranges are Range object, and can be adapted (adaptRanges)
- [ ] in model/UI: ranges are strings (to show the user)
- [ ] undo-able commands (uses this.history.update)
- [ ] multiuser-able commands (has inverse commands and transformations where needed)
- [ ] new/updated/removed commands are documented
- [ ] exportable in excel
- [ ] translations (\_t("qmsdf %s", abc))
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes
- [ ] doc is rebuild (npm run doc)
- [ ] status is correct in Odoo

PR #7380, opened 2025-10-29
[FIX] filter menu: truncate long filter values

hokolomopo:
## Description

Long values in the filter menu were not truncated, causing layout issues. There was a `text-truncate` class, but it was added to the `<label>` element instead of the `<span>` containing the text.

Task: [5219611](https://www.odoo.com/odoo/2328/tasks/5219611)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/filters/filter_menu_item/filter_menu_value_item.xml
+++ b/src/components/filters/filter_menu_item/filter_menu_value_item.xml
@@ -14,8 +14,9 @@
         name="value.toString()"
         value="this.props.isChecked"
         onChange="this.props.onClick"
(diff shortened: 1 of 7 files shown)


rrahir:
@hokolomopo 
this should  probably target sas-18.2, 
the truncate issue seems to stem in https://github.com/odoo/o-spreadsheet/pull/5287 which was aded in saas-18.2, and I could reproduce the bug in 18.2 as well

rrahir:
robodoo r+

PR #7381, opened 2025-10-29
[FIX] BottomBarSheet: sheet name should update on foreign changes

rrahir:
How to reproduce:
On Firefox,
- double click the bottom bar to rename a sheet
- Undo the change (button or through Ctrl-Z)

=> the sheetName is not rolled back to its previous value

The issue seems to lie in the fact that in FF, setting a t-esc on an Element that was previously `contenteditable=true` creates weird behaviour. I suspect some internal state of the div that is not cleared.

On the other hand, changing the contenteditable state of the span element might not be the best idea and one could consider that it's safer to simply regenerate the span altogether when switching editing state.

This commit takes the last suggested approach.

Task: 5016252

## Description:

Task: [5016252](https://www.odoo.com/odoo/2328/tasks/5016252)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/tests/bottom_bar/bottom_bar_component.test.ts
+++ b/tests/bottom_bar/bottom_bar_component.test.ts
@@ -67,6 +67,10 @@ async function mountBottomBar(
   return { parent: parent as Parent, model, env };
 }
 
+function getSheetNameSpan(): HTMLSpanElement | null {
+  return fixture.querySelector<HTMLSpanElement>(".o-sheet-name");
+}
+
 describe("BottomBar component", () => {
   test("simple rendering", async () => {
     await mountBottomBar();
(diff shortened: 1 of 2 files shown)


tests/bottom_bar/bottom_bar_component.test.ts
```diff
@@ -352,8 +356,9 @@ describe("BottomBar component", () => {
 
       sheetName.innerHTML = HTML;
       await keyDown({ key: "Enter" });
-
-      expect(sheetName.getAttribute("contenteditable")).toEqual("false");
+      expect(
+        fixture.querySelector<HTMLElement>(".o-sheet-name")!.getAttribute("contenteditable")
```
hokolomopo: use your helper getSheetNameSpan :)

pro-odoo (approved):
robodoo r+

PR #7384, opened 2025-10-29
[FIX] pivot: error on removing max number of columns/rows

hokolomopo:
## Description

When deleting the pivot style numberOfRows or numberOfColumns in the pivot design panel, the pivot would be in error. This was because we didn't handle the cases where the content on the input was not a number, and would set `NaN` values.

Task: [5220970](https://www.odoo.com/odoo/2328/tasks/5220970)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


LucasLefevre (approved):
robodoo r+

PR #7386, opened 2025-10-30
[IMP] pivot: add tabular form

hokolomopo:
## Description

This commit adds a new setting for pivot: whether to use tabular form. In tabular form, each level of row header is displayed in a separate column.

Task: [4794334](https://www.odoo.com/odoo/2328/tasks/4794334)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/packages/o-spreadsheet-engine/src/helpers/pivot/table_spreadsheet_pivot.ts
+++ b/packages/o-spreadsheet-engine/src/helpers/pivot/table_spreadsheet_pivot.ts
@@ -64,6 +64,7 @@ interface CollapsiblePivotTableColumn extends PivotTableColumn {
 export class SpreadsheetPivotTable {
   readonly columns: CollapsiblePivotTableColumn[][];
   rows: PivotTableRow[];
+  readonly rowFields: string[];
   readonly measures: string[];
   readonly fieldsType: Record<string, string | undefined>;
   readonly maxIndent: number;
@@ -87,9 +88,7 @@ export class SpreadsheetPivotTable {
       columns = this.removeCollapsedColumns(columns, measures, collapsedDomains.COL);
     }
     this.columns = columns.map((cols) => {
-      // offset in the pivot table
-      // starts at 1 because the first column is the row title
-      let offset = 1;
+      let offset = 0;
       return cols.map((col) => {
         col = { ...col, offset };
         offset += col.width;
@@ -98,6 +97,9 @@ export class SpreadsheetPivotTable {
     });
 
     this.rows = rows.filter((row) => !this.isParentCollapsed(collapsedDomains.ROW, row));
+    const numberOfRowGroupings = Math.max(...rows.map((row) => row.fields.length));
+    const rowAtMaxDepth = rows.find((row) => row.fields.length === numberOfRowGroupings);
+    this.rowFields = rowAtMaxDepth ? [...rowAtMaxDepth.fields] : [];
     this.maxIndent = Math.max(...this.rows.map((row) => row.indent));
     this.rowTree = lazy(() => this.buildRowsTree());
     this.colTree = lazy(() => this.buildColumnsTree());
@@ -159,13 +161,23 @@ export class SpreadsheetPivotTable {
 
   private getSkippedRows(pivotStyle: Required<PivotStyle>) {
     const skippedRows: Set<number> = new Set();
+    const colHeadersHeight = this.getColHeadersHeight();
     if (!pivotStyle.displayColumnHeaders) {
-      for (let i = 0; i < this.columns.length - 1; i++) {
+      for (let i = 0; i < colHeadersHeight - 1; i++) {
         skippedRows.add(i);
       }
     }
     if (!pivotStyle.displayMeasuresRow) {
-      skippedRows.add(this.columns.length - 1);
+      skippedRows.add(colHeadersHeight - 1);
+    }
+    // Skip sub-total rows in tabular form
+    if (pivotStyle.tabularForm) {
+      for (let i = 0; i < this.rows.length; i++) {
+        const indent = this.rows[i].indent;
+        if (indent !== 0 && indent !== this.maxIndent) {
+          skippedRows.add(i + colHeadersHeight);
+        }
+      }
     }
     return skippedRows;
   }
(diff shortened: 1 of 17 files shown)


packages/o-spreadsheet-engine/src/registries/icons_on_cell_registry.ts
```diff
@@ -139,14 +139,15 @@ iconsOnCellRegistry.add("conditional_formatting", (getters, position) => {
...
     return undefined;
   }
   const pivotCell = getters.getPivotCellFromPosition(position);
-  const pivotId = getters.getPivotIdFromPosition(position);
+  const definition = getters.getPivotCoreDefinition(pivotId);
+  const tabularForm = definition.style?.tabularForm ?? DEFAULT_PIVOT_STYLE.tabularForm;
 
-  if (pivotCell.type === "HEADER" && pivotId && pivotCell.domain.length) {
-    const definition = getters.getPivotCoreDefinition(pivotId);
+  if (!tabularForm && pivotCell.type === "HEADER" && pivotId && pivotCell.domain.length) {
```
LucasLefevre, line 130: The icon is still displayed when it's not in tabular form in the definition, but it is in the PIVOT formula (and nothing works)

hokolomopo: Yeah .... I think I'll wait for the pivot style PR before fixing this, in there we should have a strategy to get the actual pivot style for a pivot cell (+ the pivot style will need to be adapted for tabular form)

packages/o-spreadsheet-engine/src/helpers/table_helpers.ts
```diff
@@ -281,24 +281,42 @@ function getTableElementZones(
...
-      for (const row of tableMetaData.mainSubHeaderRows || []) {
-        zones.push({ top: row, bottom: row, left: 0, right: tableMetaData.numberOfCols - 1 });
+      if (tableMetaData.isTabular) {
+        for (const col of tableMetaData.mainSubHeaderRows || []) {
+          zones.push({ top: headerRows, bottom: lastRow - totalRows, left: col, right: col });
+        }
+      } else {
+        for (const row of tableMetaData.mainSubHeaderRows || []) {
+          zones.push({ top: row, bottom: row, left: 0, right: numberOfCols - 1 });
+        }
```
hokolomopo, line 291: I'm half of a mind to do something extensible for table styles instead of handling all the stuff there, and having to pass isTabular and mainSubHeaderRows. But that might be more complexity for not much result.

src/helpers/pivot/pivot_menu_items.ts
```diff
@@ -371,7 +371,12 @@ export function sortPivot(
 ) {
   const pivotId = env.model.getters.getPivotIdFromPosition(position);
   const pivotCell = env.model.getters.getPivotCellFromPosition(position);
-  if (pivotCell.type === "EMPTY" || pivotCell.type === "HEADER" || !pivotId) {
+  if (
+    pivotCell.type === "EMPTY" ||
+    pivotCell.type === "HEADER" ||
+    pivotCell.type === "ROW_GROUP_NAME" ||
+    !pivotId
+  ) {
```
pro-odoo, line 382: Nitpick, wouldn't it be easier to read like that:
```js
if (!pivotId || !["MEASURE_HEADER", "VALUE"].includes(pivotCell.type))
```
With that, we keep a coherence with L364

src/components/side_panel/pivot/pivot_side_panel/pivot_design_panel/pivot_design_panel.xml
```diff
@@ -55,6 +55,16 @@
               />
             </div>
           </div>
+          <div class="row mb-2 align-items-center">
+            <div class="col-6">Use tabular form:</div>
+            <div class="col-6 d-flex align-items-center">
+              <Checkbox
+                name="'tabularForm'"
+                value="pivotStyle.tabularForm ?? defaultStyle.tabularForm"
```
pro-odoo, line 63: Shouldn't we start to use `this` ? In order to ease the future migration to owl3

pro-odoo (commented):
Didn't read the whole diff, but I added a point in the task pad

packages/o-spreadsheet-engine/src/types/pivot.ts
```diff
@@ -154,6 +154,11 @@ export interface PivotHeaderCell {
   dimension: Dimension;
 }
 
+export interface PivotGroupNameCell {
+  type: "ROW_GROUP_NAME";
+  rowField: string;
```
pro-odoo, line 159: Why don't we store directly the displayName here ?

packages/o-spreadsheet-engine/src/helpers/pivot/table_spreadsheet_pivot.ts
```diff
@@ -273,20 +322,28 @@ export class SpreadsheetPivotTable {
...
+  private getColMeasure(colIndex: number) {
+    const domain = this.getColHeaderDomain(colIndex, this.getColHeadersHeight() - 1);
     const measure = domain?.at(-1)?.value;
     if (measure === undefined || measure === null) {
       throw new Error("Measure is missing");
     }
     return measure.toString();
   }
 
+  getRowHeadersWidth(pivotStyle: Required<PivotStyle>) {
```
pro-odoo: private ?

pro-odoo (commented):
LGTM for the rest

packages/o-spreadsheet-engine/src/types/pivot.ts
hokolomopo: Other pivot cells have only core info such as domain or measure id, I figured it'd be strange to have a translated value for this one

src/helpers/pivot/pivot_menu_items.ts
hokolomopo: Typing doesn't work if I do that :x

src/components/side_panel/pivot/pivot_side_panel/pivot_design_panel/pivot_design_panel.xml
hokolomopo: I'd prefer leaving the file consistent, other checkboxes don't have `this.`

packages/o-spreadsheet-engine/src/helpers/pivot/table_spreadsheet_pivot.ts
```diff
@@ -212,38 +227,72 @@ export class SpreadsheetPivotTable {
...
       return { type: "VALUE", domain, measure };
     }
+
+    return EMPTY_PIVOT_CELL;
   }
 
-  private getColHeaderDomain(col: number, row: number) {
-    if (col === 0) {
-      return undefined;
+  private getTabularFormPivotCell(
```
rrahir: is this functionreally necesary ? the two sub cases could be easily inserted in `getPivotCell` and I feel like it's easier to follow the flow through a single function rather than having 2 niche case in one function and then fallback to another function

hokolomopo: I feel like adding even more `if` cases in `getPivotCell` is worse in terms of readability, but sure 🤷

rrahir:
robodoo r+

PR #7398, opened 2025-11-03
[FIX] NumberEditor: drop unused rule

rrahir:
The rule `height: calc(100%-4px)` was never actually used. Before we forced a switch to .css files, the rule was overriden by a flat `height:30px` and this was meant to be.

The shift to .css files activated that unused rule by accident only because it was always conflicting and the change of loading order of the css files.

Concretely, the related component `NumberEditor` is only used in two places:
- `Topbar`, in which all items should have a size of 30 px,
- `TextStyler`, where we add a class that forces the size to 30px as well.

So apparently don't use it at all.

Task: 0

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/number_editor/number_editor.css
+++ b/src/components/number_editor/number_editor.css
@@ -1,7 +1,7 @@
 .o-spreadsheet {
   .o-number-editor {
     width: max-content !important;
-    height: calc(100% - 4px);
+
     input.o-font-size {
       outline: none;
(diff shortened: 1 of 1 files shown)


pro-odoo:
robodoo r+

PR #7400, opened 2025-11-04
[FIX] composer: change mode to editing when composer cursor selection changes

rmbh-odoo:
## Description:

Current behavior before PR:
- When typing `=sum(` and pressing the Home key, selecting a range caused weird
  behavior in the composer.
- The same issue occurred when using Ctrl+A to select the entire text.

- Changing the cursor selection now correctly switches the mode to 'editing'.

Task: [5226615](https://www.odoo.com/odoo/2328/tasks/5226615)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/composer/composer/composer.ts
+++ b/src/components/composer/composer/composer.ts
@@ -485,10 +485,7 @@ export class Composer extends Component<ComposerProps, SpreadsheetChildEnv> {
     }
     const newSelection = this.contentHelper.getCurrentSelection();
 
-    this.env.model.dispatch("STOP_COMPOSER_RANGE_SELECTION");
     this.props.onComposerContentFocused();
-    if (this.props.focus === "inactive") {
-    }
     this.env.model.dispatch("CHANGE_COMPOSER_CURSOR_SELECTION", newSelection);
     this.processTokenAtCursor();
   }
--- a/src/plugins/ui_stateful/edition.ts
+++ b/src/plugins/ui_stateful/edition.ts
@@ -144,6 +144,7 @@ export class EditionPlugin extends UIPlugin {
       case "CHANGE_COMPOSER_CURSOR_SELECTION":
         this.selectionStart = cmd.start;
         this.selectionEnd = cmd.end;
+        this.mode = "editing";
         break;
       case "STOP_COMPOSER_RANGE_SELECTION":
         if (this.isSelectingForComposer()) {
--- a/tests/composer/autocomplete_dropdown_component.test.ts
+++ b/tests/composer/autocomplete_dropdown_component.test.ts
@@ -338,7 +338,6 @@ describe("Autocomplete parenthesis", () => {
   test("=S( + edit S with autocomplete does not add left parenthesis", async () => {
     await typeInComposer("=S(");
     // go behind the letter "S"
-    model.dispatch("STOP_COMPOSER_RANGE_SELECTION");
     model.dispatch("CHANGE_COMPOSER_CURSOR_SELECTION", { start: 2, end: 2 });
     await nextTick();
     // show autocomplete
--- a/tests/composer/edition_plugin.test.ts
+++ b/tests/composer/edition_plugin.test.ts
@@ -103,6 +103,14 @@ describe("edition", () => {
     expect(model.getters.getEditionMode()).toBe("inactive");
   });
 
+  test("should switch to editing mode when composer cursor selection changes", () => {
+    const model = new Model();
+    model.dispatch("START_EDITION", { text: "=sum(" });
+    expect(model.getters.getEditionMode()).toBe("selecting");
+    model.dispatch("CHANGE_COMPOSER_CURSOR_SELECTION", { start: 0, end: 5 });
+    expect(model.getters.getEditionMode()).toBe("editing");
+  });
+
   test("Stopping the edition should complete the missing parenthesis of a formula", async () => {
     const model = new Model();
     model.dispatch("START_EDITION", { text: "=sum(sum(1,2" });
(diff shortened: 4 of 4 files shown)


doc/tsdoc/README.md
```diff
@@ -205,13 +205,13 @@ ___
 
 ### invalidateEvaluationCommands
```
rrahir: should not be part of the diff. the doc has been outdated for a while and if we really wanted to update it, we should do that in a dedicated commit

src/plugins/ui_stateful/edition.ts
```diff
@@ -145,8 +145,10 @@ export class EditionPlugin extends UIPlugin {
         this.selectionStart = cmd.start;
         this.selectionEnd = cmd.end;
         break;
-      case "STOP_COMPOSER_RANGE_SELECTION":
-        if (this.isSelectingForComposer()) {
+      case "UPDATE_COMPOSER_EDITION_MODE":
+        if (this.canStartComposerRangeSelection()) {
+          this.startComposerRangeSelection();
```
rrahir: this is a change of behaviour as well and not a welcome one,  I fear:
- start writing a formula with "=A1+"
- the selection carret is visible 
- if I want to crrect what I already input, I click inside the composer, and the carret goes away, I can now use the direction arrow to navigater inside the composer and not the grid
- as soon as my cursor reaches after the "+" sign, the selection carret will reappear and now, useing the arrow will make me navigate in the grid, which is not what I intended for at all

I'm not a huge fan of a command rename either specifically in a stable and old version.

On my end, I thougth about capturing the Ctrl+A event and manually store the edition and setting up the selection by hand
```diff
index faac14bbd5..29ae21a588 100644
--- a/src/components/composer/composer/composer.ts
+++ b/src/components/composer/composer/composer.ts
@@ -12,7 +12,7 @@ import {
   SpreadsheetChildEnv,
 } from "../../../types/index";
 import { css, cssPropertiesToCss } from "../../helpers/css";
-import { getBoundingRectAsPOJO } from "../../helpers/dom_helpers";
+import { getBoundingRectAsPOJO, isCtrlKey } from "../../helpers/dom_helpers";
 import { getHtmlContentFromPattern } from "../../helpers/html_content_helpers";
 import { updateSelectionWithArrowKeys } from "../../helpers/selection_helpers";
 import { Popover, PopoverProps } from "../../popover/popover";
@@ -189,6 +189,15 @@ export class Composer extends Component<ComposerProps, SpreadsheetChildEnv> {
     F2: () => console.warn("Not implemented"),
     F4: (ev: KeyboardEvent) => this.processF4Key(ev),
     Tab: (ev: KeyboardEvent) => this.processTabKey(ev),
+    a: (ev: KeyboardEvent) => {
+      if (isCtrlKey(ev)) {
+        this.env.model.dispatch("STOP_COMPOSER_RANGE_SELECTION");
+        this.env.model.dispatch("CHANGE_COMPOSER_CURSOR_SELECTION", {
+          start: 0,
+          end: this.contentHelper.getCurrentSelection().end,
+        });
+      }
+    },
   };
 ```Is this something you considered? Do you think it would work?

Comment (rrahir) [reply src/plugins/ui_stateful/edition.ts]: we could also consider the fact that setting the composer cursor selection (CHANGE_COMPOSER_CURSOR_SELECTION) should trigger a recomputation of the mode inside the plugin

Comment (rrahir) [reply src/plugins/ui_stateful/edition.ts]: the diff could actually look something like
```diff
diff --git a/src/plugins/ui_stateful/edition.ts b/src/plugins/ui_stateful/edition.ts
index 4d240076c2..a731356839 100644
--- a/src/plugins/ui_stateful/edition.ts
+++ b/src/plugins/ui_stateful/edition.ts
@@ -144,6 +144,7 @@ export class EditionPlugin extends UIPlugin {
       case "CHANGE_COMPOSER_CURSOR_SELECTION":
         this.selectionStart = cmd.start;
         this.selectionEnd = cmd.end;
+        this.mode = "editing";
         break;
       case "STOP_COMPOSER_RANGE_SELECTION":
         if (this.isSelectingForComposer()) {
diff --git a/tests/composer/edition_plugin.test.ts b/tests/composer/edition_plugin.test.ts
index 65021acd0f..bd36ffc90f 100644
--- a/tests/composer/edition_plugin.test.ts
+++ b/tests/composer/edition_plugin.test.ts
@@ -422,10 +422,7 @@ describe("edition", () => {
     model.dispatch("START_EDITION");
     model.dispatch("SET_CURRENT_CONTENT", {
       content: "=",
-    });
-    model.dispatch("CHANGE_COMPOSER_CURSOR_SELECTION", {
-      start: 1,
-      end: 1,
+      selection: { start: 1, end: 1 },
     });
 
     setSelection(model, ["A1:A3"]);
```

rmbh-odoo: Got your points. :wink: 
I’ve updated it, and it looks good for the stable versions - hopefully no new bugs!

Also, from your POV, I got an idea for master: make `onClick` only block arrow key navigation (not make the composer editable), while still allowing selection with mouse down - like in GSheets and Excel 😄

I’ll discuss this with the PO and create an imp task in master if needed.

src/components/composer/composer/composer.ts
```diff
@@ -485,10 +485,7 @@ export class Composer extends Component<ComposerProps, SpreadsheetChildEnv> {
     }
     const newSelection = this.contentHelper.getCurrentSelection();
 
-    this.env.model.dispatch("STOP_COMPOSER_RANGE_SELECTION");
```
rrahir, line 488: is this related to the fix or rather the improvement you mentioned?

rmbh-odoo: Now that we have changed the mode directly in `CHANGE_COMPOSER_CURSOR_SELECTION`,
This `STOP_COMPOSER_RANGE_SELECTION` action has become redundant,
since the composer mode update is already handled afterward.
So it's not an improvement

VincentSchippefilt:
robodoo r+

PR #7401, opened 2025-11-04
[IMP] formulas: add spilled range operator

laa-odoo:
This commit introduces the `#` spilled range operator.

Instead of extending the range syntax, `#` is implemented as a regular postfix unary operator.
This allows it to be used on any expression that returns a reference, not only on static
cell references. For example, constructs like `=IF(condition, A1, H10)#` become possible,
where `#` applies to the result of the formula rather than to a single literal reference.

On the UI side, the composer still tokenizes `<ref>#` as a single token so that users can
select it as a whole, while the engine compiles `#` as a normal unary operator bound to
SPILLED_RANGE formula.

Task: [5365642](https://www.odoo.com/odoo/2328/tasks/5365642)

### DIFF
--- a/packages/o-spreadsheet-engine/src/formulas/composer_tokenizer.ts
+++ b/packages/o-spreadsheet-engine/src/formulas/composer_tokenizer.ts
@@ -230,7 +230,49 @@ function addArgsAST(tokens: EnrichedToken[]): EnrichedToken[] {
  * @param formula
  */
 export function composerTokenize(formula: string, locale: Locale): EnrichedToken[] {
-  const tokens = rangeTokenize(formula, locale);
-
+  // Note:
+  // Ideally, rangeTokenize would only identify ranges for the composer:
+  // Ranges should be treated as single tokens in the composer, so users can select
+  // ranges on the grid. And compilation should treats ranges as operations between
+  // references. ( because ":" can be used as an operator between other things than
+  // references).
+  //
+  // The same applies to spilled ranges: although they are handled as operators
+  // during compilation, for the composer, we already treat them as part of the
+  // same token as the reference.
+  //
+  // Therefore, in the future, spillRangeTokenize and rangeTokenize will need to
+  // be combined into a single process specific to the composer and to the token
+  // creation process specific to the composer selection mode.
+  const tokens = spilledRangeTokenize(rangeTokenize(formula, locale));
   return addArgsAST(mapParentFunction(mapParenthesisCode(enrichTokens(tokens))));
+}
+
+function spilledRangeTokenize(tokens: Token[]): Token[] {
+  const result: Token[] = [];
+  while (tokens.length) {
+    result.push(matchSpilledRange(tokens) || tokens.shift()!);
+  }
+  return result;
+}
+
+function matchSpilledRange(tokens: Token[]): Token | null {
+  if (tokens.length < 2) {
+    return null;
+  }
+  const firstToken = tokens[0];
+  const secondToken = tokens[1];
+  if (
+    firstToken.type === "REFERENCE" &&
+    secondToken.type === "OPERATOR" &&
+    secondToken.value === "#" &&
+    !firstToken.value.includes(":")
+  ) {
+    tokens.splice(0, 2);
+    return {
+      type: firstToken.type,
+      value: firstToken.value + "#",
+    };
+  }
+  return null;
 }
--- a/packages/o-spreadsheet-engine/src/functions/module_operators.ts
+++ b/packages/o-spreadsheet-engine/src/functions/module_operators.ts
@@ -1,9 +1,9 @@
 import { _t } from "../translation";
-import { DivisionByZeroError } from "../types/errors";
+import { DivisionByZeroError, EvaluationError, InvalidReferenceError } from "../types/errors";
 import { AddFunctionDescription } from "../types/functions";
-import { FunctionResultNumber, FunctionResultObject, Maybe } from "../types/misc";
+import { FunctionResultNumber, FunctionResultObject, Matrix, Maybe } from "../types/misc";
 import { arg } from "./arguments";
-import { isEvaluationError, toNumber, toString } from "./helpers";
+import { generateMatrix, isEvaluationError, toNumber, toString } from "./helpers";
 import { POWER } from "./module_math";
 
 // -----------------------------------------------------------------------------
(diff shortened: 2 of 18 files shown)


packages/o-spreadsheet-engine/src/functions/module_operators.ts
```diff
@@ -288,6 +288,77 @@ export const POW = {
...
 } satisfies AddFunctionDescription;
 
+// -----------------------------------------------------------------------------
+// SPILLED_RANGE
+// -----------------------------------------------------------------------------
+
+/**
+ * Internal formula for constructing reference ranges dynamically.
+ * Technically, this formula was created to be associated with the "#" operator.
+ * We chose "#" as an operator because it enables the construction of references
```
LucasLefevre: we didn't really "chose" `#`, it's just what competitors do

packages/o-spreadsheet-engine/src/functions/module_operators.ts
```diff
@@ -288,6 +288,77 @@ export const POW = {
...
+ * in a "dynamic" sense, rather than being tied to a static reference.
+ * Thus, formulas like:
+ *  =IF(condition, A1, H10)# can be imagined, instead of a simple static reference like A1#.
+ **/
```
LucasLefevre: Smaller and good enough ?
```ts
/**
 * Internal formula implementing the # operator.
 * It allows to get the spilled range of an array formula from any reference, including
 * references returned by other functions.
 * e.g. =IF(condition, A1, H10)#
 **/
```

packages/o-spreadsheet-engine/src/functions/module_operators.ts
```diff
@@ -288,6 +288,77 @@ export const POW = {
...
+export const SPILLED_RANGE = {
+  description: _t("Gets the spilled range of an array formula."),
+  args: [arg("ref (any, range<any>)", _t("The reference to get the spilled range from."))],
```
LucasLefevre: for clarity when reading this (without reading the compiler)
```suggestion
  args: [arg("ref (any, range<any>, meta)", _t("The reference to get the spilled range from."))],
```
even though the days of `meta` are counted

packages/o-spreadsheet-engine/src/functions/module_operators.ts
```diff
@@ -288,6 +288,77 @@ export const POW = {
...
+    }
+
+    const spilledZone = this.getters.getSpreadZone(cellPosition);
+    if (spilledZone === undefined) {
+      return new InvalidReferenceError();
+    }
+    const spilledRange = this.getters.getRangeFromZone(initialRange.sheetId, spilledZone);
+    if (spilledRange.invalidXc || spilledRange.invalidSheetName) {
+      return new InvalidReferenceError();
+    }
```
LucasLefevre: I don't think this check is needed. Ranges are always valid when built from a sheet and a zone

packages/o-spreadsheet-engine/src/functions/module_operators.ts
```diff
@@ -288,6 +288,77 @@ export const POW = {
...
+    if (originPosition) {
+      this.addDependencies?.(originPosition, [spilledRange]);
+    }
+
+    // Iterate over all positions in the spilledZone and get the evaluated cell for each
```
LucasLefevre: useless comment. The code is readable and tells the same story

packages/o-spreadsheet-engine/src/plugins/ui_stateful/selection.ts
```diff
@@ -442,6 +446,47 @@ export class GridSelectionPlugin extends UIPlugin {
...
+          ...expandedZone,
+          bottom: isFullColRange(range) ? undefined : expandedZone.bottom,
+          right: isFullRowRange(range) ? undefined : expandedZone.right,
+        },
+      },
+      this.getters.getSheetSize
+    );
+    const rangeString = this.getters.getRangeString(expandedRange, forSheetId);
+    if (this.getters.isSingleCellOrMerge(range.sheetId, range.zone)) {
+      // reconstruct in this way to handle possible ":" in the sheet name
```
LucasLefevre: I'm not sure this comment adds any value.
...or we should add it everywhere `splitReference` is used for consistency.
Where it would be useful is where we forget to use `splitReference`, but then, well, ...

packages/o-spreadsheet-engine/src/plugins/ui_stateful/selection.ts
```diff
@@ -442,6 +446,47 @@ export class GridSelectionPlugin extends UIPlugin {
...
+    const spreaderPosition = this.getters.getArrayFormulaSpreadingOn({
+      sheetId: expandedRange.sheetId,
+      col: expandedRange.zone.left,
+      row: expandedRange.zone.top,
+    });
+
+    const spreadZone =
+      spreaderPosition && this.getters.getSpreadZone(spreaderPosition, { ignoreSpillError: true });
+    if (spreadZone && isEqual(spreadZone, expandedRange.zone)) {
+      // reconstruct in this way to handle possible ":" in the sheet name
```
LucasLefevre: same here about the comment

tests/evaluation/compiler.test.ts
```diff
@@ -87,6 +87,11 @@ describe("expression compiler", () => {
     const compiledFormula = compiledBaseFunction("={1;2;3;4}");
     expect(compiledFormula.execute.toString()).toMatchSnapshot();
   });
+
+  test("expression with ref#", () => {
+    const compiledFormula = compiledBaseFunction("=A1:B2#");
```
LucasLefevre: also add a test with a single cell
```suggestion
  test("expression with cell#", () => {
    const compiledFormula = compiledBaseFunction("=A1#");
```

tests/evaluation/evaluation.test.ts
```diff
@@ -384,6 +384,31 @@ describe("evaluateCells", () => {
     expect(evaluateCell("A1", { A1: "=A2<>A3", A2: "abc", A3: "def" })).toBe(true);
   });
 
+  test("# operator", () => {
```
LucasLefevre, line 387: add a test where the target in on another sheet

tests/evaluation/tokenizer.test.ts
```diff
@@ -140,6 +140,14 @@ describe("tokenizer", () => {
     ]);
   });
 
+  test("spill range operator", () => {
```
LucasLefevre, line 143: you can also add tests for `composerTokenize` in `composer_tokenizer.test.ts`

packages/o-spreadsheet-engine/src/formulas/composer_tokenizer.ts
```diff
@@ -230,7 +230,49 @@ function addArgsAST(tokens: EnrichedToken[]): EnrichedToken[] {
...
+}
+
+function matchSpilledRange(tokens: Token[]): Token | null {
+  if (tokens.length < 2) {
+    return null;
+  }
+  const firstToken = tokens[0];
+  const secondToken = tokens[1];
+  if (
+    (firstToken.type === "REFERENCE" || firstToken.type === "INVALID_REFERENCE") &&
```
LucasLefevre: c'est vrai ce `INVALID_REFERENCE` ? not test is failing when removed
```suggestion
    firstToken.type === "REFERENCE" &&
```

LucasLefevre (commented):
good good 👌 
only details and let's merge :)

tests/evaluation/tokenizer.test.ts
laa-odoo: indeed :)

laa-odoo:
@LucasLefevre, thx for the review
comments taking into account :)

PR #7402, opened 2025-11-04
[PERF] Renderer: speed-up box generation

rrahir:
when invalidating a lot of cells at once (typically happens during a MOVE_COLUMN_ROW command with a lot of headers), we compute the zones in which we can prevent animations during the next rendering. Each updated cell will need to be compared to the boxes that we want to rerender. The more the updated cells, the more comparisons will take place.

This revision adds a performance step to recompute the animation-invalidated zones in order to limit the overall comparisons.

Performance benchmark steps:
- Create a sheet with 5000 rows
- In the first column, each cell has a reference to its neighbour (=B1,=B2, etc)
- Move the first column to column C

before - 1300 ms
after - 540 ms

Task: 5213090

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/zones.ts
+++ b/src/helpers/zones.ts
@@ -405,7 +405,12 @@ export function isInside(col: number, row: number, zone: Zone): boolean {
  * Check if a zone is inside another
  */
 export function isZoneInside(smallZone: Zone, biggerZone: Zone): boolean {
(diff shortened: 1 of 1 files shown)


LucasLefevre (commented):
robodoo rebase-ff r+

PR #6174, opened 2025-04-24
[IMP] Spreadsheet: Mobile-tte (pouet pouet 🛵)

rrahir:
This pull request (finally) introduces a mobile edition mode for Spreadsheet!

The main points tackled were:

### Grid manipulations
The main issues were:
- the impossibility to properly navigate in the spreadsheet without messing
   the selection,
- a nightmare to select ranges for a formula
- the structure did not tolerate narrow viewports and it was a pain to find
  the information you were looking for
- Drag-and-drop elements (figures for instance) would conflict with the
  ability to scroll the viewport

The first three points were addressed by defining a new flow for specific
mobile/touchscreen-only devices.

All drag-and-drop features were deactivated at this time and will come back
in the future once we develop some tools to detect longtouches.

### Air out the top bar:
The top bar is too cluttered on a narrow screen and the edition of cells
from a top bar composer is counter intuitive as the virtual keyboard is
placed at the bottom of a screen device. We introduce a dedicated bottom
bar for this specific situation that will allow to:
- edit cell contents
- navigate through the menu bar (which is extracted from the top bar)
- navigate through the different sheets (standard bottom bar)

This arrangement leaves the grid at the center of the screen, where the
user attention is focused.

## Current limitations
Some features were disabled because their current behaviour/implementation
is quite complicated to adapt to the mobile experience.
These are:
- drag and dropping elements in the grid/bottom bar
- resizing headers
- every feature that requires the use of a side panel (pivot/figure edition,
  find & replace, data validation)

They will come in time but we first set our focus on two main uses:
- grid consultation
- light cell edition

Task: [2756003](https://www.odoo.com/odoo/2328/tasks/2756003)

PR #7404, opened 2025-11-05
[FIX] dashboard: properly handle grid resize in dashboard mode

rmbh-odoo:
## Description:

Current behavior before PR:
- The canvas size was computed from the grid, which caused visible white margins
  after a snappy resize (commit [3c8df4b9](https://github.com/odoo/o-spreadsheet/pull/6174/commits/3c8df4b9fdd118a34eceed19d9fc96553a62b861)).
- Some unused props were being passed, creating unnecessary code clutter.

- The dashboard canvas size is now computed from the gridRef, ensuring only the
  visible parts are displayed and avoiding redundant canvas computations.

Task: [5223156](https://www.odoo.com/odoo/2328/tasks/5223156)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/dashboard/dashboard.ts
+++ b/src/components/dashboard/dashboard.ts
@@ -21,7 +21,9 @@ import { Popover } from "../popover/popover";
 import { HorizontalScrollBar, VerticalScrollBar } from "../scrollbar/";
 import { ClickableCell, ClickableCellsStore } from "./clickable_cell_store";
 
-interface Props {}
+interface Props {
+  getGridSize: () => DOMDimension;
+}
 
 css/* scss */ `
   .o-dashboard-clickable-cell {
@@ -79,10 +81,8 @@ export class SpreadsheetDashboard extends Component<Props, SpreadsheetChildEnv> 
   }
 
   get gridContainer() {
-    const sheetId = this.env.model.getters.getActiveSheetId();
-    const { right } = this.env.model.getters.getSheetZone(sheetId);
-    const { end } = this.env.model.getters.getColDimensions(sheetId, right);
-    return cssPropertiesToCss({ "max-width": `${end}px` });
+    const maxWidth = this.getMaxSheetWidth();
+    return cssPropertiesToCss({ "max-width": `${maxWidth}px` });
   }
 
   get gridOverlayDimensions() {
@@ -120,10 +120,12 @@ export class SpreadsheetDashboard extends Component<Props, SpreadsheetChildEnv> 
     this.cellPopovers.close();
   }
 
-  onGridResized({ height, width }: DOMDimension) {
+  onGridResized() {
+    const { height, width } = this.props.getGridSize();
+    const maxWidth = this.getMaxSheetWidth();
     this.env.model.dispatch("RESIZE_SHEETVIEW", {
-      width: width,
-      height: height,
+      width: Math.min(maxWidth, width),
+      height,
       gridOffsetX: 0,
       gridOffsetY: 0,
     });
@@ -142,5 +144,11 @@ export class SpreadsheetDashboard extends Component<Props, SpreadsheetChildEnv> 
       ...getRefBoundingRect(this.gridRef),
       ...this.env.model.getters.getSheetViewDimensionWithHeaders(),
     };
+  }
+
+  private getMaxSheetWidth(): Pixel {
+    const sheetId = this.env.model.getters.getActiveSheetId();
+    const { right } = this.env.model.getters.getSheetZone(sheetId);
+    return this.env.model.getters.getColDimensions(sheetId, right).end;
   }
 }
--- a/src/components/spreadsheet/spreadsheet.ts
+++ b/src/components/spreadsheet/spreadsheet.ts
@@ -544,26 +544,25 @@ export class Spreadsheet extends Component<SpreadsheetProps, SpreadsheetChildEnv
   }
 
   getGridSize() {
-    const topBarHeight =
-      this.spreadsheetRef.el
-        ?.querySelector(".o-spreadsheet-topbar-wrapper")
-        ?.getBoundingClientRect().height || 0;
-    const bottomBarHeight =
-      this.spreadsheetRef.el
-        ?.querySelector(".o-spreadsheet-bottombar-wrapper")
-        ?.getBoundingClientRect().height || 0;
+    const el = this.spreadsheetRef.el;
+    if (!el) {
+      return { width: 0, height: 0 };
+    }
+    const getHeight = (selector) => el.querySelector(selector)?.getBoundingClientRect().height || 0;
+    const getWidth = (selector) => el.querySelector(selector)?.getBoundingClientRect().width || 0;
 
-    const gridWidth =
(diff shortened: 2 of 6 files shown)


rrahir:
if I understand the issue correctly, we miss the actual computation of the spreadsheet width when in dashboard mode. have you looked into a way to correct that as well? because right now we have a discrepency between the canvas size and the actual part displayed to the client and I'm not sure that just hiding it is the best solution (maybe it's too complicated to fix and it's the right call, but i'm curious to know if looked into it)

src/components/spreadsheet/spreadsheet.ts
```diff
@@ -547,26 +547,25 @@ export class Spreadsheet extends Component<SpreadsheetProps, SpreadsheetChildEnv
...
-      (this.spreadsheetRef.el?.getBoundingClientRect().height || 0) -
-      (this.spreadsheetRef.el?.querySelector(".o-column-groups")?.getBoundingClientRect().height ||
-        0) -
-      topBarHeight -
-      bottomBarHeight;
-    return {
-      width: Math.max(gridWidth - SCROLLBAR_WIDTH, 0),
-      height: Math.max(gridHeight - SCROLLBAR_WIDTH, 0),
-    };
+    const el = this.spreadsheetRef.el;
```
rmbh-odoo, line 547: Refactored this function for two main reasons:
1. Reduced multiple direct DOM queries to a single root element reference to improve performance and avoid redundant layout calculations.
2. Cleaned up the logic with helper functions for better readability

src/components/dashboard/dashboard.ts
```diff
@@ -111,10 +102,14 @@ export class SpreadsheetDashboard extends Component<Props, SpreadsheetChildEnv>
     this.cellPopovers.close();
   }
 
-  onGridResized({ height, width }: DOMDimension) {
+  onGridResized() {
+    const rect = this.gridRef.el?.getBoundingClientRect();
```
rrahir: I fear this could break in later versions because of the introduction of the zoom (which still has to be implemented properly for dashboard but still) due to some weird stuff happening in safari which pretty much forces us to rely on the <Spreadsheet> width and height because it is not affected similarly by the zoom depending on different browsers, see https://www.odoo.com/odoo/2328/tasks/5232092

tests/grid/dashboard_grid_component.test.ts
```diff
@@ -70,9 +72,9 @@ describe("Grid component in dashboard mode", () => {
     model.updateMode("dashboard");
     await nextTick();
 
-    const y = DEFAULT_CELL_HEIGHT + 1 + MIN_CELL_TEXT_MARGIN; // +1 to skip grid lines
-    const leftB = DEFAULT_CELL_WIDTH * 2 - GRID_ICON_EDGE_LENGTH - GRID_ICON_MARGIN;
-    const leftC = DEFAULT_CELL_WIDTH * 3 - GRID_ICON_EDGE_LENGTH - GRID_ICON_MARGIN;
+    const y = HEADER_HEIGHT + DEFAULT_CELL_HEIGHT + 1 + MIN_CELL_TEXT_MARGIN; // +1 to skip grid lines
```
rrahir: the headers are not present in dashboard, it should clearly not apprea in this kind of computation

src/components/dashboard/dashboard.ts
rrahir:
```suggestion
     const { height }  = this.props.getGridSize();
     const sheetId = this.env.model.getters.getActiveSheetId();
     const { right } = this.env.model.getters.getSheetZone(sheetId);
     const width = this.env.model.getters.getColDimensions(sheetId, right).end;
```

this takes advantage of the future zoom issue and the computed width (something we already do in `SpreadsheetDashboard.gridContainer`) Wdyt?

src/components/spreadsheet/spreadsheet.ts
rrahir: Not sure about the actual performance gain but it's definitely less messy :)

src/components/dashboard/dashboard.ts
rmbh-odoo: That solution doesn’t work because it sets a fixed max-width on the canvas on every resize, which doesn’t make sense and causes the layout bug. See the related issue [here](https://drive.google.com/file/d/1-eNLVzdUGK5MO1h3c95oKQ30B2J4D4e8/view)

I think we should fix this issue here, since this version doesn’t include Zoom functionality.
In the master, we’ll likely need to refactor the layout to align with the zoom behavior.
I’ll take some time to investigate what’s happening in the master :smiling_face_with_tear:

tests/grid/dashboard_grid_component.test.ts
rmbh-odoo: Headers aren’t present in the dashboard, so they shouldn’t affect this computation.
Not sure why the test still includes header size, since it’s not actually rendered. :confounded: 

Addition:
In dashboard tests, `gridRef.el.getBoundingClientRect()` returns 0 width and height, preventing `RESIZE_SHEETVIEW` dispatch and leaving the canvas size unupdated.

src/components/dashboard/dashboard.ts
rrahir: Here is the diff I applied on my end, I did not experience the same issue as you do and when pushed it in Odoo ,I get the right positioning (and style) at first render
image

I rushed my explanation in my previous post but I think that it's a correct approach?


```diff
diff --git a/src/components/dashboard/dashboard.ts b/src/components/dashboard/dashboard.ts
index a54b218d57..96bf59dd45 100644
--- a/src/components/dashboard/dashboard.ts
+++ b/src/components/dashboard/dashboard.ts
@@ -103,8 +103,11 @@ export class SpreadsheetDashboard extends Component<Props, SpreadsheetChildEnv>
   }
 
   onGridResized({ height, width }: DOMDimension) {
+    const sheetId = this.env.model.getters.getActiveSheetId();
+    const { right } = this.env.model.getters.getSheetZone(sheetId);
+    const width2 = this.env.model.getters.getColDimensions(sheetId, right).end;
     this.env.model.dispatch("RESIZE_SHEETVIEW", {
-      width: width,
+      width: Math.min(width, width2),
       height: height,
       gridOffsetX: 0,
       gridOffsetY: 0,
```

rrahir: the diff was aplied in master but it's the same idea

rmbh-odoo: Honestly, that’s a brilliant move 🫡

src/components/dashboard/dashboard.ts
```diff
@@ -111,9 +113,13 @@ export class SpreadsheetDashboard extends Component<Props, SpreadsheetChildEnv>
     this.cellPopovers.close();
   }
 
-  onGridResized({ height, width }: DOMDimension) {
+  onGridResized() {
+    const { height, width } = this.props.getGridSize();
```
rrahir, line 124: I guess we could go a bit further and make a helper because those lines are a repetition of the computation in `get gridContainer()` ;-)

rrahir:
robodoo r+

PR #7415, opened 2025-11-06
[FIX] TopbarMenu: specify `isReadonlyAllowed`

rrahir:
The logic that displays the top-level items of the topbarmenu is isolated from the `Menu` component logic. More specifically, it never checks if the item can be used in readonly and as such, we never took the time to define the key `isReadonlyAllowed` on those menu items.

However `RibbonMenu` relies on the logic of `Menu` so in order to display the topbar menu items properly in mobile mode, we need the key `isReadonlyAllowed` to be defined.

Task: [5245432](https://www.odoo.com/odoo/2328/tasks/5245432)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/registries/menus/topbar_menu_registry.ts
+++ b/src/registries/menus/topbar_menu_registry.ts
@@ -21,6 +21,7 @@ topbarMenuRegistry
   .add("file", {
     name: _t("File"),
     sequence: 10,
+    isReadonlyAllowed: true,
   })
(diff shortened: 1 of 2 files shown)


rrahir:
@nseinlet if you want to follow the fix ^^

PR #7427, opened 2025-11-07
Master pivot cache adrm

hokolomopo:
## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/packages/o-spreadsheet-engine/src/functions/helpers.ts
+++ b/packages/o-spreadsheet-engine/src/functions/helpers.ts
@@ -11,7 +11,7 @@ import {
   NotAvailableError,
   errorTypes,
 } from "../types/errors";
-import { LookupCaches } from "../types/functions";
+import { EvalContext, LookupCaches } from "../types/functions";
 import { Locale } from "../types/locale";
 import {
   Arg,
@@ -499,6 +499,7 @@ type VectorArgType = "horizontal" | "vertical" | "matrix";
  *   as ranges and invoke this helper directly within your `compute` implementation.
  */
 export function applyVectorization(
+  evalCtx: EvalContext,
   formula: (...args: Arg[]) => Matrix<FunctionResultObject> | FunctionResultObject,
   args: Arg[],
   acceptToVectorize: boolean[] | undefined = undefined
@@ -541,7 +542,7 @@ export function applyVectorization(
 
   if (countVectorizedCol === 1 && countVectorizedRow === 1) {
     // either this function is not vectorized or it ends up with a 1x1 dimension
-    return formula(...args);
+    return formula.call(evalCtx, ...args);
   }
 
   const getArgOffset: (i: number, j: number) => Arg[] = (i, j) =>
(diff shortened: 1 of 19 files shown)


hokolomopo:
We can also probably remove the PivotPresencePlugin and the getter/helper getFirstPivotFunction - didn't try yet

packages/o-spreadsheet-engine/src/functions/helpers.ts
```diff
@@ -541,7 +542,7 @@ export function applyVectorization(
 
   if (countVectorizedCol === 1 && countVectorizedRow === 1) {
     // either this function is not vectorized or it ends up with a 1x1 dimension
-    return formula(...args);
+    return formula.call(evalCtx, ...args);
```
LucasLefevre:
```suggestion
    return formula.apply(evalCtx, args);
```
to avoid spreading the `args`

packages/o-spreadsheet-engine/src/functions/helpers.ts
```diff
@@ -564,7 +565,16 @@ export function applyVectorization(
...
+    const basePosition = evalCtx.__originCellPosition;
+    const ctx = { ...evalCtx };
+    if (basePosition) {
+      ctx.__originCellPosition = {
+        col: basePosition.col + col,
+        row: basePosition.row + row,
+        sheetId: basePosition.sheetId,
+      };
+    }
+    const singleCellComputeResult = formula.call(ctx, ...getArgOffset(col, row));
```
LucasLefevre:
```suggestion
    const singleCellComputeResult = formula.apply(ctx, getArgOffset(col, row));
```

packages/o-spreadsheet-engine/src/plugins/ui_core_views/cell_evaluation/evaluation_plugin.ts
```diff
@@ -159,6 +161,8 @@ export class EvaluationPlugin extends CoreViewPlugin {
     "getSpreadZone",
     "getArrayFormulaSpreadingOn",
     "isEmpty",
+    "getPivotCache",
```
LucasLefevre: `getPivotCache` doesn't mean anything from the business POV.
If the data is cached or not is an implementation detail and the name doesn't tell what's inside this "cache"

packages/o-spreadsheet-engine/src/plugins/ui_core_views/pivot_ui.ts
```diff
@@ -42,15 +42,21 @@ export class PivotUIPlugin extends CoreViewPlugin {
...
   private custom: UIPluginConfig["custom"];
 
+  private pivotCellsCache = new PositionMap<PivotCacheItem>();
+
   constructor(config: CoreViewPluginConfig) {
     super(config);
     this.custom = config.custom;
+    evaluationListenerRegistry.replace("PivotUIPlugin", {
+      handleEvaluationMessage: this.handleEvaluationMessage.bind(this),
+    });
```
hokolomopo, line 59: using a registry is maybe foireux, it will explode if we have multiple models (does it happen in practice ? maybe with dashboard/version history ?)

PR #7444, opened 2025-11-11
[IMP] Composer: select text in the formula assistant

rrahir:
This commit introduces the possibility for the user to select the text content of the formula assistant.

Task: 5169514

## Description:

Task: [5169514](https://www.odoo.com/odoo/2328/tasks/5169514)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/composer/composer/composer.xml
+++ b/src/components/composer/composer/composer.xml
@@ -1,6 +1,6 @@
 <templates>
   <t t-name="o-spreadsheet-Composer">
-    <div class="o-composer-container w-100 h-100">
+    <div class="o-composer-container w-100 h-100" t-ref="composerContainer">
       <t t-set="autoCompleteProposals" t-value="props.composerStore.autoCompleteProposals"/>
       <t t-set="canBeToggled" t-value="props.composerStore.canBeToggled"/>
       <t
@@ -12,10 +12,9 @@
           t-if="props.focus !== 'inactive' and assistantIsAvailable and canBeToggled and assistant.forcedClosed"
           role="button"
           title="Show formula help"
-          t-on-click="openAssistant"
+          t-on-click.stop="openAssistant"
           t-on-pointerdown.prevent.stop=""
-          t-on-click.prevent.stop=""
-          t-on-pointerup.prevent.stop=""
+          t-on-pointerup.stop=""
           class="fa-stack position-absolute translate-middle force-open-assistant fs-4">
           <i class="fa fa-circle fa-stack-1x fa-inverse"/>
           <i class="fa fa-question-circle fa-stack-1x"/>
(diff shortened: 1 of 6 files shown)


src/components/composer/formula_assistant/formula_assistant.xml
```diff
@@ -1,6 +1,12 @@
 <templates>
   <t t-name="o-spreadsheet-FunctionDescriptionProvider">
-    <div class="o-formula-assistant-container user-select-none shadow">
+    <div
+      class="o-formula-assistant-container shadow"
+      tabindex="0"
```
hokolomopo: if the focus is in the formula assistant and the we press arrows/a random key, the assistant gets a black outline. I'm guessing this isn't intended ?

hokolomopo: Also pressing tab when the assistant is focused:
- on the grid composer puts the focus on the grid scrollbar(?), then back on the composer
- on the topbar composer it puts the focus on the "add rows footer", breaking the display
image

src/components/composer/composer/composer.xml
```diff
@@ -51,13 +50,14 @@
         t-att-style="assistantContainerStyle"
         t-if="props.focus !== 'inactive' and assistantIsAvailable and !(canBeToggled and assistant.forcedClosed)"
         t-on-wheel.stop=""
-        t-on-pointerdown.prevent.stop=""
-        t-on-pointerup.prevent.stop=""
-        t-on-click.prevent.stop="">
+        t-on-pointerdown.stop=""
```
hokolomopo, line 53: In the **topbar** composer, writing =S and clicking on a formula proposition closes the composer

src/components/composer/formula_assistant/formula_assistant.xml
```diff
@@ -1,6 +1,12 @@
...
+      t-on-pointerdown=""
```
hokolomopo, line 6: hovering a cell in the grid puts the focus back on the composer, even if some text of the formula assistant was selected. I'm guessing anything that triggers a render does that ? 

we probably don't care ?

src/components/composer/formula_assistant/formula_assistant.xml
rrahir: I guess blocking every keyboard input would do the trick :p I guess we would only miss the "Ctrl.A" feature ( and I really don't see the point of doing this tbh)

rrahir: honestly, all this is related to the onblur event taht we want to block no matter what. If we disable it on the grid /topbar composer (which I think we should do anyway), it fixes the issue. Did the onblur was absolutely required outside of the standalone composer? I don't recall that well

hokolomopo: I think the onBlur give the composers a better behavior (no random focus regain after clicking in the topbar), but it's not 100% mandatory. But in 6 month we'll get a new task "cannot select text in the formula assistant in side panel composer" so I'm not sure it fixes the issue :stuck_out_tongue:

rrahir: the issue I see with the onBlur is that is some real world cases, like switching tabs to copy/paste a content, you automatically lose the focus, which closes the composer. I stumbled upon it a few times and it wasa bit frustrating. And I obviously won't mention the debug behaviour.  I will work around it for this task ofc

rrahir: > on the topbar composer it puts the focus on the "add rows footer", breaking the

that is also true for other inputs (zoom %, fontsize in the toolbar, I will create another task to address it)

src/components/composer/formula_assistant/formula_assistant.xml
rrahir: I got a shitty fix for that specific situation, i'll let you judge

rrahir: needs some tests though

src/components/composer/formula_assistant/formula_assistant.xml
hokolomopo: I have no strong feeling on onBlur in the grid/topbar composer either way, if you can convince francois i'm fine with removing it :shrug:

src/components/composer/formula_assistant/formula_assistant.xml
hokolomopo: I'm leaning towards "not worth the ugly fix"

pro-odoo:
robodoo r+

PR #7455, opened 2025-11-12
[IMP] pivots: implement pivot table styles

hokolomopo:
### [IMP] pivots: implement pivot table styles

With this commit, we don't need to manually create a dynamic table
to a pivot to have a style applied. Instead we can add a style in the
pivot definition, and dynamic tables will automatically be created on
the dynamic pivot formulas.

Those new pivot styles are better than traditional tables styles because:
- they are directly linked to the pivot, taking into account the
number of headers, the presence of totals, etc.
- they are automatically added on `=PIVOT()` formulas, without the need to
create a dynamic table first.
- they are more powerful than the old table styles, they can have a
style for the sub-headers, the measure headers, etc.

### [IMP] renderer: draw cell background over grid lines

With this commit, the cell background are now drawn totally covering
the grid lines. It's way prettier for table/pivots to have an uniform
background color without grid lines breaking it.

The downside is we cannot draw the hover overlay with a color alpha,
because it would overlap with the next cell, making a darker color
at the border. We can workaround by merging the colors programmatically
and avoid drawing with alpha.

### [IMP] style: add `skipCellGridLines` style option

This commit introducing cell animations added an option
`skipCellGridLines` to the rendering boxes. The option allows to
skip drawing grid lines for specific cells.

This commit changes this rendering box option into a standard cell
style option, so it can be used in other parts of the codebase as well.

### [REF] table style: add `bold` to table style presets

Before this commit, the table style presets contained all the style of
a table, except the `bold` property, which was handled separately and
applied to every total/header row or first/last column.

This made the code harder to read with two sources of truth for the
style, and with the introduction of pivot table styles, every header
row won't be in bold anymore.

This commit adds the `bold` property to the table style presets,
os it is handled like every other style property.

Task: [4552232](https://www.odoo.com/odoo/2328/tasks/4552232)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


packages/o-spreadsheet-engine/src/types/misc.ts
```diff
@@ -119,6 +119,7 @@ export interface Style {
   fillColor?: Color;
   textColor?: Color;
   fontSize?: number; // in pt, not in px!
+  skipCellGridLines?: boolean;
```
LucasLefevre: I'm really not a fan of adding this to `Style` which is supposed to be the core, raw data for style, while `skipCellGridLines` seems to be something derived from other things

EDIT: mouais...I see

packages/o-spreadsheet-engine/src/helpers/pivot_table_presets.ts
```diff
@@ -0,0 +1,469 @@
...
+  headerRow: {
+    style: { bold: true, textColor: "#000000" },
+    border: {
+      top: { color: colorSet.highlight, style: "medium" },
+      bottom: { color: colorSet.highlight, style: "medium" },
+      left: { color: colorSet.highlight, style: "medium" },
+      right: { color: colorSet.highlight, style: "medium" },
+    },
+  },
+  measureHeaderRow: { border: { top: { color: colorSet.light, style: "thin" } } },
```
LucasLefevre: all `measureHeaderRow` seem to be unused

EDIT: or `measureHeaderRow` is supposed to be `measureHeader` ? (or the other way around)

packages/o-spreadsheet-engine/src/types/table.ts
```diff
@@ -79,24 +94,19 @@ export interface TableStyle {
   headerRow?: TableElementStyle;
   totalRow?: TableElementStyle;
 
+  measureHeader?: TableElementStyle;
```
LucasLefevre, line 97: `measureHeader` is also never useful ?

packages/o-spreadsheet-engine/src/helpers/table_helpers.ts
```diff
@@ -202,21 +202,12 @@ function getAllTableStyles(
...
-  return (
-    tableElement === "firstColumn" ||
-    tableElement === "lastColumn" ||
-    tableElement === "headerRow" ||
-    tableElement === "totalRow"
-  );
-}
-
-function getTableElementZones(
+export function getTableElementZones(
```
LucasLefevre:
```suggestion
function getTableElementZones(
```

packages/o-spreadsheet-engine/src/plugins/ui_core_views/pivot_ui.ts
```diff
@@ -346,4 +348,54 @@ export class PivotUIPlugin extends CoreViewPlugin {
...
+    position: CellPosition;
+    pivotStyle: Required<PivotStyle>;
+    pivotId: UID;
+  }[] {
+    const result: { position: CellPosition; pivotStyle: Required<PivotStyle>; pivotId: UID }[] = [];
+    for (const cellId of this.getters.getCellsWithTrackedFormula("PIVOT")) {
+      const position = this.getters.getCellPosition(cellId);
+      if (!position) {
+        continue;
+      }
```
LucasLefevre: position is always truthy ?

packages/o-spreadsheet-engine/src/plugins/ui_core_views/pivot_ui.ts
```diff
@@ -346,4 +348,54 @@ export class PivotUIPlugin extends CoreViewPlugin {
...
+
+  getPivotStyleAtPosition(
+    position: CellPosition
+  ): { pivotStyle: Required<PivotStyle>; pivotId: UID } | undefined {
+    const cell = this.getters.getCell(position);
+    if (!cell || !cell.isFormula) {
+      return undefined;
+    }
+    const pivotFunction = this.getFirstPivotFunction(position.sheetId, cell.compiledFormula.tokens);
+    if (!pivotFunction || pivotFunction.functionName !== "PIVOT") {
```
LucasLefevre, line 377: Shouldn't it also check that PIVOT is the main formula ?
=PIVOT(), not =CHOOSEROWS(PIVOT())

packages/o-spreadsheet-engine/src/types/table.ts
```diff
@@ -44,6 +45,16 @@ export interface TableConfig {
   styleId: string;
 }
 
+export interface TableMetaData {
+  mode: "pivot" | "table";
+  numberOfCols: number;
+  numberOfRows: number;
+  measureHeaderIndex?: number;
```
LucasLefevre:
```suggestion
  measureRow?: number;
```
?

packages/o-spreadsheet-engine/src/helpers/pivot_table_presets.ts
```diff
@@ -0,0 +1,469 @@
...
+    },
+  },
+});
+
+const pivotMediumBlackHeaders: TableStyleTemplate = (colorSet) => ({
+  category: "medium",
+  templateName: "pivotMediumBlackHeaders",
+  primaryColor: colorSet.highlight,
+  wholeTable: {
+    style: { skipCellGridLines: true, fillColor: colorSet.light },
```
LucasLefevre: `skipCellGridLines: true` not useful if there's a fill color ?

tests/test_helpers/helpers.ts
```diff
@@ -414,12 +418,12 @@ export function getGridFormat(model: Model): GridFormatDescr {
...
 }
 
-export function getGridStyle(model: Model): GridStyleDescr {
+export function getGridStyle(model: Model, range?: string): GridStyleDescr {
   const result: GridStyleDescr = {};
   const sheetId = model.getters.getActiveSheetId();
-  for (const cellId of Object.keys(model.getters.getCells(sheetId))) {
-    const { col, row } = model.getters.getCellPosition(cellId);
-    result[toXC(col, row)] = model.getters.getCellStyle({ sheetId, col, row });
+  for (const [xc] of Object.entries(getCellGrid(model, range))) {
```
LucasLefevre: if you don't use the value...
```suggestion
  for (const xc of Object.keys(getCellGrid(model, range))) {
```

LucasLefevre:
And the feature is really cool :+1:

packages/o-spreadsheet-engine/src/helpers/pivot_table_presets.ts
hokolomopo: Yeah, renaming gone wrong ...

packages/o-spreadsheet-engine/src/plugins/ui_core_views/pivot_ui.ts
hokolomopo: Yeah you're right. Added a test.

packages/o-spreadsheet-engine/src/helpers/pivot_table_presets.ts
hokolomopo: But drawing grid lines just to paint a color over it is also useless :shrug:

packages/o-spreadsheet-engine/src/plugins/ui_core_views/formula_tracker.ts
```diff
@@ -0,0 +1,52 @@
+import { doesCellContainFormula } from "../../helpers";
+import { Command } from "../../types/commands";
+import { CoreViewPlugin } from "../core_view_plugin";
+
+const trackedFormulas = ["SUBTOTAL", "PIVOT"];
+
+export class FormulaTrackerPlugin extends CoreViewPlugin {
+  static getters = ["getCellsWithTrackedFormula"] as const;
+
+  private trackedCells: Record<string, Record<string, boolean | undefined>> = {};
```
pro-odoo: Nitpick

```ts
private trackedCells: Record<string, Record<UID, boolean | undefined>> = {};
```

packages/o-spreadsheet-engine/src/plugins/ui_core_views/formula_tracker.ts
```diff
@@ -0,0 +1,52 @@
+import { doesCellContainFormula } from "../../helpers";
+import { Command } from "../../types/commands";
+import { CoreViewPlugin } from "../core_view_plugin";
+
+const trackedFormulas = ["SUBTOTAL", "PIVOT"];
```
pro-odoo, line 6: Not a big deal, but `PIVOT` should not be added in this commit

packages/o-spreadsheet-engine/src/constants.ts
```diff
@@ -179,6 +179,7 @@ export const DEFAULT_STYLE = {
   fillColor: "",
```
pro-odoo, line 186: In commit `[IMP] style: add skipCellGridLines style option`
Type in the commit message: `This commit introducing` => `The commit introducing`
Also, it's worth to add the sha of the commit

src/stores/grid_renderer_store.ts
```diff
@@ -211,31 +212,31 @@ export class GridRenderer extends SpreadsheetStore {
       const style = box.style;
       if (style.fillColor && style.fillColor !== "#ffffff") {
         ctx.fillStyle = style.fillColor || "#ffffff";
-        ctx.fillRect(box.x, box.y, box.width, box.height);
+        ctx.fillRect(box.x - 0.5, box.y - 0.5, box.width + 1, box.height + 1);
```
pro-odoo: Maybe take the opportunity to make a constant with the full explanation ?

packages/o-spreadsheet-engine/src/helpers/misc.ts
```diff
@@ -717,3 +718,12 @@ export function chartStyleToCellStyle(style: ChartStyle): Style {
     align: style.align,
   };
 }
+
+export function doesCellContainFormula(cell: Cell, formula: string): boolean {
```
rrahir:
```suggestion
export function doesCellContainFunction(cell: Cell, function: string): boolean {
```

packages/o-spreadsheet-engine/src/constants.ts
rrahir: +1 on the original commit hash -> 7729508be41ef5e6abb244923be1cce8031bc928

tests/renderer/renderer_store.test.ts
```diff
@@ -75,6 +81,20 @@ function getBoxFromText(gridRenderer: GridRenderer, text: string): Box {
...
 }
 
+/**
+ * Cell fills are drawn with a 0.5 offset from the cell rect so they fill all of the cell and are not
+ * affected with the 0.5 offset we use to make the canvas lines sharp.
+ */
+function removeOffsetOfFillStyles(fillStyles: any[]): any[] {
+  return fillStyles.map((fs) => ({
+    ...fs,
+    x: fs.x + 0.5,
```
rrahir: use the CANVAS_OFFSET constant then ^^

packages/o-spreadsheet-engine/src/helpers/pivot_table_presets.ts
```diff
@@ -0,0 +1,489 @@
...
+    ...TABLE_COLOR_SETS.black,
+    highlight: TABLE_COLOR_SETS.gray.dark,
+    mediumBorder: TABLE_COLOR_SETS.black.medium,
+  }),
+};
+
+export function pivotTableStyleIdToTableStyleId(pivotStyleId: string): string {
+  const template = PIVOT_TABLE_PRESETS[pivotStyleId];
+  if (!template) {
+    return "None";
```
rrahir, line 479: consistency with None and none
https://github.com/odoo/o-spreadsheet/pull/7455/commits/28ba2934dce0b64f171c566312a69ddfe2d3468a#diff-d373219fc92e57efd2ffc091a49b752389519807b77d53df6d25643adfe61c5aR388

packages/o-spreadsheet-engine/src/helpers/pivot/table_spreadsheet_pivot.ts
```diff
@@ -418,6 +418,27 @@ export class SpreadsheetPivotTable {
...
+      headerRows = this.columns.length - 1;
+    }
+    if (pivotStyle.displayMeasuresRow) {
+      headerRows++;
+    }
+
+    return {
+      numberOfCols: Math.min(1 + pivotStyle.numberOfColumns, cells.length),
+      numberOfRows: Math.min(headerRows + pivotStyle.numberOfRows, cells[0].length),
+      headerRows,
```
LucasLefevre: nitpick: `headerRows` -> `numberOfHeaderRows` to be consistent with the two other properties ?

packages/o-spreadsheet-engine/src/plugins/ui_core_views/dynamic_tables.ts
```diff
@@ -93,6 +87,66 @@ export class DynamicTablesPlugin extends CoreViewPlugin {
     return tables;
   }
 
+  private getDynamicTables(sheetId: UID): DynamicTable[] {
+    const dynamicTablesFromCore = this.getters
+      .getCoreTables(sheetId)
+      .filter((table): table is DynamicTable => table.type === "dynamic");
```
LucasLefevre: explicit type guard not needed (since Typescript 5.5, it can infer this)
```suggestion
      .filter((table) => table.type === "dynamic");
```

packages/o-spreadsheet-engine/src/plugins/ui_core_views/formula_tracker.ts
```diff
@@ -0,0 +1,57 @@
...
+        }
+        break;
+      }
+      case "UPDATE_CELL": {
+        if (!("content" in cmd)) {
+          return;
+        }
+        const cell = this.getters.getCell(cmd);
+        if (!cell) {
+          return;
```
LucasLefevre, line 40: If the UPDATE_CELL has the effect of clearing the cell, it's not untracked because of this early return.
It doesn't look to be an issue right now in practice, but it's an unexpected behavior.

=> `beforeHandle` ?

packages/o-spreadsheet-engine/src/plugins/ui_core_views/formula_tracker.ts
```diff
@@ -0,0 +1,57 @@
...
+          }
+        }
+        break;
+      }
+    }
+  }
+
+  getCellsWithTrackedFormula(formula: string): string[] {
+    return Object.keys(this.trackedCells[formula] || {}).filter(
+      (cellId) => this.trackedCells[formula][cellId] && this.getters.tryGetCellPosition(cellId)
```
LucasLefevre, line 56: Oh ok, they are filtered here...
It would still be less hacky to reset with `beforeHandle` I think

src/components/tables/table_style_preview/table_style_preview.ts
```diff
@@ -40,16 +47,45 @@ export class TableStylePreview extends Component<Props, SpreadsheetChildEnv> {
         this.drawTable(nextProps);
       }
     });
-    onMounted(() => this.drawTable(this.props));
+    onMounted(() => {
+      resizeObserver.observe(this.canvasRef.el!);
+    });
+    onWillUnmount(() => {
+      resizeObserver.disconnect();
```
LucasLefevre: Could you explain why a resize observer is needed ?

hokolomopo: IIRC, when we open the pivot side panel, the design tab is in `d-none` so this component is created and try to draw the table preview on a canvas of size 0. So w/o the observer when we open the design side panel, the table is wrongly drawn.

The alternative would be to re-draw at each useEffect() instead of the current onWillUpdateProps, not sure which one is better

rrahir: you could use Element.isVisible as the condition of a useEffect to limit the renderings?

hokolomopo: I'd rather have a resizeObserver than an useless re-draw at each spreadsheet render, but I can change if we prefer the useEffect variation :shrug:

rrahir: Sorry, the correct API is checkVisibility  <https://developer.mozilla.org/en-US/docs/Web/API/Element/checkVisibility>

rrahir: As you prefer, I just saw the existence of that API very recently ^^

packages/o-spreadsheet-engine/src/plugins/ui_core_views/formula_tracker.ts
hokolomopo: cells are deleted in the beforeHandle of the sheet plugin, so we don't have the cell.id anymore in the beforeHandle of this plugin ...

src/components/tables/table_style_preview/table_style_preview.ts
LucasLefevre: I think I already made that comment somewhere.
It should be a `useEffect` because `this.canvasRef.el` might change

LucasLefevre:
robodoo rebase-ff r+

PR #7461, opened 2025-11-14
[FIX] pivot: raise pivot size limit

LucasLefevre:
## Description:

Currently, if a dynamic pivot table is more than 100k cells, the pivot is not displayed and the formula is in error.

When we upgraded odoo.com, we had to wait ~0 minute before finding important spreadsheets where the limit was reached.

This commit increase the limit from 100k to 500k

If it's not enough/too much, we'll add a setting to allow users to tune it.

Task: 0
## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


rrahir:
robodoo r+

PR #7467, opened 2025-11-14
[IMP] dark mode: improve contrasts

hokolomopo:
## Description

With this commit, the main background of the dark mode changed from `os-gray-100` to `os-gray-200`. The border colors were also changed to have an higher contrast with the new background color.

Task: [5265478](https://www.odoo.com/odoo/2328/tasks/5265478)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/spreadsheet/spreadsheet.css
+++ b/src/components/spreadsheet/spreadsheet.css
@@ -1,8 +1,32 @@
 .o-spreadsheet {
+  .border {
+    border: 1px solid var(--os-border-color) !important;
+  }
+
+  .border-bottom {
+    border-bottom: 1px solid var(--os-border-color) !important;
+  }
+
+  .border-top {
+    border-top: 1px solid var(--os-border-color) !important;
+  }
+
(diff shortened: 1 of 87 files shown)


src/components/spreadsheet/spreadsheet.css
```diff
@@ -1,16 +1,40 @@
 .o-spreadsheet {
+  .border {
+    border: 1px solid var(--os-border-color) !important;
+  }
```
hokolomopo, line 4: At the start I wanted to simply override `--bs-border-color` so I could use bootstrap classes ... but odoo overrides the `border*` classes so it doesn't use `--bs-border-color` ...

LucasLefevre: Are we sure it allows design theme to work ? (I have no idea how they work)

rrahir:
robodoo r+

PR #7468, opened 2025-11-17
[FIX] Demo: Fix wrapper style

rrahir:
When we introduced the mobile mode, we started relying on the dynamic vewport height (dvh) to account for the resizing of the viewport when the virtual keyboard of a smarphone would pop up.

However, the rule we chose was to set the height of spreadsheet wrapper to 100dvh, which stands for 100% of the dynamic viewport height. As such, if we were to add some header before the Spreadsheet component, the latter would keep its size as 100% of the viewport, meaning that the full component is pushed downwards and it overflows from the page.

This revision places the 100dvh rule over the full page to account for any header that could be added in the future.

Task: 5212448

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


pro-odoo (approved):
robodoo r+

PR #7471, opened 2025-11-18
[FIX] borders: preserve side borders when combining External and All

dhrp-odoo:
## Description:

When applying an External border on a range and then applying All borders on an inner/overlapping range, the side borders (left/right) of the inner cells were lost after re-opening the spreadsheet.

The exported border data was correct, but the BordersPlugin recomputation logic in addBorder incorrectly cleared adjacent borders on import/update. The adjacency check (adjacent(existingBorder.zone, zone)) reported the side of the existing zone, but we were deciding whether to clear it based on the same side key in the new border, instead of the opposite side (shared edge) on the new zone. As a result, importing the combination of:

C2:C4 with All borders
B2:B4 with left/top/bottom
D2:D4 with right/top/bottom

ended up clearing the left and right borders of C2:C4.

This PR adjusts the adjacent clearing logic to:

Map the adjacent side of the existing zone to the opposite side on the new zone.

Only clear the existing side if the corresponding opposite side is actually being written on the new border.

Task: [5270171](https://www.odoo.com/odoo/2328/tasks/5270171)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/core/borders.ts
+++ b/src/plugins/core/borders.ts
@@ -342,22 +342,35 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
   ) {
     const borders: ZoneBorder[] = [];
     const plannedBorder = newBorder ? { zone, style: newBorder } : undefined;
-    const sideToClear = {
-      left: force || !!newBorder?.left,
-      right: force || !!newBorder?.right,
-      top: force || !!newBorder?.top,
-      bottom: force || !!newBorder?.bottom,
+
+    // For each side, decide if we must clear the border on the *adjacent*
+    // existing cell when we draw on the opposite side of the new zone.
+    //
+    // Example:
+    //  - newBorder.right is set → we draw border on the RIGHT side of `zone`
+    //  - the cell on the right may already have a LEFT border on that edge
+    // In that case we clear that LEFT border, so only the new RIGHT border
+    // remains on the shared edge.
+    //
+    // existingBorderSideToClear[side] = true means we should clear the border on that
+    // side of the existing adjacent zone before adding the new border.
+    const existingBorderSideToClear = {
+      left: force || !!newBorder?.right,
+      right: force || !!newBorder?.left,
+      top: force || !!newBorder?.bottom,
+      bottom: force || !!newBorder?.top,
     };
     let editingZone: Zone[] = [zone];
     for (const existingBorder of this.borders[sheetId] ?? []) {
       const inter = intersection(existingBorder.zone, zone);
       if (!inter) {
-        // Clear adjacent borders on which you write
+        // Check if the existing border is adjacent to the new zone
         const adjacentEdge = adjacent(existingBorder.zone, zone);
-        if (adjacentEdge && sideToClear[adjacentEdge.position]) {
+        if (adjacentEdge && existingBorderSideToClear[adjacentEdge.position]) {
           for (const newZone of splitIfAdjacent(existingBorder.zone, zone)) {
             const border = this.computeBorderFromZone(newZone, existingBorder);
             const adjacentEdge = adjacent(newZone, zone);
+            // Clear the existing border on the side that touches the new zone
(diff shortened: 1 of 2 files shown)


rrahir:
I feel like applying the opposite border strategy (left <>right, top<>bottom) should actually be done in https://github.com/odoo/o-spreadsheet/pull/7471/files#diff-775eae5d11f82632d3abbe31764630b6ee83acd820f4251d4e2a10c2c1e4a122R345 (`sideToClear` definition), which seems to do the same thing you do in your PR, wdyt? 

I still fail to understand why this only impacts the import and is not triggered during the actual use but it might come from the fact that we never functionally have the opportunity to create a border with  external borders only applying to top,left,bottom and nothing on right. it might be interesting to see if such a command can be dispatched, and if so, it should be tested as well for good measure and the test might be more explicit than an import/export (at first glance, i think this can be achieved by dispatching `SET_BORDERS_ON_TARGET`).

dhrp-odoo:
@rrahir right, I just inverted the mapping in `sideToClear` (left/right, top/bottom) so it behaves as it was meant to.

Added a test to cover this via commands (and removed the import test, since it felt almost like a duplicate)

Thanks for taking the time to review this PR 🙂

tests/borders/border_plugin.test.ts
```diff
@@ -147,6 +147,36 @@ describe("borders", () => {
     });
   });
 
+  test("Preserves side borders when combining external and all via command", () => {
```
rrahir, line 150: I think we should be very thorough here because it also affects top/bottom similarly., add some dispatch with "empty" border adjacents to C2:C4 and check that every single cell is still protected.

tests/borders/border_plugin.test.ts
```diff
@@ -147,6 +147,36 @@ describe("borders", () => {
...
+    const model = new Model();
+    const defaultBorder = DEFAULT_BORDER_DESC;
+
+    setBordersOnTarget(model, ["C2:C4"], {
+      top: defaultBorder,
+      bottom: defaultBorder,
+      left: defaultBorder,
+      right: defaultBorder,
+    });
+    setZoneBorders(model, { position: "hv" }, ["C2:C4"]);
```
rrahir:
```suggestion
    setZoneBorders(model, { position: "all" }, ["C2:C4"]);
```
is equivalent to both commands you dispatch above, isn't it?

tests/borders/border_plugin.test.ts
```diff
@@ -147,6 +147,36 @@ describe("borders", () => {
...
+      bottom: defaultBorder,
+      left: defaultBorder,
+    });
+    setBordersOnTarget(model, ["D2:D4"], {
+      top: defaultBorder,
+      bottom: defaultBorder,
+      right: defaultBorder,
+    });
+
+    expect(getBorder(model, "C2")).toEqual({
```
rrahir, line 182: you should check every single cell and verify that they were not altered ;-)

src/plugins/core/borders.ts
```diff
@@ -343,10 +343,10 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
     const borders: ZoneBorder[] = [];
     const plannedBorder = newBorder ? { zone, style: newBorder } : undefined;
     const sideToClear = {
```
rrahir: this corresponds to the side to clear in the existing borders in case of adjacency (maybe the right term is contiguity, dunno) and no intersection
```suggestion
    const existingBorderSideToClear = {
```
maybe?

dhrp-odoo:
@rrahir Indeed, the implementation is a bit confusing, especially because the bug only shows up on the opposite side.

I’ve added a comment to clarify the logic and extended the test to also cover the top/bottom cases (took a bit of time to reproduce, it wasn’t obvious at first glance :smiley:).

rrahir:
robodoo r+

PR #7488, opened 2025-11-19
[FIX] side panel: close selection input when switching tab

hokolomopo:
When focusing the selection input in the chart side panel, changing tab would not close the selection input, even though the input isn't visible anymore.

It's because since 91c0583, changing the tab would keep the component mounted, only hiding it with `d-none` (to keep its state). Relying on the selection input `onWillUnmount` to reset the state is not enough.

We now rely on a `useEffect` that checks the `input.offsetParent` property, which is `null` when an ancestor has `display: none`.

Task: 5249165

Task: [5249165](https://www.odoo.com/odoo/2328/tasks/5249165)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/tests/setup/jest.setup.ts
+++ b/tests/setup/jest.setup.ts
@@ -87,6 +87,31 @@ beforeEach(() => {
       const blob = new window.Blob([data], { type });
       setTimeout(() => callback(blob), 0);
     });
+
+  // offsetParent should return the nearest positioned ancestor, or null if an ancestor has `display: none`
+  jest
+    .spyOn(HTMLElement.prototype, "offsetParent", "get")
+    .mockImplementation(function (this: HTMLElement) {
+      for (let element: HTMLElement | null = this; element; element = element.parentElement) {
+        if (getComputedStyle(element).display === "none" || element.classList.contains("d-none")) {
+          return null;
+        }
+      }
+
+      if (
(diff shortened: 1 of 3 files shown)


tests/setup/jest.setup.ts
```diff
@@ -87,6 +87,24 @@ beforeEach(() => {
       const blob = new window.Blob([data], { type });
       setTimeout(() => callback(blob), 0);
     });
+
+  // offsetParent should return the nearest positioned ancestor, or null if an ancestor has `display: none`
+  jest
+    .spyOn(HTMLElement.prototype, "offsetParent", "get")
+    .mockImplementation(function (this: HTMLElement) {
```
rrahir, line 94: did you have a look at https://github.com/jsdom/jsdom/issues/1261, the polyfill suggested seems to stick close to the spec I could see on MDN

hokolomopo: Sure I can add conditions for fixed position and html/body tags. But note that in our case we also need to manually check for bootstrap classes `d-none/position-fixed`

rrahir: >< indeed, forgot that...

rrahir:
robodoo r+

PR #7489, opened 2025-11-19
[IMP] Menu: Delegate enabled to children items

rrahir:
Currently, if a menu is evaluated as disabled, we disable it and prevent the possibility to see its children by hover. However, there are some issues with this:
- the user is mislead because we keep showing the arrow that suggest the presence of children items
- the user can simply not navigate through the submenus to see what kind of feature could be available
- if a menu has several children some of which are readonlyAllowed, then the readonlyAllowed condition has to be set on the parent as well; the 'enabling' condition therefore becomes a combination set on both the parent and children
- if a parent menu is computed as 'enabled' it will not be greyed out (suggests that it has useful children) even if the said children are not enabled.

This revision changes the computation of 'enabled' state of item by looking at the full tree of children and applies the following logic; A parent menu is disabled if every leaf of the tree is disabled. A single 'enabled' leaf (that is a child without children of its own) will make all its parent enabled as well, up to the root. We also allow all submenus to be opened regardless of the state of the parent.

The user will therefore have a visual hint that there are active submenus or not (if all children are disabled, so will be the parent) and they will also have the possibility to have a look at all children to explore the possibilities.

Task: 5331717

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/menu/menu.ts
+++ b/src/components/menu/menu.ts
@@ -118,10 +118,15 @@ export class Menu extends Component<MenuProps, SpreadsheetChildEnv> {
   }
 
   isEnabled(menu: Action) {
-    if (menu.isEnabled(this.env)) {
-      return this.env.model.getters.isReadonly() ? menu.isReadonlyAllowed : true;
+    const children = menu.children?.(this.env);
+    if (children.length) {
+      return children.some((child) => this.isEnabled(child));
+    } else {
+      if (menu.isEnabled(this.env)) {
+        return this.env.model.getters.isReadonly() ? menu.isReadonlyAllowed : true;
+      }
+      return false;
     }
-    return false;
   }
 
   get menuStyle() {
--- a/src/components/menu_popover/menu_popover.ts
+++ b/src/components/menu_popover/menu_popover.ts
@@ -245,17 +245,15 @@ export class MenuPopover extends Component<Props, SpreadsheetChildEnv> {
   }
 
   onMouseOver(menu: Action, ev: MouseEvent) {
-    if (this.isEnabled(menu)) {
-      if (this.isParentMenu(this.subMenu, menu)) {
-        this.openingTimeOut.clear();
-        return;
-      }
-      const currentTarget = ev.currentTarget as HTMLElement;
-      if (this.isRoot(menu)) {
-        this.openingTimeOut.schedule(() => {
-          this.openSubMenu(menu, currentTarget);
-        }, TIMEOUT_DELAY);
-      }
+    if (this.isParentMenu(this.subMenu, menu)) {
+      this.openingTimeOut.clear();
+      return;
+    }
+    const currentTarget = ev.currentTarget as HTMLElement;
+    if (this.isRoot(menu)) {
+      this.openingTimeOut.schedule(() => {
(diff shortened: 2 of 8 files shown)


src/components/menu_popover/menu_popover.ts
```diff
@@ -245,17 +245,15 @@ export class MenuPopover extends Component<Props, SpreadsheetChildEnv> {
...
-      }
+    if (this.isParentMenu(this.subMenu, menu)) {
+      this.openingTimeOut.clear();
+      return;
+    }
+    const currentTarget = ev.currentTarget as HTMLElement;
+    if (this.isRoot(menu)) {
+      this.openingTimeOut.schedule(() => {
+        this.openSubMenu(menu, currentTarget);
+      }, TIMEOUT_DELAY);
```
hokolomopo, line 256: I find it a bit strange UX-wise that the menu opens while there is a `cursor: not-allowed` and no hover/effect background. I'd remove the cursor and add back the hover effect for disabled parent menus

src/components/menu/menu.ts
```diff
@@ -118,10 +118,19 @@ export class Menu extends Component<MenuProps, SpreadsheetChildEnv> {
   }
 
   isEnabled(menu: Action) {
-    if (menu.isEnabled(this.env)) {
-      return this.env.model.getters.isReadonly() ? menu.isReadonlyAllowed : true;
+    const children = menu.children?.(this.env);
+    if (children.length) {
+      const areChildrenEnabled = children.length
```
hokolomopo: you check children.length twice

hokolomopo: can just write `if(chidren.length) return children.some((child) => this.isEnabled(child)); `

tests/menus/context_menu_component.test.ts
```diff
@@ -449,33 +449,51 @@ describe("Context MenuPopover internal tests", () => {
...
-        isEnabled: () => false,
-        children: [makeTestMenuItem("subMenu")],
-      }),
-    ]);
-    await renderContextMenu(300, 300, { menuItems });
-    expect(fixture.querySelector(".o-menu div[data-name='root']")!.classList).toContain("disabled");
-    await simulateClick(".o-menu div[data-name='root']");
-    expect(fixture.querySelector(".o-menu div[data-name='subMenu']")).toBeFalsy();
-  });
+  describe("'IsEnabled' is ignored on parent menu items", () => {
```
hokolomopo: nitpick: don't use quotes in test names, we cannot use jest debugger then

hokolomopo (commented):
👋

src/actions/action.ts
```diff
@@ -117,14 +117,15 @@ export function createAction(item: ActionSpec): Action {
...
-    children: children
-      ? (env) => {
-          return children
-            .map((child) => (typeof child === "function" ? child(env) : child))
-            .flat()
-            .map(createAction);
-        }
-      : () => [],
+    children:
+      children.length > 0
```
LucasLefevre: as investigated: remove `createActions` from `formatNumberMenuItemSpec`

that was a funny one

PR #7490, opened 2025-11-19
[IMP] CF: add new conditional formats for dates

hokolomopo:
### [IMP] CF: add new conditional formats for dates

This commit adds the `dateIs/dateIsBefore/dateIsAfter` conditional
formats, as well as the excel export for these.

Note:
The excel export doesn't use Excel's date conditional formats, because
they don't have the same behaviour as ours. We'll use custom formulas
instead.

Example: today is 2025/11/19
- Our date is last month: 2025/10/19 -> 2025/11/18
- Excel's date is last month: 2025/10/01 -> 2025/10/30

Task: [5343283](https://www.odoo.com/odoo/2328/tasks/5343283)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/packages/o-spreadsheet-engine/src/types/conditional_formatting.ts
+++ b/packages/o-spreadsheet-engine/src/types/conditional_formatting.ts
@@ -1,3 +1,4 @@
+import { DateCriterionValue } from "./generic_criterion";
 import { Style, UID } from "./misc";
 import { Range } from "./range";
 
@@ -54,6 +55,7 @@ export interface CellIsRule extends SingleColorRule {
   operator: ConditionalFormattingOperatorValues;
   // can be one value for all operator except between, then it is 2 values
   values: string[];
+  dateValue?: DateCriterionValue;
 }
 export interface ExpressionRule extends SingleColorRule {
   type: "ExpressionRule";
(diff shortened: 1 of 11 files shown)


demo/main.js
```diff
@@ -263,13 +263,13 @@ class Demo extends Component {
...
     }
-    this.createModel(data || demoData);
+    // this.createModel(data || demoData);
     // this.createModel(makePivotDataset(10_000));
     // this.createModel(makeLargeDataset(26, 10_000, ["numbers"]));
     // this.createModel(makeLargeDataset(26, 10_000, ["formulas"]));
     // this.createModel(makeLargeDataset(26, 10_000, ["arrayFormulas"]));
     // this.createModel(makeLargeDataset(26, 10_000, ["vectorizedFormulas"]));
-    // this.createModel({});
+    this.createModel({});
```
anhe-odoo: @hokolomopo Are you sure about this ?

packages/o-spreadsheet-engine/src/types/conditional_formatting.ts
```diff
@@ -169,6 +171,11 @@ export type ConditionalFormattingOperatorValues =
...
   | "notContainsText"
   | "isNotEqual"
+  | "dateIs"
+  | "dateIsBefore"
+  | "dateIsAfter"
+  | "dateIsOnOrBefore"
+  | "dateIsOnOrAfter"
   | "customFormula";
 
 export const availableConditionalFormatOperators: Set<ConditionalFormattingOperatorValues> =
```
anhe-odoo, line 185: Maybe we can deduct the type from the following set to avoid having to change at everytime we add/remove a new conditional format operator ?

src/components/spreadsheet/spreadsheet.ts
```diff
@@ -191,6 +191,29 @@ export class Spreadsheet extends Component<SpreadsheetProps, SpreadsheetChildEnv
 
     const render = batched(this.render.bind(this, true));
     onMounted(() => {
+      // setTimeout(() => {
```
anhe-odoo: Commented code ?

packages/o-spreadsheet-engine/src/xlsx/functions/conditional_formatting.ts
```diff
@@ -116,6 +118,50 @@ function cellRuleFormula(ranges: string[], rule: CellIsRule): string[] {
...
+        case "lastMonth":
+          return [`AND(${firstCell}>=EDATE(TODAY(),-1),${firstCell}<TODAY())`];
+        case "lastYear":
+          return [`AND(${firstCell}>=EDATE(TODAY(),-12),${firstCell}<TODAY())`];
+      }
+    case "dateIsBefore":
+    case "dateIsAfter":
+    case "dateIsOnOrAfter":
+    case "dateIsOnOrBefore":
+      switch (rule.dateValue || "exactDate") {
```
pro-odoo: Having `rule.dateValue` undefined at this point is an error, right ?
I would add
```js
case undefined:
    throw new Error("dateValue should be defined at this point")
```
It avoid a future hidden bug :D 

Same for `dateIs`

pro-odoo (commented):
Just one little comment, then go

packages/o-spreadsheet-engine/src/xlsx/functions/conditional_formatting.ts
hokolomopo: Done :)

PR #7493, opened 2025-11-19
[FIX] GridOverlay: Prevent mousedown default when selecting range

rrahir:
When selecting a range in the grid and the composer is open, passing your selection around the composer will see its content selected as if we were actually selecting text inside the composer.

This behaviour is a direct effect of the default behaviour of the mousedown event that starts the selection. Since the focus is kept on the composer at all times, the browser behaves as if we had clicked on the composer itself even if from a user point of view, it makes little sense.

Task: 5344285

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/grid_overlay/grid_overlay.ts
+++ b/src/components/grid_overlay/grid_overlay.ts
@@ -228,6 +228,7 @@ export class GridOverlay extends Component<Props, SpreadsheetChildEnv> {
       // not main button, probably a context menu
       return;
     }
+    ev.preventDefault();
     const [col, row] = this.getCartesianCoordinates(ev);
     this.props.onCellClicked(col, row, {
       expandZone: ev.shiftKey,


rrahir:
robodoo fw no

rrahir:
robodoo fw=no

src/components/composer/composer/composer.xml
```diff
@@ -12,7 +12,7 @@
         t-on-keydown="onKeydown"
         t-on-mousewheel.stop=""
         t-on-input="onInput"
-        t-on-mousedown="onMousedown"
+        t-on-mousedown.prevent="onMousedown"
```
hokolomopo: I really don't think we want to do that, that prevents clicks from selecting text in the composer. And I got tracebacks 

https://drive.google.com/file/d/1hmY3Cq8ZxzGQuadOd7zTwrz6zBwtnwDp/view?usp=drive_link

rrahir: I pushed a fucked up wip branch everythingis supposed to take place in the dnd hook (and I even had tests ...)

rrahir: correction: I mixed 2 tasks, this PR did not have any tests but the fix is. supposed to be a one liner

PR #7495, opened 2025-11-20
[FIX] topbar: close font size editor dropdown and focus grid

dhrp-odoo:
## Description:

Pressing Tab in the font size editor moves focus to the hidden “add more rows” footer (the next focusable element in the DOM). This caused the browser to auto-scroll to the footer while the grid viewport state remained unchanged, making the footer appear to float over the grid.

Now, when the font size editor loses focus (via blur/Tab), it closes its dropdown and redirects focus back to the grid instead of the footer.

Task: [5263792](https://www.odoo.com/odoo/2328/tasks/5263792)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/font_size_editor/font_size_editor.ts
+++ b/src/components/font_size_editor/font_size_editor.ts
@@ -1,6 +1,8 @@
 import { Component, useExternalListener, useRef, useState } from "@odoo/owl";
 import { FONT_SIZES } from "../../constants";
 import { clip } from "../../helpers/index";
+import { Store, useStore } from "../../store_engine";
+import { DOMFocusableElementStore } from "../../stores/DOM_focus_store";
 import { SpreadsheetChildEnv } from "../../types/index";
 import { css } from "../helpers/css";
 import { isChildEvent } from "../helpers/dom_helpers";
@@ -62,7 +64,10 @@ export class FontSizeEditor extends Component<Props, SpreadsheetChildEnv> {
   private rootEditorRef = useRef("FontSizeEditor");
   private fontSizeListRef = useRef("fontSizeList");
 
+  private DOMFocusableElementStore!: Store<DOMFocusableElementStore>;
+
   setup() {
+    this.DOMFocusableElementStore = useStore(DOMFocusableElementStore);
     useExternalListener(window, "click", this.onExternalClick, { capture: true });
   }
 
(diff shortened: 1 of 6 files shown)


tests/top_bar_component.test.ts
```diff
@@ -339,6 +339,19 @@ describe("TopBar component", () => {
     expect(getStyle(model, "A1").fontSize).toBe(8);
   });
 
+  test("Tab from font size editor closes the dropdown", async () => {
+    const { fixture } = await mountSpreadsheet();
+    const input = fixture.querySelector("input.o-font-size") as HTMLInputElement;
+    input.focus();
+    await nextTick();
+    expect(fixture.querySelector(".o-dropdown-content.o-text-options")).toBeTruthy();
+    input.blur();
```
dhrp-odoo: I think jsdom doesn’t implement real Tab-based focus navigation, so we call blur() directly here to reliably trigger the blur handler in this test.

src/components/font_size_editor/font_size_editor.ts
```diff
@@ -70,7 +69,7 @@ export class FontSizeEditor extends Component<Props, SpreadsheetChildEnv> {
     const isOpen = this.dropdown.isOpen;
     if (!isOpen) {
       this.props.onToggle();
-      this.inputRef.el!.focus();
```
hokolomopo, line 88: why remove this ? don't we want the input focused when clicking on the arrow ?

dhrp-odoo: Actually, if we keep the font editor input focused when clicking the arrow, the dropdown becomes impossible to close.
When we click the arrow to close it, the input first triggers a blur event (which closes the dropdown), and then the click handler on the arrow runs and opens it again. Since blur happens before click, the two handlers cancel each other out.

I tried the mousedown event as well (as it is called before the blur), but then we need to replace all the click events with mousedown in this component. I am not sure if that is good or not.

If we really want the focus on the font editor input, then I guess better to remove the blur handler and directly handle the `Tab` event inside the component? What do you think? Let me know if you have a better approach

hokolomopo: Oh yeah this sounds annoying ... maybe we could close the dropwdown on an external click event rather than on blur ?

dhrp-odoo: It’s already handled; I just also wanted it to close when we press Tab in the font size input, so focus returns to the grid, which selects the cell and moves to the right.

hokolomopo: So handling tab manually with a prevent default should do the trick I guess. Not that pretty but :shrug:

dhrp-odoo: Yeah, I tried all the other approaches but couldn’t find anything better. All the solutions feel hacky (creating a boolean to skip the calls, applying mousedowns)

LucasLefevre (commented):
Thanks :)
robodoo r+

PR #7496, opened 2025-11-20
[FIx] data validation: fix date criterion

hokolomopo:
## Description

The relative date criterion 'pastWeek/pastMonth/pastYear' were not correct. For example the 'pastWeek' criterion would span a period of 8 days instead of 7. The dates numbers were also not always rounded down, which could lead to some issues.

Task: [5343580](https://www.odoo.com/odoo/2328/tasks/5343580)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/data_validation_helpers.ts
+++ b/src/helpers/data_validation_helpers.ts
@@ -1,22 +1,44 @@
 import { tryToNumber } from "../functions/helpers";
 import { DataValidationCriterion, DateCriterionValue, Locale } from "../types";
-import { DateTime, jsDateToNumber, valueToDateNumber } from "./dates";
+import { DateTime, getDaysInMonth, jsDateToNumber, valueToDateNumber } from "./dates";
 
 function toCriterionDateNumber(dateValue: Exclude<DateCriterionValue, "exactDate">): number {
   const today = DateTime.now();
   switch (dateValue) {
     case "today":
-      return jsDateToNumber(today);
-    case "yesterday":
-      return jsDateToNumber(DateTime.fromTimestamp(today.setDate(today.getDate() - 1)));
-    case "tomorrow":
-      return jsDateToNumber(DateTime.fromTimestamp(today.setDate(today.getDate() + 1)));
+      return Math.floor(jsDateToNumber(today));
+    case "yesterday": {
+      today.setDate(today.getDate() - 1);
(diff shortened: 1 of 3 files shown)


src/helpers/data_validation_helpers.ts
```diff
@@ -6,17 +6,27 @@ function toCriterionDateNumber(dateValue: Exclude<DateCriterionValue, "exactDate
...
     case "lastWeek":
-      return jsDateToNumber(DateTime.fromTimestamp(today.setDate(today.getDate() - 7)));
-    case "lastMonth":
-      return jsDateToNumber(DateTime.fromTimestamp(today.setMonth(today.getMonth() - 1)));
+      today.setDate(today.getDate() - 6);
+      return Math.floor(jsDateToNumber(today));
+    case "lastMonth": {
+      today.setDate(today.getDate() + 1);
+      today.setMonth(today.getMonth() - 1);
+      return Math.floor(jsDateToNumber(today));
```
LucasLefevre, line 30: I don't think this works.
I simulated if we were the 2025-03-30 (`const today = new DateTime(2025,2,30);`) and this gives `Mon Mar 03 2025` as a result

hokolomopo: Yeah dates are annoying ....

src/helpers/data_validation_helpers.ts
```diff
@@ -1,22 +1,44 @@
...
+      today.setDate(today.getDate() + 1);
+      return Math.floor(jsDateToNumber(today));
+    }
     case "lastWeek":
-      return jsDateToNumber(DateTime.fromTimestamp(today.setDate(today.getDate() - 7)));
-    case "lastMonth":
-      return jsDateToNumber(DateTime.fromTimestamp(today.setMonth(today.getMonth() - 1)));
+      today.setDate(today.getDate() - 6);
+      return Math.floor(jsDateToNumber(today));
+    case "lastMonth": {
```
LucasLefevre, line 21: it would be a behavior change (for master) but excel works differently (for CF).
For "last month", if you are any day in December, it targets any day between the 1st November to 31th November

LucasLefevre:
robodoo r-

LucasLefevre:
robodoo r+
fixed typo in commit title

PR #7506, opened 2025-11-21
[FIX] clickable cells: prevent overlap with grid icons

hokolomopo:
## Description

If a cell has both a clickable cell and a clickable grid icon (eg. a link in a cell with a filter icon), clicking the icon would click the clickable cell.

With this commit, we reduce the clickable cell size based on the position of the icons on the cell to prevent any overlap.

Task: [4930803](https://www.odoo.com/odoo/2328/tasks/4930803)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


pro-odoo (approved):
robodoo r+

PR #7523, opened 2025-11-24
[FIX] composer: keep edition mode inactive on cursor selection change

rmbh-odoo:
## Description:

Current behavior before PR:
- Changing the composer cursor selection triggered a switch to edition mode,
even when the composer was inactive.
- This caused unwanted activation of editing while simply moving the selection.

- Selection changes no longer modify the edition mode when the composer is inactive,
ensuring the edition state remains stable.

Task: [5354541](https://www.odoo.com/odoo/2328/tasks/5354541)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_stateful/edition.ts
+++ b/src/plugins/ui_stateful/edition.ts
@@ -144,7 +144,9 @@ export class EditionPlugin extends UIPlugin {
       case "CHANGE_COMPOSER_CURSOR_SELECTION":
         this.selectionStart = cmd.start;
         this.selectionEnd = cmd.end;
-        this.mode = "editing";
+        if (this.isSelectingForComposer()) {
+          this.mode = "editing";
+        }
         break;
       case "STOP_COMPOSER_RANGE_SELECTION":
         if (this.isSelectingForComposer()) {
--- a/tests/composer/edition_plugin.test.ts
+++ b/tests/composer/edition_plugin.test.ts
@@ -103,6 +103,13 @@ describe("edition", () => {
     expect(model.getters.getEditionMode()).toBe("inactive");
   });
 
+  test("should keep edition mode inactive when selection changes while composer is inactive", () => {
+    const model = new Model();
+    expect(model.getters.getEditionMode()).toBe("inactive");
+    model.dispatch("CHANGE_COMPOSER_CURSOR_SELECTION", { start: 0, end: 0 });
+    expect(model.getters.getEditionMode()).toBe("inactive");
+  });
+
   test("should switch to editing mode when composer cursor selection changes", () => {
     const model = new Model();
     model.dispatch("START_EDITION", { text: "=sum(" });
--- a/tests/composer/composer_integration_component.test.ts
+++ b/tests/composer/composer_integration_component.test.ts
@@ -27,6 +27,7 @@ import {
   getElComputedStyle,
   gridMouseEvent,
   keyDown,
+  keyUp,
   rightClickCell,
   selectColumnByClicking,
   simulateClick,
@@ -336,6 +337,15 @@ describe("Composer interactions", () => {
     composerEl.dispatchEvent(new Event("keyup"));
     await clickCell(model, "C8");
     expect(getSelectionAnchorCellXc(model)).toBe("C8");
+    expect(model.getters.getEditionMode()).toBe("inactive");
+  });
+
+  test("should switch topbar composer from editing to inactive when pressing Escape on cell A1 containing '=A2'", async () => {
+    setCellContent(model, "A1", "=A2");
+    await click(fixture, ".o-spreadsheet-topbar .o-composer");
+    expect(model.getters.getEditionMode()).toBe("editing");
+    keyDown({ key: "Escape" });
+    keyUp({ key: "Escape" });
     expect(model.getters.getEditionMode()).toBe("inactive");
   });
 


tests/composer/edition_plugin.test.ts
```diff
@@ -103,6 +103,13 @@ describe("edition", () => {
     expect(model.getters.getEditionMode()).toBe("inactive");
   });
 
+  test("should keep edition mode inactive when selection changes while composer is inactive", () => {
```
rrahir, line 106: Could you also add a test that tests against the component as well ? right now, if I set =A2 in A1 then select the topbar composer then hit escape, I see a traceback because of the inconsistent state.

src/plugins/ui_stateful/edition.ts
```diff
@@ -144,7 +144,9 @@ export class EditionPlugin extends UIPlugin {
       case "CHANGE_COMPOSER_CURSOR_SELECTION":
         this.selectionStart = cmd.start;
         this.selectionEnd = cmd.end;
-        this.mode = "editing";
+        if (this.isSelectingForComposer()) {
```
rrahir, line 147: there is another issue which is worth being taken care of.
during the onKeyup handler after pressing Escape, the composer selection is not reset,I think we should add this behaviour as well.
here, we should reset the selection as well when clearing the edition
https://github.com/odoo/o-spreadsheet/blob/17.0/src/plugins/ui_stateful/edition.ts#L449

edit: the dispatch CHANGE_COMPOSER_CURSOR_SELECTION should never be called in the first place and it wouldn't if the slection was properly reset when we stop the edition

rrahir (changes requested):
nice catch, I was fixing that in 18.0 but did not realise it was already an issue uin 17.0 as well.
Could you addres the second point I mentionned in a separate commit?

src/plugins/ui_stateful/edition.ts
rmbh-odoo: I investigated the issue you pointed out, and here is what I observed:
We are not dispatching `stopEdition` on Escape key press.
So, we can add the selection reset in `cancelEdition()`, because this method is triggered when pressing Escape: [composer.ts#L312](https://github.com/odoo/o-spreadsheet/blob/17.0/src/components/composer/composer/composer.ts#L312)

I added the following code to `cancelEdition()`:
`this.selectionStart = this.selectionEnd = 0;`

This solution works perfectly when the composer contains a single-token formula like '=A2'.
In that case, start and end are already 0 after pressing Escape, so there is no dispatch.

However, with a multi-token formula (e.g. '=A1+A2'), the old values oldStart and oldEnd are 0, but the values retrieved from [composer.ts#L418](https://github.com/odoo/o-spreadsheet/blob/17.0/src/components/composer/composer/composer.ts#L418)
are equal to the formula length. Because they do not match, the system ends up dispatching at the end.

As of now, I do not see any UI bug caused by this behavior.
Let me know if I missed anything in the R&D part. :smiling_face_with_tear:

rrahir: > However, with a multi-token formula (e.g. '=A1+A2'), the old values oldStart and oldEnd are 0, but the values retrieved from composer.ts#L418

The issue we have at this specific place is that we mock  completely the content helper (this was removed in upcoming versions 190+ I think) which might explain what you witness ? on my end, experimentally,  I see that on any type of formula (single or multi token) `this.contentHelper.getCurrentSelection()` is set to 0, 0.

could you confirm it is a test issue? if so, the mock could maybe be adapted ? or maybe it cant eassily, in which case we could consider adding a proper test in later versions.

What we drastically lack is a check on the coherence of the composer state (be it in 17.0 with th plugin or in later versions with the store) which brings those shitty behaviours. and it turns out that when both validating and canceling the edition, we do not reset the selection, I think it makes sense to address both cases. In 17.0, if you look at the state of the plugin once we stop the edition, its selection is still set to the value before we stopped, it's not reset to 0,0 but resetting that value in `EditionPlugin.cancelEdition` does the trick because Sthe handler ends up using cancelEdition

rrahir: Another point, the more I think about it, the more it makes no sense to try to post process something once edition is stopped. IMHO, any keyup event should be blatantly ignored once the eiditon is stopped, but we are doiong enough risky changes in stable at the moment. we could consider a change in master

rmbh-odoo: > could you confirm it is a test issue? if so, the mock could maybe be adapted ? or maybe it cant eassily, in which case we could consider adding a proper test in later versions.

Sorry, this isn’t related to single vs. multi token, and it’s not related to test cases.
In the [video](https://drive.google.com/file/d/17B6Na6FCiCk2HyA7IewxbsMeR80z7X1e/view?usp=sharing), pressing Esc on the `top composer` resets the selection to 0, but pressing Esc on the `grid composer` gives a length of 6. I’m not sure why this difference occurs.
We can’t rely on resetSelection to 0 because of this inconsistent behavior. it might be work for `stopSelection`
For the stable version, we can simply avoid dispatching in the keyup handler when the **composer is inactive**. In that case, we don’t need to worry about the selection at all. WDYT?

rrahir: > For the stable version, we can simply avoid dispatching in the keyup handler when the composer is inactive

this is exacttly what I think we should not do becasue we will notice later on that it was used on purpose (or not) and we will need to fix it again.  Let's keep the fix as it is right now  to stop breaking the stable version  :p

rrahir:
robodoo r+

PR #7527, opened 2025-11-25
[FIX] header_size: wrong row size from wrapped number

hokolomopo:
## Description

If a literal cell containing a number was wrapped (because the number was large, or because it had a long format), the row height was wrong.

Task: [4878338](https://www.odoo.com/odoo/2328/tasks/4878338)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/text_helper.ts
+++ b/src/helpers/text_helper.ts
@@ -8,7 +8,10 @@ import {
   NEWLINE,
   PADDING_AUTORESIZE_VERTICAL,
 } from "../constants";
-import { Cell, Pixel, PixelPosition, Style } from "../types";
+import { Cell, Locale, Pixel, PixelPosition, Style } from "../types";
+import { CellErrorType } from "../types/errors";
+import { parseLiteral } from "./cells";
+import { formatValue } from "./format";
 import { isMarkdownLink, parseMarkdownLink } from "./misc";
 
 export function computeTextLinesHeight(textLineHeight: number, numberOfLines: number = 1) {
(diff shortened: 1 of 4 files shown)


src/helpers/text_helper.ts
```diff
@@ -21,12 +24,22 @@ export function computeTextLinesHeight(textLineHeight: number, numberOfLines: nu
...
-  const content = cell.isFormula ? "" : cell.content;
+  let content = "";
+
+  try {
+    if (!cell.isFormula) {
+      const localeFormat = { format: cell.format, locale: DEFAULT_LOCALE };
+      content = formatValue(parseLiteral(cell.content, locale), localeFormat);
+    }
+  } catch {
+    content = CellErrorType.GenericError;
```
rrahir, line 41: can this fail for other reasons than an invalid format? 
and is it possible in real life conditions to have an invalid format?

hokolomopo: Yeah only for invalid formats. I don't think it should happens in practice, but we do have a test with invalid format that fails without the try/catch so ... (`Update cell with a format is correctly set`)

rrahir: we should discuss the possibility/necessity to reject invalid formats then ^^

hokolomopo: Probably yeah. Not sure we want to do it in 17.0 tho

src/helpers/text_helper.ts
```diff
@@ -21,12 +24,22 @@ export function computeTextLinesHeight(textLineHeight: number, numberOfLines: nu
...
 ) {
   if (!cell || (!cell.isFormula && !cell.content)) {
     return DEFAULT_CELL_HEIGHT;
   }
-  const content = cell.isFormula ? "" : cell.content;
+  let content = "";
+
+  try {
+    if (!cell.isFormula) {
+      const localeFormat = { format: cell.format, locale: DEFAULT_LOCALE };
```
rrahir:
```suggestion
      const localeFormat = { format: cell.format, locale: locale };
```
If I create a (garbage) format where the thousand separators are more than 1 charater long or the charater is bigger than a comma (as in the default one) the computation is wrong

rrahir: I think it also means that the size should be recomputed on locale change, don't think it's the case rn

tests/headers/resizing_plugin.test.ts
```diff
@@ -441,6 +448,37 @@ describe("Model resizer", () => {
...
+    });
+
+    test("updating locale updates the row size", () => {
+      setStyle(model, "C1", { fontSize: 10, wrapping: "wrap" });
+      resizeColumns(model, ["C"], 100);
+      setCellContent(model, "C1", "2000");
+      setFormat(model, "#,##", target("C1"));
+
+      expect(model.getters.getRowSize(sheet.id, 0)).toBe(DEFAULT_CELL_HEIGHT);
+      const locale = { ...DEFAULT_LOCALE, thousandsSeparator: "garbageLongThousandSeparator" };
```
rrahir, line 472: image

rrahir:
robodoo r+

PR #7531, opened 2025-11-25
[IMP] calendar chart: automatically detect calendar charts

hokolomopo:
### [FIX] calendar chart: hide scales borders

The calendar chart grid lines aren't shown, but the borders between
chartArea and the ticks labels was still shown. Depending on the
screen, zoom, and size of the chart, the border would be more or less
visible. This commit hides these borders.

### [FIX] calendar chart: fix typing

The type `CalendarChartGranularity` would accept granularities that
are only valid for pivots and not for calendar charts.

### [IMP] calendar chart: automatically detect calendar chart

If a chart is created on a dataset with dates in the first column
and numbers in the second column, we should create a calendar chart
by default.

Task: [5357420](https://www.odoo.com/odoo/2328/tasks/5357420)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/chart/chart_type_picker/chart_previews.xml
+++ b/src/components/side_panel/chart/chart_type_picker/chart_previews.xml
@@ -377,18 +377,21 @@
   </t>
   <t t-name="o-spreadsheet-ChartPreview.CALENDAR_CHART">
     <svg class="o-chart-preview" viewBox="0 0 48 48">
-      <path fill="#444" d="M3 3h42v42H3zh1v41h40V4H4"/>
+      <path fill="var(--os-gray-600)" d="M3 3h42v42H3zh1v41h40V4H4"/>
       <path
-        fill="#eb6d00"
+        fill="var(--os-chart-preview-orange-line)"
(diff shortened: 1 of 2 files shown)


src/helpers/figures/charts/smart_chart_engine.ts
```diff
@@ -198,14 +198,16 @@ function buildTwoColumnChart(columns: ColumnInfo[], getters: Getters): ChartDefi
...
-  // TODO: Handle date + number with calendar chart when implemented (and change the docstring)
   if (columns[0].type === "date" && columns[1].type === "number") {
     return {
-      ...DEFAULT_LINE_CHART_CONFIG,
-      type: "line",
+      type: "calendar",
       dataSets: [{ dataRange: getUnboundRange(getters, columns[1].zone) }],
       labelRange: getUnboundRange(getters, columns[0].zone),
       dataSetsHaveTitle: isDatasetTitled(getters, columns[0]),
+      title: {},
```
rrahir: functionally speaking, I am not convinced at all. If I want to see a day-to-day evolution of the sales of my shop, I would not be happy seeing a calendar chart which would make 0 sense.
Assuming that we want that feature, it should at least look for dates that span over several days, otherwise it becomes really pointless

rrahir (changes requested):

pro-odoo:
robodoo rebase-ff r+

PR #7533, opened 2025-11-26
[FIX] spreadsheet: dark mode display issues

dhrp-odoo:
## Description:

- Update rotation icons to use `currentColor` so they follow the active theme palette.
- Force data bar conditional format preview text to use the dark `TEXT_BODY` color, regardless of theme, as it was previously changing with the theme and becoming hard to read.

Task: [5322179](https://www.odoo.com/odoo/2328/tasks/5322179)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/components/side_panel/conditional_formatting/cf_preview/cf_preview.ts
```diff
@@ -40,8 +40,9 @@ export class ConditionalFormatPreview extends Component<Props, SpreadsheetChildE
         ? baseString + minColor + ", " + midColor + ", " + maxColor + ")"
         : baseString + minColor + ", " + maxColor + ")";
     } else if (rule.type === "DataBarRule") {
-      const color = colorNumberToHex(rule.color);
-      return `background-image: linear-gradient(to right, ${color} 50%, white 50%)`;
+      const barColor = colorNumberToHex(rule.color);
+      const gradient = `background-image: linear-gradient(to right, ${barColor} 50%, white 50%)`;
+      return `${gradient}; color: ${TEXT_BODY};`;
```
dhrp-odoo, line 45: Image

I don’t think this is actually an issue. The text color should be white and the preview’s right-side background should be black to match the grid, so that users get an accurate idea of how the data bar will look. 

Since the dark grid hasn’t been implemented yet, it’s probably better to address this as part of that task ???

hokolomopo: Yeah the preview will need to be adapted when we implement the grid in dark theme. In the meantime, we should match the grid as you said, so a black text is probably better :)

PR #7534, opened 2025-11-26
[FIX] pivot: Ensure computed measure range adaptation

rrahir:
## Description:
Currently, ODOO pivots computed measures are not properly updated upon
sheet structure modification. To be precise, their definition, which is
stored in the core plugin `PivotCorePlugin` is properly updated but the
runtime definition, stored in `PivotUIPlugin`, is not.

This occurs because the mecanism to invalidate the runtime definition
explicitely ignores the ODOO pivots. histoically, this was set up to
avoid useless reloading of ODOO pivots which could end up making server
calls but this logic is properly handled in the function `onDefinitionChange`.

We can see that in the case of spreadsheet pivots, we already
notify all plugins of such a change, but by "pure accident", as we
dispatch an "UPDATE_PIVOT" command at every range adaptation, regardless
of whether it was necessary or not. This means that the spreadsheet
pivots beneficiated of two mecanisms to update their runtime (in core,
an UPDATE_PIVOT, and the `invalidateEvaluationCommands` mecanism) which
means that invalidation work was done two times.

The investigation also led to the discovery of a missing check on the
command "ADD_PIVOT" which has been reported in  https://www.odoo.com/odoo/2328/tasks/5360591

We also noted that there is a double handling of commands between the
handling of `invalidateEvaluationCommands` and the specific command
handlers in `PivotUIPlugin`. We could clean this up in master.

Note that additional tests regarding the Odoo pivots will be added in
Odoo repository to ensure the validity of the fix.
Task: [5358213](https://www.odoo.com/odoo/2328/tasks/5358213)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_core_views/pivot_ui.ts
+++ b/src/plugins/ui_core_views/pivot_ui.ts
@@ -2,6 +2,7 @@ import { Token } from "../../formulas";
 import { astToFormula } from "../../formulas/parser";
 import { toScalar } from "../../functions/helper_matrices";
 import { toBoolean } from "../../functions/helpers";
+import { deepCopy } from "../../helpers";
 import {
   getFirstPivotFunction,
   getNumberOfPivotFunctions,
@@ -65,9 +66,7 @@ export class PivotUIPlugin extends UIPlugin {
   handle(cmd: Command) {
     if (invalidateEvaluationCommands.has(cmd.type)) {
       for (const pivotId of this.getters.getPivotIds()) {
-        if (!pivotRegistry.get(this.getters.getPivotCoreDefinition(pivotId).type).externalData) {
-          this.setupPivot(pivotId, { recreate: true });
-        }
+        this.setupPivot(pivotId, { recreate: true });
       }
     }
     switch (cmd.type) {
@@ -302,7 +301,7 @@ export class PivotUIPlugin extends UIPlugin {
   }
 
   setupPivot(pivotId: UID, { recreate } = { recreate: false }) {
-    const definition = this.getters.getPivotCoreDefinition(pivotId);
+    const definition = deepCopy(this.getters.getPivotCoreDefinition(pivotId));
     if (!(pivotId in this.pivots)) {
       const Pivot = withPivotPresentationLayer(pivotRegistry.get(definition.type).ui);
       this.pivots[pivotId] = new Pivot(this.custom, { definition, getters: this.getters });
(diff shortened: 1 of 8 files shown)


src/plugins/ui_core_views/pivot_ui.ts
```diff
@@ -63,13 +63,6 @@ export class PivotUIPlugin extends UIPlugin {
   }
 
   handle(cmd: Command) {
-    if (invalidateEvaluationCommands.has(cmd.type)) {
```
rrahir, line 66: most likely breaking the undo/redo reevaluation :o

src/plugins/ui_core_views/pivot_ui.ts
```diff
@@ -65,9 +66,7 @@ export class PivotUIPlugin extends UIPlugin {
   handle(cmd: Command) {
     if (invalidateEvaluationCommands.has(cmd.type)) {
       for (const pivotId of this.getters.getPivotIds()) {
-        if (!pivotRegistry.get(this.getters.getPivotCoreDefinition(pivotId).type).externalData) {
```
LucasLefevre, line 68: If I understand correctly, there's no bug in o-spreadsheet (all tests in this PR are :heavy_check_mark: in without the diff in the code), but this is this line which prevents from updating odoo pivots ? and also the deepcopy

But then, are the tests introduced/changed here useful ?

LucasLefevre:
Ah non, je voulais pas r+ :sweat_smile: 
I just wanted to ask the question

rrahir:
robodoo r-

@LucasLefevre  I tried a first implementatioon which ended up breaking the evaluation (on undo/redo) and all the tests passed, I caught it by testing this morning in Odoo so I added a bit of coverage.
the test "references are adapted with sheet (https://github.com/odoo/o-spreadsheet/pull/7534/files#diff-75d3b99b661635620faf9ae3eb07a99dcb16898c41607c9b488d900b7aa27d50R1068) was changed because the sheet change impacted both the computed measure AND the pivot dataset, the latter creates an invalidation of the pivot runtime on its own (see `SpreadsheetPivotCorePlugin.adaptRanges`) which is why I tried to test the computed measure dependency only/

LucasLefevre (commented):
robodoo r+

rrahir:
robodoo rebase-ff

PR #7590, opened 2025-12-02
[IMP] export: export align left when the cell content is a number;

fdamhaut:
Task: 5368130

## Description:

Task: [5368130](https://www.odoo.com/odoo/2328/tasks/5368130)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/xlsx/functions/styles.ts
+++ b/src/xlsx/functions/styles.ts
@@ -154,6 +154,9 @@ export function addStyles(styles: XLSXStyle[]): XMLString {
     if (style.alignment && style.alignment.wrapText) {
       alignAttrs.push(["wrapText", "1"]);
     }
+    if (style.alignment && style.alignment.shrinkToFit) {
+      alignAttrs.push(["shrinkToFit", "1"]);
+    }
 
     if (alignAttrs.length > 0) {
       attributes.push(["applyAlignment", "1"]); // for Libre Office
--- a/src/xlsx/helpers/content_helpers.ts
+++ b/src/xlsx/helpers/content_helpers.ts
@@ -110,6 +110,7 @@ export function extractStyle(cell: ExcelCellData, data: WorkbookData): Extracted
         ? V_ALIGNMENT_EXPORT_CONVERSION_MAP[style.verticalAlign]
         : undefined,
       wrapText: style.wrapping === "wrap" || cell.content?.includes(NEWLINE) ? true : undefined,
+      shrinkToFit: style.wrapping === "clip" ? true : undefined,
     },
   };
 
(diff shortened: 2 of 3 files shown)


src/plugins/core/cell.ts
```diff
@@ -295,10 +295,13 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
     this.export(data);
   }
 
-  private removeDefaultStyleValues(style: Style | undefined): Style {
-    const cleanedStyle = { ...style };
-    for (const property in DEFAULT_STYLE) {
-      if (cleanedStyle[property] === DEFAULT_STYLE[property]) {
+  private extractCustomStyle(cell: Cell): Style {
+    const cleanedStyle = { ...cell.style };
+    const defaultStyle = isNumber(cell.content, this.getters.getLocale())
```
hokolomopo: i don't think that works for formulas, if the cell is a formula evaluated to a number, we should still export `align: left` style.

anhe-odoo: Up ?

fdamhaut: I now also always export formula alignment as it may switch based on the sheet usage

src/plugins/core/cell.ts
```diff
@@ -295,10 +295,14 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
...
```
LucasLefevre: in core plugins, `cell.content` is always normalized using en_US locale

src/plugins/core/cell.ts
```diff
@@ -295,10 +295,14 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
...
+      ? DEFAULT_NUMBER_STYLE
+      : DEFAULT_STYLE;
+    for (const property in defaultStyle) {
```
LucasLefevre: like you did in your other PR, iterate over the style which contains less properties (in average)
```suggestion
    for (const property in cell.style) {
```

src/plugins/core/cell.ts
```diff
@@ -295,10 +295,14 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
...
+      if (property === "align" && cell.isFormula) continue;
+      if (cleanedStyle[property] === defaultStyle[property]) {
```
LucasLefevre: one line `if` 😠 
```suggestion
      if (cleanedStyle[property] === defaultStyle[property] && !(property === "align" && cell.isFormula)) {
```

PR #7592, opened 2025-12-02
[Data filter] Trim whitespace

matho-odoo:
## Description:

[Data filter] Trim whitespace and don't bother of caps vs low case when filtering

Task: [5375214](https://www.odoo.com/odoo/2328/tasks/5375214)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/packages/o-spreadsheet-engine/src/helpers/text_helper.ts
+++ b/packages/o-spreadsheet-engine/src/helpers/text_helper.ts
@@ -316,9 +316,9 @@ export function getFontSizeMatchingWidth(
   return fontSize;
 }
 
-/** Transform a string to lower case. If the string is undefined, return an empty string */
-export function toLowerCase(str: string | undefined): string {
-  return str ? str.toLowerCase() : "";
+/** Transform a string to lowercase and removes whitespace from both ends of the string*/
+export function toTrimmedLowerCase(str: string | undefined): string {
+  return str ? str.toLowerCase().trim() : "";
 }
 
 /**
--- a/packages/o-spreadsheet-engine/src/plugins/ui_stateful/filter_evaluation.ts
+++ b/packages/o-spreadsheet-engine/src/plugins/ui_stateful/filter_evaluation.ts
@@ -2,7 +2,7 @@ import { isMultipleElementMatrix, toScalar } from "../../functions/helper_matric
 import { parseLiteral } from "../../helpers/cells/cell_evaluation";
 import { toXC } from "../../helpers/coordinates";
 import { deepCopy, getUniqueText, range } from "../../helpers/misc";
-import { toLowerCase } from "../../helpers/text_helper";
+import { toTrimmedLowerCase } from "../../helpers/text_helper";
 import { positions, toZone, zoneToDimension } from "../../helpers/zones";
 import { criterionEvaluatorRegistry } from "../../registries/criterion_registry";
 import { Command, CommandResult, LocalCommand, UpdateFilterCommand } from "../../types/commands";
@@ -176,7 +176,7 @@ export class FilterEvaluationPlugin extends UIPlugin {
         continue;
       }
       if (filterValue.filterType === "values") {
-        const filteredValues = filterValue.hiddenValues?.map(toLowerCase);
+        const filteredValues = filterValue.hiddenValues?.map(toTrimmedLowerCase);
         if (!filteredValues) {
           continue;
         }
@@ -227,7 +227,7 @@ export class FilterEvaluationPlugin extends UIPlugin {
 
   private getCellValueAsString(sheetId: UID, col: number, row: number): string {
     const value = this.getters.getEvaluatedCell({ sheetId, col, row }).formattedValue;
-    return value.toLowerCase();
+    return toTrimmedLowerCase(value);
   }
 
   exportForExcel(data: ExcelWorkbookData) {
--- a/src/components/filters/filter_menu_value_list/filter_menu_value_list.ts
+++ b/src/components/filters/filter_menu_value_list/filter_menu_value_list.ts
@@ -1,6 +1,6 @@
 import { SpreadsheetChildEnv } from "@odoo/o-spreadsheet-engine/types/spreadsheet_env";
 import { Component, onWillUpdateProps, useRef, useState } from "@odoo/owl";
-import { deepEquals, positions, toLowerCase } from "../../../helpers";
+import { deepEquals, positions, toTrimmedLowerCase } from "../../../helpers";
 import { fuzzyLookup } from "../../../helpers/search";
 import { Position } from "../../../types";
 import { FilterMenuValueItem } from "../filter_menu_item/filter_menu_value_item";
@@ -76,12 +76,12 @@ export class FilterMenuValueList extends Component<Props, SpreadsheetChildEnv> {
 
     const cellValues = cells.map((val) => val.cellValue);
     const filterValues = filterValue?.filterType === "values" ? filterValue.hiddenValues : [];
-    const normalizedFilteredValues = new Set(filterValues.map(toLowerCase));
+    const normalizedFilteredValues = new Set(filterValues.map(toTrimmedLowerCase));
 
     const set = new Set<string>();
     const values: (Value & { normalizedValue: string })[] = [];
     const addValue = (value: string) => {
-      const normalizedValue = toLowerCase(value);
+      const normalizedValue = toTrimmedLowerCase(value);
       if (!set.has(normalizedValue)) {
         values.push({
           string: value || "",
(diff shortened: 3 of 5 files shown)


src/components/filters/filter_menu_value_list/filter_menu_value_list.ts
```diff
@@ -76,12 +76,12 @@ export class FilterMenuValueList extends Component<Props, SpreadsheetChildEnv> {
 
     const cellValues = cells.map((val) => val.cellValue);
     const filterValues = filterValue?.filterType === "values" ? filterValue.hiddenValues : [];
-    const normalizedFilteredValues = new Set(filterValues.map(toLowerCase));
+    const normalizedFilteredValues = new Set(filterValues.map(normalize));
```
anhe-odoo: Not sure about this, but maybe we can store the normalized values in `filterValue.hiddenValues` instead of applying the `normalize` function each time we reevaluate a filter ?

LucasLefevre: Agreed.
The values should be normalized in `filter_evaluation.ts` plugin when the values are stored in the plugin.
That way, you don't have to think (and forget) about normalizing every time you use `getFilterValue`

tests/table/filter_evaluation_plugin.test.ts
```diff
@@ -352,6 +361,30 @@ describe("Filter criterion test", () => {
...
+        A5: "A",
+      };
+      setGrid(model, grid);
+      createTableWithFilter(model, "A1:A5");
+      updateFilterCriterion(model, "A1", {
+        type: type as FilterCriterionType,
+        values: criterionValues,
+      });
+
+      expect(getFilteredRows()).toEqual(expectedFilteredRows);
```
LucasLefevre, line 386: if `["a"]` and `[]` are always the same for each test case, then they shouldn't be part of the `each`.

Also, you can use `as const` to avoid the type casting 😉 
```suggestion
test.each(["beginsWithText", "endsWithText", "isEqualText"] as const)(
    "Can filter based on a text criterion %s ignoring lowercase/uppercase and whitespaces",
    (type) => {
      const grid = {
        A2: "a",
        A3: " a",
        A4: "a ",
        A5: "A",
      };
      setGrid(model, grid);
      createTableWithFilter(model, "A1:A5");
      updateFilterCriterion(model, "A1", {
        type,
        values: ["a"],
      });

      expect(getFilteredRows()).toEqual([]);
```

packages/o-spreadsheet-engine/src/helpers/text_helper.ts
LucasLefevre: This is no longer used and can be removed.

packages/o-spreadsheet-engine/src/helpers/text_helper.ts
```diff
@@ -297,6 +297,11 @@ export function toLowerCase(str: string | undefined): string {
   return str ? str.toLowerCase() : "";
 }
 
+/** Transform a string to lowercase and removes whitespace from both ends of the string*/
+export function normalize(value: string): string {
```
LucasLefevre: The word "normalize" is very generic and doesn't mean much (or rather, it depends a lot on the context). We already have `normalizeValue`, `normalizeString`

What about naming this very very clearly ? even if it's a bit longer 
```suggestion
export function toTrimmedLowerCase(value: string): string {
```

tests/table/filter_evaluation_plugin.test.ts
```diff
@@ -352,6 +361,30 @@ describe("Filter criterion test", () => {
...
+    (type: string, criterionValues: string[], expectedFilteredRows: number[]) => {
+      const grid = {
+        A2: "a",
+        A3: " a",
+        A4: "a ",
+        A5: "A",
+      };
+      setGrid(model, grid);
+      createTableWithFilter(model, "A1:A5");
+      updateFilterCriterion(model, "A1", {
```
LucasLefevre, line 381: this test passes even without this `updateFilterCriterion`, so I'm not sure what's the point

packages/o-spreadsheet-engine/src/plugins/ui_stateful/filter_evaluation.ts
```diff
@@ -139,6 +139,11 @@ export class FilterEvaluationPlugin extends UIPlugin {
     const id = this.getters.getFilterId({ sheetId, col, row });
     if (!id) return;
     if (!this.filterValues[sheetId]) this.filterValues[sheetId] = {};
+    if (value.filterType === "values") {
+      value.hiddenValues = value.hiddenValues.map(toTrimmedLowerCase);
+    } else if (value.filterType === "criterion") {
+      value.values = value.values.map(toTrimmedLowerCase);
```
hokolomopo: criterion values can be formulas (eg. `=A1`), in which case lowercasing the formula is useless. You should lowercase the evaluated formula value instead (`evaluatedCriterionValues`)

packages/o-spreadsheet-engine/src/registries/criterion_registry.ts
```diff
@@ -663,8 +664,8 @@ criterionEvaluatorRegistry.add("customFormula", {
 criterionEvaluatorRegistry.add("beginsWithText", {
   type: "beginsWithText",
   isValueValid: (value: CellValue, criterion: EvaluatedCriterion) => {
-    const strValue = String(value);
-    return strValue.toLowerCase().startsWith(String(criterion.values[0]).toLowerCase());
+    const strValue = toTrimmedLowerCase(String(value));
+    return strValue.startsWith(String(criterion.values[0]));
```
hokolomopo: Don't change those, they are also used for conditional formats/data validation. And for conditional format, we don't want to trim the values as far as I know (it doesn't seems to be the case in excel at least). Just trim the value in the filter evaluation plugin before passing it here.

tests/table/filter_evaluation_plugin.test.ts
```diff
@@ -161,6 +161,15 @@ describe("Filter Evaluation", () => {
...
+  test("Filters ignore whitespaces", () => {
+    setCellContent(model, "A2", "a");
+    setCellContent(model, "A3", " a");
+    setCellContent(model, "A4", "a ");
+    updateFilter(model, "A2", ["a"]);
+    expect(model.getters.isRowHidden(sheetId, 1)).toEqual(true);
+    expect(model.getters.isRowHidden(sheetId, 2)).toEqual(true);
+    expect(model.getters.isRowHidden(sheetId, 3)).toEqual(true);
+  });
   test("Header is not filtered", () => {
```
hokolomopo, line 174:
```suggestion
  });
  
  test("Header is not filtered", () => {
```
nitpick

packages/o-spreadsheet-engine/src/plugins/ui_stateful/filter_evaluation.ts
```diff
@@ -178,7 +181,7 @@ export class FilterEvaluationPlugin extends UIPlugin {
 
         const evaluatedCriterionValues = filterValue.values.map((value) => {
           if (!value.startsWith("=")) {
-            return parseLiteral(value, DEFAULT_LOCALE);
+            return parseLiteral(toTrimmedLowerCase(value), DEFAULT_LOCALE);
           }
           return this.getters.evaluateFormula(sheetId, value) ?? "";
```
hokolomopo, line 205: we should also trimLowercase the results of the evaluated formulas, no ?

packages/o-spreadsheet-engine/src/registries/criterion_registry.ts
```diff
@@ -128,8 +116,8 @@ criterionEvaluatorRegistry.add("notContainsText", {
 criterionEvaluatorRegistry.add("isEqualText", {
   type: "isEqualText",
   isValueValid: (value: CellValue, criterion: EvaluatedCriterion) => {
-    const strValue = String(value);
-    return strValue.toLowerCase() === String(criterion.values[0]).toLowerCase();
+    const strValue = toTrimmedLowerCase(String(value));
```
LucasLefevre: why do you `toTrimmedLowerCase` only for those 2 types of criterion and many others are still using `toLowerCase`?

`value` is already trimmed at this point, isn't it?

hokolomopo: because CFs/DVs also use these criterions, but in CFs/DVs the values shouldn't be trimmed, the spaces at the start matter (at least it's excel behavior). 

This file shouldn't be changed at all, i'm pretty sure you left a change by mistake @matho-odoo :slightly_smiling_face:  

(also the PR need a rebase :upside_down_face: )

PR #7595, opened 2025-12-02
[IMP] formats: Set format when writing date on number format

fdamhaut:
Task: 5349524

## Description:

Task: [5349524](https://www.odoo.com/odoo/2328/tasks/5349524)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/composer/composer/cell_composer_store.ts
+++ b/src/components/composer/composer/cell_composer_store.ts
@@ -9,6 +9,7 @@ import { getDateTimeFormat, localizeFormula } from "@odoo/o-spreadsheet-engine/h
 import { criterionEvaluatorRegistry } from "@odoo/o-spreadsheet-engine/registries/criterion_registry";
 import { _t } from "@odoo/o-spreadsheet-engine/translation";
 import {
+  detectDateFormat,
   formatValue,
   isDateTimeFormat,
   isFormula,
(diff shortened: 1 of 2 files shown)


packages/o-spreadsheet-engine/src/plugins/core/cell.ts
```diff
@@ -510,8 +511,18 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
 
     // Compute the new cell properties
     const afterContent = hasContent ? replaceNewLines(after?.content) : before?.content || "";
-    const format = "format" in after ? after.format : before && before.format;
 
+    let format: Format | undefined = undefined;
+    if (after.format !== undefined) {
```
rrahir: this breaks the "paste as value" behaviour. a user will not be able to just paste the datevalue and will have tomanually set an automatic style
+ it will alter existing sheets when replaying commands/ have unexpected effects upon concurrent revisions
This kind of non-explicit side effects should be avoided, it'd be beter to implement that logic in the composer itself imho

LucasLefevre: I agree with Rémi, I'd do it in the composer.
Mainly to keep UPDATE_CELL as simple as possible since it's one of the base foundation of everything.

Imagine we go with this, and want to change it later: we can't

And because it alter the behavior of existing commands, and could break things in subtle ways

packages/o-spreadsheet-engine/src/plugins/core/cell.ts
```diff
@@ -510,7 +510,7 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
 
     // Compute the new cell properties
     const afterContent = hasContent ? replaceNewLines(after?.content) : before?.content || "";
-    const format = "format" in after ? after.format : before && before.format;
+    const format = after.format !== undefined ? after.format : before && before.format;
```
rrahir, line 576: this means the behaviour is broken in stable. indeed
- Alice adds a % format in A2
- Alice copies a numerical value from another cell that has no format
- pastes it *as value* to A2
Alice sees no % format
Bob sees a %format


the fix is correct but should be backported in all stable versions as the states of the users diverge ATM

rrahir: https://www.odoo.com/odoo/2328/tasks/5499921

rrahir (changes requested):
we could either put this task on hold and wait for https://www.odoo.com/odoo/2328/tasks/5499921 or add a test  for the change of behaviour of the formats, like it's done here 
https://github.com/odoo/o-spreadsheet/pull/7795/files#diff-df7329cc5bd8320c95399e95a2bdad414692333c6f12ad95cdf21251ea517422R1185-R1193

I'd rather put the task on hold but really don"t mind either way

rrahir:
robodoo r+

PR #7603, opened 2025-12-03
[FIX] CorePlugins: Prevent dispatch during adaptRange

rrahir:
The method `adaptRange` was originally designed for each plugin to update their own ranges, whithout having to manually react to the different commands that alter the sheets structure.

We have abused the system when we introduced the spreadsheet pivots to allow the modification of a pivot dataset (which is stored in PivotCorePlugin) from another entry point `SpreadsheetPivotCorePlugin`) by dispatching an UPDATE_PIVOT command but there are some side effects that we did not account for, the dispatched command is forwarded to EVERY other plugin. Meaning that during an adaptRange, which should only concern a specific plugin, we end up notifying a change to the whole core/coreview stack.

This revision removes the access to 'dispatch' during the handling of `adaptRange` so we ensure that the changes are kept locally.

Task: 5380747

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/pivot/pivot_registry.ts
+++ b/src/helpers/pivot/pivot_registry.ts
@@ -1,6 +1,6 @@
 import { ModelConfig } from "../../model";
 import { Registry } from "../../registries/registry";
-import { CoreGetters, Getters } from "../../types";
+import { ApplyRangeChange, CoreGetters, Getters, Range } from "../../types";
 import { PivotCoreDefinition, PivotField, PivotFields } from "../../types/pivot";
 import { Pivot } from "../../types/pivot_runtime";
 import { PivotRuntimeDefinition } from "./pivot_runtime_definition";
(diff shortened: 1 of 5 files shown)


src/plugins/core_plugin.ts
```diff
@@ -51,7 +51,14 @@ export class CorePlugin<State = any>
...
   }: CorePluginConfig) {
     super(stateObserver, dispatch, canDispatch);
-    range.addRangeProvider(this.adaptRanges.bind(this));
+
+    // Bind adaptRanges to a version of `this` where `dispatch` always throws
+    const thisWithThrowingDispatch = Object.create(this);
+    thisWithThrowingDispatch.dispatch = () => {
+      throw new Error("dispatch is not allowed in adaptRanges context");
+    };
+    range.addRangeProvider(this.adaptRanges.bind(thisWithThrowingDispatch));
```
hokolomopo: Add a test for this 🙂 

And maybe it should be the responsibility of the range plugin, rather than every other plugin ? With a boolean `isAdaptingRange` or something that would throw on a command `handle`  ... not sure tho.

hokolomopo (commented):
👋

src/plugins/core_plugin.ts
rrahir: > And maybe it should be the responsibility of the range plugin

I thought about it as well, I liked that implementation a bit but the shallow copy might not be the brightest idea. Maybe others will have a strong opinion about this?

LucasLefevre:
oups, I didn't see the second commit.

I'm ok with the second implementation

rrahir:
@LucasLefevre  it's updated :)

LucasLefevre (commented):
robodoo r+

PR #7605, opened 2025-12-04
[IMP] spreadsheet: add named ranges feature

hokolomopo:
## Description:

This commits adds the named ranges feature, that allows to defined a
custom name for a range of cells, and use that name in formulas.

Task: [5380498](https://www.odoo.com/odoo/2328/tasks/5380498)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/components/composer/composer/abstract_composer_store.ts
```diff
@@ -94,6 +99,7 @@ export abstract class AbstractComposerStore extends SpreadsheetStore {
   protected initialContent: string | undefined = "";
   private colorIndexByRange: { [xc: string]: number } = {};
   private autoComplete: Store<AutoCompleteStore> = new AutoCompleteStore(this.get);
+  private hasSelectedAProposal: boolean = false;
```
hokolomopo, line 99: This kinda sucks ... but I don't have a better idea.

The problem is if we open a composer on a formula like `=MyNamedRange`. The autocomplete dropdown is open, and `MyNamedRange` is selected. If we press enter, `provider.selectProposal` will be called, do nothing, the dropdown will be closed ... and the promptly re-opened since nothing changed, so pressing enter will do absolutely nothing and we cannot close the composer.

And we don't want to hide the dropdown if `this._currentContent === this.initialContent` otherwise we don't have functions auto-complete either ...

packages/o-spreadsheet-engine/src/collaborative/ot/ot_specific.ts
```diff
@@ -107,6 +109,12 @@ otRegistry.addTransformation(
   pivotZoneTransformation
 );
 
+otRegistry.addTransformation(
+  "UPDATE_NAMED_RANGE",
+  ["UPDATE_NAMED_RANGE", "DELETE_NAMED_RANGE"],
```
LucasLefevre, line 114: also add a transformation to transform a UPDATE_NAMED_RANGE when it's been deleted by a DELETE_NAMED_RANGE ?

packages/o-spreadsheet-engine/src/plugins/core/named_range.ts
```diff
@@ -0,0 +1,209 @@
...
+      });
+    }
+  }
+
+  export(data: WorkbookData) {
+    data.namedRanges = [];
+    for (const namedRange of this.namedRanges) {
+      data.namedRanges.push({
+        rangeName: namedRange.rangeName,
+        rangeString: this.getters.getRangeString(namedRange.range, "forceSheetReference"),
```
LucasLefevre: would it feel more natural and shorter with this?

```json
"namedRanges": {
  "name1": "Sheet1!A1",
  "name2": "Sheet1!A2",
}
```

packages/o-spreadsheet-engine/src/types/misc.ts
```diff
@@ -419,3 +422,8 @@ export interface ValueAndLabel<T = string> {
   value: T;
   label: string;
 }
+
+export interface NamedRange {
+  rangeName: string;
```
LucasLefevre: `rangeName` => `name` ?

packages/o-spreadsheet-engine/src/plugins/core/named_range.ts
```diff
@@ -0,0 +1,209 @@
...
+interface NamedRangeState {
+  readonly namedRanges: Array<NamedRange>;
+}
+
+const invalidNamedRangeCharacterRegex = /[^a-zA-Z0-9_.]/;
+
+export class NamedRangesPlugin extends CorePlugin<NamedRangeState> implements NamedRangeState {
+  static getters = ["getNamedRange", "getNamedRangeFromZone", "getNamedRanges"] as const;
+
+  readonly namedRanges: Array<NamedRange> = [];
```
LucasLefevre, line 22: wouldn't it be easier to store this as an object `{  [lowerCaseName]: range }`, instead of playing with indexes ? (and many array copies)

Lookup would also be O(1) instead of O(n)

packages/o-spreadsheet-engine/src/plugins/core/named_range.ts
```diff
@@ -0,0 +1,209 @@
...
+    if (this.getNamedRange(name)) {
+      return CommandResult.NamedRangeNameAlreadyExists;
+    }
+
+    if (invalidNamedRangeCharacterRegex.test(name) || isNumber(name, DEFAULT_LOCALE)) {
+      return CommandResult.NamedRangeNameWithInvalidCharacter;
+    }
+
+    if (rangeReference.test(name)) {
+      return CommandResult.NamedRangeNameLooksLikeCellReference;
```
LucasLefevre, line 144: TODO check if it still does what we want with `A1#` notation when this is rebased

packages/o-spreadsheet-engine/src/plugins/ui_core_views/cell_evaluation/evaluator.ts
```diff
@@ -560,13 +567,38 @@ export class Evaluator {
    * and error handling.
    */
   private buildSafeGetSymbolValue(getContextualSymbolValue?: GetSymbolValue): GetSymbolValue {
-    const getSymbolValue = (symbolName: string) => {
+    const getSymbolValue = (
```
LucasLefevre: could all this `getSymbolValue` thing be moved to `CompilationParametersBuilder` ?

It wouldn't be necessary to move around `refFn`, `range` and `ctx`.

packages/o-spreadsheet-engine/src/types/workbook_data.ts
```diff
@@ -115,6 +117,10 @@ export interface DataValidationRuleData extends Omit<DataValidationRule, "ranges
   ranges: string[];
 }
 
+export interface NamedRangeData extends Omit<NamedRange, "range"> {
```
LucasLefevre:
```suggestion
interface NamedRangeData extends Omit<NamedRange, "range"> {
```

src/components/helpers/listener_hook.ts
```diff
@@ -26,5 +26,18 @@ export function useHoveredElement(ref: Ref<HTMLElement>) {
...
+  // but it's better than having a stuck hover state.
+  const resizeObserver = new ResizeObserver(() => {
+    state.hovered = false;
+  });
+  onMounted(() => {
+    resizeObserver.observe(ref.el!);
+  });
+  onWillUnmount(() => {
+    resizeObserver.disconnect();
+  });
```
LucasLefevre: it should be a `useEffect` (similar to `useRefListener`) because `ref.el` can change in the component lifetime .

demo/data.js
```diff
@@ -3898,6 +3898,12 @@ export const demoData = {
   },
   pivotNextId: 2,
   customTableStyles: {},
+  namedRanges: [
+    {
+      rangeName: "MyNamedRange",
+      rangeString: "sh1!C4:D10",
```
LucasLefevre: `sh1` => `Sheet1`

src/components/side_panel/named_ranges_panel/named_ranges_panel.ts
```diff
@@ -0,0 +1,38 @@
...
+import { Component, useState } from "@odoo/owl";
+import { SelectionInput } from "../../selection_input/selection_input";
+import { TextInput } from "../../text_input/text_input";
+import { NamedRangePreview } from "./named_range_preview/named_range_preview";
+
+interface Props {
+  onCloseSidePanel: () => void;
+}
+
+interface State {}
```
LucasLefevre: best state ever 😄

src/registries/auto_completes/function_auto_complete.ts
```diff
@@ -2,9 +2,11 @@ import { COMPOSER_ASSISTANT_COLOR } from "@odoo/o-spreadsheet-engine/constants";
 import { functionRegistry } from "@odoo/o-spreadsheet-engine/functions/function_registry";
 import { getHtmlContentFromPattern } from "../../components/helpers/html_content_helpers";
 import { isFormula } from "../../helpers";
-import { autoCompleteProviders } from "./auto_complete_registry";
+import { AutoCompleteProposal, autoCompleteProviders } from "./auto_complete_registry";
 
-autoCompleteProviders.add("functions", {
+type FunctionAutoCompleteProposal = AutoCompleteProposal & { type: "function" | "named_range" };
+
+autoCompleteProviders.add("functions_and_named_ranges", {
```
LucasLefevre, line 9: we might want to change the behavior for all autocomplete if we want to start mixing several sources, do we want to mix all of them ?

packages/o-spreadsheet-engine/src/collaborative/ot/ot_specific.ts
hokolomopo: That's handled by the allowDispatch no ?

packages/o-spreadsheet-engine/src/plugins/core/named_range.ts
hokolomopo: I did that at first, but then renaming a named range would change the order of the ranges and they would jump around in the side panel. And we'd need to store `{  [lowerCaseName]: {range: Range, actualNameWithUpperCase] }`. 

But maybe the side panel should be more complex & smart to handle that, and the plugin should still use an object ? Idk I think I prefer an array, there shouldn't be enough ranges in a spreadsheet for the lookup performance to really matter.

src/registries/auto_completes/function_auto_complete.ts
hokolomopo: Yeah I wondered about that, but I don't think so, I don't think we want to function/named range autocomplete at the same time as the pivot id autocomplete for example ...

packages/o-spreadsheet-engine/src/collaborative/ot/ot_specific.ts
LucasLefevre: yes

packages/o-spreadsheet-engine/src/plugins/core/named_range.ts
hokolomopo: Rejects the command because I don't accept # as a valid character in named range name, so we never even get there

packages/o-spreadsheet-engine/src/plugins/core/named_range.ts
```diff
@@ -0,0 +1,210 @@
...
+    return this.namedRanges.findIndex((r) => r && r.name.toLowerCase() === name.toLowerCase());
+  }
+
+  private renameNamedRangeInFormulas(oldName: string, newName: string) {
+    const lowerCaseOldName = oldName.toLowerCase();
+    const isOldNamedRangeToken = (token: Token, nextToken: Token | undefined) => {
+      return (
+        token.type === "SYMBOL" &&
+        token.value.toLowerCase() === lowerCaseOldName &&
+        !isFuncallToken(token, nextToken)
```
LucasLefevre: this doesn't work if there's a space between the function name and the parenthesis

packages/o-spreadsheet-engine/src/plugins/core/named_range.ts
```diff
@@ -0,0 +1,210 @@
...
+      }
+      case "UPDATE_NAMED_RANGE": {
+        const index = this.getNamedRangeIndex(cmd.oldRangeName);
+        if (index !== -1) {
+          const range = this.getters.getRangeFromRangeData(cmd.ranges[0]);
+          const newNamedRanges = [...this.namedRanges];
+          newNamedRanges[index] = { name: cmd.newRangeName, range };
+          this.history.update("namedRanges", newNamedRanges);
+          if (cmd.oldRangeName !== cmd.newRangeName) {
+            this.renameNamedRangeInFormulas(cmd.oldRangeName, cmd.newRangeName);
```
LucasLefevre: actually, I'm wondering if it shouldn't be a new `name` property on `Range`.

At least, it should be part of the range plugin `adaptRanges`.

With the current approach, named ranges are only renamed in cells, but they are not renamed in CF formulas,, DV formulas, calculated measures, etc.

Note that https://github.com/odoo/o-spreadsheet/pull/7751 introduces more flexibility/convenience in the way plugins can adapt their ranges. It'll probably help (with `adaptFormulaString` and we could introduce `adaptCompiledFormula` if needed)

packages/o-spreadsheet-engine/src/plugins/core/named_range.ts
hokolomopo: Oh yeah I copied the parser, but the parser filters out spaces ....

src/components/composer/composer/composer.ts
```diff
@@ -848,11 +849,11 @@ export class Composer extends Component<CellComposerProps, SpreadsheetChildEnv>
     return [...new Set(argsToFocus)];
   }
 
-  autoComplete(value: string) {
-    if (!value || (this.assistant.forcedClosed && this.props.composerStore.canBeToggled)) {
+  autoComplete(proposal: AutoCompleteProposal) {
+    if (!proposal || (this.assistant.forcedClosed && this.props.composerStore.canBeToggled)) {
```
anhe-odoo: Not sure about it, but as we checked `value` before, shouldn't we check `proposal.text` here ?

packages/o-spreadsheet-engine/src/plugins/core/pivot.ts
```diff
@@ -171,20 +183,30 @@ export class PivotCorePlugin extends CorePlugin<CoreState> implements CoreState
...
-              "compiledMeasureFormulas",
-              pivotId,
-              measureId,
-              "formula",
-              "rangeDependencies",
-              i,
-              updatedRange.range
-            );
-          }
+        // for (let i = 0; i < compiledFormula.rangeDependencies.length; i++) {
```
anhe-odoo: Forgotten comments ?

packages/o-spreadsheet-engine/src/plugins/core/range.ts
```diff
@@ -71,9 +73,46 @@ export class RangeAdapterPlugin implements CommandHandler<CoreCommand> {
...
-    const rangeAdapter = getRangeAdapter(cmd);
-    if (rangeAdapter) {
-      this.executeOnAllRanges(rangeAdapter);
+
+    let rangeAdapter: RangeAdapter | undefined;
+    let applyRenameNamedRange: ApplyRenameNamedRange | undefined;
+
+    switch (cmd.type) {
+      case "UPDATE_NAMED_RANGE":
+        // if (cmd.newRangeName !== cmd.oldRangeName) {
```
anhe-odoo: ?

src/components/named_range_selector/named_range_selector.ts
```diff
@@ -0,0 +1,184 @@
...
+    this.topBarToolStore = useToolBarDropdownStore();
+    this.DOMFocusableElementStore = useStore(DOMFocusableElementStore);
+  }
+
+  onInput(value: string) {
+    this.menuState.searchedText = value || undefined;
+    this.menuState.menuItems = this.getNamedRangeMenuItems();
+  }
+
+  changeInputValue(newValue: string) {
```
anhe-odoo, line 48: In the plugin, you test the validity of the new name. Is there any place where you display the error if we have an invalid name ? Maybe we should check it here directly, in the same way we check it for the selection_input ?

anhe-odoo: Nevermind, you have it in the named_range_interactive :)

tests/range_plugin.test.ts
```diff
@@ -531,6 +531,13 @@ describe("range plugin", () => {
       expect(m.getters.getRangeString(r, "s1")).toBe("NOTTHERE!A1");
     });
 
+    test("can create a range without giving a defaultSheetId if a sheet name is in the XC", () => {
```
anhe-odoo: "can create a range without giving a defaultSheetId **ONLY** if a sheet name is in the XC" ?

packages/o-spreadsheet-engine/src/formulas/compiler.ts
```diff
@@ -174,6 +195,64 @@ export class CompiledFormula implements Omit<ICompiledFormula, "tokens" | "depen
...
+    const newDependencies: Range[] = [];
+    let hasChanges = false;
+    for (const range of this.rangeDependencies) {
+      const change = applyChange(range);
+      if (change.changeType === "NONE") {
+        newDependencies.push(range);
+      } else {
+        hasChanges = true;
+        newDependencies.push(change.range);
+      }
```
LucasLefevre, line 209: can be simplified
```ts
      const change = applyChange(range);
      newDependencies.push(change.range);
      if (change.changeType !== "NONE") {
        hasChanges = true;
      }
```

packages/o-spreadsheet-engine/src/formulas/compiler.ts
```diff
@@ -174,6 +195,64 @@ export class CompiledFormula implements Omit<ICompiledFormula, "tokens" | "depen
...
+    }
+
+    const compiledFormula = hasChanges
+      ? CompiledFormula.CopyWithDependencies(this, this.sheetId, newDependencies)
+      : this;
+
+    return compiledFormula.renameNamedRangeTokens(applyUpdateNamedRange);
```
LucasLefevre: we uselessly create two `CompiledFormula` here. Can you try to optimize it ?

packages/o-spreadsheet-engine/src/helpers/formulas.ts
```diff
@@ -1,12 +1,24 @@
...
 import { CellErrorType } from "../types/errors";
-import { ApplyRangeChangeResult, RangeAdapter, UID } from "../types/misc";
+import { ApplyRangeChangeResult, ApplyRenameNamedRange, RangeAdapter, UID } from "../types/misc";
 import { Range } from "../types/range";
 import { concat } from "./misc";
 import { createInvalidRange, createRangeFromXc, getRangeString } from "./range";
 import { rangeReference, splitReference } from "./references";
 import { isSheetNameEqual } from "./sheet";
 
+export function adaptFormulaString(
```
LucasLefevre, line 12: All uses of `adaptFormulaStringRanges` in `srt_specific.ts` should use this, isn't it?

-> `adaptFormulaStringRanges` should never be used anywhere in the business code

src/helpers/figures/charts/gauge_chart.ts
```diff
@@ -182,6 +182,8 @@ export class GaugeChart extends AbstractChart {
         dataRange = adaptedRange;
       }
     }
+
+    // ADRM TODO: use range adapter ??
```
LucasLefevre: yes, but I'm breaking everything in my refactoring. It is one of the next planned step

packages/o-spreadsheet-engine/src/helpers/formulas.ts
```diff
@@ -33,6 +45,44 @@ export function adaptFormulaStringRanges(
...
+    if (tokens[i].type !== "SPACE") {
+      nextNonSpaceToken = tokens[i];
+      break;
+    }
+  }
+
+  return token.type === "SYMBOL" && !isFuncallToken(token, nextNonSpaceToken);
+}
+
+export function adaptFormulaStringNamedRanges(
```
LucasLefevre: remove `export`

packages/o-spreadsheet-engine/src/helpers/formulas.ts
```diff
@@ -1,12 +1,24 @@
...
 import { isSheetNameEqual } from "./sheet";
 
+export function adaptFormulaString(
+  defaultSheetId: string,
+  formula: string,
+  applyChange: RangeAdapter,
+  namedRangeAdapter: ApplyRenameNamedRange
+): string {
+  const adaptedFormulaStringRanges = adaptFormulaStringRanges(defaultSheetId, formula, applyChange);
+  return adaptFormulaStringNamedRanges(adaptedFormulaStringRanges, namedRangeAdapter);
```
LucasLefevre: could you tried to merge things together here instead of tokenizing twice and iterating over all tokens twice?

packages/o-spreadsheet-engine/src/helpers/formulas.ts
```diff
@@ -33,6 +45,44 @@ export function adaptFormulaStringRanges(
...
+  const newTokens: Token[] = [];
+  for (let i = 0; i < tokens.length; i++) {
+    let newToken = tokens[i];
+    if (canBeNamedRangeToken(tokens, i)) {
+      const newName = namedRangeAdapter(tokens[i].value);
+      if (newName !== tokens[i].value) {
+        newToken = { ...tokens[i], value: newName };
+      }
+    }
+    newTokens.push(newToken);
```
LucasLefevre: maybe use the same strategy as `adaptFormulaStringRanges` to avoid creating yet another array of tokens

packages/o-spreadsheet-engine/src/plugins/core/range.ts
```diff
@@ -71,9 +73,44 @@ export class RangeAdapterPlugin implements CommandHandler<CoreCommand> {
...
+
+    let rangeAdapter: RangeAdapter | undefined;
+    let applyRenameNamedRange: ApplyRenameNamedRange | undefined;
+
+    switch (cmd.type) {
+      case "UPDATE_NAMED_RANGE":
+        rangeAdapter = {
+          applyChange: (range) => ({ changeType: "NONE", range }),
+          sheetId: cmd.ranges[0]._sheetId,
+          sheetName: { old: "ADRM TODO", current: "ADRM TODO" },
```
LucasLefevre: "ADRM TODO" 👀
idea: instead of faking a rangeAdapter in a ~hackish way, you could instead build a fully different adapterFunctions ?
with applyChange and adaptStringRange being identity functions and adaptFormulaString and adaptCompiledFormula being dedicated functions for named ranges (and also dedicated to only ranges in the default case)

It would also simplify a bunch of other helper functions where you trimbale both a `RangeAdapter` and a `ApplyRenameNamedRange` where in really only one of the two do real work.

what do you think?

packages/o-spreadsheet-engine/src/plugins/core/range.ts
```diff
@@ -71,9 +73,44 @@ export class RangeAdapterPlugin implements CommandHandler<CoreCommand> {
...
-    if (rangeAdapter) {
-      this.executeOnAllRanges(rangeAdapter);
+
+    let rangeAdapter: RangeAdapter | undefined;
+    let applyRenameNamedRange: ApplyRenameNamedRange | undefined;
+
+    switch (cmd.type) {
+      case "UPDATE_NAMED_RANGE":
+        rangeAdapter = {
+          applyChange: (range) => ({ changeType: "NONE", range }),
```
LucasLefevre: the same thing needs to be done for `srt_registry.ts`

packages/o-spreadsheet-engine/src/plugins/core/named_range.ts
```diff
@@ -0,0 +1,158 @@
...
+        this.history.update("namedRanges", newNamedRanges);
+        break;
+      }
+      case "UPDATE_NAMED_RANGE": {
+        const index = this.getNamedRangeIndex(cmd.oldRangeName);
+        if (index !== -1) {
+          const range = this.getters.getRangeFromRangeData(cmd.ranges[0]);
+          const newNamedRanges = [...this.namedRanges];
+          newNamedRanges[index] = { name: cmd.newRangeName, range };
+          this.history.update("namedRanges", newNamedRanges);
```
LucasLefevre:
```suggestion
          this.history.update("namedRanges", index, { name: cmd.newRangeName, range });
```

packages/o-spreadsheet-engine/src/plugins/core/named_range.ts
```diff
@@ -0,0 +1,158 @@
...
+        name: name,
+        range: this.getters.getRangeFromSheetXC(undefined, data.namedRanges[name]),
+      });
+    }
+  }
+
+  export(data: WorkbookData) {
+    data.namedRanges = {};
+    for (const namedRange of this.namedRanges) {
+      const xc = this.getters.getRangeString(namedRange.range, "forceSheetReference");
```
LucasLefevre: since you made it optional...
```suggestion
      const xc = this.getters.getRangeString(namedRange.range);
```

src/components/named_range_selector/named_range_selector.ts
```diff
@@ -0,0 +1,184 @@
...
+        (action) => action.name as string
+      );
+    }
+
+    if (actionsSpecs.length > 0) {
+      actionsSpecs.at(-1)!.separator = true;
+    }
+
+    actionsSpecs.push({
+      name: _t("Manage named ranges"),
```
LucasLefevre: should not appear in mobile since side panels are disabled

tests/named_ranges/named_ranges_plugin.test.ts
```diff
@@ -0,0 +1,286 @@
...
+      ["こんにちは", true],
+      ["안녕하세요", true],
+      ["你好", true],
+      ["éàö", true],
+    ])("Valid range names %s", (name, isValid) => {
+      const result = createNamedRange(model, name, "A1");
+      if (isValid) {
+        expect(result).toBeSuccessfullyDispatched();
+      } else {
+        expect(result).not.toBeSuccessfullyDispatched();
```
LucasLefevre: this calls for two distinct tests with valid/invalid

packages/o-spreadsheet-engine/src/helpers/formulas.ts
hokolomopo: Yeah I forgot about ot_transforms ... oops

packages/o-spreadsheet-engine/src/plugins/core/range.ts
hokolomopo: Not concluant

packages/o-spreadsheet-engine/src/collaborative/ot/ot_specific.ts
```diff
@@ -348,3 +356,22 @@ function groupHeadersTransformation(
...
+  if (toTransform.type === "DELETE_NAMED_RANGE" && toTransform.name === executed.oldRangeName) {
+    return { ...toTransform, name: executed.newRangeName };
+  }
+  if (
+    toTransform.type === "UPDATE_NAMED_RANGE" &&
+    toTransform.oldRangeName === executed.oldRangeName
+  ) {
+    return { ...toTransform, oldRangeName: executed.newRangeName };
+  }
+  return executed;
```
LucasLefevre: oops, wrong command 🙃 
looks like it's missing a test
```suggestion
  return toTransform;
```

packages/o-spreadsheet-engine/src/plugins/core/named_range.ts
```diff
@@ -0,0 +1,156 @@
...
+interface NamedRangeState {
+  readonly namedRanges: Array<NamedRange>;
+}
+
+/**
+  - \p{L} is for any letter (from any language)
+  - \p{N} is for any number
+  - the u flag at the end is for unicode, which enables the `\p{...}` syntax
+ */
+const validNamedRangeNameRegex = /^[\p{L}|\p{N}|_|\.]*$/u;
```
LucasLefevre: `"my|range"` is a valid name, but it should not
```suggestion
const validNamedRangeNameRegex = /^[\p{L}\p{N}_\.]+$/u;
```

LucasLefevre:
robodoo rebase-ff

PR #7606, opened 2025-12-04
[FIX] charts: remove zoom slicer for scatter plot

anhe-odoo:
## Task Description

This PR aims to remove the zoomable feature for the scatter plot, as it's kind of non-sense to be able to zoom on an axis and not on the other for this type of chart. Moreover, we will soon be able to manually set the min/max of each axis manually (in master).

## Related Task
- Task: [5388389](https://www.odoo.com/odoo/2328/tasks/5388389)

### DIFF
--- a/tests/figures/chart/scatter_chart_plugin.test.ts
+++ b/tests/figures/chart/scatter_chart_plugin.test.ts
@@ -1,0 +1,27 @@
+import { ChartCreationContext } from "../../../src";
+import { ScatterChart } from "../../../src/helpers/figures/charts/scatter_chart";
+import { GENERAL_CHART_CREATION_CONTEXT } from "../../test_helpers/chart_helpers";
+
+describe("scatter chart", () => {
+  test("create scatter chart from creation context", () => {
+    const context: Required<ChartCreationContext> = {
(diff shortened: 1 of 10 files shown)


tests/figures/chart/scatter_chart_plugin.test.ts
```diff
@@ -0,0 +1,27 @@
+import { ChartCreationContext } from "../../../src";
+import { ScatterChart } from "../../../src/helpers/figures/charts/scatter_chart";
+import { GENERAL_CHART_CREATION_CONTEXT } from "../../test_helpers/chart_helpers";
+
+describe("scatter chart", () => {
+  test("create scatter chart from creation context", () => {
+    const context: Required<ChartCreationContext> = {
+      ...GENERAL_CHART_CREATION_CONTEXT,
+      range: [{ dataRange: "Sheet1!B1:B4", yAxisId: "y1" }],
+    };
```
hokolomopo, line 10: set zoomable = true in the ChartCreationContext to make sure it doesn't end up in the definiton

rrahir:
robodoo r+

PR #7608, opened 2025-12-05
showValue in excel (import/export)

matho-odoo:
## Description:

In excel, we can setup each chart dataset and add the value as a label 
We should properly export this option as well

Task: [5213598](https://www.odoo.com/odoo/2328/tasks/5213598)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


tests/xlsx/xlsx_import_export.test.ts
```diff
@@ -311,6 +311,7 @@ describe("Export data to xlsx then import it", () => {
       legendPosition: "top" as const,
       stacked: false,
       labelsAsText: false,
+      showValues: true,
```
anhe-odoo: Maybe we could add some chart without showValues ? So that we have `showValues: false` at some places here ?

packages/o-spreadsheet-engine/src/xlsx/functions/charts.ts
```diff
@@ -992,23 +1011,28 @@ function addDoughnutChart(
...
       }" />
-      ${insertDataLabels()}
       ${joinXmlNodes(dataSetsNodes)}
+      ${insertDataLabels({ showValues: chart.showValues })}
     </c:doughnutChart>
   `;
 }
 
-function insertDataLabels({ showLeaderLines } = { showLeaderLines: false }): XMLString {
+function insertDataLabels(
```
anhe-odoo, line 1022: I know it was done this way before but I'm not sure we should pass and object here for only two arguments

tests/xlsx/xlsx_import_export.test.ts
anhe-odoo, line 314: Maybe we could add some chart without showValues ? So that we have `showValues: false` at some places here ?

LucasLefevre (commented):
robodoo r+

PR #7610, opened 2025-12-05
[FIX] Pivots: Recompute measure on indirect dependency update 

rrahir:
## Description:

Task: [5349782](https://www.odoo.com/odoo/2328/tasks/5349782)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/functions/helper_lookup.ts
+++ b/src/functions/helper_lookup.ts
@@ -2,7 +2,7 @@ import { isZoneInside, positionToZone, zoneToXc } from "../helpers";
 import { _t } from "../translation";
 import { EvalContext, FunctionResultObject, Getters, Maybe, Range, UID } from "../types";
 import { CircularDependencyError, EvaluationError, InvalidReferenceError } from "../types/errors";
-import { PivotCoreDefinition, PivotCoreMeasure } from "../types/pivot";
+import { PivotCoreMeasure } from "../types/pivot";
 
 /**
  * Get the pivot ID from the formula pivot ID.
@@ -37,11 +37,12 @@ export function assertDomainLength(domain: Maybe<FunctionResultObject>[]) {
 
 export function addPivotDependencies(
   evalContext: EvalContext,
-  coreDefinition: PivotCoreDefinition,
+  pivotId: UID,
   forMeasures: PivotCoreMeasure[]
 ) {
   //TODO This function can be very costly when used with PIVOT.VALUE and PIVOT.HEADER
   const dependencies: Range[] = [];
+  const coreDefinition = evalContext.getters.getPivotCoreDefinition(pivotId);
 
   if (coreDefinition.type === "SPREADSHEET" && coreDefinition.dataSet) {
     const { sheetId, zone } = coreDefinition.dataSet;
@@ -62,8 +63,7 @@ export function addPivotDependencies(
 
   for (const measure of forMeasures) {
     if (measure.computedBy) {
-      const formula = evalContext.getters.getMeasureCompiledFormula(measure);
-      dependencies.push(...formula.dependencies.filter((range) => !range.invalidXc));
+      dependencies.push(...evalContext.getters.getMeasureFullDependencies(pivotId, measure));
     }
   }
   const originPosition = evalContext.__originCellPosition;
--- a/src/plugins/core/pivot.ts
+++ b/src/plugins/core/pivot.ts
@@ -1,5 +1,5 @@
 import { compile } from "../../formulas";
-import { deepCopy, deepEquals } from "../../helpers";
+import { deepCopy, deepEquals, getCanonicalSymbolName } from "../../helpers";
 import { createPivotFormula, getMaxObjectId } from "../../helpers/pivot/pivot_helpers";
 import { pivotRegistry } from "../../helpers/pivot/pivot_registry";
 import { SpreadsheetPivotTable } from "../../helpers/pivot/table_spreadsheet_pivot";
@@ -22,11 +22,16 @@ interface Pivot {
   formulaId: string;
 }
 
+interface MeasureState {
+  formula: RangeCompiledFormula;
+  dependencies: Range[];
+}
+
 interface CoreState {
   nextFormulaId: number;
   pivots: Record<UID, Pivot | undefined>;
   formulaIds: Record<UID, string | undefined>;
-  compiledMeasureFormulas: Record<UID, Record<string, RangeCompiledFormula | undefined>>;
+  compiledMeasureFormulas: Record<UID, Record<string, MeasureState | undefined>>;
 }
 
 export class PivotCorePlugin extends CorePlugin<CoreState> implements CoreState {
@@ -39,6 +44,7 @@ export class PivotCorePlugin extends CorePlugin<CoreState> implements CoreState 
     "getMeasureCompiledFormula",
     "getPivotName",
     "isExistingPivot",
+    "getMeasureFullDependencies",
   ] as const;
 
   readonly nextFormulaId: number = 1;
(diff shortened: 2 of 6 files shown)


src/plugins/core/pivot.ts
```diff
@@ -228,27 +246,73 @@ export class PivotCorePlugin extends CorePlugin<CoreState> implements CoreState
...
+      if (
+        !existingMeasure ||
+        exploredMeasures.has(existingMeasure.id) ||
+        !existingMeasure.computedBy
+      )
+        continue;
+
+      rangeList.push(
+        ...this.computeMeasureFullDependencies(pivotId, existingMeasure, exploredMeasures)
+      );
```
hokolomopo: this will create an infinite loop if we have cyclic dependencies in measures no ?

src/plugins/core/pivot.ts
```diff
@@ -228,27 +246,73 @@ export class PivotCorePlugin extends CorePlugin<CoreState> implements CoreState
...
+          measure.id !== measureCandidate.id
+      );
+
+      if (
+        !existingMeasure ||
+        exploredMeasures.has(existingMeasure.id) ||
+        !existingMeasure.computedBy
+      )
+        continue;
+
```
hokolomopo: nitpick: brackets

src/plugins/core/pivot.ts
```diff
@@ -314,22 +378,23 @@ export class PivotCorePlugin extends CorePlugin<CoreState> implements CoreState
...
-      this.compileMeasureFormula(sheetId, newFormulaString)
-    );
+  private replaceMeasureFormula(formulaString: string, newFormulaString: string) {
+    // this.history.update("compiledMeasureFormulas", sheetId, formulaString, undefined);
+    // this.history.update(
+    //   "compiledMeasureFormulas",
+    //   sheetId,
+    //   newFormulaString,
+    //   this.compileMeasureFormula(sheetId, newFormulaString)
+    // );
```
hokolomopo: comment

src/plugins/core/pivot.ts
```diff
@@ -22,11 +22,16 @@ interface Pivot {
   formulaId: string;
 }
 
+interface MeasureState {
+  formula: RangeCompiledFormula;
+  fullDependencies: Range[];
```
hokolomopo: I'd remove the "full" in the name of this/the getter

src/plugins/core/pivot.ts
```diff
@@ -228,27 +246,73 @@ export class PivotCorePlugin extends CorePlugin<CoreState> implements CoreState
...
+          measure.id,
+          "formula",
           compiledFormula
         );
       }
     }
+    for (const measure of measures) {
+      if (measure.computedBy) {
+        const dependencies = this.computeMeasureFullDependencies(pivotId, measure);
+        this.history.update(
```
hokolomopo, line 273: Can probably be done in the same loop as the complied formula, if `computeMeasureFullDependencies` returned the whole `MeasureState` object.

rrahir: if measure1 refers to measure2 and the formula of measure2 was not already compiled, you'll have a problem ^^

regarding the generic remark "always return all of the dependencies of all of the measures of the pivot " this would only have an impact on split pivot formulas which would end up having the dependencies of every measure, even those that have 0 relation to it. I don't know if it's a deal breaker

rrahir:
@hokolomopo I have implemented your suggestion, we now store a single array of range dependecies for every pivot. it will trigger the unwanted recomputation of static pivot formulas but should not be too troublesome

src/plugins/core/pivot.ts
```diff
@@ -26,7 +26,8 @@ interface CoreState {
   nextFormulaId: number;
   pivots: Record<UID, Pivot | undefined>;
   formulaIds: Record<UID, string | undefined>;
-  compiledMeasureFormulas: Record<UID, Record<string, RangeCompiledFormula | undefined>>;
+  compiledMeasureFormulas: Record<UID, Record<string, RangeCompiledFormula>>;
+  pivotMeasureDependencies: Record<UID, Range[]>;
```
LucasLefevre: nitpicking to make it shorter, we know we are in the context of pivots
```suggestion
  measureDependencies: Record<UID, Range[]>;
```

src/plugins/core/pivot.ts
```diff
@@ -228,27 +243,30 @@ export class PivotCorePlugin extends CorePlugin<CoreState> implements CoreState
...
-        );
+        this.history.update("compiledMeasureFormulas", pivotId, measure.id, compiledFormula);
       }
     }
+
+    const ranges: Range[] = [];
+    for (const measure of measures) {
+      if (measure.computedBy) {
+        const compiledFormula = this.getMeasureCompiledFormula(pivotId, measure);
+        ranges.push(...compiledFormula.dependencies);
```
LucasLefevre: this could be mixed in the for loop just above

src/functions/helper_lookup.ts
```diff
@@ -62,8 +63,7 @@ export function addPivotDependencies(
 
   for (const measure of forMeasures) {
     if (measure.computedBy) {
-      const formula = evalContext.getters.getMeasureCompiledFormula(measure);
-      dependencies.push(...formula.dependencies.filter((range) => !range.invalidXc));
+      dependencies.push(...evalContext.getters.getMeasureFullDependencies(pivotId, measure));
```
LucasLefevre, line 66: Since the new approach where we get all dependencies, this adds the same dependencies multiple times.

src/plugins/core/pivot.ts
```diff
@@ -228,27 +246,75 @@ export class PivotCorePlugin extends CorePlugin<CoreState> implements CoreState
...
+  ): Range[] {
+    const rangeList: Range[] = [];
+    const definition = this.getPivotCoreDefinition(pivotId);
+    const formula = this.getMeasureCompiledFormula(pivotId, measure);
+    exploredMeasures.add(measure.id);
+    for (const token of formula.tokens) {
+      if (token.type !== "SYMBOL") {
+        continue;
+      }
+      const existingMeasure = definition.measures.find(
```
LucasLefevre: nitpick ?
```suggestion
      const otherMeasure = definition.measures.find(
```

src/plugins/core/pivot.ts
```diff
@@ -228,27 +246,75 @@ export class PivotCorePlugin extends CorePlugin<CoreState> implements CoreState
...
+
+      if (
+        !existingMeasure ||
+        exploredMeasures.has(existingMeasure.id) ||
+        !existingMeasure.computedBy
+      ) {
+        continue;
+      }
+
+      // TODORAR this will fail with cyclic dependencies
```
LucasLefevre: 👀

src/plugins/core/pivot.ts
LucasLefevre: Same. I'd remove "full"

src/plugins/core/pivot.ts
```diff
@@ -314,22 +380,16 @@ export class PivotCorePlugin extends CorePlugin<CoreState> implements CoreState
...
 
-  private replaceMeasureFormula(sheetId: UID, formulaString: string, newFormulaString: string) {
-    this.history.update("compiledMeasureFormulas", sheetId, formulaString, undefined);
-    this.history.update(
-      "compiledMeasureFormulas",
-      sheetId,
-      newFormulaString,
-      this.compileMeasureFormula(sheetId, newFormulaString)
-    );
+  private replaceMeasureFormula(formulaString: string, newFormulaString: string) {
```
LucasLefevre: with the new data structure, this could take `pivotId` and `measure` and it would allow simplifying this method I think (and remove `shouldUpdateMeasures`)

src/plugins/core/pivot.ts
```diff
@@ -228,27 +246,75 @@ export class PivotCorePlugin extends CorePlugin<CoreState> implements CoreState
...
+      }
+    }
+  }
+
+  private computeMeasureFullDependencies(
+    pivotId: UID,
+    measure: PivotCoreMeasure,
+    exploredMeasures: Set<string> = new Set()
+  ): Range[] {
+    const rangeList: Range[] = [];
```
LucasLefevre: `dependencyRanges` ? `rangeDependencies` ?

PR #7618, opened 2025-12-09
[FIX] data_validation: should not hide dv suggations on mouse up

rmbh-odoo:
## Description:
Current behavior before PR:
- Allowing zone de-selection introduced commitSelection on mouse up,
  which recalculates the selection and updates the zone.
- As a result, data validation suggestions were hidden on mouse up.

- Suggestions are hidden only when the zone actually changes during
  commitSelection.

Task: [5392156](https://www.odoo.com/odoo/2328/tasks/5392156)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/composer/composer/abstract_composer_store.ts
+++ b/src/components/composer/composer/abstract_composer_store.ts
@@ -115,7 +115,12 @@ export abstract class AbstractComposerStore extends SpreadsheetStore {
   abstract stopEdition(direction?: Direction): void;
 
   private handleEvent(event: SelectionEvent) {
-    this.hideHelp();
+    const hasSelectionChanged =
+      event.mode === "commitSelection" && !isEqual(event.anchor.zone, event.previousAnchor.zone);
+    if (hasSelectionChanged) {
+      this.hideHelp();
+    }
+
     const sheetId = this.getters.getActiveSheetId();
     let unboundedZone: UnboundedZone;
     if (event.options.unbounded) {
--- a/src/components/grid_overlay/grid_overlay.ts
+++ b/src/components/grid_overlay/grid_overlay.ts
@@ -1,3 +1,4 @@
+import { EditionMode } from "@odoo/o-spreadsheet-engine/types/misc";
 import { SpreadsheetChildEnv } from "@odoo/o-spreadsheet-engine/types/spreadsheet_env";
 import { Component, onMounted, onWillUnmount, useExternalListener, useRef } from "@odoo/owl";
 import { deepEquals, positionToZone } from "../../helpers";
@@ -149,6 +150,7 @@ interface Props {
   onGridMoved: (deltaX: Pixel, deltaY: Pixel) => void;
   gridOverlayDimensions: string;
   getGridSize: () => { width: number; height: number };
+  composerEditionMode: EditionMode;
 }
 
 export class GridOverlay extends Component<Props, SpreadsheetChildEnv> {
@@ -162,6 +164,7 @@ export class GridOverlay extends Component<Props, SpreadsheetChildEnv> {
     gridOverlayDimensions: String,
     slots: { type: Object, optional: true },
     getGridSize: Function,
+    composerEditionMode: { type: String, optional: true },
   };
   static components = {
     FiguresContainer,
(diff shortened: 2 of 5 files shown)


src/components/grid_overlay/grid_overlay.ts
```diff
@@ -248,12 +248,14 @@ export class GridOverlay extends Component<Props, SpreadsheetChildEnv> {
...
     const clickedIcon = this.getInteractiveIconAtEvent(zoomedMouseEvent);
-    if (clickedIcon) {
+    if (clickedIcon?.onClick) {
       this.env.model.selection.getBackToDefault();
+      clickedIcon.onClick(clickedIcon.position, this.env);
+      return;
     }
+
+    const [col, row] = this.getCartesianCoordinates(zoomedMouseEvent);
     this.props.onCellClicked(
```
hokolomopo: The icon click triggering the cell onClick was intentional, it's not the biggest deal bu we want to selection to change when clicking on a data validation checkbox for example. Or clicking on a checkbox should close the filter menu popover.

rmbh-odoo: Yes, you’re right, the cell `onClick` was intentional. I’ve updated the data validation behavior so we don’t capture the selection on edit start, as it’s not needed there. This fixes the issue, and the DV icon no longer blocks selection. 
Let me know if I missed anything. :smiling_face_with_tear:

src/components/composer/composer/abstract_composer_store.ts
rmbh-odoo, line 434: Currently, clicking the data validation icon only `startCellEdition()` and immediately captures the selection. As a result, dragging after clicking the icon does not expand the selection.

I’m not sure about the real use case for this behavior, but for other icons (such as filter and checkbox), dragging does expand the selection.
I attempted to align the behavior in this commit: [c9077c1](https://github.com/odoo/o-spreadsheet/commit/c9077c15ad305620aa37efa3189c5f7aaab1a74e)

However, I’m not sure whether it’s worth passing additional props just to support this use case 😕

src/components/grid_overlay/grid_overlay.ts
```diff
@@ -325,6 +328,6 @@ export class GridOverlay extends Component<Props, SpreadsheetChildEnv> {
 
       return isPointInsideRect(x, y, this.env.model.getters.getCellIconRect(icon, cellRect));
     });
-    return icon?.onClick ? icon : undefined;
+    return icon?.onClick && this.props.composerEditionMode !== "selecting" ? icon : undefined;
```
rmbh-odoo, line 331: This condition was added intentionally.
When the composer is in selecting mode `=SUM(`, clicking on the DV or other icons should not
trigger their behavior. Instead, the click should act as a normal grid capture.

rrahir: This goes to the opposite way of https://github.com/odoo/o-spreadsheet/pull/6480/commits/2d7e0ebc1bdec050ecb4cb083cda59033020e0ea

Not saying the decision is good or bad but I m positive that both selectionInput and composer should behave similarly (so in doubt, I 'd do this change in master ^^)

edit: this condition should be move. in `onCellClicked` , the icon does exist and should be returned all the time

src/components/composer/composer/abstract_composer_store.ts
```diff
@@ -115,7 +115,12 @@ export abstract class AbstractComposerStore extends SpreadsheetStore {
   abstract stopEdition(direction?: Direction): void;
 
   private handleEvent(event: SelectionEvent) {
-    this.hideHelp();
+    const hasSelectionChanged =
+      event.mode === "commitSelection" && !isEqual(event.anchor.zone, event.previousAnchor.zone);
```
rrahir, line 119: is the condition on zone equality really necessary ? what about the anchor?
At first glance, early return if event.mode === "commitSelection" sems to do the trick and every test passes

src/components/grid_overlay/grid_overlay.ts
rmbh-odoo: Oops, I wasn’t aware of that. Thanks for pointing it out. :+1: 
For now, we can move forward without this change. If this flow is needed, we can handle it in master. Since the target branch has changed, it makes sense to continue with a new PR.

src/components/composer/composer/abstract_composer_store.ts
rmbh-odoo: After a deeper investigation, I realized this is not limited to commitSelection only. I’ve added detailed logs and notes in the task for more context.
Hope this makes sense. :slightly_smiling_face:

rmbh-odoo:
Closing this PR since the fix target has been changed to saas-18.4.
The updated fix is available in the new PR: https://github.com/odoo/o-spreadsheet/pull/7712

PR #7619, opened 2025-12-09
[IMP] targetDependantCommand: avoid dispatch if target is empty

fdamhaut:
## Description:

Task: [4863945](https://www.odoo.com/odoo/2328/tasks/4863945)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/packages/o-spreadsheet-engine/src/plugins/core/sheet.ts
+++ b/packages/o-spreadsheet-engine/src/plugins/core/sheet.ts
@@ -26,6 +26,7 @@ import {
   CreateSheetCommand,
   FreezeColumnsCommand,
   FreezeRowsCommand,
+  isRangeDependant,
   isTargetDependent,
   RenameSheetCommand,
   UpdateCellPositionCommand,
@@ -1048,6 +1049,9 @@ export class SheetPlugin extends CorePlugin<SheetState> implements SheetState {
    * not outside the sheet.
    */
   private checkZonesAreInSheet(cmd: CoreCommand): CommandResult {
+    if (isRangeDependant(cmd) && cmd.ranges.length === 0) {
+      return CommandResult.EmptyRange;
+    }
     if (!("sheetId" in cmd)) {
       return CommandResult.Success;
     }
--- a/tests/sheet/sheets_plugin.test.ts
+++ b/tests/sheet/sheets_plugin.test.ts
@@ -34,7 +34,10 @@ import {
   unMerge,
   undo,
 } from "../test_helpers/commands_helpers";
-import { TEST_COMMANDS_TARGET_DEPENDENT } from "../test_helpers/constants";
+import {
+  TEST_COMMANDS_RANGE_DEPENDENT,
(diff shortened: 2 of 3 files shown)


tests/sheet/sheets_plugin.test.ts
```diff
@@ -1201,6 +1202,17 @@ describe("sheets", () => {
     }
   );
 
+  test.each(Object.entries(TEST_COMMANDS).filter(([cmdName, cmd]) => isTargetDependent(cmd)))(
```
hokolomopo: We already have OT_TESTS_TARGET_DEPENDANT_COMMANDS constant. But filtering on the command content sounds like a better idea than a static list, could you maybe change OT_TESTS_TARGET_DEPENDANT_COMMANDS (and similar constants) to do that ?

packages/o-spreadsheet-engine/src/plugins/core/sheet.ts
```diff
@@ -1051,6 +1052,9 @@ export class SheetPlugin extends CorePlugin<SheetState> implements SheetState {
     ) {
       return CommandResult.InvalidSheetId;
     }
+    if (isTargetDependent(cmd) && cmd.target.length === 0) {
+      return CommandResult.EmptyTarget;
+    }
```
hokolomopo, line 1068: could we do the same for the rangeDependantCommands ? I don't quite get @rrahir comment in the pad about `ADD_CONDITIONAL_FORMAT ` being valid with an empty list of range, we have an explicit check to reject it if the ranges are empty.

rrahir: don't know either where this came from, honestly don't recall the task in the first place ,and indeed my comment was 100% false :/

packages/o-spreadsheet-engine/src/components/translations_terms.ts
```diff
@@ -111,7 +112,7 @@ export const RemoveDuplicateTerms = {
   Errors: {
     Unexpected: _t("Cannot remove duplicates for an unknown reason"),
     [CommandResult.MoreThanOneRangeSelected]: _t("Please select only one range of cells"),
-    [CommandResult.EmptyTarget]: _t("Please select a range of cells containing values."),
+    [CommandResult.EmptySelectedRange]: _t("Please select a range of cells containing values."),
```
rrahir, line 115: the "EmptyTarget" was actually correct for the clipboard events, you have to have an active selection in order to aste but functionally speaking, it's always true. the error translation was only meant for the data_cleanup feature, especially, the remove duplicate feature, for which you need your selected range not to be empty ^^

the clipboards should keep returning CommandResult.EmptyTarget

rrahir (changes requested):

rrahir:
image

classic :p

robodoo r+

rrahir:
robodoo rebase-ff

PR #7622, opened 2025-12-10
[PERF] evaluation: cached dynamic dependencies

LucasLefevre:
Steps to reproduce:
- create a (large) pivot from a range
- convert it to individual formulas

=> for each formula, the formula dynamically adds (at evaluation time) its dependency on the underlying pivot range.

The evaluation engine ensures this dependency has already been evaluated.

Before this commit, we were checking every single position one by one, for every pivot formula. Even if `computeAndSave` was doing nothing because it was already computed, the code still iterated over the entire range.

With this commit, we rely on the `range` function from the compilation parameters, which is a built-in cache of ranges that were previously evaluated.

With the large demo pivot data set:

- grouped by "Month & Year" (8*183=1464 formulas)

before: ~3750ms
after:   ~787ms

- grouped by "Day" (8*4616=36928 formulas)

before: ~79500ms
after:   ~4612ms

Task: 5407156

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_core_views/cell_evaluation/evaluator.ts
+++ b/src/plugins/ui_core_views/cell_evaluation/evaluator.ts
@@ -116,13 +116,8 @@ export class Evaluator {
   private addDependencies(position: CellPosition, dependencies: Range[]) {
     this.formulaDependencies().addDependencies(position, dependencies);
     for (const range of dependencies) {
-      const sheetId = range.sheetId;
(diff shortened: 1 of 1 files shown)


pro-odoo (approved):
robodoo r+

PR #7634, opened 2025-12-12
[IMP] Excel: export link colors

rrahir:
Currently, cells that contain a hyperlink are given a default color
after evaluation but this coloration is not propagated when we export
the data as xlsx file.

This revision adds the default link color to the excel export.

Task: 5410289

## Description:

Task: [5410289](https://www.odoo.com/odoo/2328/tasks/5410289)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/packages/o-spreadsheet-engine/src/plugins/ui_feature/cell_computed_style.ts
+++ b/packages/o-spreadsheet-engine/src/plugins/ui_feature/cell_computed_style.ts
@@ -1,6 +1,10 @@
 import { LINK_COLOR } from "../../constants";
 import { PositionMap } from "../../helpers/cells/position_map";
+import { toCartesian } from "../../helpers/coordinates";
+import { getItemId } from "../../helpers/data_normalization";
 import { isObjectEmptyRecursive, removeFalsyAttributes } from "../../helpers/misc";
+import { recomputeZones } from "../../helpers/recompute_zones";
+import { isZoneInside, toZone, zoneToXc } from "../../helpers/zones";
 import {
   Command,
   invalidateBordersCommands,
@@ -8,6 +12,7 @@ import {
   invalidateEvaluationCommands,
 } from "../../types/commands";
 import { Border, CellPosition, Style } from "../../types/misc";
+import { ExcelWorkbookData } from "../../types/workbook_data";
 import { UIPlugin } from "../ui_plugin";
 import { doesCommandInvalidatesTableStyle } from "./table_computed_style";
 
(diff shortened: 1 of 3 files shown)


packages/o-spreadsheet-engine/src/migrations/migration_steps.ts
```diff
@@ -600,3 +606,56 @@ function fixOverlappingFilters(data: any): any {
   }
   return data;
 }
+
+function addHyperlinkStyle(data: WorkbookData): WorkbookData {
+  for (const sheet of data.sheets || []) {
+    for (const xc in sheet.cells || {}) {
+      const content = sheet.cells[xc];
+      if (content && (isMarkdownLink(content) || isWebLink(content))) {
+        // debugger;
```
fdamhaut: 👀

src/components/link/link_editor/link_editor.ts
```diff
@@ -86,16 +87,20 @@ export class LinkEditor extends Component<LinkEditorProps, SpreadsheetChildEnv>
   }
 
   save() {
+    const sheetId = this.env.model.getters.getActiveSheetId();
     const { col, row } = this.props.cellPosition;
     const locale = this.env.model.getters.getLocale();
     const label = this.link.label
       ? canonicalizeNumberContent(this.link.label, locale)
       : this.link.url;
+    // TODORAR add this feature to UPDATE_CELL in the composr as well ?
```
fdamhaut: TODO ?

packages/o-spreadsheet-engine/src/migrations/migration_steps.ts
fdamhaut: up

packages/o-spreadsheet-engine/src/helpers/data_normalization.ts
```diff
@@ -15,7 +15,8 @@ const globalIdCounter = new WeakMap<ItemsDic<any>, number>();
 export function getItemId<T>(item: T, itemsDic: ItemsDic<T>) {
   if (!globalReverseLookup.has(itemsDic)) {
     globalReverseLookup.set(itemsDic, new Map());
-    globalIdCounter.set(itemsDic, 0);
+    const x = Math.max(...Object.keys(itemsDic).map((k) => Number(k)), 0);
+    globalIdCounter.set(itemsDic, x);
```
fdamhaut: If you start with a IdCounter > 0 you should also populate the reverse Lookup otherwise you keep the current behaviour with the only difference being the starting index of the idCounter

packages/o-spreadsheet-engine/src/migrations/migration_steps.ts
```diff
@@ -576,6 +577,11 @@ migrationStepRegistry
       }
       return data;
     },
+  })
+  .add("19.2.0", {
+    migrate(data: WorkbookData): any {
+      return addHyperlinkStyle(data);
```
hokolomopo: you'd also have to do an odoo migration PR if we want to be correct

packages/o-spreadsheet-engine/src/migrations/migration_steps.ts
```diff
@@ -600,3 +606,37 @@ function fixOverlappingFilters(data: any): any {
   }
   return data;
 }
+
+function addHyperlinkStyle(data: WorkbookData): WorkbookData {
+  for (const sheet of data.sheets || []) {
+    for (const xc in sheet.cells || {}) {
+      const content = sheet.cells[xc];
+      if (content && (isMarkdownLink(content) || isWebLink(content))) {
```
hokolomopo: hasHyperlinkContent ?

packages/o-spreadsheet-engine/src/helpers/misc.ts
```diff
@@ -236,6 +236,10 @@ export function parseSheetUrl(sheetLink: string) {
   throw new Error(`${sheetLink} is not a valid sheet link`);
 }
 
+export function hasHyperlinkContent(content: string): boolean {
+  return isMarkdownLink(content) || isWebLink(content) || content.startsWith("=HYPERLINK(");
```
hokolomopo: So it doens't work anymore for `=IF(A1, HYPERLINK(...))`

hokolomopo: Edit: oh it never worked in the first place. Probably it should work ?

rrahir: it does work but requires the protocol prefix ^^ 
`=IF(TRUE, CONCAT("https://", "google.com"))` works

packages/o-spreadsheet-engine/src/plugins/ui_core_views/cell_evaluation/evaluation_plugin.ts
```diff
@@ -382,6 +383,22 @@ export class EvaluationPlugin extends CoreViewPlugin {
...
+        ...dataValidationStyle,
+        ...cell?.style,
+        ...cfStyle,
+      };
+      if (evaluatedCell.link && !computedStyle.textColor) {
+        const style = { ...cell?.style, textColor: LINK_COLOR };
+        const newStyleId = getItemId(style, data.styles);
+        exportedSheetData.styles[xc] = newStyleId;
+      }
+
```
hokolomopo: Would it be possible to do that in `CellComputedStylePlugin` ?

packages/o-spreadsheet-engine/src/plugins/ui_feature/cell_computed_style.ts
```diff
@@ -96,21 +99,40 @@ export class CellComputedStylePlugin extends UIPlugin {
...
 
-    return computedStyle;
+  exportForExcel(data: ExcelWorkbookData) {
+    for (const sheet of data.sheets) {
+      for (const xc in sheet.cells) {
+        const position = { sheetId: sheet.id, ...toCartesian(xc) };
+        const evaluatedCell = this.getters.getEvaluatedCell(position);
+        const computedStyle = this.getComputedStyle(position);
+        const cell = this.getters.getCell(position);
+        if (evaluatedCell.link && !computedStyle.textColor) {
```
LucasLefevre: nitpick: `getters.getCell` could be inside the `if` to avoid calling it uselessly 99%
 of the time.

packages/o-spreadsheet-engine/src/plugins/ui_feature/cell_computed_style.ts
```diff
@@ -96,21 +99,40 @@ export class CellComputedStylePlugin extends UIPlugin {
...
+          const style = { ...cell?.style, textColor: LINK_COLOR };
+          const newStyleId = getItemId(style, data.styles);
+          sheet.styles[xc] = newStyleId;
```
LucasLefevre: If `xc` is A1, but there's also a style on A1:A3, the style set for A1:A3 might takes precedence over this one depending on the insertion order

rrahir: the 'non-dynamic' style being added before, we are somehow guaranteed to be defined after but I agree this could be better. My only fear is that to be correct, we'll need to split the zones (pretty much already done in a previous version https://github.com/odoo/o-spreadsheet/commit/7b0aec20d89f2cfcb7fe297e27974e6d4e1694d5#diff-7bec0b4ad252022ab40ce3b573e008777662e0d3c090640e09b1b96e3c2238a8R610) and in worse case, this can crush the performances :'(

packages/o-spreadsheet-engine/src/plugins/ui_feature/cell_computed_style.ts
```diff
@@ -71,22 +76,79 @@ export class CellComputedStylePlugin extends UIPlugin {
...
       ...removeFalsyAttributes(cfStyle),
     };
-    const evaluatedCell = this.getters.getEvaluatedCell(position);
-    if (evaluatedCell.link && !computedStyle.textColor) {
-      computedStyle.textColor = LINK_COLOR;
-    }
+  }
 
-    return computedStyle;
+  exportForExcel(data: ExcelWorkbookData) {
```
hokolomopo, line 100: The logic here is quite complex, it would deserve a non-snapshot test to test everything is correct. Check that the style zones are correct when a link is in the middle, and that standalone links have the correct style

packages/o-spreadsheet-engine/src/plugins/ui_feature/cell_computed_style.ts
```diff
@@ -71,22 +76,79 @@ export class CellComputedStylePlugin extends UIPlugin {
...
       ...removeFalsyAttributes(tableStyle),
       ...removeFalsyAttributes(dataValidationStyle),
       ...removeFalsyAttributes(cell?.style),
       ...removeFalsyAttributes(cfStyle),
     };
-    const evaluatedCell = this.getters.getEvaluatedCell(position);
-    if (evaluatedCell.link && !computedStyle.textColor) {
-      computedStyle.textColor = LINK_COLOR;
-    }
+  }
```
hokolomopo, line 98: those change are useless right ?

rrahir: I don't think so ^^ the method getComputedStyle is used at 2 places

packages/o-spreadsheet-engine/src/plugins/ui_feature/cell_computed_style.ts
rrahir: added a test :)

PR #7643, opened 2025-12-16
[IMP] Figure: Allow dragNdrop of multiple figures

fdamhaut:
Task: 3323871

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/actions/figure_menu_actions.ts
+++ b/src/actions/figure_menu_actions.ts
@@ -13,6 +13,7 @@ export function getChartMenuActions(figureId: UID, env: SpreadsheetChildEnv): Ac
     return [];
   }
   const menuItemSpecs: ActionSpec[] = [
+    getMergeCarouselMenuItem(figureId, env),
     {
       id: "edit",
       name: _t("Edit"),
@@ -184,7 +185,9 @@ function getCopyMenuItem(
     name: _t("Copy"),
     shortcut: "Ctrl+C",
     execute: async () => {
-      env.model.dispatch("SELECT_FIGURE", { figureId });
+      if (!env.model.getters.getSelectedFigureIds().includes(figureId)) {
+        env.model.dispatch("SELECT_FIGURE", { figureId });
+      }
       env.model.dispatch("COPY");
       const osClipboardContent = await env.model.getters.getClipboardTextAndImageContent();
       await env.clipboard.write(osClipboardContent);
@@ -203,7 +206,9 @@ function getCutMenuItem(figureId: UID, env: SpreadsheetChildEnv): ActionSpec {
     name: _t("Cut"),
     shortcut: "Ctrl+X",
     execute: async () => {
-      env.model.dispatch("SELECT_FIGURE", { figureId });
+      if (!env.model.getters.getSelectedFigureIds().includes(figureId)) {
+        env.model.dispatch("SELECT_FIGURE", { figureId });
+      }
       env.model.dispatch("CUT");
       await env.clipboard.write(await env.model.getters.getClipboardTextAndImageContent());
     },
@@ -251,6 +256,7 @@ function getCopyAsImageMenuItem(figureId: UID, env: SpreadsheetChildEnv): Action
       });
       env.notifyUser({ sticky: false, type: "success", text: _t("Chart copied to clipboard") });
     },
+    isVisible: (env) => env.model.getters.getSelectedFigureIds().length <= 1,
     isReadonlyAllowed: true,
     isEnabledOnLockedSheet: true,
   };
@@ -276,6 +282,7 @@ function getDownloadChartMenuItem(figureId: UID, env: SpreadsheetChildEnv): Acti
       }
       downloadFile(url, "chart");
     },
+    isVisible: (env) => env.model.getters.getSelectedFigureIds().length <= 1,
     isReadonlyAllowed: true,
     isEnabledOnLockedSheet: true,
   };
@@ -286,11 +293,45 @@ function getDeleteMenuItem(figureId: UID, env: SpreadsheetChildEnv): ActionSpec 
     id: "delete",
     name: _t("Delete"),
     execute: () => {
-      env.model.dispatch("DELETE_FIGURE", {
-        sheetId: env.model.getters.getActiveSheetId(),
-        figureId,
-      });
+      const selectedFiguresIds = env.model.getters.getSelectedFigureIds();
+      if (selectedFiguresIds.includes(figureId)) {
+        env.model.dispatch("DELETE_FIGURES", {
+          sheetId: env.model.getters.getActiveSheetId(),
+          figureIds: selectedFiguresIds,
+        });
+      } else {
+        env.model.dispatch("DELETE_FIGURE", {
+          sheetId: env.model.getters.getActiveSheetId(),
+          figureId,
+        });
+      }
     },
     icon: "o-spreadsheet-Icon.TRASH",
+  };
+}
+
+function getMergeCarouselMenuItem(figureId: UID, env: SpreadsheetChildEnv): ActionSpec {
+  return {
+    id: "mergeCarousel",
+    name: _t("Create carousel"),
+    isVisible: (env) => {
+      const selectedFiguresIds = env.model.getters.getSelectedFigureIds();
+      if (selectedFiguresIds.length < 2 || !selectedFiguresIds.includes(figureId)) {
+        return false;
+      }
+      const sheetId = env.model.getters.getActiveSheetId();
+      const figures = selectedFiguresIds.map((id) => env.model.getters.getFigure(sheetId, id));
+      return !figures.some((f) => f === undefined || f.tag !== "chart");
+    },
+    execute: () => {
+      const sheetId = env.model.getters.getActiveSheetId();
+      const chartFigureIds = env.model.getters.getSelectedFigureIds();
+      env.model.dispatch("MERGE_CHART_FIGURES_INTO_CAROUSEL", {
+        sheetId,
+        baseFigureId: figureId,
+        chartFigureIds,
+      });
+    },
+    icon: "o-spreadsheet-Icon.CAROUSEL",
   };
 }
--- a/src/clipboard_handlers/carousel_clipboard.ts
+++ b/src/clipboard_handlers/carousel_clipboard.ts
@@ -1,113 +1,136 @@
 import { SpreadsheetChart } from "../helpers/figures/chart";
+import { boundColRowOffsetInSheet } from "../helpers/figures/figure/figure";
 import { deepCopy } from "../helpers/misc";
 import { UuidGenerator } from "../helpers/uuid";
 import { ClipboardFigureData, ClipboardOptions, ClipboardPasteTarget } from "../types/clipboard";
 import { CommandResult } from "../types/commands";
 import { Carousel, Figure } from "../types/figure";
-import { UID, Zone } from "../types/misc";
+import { PixelPosition, UID, Zone } from "../types/misc";
 import { AbstractFigureClipboardHandler } from "./abstract_figure_clipboard_handler";
 
 type ClipboardContent = {
+  figures: ClipboardCarousel[];
+  figureIds: UID[];
+};
+
+type ClipboardCarousel = {
   figureId: UID;
   copiedSheetId: UID;
   copiedFigure: Figure;
   copiedCarousel: Carousel;
   copiedCharts: { [chartId: UID]: SpreadsheetChart };
+  offset: PixelPosition;
 };
 
 export class CarouselClipboardHandler extends AbstractFigureClipboardHandler<ClipboardContent> {
   copy(data: ClipboardFigureData): ClipboardContent | undefined {
     const sheetId = data.sheetId;
-    const figure = this.getters.getFigure(sheetId, data.figureId);
-    if (!figure) {
-      throw new Error(`No figure for the given id: ${data.figureId}`);
-    }
-    if (figure.tag !== "carousel") {
-      return;
-    }
-    const copiedFigure = { ...figure };
-    const copiedCarousel = this.getters.getCarousel(data.figureId);
-    const copiedCharts: { [chartId: UID]: SpreadsheetChart } = {};
-    for (const item of copiedCarousel.items) {
-      if (item.type === "chart") {
-        const chart = this.getters.getChart(item.chartId);
-        if (!chart) {
-          throw new Error(`No chart for the given id: ${item.chartId}`);
-        }
-        copiedCharts[item.chartId] = chart;
+    const figures: ClipboardCarousel[] = [];
+    const copyRect = this.getCopyRect(data);
+    for (const figureId of data.figureIds) {
+      const figure = this.getters.getFigure(sheetId, figureId);
+      if (!figure) {
+        throw new Error(`No figure for the given id: ${figureId}`);
       }
+      if (figure.tag !== "carousel") {
+        continue;
+      }
+      const copiedFigure = { ...figure };
+      const copiedCarousel = this.getters.getCarousel(figureId);
+      const { x, y } = this.getters.getFigureUI(sheetId, figure);
+      const copiedCharts: { [chartId: UID]: SpreadsheetChart } = {};
+      for (const item of copiedCarousel.items) {
+        if (item.type === "chart") {
+          const chart = this.getters.getChart(item.chartId);
+          if (!chart) {
+            throw new Error(`No chart for the given id: ${item.chartId}`);
+          }
+          copiedCharts[item.chartId] = chart;
+        }
+      }
+      figures.push({
+        figureId,
+        copiedFigure,
+        copiedCarousel,
+        copiedCharts,
+        copiedSheetId: sheetId,
+        offset: { x: x - copyRect.x, y: y - copyRect.y },
+      });
     }
-    return {
-      figureId: data.figureId,
-      copiedFigure,
-      copiedCarousel,
-      copiedCharts,
-      copiedSheetId: sheetId,
-    };
+    return { figures, figureIds: data.figureIds };
   }
 
-  getPasteTarget(sheetId: UID): ClipboardPasteTarget {
-    const newId = UuidGenerator.smallUuid();
-    return { zones: [], figureId: newId, sheetId };
+  getPasteTarget(
+    sheetId: UID,
+    target: Zone[],
+    content: ClipboardContent,
+    options?: ClipboardOptions
+  ): ClipboardPasteTarget {
+    const figureIds = {};
+    for (const figure of content.figures) {
+      figureIds[figure.figureId] = UuidGenerator.smallUuid();
+    }
+    return { zones: [], figureIds, sheetId };
   }
 
   paste(target: ClipboardPasteTarget, clippedContent: ClipboardContent, options: ClipboardOptions) {
-    if (!target.figureId) {
+    if (!target.figureIds) {
       return;
     }
-    const { zones, figureId } = target;
+    const { zones } = target;
     const sheetId = target.sheetId;
-    const { width, height } = clippedContent.copiedFigure;
-    const maxPosition = this.getters.getMaxAnchorOffset(sheetId, height, width);
-    let { left: col, top: row } = zones[0];
-    const offset = { x: 0, y: 0 };
-    if (col > maxPosition.col) {
-      col = maxPosition.col;
-      offset.x = maxPosition.offset.x;
-    }
-    if (row > maxPosition.row) {
-      row = maxPosition.row;
-      offset.y = maxPosition.offset.y;
-    }
-    this.dispatch("CREATE_CAROUSEL", {
-      figureId,
-      sheetId,
-      definition: { items: [] },
-      col,
-      row,
-      offset,
-      size: { height, width },
-    });
-
-    const items = deepCopy(clippedContent.copiedCarousel.items);
-    for (const item of items) {
-      if (item.type !== "chart") {
-        continue;
-      }
-      const chart = clippedContent.copiedCharts[item.chartId];
-      const newId = UuidGenerator.smallUuid();
-      const definition = SpreadsheetChart.fromDefinition(
+    for (const clippedFigure of clippedContent.figures) {
+      const figureId = target.figureIds[clippedFigure.figureId];
+      const { width, height } = clippedFigure.copiedFigure;
+      const { col, row, offset } = boundColRowOffsetInSheet(
         this.getters,
         sheetId,
-        chart.copyInSheetId(sheetId)
-      ).getDefinition();
-      this.dispatch("CREATE_CHART", { figureId, chartId: newId, sheetId, definition });
-      item.chartId = newId;
-    }
+        { col: zones[0].left, row: zones[0].top },
+        { ...clippedFigure.copiedFigure, offset: clippedFigure.offset }
+      );
+      this.dispatch("CREATE_CAROUSEL", {
+        figureId,
+        sheetId,
+        definition: { items: [] },
+        col,
+        row,
+        offset,
+        size: { height, width },
+      });
 
-    this.dispatch("UPDATE_CAROUSEL", {
-      sheetId,
-      figureId,
-      definition: { ...clippedContent.copiedCarousel, items },
-    });
+      const items = deepCopy(clippedFigure.copiedCarousel.items);
+      for (const item of items) {
+        if (item.type !== "chart") {
+          continue;
+        }
+        const chart = clippedFigure.copiedCharts[item.chartId];
+        const newId = UuidGenerator.smallUuid();
+        const definition = SpreadsheetChart.fromDefinition(
+          this.getters,
+          sheetId,
+          chart.copyInSheetId(sheetId)
+        ).getDefinition();
+        this.dispatch("CREATE_CHART", { figureId, chartId: newId, sheetId, definition });
+        item.chartId = newId;
+      }
 
-    if (options.isCutOperation) {
-      this.dispatch("DELETE_FIGURE", {
-        sheetId: clippedContent.copiedSheetId,
-        figureId: clippedContent.copiedFigure.id,
+      this.dispatch("UPDATE_CAROUSEL", {
+        sheetId,
+        figureId,
+        definition: { ...clippedFigure.copiedCarousel, items },
+      });
+
+      if (options.isCutOperation) {
+        this.dispatch("DELETE_FIGURE", {
+          sheetId: clippedFigure.copiedSheetId,
+          figureId: clippedFigure.copiedFigure.id,
+        });
+      }
+      this.dispatch("SELECT_FIGURE", {
+        figureId,
+        selectMultiple: clippedContent.figureIds.length > 1,
       });
     }
-    this.dispatch("SELECT_FIGURE", { figureId });
   }
 
   isPasteAllowed(sheetId: UID, target: Zone[], content: any, option?: ClipboardOptions) {
--- a/src/clipboard_handlers/chart_clipboard.ts
+++ b/src/clipboard_handlers/chart_clipboard.ts
@@ -1,38 +1,51 @@
 import { SpreadsheetChart } from "../helpers/figures/chart";
+import { boundColRowOffsetInSheet } from "../helpers/figures/figure/figure";
 import { UuidGenerator } from "../helpers/uuid";
 import { ClipboardFigureData, ClipboardOptions, ClipboardPasteTarget } from "../types/clipboard";
 import { CommandResult } from "../types/commands";
 import { Figure } from "../types/figure";
-import { UID, Zone } from "../types/misc";
+import { PixelPosition, UID, Zone } from "../types/misc";
 import { AbstractFigureClipboardHandler } from "./abstract_figure_clipboard_handler";
 
 type ClipboardContent = {
+  figures: ClipboardCharts[];
+  figureIds: UID[];
+};
+
+type ClipboardCharts = {
   figureId: UID;
   copiedFigure: Figure;
   copiedChart: SpreadsheetChart;
+  offset: PixelPosition;
 };
 
 export class ChartClipboardHandler extends AbstractFigureClipboardHandler<ClipboardContent> {
   copy(data: ClipboardFigureData): ClipboardContent | undefined {
     const sheetId = data.sheetId;
-    const figure = this.getters.getFigure(sheetId, data.figureId);
-    if (!figure) {
-      throw new Error(`No figure for the given id: ${data.figureId}`);
+    const figures: ClipboardCharts[] = [];
+    const copyRect = this.getCopyRect(data);
+    for (const figureId of data.figureIds) {
+      const figure = this.getters.getFigure(sheetId, figureId);
+      if (!figure) {
+        throw new Error(`No figure for the given id: ${figureId}`);
+      }
+      if (figure.tag !== "chart") {
+        continue;
+      }
+      const copiedFigure = { ...figure };
+      const { x, y } = this.getters.getFigureUI(sheetId, figure);
+      const chart = this.getters.getChartFromFigureId(figureId);
+      if (!chart) {
+        throw new Error(`No chart for the given id: ${figureId}`);
+      }
+      figures.push({
+        figureId,
+        copiedFigure,
+        copiedChart: chart,
+        offset: { x: x - copyRect.x, y: y - copyRect.y },
+      });
     }
-    if (figure.tag !== "chart") {
-      return;
-    }
-    const copiedFigure = { ...figure };
-    const chart = this.getters.getChartFromFigureId(data.figureId);
-    if (!chart) {
-      throw new Error(`No chart for the given id: ${data.figureId}`);
-    }
-    const copiedChart = chart;
-    return {
-      figureId: data.figureId,
-      copiedFigure,
-      copiedChart,
-    };
+    return { figures, figureIds: data.figureIds };
   }
 
   getPasteTarget(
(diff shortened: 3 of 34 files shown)


packages/o-spreadsheet-engine/src/helpers/rectangle.ts
```diff
@@ -0,0 +1,37 @@
+import { Rect, Zone } from "..";
```
hokolomopo: duplicated file ? there is two `rectangle.ts` helper files

src/plugins/ui_feature/figure.ts
```diff
@@ -0,0 +1,39 @@
+import { Command, CommandResult } from "../../types/commands";
+import { UIPlugin } from "../ui_plugin";
+
+export class FigureUIPlugin extends UIPlugin {
```
hokolomopo, line 5: I think we're trying to replace feature UI plugins by stores.

packages/o-spreadsheet-engine/src/plugins/ui_stateful/carousel_ui.ts
```diff
@@ -25,10 +25,12 @@ export class CarouselUIPlugin extends UIPlugin {
 
   allowDispatch(cmd: LocalCommand): CommandResult | CommandResult[] {
     switch (cmd.type) {
-      case "ADD_FIGURE_CHART_TO_CAROUSEL":
+      case "ADD_FIGURES_CHART_TO_CAROUSEL":
         if (
           !this.getters.doesCarouselExist(cmd.carouselFigureId) ||
-          this.getters.getFigure(cmd.sheetId, cmd.chartFigureId)?.tag !== "chart"
+          cmd.chartFigureIds.find(
```
hokolomopo:
```suggestion
          cmd.chartFigureIds.some(
```
nitpick

packages/o-spreadsheet-engine/src/plugins/ui_stateful/carousel_ui.ts
```diff
@@ -68,8 +70,10 @@ export class CarouselUIPlugin extends UIPlugin {
       case "ADD_NEW_CHART_TO_CAROUSEL":
         this.addNewChartToCarousel(cmd.figureId, cmd.sheetId);
         break;
-      case "ADD_FIGURE_CHART_TO_CAROUSEL":
-        this.addFigureChartToCarousel(cmd.carouselFigureId, cmd.chartFigureId, cmd.sheetId);
+      case "ADD_FIGURES_CHART_TO_CAROUSEL":
+        cmd.chartFigureIds.forEach((fid) => {
```
hokolomopo: nitpick: i'd rename all the `fid` to `figureId`, or even `chartFigureId`

src/plugins/ui_stateful/clipboard.ts
```diff
@@ -519,8 +520,8 @@ export class ClipboardPlugin extends UIPlugin {
     let innerHTML: string = "";
     const cells = this.copiedData?.cells;
     if (!cells) {
-      if (this.copiedData?.figureId) {
-        const figureId = this.copiedData.figureId;
+      if (this.copiedData?.figureIds && this.copiedData.figureIds.length) {
+        const figureId = this.copiedData.figureIds[0];
```
hokolomopo, line 588: can we add all the copied charts inside the HTML ?

packages/o-spreadsheet-engine/src/plugins/ui_stateful/selection.ts
```diff
@@ -60,7 +60,7 @@ export class GridSelectionPlugin extends UIPlugin {
     "getSelectedZones",
     "getSelectedZone",
     "getSelectedCells",
-    "getSelectedFigureId",
+    "getSelectedFiguresIds",
```
hokolomopo: There probably need to be some adaptation in odoo if we change the getter name

src/plugins/ui_stateful/selection.ts
```diff
@@ -451,6 +443,7 @@ export class GridSelectionPlugin extends UIPlugin {
     this.sheetsData[sheetIdFrom] = {
       gridSelection: deepCopy(this.gridSelection),
     };
+    this.selectedFiguresIds = [];
```
hokolomopo, line 532: wait we didn't deselect the selected figure when changing sheet before ? O_o

src/types/commands.ts
```diff
@@ -985,6 +1000,7 @@ export interface AutofillAutoCommand {
 export interface SelectFigureCommand {
   type: "SELECT_FIGURE";
   figureId: UID | null;
+  selectMultiple?: boolean;
```
hokolomopo, line 1079: maybe always giving an array of figure as argument would be clearer than a boolean selectMultiple

packages/o-spreadsheet-engine/src/plugins/ui_stateful/clipboard.ts
```diff
@@ -674,9 +675,14 @@ export class ClipboardPlugin extends UIPlugin {
...
     const sheetId = this.getters.getActiveSheetId();
-    const selectedFigureId = this.getters.getSelectedFigureId();
-    if (selectedFigureId) {
-      return { figureId: selectedFigureId, sheetId };
+    const selectedFiguresIds = this.getters.getSelectedFiguresIds();
+    const figuresUI = selectedFiguresIds
+      .map((id) => this.getters.getFigure(sheetId, id))
+      .filter(isDefined)
+      .map((f) => this.getters.getFigureUI(sheetId, f));
+    const topLeft = rectUnion(...figuresUI);
```
hokolomopo: I'd say it's the responsibility of the clipboard handlers to figure the top-left, not of the clipboard plugin

src/components/figures/figure/figure.ts
```diff
@@ -169,14 +169,26 @@ export class FigureComponent extends Component<Props, SpreadsheetChildEnv> {
...
-        this.env.model.dispatch("UPDATE_FIGURE", {
-          sheetId: this.env.model.getters.getActiveSheetId(),
-          figureId: this.props.figureUI.id,
-          offset,
-          col,
-          row,
-        });
+        const sheetId = this.env.model.getters.getActiveSheetId();
+        const figureIds = this.env.model.getters.getSelectedFiguresIds();
+        if (figureIds.length <= 1) {
```
hokolomopo: just use a MOVE_FIGURES even for a single figure, its the same thing and is shorter code no ?

src/components/figures/figure/figure.ts
```diff
@@ -169,14 +169,26 @@ export class FigureComponent extends Component<Props, SpreadsheetChildEnv> {
...
+            row,
+          });
+        } else {
+          const commands = this.env.model.getters
+            .getFigures(sheetId, figureIds)
+            .map((f) => this.env.model.getters.getFigureUI(sheetId, f))
+            .map((f) => {
+              return { sheetId, figureId: f.id, ...this.postionInBoundary(f, ev.key) };
+            });
+          this.env.model.dispatch("MOVE_FIGURES", { commands });
```
hokolomopo: If you create a store, I'd move all of the `postionInBoundary` in the store to simplify the component

src/components/figures/figure_container/figure_container.ts
```diff
@@ -36,6 +37,7 @@ interface Snap<T extends HFigureAxisType | VFigureAxisType> {
 
 interface DndState {
   draggedFigure?: FigureUI;
+  selectedFigure?: FigureUI[];
```
hokolomopo:
```suggestion
  selectedFigures?: FigureUI[];
```

src/components/figures/figure_container/figure_container.ts
```diff
@@ -144,14 +146,11 @@ export class FiguresContainer extends Component<Props, SpreadsheetChildEnv> {
 
   private getVisibleFigures(): FigureUI[] {
     const visibleFigures = this.env.model.getters.getVisibleFigures();
-    if (
-      this.dnd.draggedFigure &&
-      !visibleFigures.some((figureUI) => figureUI.id === this.dnd.draggedFigure?.id)
-    ) {
-      if (this.dnd.draggedFigure) {
-        visibleFigures.push(this.dnd.draggedFigure);
+    this.dnd.selectedFigure?.forEach((f) => {
```
hokolomopo: some people don't like `foreach` here 🙃

hokolomopo (commented):
A bunch of small issues, and added some testing notes in the pad. But the feature is very cool :D

src/plugins/ui_stateful/clipboard.ts
fdamhaut: The reason for this is that I want `getClipboardTextAndImageContent` to have matching HTML and Image content.
But I cannot add more image to the OS clipboard without a larger refactor of the clipboard to have multiple Image (which i plan to do another task for). So in the mean time, both method only copy the first image to the OS clipboard

src/plugins/ui_stateful/selection.ts
fdamhaut: We did but the file had two sequential switch case with "ACTIVATE_SHEET" being present in both of them, It come from the one i deleted (see line 176-178 of the old file)

src/types/commands.ts
fdamhaut: Idk why I though it was a coreCommand and wanted to avoid migration, It will indeed be clearer
*Edit: After testing, I can't because when selecting figure created by the copy, I need the newly created figureIds from multiple != handlers. The cleanest solution might be to refactor the selection plugin to a store but I think it's outside of this PR scope*

src/plugins/ui_feature/figure.ts
fdamhaut: In this case as there is no state and the plugin is only used to group the commands. I feel a plugin is a better choice according to [this](https://github.com/odoo/o-spreadsheet/tree/master/src/store_engine#when-to-use-a-store-and-when-to-use-a-plugin)

packages/o-spreadsheet-engine/src/plugins/ui_stateful/clipboard.ts
fdamhaut: I feel like I would rather leave it there compared to having this copied thrice in the three figure clipboard handlers (which would then need to act based on data they don't copy).

src/plugins/core/figures.ts
```diff
@@ -325,11 +325,11 @@ export class FigurePlugin extends CorePlugin<FigureState> implements FigureState
   // Getters
   // ---------------------------------------------------------------------------
 
-  getFigures(sheetId: UID): Figure[] {
+  getFigures(sheetId: UID, figureIds?: UID[]): Figure[] {
```
pro-odoo: I'm not a fan of this kind of thing, it's easier to manually filter the figures after the call to `getFigures`

packages/o-spreadsheet-engine/src/types/commands.ts
```diff
@@ -620,6 +635,12 @@ export interface PopOutChartFromCarouselCommand extends SheetDependentCommand {
   chartId: UID;
 }
 
+export interface MergeIntoCarouselCommand extends SheetDependentCommand {
+  type: "MERGE_INTO_CAROUSEL";
```
rrahir:
```suggestion
  type: "MERGE_CHART_FIGURES_INTO_CAROUSEL";
```

packages/o-spreadsheet-engine/src/plugins/ui_stateful/selection.ts
```diff
@@ -63,7 +64,7 @@ export class GridSelectionPlugin extends UIPlugin {
...
```
rrahir:
```suggestion
    "getSelectedFigureIds",
```
Don"t want to restart the whole groups_id vs group_ids debate but in Ddoo it's pretty much group_ids (except in one place, don't know why/when/are they mad?)  and we already have similar getters that follow the same pattern `getChartIds`, `getSheetIds`...

src/components/figures/figure/figure.ts
```diff
@@ -171,14 +171,15 @@ export class FigureComponent extends Component<Props, SpreadsheetChildEnv> {
...
-          sheetId: this.env.model.getters.getActiveSheetId(),
-          figureId: this.props.figureUI.id,
-          offset,
-          col,
-          row,
-        });
+        const sheetId = this.env.model.getters.getActiveSheetId();
+        const figureIds = this.env.model.getters.getSelectedFiguresIds();
+        const commands = this.env.model.getters
+          .getFigures(sheetId, figureIds)
```
rrahir: i'm not 100% convinced by the new signature of the getter, it's only used once, here, and the current implementation is such that we pass 3 times over the same array by chaining `map` . It feels like a simple loop over figureIds would do the trick

tests/figures/figures_plugin.test.ts
```diff
@@ -362,7 +485,7 @@ describe("figure plugin", () => {
     expect(result).toBeCancelledBecause(CommandResult.FigureDoesNotExist);
   });
 
-  test("can delete a figure", () => {
+  test("can delete a figure and it will reselect it", () => {
```
rrahir:
```suggestion
  test("can delete a figure and UNDO will reselect it", () => {
```

src/clipboard_handlers/carousel_clipboard.ts
```diff
@@ -8,108 +8,136 @@ import {
...
+      const maxPosition = this.getters.getMaxAnchorOffset(sheetId, height, width);
+      let { left: col, top: row } = zones[0];
+      const offset = clippedFigure.offset;
+      if (col > maxPosition.col) {
+        col = maxPosition.col;
+        offset.x = maxPosition.offset.x;
+      }
+      if (row > maxPosition.row) {
+        row = maxPosition.row;
+        offset.y = maxPosition.offset.y;
```
rrahir: this overrides the content of the clipboard, if you paste several times,it will affect the applied offset

rrahir (changes requested):

packages/o-spreadsheet-engine/src/plugins/ui_stateful/clipboard.ts
rrahir: this could be implemented in the `AbstractFigureClipboardHandler` though

packages/o-spreadsheet-engine/src/plugins/ui_feature/figure.ts
```diff
@@ -0,0 +1,68 @@
...
+import { UIPlugin } from "../ui_plugin";
+
+const uuidGenerator = new UuidGenerator();
+
+export class FigureUIPlugin extends UIPlugin {
+  allowDispatch(cmd: Command): CommandResult | CommandResult[] {
+    switch (cmd.type) {
+      case "MOVE_FIGURES":
+        for (const subCommand of cmd.commands) {
+          const result = this.allowDispatch({ type: "UPDATE_FIGURE", ...subCommand });
```
rrahir: you call yourself here, so the result will always be CommandResult.Success since you don't handle UPDATE_FIGURE ^^
```suggestion
          const result = this.canDispatch({ type: "UPDATE_FIGURE", ...subCommand });
```

packages/o-spreadsheet-engine/src/types/commands.ts
```diff
@@ -540,11 +540,26 @@ export interface UpdateFigureCommand
...
 }
 
+interface MoveFiguresSubCommand extends PositionDependentCommand {
+  figureId: UID;
+  offset: PixelPosition;
+}
+
+export interface MoveFiguresCommand {
+  type: "MOVE_FIGURES";
+  commands: MoveFiguresSubCommand[];
```
rrahir: 😨  is it really necessary to generate the commands ? couldn't the payload be deltaX, deltaY and have the plugin sort it out for each figures?

packages/o-spreadsheet-engine/src/helpers/figures/figure/figure.ts
```diff
@@ -28,3 +29,30 @@ export function getMaxFigureSize(getters: Getters, figureSize: FigureSize): Figu
   }
   return size;
 }
+
+export function getColRowOffset(
```
LucasLefevre: what's the difference between this and `getters.getColRowOffset` ?
The names should be different if they do something different

packages/o-spreadsheet-engine/src/types/clipboard.ts
```diff
@@ -33,6 +33,7 @@ export interface ClipboardOptions {
   isCutOperation: boolean;
   pasteOption?: ClipboardPasteOptions;
   selectTarget?: boolean;
+  fromSheetid?: UID;
```
LucasLefevre: never used

packages/o-spreadsheet-engine/src/types/clipboard.ts
```diff
@@ -48,21 +49,21 @@ export type ClipboardCellData = {
...
+  figureIds: UID[];
 };
 
 export type ClipboardData = ClipboardCellData | ClipboardFigureData;
 
 export type ClipboardPasteTarget = {
   sheetId: UID;
   zones: Zone[];
-  figureId?: UID;
+  figureIds?: Record<UID, UID>;
```
LucasLefevre: worth adding a comment to explain what are both UID here.

src/clipboard_handlers/image_clipboard.ts
```diff
@@ -45,46 +60,61 @@ export class ImageClipboardHandler extends AbstractFigureClipboardHandler<Clipbo
...
+      const figureId = target.figureIds[clippedFigure.figureId];
+      const sheetId = this.getters.getActiveSheetId();
+      const { width, height } = clippedFigure.copiedFigure;
+      const copy = deepCopy(clippedFigure.copiedImage);
+      const maxPosition = this.getters.getMaxAnchorOffset(sheetId, height, width);
+      let { col, row, offset } = getColRowOffset(
+        this.getters,
+        sheetId,
+        zones[0].left,
+        zones[0].top,
```
LucasLefevre: These exact 15 lines are repeated 3 times in 3 different places. Any way to factorize them?

src/components/figures/figure/figure.ts
```diff
@@ -200,41 +216,44 @@ export class FigureComponent extends Component<Props, SpreadsheetChildEnv> {
     }
   }
 
-  private postionInBoundary(position: AnchorOffset, key: string): AnchorOffset {
+  private postionInBoundary(position: AnchorOffset, key: string, shift: boolean): AnchorOffset {
     const sheetId = this.env.model.getters.getActiveSheetId();
+    const shiftAmount = shift ? 1 : 5;
```
LucasLefevre, line 218: best way ever to spam revisions: hold shift and move 🙈

src/components/helpers/figure_snap_helper.ts
```diff
@@ -30,20 +31,22 @@ interface SnapReturn {
...
+  figuresToSnap: FigureUI[],
   otherFigures: FigureUI[]
-): SnapReturn {
+): SnapReturn<FigureUI[]> {
+  const aggregate = rectUnion(...figuresToSnap);
+
   const verticalSnapLine = getSnapLine(
     getters,
-    figureToSnap,
+    aggregate,
```
LucasLefevre:
```suggestion
    aggregatedRect,
```

src/components/figures/figure_container/figure_container.ts
```diff
@@ -417,15 +463,17 @@ export class FiguresContainer extends Component<Props, SpreadsheetChildEnv> {
...
+      const otherFigures = this.getOtherFigures([figureUI.id]);
       const snapResult = snapForResize(
         this.env.model.getters,
         dirX,
         dirY,
         draggedFigure,
         otherFigures
       );
-      this.dnd.draggedFigure = snapResult.snappedFigure;
+      this.dnd.draggedFigure = snapResult.snapResult;
```
LucasLefevre: `snapResult.snapResult` or `snapReturn.snapResult` are particularly 😵‍💫 (and inconsistent)

Any way to rename `snapResult` to something that it is -> figureIds
Maybe it could always be an array of figures to remove the generic and help with the name

LucasLefevre (commented):
mainly details, the feature works great 👌

src/components/figures/figure/figure.ts
rrahir: right now, shift or no shift, you spam  :D

src/components/figures/figure_container/figure_container.ts
```diff
@@ -314,55 +340,75 @@ export class FiguresContainer extends Component<Props, SpreadsheetChildEnv> {
...
       );
+      const draggedFigure = selectedFigures.find((f) => f.id === draggedFigureId);
 
-      const otherFigures = this.getOtherFigures(initialFigure.id);
-      const overlappingCarousel = this.getCarouselOverlappingChart(draggedFigure, otherFigures);
+      let overlappingCarousel: FigureUI | undefined = undefined;
+      const otherFigures = this.getOtherFigures(selectedFigures.map((f) => f.id));
+      if (draggedFigure && !selectedFigures.find((f) => f.tag !== "chart")) {
+        overlappingCarousel = this.getCarouselOverlappingChart(draggedFigure, otherFigures);
+        this.dnd.overlappingCarousel = overlappingCarousel;
```
pro-odoo: You can remove this line, the assignation is done L359

src/types/commands.ts
```diff
@@ -568,11 +568,26 @@ export interface UpdateFigureCommand
   figureId: UID;
 }
 
+export interface MoveFiguresSubCommand extends PositionDependentCommand {
```
pro-odoo: An interface named `Command` without a `type` is unclear, let's name it differently
`MoveFiguresPayload` ?
`UpdateFigurePayload` ? (When I read handle of `MOVE_FIGURES`)

packages/o-spreadsheet-engine/src/plugins/ui_feature/figure.ts
```diff
@@ -0,0 +1,68 @@
...
+import { Command, CommandResult } from "../../types/commands";
+import { UIPlugin } from "../ui_plugin";
+
+const uuidGenerator = new UuidGenerator();
+
+export class FigureUIPlugin extends UIPlugin {
+  allowDispatch(cmd: Command): CommandResult | CommandResult[] {
+    switch (cmd.type) {
+      case "MOVE_FIGURES":
+        for (const position of cmd.figures) {
```
pro-odoo:
```suggestion
        for (const updateFigurePayload of cmd.figures) {
```
?
At least it's not a position

packages/o-spreadsheet-engine/src/plugins/ui_feature/figure.ts
```diff
@@ -0,0 +1,68 @@
...
+          return CommandResult.FigureDoesNotExist;
+        }
+    }
+    return CommandResult.Success;
+  }
+
+  handle(cmd: Command) {
+    switch (cmd.type) {
+      case "MOVE_FIGURES":
+        for (const position of cmd.figures) {
```
pro-odoo: Same as above

pro-odoo (commented):
Maybe add a test to ensure we cannot move figure in readonly mode ?

src/plugins/ui_stateful/selection.ts
```diff
@@ -241,7 +230,17 @@ export class GridSelectionPlugin extends UIPlugin {
         }
         break;
       case "SELECT_FIGURE":
-        this.selectedFigureId = cmd.figureId;
+        if (cmd.selectMultiple) {
+          if (cmd.figureId) {
```
rrahir, line 244: i think we should not reorder the figures if `cmd.figureId` is already in the list. 
I open the chart sidepanel
- select chart1 -> sidepanel switches to chart1
- shift-select chart 2 -> sidepanel switches to chart2
- shift-mousedown chart1 -> sidepanel switches to chart1
- shift-mouseup chart2 -> sidepanel switches back to chart2

this creates a glich in the sidepanel when we try to unselect a figure
```suggestion
          if (cmd.figureId & !this.selectedFigureIds.includes(cmd.figureId)) {
```

src/actions/figure_menu_actions.ts
rrahir, line 95: nitpick maybe to discuss with FRGI for later - I don't  think this action should unselect other figures, same for `edit_chart`.
I might want to check the definition of one of the selected chart to see if it's the right one to copy/move/w-h

src/clipboard_handlers/chart_clipboard.ts
```diff
@@ -7,37 +8,51 @@ import {
...
 type ClipboardContent = {
+  figures: ClipboardCharts[];
+  figureIds: UID[];
+};
+
+type ClipboardCharts = {
   figureId: UID;
   copiedFigure: Figure;
   copiedChart: AbstractChart;
+  offset: PixelPosition;
```
rrahir, line 19: this is never used at paste, which might explain why the charts are
stacked together after the paste. this comment is also valid for carousels and images

src/clipboard_handlers/carousel_clipboard.ts
```diff
@@ -8,108 +9,132 @@ import {
...
-  getPasteTarget(sheetId: UID): ClipboardPasteTarget {
-    const newId = new UuidGenerator().smallUuid();
-    return { zones: [], figureId: newId, sheetId };
+  getPasteTarget(
+    sheetId: UID,
+    target: Zone[],
+    content: ClipboardContent,
+    options?: ClipboardOptions
+  ): ClipboardPasteTarget {
+    const uuids = new UuidGenerator();
```
rrahir: nitpick, no need to create multiple generators, we could have a single one created at the root

rrahir (changes requested):

packages/o-spreadsheet-engine/src/plugins/index.ts
LucasLefevre, line 1: file reappeared after rebase

src/helpers/figures/figure/figure.ts
```diff
@@ -28,3 +29,59 @@ export function getMaxFigureSize(getters: Getters, figureSize: FigureSize): Figu
   }
   return size;
 }
+
+/*
+ * Return col, row and offset where offset is within the cell at col/row
+ * keeping the position similar
+ */
+export function boundColRowOffset(
```
LucasLefevre:
```suggestion
function boundColRowOffset(
```

src/types/commands.ts
LucasLefevre: up. `MoveFiguresPayload` ?

src/components/figures/figure_container/figure_container.ts
```diff
@@ -109,6 +113,8 @@ export class FiguresContainer extends Component<Props, SpreadsheetChildEnv> {
 
   dnd = useState<DndState>({
     draggedFigure: undefined,
+    selectedFigures: undefined,
+    selectedRect: undefined,
```
LucasLefevre, line 118: does `selectedRect` need to be part of the state ? Can it be derived ?

src/components/figures/figure_container/figure_container.ts
```diff
@@ -314,55 +350,74 @@ export class FiguresContainer extends Component<Props, SpreadsheetChildEnv> {
...
-      );
       if (!this.dnd.overlappingCarousel) {
-        this.env.model.dispatch("UPDATE_FIGURE", {
-          sheetId,
-          figureId: figureUI.id,
-          offset,
-          col,
-          row,
-        });
+        const commands =
```
LucasLefevre: these are not commands

src/components/figures/figure_container/figure_container.xml
```diff
@@ -1,14 +1,17 @@
 <templates>
   <t t-name="o-spreadsheet-FiguresContainer">
     <div class="position-absolute">
-      <t t-foreach="containers" t-as="container" t-key="container.type">
+      <t t-foreach="containers" t-as="container" t-key="container.id">
         <div
           class="o-figure-container position-absolute pe-none overflow-hidden"
           t-att-style="container.style"
           t-att-data-id="container.type + 'Container'">
```
LucasLefevre, line 8: it looks like this is the last use of ´container.type´. What's the purpose of this data id ? can `type` be removed now that there's `id` ?

src/components/helpers/figure_snap_helper.ts
```diff
@@ -18,8 +19,8 @@ export interface SnapLine<T extends HFigureAxisType | VFigureAxisType> {
   position: Pixel;
 }
 
-interface SnapReturn {
-  snappedFigure: FigureUI;
+interface SnapReturn<T> {
+  snappedFigures: T;
```
LucasLefevre: Looks like it's always `FigureUI[];` which means it doesn't need to be generic
```suggestion
  snappedFigures: FigureUI[];
```

src/components/popover/popover.ts
```diff
@@ -1,10 +1,10 @@
 import { Component, onMounted, onWillUnmount, useEffect, useRef } from "@odoo/owl";
-import { rectIntersection } from "../../helpers/rectangle";
 import { DOMCoordinates, DOMDimension, Pixel, Rect } from "../../types";
 import { PopoverPropsPosition } from "../../types/cell_popovers";
 import { CSSProperties } from "../../types/misc";
 import { SpreadsheetChildEnv } from "../../types/spreadsheet_env";
 import { usePopoverContainer, useSpreadsheetRect } from "../helpers/position_hook";
+import { rectIntersection } from "../../helpers/rectangle";
```
LucasLefevre: there are some strange import reorders in random files here and there; don't know what happened

src/helpers/viewport_collection.ts
```diff
@@ -683,10 +683,15 @@ export class ViewportCollection {
     const { scrollX, scrollY } = this.getSheetScrollInfo(sheetId);
     const x = position.x - scrollX;
     const y = position.y - scrollY;
-    const col =
+    let col =
       x >= 0 ? this.getColIndex(sheetId, x) : this.getColIndexLeftOfMainViewport(sheetId, x);
-    const row =
-      y >= 0 ? this.getRowIndex(sheetId, y) : this.getRowIndexTopOfMainViewport(sheetId, y);
+    if (col < 0) {
+      col = this.getMainInternalViewport(sheetId).searchHeaderIndex("COL", x);
```
LucasLefevre, line 690: I don't understand this.
if we are inside the `if` does it mean `col` was not found ? but then why can we search ?

src/plugins/core/figures.ts
```diff
@@ -331,11 +331,11 @@ export class FigurePlugin extends CorePlugin<FigureState> implements FigureState
...
```
LucasLefevre: `figureIds` never used

src/plugins/ui_stateful/clipboard.ts
```diff
@@ -451,6 +451,10 @@ export class ClipboardPlugin extends UIPlugin {
         zone.top
       );
     }
+    if (copiedData.figureIds) {
+      // Deselect current figure before adding new figure to selection
+      this.dispatch("SELECT_FIGURE", { figureId: null });
```
LucasLefevre, line 461: not covered by a test

src/plugins/ui_stateful/sheetview.ts
```diff
@@ -152,11 +152,13 @@ export class SheetViewPlugin extends UIPlugin {
         break;
       case "UNDO":
       case "REDO":
-        this.viewports.cleanViewports();
-        for (const sheetId of this.getters.getSheetIds()) {
-          this.sheetsWithDirtyViewports.add(sheetId);
+        if (!this.getters.getSelectedFigureIds().length) {
```
LucasLefevre: why ?

src/registries/side_panel_registry.ts
```diff
@@ -63,7 +63,10 @@ sidePanelRegistry.add("ChartPanel", {
   title: _t("Chart"),
   Body: ChartPanel,
   computeState: (getters: Getters, initialProps: { chartId: UID }) => {
-    const figureId = getters.getSelectedFigureId();
+    if (getters.getSelectedFigureIds().length > 1) {
+      return { isOpen: false };
+    }
+    const figureId = getters.getSelectedFigureIds().length && getters.getSelectedFigureIds()[0];
```
LucasLefevre: 3 calls to `getters.getSelectedFigureIds()` for the price of one 😄 
```suggestion
    const figureIds = getters.getSelectedFigureIds();
    if (figureIds.length > 1) {
      return { isOpen: false };
    }
    const figureId = figureIds[0];
```

tests/figures/figure_component.test.ts
```diff
@@ -216,6 +238,17 @@ describe("figures", () => {
     expect(fixture.querySelector(".o-figure")).toBeNull();
   });
 
+  test("Can delete multiple figure with `Backspace`", async () => {
+    createFigure(model, { id: "fig1" });
+    createFigure(model, { id: "fig2" });
+    await nextTick();
+    model.dispatch("SELECT_FIGURE", { figureId: "fig1" });
```
LucasLefevre: use `selectFigure` everywhere

LucasLefevre (commented):
👋 
Nice feature :)

src/helpers/viewport_collection.ts
fdamhaut: getColIndex only return a value if the columns is visible in the viewport, so if the position is to the right of every viewport, the previous line would set col to -1.
Search does not have that issue.

src/components/figures/figure_container/figure_container.xml
fdamhaut: I would rather remove id then to avoidhaving a useless diff that rename it and change it's type

src/plugins/ui_stateful/sheetview.ts
fdamhaut: If you redo and have selected figures, it means you have "redone" the creation of figures, in that case you don't want to jump to selected cell but would rather stay were you are (as it's most likely to be a position near the figure you are redoing)

src/plugins/ui_stateful/clipboard.ts
fdamhaut: The behaviour was already tested by `Copied figure ${type} are selected` which ensure the original figure is not selected anymore

src/plugins/ui_feature/figure.ts
```diff
@@ -0,0 +1,68 @@
...
+        for (const figureId of cmd.figureIds) {
+          if (!this.getters.getFigure(cmd.sheetId, figureId)) {
+            return CommandResult.FigureDoesNotExist;
+          }
+        }
+        break;
+      case "MERGE_CHART_FIGURES_INTO_CAROUSEL":
+        const figures = cmd.chartFigureIds.map((id) => this.getters.getFigure(cmd.sheetId, id));
+        if (figures.some((f) => f === undefined || f.tag !== "chart")) {
+          return CommandResult.FigureDoesNotExist;
```
LucasLefevre, line 28: nitpick: maybe also validate baseFigureId ?

src/components/figures/figure/figure.ts
```diff
@@ -199,41 +219,44 @@ export class FigureComponent extends Component<Props, SpreadsheetChildEnv> {
...
+          offset.x = this.env.model.getters.getColSize(sheetId, col) - shiftAmount + offset.x;
         } else {
-          offset.x--;
+          offset.x -= shiftAmount;
         }
         break;
       case "ArrowDown":
-        if (offset.y === this.env.model.getters.getRowSize(sheetId, row)) {
+        const rowSize = this.env.model.getters.getRowSize(sheetId, row);
+        if (offset.y + shiftAmount > rowSize) {
```
LucasLefevre, line 246:
```suggestion
        if (offset.y + shiftAmount >= rowSize) {
```
to keep the offset strictly inside the row and have the same behavior as before ?

src/components/figures/figure/figure.ts
```diff
@@ -199,41 +219,44 @@ export class FigureComponent extends Component<Props, SpreadsheetChildEnv> {
...
+          offset.y = offset.y + shiftAmount - rowSize;
         } else {
-          offset.y++;
+          offset.y += shiftAmount;
         }
         break;
       case "ArrowRight":
-        if (offset.x === this.env.model.getters.getColSize(sheetId, row)) {
+        const colSize = this.env.model.getters.getColSize(sheetId, col);
+        if (offset.x + shiftAmount > colSize) {
```
LucasLefevre, line 255:
```suggestion
        if (offset.x + shiftAmount >= colSize) {
```

src/plugins/ui_stateful/selection.ts
```diff
@@ -291,7 +290,15 @@ export class GridSelectionPlugin extends UIPlugin {
           this.gridSelection.anchor.zone
         );
         this.setSelectionMixin(this.gridSelection.anchor, this.gridSelection.zones);
-        this.selectedFigureId = null;
+        this.selectedFiguresIds =
+          cmd.type === "UNDO"
+            ? cmd.commands
+                .filter(
+                  (cmd): cmd is DeleteFigureCommand =>
+                    cmd.type === "DELETE_FIGURE" && cmd.sheetId === sheetId
```
LucasLefevre, line 298: shoudn't we also reselect the figure if a CREATE_FIGURE is redone ?

Create a new chart, then undo, then redo -> the chart should be selected ?

src/plugins/ui_stateful/sheetview.ts
LucasLefevre: I'm fairly sure you at least want to call `this.viewports.cleanViewports();` to avoid leaving viewports linked to dead sheets.

And I'm also quite sure we need to make sure viewports are all still in a consistent state (columns could have disappeared, ...).

We've already had crashes because of that before (hence the cleaning/reset at undo/redo)

EDIT: after some digging, it's not a real issue right now because the only way to have a selected figure at this point is if you `UNDO` a `DELETE_FIGURE` (see `selection.ts`), which doesn't change the structure of sheets/viewports (but from a theoretical pov, there could be a revision with DELETE_COLS + DELETE_FIGURE).

Connecting all the pieces and understanding why "it works" is not straight forward. Do we really need this feature ? or could it be done in another way ?

src/actions/figure_menu_actions.ts
```diff
@@ -286,11 +293,45 @@ function getDeleteMenuItem(figureId: UID, env: SpreadsheetChildEnv): ActionSpec
...
+      }
     },
     icon: "o-spreadsheet-Icon.TRASH",
   };
 }
+
+function getMergeCarouselMenuItem(figureId: UID, env: SpreadsheetChildEnv): ActionSpec {
+  return {
+    id: "mergeCarousel",
+    name: _t("Create carousel"),
```
LucasLefevre, line 316: not tested at all ?

LucasLefevre (approved):
robodoo r+

PR #7644, opened 2025-12-16
[FIX] grid: close popovers on external clicks

hokolomopo:
## Description

The filter menu/link editor popovers were not closing when clicking outside of the grid.

Note: the external click listener is in the `GridPopover` component, but this component does not have access to the HTMLElement of the popover (since it is rendered in a portal). We have thus rely on checking if the click target is inside a `.o-popover` element, which may not be the most robust solution. But it's probably acceptable, and is easy. Having an external click listener in each popover component would be more robust, but we'd risk forgetting it when creating new popovers components.

Also the external listener is on both `Grid` and `Dashboard` components, because we need the reference to the parent of the filter icon. Otherwise a click on an icon would first close the popover with the external click, and immediately re-open it with the icon click handler.

Task: [5418367](https://www.odoo.com/odoo/2328/tasks/5418367)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/dashboard/dashboard.ts
+++ b/src/components/dashboard/dashboard.ts
@@ -1,4 +1,4 @@
-import { Component, useChildSubEnv, useRef, useState } from "@odoo/owl";
+import { Component, useChildSubEnv, useExternalListener, useRef, useState } from "@odoo/owl";
 import { positionToZone } from "../../helpers/zones";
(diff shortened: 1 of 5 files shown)


pro-odoo (approved):
robodoo r+

pro-odoo:
robodoo fw=no

PR #7645, opened 2025-12-16
[FIX] evaluation: prevent empty matrix

LucasLefevre:
## Description:

The evaluation doesn't support well empty matrices (`[[]]`). A range is always at least one value and the result of a function should be at least one value.

There are plenty of uses of patterns like `array[0].length` which throws an exception if the array is empty.

This is not robust.

This commit ensures the input of a function is never an empty matrix. If a `compute` returns an empty matrix, it's transformed to an empty value.

Task: [5421196](https://www.odoo.com/odoo/2328/tasks/5421196)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


pro-odoo:
robodoo r+

PR #7667, opened 2025-12-17
[FIX] f&r: the searched range should follow the active sheet

hokolomopo:
## Description

When searching a value in a specific range in the Find & Replace panel, the range should be updated when changing the active sheet.

We actually had a test for this behavior, but the test was skipped 2 years ago during a refactoring, and never fixed.

Task: [5423885](https://www.odoo.com/odoo/2328/tasks/5423885)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


LucasLefevre:
robodoo retry

PR #7669, opened 2025-12-18
Formula results can be used as reference

laa-odoo:
Enabling the reuse of functions returning references as arguments in functions expecting references.

Task: [5428577](https://www.odoo.com/odoo/2328/tasks/5428577)

### DIFF
--- a/src/formulas/compiler.ts
+++ b/src/formulas/compiler.ts
@@ -493,7 +493,7 @@ function compileTokensOrThrow(tokens: Token[]): ICompiledFormula {
         }
         case "UNARY_OPERATION": {
           const fnName = UNARY_OPERATOR_MAP[ast.value];
-          const operand = compileAST(ast.operand, ast.value === "#").assignResultToVariable();
+          const operand = compileAST(ast.operand, ast.value === "#").assignResultToVariable(); // hasRange is true to avoid vectorization of SPILLED.RANGE
           code.append(operand);
           return code.return(`ctx['${fnName}'](${operand.returnExpression})`);
         }
--- a/src/helpers/cells/cell_evaluation.ts
+++ b/src/helpers/cells/cell_evaluation.ts
@@ -35,7 +35,7 @@ export function evaluateLiteral(
       ? literalCell.content
       : literalCell.parsedValue;
   const functionResult = { value, format: localeFormat.format, origin: position };
-  return createEvaluatedCell(functionResult, localeFormat.locale);
+  return createEvaluatedCell(functionResult, localeFormat.locale, position);
 }
 
 export function parseLiteral(content: string, locale: Locale): CellValue {
@@ -67,13 +67,13 @@ export function parseLiteral(content: string, locale: Locale): CellValue {
 export function createEvaluatedCell(
   functionResult: FunctionResultObject,
   locale: Locale = DEFAULT_LOCALE,
-  cell?: Cell,
-  origin?: CellPosition
+  position?: CellPosition,
+  cell?: Cell
 ): EvaluatedCell {
   const link = detectLink(functionResult.value);
   if (!link) {
-    const evaluateCell = _createEvaluatedCell(functionResult, locale, cell);
-    return addOrigin(evaluateCell, functionResult.origin ?? origin);
+    const evaluateCell = _createEvaluatedCell(functionResult, locale, position, cell);
+    return addOrigin(evaluateCell, functionResult.origin ?? position);
   }
   const value = parseLiteral(link.label, locale);
   const format =
@@ -87,16 +87,17 @@ export function createEvaluatedCell(
   };
   return addOrigin(
     {
-      ..._createEvaluatedCell(linkPayload, locale, cell),
+      ..._createEvaluatedCell(linkPayload, locale, position, cell),
       link,
     },
-    functionResult.origin ?? origin
+    functionResult.origin ?? position
   );
 }
 
 function _createEvaluatedCell(
   functionResult: FunctionResultObject,
   locale: Locale,
+  position?: CellPosition,
   cell?: Cell
 ): EvaluatedCell {
   let { value, format, message, errorOriginPosition } = functionResult;
@@ -104,7 +105,7 @@ function _createEvaluatedCell(
 
   const formattedValue = formatValue(value, { format, locale });
   if (isEvaluationError(value)) {
-    return errorCell(value, message, errorOriginPosition);
+    return errorCell(value, message, position, errorOriginPosition);
   }
   if (value === null) {
     return emptyCell(format);
@@ -114,47 +115,56 @@ function _createEvaluatedCell(
     // with the next line, the value of the cell is transformed depending on the format.
     // This shouldn't happen, by doing this, the formulas handling numbers are not able
     // to interpret the value as a number.
-    return textCell(toString(value), format, formattedValue);
+    return textCell(toString(value), format, formattedValue, position);
   }
 
   if (typeof value === "number") {
     if (isDateTimeFormat(format || "")) {
-      return dateTimeCell(value, format, formattedValue);
+      return dateTimeCell(value, format, formattedValue, position);
     }
-    return numberCell(value, format, formattedValue);
+    return numberCell(value, format, formattedValue, position);
   }
   if (typeof value === "boolean") {
-    return booleanCell(value, format, formattedValue);
+    return booleanCell(value, format, formattedValue, position);
   }
-  return textCell(value, format, formattedValue);
+  return textCell(value, format, formattedValue, position);
 }
 
 function textCell(
   value: string,
   format: string | undefined,
-  formattedValue: string
+  formattedValue: string,
+  position?: CellPosition
 ): EvaluatedCell {
   return {
     value,
     format,
     formattedValue,
+    position,
     type: CellValueType.text,
     isAutoSummable: true,
     defaultAlign: "left",
   };
 }
 
-function numberCell(value: number, format: string | undefined, formattedValue: string): NumberCell {
+function numberCell(
+  value: number,
+  format: string | undefined,
+  formattedValue: string,
+  position?: CellPosition
+): NumberCell {
   return {
     value: value || 0, // necessary to avoid "-0" and NaN values,
     format,
     formattedValue,
+    position,
     type: CellValueType.number,
     isAutoSummable: true,
     defaultAlign: "right",
   };
 }
 
+// do not create empty cells based on their position to allow memoization
 const emptyCell = memoize(function emptyCell(format: string | undefined): EmptyCell {
   return {
     value: null,
@@ -169,12 +179,14 @@ const emptyCell = memoize(function emptyCell(format: string | undefined): EmptyC
 function dateTimeCell(
   value: number,
   format: string | undefined,
-  formattedValue: string
+  formattedValue: string,
+  position?: CellPosition
 ): NumberCell {
   return {
     value,
     format,
     formattedValue,
+    position,
     type: CellValueType.number,
     isAutoSummable: false,
     defaultAlign: "right",
@@ -184,23 +196,31 @@ function dateTimeCell(
 function booleanCell(
   value: boolean,
   format: string | undefined,
-  formattedValue: string
+  formattedValue: string,
+  position?: CellPosition
 ): BooleanCell {
   return {
     value,
     format,
     formattedValue,
+    position,
     type: CellValueType.boolean,
     isAutoSummable: false,
     defaultAlign: "center",
   };
 }
 
-function errorCell(value: string, message?: string, errorOriginPosition?: CellPosition): ErrorCell {
+function errorCell(
+  value: string,
+  message?: string,
+  position?: CellPosition,
+  errorOriginPosition?: CellPosition
+): ErrorCell {
   return {
     value,
     formattedValue: value,
     message,
+    position,
     type: CellValueType.error,
     isAutoSummable: false,
     defaultAlign: "center",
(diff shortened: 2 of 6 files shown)


packages/o-spreadsheet-engine/src/types/misc.ts
```diff
@@ -208,9 +203,25 @@ export type Matrix<T = unknown> = T[][];
 export type FunctionResultObject = {
   value: CellValue;
   format?: Format;
-  errorOriginPosition?: CellPosition;
+  errorOriginPosition?: CellPosition /**bla */;
```
pro-odoo: 👀

pro-odoo (commented):
Don't forget to add a runbot in the task ;)

packages/o-spreadsheet-engine/src/plugins/ui_core_views/cell_evaluation/compilation_parameters.ts
```diff
@@ -67,24 +66,15 @@ class CompilationParametersBuilder {
...
     // the compiler guarantees only single cell ranges reach this part of the code
     const position = { sheetId: range.sheetId, col: range.zone.left, row: range.zone.top };
-    if (isMeta) {
-      this.computeCell(position); // ensure the cell is computed if the function using the meta parameter uses the value
-      // Use zoneToXc of zone instead of getRangeString to avoid sending unbounded ranges
-      const sheetName = this.getters.getSheetName(range.sheetId);
-      return { value: getFullReference(sheetName, zoneToXc(range.zone)) };
-    }
-    return this.computeCell(position);
+    return { ...this.computeCell(position), position };
```
LucasLefevre: This branch is significantly slower than master. Probably in part because an new object is created (and later GC'ed) every time here.

Could you consider storing `position` in `EvaluatedCell` ?

```
evaluate all cells
  master:                                                Mean: 343.44 ms, StdErr: 3.89 ms, n=20 (-8%)
  master-use-cell-position-on-FunctionResultObject-laa:  Mean: 372.15 ms, StdErr: 7.57 ms, n=20

```

laa-odoo: We obtain this performance with your suggestion (in the new commit).

```
evaluate all cells
  master:                                                Mean: 715.31 ms, StdErr: 6.21 ms, n=20 (-2%)
  master-use-cell-position-on-FunctionResultObject-laa:  Mean: 727.38 ms, StdErr: 7.91 ms, n=20

```

packages/o-spreadsheet-engine/src/functions/module_info.ts
```diff
@@ -35,45 +34,43 @@ export const CELL = {
...
+        return firstReference.format || "";
       }
       case "row":
-        return range.zone.top + 1;
+        return position.row + 1;
       case "type": {
-        const position = { sheetId: range.sheetId, col: range.zone.left, row: range.zone.top };
-        const type = this.getters.getEvaluatedCell(position).type;
-        if (type === CellValueType.empty) {
+        if (firstReference.type === "empty") {
```
anhe-odoo: Nitpick but I'd prefer a `switch` here. But feel free to ignore that

src/functions/module_lookup.ts
```diff
@@ -1019,26 +1018,29 @@ export const OFFSET = {
...
   ],
   compute: function (
-    cellReference: Matrix<{ value: string }>,
+    cellReference: Arg,
     offsetRows: Maybe<FunctionResultObject>,
     offsetColumns: Maybe<FunctionResultObject>,
     height: Maybe<FunctionResultObject>,
     width: Maybe<FunctionResultObject>
   ) {
-    if (isEvaluationError(cellReference[0][0].value)) {
```
anhe-odoo, line 1033: Image

src/functions/module_lookup.ts
```diff
@@ -1019,26 +1018,29 @@ export const OFFSET = {
...
+    const _cellReference = toMatrix(cellReference);
+    const firstCell = _cellReference[0][0];
 
-    const ref0 = cellReference[0][0].value;
-    if (!ref0) {
-      return new EvaluationError(
-        "In this context, the function OFFSET needs to have a cell or range in parameter."
-      );
+    if (firstCell === undefined) {
+      // to do: should work with ranges out of bounds
```
anhe-odoo, line 1039: Why did you change the error message here ?

src/functions/module_lookup.ts
```diff
@@ -545,20 +544,20 @@ export const ROW = {
...
-    if (cellReference[0][0] === undefined) {
+    const _cellReference = toMatrix(cellReference);
+    const firstCell = _cellReference[0][0];
+    if (firstCell === undefined) {
+      // to do: should work with ranges out of bounds
       return new EvaluationError(_t("The range is out of bounds."));
     }
-    if (cellReference[0][0].value === CellErrorType.InvalidReference) {
-      return cellReference[0][0]; // return the same error
+    if (firstCell.position === undefined) {
```
anhe-odoo, line 556: Same question here, why don't you return the original error anymore ?

packages/o-spreadsheet-engine/src/types/misc.ts
anhe-odoo: Still there :D You removed it in the wrong commit

packages/o-spreadsheet-engine/src/formulas/compiler.ts
```diff
@@ -209,18 +193,18 @@ function compileTokensOrThrow(tokens: Token[]): CompiledFormula {
         }
         case "UNARY_OPERATION": {
           const fnName = UNARY_OPERATOR_MAP[ast.value];
-          const { isMeta, hasRange } =
+          const hasRange =
             ast.value === "#"
-              ? { isMeta: true, hasRange: true } // hasRange is true to avoid vectorization of SPILLED.RANGE
-              : { isMeta: false, hasRange: false };
-          const operand = compileAST(ast.operand, isMeta, hasRange).assignResultToVariable();
+              ? true // hasRange is true to avoid vectorization of SPILLED.RANGE
```
anhe-odoo: Do we really need a ternary that can return `true` or `false` ? :D 

`const hasRange = ast.value === "#";` ?

src/functions/module_lookup.ts
laa-odoo: Here, the same error occurs (InvalidReference). But why the change?

Here, for the first time, we have the ability to access the reference without needing to query the value, a value which here represents a special case of the evaluation (meta value). So we are making this change so that we can later remove the concept of meta arguments.

src/functions/module_lookup.ts
laa-odoo: Although the code is a general way of handling errors here, it was specifically designed to handle the case where we had a reference error (because of a meta value in argsdefinition). We retain the concept here with the new code `return new InvalidReferenceError(expectReferenceError);`

src/functions/module_lookup.ts
laa-odoo: To ensure consistency with other functions:

The error message returned here was incorrect. In fact, the reference problem was handled by the more general error handling above. 

Also, error handling was missing to address this case: https://github.com/odoo/o-spreadsheet/pull/7669/changes/6697f5f7ddf3899bd27d892b1aaabb88fc6e872c#diff-f7cca7a658bccd4d5292ae885516f655ee718c7f654c68100668ebc5d8a8a639L110-L112
So I have therefore added it here.

packages/o-spreadsheet-engine/src/formulas/compiler.ts
laa-odoo: indeed xD

packages/o-spreadsheet-engine/src/functions/module_operators.ts
```diff
@@ -343,7 +350,7 @@ export const SPILLED_RANGE = {
       spilledZone.right - spilledZone.left + 1,
       spilledZone.bottom - spilledZone.top + 1,
       (col: number, row: number): FunctionResultObject =>
-        this.getters.getEvaluatedCell({
+        this.getRef({
```
LucasLefevre: `getRef` doesn't mean much. It's much less descriptive than `getEvaluatedCell`
But I don't have brilliant ideas.
It could be `this.getFormulaResult` ?

packages/o-spreadsheet-engine/src/types/misc.ts
```diff
@@ -203,9 +203,25 @@ export type Matrix<T = unknown> = T[][];
...
```
rrahir: nitpick: should probably not appear  in. the commits at all

packages/o-spreadsheet-engine/src/plugins/ui_core_views/cell_evaluation/evaluator.ts
```diff
@@ -371,7 +370,7 @@ export class Evaluator {
       e.value = e?.value || CellErrorType.GenericError;
       e.message = e?.message || implementationErrorMessage;
       e.origin = position;
-      return createEvaluatedCell(e);
+      return createEvaluatedCell(e, undefined, position);
```
LucasLefevre: there's no reason not to give it
```suggestion
      return createEvaluatedCell(e, locale, position);
```

packages/o-spreadsheet-engine/src/helpers/cells/cell_evaluation.ts
```diff
@@ -67,12 +67,13 @@ export function parseLiteral(content: string, locale: Locale): CellValue {
 export function createEvaluatedCell(
   functionResult: FunctionResultObject,
   locale: Locale = DEFAULT_LOCALE,
+  position?: CellPosition,
   cell?: Cell,
   origin?: CellPosition
```
LucasLefevre: do we still need the `origin` parameter here ? isn't it always the same as `position` ?

All tests pass with this
```diff
diff --git a/packages/o-spreadsheet-engine/src/helpers/cells/cell_evaluation.ts b/packages/o-spreadsheet-engine/src/helpers/cells/cell_evaluation.ts
index 5a5c0b034..f5f0cc903 100644
--- a/packages/o-spreadsheet-engine/src/helpers/cells/cell_evaluation.ts
+++ b/packages/o-spreadsheet-engine/src/helpers/cells/cell_evaluation.ts
@@ -74,7 +74,7 @@ export function createEvaluatedCell(
   const link = detectLink(functionResult.value);
   if (!link) {
     const evaluateCell = _createEvaluatedCell(functionResult, locale, position, cell);
-    return addOrigin(evaluateCell, functionResult.origin ?? origin);
+    return addOrigin(evaluateCell, functionResult.origin ?? position);
   }
   const value = parseLiteral(link.label, locale);
   const format =
@@ -91,7 +91,7 @@ export function createEvaluatedCell(
       ..._createEvaluatedCell(linkPayload, locale, position, cell),
       link,
     },
-    functionResult.origin ?? origin
+    functionResult.origin ?? position
   );
 }
```

packages/o-spreadsheet-engine/src/types/misc.ts
```diff
@@ -206,6 +201,22 @@ export type FunctionResultObject = {
...
+   * In such cases, we use the `position` property to store the referenced cell's
+   * position.
+   *
+   * Why do this: some formulas specifically expect cell references as input. By storing
+   * the position of the referenced cell, we can reuse, for example, the result of
+   * the previous lookup formula as input to another formula that specifically expects
+   * a reference.
+   */
+  position?: CellPosition;
+  type?: CellValueType;
```
LucasLefevre: `type` ? It is used ?
EDIT: yes, it's used. I think it shouldn't be optional -> if it's optional, we don't know how/when we can rely on its value or not.
It should be either required or removed (probably easier to remove it and adapt the two places where it's used).

src/types/misc.ts
```diff
@@ -206,6 +201,22 @@ export type FunctionResultObject = {
...
+   * value from C1:C9, but a reference to a cell in C1:C9.
+   * This is much more powerful than returning just the value.
+   * In such cases, we use the `position` property to store the referenced cell's
+   * position.
+   *
+   * Why do this: some formulas specifically expect cell references as input. By storing
+   * the position of the referenced cell, we can reuse, for example, the result of
+   * the previous lookup formula as input to another formula that specifically expects
+   * a reference.
+   */
```
LucasLefevre, line 206: I asked Claude to try simplifying the comment (make it shorter).
is it better ?
```
  /**
   * When a formula returns a reference to another cell (rather than just a value),
   * this stores the referenced cell's position.
   *
   * Example: `LOOKUP(42, B1:B9, C1:C9)` might resolve to cell C5, setting
   * `position` to C5. Formulas that specifically expect a cell reference as input
   * can then receive C5 as a reference rather than just its value
   * (e.g. `=CELL("format", LOOKUP(42, B1:B9, C1:C9))`).
   */
```

packages/o-spreadsheet-engine/src/helpers/cells/cell_evaluation.ts
laa-odoo: no no,

take this: 
```
A1: "42"
A2: "=A1"
A3: "=A2"
```

in the case `A2`:
- position will be `A1`
- origin will be `A1`

but in the case `A3`
- position will be `A2`
- origin will be `A1`

We need to distinguish between the two cases.
I'm going to add a test

laa-odoo: EDIT: 

You're right, at this scale it's not a problem

laa-odoo:
> mainly details :)

thx @LucasLefevre, 
review taken into account :)

tests/functions/module_info.test.ts
```diff
@@ -409,6 +409,8 @@ describe("NA formula", () => {
 });
 
 describe("ISFORMULA formula", () => {
+  // test
```
LucasLefevre: 👀

pro-odoo:
robodoo r+

pro-odoo:
robodoo r-

pro-odoo:
robodoo r+

PR #7674, opened 2025-12-22
[FIX] side_panel: reset format type to Automatic on unformatted cells

rmbh-odoo:
## Description:
Current behavior before PR:
- Opening the More Formats panel on a formatted cell correctly updated the selected format type.
- Changing the selection to an unformatted cell kept the previously selected format type.
- The format type did not reset to Automatic, unlike the top bar.

- Focusing a cell without any format resets the format type to Automatic.
- The panel stays in sync with the current selection and matches the top bar behavior.

Task: [5398649](https://www.odoo.com/odoo/2328/tasks/5398649)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/more_formats/more_formats_store.ts
+++ b/src/components/side_panel/more_formats/more_formats_store.ts
@@ -35,6 +35,7 @@ export class MoreFormatsStore extends SpreadsheetStore {
   ] as const;
 
   invalidFormat = false;
+  isApplyingFormatFromPanel = false;
   currentFormat = this.formatInSelection;
   category: CustomFormatCategory = this.detectFormatCategory(this.formatInSelection);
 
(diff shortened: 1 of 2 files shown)


src/components/side_panel/more_formats/more_formats_store.ts
dhrp-odoo: You need to modify this condition as well; otherwise, the same issue will occur when the side panel is opened, and we modify the format from the top bar.

src/components/side_panel/more_formats/more_formats_store.ts
```diff
@@ -35,6 +35,7 @@ export class MoreFormatsStore extends SpreadsheetStore {
   ] as const;
 
   invalidFormat = false;
+  isApplyingFormatFromPanel = false;
```
rmbh-odoo, line 38: This flag prevents the **active format** from being re-synced from the **handle()** while a format is applied from the side panel. It avoids unintended UI resets to Automatic (e.g., when switching to a custom currency temporarily sets the format to undefined) and keeps the side panel state consistent.

src/components/side_panel/more_formats/more_formats_store.ts
```diff
@@ -61,20 +62,22 @@ export class MoreFormatsStore extends SpreadsheetStore {
   }
 
   handle() {
-    const selectedFormat = this.formatInSelection;
-    if (selectedFormat && this.lastFormatInSelection !== selectedFormat) {
-      this.setActiveFormat(selectedFormat);
-      this.lastFormatInSelection = selectedFormat;
+    if (!this.isApplyingFormatFromPanel) {
```
rrahir, line 65: the condition looks superfluous here (all tests pass without it) Wasn't it only meant to counter the change of selection?

src/components/side_panel/more_formats/more_formats_store.ts
rrahir: is this particulair behaviour tested?

src/components/side_panel/more_formats/more_formats_store.ts
rmbh-odoo: When this flag is removed, around 25 test cases fail in `custom_currency_side_panel_component.test.ts`. So this behavior is already covered by existing tests, and there’s no need to add an explicit test for it. You can verify this directly in that test file.
Let me know if I’m missing anything. 😄

src/components/side_panel/more_formats/more_formats_store.ts
rmbh-odoo: :arrow_down:

src/components/side_panel/more_formats/more_formats_store.ts
rrahir: well, don't really know what/how I tested it then :o sorry for the false statement

LucasLefevre (approved):
robodoo r+
Thanks

PR #7691, opened 2025-12-24
[FIX] range: adapt string XC on sheet rename

LucasLefevre:
## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/packages/o-spreadsheet-engine/src/plugins/core/chart.ts
+++ b/packages/o-spreadsheet-engine/src/plugins/core/chart.ts
@@ -11,13 +11,7 @@ import {
   DeleteChartCommand,
   UpdateChartCommand,
 } from "../../types/commands";
-import {
-  AdaptSheetName,
-  ApplyRangeChange,
-  HeaderIndex,
-  PixelPosition,
-  UID,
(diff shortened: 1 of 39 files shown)


packages/o-spreadsheet-engine/src/plugins/core/chart.ts
```diff
@@ -48,7 +48,11 @@ export class ChartPlugin extends CorePlugin<ChartState> implements ChartState {
   private validateChartDefinition = (cmd: CreateChartCommand | UpdateChartCommand) =>
     validateChartDefinition(this, cmd.definition);
 
-  adaptRanges(applyChange: ApplyRangeChange, sheetId: UID, adaptSheetName: AdaptSheetName) {
+  adaptRanges(
```
rrahir: the methods `updateRanges` would benefit of adaptFormulaStringRanges as well, specifically the gauge chart, which do it on their own https://github.com/odoo/o-spreadsheet/blob/19.0/src/helpers/figures/charts/gauge_chart.ts#L273

PR #7693, opened 2025-12-24
[IMP] selection_stream_processor: cycle within selection on Tab/Enter

dhrp-odoo:
## Description:

When one or more zones are selected, pressing **Tab** or **Enter** moves the anchor **within the current selection**, rather than extending the selection or replacing it.

In-selection navigation:

- Tab / Shift+Tab: traverse horizontally (right/left), wrap to next row
- Enter / Shift+Enter: traverse vertically (down/up), wrap to next col
- With multiple zones, jump to the next zone entry in navigation order
- Merges are treated as a single target (visited once)

Task: [4647116](https://www.odoo.com/odoo/project/2328/tasks/4647116)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/packages/o-spreadsheet-engine/src/selection_stream/selection_stream_processor.ts
+++ b/packages/o-spreadsheet-engine/src/selection_stream/selection_stream_processor.ts
@@ -139,6 +139,21 @@ export class SelectionStreamProcessorImpl implements SelectionStreamProcessor {
   }
 
   /**
+   * Update the anchor cell while keeping the current anchor zone unchanged.
+   */
+  updateAnchorCell(
+    col: HeaderIndex,
+    row: HeaderIndex,
+    options: SelectionEventOptions = { scrollIntoView: true }
+  ): DispatchResult {
+    return this.processEvent({
+      mode: "updateAnchor",
+      anchor: { zone: this.anchor.zone, cell: { col, row } },
+      options,
+    });
+  }
+
+  /**
    * Update the current anchor such that it includes the given
    * cell position.
    */
--- a/src/components/grid/grid.ts
+++ b/src/components/grid/grid.ts
@@ -50,10 +50,10 @@ import {
   CellValueType,
   Client,
   ClipboardMIMEType,
-  Dimension,
-  Direction,
   DOMCoordinates,
   DOMDimension,
+  Dimension,
+  Direction,
   GridClickModifiers,
   HeaderIndex,
   Pixel,
@@ -73,7 +73,10 @@ import { cssPropertiesToCss } from "../helpers";
 import { getRefBoundingRect, keyboardEventToShortcutString } from "../helpers/dom_helpers";
 import { useDragAndDropBeyondTheViewport } from "../helpers/drag_and_drop_grid_hook";
 import { useGridDrawing } from "../helpers/draw_grid_hook";
-import { updateSelectionWithArrowKeys } from "../helpers/selection_helpers";
+import {
+  moveAnchorWithinSelection,
+  updateSelectionWithArrowKeys,
+} from "../helpers/selection_helpers";
 import { useTouchScroll } from "../helpers/touch_scroll_hook";
 import { useWheelHandler } from "../helpers/wheel_hook";
 import { ZoomedMouseEvent } from "../helpers/zoom";
@@ -240,19 +243,12 @@ export class Grid extends Component<Props, SpreadsheetChildEnv> {
   // this map will handle most of the actions that should happen on key down. The arrow keys are managed in the key
   // down itself
   private keyDownMapping: { [key: string]: Function } = {
-    Enter: () => {
-      const cell = this.env.model.getters.getActiveCell();
-      cell.type === CellValueType.empty
-        ? this.onComposerCellFocused()
-        : this.onComposerContentFocused();
-    },
-    Tab: () => this.env.model.selection.moveAnchorCell("right", 1),
-    "Shift+Tab": () => this.env.model.selection.moveAnchorCell("left", 1),
+    Enter: () => this.editOrMoveInSelection("down"),
+    "Shift+Enter": () => this.editOrMoveInSelection("up"),
+    Tab: () => this.moveInSelection("right"),
+    "Shift+Tab": () => this.moveInSelection("left"),
     F2: () => {
-      const cell = this.env.model.getters.getActiveCell();
-      cell.type === CellValueType.empty
-        ? this.onComposerCellFocused()
-        : this.onComposerContentFocused();
+      this.focusComposerFromActiveCell();
     },
     Delete: () => {
       this.env.model.dispatch("DELETE_UNFILTERED_CONTENT", {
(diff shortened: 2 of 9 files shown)


src/components/grid/grid.ts
```diff
@@ -477,6 +473,20 @@ export class Grid extends Component<Props, SpreadsheetChildEnv> {
...
+    const sheetId = this.env.model.getters.getActiveSheetId();
+    const zone = this.env.model.getters.getSelectedZone();
+    if (this.env.model.getters.isSingleCellOrMerge(sheetId, zone)) {
+      const cell = this.env.model.getters.getActiveCell();
+      cell.type === CellValueType.empty
+        ? this.onComposerCellFocused()
+        : this.onComposerContentFocused();
+      return;
+    }
+    const moveDirection = direction === "down" ? "down" : "up";
```
hokolomopo: moveDirection is the exact same as direction, no ?

packages/o-spreadsheet-engine/src/types/event_stream/selection_events.ts
```diff
@@ -8,6 +8,11 @@ export type SelectionEventOptions = {
 export interface SelectionEvent {
   anchor: AnchorZone;
   previousAnchor: AnchorZone;
-  mode: "newAnchor" | "overrideSelection" | "updateAnchor" | "commitSelection";
+  mode:
+    | "newAnchor"
+    | "overrideSelection"
+    | "updateAnchor"
+    | "commitSelection"
+    | "moveAnchorWithinSelection";
```
hokolomopo: I don't think `moveAnchorWithinSelection` should be a type of selection event. It being a selection event means that every handler of selection event should be aware of the event, and be able to react accordingly. But we don't really care if we moved the anchor because of a tab/enter or a click. 

The logic `moveAnchorWithinSelection` should be in either the grid/selection plugin, and it should fire updateAnchor events

packages/o-spreadsheet-engine/src/selection_stream/selection_stream_processor.ts
```diff
@@ -138,6 +143,47 @@ export class SelectionStreamProcessorImpl implements SelectionStreamProcessor {
...
   }
 
+  /**
+   * Move the anchor inside the current selection without altering selected zones.
+   *
+   * This navigates within the set of currently selected zones, attempting to move to the next
+   * navigable cell inside the current zone first, and otherwise wrapping to the next zone.
+   * Falls back to moving the anchor cell outside the selection when no in-selection target exists.
+   */
+  moveAnchorWithinSelection(direction: Direction): DispatchResult {
```
hokolomopo: As I said, this should not be in the selectio processor.

packages/o-spreadsheet-engine/src/selection_stream/selection_stream_processor.ts
```diff
@@ -138,6 +143,47 @@ export class SelectionStreamProcessorImpl implements SelectionStreamProcessor {
...
+    const start = this.getSelectionNavigationStart(plan, zones);
+    if (!start) {
+      return this.moveAnchorCell(direction, 1);
+    }
+    const nextInZone = this.getNextPositionInZone(start.position, start.zone, plan);
+    const target =
+      nextInZone !== null
+        ? { position: nextInZone, zone: start.zone }
+        : this.findNextZoneTarget(zones, start.zoneIndex, plan);
+
```
hokolomopo: I feel like the curent code is waaay too complex. It shouldn't be 500 lines of code to go the the next/previous cell in a zone. 

A simple logic should do the job I think
```js
currentZone = zone;
while(!cellInMergeButNotMainCell && !isCellHidde,){
     if(lastCellOfZone)  currentZone = nextZone;
     currentCell = getNextCell(currentCell, currentZone, direction) // simple function that does position +-1, and jump to the next col/row
}
```
the real code will be a bit more complex obviously, but really think it could be way simpler that the current implementation

hokolomopo (commented):
:wave: :wave: :wave:

packages/o-spreadsheet-engine/src/types/event_stream/selection_events.ts
dhrp-odoo: There are two reasons I introduced a dedicated `moveAnchorWithinSelection` mode:

`updateAnchor` mutates zones. `updateAnchor` replaces the zone at `previousAnchor.zone`. That breaks multi‑zone selections when you navigate inside the selection.
Example:
- Select two zones: A1:B2 and D1:E2.
- Press Tab to move from A1 into D1.
- updateAnchor replaces the previous zone → you lose the first zone. This breaks multi‑zone selection.

`moveAnchorWithinSelection` keeps zones unchanged and only moves the anchor cell. No existing mode does that.

In sheetview.ts - handleEvent, updateAnchor uses findCellInNewZone(...), which returns a boundary cell based on zone changes (not the anchor cell). 

For in‑selection navigation the zone usually doesn’t change, only the anchor does, so the viewport can scroll to the wrong place or not scroll at all.
`moveAnchorWithinSelection` explicitly scrolls to event.anchor.cell, which is the desired behavior.

Let me know if I misunderstood anything :)

packages/o-spreadsheet-engine/src/selection_stream/selection_stream_processor.ts
dhrp-odoo: FRGI agreed we can keep this single-zone only, so I simplified the implementation based on that.

I kept a minimal merge check since it was trivial. IMO, should be better now?

src/components/grid/grid.ts
```diff
@@ -424,6 +420,27 @@ export class Grid extends Component<Props, SpreadsheetChildEnv> {
...
+      : this.onComposerContentFocused();
+  }
+
+  private moveAnchor(direction: Direction) {
+    const sheetId = this.env.model.getters.getActiveSheetId();
+    const selectedZone = this.env.model.getters.getSelectedZone();
+    if (this.env.model.getters.isSingleCellOrMerge(sheetId, selectedZone)) {
+      if (direction === "up" || direction === "down") {
+        this.focusComposerFromActiveCell(); // Single cell/merge: up/down (Enter) opens the composer.
+      } else {
```
hokolomopo: nitpick: i'd put this condition in a `handleEnterKey` handler that would either open a composer or call `moveAnchor`. Having the helper named "move anchor" not always moving the anchor is a bit confusing

src/components/helpers/selection_helpers.ts
```diff
@@ -19,3 +21,78 @@ export function updateSelectionWithArrowKeys(
...
+
+export function moveAnchorWithinSelection(
+  getters: Getters,
+  selection: SelectionStreamProcessor,
+  direction: Direction
+) {
+  const {
+    anchor: { zone, cell },
+  } = getters.getSelection();
+  let currentCell = { ...cell };
```
hokolomopo:
```suggestion
  let currentPosition = { ...cell };
```

src/components/helpers/selection_helpers.ts
```diff
@@ -19,3 +21,78 @@ export function updateSelectionWithArrowKeys(
...
+  const maxChecks = getZoneArea(zone);
+  for (let i = 0; i < maxChecks; i++) {
+    currentCell = getNextCellInZone(currentCell, zone, direction);
+    if (!isNavigablePosition(getters, currentCell)) {
+      continue;
+    }
```
hokolomopo: nitpick: a loop `while(!isNaviguablePosition && isPositionInsindeZone)` would probably be easier to read

hokolomopo (commented):
It's way simpler now indeed ! 

some nitpicks, otherwise LGTM :)

src/components/grid/grid.ts
```diff
@@ -424,6 +420,35 @@ export class Grid extends Component<Props, SpreadsheetChildEnv> {
...
   };
 
+  private focusComposerFromActiveCell() {
+    const cell = this.env.model.getters.getActiveCell();
+    cell.type === CellValueType.empty
+      ? this.onComposerCellFocused()
+      : this.onComposerContentFocused();
+  }
+
+  private handleEnterKey(direction: "up" | "down") {
```
rrahir:
```suggestion
  private editOrMoveInSelection(direction: "up" | "down") {
```

src/components/grid/grid.ts
```diff
@@ -424,6 +420,35 @@ export class Grid extends Component<Props, SpreadsheetChildEnv> {
...
+    if (this.isSingleCellOrMergeSelection()) {
+      this.focusComposerFromActiveCell();
+      return;
+    }
+    moveAnchorWithinSelection(this.env.model.getters, this.env.model.selection, direction);
+  }
+
+  private handleTabKey(direction: "left" | "right") {
```
rrahir:
```suggestion
  private moveInSelection(direction: "left" | "right") {
```
or something of the sort

dhrp-odoo: make sense :+1: updated the code

rrahir:
robodoo r+

PR #7714, opened 2026-01-05
[FIX] pivot: carry forward running totals for missing date buckets

dhrp-odoo:
## Description:

When a measure is displayed as "running total" and the pivot is grouped by a date granularity (month/day/etc), PIVOT.VALUE returned an empty value for a requested bucket that does not exist in the pivot. This happened because the running-total cache only stores values for materialized pivot buckets, so a missing date key had no direct lookup.

With this commit, for date/datetime dimensions, we build an ordered index of existing buckets per secondary domain and running-total group, and use a binary search to find the nearest previous bucket. If found, we return its cumulative value (carry-forward); if the requested bucket is before the first one, we return empty.

Task: [5420999](https://www.odoo.com/odoo/2328/tasks/5420999)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/tests/pivots/pivot_measure/pivot_measure_display_model.test.ts
+++ b/tests/pivots/pivot_measure/pivot_measure_display_model.test.ts
@@ -1393,6 +1393,135 @@ describe("Measure display", () => {
         A25: "Total",      B25: "230200",  C25: "230200",  D25: "90000",  E25: "320200",  F25: "320200",  G25: "",
       });
     });
+
+    test("PIVOT.VALUE running total falls back to the previous value for a missing date bucket", () => {
+      const model = createModelWithTestPivotDataset();
+      setCellContent(model, "A40", `=PIVOT.VALUE(1, "${measureId}", "Created on:month_number", 5)`);
+      expect(getEvaluatedCell(model, "A40").value).toBe("");
+      updatePivotMeasureDisplay(model, pivotId, measureId, {
+        type: "running_total",
+        fieldNameWithGranularity: "Created on:month_number",
+      });
+      expect(getEvaluatedCell(model, "A40").value).toBe(320200);
+    });
+
+    test("PIVOT.VALUE running total supports child row and column domains", () => {
+      const model = createModelWithTestPivotDataset({
+        rows: [
+          { fieldName: "Created on", granularity: "month_number", order: "asc" },
+          { fieldName: "Stage", order: "asc" },
+        ],
+        columns: [{ fieldName: "Salesperson", order: "asc" }],
+      });
+      setCellContent(
+        model,
+        "A40",
+        `=PIVOT.VALUE(1, "${measureId}", "Created on:month_number", 4, "Stage", "New", "Salesperson", "Alice")`
+      );
+      setCellContent(
+        model,
+        "A41",
+        `=PIVOT.VALUE(1, "${measureId}", "Created on:month_number", 5, "Stage", "New", "Salesperson", "Alice")`
+      );
+
+      expect(getEvaluatedCell(model, "A40").value).toBe(49000);
+      expect(getEvaluatedCell(model, "A41").value).toBe("");
+
+      updatePivotMeasureDisplay(model, pivotId, measureId, {
+        type: "running_total",
+        fieldNameWithGranularity: "Created on:month_number",
+      });
+
+      expect(getEvaluatedCell(model, "A40").value).toBe(154600);
+      expect(getEvaluatedCell(model, "A41").value).toBe(154600);
+    });
+
+    test("PIVOT.VALUE running total fallback works with multi-level col domain", () => {
+      const model = createModelWithTestPivotDataset({
+        rows: [{ fieldName: "Created on", granularity: "month_number", order: "asc" }],
+        columns: [
+          { fieldName: "Salesperson", order: "asc" },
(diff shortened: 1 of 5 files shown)


src/helpers/pivot/pivot_presentation.ts
```diff
@@ -678,6 +693,158 @@ export default function (PivotClass: PivotUIConstructor) {
...
+        : runningTotals[previousMainDomainKey]?.[secondaryDomainKey];
+    }
+
+    private getRunningTotalOrderIndex(
+      measure: PivotMeasure,
+      fieldNameWithGranularity: string
+    ): RunningTotalOrderIndex {
+      let orderIndex = this.runningTotalOrderIndex[measure.id]?.[fieldNameWithGranularity];
+      if (!orderIndex) {
+        orderIndex = this.computeRunningTotalOrderIndex(measure.id, fieldNameWithGranularity);
```
hokolomopo: Instead of creating and maintaining this sorted orderIndex, and doing binary search in it, would it be possible to just compute the previous value on the fly by looping on all the existing value and picking the correct one ?

Obviously it will be slower than a binary search, but I don't think it's very performance-critical (there probably won't ever be a lot of dates groups, and even less PIVOT.VALUE referencing a non-existing group with running totals enabled).

It will probably simplify the code a lot no ? What do you think ?

dhrp-odoo: Yes, a linear search keeps the code much simpler. It does traverse all pivot cells, but this path is rare, so I agree to keep it simple. I’ve updated the code, please review again :)

src/helpers/pivot/pivot_presentation.ts
```diff
@@ -681,6 +691,97 @@ export default function (PivotClass: PivotUIConstructor) {
...
+      }
+
+      const secondaryDomainKey = domainToString(secondaryDomain);
+      const runningTotalKey = getRunningTotalDomainKey(mainDomain, fieldNameWithGranularity);
+      let previousMainDomainKey: string | undefined;
+      let previousValue: CellValue | undefined;
+
+      const pivotCells = this.getPivotValueCells(measure.id);
+      for (const col of pivotCells) {
+        for (const cell of col) {
```
hokolomopo: We probably can get all the relevant domains w/o looping on all the cells of the pivot, no ? Probably by looking in the `colTree/rowTree` ?

dhrp-odoo: Yes, I updated the code to use the row/col tree. Is this what you had in mind?

tests/pivots/pivot_measure/pivot_measure_display_model.test.ts
```diff
@@ -1389,6 +1389,43 @@ describe("Measure display", () => {
...
       });
     });
+
+    test("PIVOT.VALUE running total falls back to the previous value for a missing date bucket", () => {
+      const model = createModelWithTestPivotDataset();
+      updatePivotMeasureDisplay(model, pivotId, measureId, {
+        type: "running_total",
+        fieldNameWithGranularity: "Created on:month_number",
+      });
+      setCellContent(model, "A40", `=PIVOT.VALUE(1, "${measureId}", "Created on:month_number", 5)`);
```
LucasLefevre: you can add the formula before and show the empty value. It makes the test slightly easier to read to understand before/after the "display as"
```ts
      const model = createModelWithTestPivotDataset();
      setCellContent(model, "A40", `=PIVOT.VALUE(1, "${measureId}", "Created on:month_number", 5)`);
      expect(getEvaluatedCell(model, "A40").value).toBe("");
      updatePivotMeasureDisplay(model, pivotId, measureId, {
        type: "running_total",
        fieldNameWithGranularity: "Created on:month_number",
      });
```

LucasLefevre: could you add basic tests where `Created on` is not the only part of the domain in the formula ?
with parent/child domain, and col domain

src/helpers/pivot/pivot_presentation.ts
```diff
@@ -681,6 +690,98 @@ export default function (PivotClass: PivotUIConstructor) {
...
+        return 0;
+      }
+      if (aIsNull) {
+        return order === "asc" ? 1 : -1;
+      }
+      if (bIsNull) {
+        return order === "asc" ? -1 : 1;
+      }
+
+      const diff = toNumber(a, DEFAULT_LOCALE) - toNumber(b, DEFAULT_LOCALE);
```
LucasLefevre: Hmm 🤔 
In odoo, there's the "Week & Year" granularity. And the values look like `"48/2026"` for week 48.
But `"48/2026"` cannot be coersed to a number 

Screenshot from your branch:

Image

Also, for quarters (in odoo), it works "by chance" because `"2/2025"` can be interpreted (as a month!) and March (Q3 `"3/2025"`) is after February (and the max quarter number is <12).

The only thing I can think of right now that wouldn't be too dirty is to add a new "compare" thing in `pivotTimeAdapterRegistry`
But if you have other ideas, there're welcome...

LucasLefevre: if the diff starts to grow and you don't want to go through fw-ports, feel free to retarget to master :wink:

PR #7717, opened 2026-01-06
[FIX] session: recover from errors when sending messages

VincentSchippefilt:
If, for any reason, the sending of a revision creates an error b.e. when the user is disconnected, we never reset the "waitingForAck" flag, and all future messages will be queued but never send. The result of this behavior was that any user action during any disconnection will set the spreadsheet in a state where all changes from that point are lost.

This fix handles the error while sending a message and always reset the flag, so that at next change, we will try to re-send the same message until it works.

Task: 5469116

### DIFF
--- a/tests/__mocks__/transport_service_async.ts
+++ b/tests/__mocks__/transport_service_async.ts
@@ -1,0 +1,12 @@
+import { CollaborationMessage } from "../../src/types/collaborative/transport_service";
+import { MockTransportService } from "./transport_service";
+
+export class MockTransportServiceAsync extends MockTransportService {
+  async notifyListeners(message: CollaborationMessage) {
+    return Promise.all(
+      this.listeners.map(({ callback }) => {
+        return Promise.resolve().then((x) => callback(message));
+      })
+    );
+  }
+}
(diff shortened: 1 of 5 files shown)


tests/__mocks__/transport_service_async.ts
```diff
@@ -0,0 +1,53 @@
...
+import {
+  CollaborationMessage,
+  NewMessageCallback,
+} from "../../src/types/collaborative/transport_service";
+import { MockTransportService } from "./transport_service";
+
+export class MockTransportServiceAsync extends MockTransportService {
+  onNewMessage(id: UID, callback: NewMessageCallback) {
+    this.listeners.push({ id, callback });
+  }
```
LucasLefevre: not needed. Same implementation from its parent

tests/__mocks__/transport_service_async.ts
```diff
@@ -0,0 +1,53 @@
...
+  async sendMessage(message: CollaborationMessage) {
```
LucasLefevre: not needed either

tests/__mocks__/transport_service_async.ts
```diff
@@ -0,0 +1,53 @@
...
+          this.snapshot = msg.data;
+        }
+        break;
+      default:
+        this.notifyListeners(msg);
+        break;
+    }
+  }
+
+  leave(id: UID) {
```
LucasLefevre: same

LucasLefevre (commented):
only the override of `notifyListeners` is needed

PR #7722, opened 2026-01-07
[IMP] named ranges: add support of xlsx import/export

hokolomopo:
### [IMP] named ranges: add support of xlsx import/export

This commit introduces the support for named ranges in the import and
export of XLSX files.

Task: [5462359](https://www.odoo.com/odoo/2328/tasks/5462359)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/xlsx/conversion/table_conversion.ts
```diff
@@ -191,6 +194,33 @@ function deconstructSheets(convertedSheets: SheetData[]): DeconstructedSheets {
   return deconstructedSheets;
 }
 
+function convertTableFormulaReferencesInNamedRanges(
```
rrahir, line 197: seems a bit weird to do this in the table convertion and not have it done when converting the named range. it knid of implies that the information disappears after the table convertion and tthat the conversions need to be executed in a precise order not to break. maybe unavoidable though :'(

rrahir (commented):
👋  Chtit rebase et chtite question

src/xlsx/conversion/table_conversion.ts
hokolomopo: Wee it's already the conversion of table that change the cell content, by your logic it should be the cell conversion that does that. Should it all be done here, or should all the different parts of the conversion worry about table references ? I' think I prefer to have it centralized, but I don't really care :man_shrugging: 

(And now that I think about it, we need to check if excel allows table references in CF/DV/charts/..., because we don't handle that at all)

rrahir:
robodoo r+

PR #7732, opened 2026-01-08
[FIX] helpers: export chart check type from runtime config

rmbh-odoo:
## Description:

Current behavior before PR:
- Chart image generation could fail when the provided chart type was
  not a valid Chart.js controller (e.g. combo, waterfall, geo, ...).
- Helpers previously relied on given chart types, which do not always
  map to generic Chart.js types, causing image export to return undefined.

- The export logic now uses the runtime config type, which resolves to a
  valid Chart.js controller and fixes image and XLSX exports for all
  chart types.

Task: [5446988](https://www.odoo.com/odoo/2328/tasks/5446988)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/packages/o-spreadsheet-engine/src/helpers/figures/charts/chart_ui_common.ts
+++ b/packages/o-spreadsheet-engine/src/helpers/figures/charts/chart_ui_common.ts
@@ -48,16 +48,16 @@ export async function chartToImageUrl(
     if (!extensionsLoaded) {
       registerChartJSExtensions();
     }
-    if (!globalThis.Chart.registry.controllers.get(type)) {
-      console.log(`Chart of type "${type}" is not registered in Chart.js library.`);
+
+    const config = deepCopy(runtime.chartJsConfig);
+    config.plugins = [backgroundColorChartJSPlugin];
+    if (!globalThis.Chart.registry.controllers.get(config.type)) {
+      console.log(`Chart of type "${config.type}" is not registered in Chart.js library.`);
       if (!extensionsLoaded) {
         unregisterChartJsExtensions();
       }
       return imageUrl;
     }
-
-    const config = deepCopy(runtime.chartJsConfig);
-    config.plugins = [backgroundColorChartJSPlugin];
 
     const chart = new globalThis.Chart(
       canvas as unknown as HTMLCanvasElement,
@@ -106,16 +106,16 @@ export async function chartToImageFile(
     if (!extensionsLoaded) {
       registerChartJSExtensions();
     }
-    if (!globalThis.Chart.registry.controllers.get(type)) {
-      console.log(`Chart of type "${type}" is not registered in Chart.js library.`);
+
+    const config = deepCopy(runtime.chartJsConfig);
+    config.plugins = [backgroundColorChartJSPlugin];
+    if (!globalThis.Chart.registry.controllers.get(config.type)) {
+      console.log(`Chart of type "${config.type}" is not registered in Chart.js library.`);
       if (!extensionsLoaded) {
         unregisterChartJsExtensions();
       }
       return chartBlob;
     }
-
-    const config = deepCopy(runtime.chartJsConfig);
-    config.plugins = [backgroundColorChartJSPlugin];
 
     const chart = new globalThis.Chart(
       canvas as unknown as HTMLCanvasElement,
--- a/tests/figures/chart/chart_conversion.test.ts
+++ b/tests/figures/chart/chart_conversion.test.ts
@@ -55,22 +55,57 @@ describe("chart conversion", () => {
     expect(spy).toHaveBeenCalledWith("Chart.js library is not loaded");
   });
 
-  test("cannot export if the chart type is not supported", async () => {
-    globalThis.Chart!.registry.controllers.get = (controllerName: string) => undefined;
-    const spy = jest.spyOn(console, "log").mockImplementation(() => {});
-
+  test("export Chart.js chart check type from runtime config, not the given chart type", async () => {
     const model = new Model();
     const sheetId = model.getters.getActiveSheetId();
-    createChart(model, TEST_CHART_DATA.basicChart, chartId, undefined, { figureId });
+
+    createChart(model, TEST_CHART_DATA.combo, chartId, undefined, { figureId });
     const figure = model.getters.getFigure(sheetId, figureId)!;
-    const runtime = model.getters.getChartRuntime(chartId);
+    const runtime = model.getters.getChartRuntime(chartId) as any;
+    expect(runtime.chartJsConfig.type).toBe("bar");
 
-    const imageUrl = await chartToImageUrl(runtime, figure, "bar");
+    const registrySpy = jest.spyOn(globalThis.Chart!.registry.controllers, "get");
+
+    const imageUrl = await chartToImageUrl(runtime, figure, "combo");
+    expect(imageUrl).toBe("data:image/png;base64,");
+    expect(registrySpy).toHaveBeenCalledWith("bar");
+    expect(registrySpy).not.toHaveBeenCalledWith("combo");
+
+    const imageFile = await chartToImageFile(runtime, figure, "combo");
+    expect(imageFile).toBeInstanceOf(Blob);
+    expect(registrySpy).toHaveBeenCalledWith("bar");
+    expect(registrySpy).not.toHaveBeenCalledWith("combo");
+  });
+
+  test("cannot export if the chart runtime config type is not supported", async () => {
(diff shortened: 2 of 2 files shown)


packages/o-spreadsheet-engine/src/helpers/figures/charts/chart_ui_common.ts
```diff
@@ -48,17 +48,17 @@ export async function chartToImageUrl(
     if (!extensionsLoaded) {
       registerChartJSExtensions();
     }
-    if (!globalThis.Chart.registry.controllers.get(type)) {
-      console.log(`Chart of type "${type}" is not registered in Chart.js library.`);
+
+    const config = deepCopy(runtime.chartJsConfig);
+    config.plugins = [backgroundColorChartJSPlugin];
+    if (!globalThis.Chart.registry.controllers.get(config.type)) {
+      console.log(`Chart of type "${config.type}" is not registered in Chart.js library.`);
```
rmbh-odoo, line 55: The Husky pre-commit check failed because of a console statement in my changes.
I’ve temporarily ignored it since console logs already exist here.
Let me know if you prefer removing the console logs; otherwise, we can move forward with this as is. 😄

tests/figures/chart/chart_conversion.test.ts
```diff
@@ -55,22 +55,57 @@ describe("chart conversion", () => {
...
+    const sheetId = model.getters.getActiveSheetId();
+
+    createChart(model, TEST_CHART_DATA.basicChart, chartId, undefined, { figureId });
+    const figure = model.getters.getFigure(sheetId, figureId)!;
+    const runtime = model.getters.getChartRuntime(chartId) as any;
+    expect(runtime.chartJsConfig.type).toBe("bar");
+
+    const registrySpy = jest.spyOn(globalThis.Chart!.registry.controllers, "get");
+
+    const imageUrl = await chartToImageUrl(runtime, figure, "odoo_bar" as any);
```
hokolomopo: We shouldn't test thing related to odoo chart explicitly in o-spreadsheet. The bug was also present for waterfall chart, you can test with those instead

tests/figures/chart/chart_conversion.test.ts
```diff
@@ -55,22 +55,57 @@ describe("chart conversion", () => {
...
     const sheetId = model.getters.getActiveSheetId();
     createChart(model, TEST_CHART_DATA.basicChart, chartId, undefined, { figureId });
     const figure = model.getters.getFigure(sheetId, figureId)!;
-    const runtime = model.getters.getChartRuntime(chartId);
+    const originalRuntime = model.getters.getChartRuntime(chartId) as any;
+    const runtime = {
+      ...originalRuntime,
+      chartJsConfig: {
+        ...originalRuntime.chartJsConfig,
+        type: "odoo_bar",
```
hokolomopo: there's no need to change this test, no ?

packages/o-spreadsheet-engine/src/helpers/figures/charts/chart_ui_common.ts
```diff
@@ -106,17 +106,17 @@ export async function chartToImageFile(
     if (!extensionsLoaded) {
       registerChartJSExtensions();
     }
-    if (!globalThis.Chart.registry.controllers.get(type)) {
-      console.log(`Chart of type "${type}" is not registered in Chart.js library.`);
+
+    const config = deepCopy(runtime.chartJsConfig);
+    config.plugins = [backgroundColorChartJSPlugin];
```
hokolomopo, line 111: that's not something you changed there, but why do we remove all the plugins ? Don't we want the title ? The show values ? I'm confused

packages/o-spreadsheet-engine/src/helpers/figures/charts/chart_ui_common.ts
```diff
@@ -106,17 +106,17 @@ export async function chartToImageFile(
...
+    if (!globalThis.Chart.registry.controllers.get(config.type)) {
```
hokolomopo, line 112: while you're in there you could ut the condition at the top of the functions, there's no need to register then unregister the extensions in that case

packages/o-spreadsheet-engine/src/helpers/figures/charts/chart_ui_common.ts
rmbh-odoo: Hi, this was done by Anhe in SaaS-16.4 around 2 years ago, with an MD file added for reference. https://github.com/odoo/o-spreadsheet/pull/2968/changes/274877480ffe9ab54d6bbd1d985825118fc4b2a7 
Title and show values still appear in the image as expected, I guess, since they’re already rendered in the canvas. 🤔 

[docs/configuration/canvas-background.md](https://github.com/chartjs/Chart.js/blob/8fdf76f8f02d31684d34704341a5d9217e977491/docs/configuration/canvas-background.md)

tests/figures/chart/chart_conversion.test.ts
rmbh-odoo: We need to change `runtime.chartJsConfig.type` because this test specifically targets a non-Chart.js type.
Since we no longer depend on an explicitly provided type, the previous test case no longer has meaningful coverage.

packages/o-spreadsheet-engine/src/helpers/figures/charts/chart_ui_common.ts
```diff
@@ -44,21 +44,16 @@ export async function chartToImageUrl(
       console.log("Chart.js library is not loaded");
       return imageUrl;
     }
+    const config = deepCopy(runtime.chartJsConfig);
+    config.plugins = [backgroundColorChartJSPlugin];
+    if (!globalThis.Chart.registry.controllers.get(config.type)) {
+      console.log(`Chart of type "${config.type}" is not registered in Chart.js library.`);
```
rrahir, line 55: if you return before  the extensions are loaded, you miss some controllers, like for the funnel chart for instance.

rrahir:
Also, the commit message is quite misleading, it extends to more than waterfall charts, geo charts are also impacted !
It's more about the conversion of chartJs types of chart to images that is an issue.

packages/o-spreadsheet-engine/src/helpers/figures/charts/chart_ui_common.ts
hokolomopo: yeah you're right, my bad I was wrong on my previous comment. @rmbh-odoo this whole function is strange, just go back to your previous fix w/o changing the function too much, I'll have a look to see what happen there in another task :)

rmbh-odoo: Yeah, I was worried about that as well. I’ve fixed it now and tested it with combo charts.

@rrahir  From what I can see, the extensions are already loaded when I try this. Could you explain in which situations the extensions might not be loaded yet? 
I’m asking for my general understanding. :zany_face:

tests/figures/chart/chart_conversion.test.ts
```diff
@@ -55,22 +55,57 @@ describe("chart conversion", () => {
...
+    expect(registrySpy).toHaveBeenCalledWith("bar");
+    expect(registrySpy).not.toHaveBeenCalledWith("combo");
+
+    const imageFile = await chartToImageFile(runtime, figure, "combo");
+    expect(imageFile).toBeInstanceOf(Blob);
+    expect(registrySpy).toHaveBeenCalledWith("bar");
+    expect(registrySpy).not.toHaveBeenCalledWith("combo");
+  });
+
+  test("cannot export if the chart runtime config type is not supported", async () => {
```
rmbh-odoo, line 80: I’m not sure we have tests like this, but I added it for better coverage 🥲
Let me know if it’s not needed.

packages/o-spreadsheet-engine/src/helpers/figures/charts/chart_ui_common.ts
rrahir: sure :) the trick is to open a spreadsheet without loading any `ChartJsComponent` since we lazy load the extensions when we create an instance of that component. If you open the demo spreadsheet and try to export it to excel without changing sheets, the extensions are never loaded

tests/figures/chart/chart_conversion.test.ts
```diff
@@ -55,22 +55,57 @@ describe("chart conversion", () => {
...
+    globalThis.Chart!.registry.controllers.get = (type: string) => {
```
rrahir: I just amended that line to use spyOn as well. This way, the method is not overriden for subsequent tests

rrahir:
robodoo r+

PR #7733, opened 2026-01-08
[REF] lint: enforce braces for all control statements

LucasLefevre:
patterns like `if (foo) foo++;`, `if (foo) continue;` or `if (foo) return;` are less readable then the version with braces.

`continue` or `return` are easy to miss when it's too far right on the line after a long `if (...a long condition...)`

It also improves code style consistency across the codebase, which always improves readability.

I changed the order of eslint and prettier in the precommit hook. because eslint rewrites `if (foo) return;` to `if (foo) {return;}`, which then needs to be prettified.

> [!TIP]
> **The pre-commit hook automatically re-writes the pattern**. FLDA, you can still write one-liner :wink: 

I'm sorry FLDA and ADRM (but not really sorry)

Task: 0

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -15,6 +15,7 @@ export default tseslint.config(
       "no-unsafe-optional-chaining": "error", // ban unsafe optional chaining
       eqeqeq: "error", // ban non-strict equal
       "@typescript-eslint/no-non-null-asserted-optional-chain": "error", // ban non-null assertion in optional chain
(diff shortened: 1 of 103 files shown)


pro-odoo:
robodoo r+

PR #7745, opened 2026-01-09
[FIX] *: fix Safari detection

rrahir:
The userAgent on Chrome-based browser looks something like `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/143.0.0.0 Safari/537.36` and somehow refers to Safari.
Our helper to detect safari (more specifically Webkit-based) browsers was not able to make that distinction.

MDN does not recommend to rely on the userAgent for browser detection so
we propose to directly test the behaviour that was problematic with
safari: detect if the zoom css property has an impact on the computation
of an element bounding client rect.

Task: 5453400

## Description:

Task: [5453400](https://www.odoo.com/odoo/2328/tasks/5453400)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/helpers/dom_helpers.ts
+++ b/src/components/helpers/dom_helpers.ts
@@ -3,6 +3,18 @@ import { DOMCoordinates, Rect } from "@odoo/o-spreadsheet-engine/types/rendering
 
 const macRegex = /Mac/i;
 
+let zoomCssDoesNotAffectBoundingRect = false;
+
+function defineZoomCssImpactOnBoundingRect() {
+  const div = document.createElement("div");
+  div.setAttribute("style", `width:10px;height:1px;zoom:2;position:absolute;z-index:-10000`);
+  document.body.appendChild(div);
+  zoomCssDoesNotAffectBoundingRect = div.getBoundingClientRect().width !== 20;
+  document.body.removeChild(div);
+}
+
+defineZoomCssImpactOnBoundingRect();
+
 const MODIFIER_KEYS = ["Shift", "Control", "Alt", "Meta"];
 
 /**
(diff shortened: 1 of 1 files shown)


rrahir:
robodoo help

rrahir:
robodoo fw=no

src/components/helpers/dom_helpers.ts
```diff
@@ -238,14 +238,18 @@ export function downloadFile(dataUrl: string, fileName: string) {
 }
```
hokolomopo, line 250: mdn seems to say that parsing to check the useAgent is a very bad idea https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Browser_detection_using_the_user_agent .

Any way to do feature detection in place of the `isBrowserSafari` ? IIRC that was because of a wrong boundingCleintRect, it might somehow be testable

rrahir: agreed it's not the best but since it's a backport :/ I guess we could try to generate a div, apply a zoom and check if the geboundingrect is coherent ...

rrahir:
robodoo fw=default

rrahir:
@hokolomopo check out the latest commit, it's less Safari based :D

src/components/helpers/dom_helpers.ts
```diff
@@ -3,6 +3,19 @@ import { DOMCoordinates, Rect } from "@odoo/o-spreadsheet-engine/types/rendering
 
 const macRegex = /Mac/i;
 
+let zoomCssDoesNotAffectBoundingRect = false;
+
+function defineZoomCssImpactOnBoundingRect() {
+  const div = document.createElement("div");
+  div.setAttribute("style", `width:10px;height:1px;zoom:2;position:absolute;z-index:-10000`);
+  div.dataset.test = "safariDetection";
```
LucasLefevre: is this needed ?

src/components/helpers/dom_helpers.ts
```diff
@@ -31,11 +44,11 @@ export function zoomCorrectedElementRect(el: Element, zoomLevel: number): Rect {
   if (zoomedElement) {
     targetEl = zoomedElement;
     // FIXME
-    // Safari messes up the computation of getBoundingClientRect on elements subjected to a zoom
+    // Some navigators, like Safari, mess up the computation of getBoundingClientRect on elements subjected to a zoom CSS property.
```
LucasLefevre, line 46:
```suggestion
    // Some browsers, like Safari, mess up the computation of getBoundingClientRect on elements subjected to a zoom CSS property.
```

src/components/helpers/dom_helpers.ts
```diff
@@ -238,16 +251,12 @@ export function downloadFile(dataUrl: string, fileName: string) {
 }
 
 /**
- * Detects if the current browser is Firefox
+ * Detects the current browser brand and subsequent rendering engine
```
LucasLefevre, line 253: 🤔

PR #7752, opened 2026-01-12
[IMP] components: remove native `<select>` inputs

hokolomopo:
## Description

The native `<select>` inputs have multiple issues:

- Since their dropdown is rendered by the OS/browser, we have no control over its style and it can look wildly different that the rest of our UI.
- We also cannot control its size, which can make it even uglier.
- And on some browsers/OS, clicking on a `<select>` doesn't even trigger a `click` event(!?), which breaks our `onClick` external listeners.

This commit transforms all native `<select>` inputs into our custom `<Select>` component, that has a style consistent with the rest of our dropdown, and that we fully control.

Task: [5213725](https://www.odoo.com/odoo/2328/tasks/5213725)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/components/select/select.xml
```diff
@@ -0,0 +1,30 @@
...
+      <option selected="true" t-att-value="props.selectedValue" t-esc="selectedLabel"/>
+    </select>
+
+    <Popover t-if="state.isPopoverOpen" t-props="popoverProps">
+      <div
+        class="o-select-dropdown bg-white overflow-hidden py-2"
+        t-att-class="props.popoverClass"
+        t-ref="dropdownRef">
+        <t t-foreach="props.values" t-as="value" t-key="value_index">
+          <option
```
rrahir: Idon't really understand the use of option here since you are out of a <select> parent, don't think it's recommended.
 
 EDIT: yeah safari says no
 
image

src/components/select/select.xml
```diff
@@ -0,0 +1,30 @@
+<templates>
+  <t t-name="o-spreadsheet-Select">
+    <select
```
rrahir: I would go with something like a div, something that can capture the focus so that I can use my arrows to switch between the different values. I think the implementation in Odoo is not that bad, it also allows to write down some text and filters out the possible values based on that not mandantory but having the possibility to navigat the values via keyboard (arrows and tab) seems essential to me.

src/components/side_panel/chart/building_blocks/series_design/series_with_axis_design_editor.ts
```diff
@@ -113,13 +115,25 @@ export class SeriesWithAxisDesignEditor extends Component<Props, SpreadsheetChil
     this.updateTrendLineValue(index, config);
   }
 
-  getPolynomialDegrees(index: number): number[] {
-    return range(1, this.getMaxPolynomialDegree(index) + 1);
+  get trendOptions(): ValueAndLabel[] {
+    return [
+      { value: "linear", label: "Linear" },
```
rrahir:
```suggestion
      { value: "linear", label: _t("linear") },
```
etc

pro-odoo:
robodoo r+
(push-forced to resolve conflict)

PR #7756, opened 2026-01-13
[IMP] export: reduce JSON size

VincentSchippefilt:
## Description:

Reduce the JSON file size of spreadsheets by storing only changes to formulae and, in a second pass,
collecting all the formulae or changes that are the same and storing them under one single key in the JSON file.

Task: [5489478](https://www.odoo.com/odoo/2328/tasks/5489478)

## API changes:
compiler.ts/`compile(formula: string)` --> `CompiledFormula.Compile(formula: string, sheetId: UID, getters: CoreGetters)`
fomulas/helpers.ts/`getFunctionsFromTokens(tokens: Tokens[], functionNames: string[])` --> fomulas/helpers.ts/`getFunctionsFromTokens(compiledFormula: CompiledFormula, functionNames: string[])`
removed `RangeCompiledFormula` --> `CompiledFormula`
model.ts/`export()` --> `export(shouldSquish?: boolean = false)`
cell.ts/`getTranslatedFormulaCell(sheetId: UID, offsetX: number, offsetY: number, tokens: Token[])` --> `getTranslatedFormulaCell(sheetId: UID,offsetX: number,offsetY: number,compiledFormula: CompiledFormula | SerializedCompiledFormula )`
removed plugins/core/cell.ts/`ReferenceToken` class --> it is implemented in CompiledFormula`
formulas/helper.ts/`isExportableToExcel` & `getFunctionsFromTokens` --> CompiledFormula.`areAllFunctionsExportableToExcel` & `getFunctionsFromTokens`
removed FormulaCellWithDependencies and ReferenceToken

## Observed performance gains on large dashboard:
importing sheets: 
~2300ms --> ~680 ms
1560 MB ram --> 580 MB

JSON Size:
40 MB -> 147 KB
4,4 MB -> 3KB
1,6 MB-> 43 KB

### Shared formula
```cells: {
   A1: "My Text",
   A2: "=SUM(B2:B9)",
   A3: "=SUM(B2:B9)",
```
As we can see, A2 and A3 share the same formula. They can be rewritten as
```cells: {
   A1: "My Text",
   'A2:A3': "=SUM(B2:B9)",
```
removing duplication completely

### Formula with small differences
Usually formula do not repeat exactly the same, but offset their dependencies (being either references to other cells, numbers or strings) slightly. We can rewrite:
```cells: {
   A1: "My Text",
   A2: "=CONCAT(B2, $C$2, D2, "hello"),
   A3: "=CONCAT(B3, $C$2, D3, "hello"),
   A4: "=CONCAT(B4, $C$2, D4, "hello"),
```
to 
```cells: {
   A1: "My Text",
   A2: "=CONCAT(B2, $C$2, D2, "hello"),
   A3: { R : "+R1|=|+R1" },
   A4: { R : "+R1|=|+R1" },
   ...
```
now we see that A3 and A4 have the same transformation, so we can rewrite them to 
```cells: {
   A1: "My Text",
   A2: "=CONCAT(B2, $C$2, D2, "hello"),
   "A3:A4": { R : "+R1|=|+R1" },
```
We apply the same for number arguments and string arguments. 
A formula with all the arguments slightly changed might look like
```cells: {
   A1: "=MY_FORMULA(1, B1, "coucou"),
   A2: { R : "+R1", N: "+1", S: ["hello"] },
   A3:A4: { R : "+R1", N: "+1"}
```
can be read as 
```cells: {
   A1: "=MY_FORMULA(1, B1, "coucou"),
   A2: "=MY_FORMULA(2, B2, "hello"),
   A3: "=MY_FORMULA(3, B3, "hello"),
   A4: "=MY_FORMULA(4, B4, "hello"),
```

When reading the compressed JSON, the order of the keys in the JSON itself do not matter, they are sorted on the first part of the key, so either the cell reference or the left part of the range.

### DIFF
--- a/packages/o-spreadsheet-engine/src/formulas/compiler.ts
+++ b/packages/o-spreadsheet-engine/src/formulas/compiler.ts
@@ -56,23 +56,29 @@ const NO_REAL_VALUE = "__NO_REAL_VALUE__";
 // It is only exported for testing purposes
 export const functionCache: { [key: string]: FormulaToExecute } = {};
 
-export class CompiledFormula implements Omit<Omit<ICompiledFormula, "tokens">, "dependencies"> {
+/**
+ * A compiled formula is the result of the compilation of a formula string.
+ * It contains all the information needed to execute the formula, as well as some metadata
+ * about the formula (dependencies, literal values, symbols...) that can be used to rebuild a slightly different formula
+ * without recompiling it (for example when the formula is copied to another cell, or when we want to replace literal values but keep the same structure).
+ * */
+export class CompiledFormula implements Omit<ICompiledFormula, "tokens" | "dependencies"> {
   public readonly rangeDependencies: Range[];
   public hasDependencies: boolean;
 
   private constructor(
-    public sheetId: UID,
+    public readonly sheetId: UID,
     private readonly tokens: Token[],
-    public literalValues: LiteralValues,
-    public symbols: string[],
+    public readonly literalValues: LiteralValues,
+    public readonly symbols: string[],
     dependencies: Range[],
-    public isBadExpression: boolean,
-    public normalizedFormula: string,
-    public execute: FormulaToExecute
+    public readonly isBadExpression: boolean,
+    public readonly normalizedFormula: string,
+    public readonly execute: FormulaToExecute
   ) {
     this.hasDependencies = dependencies?.length > 0;
     this.tokens.forEach((t) => {
-      if (["REFERENCE", "NUMBER", "STRING"].includes(t.type)) {
+      if (["REFERENCE", "NUMBER", "STRING", "INVALID_REFERENCE"].includes(t.type)) {
         t.value = NO_REAL_VALUE;
       }
     });
@@ -89,6 +95,7 @@ export class CompiledFormula implements Omit<Omit<ICompiledFormula, "tokens">, "
     this.tokens.forEach((token: Token) => {
       switch (token.type) {
         case "REFERENCE":
+        case "INVALID_REFERENCE":
           token.value = getters.getRangeString(
             this.rangeDependencies[referenceIndex++],
             this.sheetId,
@@ -110,19 +117,27 @@ export class CompiledFormula implements Omit<Omit<ICompiledFormula, "tokens">, "
     return parseTokens(this.getTokens(getters));
   }
 
+  /**
+   * Return the string representation of the formula, with the current dependencies and literal values.
+   * This is a heavy operation as it converts the rangeDependencies to string on each call.
+   * */
   toFormulaString(getters: CoreGetters, referenceOption?: RangeStringOptions): string {
     if (this.isBadExpression) {
       return this.normalizedFormula;
     }
 
-    return concat(this.getTokens(getters, referenceOption).map((t) => t.value));
+    return concat(
+      this.getTokens(getters, referenceOption)
+        .filter((token) => token.type !== "SPACE")
+        .map((t) => t.value)
+    );
   }
 
   usesSymbol(symbol: string) {
     return this.tokens.some(
       (t) =>
         t.type === "SYMBOL" &&
-        t.value.localeCompare(symbol, undefined, {
+        t.value.localeCompare(symbol, "en", {
           sensitivity: "accent",
         }) === 0
     );
@@ -132,10 +147,10 @@ export class CompiledFormula implements Omit<Omit<ICompiledFormula, "tokens">, "
     if (this.isBadExpression) {
       return false;
     }
-    const nonExportableFunctions = iterateAstNodes(parseTokens(this.tokens)).filter(
+    const nonExportableFunctions = iterateAstNodes(parseTokens(this.tokens)).some(
       (ast) => ast.type === "FUNCALL" && !functions[ast.value.toUpperCase()]?.isExported
     );
-    return nonExportableFunctions.length === 0;
+    return !nonExportableFunctions;
   }
 
   getFunctionsFromTokens(functionNames: string[], getters: CoreGetters) {
@@ -169,6 +184,9 @@ export class CompiledFormula implements Omit<Omit<ICompiledFormula, "tokens">, "
     return compiledFormula.isBadExpression;
   }
 
+  /**
+   * Recreates a CompiledFormula based on `base` with adapted dependencies.
+   * */
   static CopyWithDependencies(
     base: CompiledFormula,
     sheetId: UID,
@@ -186,6 +204,9 @@ export class CompiledFormula implements Omit<Omit<ICompiledFormula, "tokens">, "
     );
   }
 
+  /**
+   * Recreates a CompiledFormula based on `base` with adapted dependencies and/or different literal values.
+   * */
   static CopyWithDependenciesAndLiteral(
     base: CompiledFormula,
     sheetId: UID,
@@ -205,6 +226,10 @@ export class CompiledFormula implements Omit<Omit<ICompiledFormula, "tokens">, "
     );
   }
 
+  /**
+   * When copy/pasting a formula across sheets, the formula is serialized (all it's serializable properties are kept) and deserialized on the new sheet.
+   * This function allows to recompile the formula based on the serializable properties.
+   * */
   static CompileForSerializedFormula(
     sheetId: UID,
     base: SerializedCompiledFormula
@@ -242,6 +267,9 @@ export class CompiledFormula implements Omit<Omit<ICompiledFormula, "tokens">, "
   }
 }
 
+/**
+ * A compiled formula serialized
+ * */
 export type SerializedCompiledFormula = {
   sheetId: UID;
   tokens: Token[];
--- a/packages/o-spreadsheet-engine/src/helpers/expand_range.ts
+++ b/packages/o-spreadsheet-engine/src/helpers/expand_range.ts
@@ -1,0 +1,30 @@
+import { Position } from "../types/misc";
+import { toCartesian } from "./coordinates";
+
+export function expandXc(xc: string): Iterable<Position> {
+  const pos = toCartesian(xc);
+  return [pos];
+}
+
+/**
+ * Helper to expand a range string like A1:B3 into all cell references in the range
+ * @param start
+ * @param end
+ */
+export function* expandRange(start: string, end: string): Generator<Position> {
+  const startPos = toCartesian(start);
+  const endPos = toCartesian(end);
+  for (
+    let col = Math.min(startPos.col, endPos.col);
+    col <= Math.max(startPos.col, endPos.col);
+    col++
+  ) {
+    for (
+      let row = Math.min(startPos.row, endPos.row);
+      row <= Math.max(startPos.row, endPos.row);
+      row++
+    ) {
+      yield { col, row };
+    }
+  }
+}
--- a/packages/o-spreadsheet-engine/src/model.ts
+++ b/packages/o-spreadsheet-engine/src/model.ts
@@ -2,7 +2,7 @@ import { LocalTransportService } from "./collaborative/local_transport_service";
 import { ReadonlyTransportFilter } from "./collaborative/readonly_transport_filter";
 import { Session } from "./collaborative/session";
 import { DEFAULT_REVISION_ID } from "./constants";
-import { UuidGenerator } from "./helpers";
+import { deepEquals, UuidGenerator } from "./helpers";
 import { EventBus } from "./helpers/event_bus";
 import { deepCopy, lazy } from "./helpers/misc";
 import { buildRevisionLog } from "./history/factory";
@@ -69,7 +69,7 @@ const enum Status {
  * State changes are then performed through commands.  Commands are dispatched
  * to the model, which will then relay them to each plugins (and the history
  * handler). Then, the model will trigger an 'update' event to notify whoever
- * is concerned that the command was applied (if it was not cancelled).
+ * is concerned that the command was applied (if it was not canceled).
  *
  * Also, the model has an unconventional responsibility: it actually renders the
  * visible viewport on a canvas. This is because each plugins actually manage a
@@ -225,24 +225,26 @@ export class Model extends EventBus<any> implements CommandDispatcher {
       this.uiHandlers.push(plugin);
     }
 
-    // starting plugins
-    this.dispatch("START");
-    // Model should be the last permanent subscriber in the list since he should render
-    // after all changes have been applied to the other subscribers (plugins)
-    this.selection.observe(this, {
-      handleEvent: () => this.trigger("update"),
-    });
-    // This should be done after construction of LocalHistory due to order of
-    // events
-    this.setupSessionEvents();
+    if (this.config.mode !== "export_verification") {
+      // starting plugins
+      this.dispatch("START");
+      // Model should be the last permanent subscriber in the list since he should render
+      // after all changes have been applied to the other subscribers (plugins)
+      this.selection.observe(this, {
+        handleEvent: () => this.trigger("update"),
+      });
+      // This should be done after construction of LocalHistory due to order of
+      // events
+      this.setupSessionEvents();
 
-    this.joinSession();
+      this.joinSession();
 
-    if (config.snapshotRequested || (data["[Content_Types].xml"] && !this.getters.isReadonly())) {
-      const startSnapshot = performance.now();
-      console.debug("Snapshot requested");
-      this.session.snapshot(this.exportData());
-      console.debug("Snapshot taken in", performance.now() - startSnapshot, "ms");
+      if (config.snapshotRequested || (data["[Content_Types].xml"] && !this.getters.isReadonly())) {
+        const startSnapshot = performance.now();
+        console.debug("Snapshot requested");
+        this.session.snapshot(this.exportData());
+        console.debug("Snapshot taken in", performance.now() - startSnapshot, "ms");
+      }
     }
     console.debug("Model created in", performance.now() - start, "ms");
     console.debug("######");
@@ -373,7 +375,10 @@ export class Model extends EventBus<any> implements CommandDispatcher {
       name: _t("Anonymous").toString(),
     };
     const transportService = config.transportService || new LocalTransportService();
-    const isReadonly = config.mode === "readonly" || config.mode === "dashboard";
+    const isReadonly =
+      config.mode === "readonly" ||
+      config.mode === "dashboard" ||
+      config.mode === "export_verification";
     return {
       ...config,
       mode: config.mode || "normal",
@@ -630,10 +635,36 @@ export class Model extends EventBus<any> implements CommandDispatcher {
    * export data out of the model.
    */
   exportData(): WorkbookData {
+    const squished = true;
+    const unsquished = false;
+    const exportSquished = this._exportData(squished);
+    const exportUnsquished = this._exportData(unsquished);
+    const verificationConfig = {
+      ...this.config,
+      mode: "export_verification" as Mode, // will not trigger evaluation or join the session
+      client: { id: "exporter", name: "exporter" },
+      snapshotRequested: false, // guarantee that no extra snapshot is requested
+      transportService: new LocalTransportService(),
+    };
+
+    const exportVerificationModel = new Model(
+      deepCopy(exportSquished), // the import itself modifies the data, so we need to deep copy it to keep the original one for comparison
+      verificationConfig
+    )._exportData(unsquished);
+
+    if (!deepEquals(exportUnsquished, exportVerificationModel)) {
+      exportUnsquished.isNotSquishable = true;
+      return exportUnsquished;
+    } else {
+      return exportSquished;
+    }
+  }
+
+  _exportData(shouldSquish: boolean): WorkbookData {
     let data = createEmptyWorkbookData();
     for (const handler of this.handlers) {
       if (handler instanceof CorePlugin) {
-        handler.export(data);
+        handler.export(data, shouldSquish);
       }
     }
     data.revisionId = this.session.getRevisionId() || DEFAULT_REVISION_ID;
--- a/packages/o-spreadsheet-engine/src/plugins/core/cell.ts
+++ b/packages/o-spreadsheet-engine/src/plugins/core/cell.ts
@@ -8,7 +8,7 @@ import {
 } from "../../helpers/data_normalization";
 import { deepEquals, range, replaceNewLines } from "../../helpers/misc";
 
-import { toCartesian, toXC } from "../../helpers/coordinates";
+import { toXC } from "../../helpers/coordinates";
 import { CorePlugin } from "../core_plugin";
 
 import { isInside } from "../../helpers/zones";
@@ -39,6 +39,8 @@ import { DEFAULT_LOCALE } from "../../types/locale";
 import { AdaptSheetName, Style, UpdateCellData, Zone } from "../../types/misc";
 import { Range, RangePart } from "../../types/range";
 import { ExcelWorkbookData, WorkbookData } from "../../types/workbook_data";
+import { SquishedCell, Squisher } from "./squisher";
+import { Unsquisher } from "./unsquisher";
 
 interface CoreState {
   // this.cells[sheetId][cellId] --> cell|undefined
@@ -236,6 +238,7 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
   // ---------------------------------------------------------------------------
 
   import(data: WorkbookData) {
+    const start = performance.now();
     for (const sheet of data.sheets) {
       const sheetId = sheet.id;
       const cellsData = new PositionMap<{
@@ -245,11 +248,19 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
         format?: number;
       }>();
       // cells content
-      for (const xc in sheet.cells) {
-        if (sheet.cells[xc]) {
-          const { col, row } = toCartesian(xc);
-          const position = { sheetId: sheet.id, col, row };
-          cellsData.set(position, { content: sheet.cells[xc] });
+      const unsquisher = new Unsquisher();
+      for (const unsquishedItem of unsquisher.unsquishSheet(sheet.cells, sheet.id, this.getters)) {
+        if (unsquishedItem.content || unsquishedItem.compiled) {
+          const position = {
+            sheetId: sheet.id,
+            col: unsquishedItem.position.col,
+            row: unsquishedItem.position.row,
+          };
+          if (unsquishedItem.compiled) {
+            cellsData.set(position, { compiledFormula: unsquishedItem.compiled });
+          } else {
+            cellsData.set(position, { content: unsquishedItem.content });
+          }
         }
       }
       // cells style and format
@@ -284,22 +295,22 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
         }
       }
     }
+    console.debug("cells imported in ", performance.now() - start);
   }
 
-  export(data: WorkbookData) {
+  export(data: WorkbookData, shouldSquish: boolean) {
     const styles: { [styleId: number]: Style } = {};
     const formats: { [formatId: number]: string } = {};
-
     for (const _sheet of data.sheets) {
+      const squisher = new Squisher(this.getters);
       const positionsByStyle: Record<number, CellPosition[]> = [];
       const positionsByFormat: Record<number, CellPosition[]> = [];
-      const cells: { [key: string]: string } = {};
+      const cells: { [key: string]: SquishedCell } = {};
       const positions = Object.values(this.cells[_sheet.id] || {})
         .map((cell) => this.getters.getCellPosition(cell.id))
         .sort((a, b) => (a.col === b.col ? a.row - b.row : a.col - b.col));
       for (const position of positions) {
         const cell = this.getters.getCell(position)!;
-        const xc = toXC(position.col, position.row);
         const style = this.extractCustomStyle(cell);
         if (Object.keys(style).length) {
           const styleId = getItemId<Style>(style, styles);
@@ -311,15 +322,18 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
           positionsByFormat[formatId] ??= [];
           positionsByFormat[formatId].push(position);
         }
+        const xc = toXC(position.col, position.row);
         if (cell.isFormula) {
-          cells[xc] = cell.compiledFormula.toFormulaString(this.getters);
+          cells[xc] = shouldSquish
+            ? squisher.squish(cell, _sheet.id)
+            : cell.compiledFormula.toFormulaString(this.getters);
         } else if (cell.content) {
-          cells[xc] = cell.content;
+          cells[xc] = shouldSquish ? squisher.squish(cell, _sheet.id) : cell.content;
         }
       }
       _sheet.styles = groupItemIdsByZones(positionsByStyle);
       _sheet.formats = groupItemIdsByZones(positionsByFormat);
-      _sheet.cells = cells;
+      _sheet.cells = shouldSquish ? squisher.squishSheet(cells) : cells;
     }
     data.styles = styles;
     data.formats = formats;
@@ -340,7 +354,7 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
   }
 
   exportForExcel(data: ExcelWorkbookData) {
-    this.export(data);
+    this.export(data, false);
     const incompatibleIds: number[] = [];
     for (const formatId in data.formats || []) {
       if (!isExcelCompatible(data.formats[formatId])) {
@@ -409,7 +423,7 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
     return newFormula.toFormulaString(this.getters, { useBoundedReference });
   }
 
-  /*
+  /**
    * Constructs a formula string based on an initial formula and a translation vector
    */
   getTranslatedCellFormula(
@@ -422,7 +436,7 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
       compiledFormula = CompiledFormula.CompileForSerializedFormula(sheetId, compiledFormula);
     }
     const adaptedDependencies = this.getters.createAdaptedRanges(
-      (compiledFormula as CompiledFormula).rangeDependencies,
+      compiledFormula.rangeDependencies,
       offsetX,
       offsetY,
       sheetId
--- a/packages/o-spreadsheet-engine/src/plugins/core/squisher.ts
+++ b/packages/o-spreadsheet-engine/src/plugins/core/squisher.ts
@@ -1,0 +1,298 @@
+import { CompiledFormula } from "../../formulas/compiler";
+import { deepCopy, deepEquals } from "../../helpers";
+import { toCartesian, toXC } from "../../helpers/coordinates";
+import { getRangeString } from "../../helpers/range";
+import { Cell } from "../../types/cells";
+import { CoreGetters } from "../../types/core_getters";
+import { UID } from "../../types/misc";
+import { Range } from "../../types/range";
+
+/**
+ * Example of squishing:
+ * "=concat(A1, "test", B1:C3, 5)" followed by
+ * "=concat(A2, "test", B1:C3, 7)"
+ *        __|      |_     |    |
+ *       |    _______|____|    |
+ *       v   v       v         v
+ * {R: "+R1|=", S: ["="], N: "+2"}
+ *
+ * N can be full numbers or relative change identified with a + sign before the number.
+ * R can be full references or relative change identified with a +Ck or +Rk (k being the number of columns or rows of the change).
+ * S can only be the full string if it changed, or "=" if it did not change.
+ * * */
+export interface SquishedFormula {
+  N?: string; // the numbers used in the formula, ordered by position, converted to string, separated by |
+  S?: string[]; // the strings used in the formula, ordered by position
+  R?: string | string[]; // the references used in the formula, ordered by position, converted to string, separated by | if needed
+}
+
+export type SquishedCell = string | SquishedFormula;
+
+export const SEPARATOR = "|";
+export const NO_CHANGE = "=";
+
+export class Squisher {
+  private readonly getters: CoreGetters;
+  // the base formula to compare against
+  private baseFormula: CompiledFormula | undefined;
+  // for each number in the base formula, how much offset has already been applied
+  private alreadyAppliedNumberOffsets: number[] = [];
+  // for each string in the base formula, the previous string value
+  private previousStrings: string[] = [];
+  // whether the base formula was already transformed. Formulas that have already been transformed must continue to be transformed
+  private baseFormulaWasTransformed: boolean = false;
+
+  constructor(getters: CoreGetters) {
+    this.getters = getters;
+  }
+
+  /** Build the result object based on a transformation. Joins the numbers into a single string, tries to do the same
+   * with references (unless they contain a sheet name that contains the separator). */
+  private buildResult(
+    numbers: string[],
+    strings: string[],
+    references: string[]
+  ): Partial<SquishedFormula> {
+    const res: Partial<SquishedFormula> = {};
+    if (numbers.length) {
+      res.N = numbers.join(SEPARATOR);
+    }
+    if (strings.length) {
+      res.S = strings;
+    }
+    if (references.length) {
+      let referencePattern: string | string[] | null | undefined;
+      if (references.some((x) => x.includes(SEPARATOR))) {
+        referencePattern = references;
+      } else {
+        referencePattern = references.join(SEPARATOR);
+      }
+      res.R = referencePattern;
+    }
+    return res;
+  }
+
+  /** Change the base formula to the given one, resetting all offsets and previous strings */
+  resetBaseTo(formula: CompiledFormula) {
+    this.baseFormula = CompiledFormula.CopyWithDependenciesAndLiteral(
+      formula,
+      formula.sheetId,
+      deepCopy(formula.rangeDependencies),
+      formula.literalValues.numbers,
+      formula.literalValues.strings
+    );
+    this.alreadyAppliedNumberOffsets = formula.literalValues.numbers.map((_) => 0);
+    this.previousStrings = formula.literalValues.strings.map((x) => x.value);
+    this.baseFormulaWasTransformed = false;
+  }
+
+  /** Reset the base formula to undefined, resetting all offsets */
+  resetBase() {
+    if (this.baseFormula) {
+      this.baseFormula = undefined;
+      this.alreadyAppliedNumberOffsets = [];
+      this.previousStrings = [];
+      this.baseFormulaWasTransformed = false;
+    }
+  }
+
+  /**
+   * Takes a cell and squish their formulas against the previous one (in the previous call).
+   * We should call this method for each cell in the sheet, in order from top to bottom, then from left to right (all cells of a columns, for each columns left to right).
+   *
+   * The result of this method is:
+   * - if the cell is not a formula, returns the content as is and resets the base formula
+   * - if the cell is a formula:
+   *   - if there is no previous formula, or the normalized formula is different from the previous one, resets the base formula to this one and returns the full formula string
+   *   - else, compares the literal values and range dependencies to the previous formula, and for each parameter:
+   *     - for numbers: returns a relative change (+N or -N) if possible, else the full number or "=" if unchanged
+   *     - for strings: returns the full string if changed, else "="
+   *     - for references: returns a relative change (+Ck or +Rk) if possible, else the full reference or "=" if unchanged
+   * */
+  squish(cell: Cell, forSheetId: UID): SquishedCell {
+    if (!cell.isFormula) {
+      this.resetBase();
+      return cell.content;
+    }
+
+    let numbers: string[] = [];
+    let strings: string[] = [];
+    let references: string[] = [];
+
+    if (
+      !this.baseFormula ||
+      this.baseFormula.normalizedFormula !== cell.compiledFormula.normalizedFormula
+    ) {
+      this.resetBaseTo(cell.compiledFormula);
+      return cell.compiledFormula.toFormulaString(this.getters);
+    } else {
+      if (
+        !this.baseFormulaWasTransformed &&
+        deepEquals(cell.compiledFormula.literalValues, this.baseFormula.literalValues) &&
+        deepEquals(cell.compiledFormula.rangeDependencies, this.baseFormula.rangeDependencies)
+      ) {
+        return cell.compiledFormula.toFormulaString(this.getters);
+      }
+      numbers = this.squishNumbers(cell.compiledFormula.literalValues.numbers);
+      strings = this.squishStrings(cell.compiledFormula.literalValues.strings);
+      references = this.squishReferences(cell.compiledFormula.rangeDependencies, forSheetId);
+      this.baseFormulaWasTransformed = true;
+    }
+    return this.buildResult(numbers, strings, references);
+  }
+
+  /**
+   * Read all the consecutive cells with either the same content or the same transformation and merge their key into one zone
+   * Do not join cells from different columns
+   * */
+  squishSheet(cells: { [key: string]: string | SquishedCell }): {
+    [key: string]: string | SquishedCell;
+  } {
+    const allKeys = Object.keys(cells);
+    const result: { [key: string]: string | SquishedCell } = {};
+    for (let startIndex = 0; startIndex < allKeys.length; startIndex++) {
+      const startKey = toCartesian(allKeys[startIndex]);
+      let offset = 0;
+
+      for (offset = 0; offset + startIndex + 1 < allKeys.length; offset++) {
+        const nextKey = toCartesian(allKeys[offset + startIndex + 1]);
+        if (
+          nextKey.col !== startKey.col || // different column, do not merge
+          nextKey.row !== startKey.row + offset + 1 || // not consecutive, do not merge
+          !deepEquals(cells[allKeys[offset + startIndex + 1]], cells[allKeys[startIndex]]) // different content, do not merge
+        ) {
+          break;
+        }
+      }
+
+      if (offset > 0) {
+        // we have found consecutive cells with the same pattern or content, merge them
+        const rangeKey = `${allKeys[startIndex]}:${toXC(startKey.col, startKey.row + offset)}`;
+        result[rangeKey] = cells[allKeys[startIndex]];
+        startIndex += offset;
+      } else {
+        result[allKeys[startIndex]] = cells[allKeys[startIndex]];
+      }
+    }
+
+    return result;
+  }
+
+  /**
+   * Squish the references. Result for each reference should be compared the previous formula.
+   * The result can be:
+   * - a relative change, e.g. +C2 or +R5 as string
+   * - the full reference if the reference is too different
+   * */
+  private squishReferences(references: Range[], forSheetId: UID) {
+    if (!this.baseFormula) {
+      throw new Error("No previous cell to squish against");
+    }
+    return references.map((reference, index) =>
+      this.squishOneReference(reference, this.baseFormula!.rangeDependencies, index, forSheetId)
+    );
+  }
+
+  private squishOneReference(
+    reference: Range,
+    previousReferences: Range[],
+    index: number,
+    forSheetId: UID
+  ) {
+    const previousReference = previousReferences[index];
+    if (deepEquals(previousReference, reference)) {
+      return NO_CHANGE;
+    }
+    if (
+      previousReference.sheetId !== reference.sheetId ||
+      previousReference.prefixSheet !== reference.prefixSheet ||
+      previousReference.invalidSheetName !== reference.invalidSheetName ||
+      previousReference.invalidXc !== reference.invalidXc
+    ) {
+      // sheet changed or valid/invalid changed, cannot squish
+      previousReferences[index] = deepCopy(reference);
+      return getRangeString(reference, forSheetId, this.getters.getSheetName);
+    }
+    if (
+      previousReference.unboundedZone.bottom === undefined ||
+      previousReference.unboundedZone.right === undefined ||
+      reference.unboundedZone.bottom === undefined ||
+      reference.unboundedZone.right === undefined
+    ) {
+      // unbounded ranges, cannot squish
+      previousReferences[index] = deepCopy(reference);
+      return getRangeString(reference, forSheetId, this.getters.getSheetName);
+    }
+    for (let i = 0; i < reference.parts.length; i++) {
+      if (
+        previousReference.parts[i].colFixed !== reference.parts[i].colFixed ||
+        previousReference.parts[i].rowFixed !== reference.parts[i].rowFixed
+      ) {
+        // absolute/relative parts changed, cannot squish
+        previousReferences[index] = deepCopy(reference);
+        return getRangeString(reference, forSheetId, this.getters.getSheetName);
+      }
+    }
+    const currentZone = reference.zone;
+    const previousZone = previousReference.zone;
+    if (
+      currentZone.top !== currentZone.bottom ||
+      currentZone.left !== currentZone.right ||
+      previousZone.top !== previousZone.bottom ||
+      previousZone.left !== previousZone.right
+    ) {
+      // ranges, cannot squish
+      previousReferences[index] = deepCopy(reference);
+      return getRangeString(reference, forSheetId, this.getters.getSheetName);
+    }
+
+    // 1D range squishing
+    const diffCol = reference.zone.left - previousReference.zone.left;
+    const diffRow = reference.zone.top - previousReference.zone.top;
+    previousReference.zone = deepCopy(reference.zone);
+    previousReference.unboundedZone = deepCopy(reference.unboundedZone);
+    if (diffCol !== 0 && diffRow === 0) {
+      return `${diffCol > 0 ? "+" : "-"}C${Math.abs(diffCol)}`; // ex. +C2 or -C3
+    } else if (diffRow !== 0 && diffCol === 0) {
+      return `${diffRow > 0 ? "+" : "-"}R${Math.abs(diffRow)}`; // ex. +R5 or -R4
+    }
+
+    return getRangeString(reference, forSheetId, this.getters.getSheetName);
+  }
+
+  /**
+   * Squish the number parameters. Result for each parameter should be compared the previous formula.
+   * The result can be:
+   * - a relative change, e.g. +2 or -5 as string
+   * - "=" meaning no change as string
+   * */
+  private squishNumbers(numbers: { value: number }[]) {
+    const result: string[] = numbers.map((x) => NO_CHANGE);
+    for (let i = 0; i < numbers.length; i++) {
+      const diff =
+        numbers[i].value -
+        (this.baseFormula!.literalValues.numbers[i].value + this.alreadyAppliedNumberOffsets[i] ||
+          0);
+
+      if (diff !== 0) {
+        result[i] = "+" + diff.toString();
+        this.alreadyAppliedNumberOffsets[i] = (this.alreadyAppliedNumberOffsets[i] || 0) + diff;
+      }
+    }
+    return result;
+  }
+
+  private squishStrings(strings: { value: string }[]) {
+    const result: string[] = strings.map((x) => NO_CHANGE);
+    for (let i = 0; i < strings.length; i++) {
+      const str = strings[i].value;
+      const previousStr = this.previousStrings[i];
+      if (str !== previousStr) {
+        // different strings, cannot squish
+        result[i] = str;
+        this.previousStrings[i] = str;
+      }
+    }
+    return result;
+  }
+}
--- a/packages/o-spreadsheet-engine/src/plugins/core/unsquisher.ts
+++ b/packages/o-spreadsheet-engine/src/plugins/core/unsquisher.ts
@@ -1,0 +1,247 @@
+import { CompiledFormula } from "../../formulas/compiler";
+import { toCartesian } from "../../helpers/coordinates";
+import { expandRange, expandXc } from "../../helpers/expand_range";
+import { CoreGetters } from "../../types/core_getters";
+import { Position, UID } from "../../types/misc";
+import { Range } from "../../types/range";
+import { NO_CHANGE, SEPARATOR, SquishedCell, SquishedFormula } from "./squisher";
+
+type UnsquishMethod =
+  | "NOT_A_FORMULA"
+  | "NEW_FORMULA"
+  | "COMBINE_OFFSET"
+  | "FIRST_OFFSET"
+  | "NO_CHANGE";
+
+export class Unsquisher {
+  private previousCell: CompiledFormula | undefined;
+  private alreadyAppliedNumberOffset: number[] = [];
+  private previousString: string[] = [];
+  private alreadyAppliedReferenceOffset: Range[] = [];
+  private previousOffset: SquishedFormula | undefined = undefined;
+
+  /**
+   * Expands a squished sheet object back to a full cell map.
+   * For range keys like "B2:B73", fills all cells in the range with the value.
+   * For single cell keys like "A1", copies as is.
+   * Assumes input is an object: { [ref: string]: string | object }
+   *
+   * @param squished - The squished sheet object.
+   * Yields [cellRef, value] for each cell in the expanded squished object.
+   * @param sheetId
+   * @param getters
+   */
+  *unsquishSheet(
+    squished: { [key: string]: SquishedCell | undefined },
+    sheetId: UID,
+    getters: CoreGetters
+  ): Generator<{
+    position: Position;
+    content?: string;
+    compiled?: CompiledFormula;
+  }> {
+    // sort the keys in cartesian order (col first, then row) using the first part of the range (before ":")
+    const keys = Object.keys(squished)
+      .map((x) => {
+        return { cartesian: toCartesian(x.split(":")[0]), key: x };
+      })
+      .sort((a, b) =>
+        a.cartesian.col === b.cartesian.col
+          ? a.cartesian.row - b.cartesian.row
+          : a.cartesian.col - b.cartesian.col
+      );
+
+    let strategy: UnsquishMethod | undefined;
+    for (const { key } of keys) {
+      const current = squished[key];
+      if (current === undefined || current === null || current === "") {
+        continue; // skip empty entries
+      }
+      if (typeof current === "string") {
+        if (current.startsWith("=")) {
+          strategy = "NEW_FORMULA";
+          // compile the found formula. Reset previousCell and offsets because it's a new formula
+          const compiled = CompiledFormula.Compile(current, sheetId, getters);
+          this.previousCell = compiled;
+          this.alreadyAppliedNumberOffset = compiled.literalValues.numbers.map((n) => n.value);
+          this.previousString = compiled.literalValues.strings.map((s) => s.value);
+          this.alreadyAppliedReferenceOffset = [...compiled.rangeDependencies];
+          this.previousOffset = undefined;
+        } else {
+          strategy = "NOT_A_FORMULA";
+        }
+      } else {
+        // the current cell is an object, let's see if there is a transformation
+        if (current.N || current.S || current.R) {
+          if (!strategy || strategy === "NEW_FORMULA") {
+            strategy = "FIRST_OFFSET";
+          } else {
+            strategy = "COMBINE_OFFSET";
+          }
+        } else {
+          strategy = "NO_CHANGE";
+        }
+      }
+
+      const parts = key.split(":");
+      const positionsOfKey = parts.length === 1 ? expandXc(key) : expandRange(parts[0], parts[1]);
+      switch (strategy) {
+        case "NEW_FORMULA":
+        case "NO_CHANGE":
+          for (const position of positionsOfKey) {
+            yield { position, compiled: this.previousCell };
+          }
+          break;
+        case "NOT_A_FORMULA":
+          for (const position of positionsOfKey) {
+            yield {
+              position,
+              content: current as string | undefined,
+            };
+          }
+          break;
+        case "FIRST_OFFSET":
+          this.previousOffset = current as SquishedFormula;
+          for (const position of positionsOfKey) {
+            const result = this.unsquish(current as SquishedFormula, sheetId, getters);
+            yield { position, compiled: result };
+          }
+          break;
+        case "COMBINE_OFFSET":
+          if (!this.previousOffset) {
+            throw new Error("No previous offset to combine with");
+          }
+          const currentOffset = current as SquishedFormula;
+          this.previousOffset.N = currentOffset.N ?? this.previousOffset.N;
+          this.previousOffset.S = currentOffset.S ?? this.previousOffset.S;
+          this.previousOffset.R = currentOffset.R ?? this.previousOffset.R;
+          for (const position of positionsOfKey) {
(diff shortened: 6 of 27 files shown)


packages/o-spreadsheet-engine/src/formulas/tokenizer.ts
```diff
@@ -170,7 +174,7 @@ function tokenizeNumber(chars: TokenizingChars, locale: Locale): Token | null {
   return null;
 }
 
-function tokenizeString(chars: TokenizingChars): Token | null {
+function tokenizeString(chars: TokenizingChars, stringIndex: number = 0): Token | null {
```
VincentSchippefilt: remove this change

packages/o-spreadsheet-engine/src/helpers/misc.ts
```diff
@@ -745,3 +717,52 @@ export function chartStyleToCellStyle(style: ChartStyle): Style {
...
+    } else if (has1 && has2) {
+      if (typeof o1[key] !== typeof o2[key]) {
+        diffCount++;
+      } else {
+        diffCount += countDeepDifferences(o1[key], o2[key]);
+      }
+    }
+  }
+  return diffCount;
+}
```
VincentSchippefilt: see if still usefull

packages/o-spreadsheet-engine/src/plugins/core/cell.ts
```diff
@@ -657,66 +722,27 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
...
-    private rangeIndex: number,
-    private sheetId,
-    private getRangeString: (range: Range, sheetId: UID) => string
-  ) {}
-
-  get value() {
-    const range = this.ranges[this.rangeIndex];
-    return this.getRangeString(range, this.sheetId);
+  get contentWithFixedReferences() {
+    return this.compiledFormula.toFormulaString(this.getters, { useBoundedReference: true });
```
VincentSchippefilt: see if we can remove this class

VincentSchippefilt: will do it in a second phase

tests/clipboard/clipboard_plugin.test.ts
```diff
@@ -2038,7 +2038,7 @@ describe("clipboard", () => {
 
   test.each([
     ["=A1", "=#REF"],
-    ["=SUM(A1:A2)", "=SUM(#REF)"],
+    // ["=SUM(A1:A2)", "=SUM(#REF)"],
```
VincentSchippefilt: restore this test

packages/o-spreadsheet-engine/src/formulas/compiler.ts
```diff
@@ -36,29 +38,257 @@ export const UNARY_OPERATOR_MAP = {
...
-  symbols: string[];
-};
-
 // this cache contains all compiled function code, grouped by "structure". For
 // example, "=2*sum(A1:A4)" and "=2*sum(B1:B4)" are compiled into the same
 // structural function.
 // It is only exported for testing purposes
 export const functionCache: { [key: string]: FormulaToExecute } = {};
 
+export class BananaCompiledFormula implements Omit<CompiledFormula, "tokens"> {
```
pro-odoo: 🍌

packages/o-spreadsheet-engine/src/formulas/tokenizer.ts
```diff
@@ -155,7 +155,11 @@ function tokenizeOperator(chars: TokenizingChars): Token | null {
 
 const FIRST_POSSIBLE_NUMBER_CHARS = new Set("0123456789");
 
-function tokenizeNumber(chars: TokenizingChars, locale: Locale): Token | null {
+function tokenizeNumber(
+  chars: TokenizingChars,
+  locale: Locale,
+  numberIndex: number = 0
```
LucasLefevre: `numberIndex` not used

packages/o-spreadsheet-engine/src/formulas/compiler.ts
LucasLefevre: The existing `CompiledFormula` is only used once in this file. It could be moved here and renamed, and then rename the banana to `CompiledFormula`

packages/o-spreadsheet-engine/src/formulas/compiler.ts
```diff
@@ -36,29 +38,257 @@ export const UNARY_OPERATOR_MAP = {
...
+
 // -----------------------------------------------------------------------------
 // COMPILER
 // -----------------------------------------------------------------------------
 
-export function compile(formula: string): CompiledFormula {
+export function compile(formula: string, sheetId: UID): BananaCompiledFormula {
   const tokens = rangeTokenize(formula);
-  return compileTokens(tokens);
+  const compiledFormula = compileTokens(tokens);
```
LucasLefevre: `compileTokens` is no longer used outside of this file (at least in o-spreadsheet, I didn't check odoo)
-> we can remove `export`

packages/o-spreadsheet-engine/src/formulas/compiler.ts
```diff
@@ -36,29 +38,257 @@ export const UNARY_OPERATOR_MAP = {
...
+      }
+      this._dependencies = undefined;
+      this._rangeDependencies = dependencies;
+    }
+    this.isBadExpression = isBadExpression;
+    this.normalizedFormula = normalizedFormula;
+    this.execute = execute;
+  }
+
+  getStringifiedTokens(
```
LucasLefevre: The tokens are not stringified
```suggestion
  getTokens(
```

packages/o-spreadsheet-engine/src/formulas/compiler.ts
```diff
@@ -36,29 +38,257 @@ export const UNARY_OPERATOR_MAP = {
...
+    let referenceIndex = 0;
+    let numberIndex = 0;
+    let stringIndex = 0;
+    if (this.isBadExpression) {
+      return Object.freeze(this._tokens);
+    }
+    return this._tokens.map((token: Token) => {
+      switch (token.type) {
+        case "REFERENCE":
+          token.value = getters.getRangeString(
```
LucasLefevre, line 99: If tokens are shared by all similar formulas, then it looks very error prone to give them to the external world and later mutate them.

packages/o-spreadsheet-engine/src/formulas/compiler.ts
```diff
@@ -36,29 +38,257 @@ export const UNARY_OPERATOR_MAP = {
...
+      }
+    }
+    return this._dependencies || [];
+  }
+
+  get rangeDependencies(): Range[] {
+    if (this.hasDependencies) {
+      if (!this._rangeDependencies) {
+        throw new Error("Range dependencies have not been ensured yet.");
+      }
```
LucasLefevre: ideally, this should be checked by the type system with different types depending on if it has (range) dependencies or not. 
But it might not be easy

packages/o-spreadsheet-engine/src/formulas/compiler.ts
VincentSchippefilt: I'll rename it at some point

packages/o-spreadsheet-engine/src/formulas/compiler.ts
```diff
@@ -36,29 +38,257 @@ export const UNARY_OPERATOR_MAP = {
...
+
+  isFirstNonWhitespaceToken(tokenValue: string): boolean {
+    const firstNonSpaceToken = this._tokens.find((token, i) => i > 0 && token.type !== "SPACE");
+    return (
+      firstNonSpaceToken?.type === "SYMBOL" && firstNonSpaceToken.value.toUpperCase() === "PIVOT"
+    );
+  }
+
+  static CopyWithDependencies(
+    base: BananaCompiledFormula,
```
LucasLefevre: any benefit of having this `static` instead of an instance method ?
`base` could just be `this`

src/components/composer/composer/cell_composer_store.ts
```diff
@@ -278,7 +279,11 @@ export class CellComposerStore extends AbstractComposerStore {
       this.editionMode === "inactive"
         ? Infinity // one liner
         : 80;
-    return prettify(parseTokens(cell.compiledFormula.tokens), width);
+    const tokens =
+      cell instanceof FormulaCellWithDependencies
+        ? cell.tokens
+        : cell.compiledFormula.getStringifiedTokens(this.getters);
```
LucasLefevre: this is equivalent and it allows to remove `get tokens` from `class FormulaCellWithDependencies`
```suggestion
    const tokens = cell.compiledFormula.getStringifiedTokens(this.getters);
```

packages/o-spreadsheet-engine/src/formulas/compiler.ts
```diff
@@ -36,29 +38,257 @@ export const UNARY_OPERATOR_MAP = {
...
+      if (!this._rangeDependencies) {
+        throw new Error("Range dependencies have not been ensured yet.");
+      }
+    }
+    return this._rangeDependencies || [];
+  }
+
+  convertXCDependenciesToRange(
+    getRangeFromSheetXC: (defaultSheetId: UID, sheetXC: string) => Range,
+    sheetId?: UID
```
LucasLefevre: It was unclear to me what was `sheetId` here compared to `this.sheetId`.
So I removed it and all tests pass :)

packages/o-spreadsheet-engine/src/formulas/compiler.ts
```diff
@@ -36,29 +38,257 @@ export const UNARY_OPERATOR_MAP = {
...
+  ) {
+    if (
+      this._dependencies?.length &&
+      this._dependencies.some((x: string | Range) => typeof x === "string")
```
LucasLefevre: There're all supposed to be strings already

```suggestion
```

packages/o-spreadsheet-engine/src/formulas/compiler.ts
```diff
@@ -36,29 +38,257 @@ export const UNARY_OPERATOR_MAP = {
...
+    ) {
+      this._rangeDependencies = this._dependencies.map((xc: string | Range) =>
+        typeof xc === "string" ? getRangeFromSheetXC(sheetId || this.sheetId, xc) : xc
+      );
```
LucasLefevre:
```suggestion
      this._rangeDependencies = this._dependencies.map((xc) => getRangeFromSheetXC(this.sheetId, xc));
```

packages/o-spreadsheet-engine/src/plugins/core/unsquisher.ts
```diff
@@ -0,0 +1,211 @@
...
+   * For single cell keys like "A1", copies as is.
+   * Assumes input is an object: { [ref: string]: string | object }
+   *
+   * @param squished - The squished sheet object.
+   * Yields [cellRef, value] for each cell in the expanded squished object.
+   * @param sheetId
+   * @param getRangeFromSheetXC
+   */
+  *unsquishSheet(
+    squished: { [key: string]: any },
```
LucasLefevre: TODO: replace `any` 🥴

packages/o-spreadsheet-engine/src/plugins/core/unsquisher.ts
```diff
@@ -0,0 +1,211 @@
...
+          }
+          const result = this.unsquish(current, sheetId, getRangeFromSheetXC);
+          yield { position: { col: cell[0], row: cell[1] }, compiled: result };
+        }
+      }
+    }
+  }
+
+  private unsquish(
+    squishedElement: any,
```
LucasLefevre: `any` 👀

packages/o-spreadsheet-engine/src/plugins/core/unsquisher.ts
```diff
@@ -0,0 +1,211 @@
...
+    sheetId: UID,
+    getRangeFromSheetXC: (sheetId: UID, reference: string) => Range
+  ): BananaCompiledFormula {
+    if (typeof squishedElement === "string") {
+      if (squishedElement.startsWith("=")) {
+        throw new Error(
+          "Programming Error: Formula strings should have been handled by the caller"
```
LucasLefevre: not needed if `squishedElement` is typed, the compiler will yell before it has a chance to run 😄

packages/o-spreadsheet-engine/src/plugins/core/unsquisher.ts
```diff
@@ -0,0 +1,211 @@
...
+  *unsquishSheet(
+    squished: { [key: string]: any },
+    sheetId: UID,
+    getRangeFromSheetXC: (sheetId: UID, reference: string) => Range
+  ): Generator<{
+    position: Position;
+    content?: string;
+    compiled?: BananaCompiledFormula;
+  }> {
+    for (const key in squished) {
```
LucasLefevre: we should rely on a explicit strict order that we want.
Relying on serializing/parsing order of all different layers (JS, python, ...) really isn't robust.
But it imply changing the export structure a bit

tests/model/squish.test.ts
```diff
@@ -0,0 +1,305 @@
...
+    [["=SUM(B1)", "=SUM(B$2)"], { R: "B$2" }],
+
+    //changing sheet name
+    [["=SUM(Sheet1!B1)", "=SUM(Sheet2!B1)"], { R: "Sheet2!B1" }],
+    [["=SUM(Sheet1!B1)", "=SUM(Sheet2!B2)"], { R: "Sheet2!B2" }],
+    [["=SUM(B1)", "=SUM(Sheet1!B1)"], { R: "Sheet1!B1" }],
+    [["=SUM(B1)", "=SUM('Shee  t2'!B1)"], { R: "'Shee  t2'!B1" }],
+
+    // ising a range that looks like a change
+    [["=SUM(R1)", "=SUM(R2)"], { R: "+R1" }],
```
LucasLefevre, line 159: - add tests with `$` references
`[["=SUM($R1)", "=SUM($R2)"], { R: "+R1" }],`
- tests with same sheet name
`[["=SUM(Sheet1!A1)", "=SUM(Sheet1!A2)"], { R: "+R1" }],`

- add tests with ranges `A1:A2` (also with various `$`)
- test with unbounded ranges
- test where the offset is different than 1
- test where there an offset in both col and row `=A1` and `=B2`

tests/model/squish.test.ts
```diff
@@ -0,0 +1,305 @@
...
+    [["=SUM(B2)", "=SUM(C3)"], { R: "C3" }],
+
+    // using unauthorized characters in the sheet name should still work
+    [["=SUM(B1)", "=SUM('Shee*t2'!B1)"], { R: "'Shee*t2'!B1" }],
+
+    // incorrect ranges should still work
+    [["=SUM(B1)", "=SUM(2B)"], "=SUM(2B)"],
+
+    // using the squish separator character in the sheet name is valid
+    [["=SUM(B1)", "=SUM('Shee|t2'!B1)"], { R: ["'Shee|t2'!B1"] }],
```
LucasLefevre, line 202: test with `|` in a string

packages/o-spreadsheet-engine/src/plugins/core/squisher.ts
```diff
@@ -0,0 +1,278 @@
...
+    }
+    const currentZone = reference.zone;
+    const previousZone = previousReference.zone;
+    if (
+      currentZone.top !== currentZone.bottom ||
+      currentZone.left !== currentZone.right ||
+      previousZone.top !== previousZone.bottom ||
+      previousZone.left !== previousZone.right
+    ) {
+      // ranges, cannot squish
```
LucasLefevre, line 244: 😞

packages/o-spreadsheet-engine/src/xlsx/conversion/table_conversion.ts
```diff
@@ -142,7 +142,7 @@ function deconstructSheets(convertedSheets: SheetData[]): DeconstructedSheets {
   const deconstructedSheets: DeconstructedSheets = {};
   for (const sheet of convertedSheets) {
     for (const xc in sheet.cells) {
-      const cellContent = sheet.cells[xc];
+      const cellContent = sheet.cells[xc] as string | undefined;
```
LucasLefevre, line 145: use (and adapt) `ExcelSheetData` instead of casting
```suggestion
      const cellContent = sheet.cells[xc];
```

tests/test_helpers/helpers.ts
```diff
@@ -552,13 +552,20 @@ export function getRangeCellsAsMatrix(
...
-export function createModelFromGrid(grid: GridDescr): Model {
+export function createModelFromGrid(grid: GridDescr, addExtraSheet = false): Model {
   const model = new Model();
   for (const xc in grid) {
     if (grid[xc] !== undefined) {
       setCellContent(model, xc, grid[xc]!);
     }
   }
+  if (addExtraSheet) {
+    model.dispatch("CREATE_SHEET", {
```
LucasLefevre: This is overkill and makes the API more complex.
 It's a simple line to add a sheet in a test: `createSheet(model, { sheetId: "Sheet2" });`

packages/o-spreadsheet-engine/src/plugins/core/squisher.ts
```diff
@@ -0,0 +1,278 @@
...
+    numbers: string[],
+    strings: string[],
+    references: string[]
+  ): Partial<SquishedFormula> {
+    const res: Partial<SquishedFormula> = {};
+    if (numbers.length && numbers.some((x) => x !== NO_CHANGE)) {
+      const numberPattern = numbers.join(SEPARATOR);
+      if (this.previousNumberPattern !== numberPattern) {
+        this.previousNumberPattern = numberPattern;
+        res["N"] = numberPattern;
```
LucasLefevre: this syntax is not entirely type-safe.
`res["coucou"] =  numberPattern;` also works
```suggestion
        res.N = numberPattern;
```

packages/o-spreadsheet-engine/src/plugins/core/squisher.ts
```diff
@@ -0,0 +1,278 @@
...
+  private alreadyAppliedNumberOffsets: number[] = [];
+  private previousStrings: string[] = [];
+  private previousNumberPattern: string | null = null;
+  private previousReferencePattern: string | string[] | null = null;
+
+  constructor(getSheetName: (sheetId: UID) => string) {
+    this.getSheetName = getSheetName;
+  }
+
+  buildResultExcludingIdentical(
```
LucasLefevre:
```suggestion
  private buildResultExcludingIdentical(
```

packages/o-spreadsheet-engine/src/plugins/core/squisher.ts
```diff
@@ -0,0 +1,278 @@
...
+      }
+      if (this.previousReferencePattern !== referencePattern) {
+        this.previousReferencePattern = referencePattern;
+        res["R"] = referencePattern;
+      }
+    }
+    return res;
+  }
+
+  resetBaseTo(cell: FormulaCell) {
```
LucasLefevre:
```suggestion
  private resetBaseTo(cell: FormulaCell) {
```

packages/o-spreadsheet-engine/src/plugins/core/squisher.ts
```diff
@@ -0,0 +1,278 @@
...
+
+export type SquishedCell = string | Partial<SquishedFormula>;
+
+export const SEPARATOR = "|";
+export const NO_CHANGE = "=";
+
+export class Squisher {
+  previousCell: FormulaCell | undefined;
+  private readonly getSheetName: (sheetId: UID) => string;
+  private alreadyAppliedOffsets = [
```
LucasLefevre:
```suggestion
  private alreadyAppliedRangeOffsets = [
```

packages/o-spreadsheet-engine/src/plugins/core/squisher.ts
```diff
@@ -0,0 +1,278 @@
...
+   * */
+  private squishReferences(references: Range[], forSheetId: UID) {
+    if (this.previousCell === undefined) {
+      throw new Error("No previous cell to squish against");
+    }
+    return references.map((x, index) =>
+      this.squishOneReference(
+        x,
+        this.previousCell!.compiledFormula.rangeDependencies[index],
+        forSheetId,
```
LucasLefevre: you can remove `!` by doing this
```suggestion
    }
    const previousCell = this.previousCell;
    return references.map((x, index) =>
      this.squishOneReference(
        x,
        previousCell.compiledFormula.rangeDependencies[index],
        forSheetId,
```

packages/o-spreadsheet-engine/src/formulas/compiler.ts
VincentSchippefilt: yep, that's the plan

packages/o-spreadsheet-engine/src/formulas/compiler.ts
VincentSchippefilt: they are marked as read-only, so changes to tokens are marked by the compiler as invalid

packages/o-spreadsheet-engine/src/formulas/compiler.ts
```diff
@@ -36,37 +38,271 @@ export const UNARY_OPERATOR_MAP = {
...
+  private _dependencies?: string[];
+  isBadExpression: boolean;
+  normalizedFormula: string;
+  execute: FormulaToExecute;
+  private _rangeDependencies?: Range[];
+  hasDependencies: boolean;
+  sheetId: UID;
+
+  constructor(
+    sheetId: UID,
```
pro-odoo: Nitpick:
```ts
sheetId: UID
constructor(sheetId: UID) {
    this.sheetId = sheetId;
}
```
is the same as 
```ts
constructor(public sheetId: UID) {}
```

packages/o-spreadsheet-engine/src/formulas/compiler.ts
```diff
@@ -36,37 +38,271 @@ export const UNARY_OPERATOR_MAP = {
...
+      compiledFormula.symbols,
+      base._rangeDependencies,
+      base.isBadExpression,
+      base.normalizedFormula,
+      compiledFormula.execute
+    );
+  }
+}
+
+export type SerializedBananaCompiledFormula = {
```
pro-odoo: `SerializedCompiledFormula`

packages/o-spreadsheet-engine/src/formulas/compiler.ts
```diff
@@ -36,37 +38,271 @@ export const UNARY_OPERATOR_MAP = {
...
+}
 
 // this cache contains all compiled function code, grouped by "structure". For
 // example, "=2*sum(A1:A4)" and "=2*sum(B1:B4)" are compiled into the same
 // structural function.
 // It is only exported for testing purposes
 export const functionCache: { [key: string]: FormulaToExecute } = {};
 
+export class CompiledFormula implements Omit<ICompiledFormula, "tokens"> {
+  private readonly _tokens: Token[];
```
pro-odoo: Nitpick, in o-spreadsheet no need to prefix private variable with `_` as Typescript already tell us that the variable is private

packages/o-spreadsheet-engine/src/formulas/compiler.ts
```diff
@@ -36,37 +38,271 @@ export const UNARY_OPERATOR_MAP = {
...
+      { numbers: literalNumbers, strings: literalStrings },
+      base.symbols,
+      dependencies,
+      base.isBadExpression,
+      base.normalizedFormula,
+      base.execute
+    );
+  }
+
+  static CompileForSerializedFormula(
```
pro-odoo, line 233: `CompileFromSerializedFormula` ?

packages/o-spreadsheet-engine/src/formulas/compiler.ts
pro-odoo: We do have the guarantee that no one outside the compiler will modify the value of a token. Read-only on the array does not prevent to update the content of an element of that array, so IMO we should re-create a new token instead of mutate it, no ?

packages/o-spreadsheet-engine/src/helpers/expand_range.ts
```diff
@@ -0,0 +1,25 @@
+// Helper to expand a range string like A1:B3 into all cell references in the range
+import { toCartesian } from "./coordinates";
+
+export function* expandOne(xc: string): Generator<[number, number]> {
```
pro-odoo: `expandXC` ?

packages/o-spreadsheet-engine/src/helpers/misc.ts
```diff
@@ -748,10 +748,5 @@ export function chartStyleToCellStyle(style: ChartStyle): Style {
 }
 
 export function doesCellContainFunction(cell: Cell, formula: string): boolean {
-  return (
-    cell.isFormula &&
-    cell.compiledFormula.tokens.some(
-      (t) => t.type === "SYMBOL" && t.value.toUpperCase() === formula
-    )
-  );
+  return cell.isFormula && cell.compiledFormula.usesSymbol(formula);
```
pro-odoo, line 768: 👍

packages/o-spreadsheet-engine/src/formulas/compiler.ts
```diff
@@ -36,37 +38,271 @@ export const UNARY_OPERATOR_MAP = {
...
+  isBadExpression: boolean;
+  normalizedFormula: string;
+};
+
 // -----------------------------------------------------------------------------
 // COMPILER
 // -----------------------------------------------------------------------------
 
-export function compile(formula: string): CompiledFormula {
+export function compile(formula: string, sheetId: UID): CompiledFormula {
```
pro-odoo: Should `sheetId` be optional ? It feels unnatural from the caller to specify a fake sheet id (`no_sheet`).
What do you think ?

packages/o-spreadsheet-engine/src/formulas/compiler.ts
VincentSchippefilt: the static factory pattern :)
I also like to see the name of the type on a line that creates an object: either new Brol() or Brol.Create.... or Brol.Copy....

tests/model/squish.test.ts
```diff
@@ -0,0 +1,390 @@
...
+      A6: '=IF(AND(F20820<=\'Sheet1\'!$M$1,F20820>=\'Sheet1\'!$L$1),IFERROR(MID(C20820,SEARCH("(",C20820)+1,SEARCH(")",C20820)-SEARCH("(",C20820)-1),""))',
+      A7: '=IF(AND(F20821<=\'Sheet1\'!$M$1,F20821>=\'Sheet1\'!$L$1),IFERROR(MID(C20821,SEARCH("(",C20821)+1,SEARCH(")",C20821)-SEARCH("(",C20821)-1),""))',
+      D1: '="Test"+"Test"',
+      D2: '="Test"+"Test"',
+      D3: '="Test2"',
+      D4: '="Test2"',
+      D5: '="Test2"',
+    });
+  });
+  test("should squish identical formulas", () => {
```
hokolomopo, line 34:
```suggestion
  });
  
  test("should squish identical formulas", () => {
```
nitpick

packages/o-spreadsheet-engine/src/model.ts
```diff
@@ -624,11 +624,11 @@ export class Model extends EventBus<any> implements CommandDispatcher {
    * As the name of this method strongly implies, it is useful when we need to
    * export data out of the model.
    */
-  exportData(): WorkbookData {
+  exportData(shouldSquish: boolean = false): WorkbookData {
```
hokolomopo: I'd use a string `"squished" | "unsquished"` instead of a boolean. `model.exportData(true)` isn't exactly reader-friendly 🙈

tests/model/squish.test.ts
```diff
@@ -0,0 +1,390 @@
...
+    ],
+    // identical transformations
+    [
+      { A1: "=SUM(B1)", A2: "=SUM(B2)", A3: "=SUM(B3)" },
+      { A1: "=SUM(B1)", "A2:A3": { R: "+R1" } },
+    ],
+    // identical transformations and positions not following each other
+    [
+      { A1: "=SUM(B1)", A2: "=SUM(B2)", A4: "=SUM(B3)" },
+      { A1: "=SUM(B1)", A2: { R: "+R1" }, A4: {} },
```
hokolomopo: Using an empty object feels hard to understand/easier to mess up & delete. Maybe something like this instead ?
```suggestion
      { A1: "=SUM(B1)", A2: { R: "+R1" }, A4: { R: "" } },
```

tests/model/squish.test.ts
```diff
@@ -0,0 +1,390 @@
...
+    });
+    const exportSquished = model.exportData(true);
+    expect(exportSquished.sheets[0].cells).toEqual({ A1: "=SUM(B1)", A2: { R: "+R1" } });
+    expect(exportSquished.sheets[1].cells).toEqual({ A1: "=SUM(B3)", A2: { R: "+R1" } });
+  });
+
+  test.each([
+    // different cells
+    [["=SUM(B1)", "=SUM(B2)"], { R: "+R1" }],
+    [["=SUM(B2)", "=SUM(B1)"], { R: "+R-1" }],
```
hokolomopo: Why not `R: "-R1"` ? Less characters, and easier to understand.

packages/o-spreadsheet-engine/src/plugins/core/squisher.ts
```diff
@@ -0,0 +1,278 @@
...
+import { toCartesian, toXC } from "../../helpers/coordinates";
+import { getRangeString } from "../../helpers/range";
+import { Cell, FormulaCell } from "../../types/cells";
+import { UID } from "../../types/misc";
+import { Range } from "../../types/range";
+
+export interface SquishedFormula {
+  N?: string;
+  S?: string[];
+  R?: string | string[];
```
hokolomopo: I'd add some docstring on what N or S or R mean and do

packages/o-spreadsheet-engine/src/formulas/compiler.ts
```diff
@@ -36,37 +38,271 @@ export const UNARY_OPERATOR_MAP = {
...
+
+  getFunctionsFromTokens(functionNames: string[], getters: CoreGetters) {
+    // Parsing is an expensive operation, so we first check if the
+    // formula contains one of the function names
+    if (!functionNames.some((functionName) => this.usesSymbol(functionName))) {
+      return [];
+    }
+    let ast: AST | undefined;
+    try {
+      ast = parseTokens(this.getTokens(getters));
```
hokolomopo: missing some docstring explaining why this.tokens and this.getTokens are different. Maybe another name for `this.tokens` ?

packages/o-spreadsheet-engine/src/helpers/expand_range.ts
```diff
@@ -0,0 +1,25 @@
+// Helper to expand a range string like A1:B3 into all cell references in the range
```
hokolomopo: add dosctring do the helpers, instead of a comment at the top of the file that will be wrong in 1 year when we add other helpers in the file 🙂

packages/o-spreadsheet-engine/src/helpers/expand_range.ts
```diff
@@ -0,0 +1,25 @@
...
+  ) {
+    for (
+      let row = Math.min(startPos.row, endPos.row);
+      row <= Math.max(startPos.row, endPos.row);
+      row++
+    ) {
+      yield [col, row];
+    }
+  }
+}
```
hokolomopo:
```suggestion
export function* expandRange(start: string, end: string): Position[] {
  const startPos = toCartesian(start);
  const endPos = toCartesian(end);
  return positions({
      top: startPos.row,
      bottom: endPos.row,
      ...
  });
  }
```
Isn't it just shorter & easier to read ? Not sure we need to re-invent existing helpers & use fancy iterators. And `expandXc(myXc)` could just be replaced by `[toCartesian(myXc)]`

packages/o-spreadsheet-engine/src/plugins/core/cell.ts
```diff
@@ -369,44 +403,48 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
...
+      compiledFormula.rangeDependencies
     );
-    const adaptedDependencies = this.getters.removeRangesSheetPrefix(targetSheetId, dependencies);
-    return this.getFormulaString(targetSheetId, tokens, adaptedDependencies);
+    return CompiledFormula.CopyWithDependencies(
+      compiledFormula,
+      targetSheetId,
+      adaptedDependencies
+    ).toFormulaString(this.getters);
+    //this.getFormulaString(targetSheetId, compiledFormula, adaptedDependencies);
```
hokolomopo: comment

packages/o-spreadsheet-engine/src/formulas/compiler.ts
```diff
@@ -36,37 +38,271 @@ export const UNARY_OPERATOR_MAP = {
...
+          break;
+        case "NUMBER":
+          token.value = this.literalValues.numbers[numberIndex++].value.toString();
+          break;
+        case "STRING":
+          token.value = `"${
+            this.literalValues.strings[stringIndex++].value /*.replace(
+              /"/g,
+              '""'
+            )*/
```
hokolomopo: comment

packages/o-spreadsheet-engine/src/plugins/core/cell.ts
```diff
@@ -576,44 +614,32 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
...
-    if (compiledFormula.dependencies.length) {
-      return this.createFormulaCellWithDependencies(id, compiledFormula, format, sheetId);
+    const compiledFormula = compile(content, sheetId);
+    if (
+      compiledFormula.dependencies.length ||
+      compiledFormula.literalValues.numbers.length ||
+      compiledFormula.literalValues.strings.length
+    ) {
+      compiledFormula.convertXCDependenciesToRange(this.getters.getRangeFromSheetXC);
     }
```
hokolomopo: do we need to convert xc to ranges if there are only literal number/strings ? And in this case I think `convertXCDependenciesToRange` won't to anything, the first thing it does is check `this.hasDependencies`

hokolomopo: Oh it's for `NO_REAL_VALUE` isn't it ? Well it won't work if there are no ranges 😅 

Also why is it the only place where we do this check before calling `convertXCDependenciesToRange` ?

packages/o-spreadsheet-engine/src/plugins/ui_core_views/evaluation_conditional_format.ts
```diff
@@ -111,9 +111,15 @@ export class EvaluationConditionalFormatPlugin extends CoreViewPlugin {
           }
           break;
         case "CellIsRule":
-          const formulas = cf.rule.values.map((value) =>
-            value.startsWith("=") ? compile(value) : undefined
-          );
+          const formulas = cf.rule.values.map((value) => {
+            if (value.startsWith("=")) {
+              const compilationResult = compile(value, sheetId);
+              compilationResult.convertXCDependenciesToRange(this.getters.getRangeFromSheetXC);
```
hokolomopo: might be worth creating a helper `compileWithRanges` or something

packages/o-spreadsheet-engine/src/formulas/compiler.ts
hokolomopo: +1

src/clipboard_handlers/cell_clipboard.ts
```diff
@@ -256,19 +254,19 @@ export class CellClipboardHandler extends AbstractCellClipboardHandler<
     }
 
     let content = origin?.content;
-    if (origin?.tokens && origin.tokens.length > 0 && !clipboardOption?.isCutOperation) {
+    if (
+      origin?.compiledFormula &&
+      origin.compiledFormula.hasDependencies &&
```
hokolomopo:
```suggestion
      origin?.compiledFormula?.hasDependencies  
```
nitpick

tests/cells/cell_plugin.test.ts
```diff
@@ -608,64 +608,28 @@ describe("Cell dependencies and tokens are updated", () => {
...
-    expect(getCell(model, "A1")).toMatchObject({
-      content: "=C4",
-      compiledFormula: {
-        dependencies: [{ zone: toZone("C4") }],
-        tokens: [
-          { type: "OPERATOR", value: "=" },
-          { type: "REFERENCE", value: "C4" },
-        ],
-      },
-    });
```
hokolomopo, line 619: could test compiledFormula.getTokens() instead of removing the test

tests/composer/composer_component.test.ts
```diff
@@ -1589,7 +1589,7 @@ describe("composer highlights color", () => {
   });
 
   test("highlight range", async () => {
-    setCellContent(model, "A1", "=sum(a1:a10)");
+    setCellContent(model, "A1", "=sum(A1:A10)");
```
hokolomopo: why this change ?

hokolomopo (commented):
👋 

Mostly small things. I didn't read the details of the squisher/unsquisher. Really cool feature 🥳

packages/o-spreadsheet-engine/src/formulas/compiler.ts
```diff
@@ -36,37 +38,271 @@ export const UNARY_OPERATOR_MAP = {
...
+  getFunctionsFromTokens(functionNames: string[], getters: CoreGetters) {
+    // Parsing is an expensive operation, so we first check if the
+    // formula contains one of the function names
+    if (!functionNames.some((functionName) => this.usesSymbol(functionName))) {
+      return [];
+    }
+    let ast: AST | undefined;
+    try {
+      ast = parseTokens(this.getTokens(getters));
+    } catch {
```
LucasLefevre: actually, thanks to `this.isBadExpression` we know if parsing's gonna fail or not. It can early returns `if(this.isBadExpression)` and remove try/catch

packages/o-spreadsheet-engine/src/formulas/compiler.ts
```diff
@@ -36,37 +38,271 @@ export const UNARY_OPERATOR_MAP = {
...
+    return concat(this.getTokens(getters, referenceOption).map((t) => t.value));
+  }
+
+  usesSymbol(symbol: string) {
+    return this.tokens.some((t) => t.type === "SYMBOL" && t.value === symbol);
+  }
+
+  areAllFunctionsExportableToExcel(): boolean {
+    try {
+      const nonExportableFunctions = iterateAstNodes(parseTokens(this.tokens)).filter(
```
LucasLefevre: same here

packages/o-spreadsheet-engine/src/formulas/compiler.ts
LucasLefevre: +1
the *array* of tokens is read-only, not tokens themselves (they should be), otherwise you wouldn't need to mutate them here

packages/o-spreadsheet-engine/src/formulas/tokenizer.ts
```diff
@@ -49,7 +49,7 @@ type TokenType =
 
 export interface Token {
   readonly type: TokenType;
-  readonly value: string;
+  value: string;
```
LucasLefevre, line 52:
```suggestion
  readonly value: string;
```

packages/o-spreadsheet-engine/src/plugins/core/cell.ts
```diff
@@ -576,44 +614,32 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
...
   ): FormulaCell {
-    const compiledFormula = compile(content);
-    if (compiledFormula.dependencies.length) {
-      return this.createFormulaCellWithDependencies(id, compiledFormula, format, sheetId);
+    const compiledFormula = compile(content, sheetId);
+    if (
+      compiledFormula.dependencies.length ||
+      compiledFormula.literalValues.numbers.length ||
+      compiledFormula.literalValues.strings.length
+    ) {
```
LucasLefevre: useless check
`convertXCDependenciesToRange` already does nothing if there's no dependency

packages/o-spreadsheet-engine/src/plugins/core/range.ts
```diff
@@ -346,13 +350,17 @@ export class RangeAdapter implements CommandHandler<CoreCommand> {
       return formula;
     }
 
-    const compiledFormula = compile(formula);
+    const compiledFormula = compile(formula, sheetIdFrom);
     const updatedDependencies = compiledFormula.dependencies.map((dep) => {
       const range = this.getters.getRangeFromSheetXC(sheetIdFrom, dep);
```
LucasLefevre:
```suggestion
    const compiledFormula = compile(formula, sheetIdFrom);
    compiledFormula.convertXCDependenciesToRange();
const updatedDependencies = compiledFormula.rangeDependencies.map((range) => {
```

packages/o-spreadsheet-engine/src/plugins/core/range.ts
```diff
@@ -320,13 +320,17 @@ export class RangeAdapter implements CommandHandler<CoreCommand> {
       return formula;
     }
 
-    const compiledFormula = compile(formula);
+    const compiledFormula = compile(formula, sheetId);
     const updatedDependencies = compiledFormula.dependencies.map((dep) => {
       const range = this.getters.getRangeFromSheetXC(sheetId, dep);
```
LucasLefevre:
```suggestion
    const compiledFormula = compile(formula, sheetId);
    compiledFormula.convertXCDependenciesToRange(this.getters.getRangeFromSheetXC);
const updatedDependencies = compiledFormula.rangeDependencies.map((range) => {
```

packages/o-spreadsheet-engine/src/formulas/compiler.ts
```diff
@@ -36,37 +38,271 @@ export const UNARY_OPERATOR_MAP = {
...
+            )*/
+          }"`;
+          break;
+      }
+      return token;
+    });
+  }
+
+  get dependencies(): string[] {
+    if (this.hasDependencies) {
```
LucasLefevre: I believe there's an opportunity to make this class entirely type-safe and simpler which would not be that difficult.

- `compiledFormula.dependencies` is used only 3 times and they are very easy top remove (see other comments).
- the only places where we use `compile` without `convertXCDependenciesToRange` is to check `isBadExpression` (for which we do not even need to instantiate a new Banana).

We could have
- a helper `isBadExpression(formulaStr)` (it would also allow to remove the hack-ish `"no_sheet"`)
- something like `compileWithRanges` like Adrien suggested and it would instantiate the class instance with references already converted to ranges. 

And `CompileFormula` would not need to manage strings at all.

packages/o-spreadsheet-engine/src/formulas/compiler.ts
```diff
@@ -36,37 +38,271 @@ export const UNARY_OPERATOR_MAP = {
...
+
+  constructor(
+    public sheetId: UID,
+    private readonly tokens: Token[],
+    public literalValues: LiteralValues,
+    public symbols: string[],
+    dependencies: string[] | Range[],
+    public isBadExpression: boolean,
+    public normalizedFormula: string,
+    public execute: FormulaToExecute
```
LucasLefevre:
```suggestion
  readonly hasDependencies: boolean;

  constructor(
    readonly sheetId: UID,
    private readonly tokens: Token[],
    readonly literalValues: LiteralValues,
    readonly symbols: string[],
    dependencies: string[] | Range[],
    readonly isBadExpression: boolean,
    readonly normalizedFormula: string,
    readonly execute: FormulaToExecute
```

packages/o-spreadsheet-engine/src/plugins/core/cell.ts
```diff
@@ -255,15 +256,34 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
   // ---------------------------------------------------------------------------
 
   import(data: WorkbookData) {
+    console.debug("starting loading sheets");
```
LucasLefevre: I don't think we need to log at the beginning.
The timing at the end is the interesting part.

packages/o-spreadsheet-engine/src/plugins/core/cell.ts
```diff
@@ -291,42 +312,55 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
         }
       }
     }
+    console.debug("done importing sheets in ", performance.now() - start);
```
LucasLefevre:
```suggestion
    console.debug("cells imported in ", performance.now() - start);
```

packages/o-spreadsheet-engine/src/plugins/core/cell.ts
```diff
@@ -654,69 +680,3 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
     return CommandResult.Success;
   }
 }
-
-export class FormulaCellWithDependencies implements FormulaCell {
```
LucasLefevre, line 760: 👍

packages/o-spreadsheet-engine/src/plugins/core/cell.ts
```diff
@@ -576,44 +614,32 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
...
-    for (const xc of compiledFormula.dependencies) {
-      dependencies.push(this.getters.getRangeFromSheetXC(sheetId, xc));
-    }
-    return new FormulaCellWithDependencies(
+    const getters = this.getters;
+    return {
       id,
-      compiledFormula,
+      get content() {
+        return compiledFormula.toFormulaString(getters);
```
LucasLefevre: TODO check perf:
at the time, the class was introduced because it was much faster and memory efficient than the closure. But the closure is not the same, so to check.

see https://github.com/odoo/o-spreadsheet/commit/3169b5c9ec521061e0279aa65a68aef2d17b09db

packages/o-spreadsheet-engine/src/plugins/core/cell.ts
LucasLefevre: 👀

packages/o-spreadsheet-engine/src/formulas/compiler.ts
VincentSchippefilt: ok. in this version, the tokens are not exposed anymore. the outside world do not need the tokens: 
- the composer needs the AST and was requesting the tokens to parse them
- a bunch of different functions needed to know if they were used in the formula
- the string reconstruction needed the tokens.
--> now all of that happens internally in this class.

packages/o-spreadsheet-engine/src/formulas/compiler.ts
VincentSchippefilt: the dependencies as string are now completely gone :)

packages/o-spreadsheet-engine/src/helpers/expand_range.ts
VincentSchippefilt: you are right for the expandXC, for the rest, I'm trying to be really aggressive about not allocating memory when it is not needed. `positions` requires 2 copies of the zone object and generate all the positions in memory. I don't want that.

tests/cells/cell_plugin.test.ts
VincentSchippefilt: I didn't remove the test, juste made it more correct: the previous version was testing an implementation detail: the tokens were updated, this version tests the result: the complete formula got updated.
Also, the tokens are now completely private to the compiler and the CompiledFormula

tests/model/squish.test.ts
VincentSchippefilt: j'ai enlevé tous les cas ou on génère un objet vide, ce n'était pas 100% correct

packages/o-spreadsheet-engine/src/formulas/compiler.ts
VincentSchippefilt: done

packages/o-spreadsheet-engine/src/plugins/ui_feature/datavalidation_insertion.ts
```diff
@@ -15,14 +15,21 @@ export class DataValidationInsertionPlugin extends UIPlugin {
...
-            } else if (cell?.content && evaluatedCell.type === CellValueType.empty) {
+            } else if (
+              (cell?.isFormula || cell?.content) &&
+              evaluatedCell.type === CellValueType.empty
+            ) {
               let value: string | undefined;
-              if (cell.content.startsWith("=")) {
-                const result = this.getters.evaluateFormula(position.sheetId, cell.content);
+              if (cell.isFormula) {
+                // TODO: it's ridiculous to recompile the already compiled formula here, but I want this to work for now
```
LucasLefevre: there's `getters.evaluateCompiledFormula` 😉

packages/o-spreadsheet-engine/src/formulas/tokenizer.ts
VincentSchippefilt: as the tokens are now private to CompiledFormula, I can leave them editable

packages/o-spreadsheet-engine/src/plugins/core/cell.ts
VincentSchippefilt: good catch. I updated the code to remove the closure by moving "content" property out of FormulaCell and use dedicated properties everywhere

tests/model/squish.test.ts
VincentSchippefilt: ok for the references, but for the numbers I can't just have -1 as a transformation, because it could either be the full number -1 or the transformation "previous -1", so it will stay "+-1"

packages/o-spreadsheet-engine/src/xlsx/conversion/table_conversion.ts
VincentSchippefilt: not in this PR: the entire chain of function calls must be changed to be typed correctly

packages/o-spreadsheet-engine/src/formulas/compiler.ts
VincentSchippefilt: now that compile is part of CompileFormula, the sheet is always required.

demo/main.js
```diff
@@ -317,10 +317,10 @@ class Demo extends Component {
       this.transportService = undefined;
       this.stateUpdateMessages = [];
     }
-    this.createModel(data || demoData);
+    //this.createModel(data || demoData);
     // this.createModel(makePivotDataset(10_000));
     // this.createModel(makeLargeDataset(26, 10_000, ["numbers"]));
-    // this.createModel(makeLargeDataset(26, 10_000, ["formulas"]));
+    this.createModel(makeLargeDataset(26, 10_000, ["formulasSquished"]));
```
rrahir: don't forget to remove that ;-)

packages/o-spreadsheet-engine/src/plugins/core/unsquisher.ts
```diff
@@ -0,0 +1,253 @@
...
+  | "COMBINE_OFFSET"
+  | "FIRST_OFFSET"
+  | "NO_CHANGE";
+
+export class Unsquisher {
+  private previousCell: CompiledFormula | undefined;
+  private alreadyAppliedNumberOffset: number[] = [];
+  private previousString: string[] = [];
+  private alreadyAppliedReferenceOffset: Range[] = [];
+  //private previousTransformation: SquishedFormula | undefined = undefined;
```
pro-odoo: 👀

VincentSchippefilt:
robodoo rebase-ff

packages/o-spreadsheet-engine/src/formulas/compiler.ts
```diff
@@ -36,37 +38,256 @@ export const UNARY_OPERATOR_MAP = {
...
 // It is only exported for testing purposes
 export const functionCache: { [key: string]: FormulaToExecute } = {};
 
+/**
+ * A compiled formula is the result of the compilation of a formula string.
+ * It contains all the information needed to execute the formula, as well as some metadata
+ * about the formula (dependencies, literal values, symbols...) that can be used to rebuild a slightly different formula
+ * without recompiling it (for example when the formula is copied to another cell, or when we want to replace literal values but keep the same structure).
+ * */
+export class CompiledFormula implements Omit<Omit<ICompiledFormula, "tokens">, "dependencies"> {
```
LucasLefevre:
```suggestion
export class CompiledFormula implements Omit<ICompiledFormula, "tokens" | "dependencies"> {
```

packages/o-spreadsheet-engine/src/formulas/compiler.ts
```diff
@@ -36,37 +38,256 @@ export const UNARY_OPERATOR_MAP = {
...
+  public readonly rangeDependencies: Range[];
+  public hasDependencies: boolean;
```
LucasLefevre, line 67: do we still need `hasDependencies` ?
It's used only 2 times and it's equally simple to replace it with `.rangeDependencies.length`

packages/o-spreadsheet-engine/src/formulas/compiler.ts
```diff
@@ -36,37 +38,256 @@ export const UNARY_OPERATOR_MAP = {
...
+  }
+
+  areAllFunctionsExportableToExcel(): boolean {
+    if (this.isBadExpression) {
+      return false;
+    }
+    const nonExportableFunctions = iterateAstNodes(parseTokens(this.tokens)).filter(
+      (ast) => ast.type === "FUNCALL" && !functions[ast.value.toUpperCase()]?.isExported
+    );
+    return nonExportableFunctions.length === 0;
```
LucasLefevre: it was already like that before, but there's no need to build a new array
```suggestion
    const containsNonExportedFunctions = iterateAstNodes(parseTokens(this.tokens)).some(
      (ast) => ast.type === "FUNCALL" && !functions[ast.value.toUpperCase()]?.isExported
    );
    return !containsNonExportedFunctions;
```

packages/o-spreadsheet-engine/src/formulas/compiler.ts
```diff
@@ -36,37 +38,256 @@ export const UNARY_OPERATOR_MAP = {
...
+      .map((node: ASTFuncall) => ({
+        functionName: node.value.toUpperCase(),
+        args: node.args,
+      }));
+  }
+
+  isFirstNonWhitespaceToken(tokenValue: string): boolean {
+    const firstNonSpaceToken = this.tokens.find((token, i) => i > 0 && token.type !== "SPACE");
+    return (
+      firstNonSpaceToken?.type === "SYMBOL" && firstNonSpaceToken.value.toUpperCase() === tokenValue
```
LucasLefevre: the method name isn't really accurate as it checks for symbols, not any token

packages/o-spreadsheet-engine/src/formulas/compiler.ts
```diff
@@ -36,37 +38,256 @@ export const UNARY_OPERATOR_MAP = {
...
+      base.isBadExpression,
+      base.normalizedFormula,
+      compiledFormula.execute
+    );
+  }
+
+  /**
+   * Make a new instance of CompiledFormula by compiling the formula string as input by the user.
+   * */
+  static CompileFormula(formula: string, sheetId: UID, getters: CoreGetters): CompiledFormula {
```
LucasLefevre: nitpick,

`CompileFormula` => `Compile`
do we need "formula" twice in `CompiledFormula.CompileFormula(...)`

... tu vas me dire que ça fait deux fois "compile"

> *J'voulais dire " Bande de chacals, vous allez crever comme des chacals ". Mais ça faisait deux fois " chacals "...*

packages/o-spreadsheet-engine/src/formulas/compiler.ts
```diff
@@ -36,37 +38,256 @@ export const UNARY_OPERATOR_MAP = {
...
+    }
+
+    return concat(this.getTokens(getters, referenceOption).map((t) => t.value));
+  }
+
+  usesSymbol(symbol: string) {
+    return this.tokens.some(
+      (t) =>
+        t.type === "SYMBOL" &&
+        t.value.localeCompare(symbol, undefined, {
```
LucasLefevre: Should we force a locale instead of relying on the host's locale ?

If there's a bug because of that, it could be hard to reproduce/debug
```suggestion
        t.value.localeCompare(symbol, "en", {
```

packages/o-spreadsheet-engine/src/plugins/core/cell.ts
```diff
@@ -390,44 +414,47 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
...
+    offsetX: number,
+    offsetY: number,
+    compiledFormula: CompiledFormula | SerializedCompiledFormula
+  ) {
+    if (!(compiledFormula instanceof CompiledFormula)) {
+      compiledFormula = CompiledFormula.CompileForSerializedFormula(sheetId, compiledFormula);
+    }
     const adaptedDependencies = this.getters.createAdaptedRanges(
-      compileTokens(tokens).dependencies.map((d) => this.getters.getRangeFromSheetXC(sheetId, d)),
+      (compiledFormula as CompiledFormula).rangeDependencies,
```
LucasLefevre: Looks like TS is smart and casting isn't needed
```suggestion
      compiledFormula.rangeDependencies,
```

packages/o-spreadsheet-engine/src/plugins/core/squisher.ts
```diff
@@ -0,0 +1,301 @@
...
+import { getRangeString } from "../../helpers/range";
+import { Cell } from "../../types/cells";
+import { CoreGetters } from "../../types/core_getters";
+import { UID } from "../../types/misc";
+import { Range } from "../../types/range";
+
+export interface SquishedFormula {
+  N?: string; // the numbers used in the formula, ordered by position, converted to string, separated by |
+  S?: string[]; // the strings used in the formula, ordered by position
+  R?: string | string[]; // the references used in the formula, ordered by position, converted to string, separated by | if needed
```
LucasLefevre, line 26: maybe one or two small examples here are worth a thousand words

packages/o-spreadsheet-engine/src/plugins/core/squisher.ts
```diff
@@ -0,0 +1,301 @@
...
+  // for each number in the base formula, how much offset has already been applied
+  private alreadyAppliedNumberOffsets: number[] = [];
+  // for each string in the base formula, the previous string value
+  private previousStrings: string[] = [];
+  // whether the base formula was already transformed. Formulas that have already been transformed must continue to be transformed
+  private baseFormulaWasTransformed: boolean;
+
+  constructor(getters: CoreGetters) {
+    this.getters = getters;
+    this.baseFormulaWasTransformed = false;
```
LucasLefevre: nitpick
```suggestion
  private baseFormulaWasTransformed = false;

  constructor(getters: CoreGetters) {
    this.getters = getters;
```

packages/o-spreadsheet-engine/src/plugins/core/squisher.ts
```diff
@@ -0,0 +1,301 @@
...
+   * Squish the references. Result for each reference should be compared the previous formula.
+   * The result can be:
+   * - a relative change, e.g. +C2 or +R5 as string
+   * - the full reference if the reference is too different
+   * */
+  private squishReferences(references: Range[], forSheetId: UID) {
+    if (!this.baseFormula) {
+      throw new Error("No previous cell to squish against");
+    }
+    return references.map((x, index) =>
```
LucasLefevre: `x` => `range` or `reference` or `ref`

packages/o-spreadsheet-engine/src/plugins/core/squisher.ts
```diff
@@ -0,0 +1,301 @@
...
+      previousReferences[index] = deepCopy(reference);
+      return getRangeString(reference, forSheetId, this.getters.getSheetName);
+    }
+
+    if (
+      currentZone.left === currentZone.right &&
+      currentZone.top === currentZone.bottom &&
+      previousZone.left === previousZone.right &&
+      previousZone.top === previousZone.bottom
+    ) {
```
LucasLefevre: This last `if` is useless, the condition is already checked above

packages/o-spreadsheet-engine/src/plugins/core/squisher.ts
```diff
@@ -0,0 +1,301 @@
...
+        }
+      }
+      if (offset > 0) {
+        // we have found consecutive empty cells, merge them
+        const rangeKey = `${allKeys[i]}:${toXC(startKey.col, startKey.row + offset)}`;
+        result[rangeKey] = cells[allKeys[i]];
+        i += offset;
+      } else {
+        result[allKeys[i]] = cells[allKeys[i]];
+        offset = 0;
```
LucasLefevre: useless reset. It's always reset to 0 at the beginning of the for loop

packages/o-spreadsheet-engine/src/plugins/core/squisher.ts
```diff
@@ -0,0 +1,301 @@
...
+          // same column, check if consecutive row
+          if (
+            nextKey.row === startKey.row + offset + 1 &&
+            deepEquals(cells[allKeys[j]], cells[allKeys[i]])
+          ) {
+            // consecutive cell with same transformation or same value, continue
+            offset += 1;
+          } else {
+            // not consecutive, break
+            break;
```
LucasLefevre: I had a hard-time reading the nested ifs/elses/breaks, so I tried rewriting it with a while loop.
Not sure if it's better

```ts
let offset = 0;
let j = i + 1;
let nextKey = allKeys[j] ? toCartesian(allKeys[j]) : undefined;
while (
  nextKey?.col === startKey.col && // same column
  nextKey?.row === startKey.row + offset + 1 && // consecutive row
  deepEquals(cells[allKeys[j]], cells[allKeys[i]]) // same content or same transformation
) {
    offset += 1;
    j += 1;
    nextKey = allKeys[j] ? toCartesian(allKeys[j]) : undefined;
}
```
(`j` could probably be replaced with `i+1+offset`)

packages/o-spreadsheet-engine/src/plugins/core/squisher.ts
```diff
@@ -0,0 +1,301 @@
...
+          }
+        } else {
+          // different column, break
+          break;
+        }
+      }
+      if (offset > 0) {
+        // we have found consecutive empty cells, merge them
```
LucasLefevre: > empty cells

They are not empty but they follow the same pattern, right?

packages/o-spreadsheet-engine/src/plugins/core/unsquisher.ts
LucasLefevre: 👀 commented line

packages/o-spreadsheet-engine/src/plugins/core/unsquisher.ts
```diff
@@ -0,0 +1,253 @@
...
+      if (typeof current === "string") {
+        if (current.startsWith("=")) {
+          strategy = "NEW_FORMULA";
+          // compile the found formula. Reset previousCell and offsets because it's a new formula
+          const compiled = CompiledFormula.CompileFormula(current, sheetId, getters);
+          this.previousCell = compiled;
+          this.alreadyAppliedNumberOffset = compiled.literalValues.numbers.map((n) => n.value);
+          this.previousString = compiled.literalValues.strings.map((s) => s.value);
+          this.alreadyAppliedReferenceOffset = compiled.rangeDependencies.map((r) => r);
+          //this.previousTransformation = undefined;
```
LucasLefevre: 👀

packages/o-spreadsheet-engine/src/plugins/core/unsquisher.ts
```diff
@@ -0,0 +1,253 @@
...
+      }
+      if (typeof current === "string") {
+        if (current.startsWith("=")) {
+          strategy = "NEW_FORMULA";
+          // compile the found formula. Reset previousCell and offsets because it's a new formula
+          const compiled = CompiledFormula.CompileFormula(current, sheetId, getters);
+          this.previousCell = compiled;
+          this.alreadyAppliedNumberOffset = compiled.literalValues.numbers.map((n) => n.value);
+          this.previousString = compiled.literalValues.strings.map((s) => s.value);
+          this.alreadyAppliedReferenceOffset = compiled.rangeDependencies.map((r) => r);
```
LucasLefevre: At first glance, it looks like a useless `map`.
`[...brol]` better convey the intention of copying IMO
```suggestion
          this.alreadyAppliedReferenceOffset = [...compiled.rangeDependencies];
```

packages/o-spreadsheet-engine/src/plugins/core/unsquisher.ts
```diff
@@ -0,0 +1,253 @@
...
+          strategy = "NO_CHANGE";
+        }
+      }
+
+      const parts = key.split(":");
+      const positionsOfKey = parts.length === 1 ? expandXc(key) : expandRange(parts[0], parts[1]);
+      switch (strategy) {
+        case "NEW_FORMULA":
+          for (const cell of positionsOfKey) {
+            yield { position: { col: cell[0], row: cell[1] }, compiled: this.previousCell };
```
LucasLefevre: could `expandXc` and `expandRange` generate `{ col, row }` instead of `[col, row]` ?

Building all those `{ col: cell[0], row: cell[1] }` wouldn't be needed anymore

packages/o-spreadsheet-engine/src/plugins/core/unsquisher.ts
```diff
@@ -0,0 +1,253 @@
...
+          this.previousOffset.N = currentOffset.N ? currentOffset.N : this.previousOffset.N;
+          this.previousOffset.S = currentOffset.S ? currentOffset.S : this.previousOffset.S;
+          this.previousOffset.R = currentOffset.R ? currentOffset.R : this.previousOffset.R;
+          for (const cell of positionsOfKey) {
+            const result = this.unsquish(this.previousOffset, sheetId, getters);
+            yield { position: { col: cell[0], row: cell[1] }, compiled: result };
+          }
+          break;
+        case "NO_CHANGE":
+          for (const cell of positionsOfKey) {
```
LucasLefevre: This case can be combined with `case "NEW_FORMULA":`, it's exactly the same

packages/o-spreadsheet-engine/src/plugins/core/unsquisher.ts
```diff
@@ -0,0 +1,253 @@
...
+          }
+          break;
+        case "COMBINE_OFFSET":
+          if (!this.previousOffset) {
+            throw new Error("No previous offset to combine with");
+          }
+          const currentOffset = current as SquishedFormula;
+          this.previousOffset.N = currentOffset.N ? currentOffset.N : this.previousOffset.N;
+          this.previousOffset.S = currentOffset.S ? currentOffset.S : this.previousOffset.S;
+          this.previousOffset.R = currentOffset.R ? currentOffset.R : this.previousOffset.R;
```
LucasLefevre: nitpick
slightly easier to read IMO
```suggestion
          this.previousOffset.N = currentOffset.N ?? this.previousOffset.N;
          this.previousOffset.S = currentOffset.S ?? this.previousOffset.S;
          this.previousOffset.R = currentOffset.R ?? this.previousOffset.R;
```

packages/o-spreadsheet-engine/src/plugins/core/unsquisher.ts
```diff
@@ -0,0 +1,253 @@
...
+      this.previousString[index] = str;
+      return { value: str };
+    }
+  };
+
+  private adjustNumbers = (numStr: string, index: number): { value: number } => {
+    if (numStr === NO_CHANGE) {
+      return { value: this.alreadyAppliedNumberOffset[index] };
+    } else {
+      const currentOffset = parseFloat(numStr.slice(1));
```
LucasLefevre, line 239: It wasn't obvious to me at first what was this `slice(1)`.
It's worth a comment, or:
if `"-1"` was exported instead of `"+-1"` for negative offsets, we wouldn't need this `slice(1)`

packages/o-spreadsheet-engine/src/plugins/core/unsquisher.ts
```diff
@@ -0,0 +1,253 @@
...
+      if (squishedElement.S !== undefined && squishedElement.S.length > 0) {
+        current.strings = squishedElement.S.map(this.adjustStrings);
+      } else {
+        current.strings = this.previousCell.literalValues.strings;
+      }
+      if (squishedElement.R !== undefined && this.previousCell) {
+        // references
+        let references: string[];
+        if (typeof squishedElement.R !== "string") {
+          // special case when sheets names contains the separator
```
LucasLefevre:
```suggestion
          // special case when sheet names contain the separator
```

packages/o-spreadsheet-engine/src/model.ts
```diff
@@ -630,10 +635,35 @@ export class Model extends EventBus<any> implements CommandDispatcher {
...
+      snapshotRequested: false, // guarantee that no extra snapshot is requested
+      transportService: new LocalTransportService(),
+    };
+
+    const exportVerificationModel = new Model(
+      deepCopy(exportSquished), // the import itself modifies the data, so we need to deep copy it to keep the original one for comparison
+      verificationConfig
+    )._exportData(unsquished);
+    if (!deepEquals(exportUnsquished, exportVerificationModel)) {
+      exportUnsquished.isNotSquishable = true;
```
LucasLefevre, line 656: the demo dataset has `isNotSquishable: true` 😕

packages/o-spreadsheet-engine/src/plugins/core/unsquisher.ts
VincentSchippefilt: there is confusion between "-1" the number and "-1" the offset. To clear the confusion, I add the "+" before. Same for "1" the number and "+1" the offset. "+" anything is always an offset. No "+" is always an absolute number

packages/o-spreadsheet-engine/src/plugins/core/squisher.ts
VincentSchippefilt:
```suggestion
      for (offset = 0; offset+startIndex+1 < allKeys.length; offset++) {
        const nextKey = toCartesian(allKeys[offset+startIndex+1]);
        if (
          nextKey.col !== startKey.col || // different column, do not merge
          nextKey.row !== startKey.row + offset + 1 || // not consecutive, do not merge
          !deepEquals(cells[allKeys[offset+startIndex+1]], cells[allKeys[startIndex]]) // different content, do not merge
        ) {
          break;
        }
      }
```

packages/o-spreadsheet-engine/src/model.ts
VincentSchippefilt: it was the spaces. I removed all the spaces :D

packages/o-spreadsheet-engine/src/helpers/expand_range.ts
```diff
@@ -0,0 +1,30 @@
...
+ * Helper to expand a range string like A1:B3 into all cell references in the range
+ * @param start
+ * @param end
+ */
+export function* expandRange(start: string, end: string): Generator<Position> {
+  const startPos = toCartesian(start);
+  const endPos = toCartesian(end);
+  for (
+    let col = Math.min(startPos.col, endPos.col);
+    col <= Math.max(startPos.col, endPos.col);
```
rrahir, line 19: can be evaluated a single time; same for the condition below

packages/o-spreadsheet-engine/src/plugins/core/squisher.ts
```diff
@@ -0,0 +1,298 @@
...
+    this.getters = getters;
+  }
+
+  /** Build the result object based on a transformation. Joins the numbers into a single string, tries to do the same
+   * with references (unless they contain a sheet name that contains the separator). */
+  private buildResult(
+    numbers: string[],
+    strings: string[],
+    references: string[]
+  ): Partial<SquishedFormula> {
```
rrahir, line 55: useless Partial

rrahir:
robodoo r+

LucasLefevre:
robodoo r+

PR #7758, opened 2026-01-13
[IMP] added shortcut in the ActionSpec

matho-odoo:
## Description:

In order to add the shortcuts in the menu when using  Ctrl+K in Odoo, we needed to add shortcut additionaly to description in the ActionSpec

Task: [5474239](https://www.odoo.com/odoo/2328/tasks/5474239)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/actions/action.ts
+++ b/src/actions/action.ts
@@ -12,6 +12,7 @@ export interface ActionSpec {
    */
   name: string | ((env: SpreadsheetChildEnv) => string);
   description?: string | ((env: SpreadsheetChildEnv) => string);
+  shortcut?: string;
   /**
    * which represents its position inside the
    * menus (the lower sequence it has, the upper it is in the menu)
@@ -69,6 +70,7 @@ export interface ActionSpec {
 export interface Action {
   name: (env: SpreadsheetChildEnv) => string;
   description: (env: SpreadsheetChildEnv) => string;
+  shortcut: string;
   sequence: number;
   id: string;
   isVisible: (env: SpreadsheetChildEnv) => boolean;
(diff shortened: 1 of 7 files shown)


src/actions/action.ts
```diff
@@ -131,6 +134,7 @@ export function createAction(item: ActionSpec): Action {
     iconColor: item.iconColor,
     secondaryIcon: typeof secondaryIcon === "function" ? secondaryIcon : () => secondaryIcon || "",
     description: typeof description === "function" ? description : () => description || "",
```
pro-odoo, line 136: In order to avoid duplicating all shortcuts, I wonder if we could change this line to return the shortcut if the description is empty ?
```suggestion
    description: typeof description === "function" ? description : () => (description ?? shortcut) || "",
```
Or alternatively, compute it in the component (menu.xml)
```diff
- <t t-set="description" t-value="menuItem.description(env)"/>
+ <t t-set="description" t-value="menuItem.description(env) or menuItem.shortcut"/>
``` 
With that, we could remove description when it's the same as the shortcut.

EDIT: I think I would go for the second solution

tests/__snapshots__/top_bar_component.test.ts.snap
```diff
@@ -74,7 +74,7 @@ exports[`TopBar component simple rendering 1`] = `
         >
           <span
             class="o-menu-item-button o-disabled o-hoverable-button o-toolbar-button"
-            title="Undo (Ctrl+Z)"
+            title="Undo"
```
pro-odoo: These changes are not correct, it seems that we forgot a place where description is used :)

PR #7761, opened 2026-01-13
[PERF] cells: numeric cell id

LucasLefevre:
Before: cell ids are strings
After: cell ids are integers

Integers are smaller and there hash is faster to compute (for
insertion and lookup in Set).

Not by much, but if you do it 1M times...

TBH, I don't understand how it can have so much impact on
the START command (-35% for HADA's spreadsheet!)
but I didn't dig too much.

## HADHA's spreadsheet

Big fat spreadsheet on odoo.com with 600k+ formula, which
are all evaluated twice when opening the spreadsheet.

```
START
  before Mean=7037.74 ms, StdErr=134.02 ms, n=30
  after  Mean=4601.67 ms, StdErr=69.55 ms,  n=30 (-35%)

Model created in
  before Mean=13080.04 ms, StdErr=203.93 ms, n=30
  after  Mean=10562.60 ms, StdErr=144.02 ms, n=30 (-20%)

evaluate all cells
  before Mean=3614.06 ms, StdErr=44.67 ms, n=30
  after  Mean=3487.22 ms, StdErr=42.61 ms, n=30 (-5%)
```
## Large data set

```
evaluate all cells
  before: Mean: 393.64 ms, StdErr: 6.97 ms, n=30
  after:  Mean: 382.46 ms, StdErr: 5.65 ms, n=30 (-3%)

START
  before: Mean: 586.34 ms, StdErr: 9.04 ms, n=30
  after:  Mean: 562.25 ms, StdErr: 6.89 ms, n=30 (-4%)

Model created in
  before: Mean: 2956.09 ms, StdErr: 11.67 ms, n=30
  after:  Mean: 2907.31 ms, StdErr: 10.30 ms, n=30 (-2%)
```

Task: 5491227

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/packages/o-spreadsheet-engine/src/plugins/core/cell.ts
+++ b/packages/o-spreadsheet-engine/src/plugins/core/cell.ts
@@ -349,8 +349,8 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
   // ---------------------------------------------------------------------------
   // GETTERS
   // ---------------------------------------------------------------------------
-  getCells(sheetId: UID): Record<UID, Cell> {
-    return this.cells[sheetId] || {};
+  getCells(sheetId: UID): Cell[] {
+    return Object.values(this.cells[sheetId] || {});
   }
 
   /**
--- a/packages/o-spreadsheet-engine/src/plugins/ui_core_views/pivot_ui.ts
+++ b/packages/o-spreadsheet-engine/src/plugins/ui_core_views/pivot_ui.ts
@@ -334,9 +334,7 @@ export class PivotUIPlugin extends CoreViewPlugin {
     }
     const unusedPivots = new Set(this.getters.getPivotIds());
     for (const sheetId of this.getters.getSheetIds()) {
-      const cells = this.getters.getCells(sheetId);
-      for (const cellId in cells) {
-        const cell = cells[cellId];
+      for (const cell of this.getters.getCells(sheetId)) {
         const position = this.getters.getCellPosition(cell.id);
         const pivotId = this.getPivotIdFromPosition(position);
         if (pivotId) {
--- a/packages/o-spreadsheet-engine/src/plugins/core/settings.ts
+++ b/packages/o-spreadsheet-engine/src/plugins/core/settings.ts
@@ -34,9 +34,7 @@ export class SettingsPlugin extends CorePlugin {
 
   private changeCellsDateFormatWithLocale(oldLocale: Locale, newLocale: Locale) {
     for (const sheetId of this.getters.getSheetIds()) {
-      const cells = this.getters.getCells(sheetId);
-      for (const cellId in cells) {
-        const cell = cells[cellId];
+      for (const cell of this.getters.getCells(sheetId)) {
         let formatToApply: Format | undefined;
         if (cell.format === oldLocale.dateFormat) {
           formatToApply = newLocale.dateFormat;
--- a/packages/o-spreadsheet-engine/src/plugins/core/sheet.ts
+++ b/packages/o-spreadsheet-engine/src/plugins/core/sheet.ts
@@ -755,7 +755,7 @@ export class SheetPlugin extends CorePlugin<SheetState> implements SheetState {
     this.history.update("orderedSheetIds", orderedSheetIds);
     this.history.update("sheets", Object.assign({}, this.sheets, { [newSheet.id]: newSheet }));
 
-    for (const cell of Object.values(this.getters.getCells(fromId))) {
+    for (const cell of this.getters.getCells(fromId)) {
       const { sheetId, col, row } = this.getCellPosition(cell.id);
       const style = this.getters.getCellStyle({ sheetId, col, row });
       this.dispatch("UPDATE_CELL", {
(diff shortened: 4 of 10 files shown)


packages/o-spreadsheet-engine/src/plugins/core/cell.ts
```diff
@@ -47,7 +47,7 @@ import { ExcelWorkbookData, WorkbookData } from "../../types/workbook_data";
```
hokolomopo, line 47: typo in commit msg `and there hash`

packages/o-spreadsheet-engine/src/plugins/ui_core_views/pivot_ui.ts
```diff
@@ -334,8 +334,10 @@ export class PivotUIPlugin extends CoreViewPlugin {
     }
     const unusedPivots = new Set(this.getters.getPivotIds());
     for (const sheetId of this.getters.getSheetIds()) {
-      for (const cellId in this.getters.getCells(sheetId)) {
-        const position = this.getters.getCellPosition(cellId);
+      const cells = this.getters.getCells(sheetId);
+      for (const cellId in cells) {
+        const cell = cells[cellId];
+        const position = this.getters.getCellPosition(cell.id);
```
hokolomopo: getters.getCells should probably return an array (Object.values) directly, unless it's that big of a perf hit. It's really strange to have `cellId` and `cell.id` not being the same thing, and having to worry about that in 6+ places because of a perf implementation detail.

packages/o-spreadsheet-engine/src/types/commands.ts
```diff
@@ -344,7 +344,7 @@ export interface UpdateCellCommand extends PositionDependentCommand {
  */
 export interface UpdateCellPositionCommand extends PositionDependentCommand {
   type: "UPDATE_CELL_POSITION";
-  cellId?: UID;
+  cellId?: number;
```
hokolomopo, line 347: migration ? 🙃

src/clipboard_handlers/cell_clipboard.ts
```diff
@@ -61,7 +61,7 @@ export class CellClipboardHandler extends AbstractCellClipboardHandler<
             const formulaPivotId = this.getters.getPivotFormulaId(pivotId);
             const pivotFormula = createPivotFormula(formulaPivotId, pivotCell);
             cell = {
-              id: cell?.id || "",
+              id: cell?.id ?? 0,
```
hokolomopo, line 64: we should type that better sometimes and remove id here, cell.id is not used, and it would be very wrong if we were to use it by mistake

hokolomopo:
need rebase 😄

LucasLefevre:
rebased

hokolomopo (commented):
👋

packages/o-spreadsheet-engine/src/types/commands.ts
LucasLefevre: UPDATE_CELL_POSITION is never sent to the network and never saved

packages/o-spreadsheet-engine/src/plugins/ui_core_views/pivot_ui.ts
LucasLefevre: agreed sur le principe.
But it's a pity to allocate a new array for that, especially when it's in START which must be as fast as possible, same for building the dependency graph

hokolomopo: Returning an iterator ? :sweat_smile:

LucasLefevre: I'll try, but I expect it to be also slower

LucasLefevre: It looks like V8 is really smart and is able to avoid allocating the array when using `Object.values`.

Generator (`yield cells[cellId]`) slightly slower (but not significant)

```
  START
    master-number-cell-id-lul:           Mean: 5344.05 ms, StdErr: 18.40 ms, n=20 (-1%)
    master-number-cell-id-iterator-lul:  Mean: 5376.24 ms, StdErr: 26.40 ms, n=20
```

`Object.values` no difference except noise
```
  START
    master-number-cell-id-lul:           Mean: 5442.34 ms, StdErr: 21.40 ms, n=20
    master-number-cell-id-values-lul:    Mean: 5432.91 ms, StdErr: 37.94 ms, n=20 (-0%)
```

I'm not sure if relying on a low-level engine optimization is a good idea, but it makes the code cleaner indeed

packages/o-spreadsheet-engine/src/plugins/core/cell.ts
```diff
@@ -499,7 +499,7 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
   }
 
   private getNextUid() {
```
pro-odoo: Nitpick, I would rename it to `getNextCellId`, at least to remove `uid` from the name

pro-odoo: Also, we should rename `getRowCells` to `getRowCellIds`.
It's contre-intuitif that `getCells` returns Cell[] and `getRowCells` number[]
It could be done in another PR, but as there are only 3 occurrences...

LucasLefevre:
@pro-odoo updated

pro-odoo:
robodoo r+

pro-odoo:
robodoo rebase-ff

PR #7763, opened 2026-01-13
[IMP] data: add column analysis functionnality

anhe-odoo:
## Task Description

This PR aims to add a new side panel accessible from the data menu or from a column header context menu. In this side panel, we can add statistical analysis of the data in the column : 
* Count chart of distribution (if applicable (ie if the data are all numbers))
* Sum, Min, max, mean, median values (if applicable (numbers))
* Number of (uniques) value, number of empty cell, number of cell most/least frequent values

## Related Task

Task: [[Stats] add column statistical analysis](https://www.odoo.com/odoo/2328/tasks/5410989)

### DIFF
--- a/packages/o-spreadsheet-engine/src/functions/helper_statistical.ts
+++ b/packages/o-spreadsheet-engine/src/functions/helper_statistical.ts
@@ -47,6 +47,28 @@ export function average(values: Arg[], locale: Locale) {
   return sum / count;
 }
 
+export function median(values: Arg[], locale: Locale) {
+  const nums: number[] = [];
+  visitNumbers(
+    values,
+    (a) => {
+      nums.push(a.value);
+    },
+    locale
+  );
+  nums.sort((a, b) => a - b);
+  const len = nums.length;
+  if (len === 0) {
+    return undefined;
+  }
+  const mid = Math.floor(len / 2);
+  if (len % 2 === 0) {
+    return (nums[mid - 1] + nums[mid]) / 2;
+  } else {
+    return nums[mid];
+  }
+}
+
 export function countNumbers(values: Arg[], locale: Locale) {
   let count = 0;
   for (const n of values) {
--- a/src/components/side_panel/column_stats/column_stats_panel.css
+++ b/src/components/side_panel/column_stats/column_stats_panel.css
@@ -1,0 +1,24 @@
+.o-spreadsheet {
+  .o-column-stats-title {
+    font-size: 16px;
+  }
+  .o-row-switcher {
+    cursor: pointer;
+  }
+  .o-column-stats-row {
+    padding: 1px 0;
+  }
+  .o-column-stats-row:nth-child(even) {
+    background: var(--os-gray-100);
+  }
+  .o-column-frequencies-row {
+    cursor: pointer;
+  }
+  .o-column-frequencies-row:hover span {
+    font-weight: bold;
+  }
+
+  .o-column-stats-chart {
+    height: 200px;
+  }
+}
--- a/src/components/side_panel/column_stats/column_stats_panel.ts
+++ b/src/components/side_panel/column_stats/column_stats_panel.ts
@@ -1,0 +1,316 @@
+import { Highlight } from "@odoo/o-spreadsheet-engine";
+import { _t } from "@odoo/o-spreadsheet-engine/translation";
+import { SpreadsheetChildEnv } from "@odoo/o-spreadsheet-engine/types/spreadsheet_env";
+import { Component, onWillUnmount, useEffect, useRef, useState } from "@odoo/owl";
+import { Chart, ChartConfiguration } from "chart.js/auto";
+import {
+  clipTextWithEllipsis,
+  FIRST_CHART_COLOR,
+  numberToLetters,
+  positionToZone,
+} from "../../../helpers";
+import { Store, useStore } from "../../../store_engine";
+import { useHighlights } from "../../helpers/highlight_hook";
+import { NumberInput } from "../../number_input/number_input";
+import { BadgeSelection } from "../components/badge_selection/badge_selection";
+import { SidePanelCollapsible } from "../components/collapsible/side_panel_collapsible";
+import { Section } from "../components/section/section";
+import { ColumnStatisticsStore } from "./column_stats_store";
+
+interface Props {
+  onCloseSidePanel: () => void;
+}
+
+export class ColumnStatsPanel extends Component<Props, SpreadsheetChildEnv> {
+  static template = "o-spreadsheet-ColumnStatsPanel";
+  static props = { onCloseSidePanel: Function };
+  static components = { NumberInput, SidePanelCollapsible, BadgeSelection, Section };
+
+  state = useState({
+    currentChart: "count",
+    currentFrequencyOrder: "descending",
+    highlightPositions: [] as { row: number; col: number }[],
+  });
+
+  store!: Store<ColumnStatisticsStore>;
+  private chartCanvas = useRef("columnStatsChart");
+  private chart?: Chart;
+
+  setup() {
+    this.store = useStore(ColumnStatisticsStore);
+    useHighlights(this);
+    onWillUnmount(() => this.destroyChart());
+    useEffect(
+      () => {
+        this.updateChart();
+      },
+      () => [this.store.countChartData, this.store.histogramData]
+    );
+  }
+
+  get columnLabel(): string {
+    if (this.store.selectedColumn === undefined) {
+      return "";
+    }
+    const cell = this.env.model.getters.getEvaluatedCell({
+      sheetId: this.env.model.getters.getActiveSheetId(),
+      col: this.store.selectedColumn,
+      row: 0,
+    });
+    if (cell?.type === "text" && cell.value.trim() !== "") {
+      return cell.value;
+    }
+    return _t("Column %s", numberToLetters(this.store.selectedColumn));
+  }
+
+  get charts() {
+    return [
+      { value: "count", label: _t("Count"), icon: "o-spreadsheet-Icon.COUNT_CHART" },
+      { value: "histogram", label: _t("Distribution"), icon: "o-spreadsheet-Icon.COUNT_CHART" },
+    ];
+  }
+
+  get frequencyOrders() {
+    return [
+      { value: "descending", label: _t("Descending"), icon: "o-spreadsheet-Icon.DESCENDING_SORT" },
+      { value: "ascending", label: _t("Ascending"), icon: "o-spreadsheet-Icon.ASCENDING_SORT" },
+    ];
+  }
+
+  get shouldShowChart(): boolean {
+    if (this.state.currentChart === "histogram") {
+      return this.store.numericValues.length > 0;
+    } else if (this.state.currentChart === "count") {
+      return this.store.values.length > 0;
+    }
+    return false;
+  }
+
+  get chartErrorMessage(): string | null {
+    if (this.state.currentChart === "histogram" && this.store.numericValues.length === 0) {
+      return _t("No numeric values to display.");
+    }
+    if (this.state.currentChart === "count" && this.store.values.length === 0) {
+      return _t("No values to display.");
+    }
+    return null;
+  }
+
+  private getChartConfiguration(): ChartConfiguration<"bar" | "line"> | null {
+    switch (this.state.currentChart) {
+      case "histogram":
+        return this.getHistogramChartConfiguration();
+      case "count":
+        return this.getCountChartConfiguration();
+    }
+    return null;
+  }
+
+  private createChart() {
+    if (!globalThis.Chart) {
+      throw new Error("Chart.js library is not loaded");
+    }
+    const canvas = this.chartCanvas.el as HTMLCanvasElement | null;
+    if (!canvas) {
+      return;
+    }
+    const ctx = canvas.getContext("2d");
+    if (!ctx) {
+      return;
+    }
+    const config = this.getChartConfiguration();
+    if (!config) {
+      return;
+    }
+    this.chart = new globalThis.Chart!(ctx, config);
+  }
+
+  private getCountChartConfiguration(): ChartConfiguration<"bar"> | null {
+    const count = this.store.countChartData;
+    if (!count) {
+      return null;
+    }
+
+    return {
+      type: "bar",
+      data: {
+        labels: count.labels.map(this.clipTextWithEllipsis.bind(this)),
+        datasets: [
+          {
+            type: "bar",
+            data: count.data,
+            backgroundColor: FIRST_CHART_COLOR,
+            borderSkipped: false,
+            borderWidth: 1,
+            barPercentage: 1,
+            categoryPercentage: 1,
+          },
+        ],
+      },
+      options: {
+        animation: false,
+        responsive: true,
+        maintainAspectRatio: false,
+        plugins: {
+          legend: { display: false },
+          tooltip: { enabled: true },
+        },
+        scales: {
+          x: {
+            display: true,
+            ticks: {
+              maxRotation: 90,
+              minRotation: 90,
+            },
+          },
+          y: { display: true, beginAtZero: true },
+        },
+      },
+    };
+  }
+
+  private getHistogramChartConfiguration(): ChartConfiguration<"bar" | "line"> | null {
+    const histogram = this.store.histogramData;
+    if (!histogram) {
+      return null;
+    }
+    return {
+      type: "line",
+      data: {
+        labels: histogram.tooltipLabels,
+        datasets: [
+          {
+            type: "bar",
+            data: histogram.data,
+            backgroundColor: FIRST_CHART_COLOR,
+            borderSkipped: false,
+            borderWidth: 1,
+            barPercentage: 1,
+            categoryPercentage: 1,
+          },
+        ],
+      },
+      options: {
+        animation: false,
+        responsive: true,
+        maintainAspectRatio: false,
+        plugins: {
+          legend: { display: false },
+          tooltip: { enabled: true },
+        },
+        scales: {
+          x: { display: false, stacked: true, position: "top" },
+          y: { display: true },
+          x2: {
+            type: "linear",
+            position: "bottom",
+            display: true,
+            min: 0,
+            max: histogram.tooltipLabels.length,
+            ticks: {
+              callback: (value: any) =>
+                Math.floor(value) !== value
+                  ? ""
+                  : this.clipTextWithEllipsis(histogram.tickLabels?.[value]),
+              maxRotation: 90,
+              minRotation: 90,
+            },
+          },
+        },
+      },
+    };
+  }
+
+  private clipTextWithEllipsis(text: string | undefined): string {
+    if (!text) {
+      return "";
+    }
+    const canvas = this.chartCanvas.el as HTMLCanvasElement | null;
+    if (!canvas) {
+      return text;
+    }
+    const ctx = canvas.getContext("2d");
+    if (!ctx) {
+      return text;
+    }
+    return clipTextWithEllipsis(ctx, text, 75);
+  }
+
+  private updateChart() {
+    const config = this.getChartConfiguration();
+    if (!config) {
+      this.destroyChart();
+      return;
+    }
+    if (!this.chart) {
+      this.createChart();
+      return;
+    }
+    this.chart.config.options = config.options;
+    this.chart.data = config.data;
+    this.chart.update();
+  }
+
+  private destroyChart() {
+    this.chart?.destroy();
+    this.chart = undefined;
+  }
+
+  switchChart(chartType: string) {
+    this.state.currentChart = chartType;
+    this.destroyChart();
+    this.createChart();
+  }
+
+  switchFrequencyOrder(order: string) {
+    this.state.currentFrequencyOrder = order;
+  }
+
+  updateIgnoredRows(ignoredRowsStr: string) {
+    const ignoredRows = parseInt(ignoredRowsStr, 10);
+    this.store.updateIgnoredRows(isNaN(ignoredRows) ? 0 : Math.max(0, ignoredRows));
+  }
+
+  get valueFrequencies(): { value: string; count: number }[] {
+    const orderingCriterion =
+      this.state.currentFrequencyOrder === "ascending"
+        ? (a: { count: number }, b: { count: number }) => a.count - b.count
+        : (a: { count: number }, b: { count: number }) => b.count - a.count;
+    return this.store.valueFrequencies.sort(orderingCriterion).slice(0, 5);
+  }
+
+  get highlights(): Highlight[] {
+    const column = this.store.selectedColumn;
+    if (column === undefined) {
+      return [];
+    }
+    return [
+      {
+        range: this.env.model.getters.getRangeFromZone(this.env.model.getters.getActiveSheetId(), {
+          top: this.store.ignoredRows,
+          left: column,
+          bottom: undefined,
+          right: column,
+        }),
+        color: "#a3e9a39a",
+        interactive: false,
+      },
+      ...this.state.highlightPositions.map((position) => ({
+        range: this.env.model.getters.getRangeFromZone(
+          this.env.model.getters.getActiveSheetId(),
+          positionToZone(position)
+        ),
+        color: "#ffeb3b9a",
+        interactive: false,
+      })),
+    ];
+  }
+
+  highlightFrequencyPositions(positions: { row: number; col: number }[]) {
+    this.state.highlightPositions = positions;
+  }
+
+  clearHighlights() {
+    this.state.highlightPositions = [];
+  }
+}
--- a/src/components/side_panel/column_stats/column_stats_panel.xml
+++ b/src/components/side_panel/column_stats/column_stats_panel.xml
@@ -1,0 +1,94 @@
+<templates>
+  <t t-name="o-spreadsheet-ColumnStatsPanel">
+    <div class="o-column-stats-panel">
+      <t t-if="store.hasSingleColumn">
+        <Section>
+          <div class="d-flex justify-content-between align-items-center o-panel-header mb-1">
+            <span class="o-fw-bold w-100 o-column-stats-title" t-esc="columnLabel"/>
+            <span
+              class="p-1 o-row-switcher"
+              data-test-id="previous-column-button"
+              title="Switch to previous column"
+              t-on-click="store.selectPreviousColumn">
+              <t t-call="o-spreadsheet-Icon.ANGLE_LEFT"/>
+            </span>
+            <span
+              class="p-1 o-row-switcher"
+              title="Switch to next column"
+              data-test-id="next-column-button"
+              t-on-click="store.selectNextColumn">
+              <t t-call="o-spreadsheet-Icon.ANGLE_RIGHT"/>
+            </span>
+          </div>
+          <div class="mt-3">
+            <div class="o-sidePanel-label mb-1">Rows to ignore</div>
+            <NumberInput
+              value="store.ignoredRows"
+              class="'o-ignored-rows-input'"
+              min="0"
+              onChange.bind="updateIgnoredRows"
+            />
+          </div>
+        </Section>
+        <SidePanelCollapsible title.translate="Data visualization" isInitiallyCollapsed="false">
+          <t t-set-slot="content">
+            <div class="mx-4 mb-2">
+              <BadgeSelection
+                choices="charts"
+                onChange.bind="switchChart"
+                selectedValue="state.currentChart"
+              />
+              <t t-if="shouldShowChart">
+                <div class="o-column-stats-chart mx-2 mt-2">
+                  <canvas class="w-100 h-100" t-ref="columnStatsChart"/>
+                </div>
+              </t>
+              <t t-else="">
+                <div class="text-muted mt-4" t-esc="chartErrorMessage"/>
+              </t>
+            </div>
+          </t>
+        </SidePanelCollapsible>
+        <SidePanelCollapsible title.translate="General statistics" isInitiallyCollapsed="false">
+          <t t-set-slot="content">
+            <div class="o-column-global-stats d-flex flex-column gap-2 mb-2">
+              <t t-foreach="store.statItems" t-as="stat" t-key="stat.name">
+                <div class="o-column-stats-row d-flex justify-content-between">
+                  <span class="text-muted" t-esc="stat.name"/>
+                  <span class="o-fw-bold" t-att-data-test-id="stat.name" t-esc="stat.value"/>
+                </div>
+              </t>
+            </div>
+          </t>
+        </SidePanelCollapsible>
+        <SidePanelCollapsible title.translate="Frequency" isInitiallyCollapsed="true">
+          <t t-set-slot="content">
+            <div class="mx-4 mb-2">
+              <BadgeSelection
+                choices="frequencyOrders"
+                onChange.bind="switchFrequencyOrder"
+                selectedValue="state.currentFrequencyOrder"
+              />
+            </div>
+            <div class="o-column-global-stats d-flex flex-column mx-4 mb-2">
+              <t t-foreach="valueFrequencies" t-as="value" t-key="value.value">
+                <div
+                  class="o-column-stats-row o-column-frequencies-row d-flex justify-content-between"
+                  t-on-mouseenter="() => this.highlightFrequencyPositions(value.positions)"
+                  t-on-mouseleave="() => this.clearHighlights()">
+                  <span class="frequency-label text-muted" t-esc="value.value"/>
+                  <span class="frequency-value o-fw-bold" t-esc="value.count"/>
+                </div>
+              </t>
+            </div>
+          </t>
+        </SidePanelCollapsible>
+      </t>
+      <t t-else="">
+        <div class="o-column-stats-empty text-muted p-4">
+          Select a single column to view statistics.
+        </div>
+      </t>
+    </div>
+  </t>
+</templates>
--- a/src/components/side_panel/column_stats/column_stats_store.ts
+++ b/src/components/side_panel/column_stats/column_stats_store.ts
@@ -1,0 +1,347 @@
+import { sum } from "@odoo/o-spreadsheet-engine/functions/helper_math";
+import { average, max, median, min } from "@odoo/o-spreadsheet-engine/functions/helper_statistical";
+import { _t } from "@odoo/o-spreadsheet-engine/translation";
+import { formatValue, isDefined } from "../../../helpers";
+import {
+  computeStatisticFnResults,
+  SelectionStatisticFunction,
+  StatisticFnResults,
+} from "../../../helpers/selection_statistic_functions";
+import { Get } from "../../../store_engine";
+import { SpreadsheetStore } from "../../../stores";
+import {
+  CellValue,
+  CellValueType,
+  Command,
+  EvaluatedCell,
+  invalidateEvaluationCommands,
+  LocaleFormat,
+  Position,
+} from "../../../types";
+
+const columnStatisticFunctions: SelectionStatisticFunction[] = [
+  {
+    name: _t("Total rows"),
+    types: Object.values(CellValueType),
+    compute: (values, locale) => values.length,
+    format: "0",
+  },
+  {
+    name: _t("Unique values"),
+    types: [CellValueType.number, CellValueType.text, CellValueType.boolean, CellValueType.error],
+    compute: (values, locale) => {
+      const uniqueValues = new Set<string | number | boolean>();
+      for (const cell of values) {
+        uniqueValues.add(cell.value as string | number | boolean);
+      }
+      return uniqueValues.size;
+    },
+    format: "0",
+  },
+  {
+    name: _t("Sum"),
+    types: [CellValueType.number],
+    compute: (values, locale) => sum([[values]], locale),
+  },
+  {
+    name: _t("Average"),
+    types: [CellValueType.number],
+    compute: (values, locale) => average([[values]], locale),
+  },
+  {
+    name: _t("Median"),
+    types: [CellValueType.number],
+    compute: (values, locale) => median([[values]], locale) ?? "",
+  },
+  {
+    name: _t("Minimum value"),
+    types: [CellValueType.number],
+    compute: (values, locale) => min([[values]], locale).value,
+  },
+  {
+    name: _t("Maximum value"),
+    types: [CellValueType.number],
+    compute: (values, locale) => max([[values]], locale).value,
+  },
+];
+
+function buildEmptyStatisticFnResults(
+  selectionStatisticFunctions: SelectionStatisticFunction[]
+): StatisticFnResults {
+  const statisticFnResults: StatisticFnResults = {};
+  for (const fn of selectionStatisticFunctions) {
+    statisticFnResults[fn.name] = undefined;
+  }
+  return statisticFnResults;
+}
+
+interface CountChartData {
+  data: number[];
+  labels: string[];
+  positions: Position[][];
+}
+
+interface HistogramData {
+  data: number[];
+  tooltipLabels: string[];
+  tickLabels: string[];
+}
+
+interface PositionedValue<T> extends Position {
+  value: T;
+}
+
+export class ColumnStatisticsStore extends SpreadsheetStore {
+  mutators = ["updateIgnoredRows", "selectNextColumn", "selectPreviousColumn"] as const;
+  statisticFnResults: StatisticFnResults = buildEmptyStatisticFnResults(columnStatisticFunctions);
+  selectedColumn?: number;
+  numericValues: PositionedValue<number>[] = [];
+  values: PositionedValue<CellValue>[] = [];
+  dataFormat?: string;
+  countChartData?: CountChartData;
+  histogramData?: HistogramData;
+  private isDirty = false;
+  ignoredRows = 0;
+
+  constructor(get: Get) {
+    super(get);
+    this.model.selection.observe(this, {
+      handleEvent: this.refreshStatistics.bind(this),
+    });
+    this.onDispose(() => {
+      this.model.selection.unobserve(this);
+    });
+    this.refreshStatistics();
+  }
+
+  handle(cmd: Command) {
+    if (
+      invalidateEvaluationCommands.has(cmd.type) ||
+      (cmd.type === "UPDATE_CELL" && ("content" in cmd || "format" in cmd))
+    ) {
+      this.isDirty = true;
+    }
+    switch (cmd.type) {
+      case "HIDE_COLUMNS_ROWS":
+      case "UNHIDE_COLUMNS_ROWS":
+      case "GROUP_HEADERS":
+      case "UNGROUP_HEADERS":
+      case "ACTIVATE_SHEET":
+      case "ACTIVATE_NEXT_SHEET":
+      case "ACTIVATE_PREVIOUS_SHEET":
+      case "EVALUATE_CELLS":
+      case "UNDO":
+      case "REDO":
+        this.isDirty = true;
+    }
+  }
+
+  finalize() {
+    if (this.isDirty) {
+      this.isDirty = false;
+      this.refreshStatistics();
+    }
+  }
+
+  get hasSingleColumn(): boolean {
+    return this.selectedColumn !== undefined;
+  }
+
+  private computeCountChartData(): CountChartData | undefined {
+    if (this.selectedColumn === undefined) {
+      return undefined;
+    }
+    const values = this.numericValues.length ? this.numericValues : this.values;
+    if (!values.length) {
+      return undefined;
+    }
+    const countMap = new Map<string, { positions: Position[]; count: number; value: CellValue }>();
+    for (const val of values) {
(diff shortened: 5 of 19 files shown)


packages/o-spreadsheet-engine/src/functions/helper_statistical.ts
```diff
@@ -47,6 +47,28 @@ export function average(values: Arg[], locale: Locale) {
   return sum / count;
```
hokolomopo, line 47: https://drive.google.com/file/d/12zERhNLg_R4DSVF4O4WNefdgtDpONMJi/view?usp=drive_link

Strange missing vales on the chart sometimes ?

src/components/side_panel/column_stats/column_stats_store.ts
```diff
@@ -0,0 +1,221 @@
...
+  private isDirty = false;
+  private ignoreHeader = false;
+  protected highlightStore = useLocalStore(HighlightStore);
+
+  constructor(get: Get) {
+    super(get);
+    this.model.selection.observe(this, {
+      handleEvent: this.refreshStatistics.bind(this),
+    });
+    this.highlightStore.register(this);
```
hokolomopo: if you close and re-open the side panel, you don't have highlights anymore. Probably because you use useStore instead of useLocalStore ?

src/components/side_panel/column_stats/column_stats_panel.ts
```diff
@@ -0,0 +1,392 @@
...
+  get columnLabel(): string {
+    if (this.store.selectedColumn === undefined) {
+      return "";
+    }
+    const cell = this.env.model.getters.getEvaluatedCell({
+      sheetId: this.env.model.getters.getActiveSheetId(),
+      col: this.store.selectedColumn,
+      row: 0,
+    });
+    if (typeof cell?.value === "string" && cell.value.trim() !== "") {
```
hokolomopo: use cell.type instead. You probably won't need to cast the value as a string also

src/components/side_panel/column_stats/column_stats_panel.ts
```diff
@@ -0,0 +1,392 @@
...
+  }
+
+  get statItems(): StatisticItem[] {
+    const locale = this.env.model.getters.getLocale();
+    const localeFormat = { locale, format: this.store.dataFormat };
+    return Object.entries(this.store.statisticFnResults).map(([name, fnValue]) => {
+      if (fnValue === undefined) {
+        return { name, value: "—" };
+      }
+      return { name, value: formatValue(fnValue(), localeFormat) };
```
hokolomopo: if not too annoying, useless decimals shouldn't be shown

image

src/components/side_panel/column_stats/column_stats_panel.ts
```diff
@@ -0,0 +1,392 @@
...
+
+  get valueFrequencies(): { value: any; count: number }[] {
+    const count = this.count;
+    if (!count) {
+      return [];
+    }
+    const orderingCriterion =
+      this.state.currentFrequencyOrder === "ascending"
+        ? (a, b) => a.count - b.count
+        : (a, b) => b.count - a.count;
```
hokolomopo: typing

src/components/side_panel/column_stats/column_stats_panel.ts
```diff
@@ -0,0 +1,392 @@
...
+    }
+
+    return {
+      data: bins,
+      labels,
+      tickLabels,
+    };
+  }
+
+  get count(): HistogramData | undefined {
```
hokolomopo: `count` isn't great as a variable/attribute name

src/components/side_panel/column_stats/column_stats_panel.xml
```diff
@@ -0,0 +1,88 @@
...
+                choices="frequencyOrders"
+                onChange.bind="switchFrequencyOrder"
+                selectedValue="state.currentFrequencyOrder"
+              />
+            </div>
+            <div class="o-column-global-stats d-flex flex-column gap-2 mb-2">
+              <t t-foreach="valueFrequencies" t-as="value" t-key="value.value">
+                <div class="o-column-stats-row d-flex justify-content-between">
+                  <span class="frequency_label text-muted" t-esc="value.value"/>
+                  <span class="frequency_value o-fw-bold" t-esc="value.count"/>
```
hokolomopo:
```suggestion
                  <span class="frequency-label text-muted" t-esc="value.value"/>
                  <span class="frequency-value o-fw-bold" t-esc="value.count"/>
```

src/components/side_panel/column_stats/column_stats_store.ts
```diff
@@ -0,0 +1,221 @@
...
+      this.numericValues = [];
+      this.statisticFnResults = buildEmptyStatisticFnResults(columnStatisticFunctions);
+      this.dataFormat = undefined;
+      return;
+    }
+
+    const sheetId = getters.getActiveSheetId();
+    const column = getters.getActivePosition().col;
+    this.selectedColumn = column;
+    const formatsInDataset = getters
```
hokolomopo: getters.getRangeFormats & getters.getColsZone

src/components/side_panel/column_stats/column_stats_store.ts
```diff
@@ -0,0 +1,221 @@
...
+    const cells: EvaluatedCell[] = [];
+    const numericValues: number[] = [];
+    const values: any[] = [];
+
+    let headerIgnored = false;
+
+    for (let row = 0; row < numberOfRows; row++) {
+      if (getters.isRowHidden(sheetId, row) || getters.isColHidden(sheetId, column)) {
+        continue;
+      }
```
hokolomopo, line 284: use getRangeValues, it already skips hidden cells

src/components/side_panel/column_stats/column_stats_panel.ts
```diff
@@ -0,0 +1,392 @@
...
+    if (this.state.currentChart === "histogram" && this.store.numericValues.length === 0) {
+      return _t("No numeric values to display.");
+    }
+    if (this.state.currentChart === "count" && this.store.values.length === 0) {
+      return _t("No values to display.");
+    }
+    return null;
+  }
+
+  get histogram(): HistogramData | undefined {
```
hokolomopo: try to move most of the logic of the component in the store if possible

src/components/side_panel/column_stats/column_stats_panel.ts
```diff
@@ -0,0 +1,392 @@
...
+    } else {
+      for (const value of values) {
+        const ratio = (value - minValue) / range;
+        const rawIndex = Math.floor(ratio * barCount);
+        const index = Math.min(barCount - 1, Math.max(0, rawIndex));
+        bins[index] += 1;
+      }
+    }
+
+    const localeFormat = {
```
hokolomopo: maybe create a getter `get localeFormat` in the store, you need it like 3-4  times

src/components/side_panel/column_stats/column_stats_panel.ts
```diff
@@ -0,0 +1,392 @@
...
+      .sort((a, b) => a[0] - b[0])
+      .forEach(([val, count]) => {
+        labels.push(formatValue(val, localeFormat));
+        data.push(count);
+      });
+
+    return {
+      data,
+      labels,
+    };
```
hokolomopo:
```suggestion
    return {data,labels,};
```
omega nitpick. Same in other returns. Feel free to ignore me

src/components/side_panel/column_stats/column_stats_panel.ts
```diff
@@ -0,0 +1,392 @@
...
+    const canvas = this.chartCanvas.el as HTMLCanvasElement | null;
+    if (!canvas) {
+      return;
+    }
+    const ctx = canvas.getContext("2d");
+    if (!ctx) {
+      return;
+    }
+    const chartData: ChartConfiguration<"bar" | "line"> = {
+      type: "line",
```
hokolomopo: why line ?

src/components/side_panel/column_stats/column_stats_panel.ts
```diff
@@ -0,0 +1,392 @@
...
+        responsive: true,
+        maintainAspectRatio: false,
+        plugins: {
+          legend: { display: false },
+          tooltip: { enabled: true },
+        },
+        scales: {
+          x: { display: false, stacked: true, position: "top" },
+          y: { display: true },
+          x2: {
```
hokolomopo, line 204: why do we need 2 X axis ?

src/components/side_panel/column_stats/column_stats_panel.ts
```diff
@@ -0,0 +1,392 @@
...
+            type: "linear",
+            position: "bottom",
+            display: true,
+            min: 0,
+            max: 2 * histogram.labels.length,
```
hokolomopo: Mmmh i'll trust you on that, but maybe a  small comment on why 2 * length ?

src/components/side_panel/column_stats/column_stats_panel.ts
```diff
@@ -0,0 +1,392 @@
...
+            ticks: {
+              callback: (value: any) =>
+                Math.floor(value) !== value ? "" : this.tickLabels?.[value] ?? "",
```
hokolomopo: why do you need this.tickLabels ? Can't you just use histogram.ticklabels ?

src/components/side_panel/column_stats/column_stats_panel.ts
hokolomopo: same for histogram. histogramDataset ?

src/components/side_panel/column_stats/column_stats_panel.ts
```diff
@@ -0,0 +1,392 @@
...
+      return;
+    }
+    this.chart.data.labels = count.labels;
+    if (this.chart.data.datasets[0]) {
+      this.chart.data.datasets[0].data = count.data;
+    }
+    this.chart.update();
+  }
+
+  private updateHistogramChart() {
```
hokolomopo: You could use the same method for both types of charts. Just override all the options of this.chart for both, and pass the chart data as argument

src/components/side_panel/column_stats/column_stats_panel.xml
```diff
@@ -0,0 +1,88 @@
+<templates>
+  <t t-name="o-spreadsheet-ColumnStatsPanel">
+    <div class="o-column-stats-panel">
+      <t t-if="store.hasSingleColumn">
+        <div class="d-flex justify-content-between align-items-center o-panel-header px-4 py-3">
```
hokolomopo: shouldn't you use a Section instead of a manual x padding on all the elements ?

src/components/side_panel/column_stats/column_stats_panel.ts
```diff
@@ -0,0 +1,392 @@
...
+    this.createChart();
+  }
+
+  switchFrequencyOrder(order: string) {
+    this.state.currentFrequencyOrder = order;
+  }
+
+  toggleIgnoreHeader(ignoreHeader: boolean) {
+    this.state.ignoreHeader = ignoreHeader;
+    this.store.updateIgnoreHeader(ignoreHeader);
```
hokolomopo: i'd store ignoreHeader in either the state or the store, not both

src/components/side_panel/column_stats/column_stats_panel.xml
```diff
@@ -0,0 +1,88 @@
...
+            <div class="mx-4 mb-2">
+              <BadgeSelection
+                choices="frequencyOrders"
+                onChange.bind="switchFrequencyOrder"
+                selectedValue="state.currentFrequencyOrder"
+              />
+            </div>
+            <div class="o-column-global-stats d-flex flex-column gap-2 mb-2">
+              <t t-foreach="valueFrequencies" t-as="value" t-key="value.value">
+                <div class="o-column-stats-row d-flex justify-content-between">
```
hokolomopo: should we just not display the empty values ? 
image

src/components/side_panel/column_stats/column_stats_store.ts
```diff
@@ -0,0 +1,221 @@
...
+  EvaluatedCell,
+  Highlight,
+  invalidateEvaluationCommands,
+} from "../../../types";
+
+const columnStatisticFunctions: SelectionStatisticFunction[] = [
+  {
+    name: _t("Total rows"),
+    types: Object.values(CellValueType),
+    compute: (values, locale) => values.length.toString(),
```
hokolomopo: why do you toString the first few stats, but not sum/avg/.. ?

src/components/side_panel/column_stats/column_stats_store.ts
```diff
@@ -0,0 +1,221 @@
...
+    if (this.selectedColumn === undefined) {
+      return [];
+    }
+    const column = numberToLetters(this.selectedColumn);
+    return [
+      {
+        range: this.getters.getRangeFromSheetXC(
+          this.getters.getActiveSheetId(),
+          `${column}:${column}`
+        ),
```
hokolomopo: getRangeFromZone(getters.getColumnsZone(...))

src/components/side_panel/column_stats/column_stats_store.ts
```diff
@@ -0,0 +1,221 @@
...
+        color: "#5c9e5cff",
```
hokolomopo: do we want another color than the other highlights ?

src/components/side_panel/column_stats/column_stats_store.ts
```diff
@@ -0,0 +1,221 @@
...
+        interactive: true,
```
hokolomopo: interactive ?

src/components/side_panel/column_stats/column_stats_panel.xml
```diff
@@ -0,0 +1,88 @@
...
+          onChange.bind="toggleIgnoreHeader"
+          className="'px-4 w-100 pb-2'"
+          label="'Ignore header'"
+          title="'Ignore first non-empty row'"
+        />
+        <SidePanelCollapsible title.translate="Data visualization" isInitiallyCollapsed="true">
+          <t t-set-slot="content">
+            <div class="mx-4 mb-2">
+              <BadgeSelection
+                choices="charts"
```
hokolomopo, line 37: could we imagine a different icon for both choices ? 
image

src/components/side_panel/column_stats/column_stats_store.ts
```diff
@@ -0,0 +1,221 @@
...
+const columnStatisticFunctions: SelectionStatisticFunction[] = [
+  {
+    name: _t("Total rows"),
+    types: Object.values(CellValueType),
+    compute: (values, locale) => values.length.toString(),
+  },
+  {
+    name: _t("Empty cells"),
+    types: [CellValueType.empty],
+    compute: (values, locale) => values.length.toString(),
```
hokolomopo: you copied gsheet, but I really don't see the utility of this ... for 99% of datasets, the rows below the data table will be empty and will mess up this metric

tests/column_statistics/column_statistics_side_panel.test.ts
```diff
@@ -0,0 +1,277 @@
...
+    expect(values).toEqual(["2", "1", "1", "1", "1"]);
+  });
+
+  test("Column stats side panel is correctly filled for string data", async () => {
+    setGrid(model, { A1: "a", A2: "b", A3: "c", A4: "d", A5: "e", A6: "a" });
+    selectCell(model, "A1");
+    env.openSidePanel("ColumnStats");
+    await nextTick();
+
+    expect(fixture.querySelector('[data-test-id="Total rows"]')?.textContent).toBe("100");
```
hokolomopo:
```suggestion
    expect('[data-test-id="Total rows").toHaveText("100");
```
same everywhere else

tests/column_statistics/column_statistics_side_panel.test.ts
```diff
@@ -0,0 +1,277 @@
...
+    expect(fixture.querySelector('[data-test-id="Sum"]')?.textContent).toBe("—");
+  });
+
+  test("Column stats side panel is showing an error message when selecting multiple columns", async () => {
+    setGrid(model, { A1: "10", A2: "20", B1: "30", B2: "40" });
+    setSelection(model, ["A1", "B1"]);
+    env.openSidePanel("ColumnStats");
+    await nextTick();
+
+    expect(fixture.querySelector('[data-test-id="Total rows"]')).toBeNull();
```
hokolomopo:
```suggestion
    expect(('[data-test-id="Total rows"]')).toHaveCount(0);
```

hokolomopo:
Also use the same tooltip as our other charts :D

src/components/side_panel/column_stats/column_stats_panel.ts
anhe-odoo: Because I need a second linear axis and the typing is not "content" if it's not a line chart

src/components/side_panel/column_stats/column_stats_panel.ts
anhe-odoo: To be able to show the boundaries of the range between the bars, and I couldn't find another way to do it

src/components/side_panel/column_stats/column_stats_panel.xml
anhe-odoo: No, it was a choice from FRGI

src/components/side_panel/column_stats/column_stats_store.ts
anhe-odoo: `useLocalStore` keep the highlight even after exiting the side panel. But i'll have la look at it

src/components/side_panel/column_stats/column_stats_store.ts
anhe-odoo: But here I'm computing the array of values and the array of cells at once

src/components/side_panel/column_stats/column_stats_panel.ts
anhe-odoo: Because when the data are updated the histogram is fully rebuilt, then I only update the chart data and this.tickLabels for the tooltip

src/components/side_panel/column_stats/column_stats_store.ts
anhe-odoo: To have the first 3 as full integer ignoring the dataset format

src/components/side_panel/column_stats/column_stats_store.ts
anhe-odoo:
```js
range: this.getters.getRangeFromZone(
          this.getters.getActiveSheetId(),
          {
            top: 0,
            left: this.selectedColumn,
            bottom: undefined,
            right: this.selectedColumn,
         }
        ),
``` ?

File: src/components/side_panel/column_stats/column_stats_store.ts (hunk 21 above)
Comment (hokolomopo) [reply src/components/side_panel/column_stats/column_stats_store.ts]: So the string formats of the cell will apply no ? `;;;@"myStringFormat"`

File: src/components/side_panel/column_stats/column_stats_panel.ts (hunk 16 above)
Comment (hokolomopo) [reply src/components/side_panel/column_stats/column_stats_panel.ts]: You should probably just rebuild the full runtime and apply it again to the chart, just like we do for the spreadsheet's charts

File: src/components/side_panel/column_stats/column_stats_store.ts (hunk 22 above)
Comment (hokolomopo) [reply src/components/side_panel/column_stats/column_stats_store.ts]: On one hand I'm not a fan of building XCs where we can avoid it, on the other hand it was shorter :sweat_smile: Idk, pick whichever one you prefer

File: src/components/side_panel/column_stats/column_stats_store.ts (hunk 21 above)
Comment (anhe-odoo) [reply src/components/side_panel/column_stats/column_stats_store.ts]: No, it won't
image

Comment (hokolomopo) [reply src/components/side_panel/column_stats/column_stats_store.ts]: Yes it does :sweat_smile: 

image

Comment (hokolomopo) [reply src/components/side_panel/column_stats/column_stats_store.ts]: up

**Review by hokolomopo — COMMENTED** [review]
:wave: 

almost good for me, added a note in the task pad

File: src/components/side_panel/column_stats/column_stats_panel.ts [hunk 29]
```diff
@@ -0,0 +1,337 @@
...
+    this.destroyChart();
+    this.createChart();
+  }
+
+  switchFrequencyOrder(order: string) {
+    this.state.currentFrequencyOrder = order;
+  }
+
+  get rowsToIgnoreLabel() {
+    return _t("Rows to ignore");
```
LucasLefevre: move this to the template

src/components/side_panel/column_stats/column_stats_panel.xml
```diff
@@ -0,0 +1,94 @@
...
+            <span
+              class="p-1 o-row-switcher"
+              t-att-title="'Switch to next column'"
+              data-test-id="next-column-button"
+              t-on-click="store.selectNextColumn">
+              <t t-call="o-spreadsheet-Icon.ANGLE_RIGHT"/>
+            </span>
+          </div>
+          <div class="mt-3">
+            <div class="o-sidePanel-label mb-1" t-esc="rowsToIgnoreLabel"/>
```
LucasLefevre:
```suggestion
            <div class="o-sidePanel-label mb-1">Rows to ignore</div>
```

src/components/side_panel/column_stats/column_stats_panel.xml
```diff
@@ -0,0 +1,94 @@
...
+  <t t-name="o-spreadsheet-ColumnStatsPanel">
+    <div class="o-column-stats-panel">
+      <t t-if="store.hasSingleColumn">
+        <Section>
+          <div class="d-flex justify-content-between align-items-center o-panel-header mb-1">
+            <span class="o-fw-bold w-100 o-column-stats-title" t-esc="columnLabel"/>
+            <span
+              class="p-1 o-row-switcher"
+              data-test-id="previous-column-button"
+              t-att-title="'Switch to previous column'"
```
LucasLefevre: if it's a dynamic attribute, it's not translated. And here it doesn't need to be dynamic
```suggestion
              title="Switch to previous column"
```

src/components/side_panel/column_stats/column_stats_panel.xml
```diff
@@ -0,0 +1,94 @@
...
+              t-on-click="store.selectPreviousColumn">
+              <t t-call="o-spreadsheet-Icon.ANGLE_LEFT"/>
+            </span>
+            <span
+              class="p-1 o-row-switcher"
+              t-att-title="'Switch to next column'"
```
LucasLefevre:
```suggestion
              title="Switch to next column"
```

src/components/side_panel/column_stats/column_stats_store.ts
```diff
@@ -0,0 +1,319 @@
...
+    if (this.selectedColumn === undefined) {
+      return undefined;
+    }
+    const values = this.numericValues.length ? this.numericValues : this.values;
+    if (!values.length) {
+      return undefined;
+    }
+    const countMap = new Map<
+      string,
+      { positions: { row: number; col: number }[]; count: number; value: number }
```
LucasLefevre:
```suggestion
      { positions: Position[]; count: number; value: number }
```

src/components/side_panel/column_stats/column_stats_store.ts
```diff
@@ -0,0 +1,319 @@
...
+
+  get ignoredRowsCount(): number {
+    return this.ignoredRows;
+  }
+
+  get countChartData():
+    | {
+        data: number[];
+        labels: string[];
+        positions: { row: number; col: number }[][];
```
LucasLefevre:
```suggestion
        positions: Position[][];
```

src/components/side_panel/column_stats/column_stats_store.ts
```diff
@@ -0,0 +1,319 @@
...
+        positions: [
+          ...(countMap.get(formattedValue)?.positions ?? []),
+          { row: val.row, col: val.col },
+        ],
+        count: (countMap.get(formattedValue)?.count ?? 0) + 1,
+        value: val.value,
+      });
+    }
+    const data: number[] = [];
+    const positions: { row: number; col: number }[][] = [];
```
LucasLefevre:
```suggestion
    const positions: Position[][] = [];
```

src/components/side_panel/column_stats/column_stats_store.ts
```diff
@@ -0,0 +1,319 @@
...
+      }
+    }
+
+    return { data, labels, tickLabels };
+  }
+
+  get valueFrequencies(): {
+    value: any;
+    count: number;
+    positions: { row: number; col: number }[];
```
LucasLefevre:
```suggestion
    positions: Position[];
```

src/components/side_panel/column_stats/column_stats_store.ts
```diff
@@ -0,0 +1,319 @@
...
+      { positions: { row: number; col: number }[]; count: number; value: number }
+    >();
+    for (const val of values) {
+      const formattedValue =
+        typeof val.value === "number" ? formatValue(val.value, this.localeFormat) : val.value;
+      countMap.set(formattedValue, {
+        positions: [
+          ...(countMap.get(formattedValue)?.positions ?? []),
+          { row: val.row, col: val.col },
+        ],
```
LucasLefevre: this is quadratic. Don't rebuild a new array for each new position with the same formattedValue

src/components/side_panel/column_stats/column_stats_panel.ts
```diff
@@ -0,0 +1,337 @@
...
+  private chartCanvas = useRef("columnStatsChart");
+  private chart?: Chart;
+
+  setup() {
+    this.store = useStore(ColumnStatisticsStore);
+    useHighlights(this);
+    onMounted(() => this.createChart());
+    onWillUnmount(() => this.destroyChart());
+    useEffect(() => {
+      this.updateChart();
```
LucasLefevre: Updating the chart at each render is too much.

I did a column with 10k values `=SEQUENCE(10000)` and open the side panel.
Scrolling is suuuuper slow and it ends up crashing the browser

src/components/side_panel/column_stats/column_stats_store.ts
```diff
@@ -0,0 +1,319 @@
...
+        data[index] += 1;
+      }
+    }
+
+    const localeFormat = this.localeFormat;
+
+    const tickLabels: string[] = [];
+    const labels: string[] = [];
+    for (let i = 0; i <= barCount; i++) {
+      tickLabels.push(`${formatValue(minValue + i * step, localeFormat)}`);
```
LucasLefevre:
```suggestion
      tickLabels.push(formatValue(minValue + i * step, localeFormat));
```

src/components/side_panel/column_stats/column_stats_store.ts
```diff
@@ -0,0 +1,319 @@
...
+          `${formatValue(value, localeFormat)}-${formatValue(value + step, localeFormat)}`
+        );
+      }
+    }
+
+    return { data, labels, tickLabels };
+  }
+
+  get valueFrequencies(): {
+    value: any;
```
LucasLefevre: 👀 🚨 `any`

src/components/side_panel/column_stats/column_stats_store.ts
```diff
@@ -0,0 +1,319 @@
...
+        bottom: getters.getNumberRows(sheetId) - 1,
+        right: col,
+      })
+    );
+    this.dataFormat = formatsInDataset.find(isDefined) ?? "0.00";
+
+    const numberOfRows = getters.getNumberRows(sheetId);
+    const cells: EvaluatedCell[] = [];
+    const numericValues: { row: number; col: number; value: number }[] = [];
+    const values: { row: number; col: number; value: any }[] = [];
```
LucasLefevre: 👀 🚨 `any`

src/components/side_panel/column_stats/column_stats_panel.ts
```diff
@@ -0,0 +1,337 @@
...
+  get rowsToIgnoreLabel() {
+    return _t("Rows to ignore");
+  }
+
+  updateIgnoredRows(ignoredRowsStr: string) {
+    const ignoredRows = parseInt(ignoredRowsStr, 10);
+    this.store.updateIgnoredRows(isNaN(ignoredRows) ? 0 : Math.max(0, ignoredRows));
+  }
+
+  get valueFrequencies(): { value: any; count: number }[] {
```
LucasLefevre: 👀 🚨 `any`

src/components/side_panel/column_stats/column_stats_store.ts
```diff
@@ -0,0 +1,319 @@
...
+  selectedColumn?: number;
+  numericValues: {
+    row: number;
+    col: number;
+    value: number;
+  }[] = [];
+  values: {
+    row: number;
+    col: number;
+    value: any;
```
LucasLefevre: 👀 🚨 `any`

src/components/side_panel/column_stats/column_stats_store.ts
```diff
@@ -0,0 +1,319 @@
...
+
+      const evaluatedCell = getters.getEvaluatedCell({ sheetId, col: col, row });
+      cells.push(evaluatedCell);
+      if (
+        evaluatedCell.type !== CellValueType.empty &&
+        evaluatedCell.type !== CellValueType.error
+      ) {
+        values.push({ row, col, value: evaluatedCell.value });
+        if (evaluatedCell.type === CellValueType.number) {
+          numericValues.push({ row, col, value: evaluatedCell.value as number });
```
LucasLefevre: casting is not needed
```suggestion
          numericValues.push({ row, col, value: evaluatedCell.value});
```

src/helpers/selection_statistic_functions.ts
```diff
@@ -0,0 +1,56 @@
...
+}
+
+export interface SelectionStatisticFunction {
+  name: string;
+  compute: (data: EvaluatedCell[], locale: Locale) => number | string;
+  types: CellValueType[];
+  format?: string;
+}
+
+export function buildEmptyStatisticFnResults(
```
LucasLefevre: move this in the only file where it's used

tests/column_statistics/column_statistics_side_panel.test.ts
```diff
@@ -0,0 +1,269 @@
...
+    );
+    expect(labels).toEqual(["a", "b", "c"]);
+
+    await click(fixture, 'button[data-id="ascending"]');
+
+    labels = Array.from(fixture.querySelectorAll(".frequency-label")).map((el) => el.textContent);
+    expect(labels).toEqual(["c", "b", "a"]);
+  });
+
+  test.each(["#,##0.00", "[$$]#,##0.00", "0.00%"])(
```
LucasLefevre: no need to test 3 formats. Only one is enough.

tests/column_statistics/column_statistics_side_panel.test.ts
```diff
@@ -0,0 +1,269 @@
...
+import {
+  mockChart,
+  mountComponentWithPortalTarget,
+  nextTick,
+  setGrid,
+} from "../test_helpers/helpers";
+
+mockChart();
+
+describe("column statistics sidePanel component", () => {
```
LucasLefevre, line 16: Some/most tests could be testing the store instead of the UI. They are much faster, easier to write, easier to debug.
There's 300+ LoC of business logic in the store. It can be tested independently from the UI (slow and chiant to debug)

tests/column_statistics/column_statistics_side_panel.test.ts
```diff
@@ -0,0 +1,269 @@
...
+    expect(values).toEqual(["2", "1", "1", "1", "1"]);
+  });
+
+  test("Column stats side panel is correctly filled for empty data", async () => {
+    selectCell(model, "A1");
+    env.openSidePanel("ColumnStats");
+    await nextTick();
+
+    expect('[data-test-id="Total rows"]').toHaveText("100");
+    expect('[data-test-id="Unique values"]').toHaveText("—");
```
LucasLefevre: really not a fan of this `data-test-id` and having stuff in the business code just for testing purposes

anhe-odoo: I agree with you, but before, we were having lots of class in template that were only there for testing purpose, so I personally prefer this way to access to the element. Maybe we can try to access them with a more complex query, but I'm not sure it's a good idea too ... If any, I'm totally open to other suggestions

src/components/side_panel/column_stats/column_stats_panel.xml
anhe-odoo: Not really, both are bar chart :/

src/components/side_panel/column_stats/column_stats_store.ts
```diff
@@ -0,0 +1,332 @@
...
+    }
+  }
+
+  get hasSingleColumn(): boolean {
+    return this.selectedColumn !== undefined;
+  }
+
+  get ignoredRowsCount(): number {
+    return this.ignoredRows;
+  }
```
LucasLefevre: getter not needed. Just make `this.ignoredRows` a public attribute

src/components/side_panel/column_stats/column_stats_panel.css
```diff
@@ -5,10 +5,16 @@
   .o-row-switcher {
     cursor: pointer;
```
LucasLefevre, line 6: why a pointer ? It's not clickable 

EDIT: it's for the highlights, I saw it after.

src/components/side_panel/column_stats/column_stats_store.ts
```diff
@@ -0,0 +1,332 @@
...
+    return { data, labels, positions };
+  }
+
+  private computeHistogramData(): HistogramData | undefined {
+    const values = this.numericValues;
+    if (!values.length) {
+      return undefined;
+    }
+
+    const barCount = 1 + Math.floor(Math.log2(Math.max(new Set(values).size, 1)));
```
LucasLefevre: `Math.max` is useless here because there's at least one element in the set ?

LucasLefevre: Is it expected that `values` are objects here ? When you create a set out of it, they all end up in the set

LucasLefevre: There's an issue somewhere:

Image

Here, I have 100k times the formula `=RANDBETWEEN(-10, 20)` 

The distribution should be uniform. Either the chart is not correctly computed or `RANDBETWEEN` is not uniform.
Could you check ?

src/components/side_panel/column_stats/column_stats_store.ts
```diff
@@ -0,0 +1,332 @@
...
+        const rawIndex = Math.floor(ratio * barCount);
+        const index = Math.min(barCount - 1, Math.max(0, rawIndex));
+        data[index] += 1;
+      }
+    }
+
+    const localeFormat = this.localeFormat;
+
+    const tickLabels: string[] = [];
+    const labels: string[] = [];
```
LucasLefevre: It's not obvious to me what's the difference `labels` and `tickLabels`.
Maybe you can rename `labels` to be more precise

src/components/side_panel/column_stats/column_stats_store.ts
```diff
@@ -0,0 +1,332 @@
...
+    const step = range / barCount;
+    const data = new Array(barCount).fill(0);
+
+    if (range === 0) {
+      data[0] = values.length;
+    } else {
+      for (const value of values) {
+        const ratio = (value.value - minValue) / range;
+        const rawIndex = Math.floor(ratio * barCount);
+        const index = Math.min(barCount - 1, Math.max(0, rawIndex));
```
LucasLefevre: Is `Math.max(0, ...` needed ? `rawIndex` cannot be negative, right ?

src/components/side_panel/column_stats/column_stats_panel.ts
```diff
@@ -0,0 +1,335 @@
...
+      col: this.store.selectedColumn,
+      row: 0,
+    });
+    if (cell?.type === "text" && cell.value.trim() !== "") {
+      return cell.value;
+    }
+    return _t("Column %s", numberToLetters(this.store.selectedColumn));
+  }
+
+  get statItems(): StatisticItem[] {
```
LucasLefevre: could be in the store

And `localeFormat` can be made private if it's moved

src/components/side_panel/column_stats/column_stats_panel.xml
```diff
@@ -0,0 +1,94 @@
...
+        <SidePanelCollapsible title.translate="Data visualization" isInitiallyCollapsed="false">
+          <t t-set-slot="content">
+            <div class="mx-4 mb-2">
+              <BadgeSelection
+                choices="charts"
+                onChange.bind="switchChart"
+                selectedValue="state.currentChart"
+              />
+              <t t-if="shouldShowChart">
+                <div class="o-column-stats-chart mx-2 mt-2" style="height: 200px;">
```
LucasLefevre: `style` should be in the css file

tests/column_statistics/column_statistics_store.test.ts
```diff
@@ -0,0 +1,179 @@
...
+import {
+  deleteRows,
+  selectCell,
+  setCellContent,
+  setCellFormat,
+} from "../test_helpers/commands_helpers";
+import { mountSpreadsheet, setGrid } from "../test_helpers/helpers";
+
+function getStoreStatistics(store: Store<ColumnStatisticsStore>): Record<string, any> {
+  const results: Record<string, any> = {};
```
LucasLefevre:
```suggestion
  const results: Record<string, number> = {};
```

src/helpers/selection_statistic_functions.ts
```diff
@@ -0,0 +1,46 @@
...
+  locale: Locale
+): StatisticFnResults {
+  const statisticFnResults: StatisticFnResults = {};
+  const getCells = ((typesMap: Record<string, EvaluatedCell[]>) => (typeStr: string) => {
+    if (!typesMap[typeStr]) {
+      const types = typeStr.split(",");
+      typesMap[typeStr] = cells.filter((c) => types.includes(c.type));
+    }
+    return typesMap[typeStr];
+  })({});
```
LucasLefevre: A immediately-called function that returns a function isn't easy to read. 
The previous version with `memoize` was easier to read IMO

src/components/side_panel/column_stats/column_stats_panel.ts
```diff
@@ -0,0 +1,335 @@
...
+        range: this.env.model.getters.getRangeFromZone(this.env.model.getters.getActiveSheetId(), {
+          top: this.store.ignoredRowsCount,
+          left: column,
+          bottom: undefined,
+          right: column,
+        }),
+        color: "#a3e9a39a",
+        interactive: false,
+      },
+      ...this.state.highlightPositions.map((pos) => ({
```
LucasLefevre:
```suggestion
      ...this.state.highlightPositions.map((position) => ({
```

src/components/side_panel/column_stats/column_stats_panel.ts
```diff
@@ -0,0 +1,335 @@
...
+        range: this.env.model.getters.getRangeFromZone(this.env.model.getters.getActiveSheetId(), {
+          top: pos.row,
+          left: pos.col,
+          bottom: pos.row,
+          right: pos.col,
+        }),
```
LucasLefevre:
```suggestion
        range: this.env.model.getters.getRangeFromZone(this.env.model.getters.getActiveSheetId(), positionToZone(position)),
```

src/components/side_panel/column_stats/column_stats_store.ts
anhe-odoo: It's an issue with the chart. But it depends on the number of the "intervales" and the size of them (which depends on the number). Choosing the correct number to have the best chart is quite a real complex problem (https://en.wikipedia.org/wiki/Histogram#Number_of_bins_and_width). I went for an easy/well-known solution (Sturges's rule) but it's still not perfect. I'll have a look in the code to be sure the issue comes from there, but if it's the case I'm not sure we can do something about it :/

src/components/side_panel/column_stats/column_stats_store.ts
anhe-odoo: labels are used for tooltip and are like "min-max" while ticksLabels are the labels in the x axis' ticks. I can maybe use "barLabel" and "ticksLabel" ? Or "tooltipLabel" and "ticksLabel" ?

src/components/side_panel/column_stats/column_stats_store.ts
LucasLefevre: If we understand where it comes from and it's a "known" limitation, then ok.
Could you add a comment next to your code saying it's Sturges's rule ?

LucasLefevre: > Sturges's rule [...] implicitly assumes an approximately normal distribution.

we cannot assume a normal distribution, can we ?

anhe-odoo: Not on your data coming from RANDBETWEEN, but it depends on the data :/ But I can try with the other rules (like Doane's formula which is supposed to be a better but more complex solution), at least to check if the results seems visually better

src/components/side_panel/column_stats/column_stats_store.ts
```diff
@@ -0,0 +1,347 @@
...
+    return { data, labels, positions };
+  }
+
+  private computeHistogramData(): HistogramData | undefined {
+    const values = this.numericValues.map((v) => v.value);
+    if (!values.length) {
+      return undefined;
+    }
+
+    const barCount = 1 + Math.floor(Math.log2(new Set(values).size)); // Sturge's formula for optimal bin count
```
LucasLefevre: Since values are all different objects (?), feeding them into a set is useless (?)
```suggestion
    const barCount = 1 + Math.floor(Math.log2(values.length)); // Sturge's formula for optimal bin count
```

src/components/side_panel/column_stats/column_stats_store.ts
```diff
@@ -0,0 +1,347 @@
...
+
+  get hasSingleColumn(): boolean {
+    return this.selectedColumn !== undefined;
+  }
+
+  private computeCountChartData(): CountChartData | undefined {
+    if (this.selectedColumn === undefined) {
+      return undefined;
+    }
+    const values = this.numericValues.length ? this.numericValues : this.values;
```
LucasLefevre, line 154: GSheet keeps both strings and numbers and us only numbers.
Is it intentional ?

Image

Image

src/components/side_panel/column_stats/column_stats_panel.ts
```diff
@@ -0,0 +1,311 @@
...
+
+    return {
+      type: "bar",
+      data: {
+        labels: count.labels.map(this.clipTextWithEllipsis.bind(this)),
+        datasets: [
+          {
+            type: "bar",
+            data: count.data,
+            backgroundColor: "#5B9BD5",
```
LucasLefevre: not the same blue as the default chart blue. Is it intentional ?

Image

tests/column_statistics/column_statistics_store.test.ts
```diff
@@ -0,0 +1,181 @@
...
+
+    deleteRows(model, [1]);
+
+    expect(store.statisticFnResults["Unique values"]?.value?.()).toBe(2);
+  });
+
+  test("Store use correct number format", async () => {
+    setGrid(model, { A1: "10.5", A2: "20.3", A3: "30.7" });
+    setCellFormat(model, "A1", "[$$]#,##0.00");
+    model.getters.getCellFormat = () => "[$$]#,##0.00";
```
LucasLefevre:
```suggestion
```

tests/column_statistics/column_statistics_store.test.ts
```diff
@@ -0,0 +1,181 @@
...
+  store: Store<ColumnStatisticsStore>
+): Record<string, number | string | undefined> {
+  const results: Record<string, number | string | undefined> = {};
+  for (const [fnName, result] of Object.entries(store.statisticFnResults)) {
+    results[fnName] = result?.value?.();
+  }
+  return results;
+}
+
+describe("column statistics sidePanel component", () => {
```
LucasLefevre:
```suggestion
describe("column statistics sidePanel store", () => {
```

packages/o-spreadsheet-engine/src/functions/helper_statistical.ts
```diff
@@ -47,6 +47,28 @@ export function average(values: Arg[], locale: Locale) {
...
+    values,
+    (a) => {
+      nums.push(a.value);
+    },
+    locale
+  );
+  nums.sort((a, b) => a - b);
+  const len = nums.length;
+  if (len === 0) {
+    return 0;
```
LucasLefevre: this is technically wrong. The median isn't 0 if there is no data at all
```suggestion
  if (len === 0) {
    return undefined;
```

src/components/side_panel/column_stats/column_stats_store.ts
anhe-odoo: Yes, it was discussed with FRGI

src/components/side_panel/column_stats/column_stats_store.ts
anhe-odoo: No, they are numbers (or supposed to be), they comes from this line above : 
`const values = this.numbericValues.map((v) => v.value)`

PR #7795, opened 2026-01-15
[FIX] Cell: do not rely on key presence in commands

rrahir:
In our handler of `UPDATE_CELL`, we rely on the presence of a given key in the command to decide how to process it. However, undefined keys tend to be purged from a payload when serialized, which mean different users do not receive the same command payload and their state ends up diverging.

E.g.:
- Alice write something in A1
- Alice copies an empty cell
- Alice pastes it to A1

For Alice, A1 is empty
for others, A1 content did not change

The same applies to the formats but can only be replicated in later versions (18.0 +):

- Alice writes `2` in A1
- Alice adds a `%` format to A2
- Alice copies A1 and pastes it *as value only* to A2

Alice sees no `%` in A2 while other users see it.

Task: 5499921

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/clipboard/clipboard_cells_state.ts
+++ b/src/helpers/clipboard/clipboard_cells_state.ts
@@ -434,7 +434,7 @@ export class ClipboardCellsState extends ClipboardCellsAbstractState {
     if (content !== "" || origin.cell?.format || origin.cell?.style) {
       this.dispatch("UPDATE_CELL", {
         ...target,
-        content,
+        content: content ?? "",
         style: origin.cell?.style || null,
         format: origin.cell?.format,
       });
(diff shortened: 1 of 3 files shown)


src/helpers/clipboard/clipboard_cells_state.ts
```diff
@@ -434,7 +434,7 @@ export class ClipboardCellsState extends ClipboardCellsAbstractState {
     if (content !== "" || origin.cell?.format || origin.cell?.style) {
       this.dispatch("UPDATE_CELL", {
         ...target,
-        content,
+        content: content ?? null,
```
LucasLefevre: `null` or an empty string ?

Empty string does the job I think and doesn't introduce another special case.

```suggestion
        content: content ?? "",
```

rrahir: updated  with the suggestion l:)

LucasLefevre:
robodoo rebase-ff

PR #7796, opened 2026-01-16
[FIX] XLSXImport: Fix crash on incomplete xlsx file with external ref…

rrahir:
…erence

In Excel, there is a mecanism for spreadsheets to cross-reference themselves, with cells of the form `[otherSpreadsheetFileName]Sheet1!A1`. Luckily, the saved files contain a cache of the values from other spreadsheets and we use this cache during our import process to replace the cross-sheet references with their cached value.

Google sheet, and possibly other spreadsheets engines, do not support xlsx files with cross references to other spreadsheets. More specifically, they let them *as is* even though the reference will not work, and they throw the cache away. This means that an xlsx file generated from Google Sheet will not contain the cache either.

The code was designed with the assumption that the cache would always be present, which led to crashes when trying to upload an xlsx file (with cross-references) that was obtain through Google Sheets.

Task: 5499753

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/xlsx/conversion/formula_conversion.ts
+++ b/src/xlsx/conversion/formula_conversion.ts
@@ -65,10 +65,10 @@ function convertFormula(formula: string, data: XLSXImportData): string {
   formula = formula.replace(externalReferenceRegex, (match, externalRefId, sheetName, cellRef) => {
     externalRefId = Number(externalRefId) - 1;
     cellRef = cellRef.replace(/\$/g, "");
-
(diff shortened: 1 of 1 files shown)


LucasLefevre (commented):
robodoo r+

LucasLefevre:
robodoo r+

PR #7819, opened 2026-01-20
[IMP] icons: add and uniformize some icons

anhe-odoo:
## Task Description

We add here new icons for the formulas categories, and redraw pivots and carrousel icons to make them more similar to the other icons

## Related Task
- Task: [[ICONS] Improve/add some icons](https://www.odoo.com/odoo/2328/tasks/5797703)


packages/o-spreadsheet-engine/src/functions/module_text.ts
```diff
@@ -143,6 +143,34 @@ export const FIND = {
   isExported: true,
 } satisfies AddFunctionDescription;
 
+// -----------------------------------------------------------------------------
+// FORMAT.LARGE.NUMBER
+// -----------------------------------------------------------------------------
+export const FORMAT_LARGE_NUMBER = {
```
pro-odoo, line 149: A good practice is to make a separated commit for this kind of changes. It has nothing to do with icons :)

src/actions/insert_actions.ts
```diff
@@ -277,6 +277,7 @@ export const categoriesFunctionListMenuBuilder: ActionBuilder = () => {
     return {
       name: category,
       children: createFormulaFunctions(functionsInCategory),
+      sequence: 60 + i,
```
pro-odoo, line 280: Same for this :D

PR #7821, opened 2026-01-20
[Measure] Display Show as on measure tile

matho-odoo:
## Description:

When a Show as is set on a measure, there is no indication that it is set. The idea would be to display it on the tile so that it's clearer.

Task: [5502479](https://www.odoo.com/odoo/2328/tasks/5502479)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/packages/o-spreadsheet-engine/src/components/translations_terms.ts
+++ b/packages/o-spreadsheet-engine/src/components/translations_terms.ts
@@ -3,6 +3,7 @@ import { _t } from "../translation";
 import { ChartColorScale } from "../types/chart";
 import { CommandResult } from "../types/commands";
 import { Locale } from "../types/locale";
+import { NEXT_VALUE, PREVIOUS_VALUE } from "../types/pivot";
 
 export const CfTerms = {
   Errors: {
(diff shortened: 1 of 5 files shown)


tests/pivots/pivot_side_panel.test.ts
```diff
@@ -229,4 +229,36 @@ describe("Pivot side panel", () => {
...
+      A1: "Customer",
+      B1: "Price",
+      C1: "Year",
+      A2: "Alice",
+      B2: "10",
+      C2: "2020",
+      A3: "Alice",
+      B3: "20",
+      C3: "2021",
+    };
```
pro-odoo: Nitpick, add `//prettier-ignore` and format it as grid
```suggestion
    //prettier ignore
    const grid = {
      A1: "Customer", B1: "Price", C1: "Year",
      A2: "Alice",    B2: "10",    C2: "2020",
      A3: "Alice",    B3: "20",    C3: "2021",
    };
```

packages/o-spreadsheet-engine/src/components/translations_terms.ts
```diff
@@ -206,13 +206,39 @@ export const measureDisplayTerms = {
...
+    "%_of_parent_row_total": () => _t("Displayed as % of parent row total"),
     "%_of_parent_col_total": () => _t("Displayed as % of parent column total"),
     "%_of_parent_total": (field: string) => _t('Displayed as % of parent "%s" total', field),
-    difference_from: (field: string) => _t('Displayed as difference from "%s"', field),
-    "%_difference_from": (field: string) => _t('Displayed as % difference from "%s"', field),
+    difference_from: (field: string, value: string) => {
+      switch (value) {
+        case "(next)":
+          return _t('Displayed as difference from next "%s"', field);
+        case "(previous)":
```
LucasLefevre: There are constants for this. The goal of using this instead of raw strings is to avoid typos. If you do a typo with the variable name -> the compiler will tell you. If you make a typo in the string, it can go unnoticed
```suggestion
        case NEXT_VALUE:
          return _t('Displayed as difference from next "%s"', field);
        case PREVIOUS_VALUE:
```

packages/o-spreadsheet-engine/src/components/translations_terms.ts
```diff
@@ -206,13 +206,39 @@ export const measureDisplayTerms = {
...
     "%_of_col_total": () => _t("Displayed as % of column total"),
     "%_of_row_total": () => _t("Displayed as % of row total"),
-    "%_of": (field: string) => _t('Displayed as % of "%s"', field),
-    "%_of_parent_row_total": (field: string) =>
-      _t('Displayed as % of parent row total of "%s"', field),
+    "%_of": (field: string, value: string) => {
+      switch (value) {
+        case "(next)":
+          return _t('Displayed as % of next "%s"', field);
+        case "(previous)":
```
LucasLefevre:
```suggestion
        case NEXT_VALUE:
          return _t('Displayed as % of next "%s"', field);
        case PREVIOUS_VALUE:
```

packages/o-spreadsheet-engine/src/components/translations_terms.ts
```diff
@@ -206,13 +206,39 @@ export const measureDisplayTerms = {
...
+          return _t('Displayed as difference from previous "%s"', field);
+        default:
+          return _t('Displayed as difference from "%s" : %s', field, value);
+      }
+    },
+    "%_difference_from": (field: string, value: string) => {
+      switch (value) {
+        case "(next)":
+          return _t('Displayed as % difference from next "%s"', field);
+        case "(previous)":
```
LucasLefevre: same

PR #7824, opened 2026-01-20
[FIX] run engine with node.js

LucasLefevre:
With this commit, you can instantiate a new model with node.js

It currently crashes because `OffscreenCanvas` is not defined in node.js

The main goal is to be able to script performance measurements and run the script with node.js

I do not want to install a dependency for a canbas implementation for node.s because:
- I don't want dependencies
- I don't want to care about `npm`, `node_modules`, etc. I just want to run javascript with node.js
- we don't need it in practice

Task: [5804428](https://www.odoo.com/odoo/2328/tasks/5804428)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -50,6 +50,10 @@ export default (commandLineArgs) => {
           find: "@odoo/o-spreadsheet-engine/*",
           replacement: path.resolve(__dirname, "./packages/o-spreadsheet-engine/src/*"),
         },
+        {
+          find: /.*canvas_mock$/,
+          replacement: path.resolve(__dirname, "./tools/empty-module.js"),
+        },
       ],
     }),
     nodeResolve(),
@@ -89,6 +93,10 @@ export default (commandLineArgs) => {
                 __dirname,
                 "./packages/o-spreadsheet-engine/build/js/o-spreadsheet-engine/src/*"
               ),
+            },
+            {
+              find: /.*canvas_mock$/,
+              replacement: path.resolve(__dirname, "./tools/empty-module.js"),
             },
           ],
         }),
--- a/tools/empty-module.js
+++ b/tools/empty-module.js
@@ -1,0 +1,1 @@
+// Empty module used to exclude canvas_mock from browser builds


packages/o-spreadsheet-engine/src/helpers/text_helper.ts
```diff
@@ -20,21 +20,24 @@ export function computeTextLinesHeight(textLineHeight: number, numberOfLines: nu
   return numberOfLines * (textLineHeight + MIN_CELL_TEXT_MARGIN) - MIN_CELL_TEXT_MARGIN;
 }
 
-export function getCanvas(width: number = 100, height: number = 100): Canvas2DContext {
-  // Browser environment
+export function getCanvas(width: number = 100, height: number = 100): Canvas2DContext | undefined {
+  if (typeof OffscreenCanvas === "undefined") {
```
hokolomopo: why typeof ? why not just if(!offScreenCanvas) ?

LucasLefevre: because it crashes
```
if (!OffscreenCanvas) {
^

ReferenceError: OffscreenCanvas is not defined
```

hokolomopo: Indeed, I didn't know that https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof#interaction_with_undeclared_and_uninitialized_variables

packages/o-spreadsheet-engine/src/canvas_mock.ts
```diff
@@ -0,0 +1,43 @@
...
+        }
+        throw new Error(
+          `OffscreenCanvas mock: "${String(prop)}" is not implemented.\n` +
+            `Add it to MockOffscreenCanvas or MockOffscreenCanvasRenderingContext2D if needed.`
+        );
+      },
+    });
+  }
+
+  globalThis.OffscreenCanvas = MockOffscreenCanvas as unknown as typeof OffscreenCanvas;
```
fdamhaut: Is there a reason to not have `MockOffscreenCanvas as typeof OffscreenCanvas;` ?

PR #7830, opened 2026-01-21
[PERF] evaluation: fast predicate path for empty strings

LucasLefevre:
`evaluatePredicate` is slow when evaluating on empty strings. It uses a regexp to do that, which using a bazooka to kill a fly.

This situation is more common than you would think because the default value of `value` is `""` and, in odoo, the result of an empty ODOO.LIST function is also a `""`

### Synthetic data

```js
const cells = {};
for (let row = 1; row <= 10000; row++) {
    const cellId = `B${row}`;
    cells[cellId] = '=COUNTIF(A:A, "")';
}
const data = {
    sheets: [
        { cells, rowNumber: 10000 }
    ],
};
```
```
evaluate all cells
  before: Mean: 3576.86 ms, StdErr: 5.66 ms, n=20
  after:  Mean: 1329.21 ms, StdErr: 18.03 ms, n=20 (-63%)

Model created in
  before: Mean: 3750.99 ms, StdErr: 5.81 ms, n=20
  after:  Mean: 1504.15 ms, StdErr: 18.21 ms, n=20 (-60%)
```

### Real world spreadsheet on odoo.com
 
```
    Model created in
      before Mean=7544.97 ms, StdErr=17.06 ms, n=30
      after  Mean=7408.60 ms, StdErr=24.48 ms, n=30 (-2%)

    evaluate all cells
      before Mean=2575.05 ms, StdErr=5.72 ms, n=30
      after  Mean=2418.42 ms, StdErr=6.71 ms, n=30 (-6%)
```
## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/functions/helpers.ts
+++ b/src/functions/helpers.ts
@@ -727,7 +727,10 @@ function evaluatePredicate(
   if (operator === "<>" || operator === "=") {
     let result: boolean;
     if (typeof value === typeof operand) {
-      if (typeof value === "string" && typeof operand === "string") {
+      if (value === "" && operand === "") {
+        // fast path to avoid regex evaluation
+        result = true;
+      } else if (typeof value === "string" && typeof operand === "string") {
         result = wildcardToRegExp(operand).test(value);
       } else {
         result = value === operand;


pro-odoo:
robodoo r+

PR #7845, opened 2026-01-21
[IMP] spreadsheet: improve handling of context menu events

hokolomopo:
## Description

Our handling of context menu event is all over the place. Some components prevent the default behavior (eg. bottom bar, menu) while some others do not (eg. top bar, side panels).

With this commit, the spreadsheet component intercepts all the context
menu events. It prevent the default behavior, except for editable
elements so the user can copy/paste with the browser context menu.

Task: [5499550](https://www.odoo.com/odoo/2328/tasks/5499550)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/headers_overlay/headers_overlay.xml
+++ b/src/components/headers_overlay/headers_overlay.xml
@@ -32,7 +32,6 @@
           class="o-handle"
           t-on-pointerdown="onMouseDown"
           t-on-dblclick="onDblClick"
-          t-on-contextmenu.prevent=""
           t-attf-style="top:{{state.draggerLinePosition - 2}}px;">
           <div class="dragging-resizer" t-if="state.isResizing"/>
         </div>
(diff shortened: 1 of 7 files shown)


src/components/headers_overlay/headers_overlay.xml
```diff
@@ -75,7 +74,6 @@
           class="o-handle"
           t-on-pointerdown="onMouseDown"
           t-on-dblclick="onDblClick"
-          t-on-contextmenu.prevent=""
```
rrahir, line 78: if I right-click on the resizer, the context menu opens but then if I left-click on the header overlay again, I am in a weird state where there is a visual handle that follows my curso but nothing happens if I click. I think it's worth leaving this line?

hokolomopo: I added a check in the mousedown instead 

image

rrahir:
robodoo r+

PR #7846, opened 2026-01-21
[FIX] menu: can open empty menu popover

hokolomopo:
## Description

Since the split of `Menu` and `MenuPopover`, we can open an empty menu, which is not desired.

Two issues:
- the `t-if` condition was plainly wrong, checking for the existence of the array rather than it's length.
- the logic to know what menus are visible or not was in the `Menu` component, but is needed in the parent `MenuPopover` component to know if we should open the popover or not.

Task: [5863077](https://www.odoo.com/odoo/2328/tasks/5863077)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/actions/action.ts
+++ b/src/actions/action.ts
@@ -1,5 +1,7 @@
 import { Color, SpreadsheetChildEnv } from "../types";
 
+export type MenuItemOrSeparator = Action | "separator";
+
 /*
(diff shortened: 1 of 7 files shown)


pro-odoo (approved):
robodoo r+

PR #7850, opened 2026-01-22
[FIX] Evaluation: Provide cell position on single formula evaluation

rrahir:
How to reproduce:

- in a spreadsheet, go to the last row
- in a cell of the last row, write the formula `=MUNIT(ROW())` -> the formula is in #SPILL error because we did not extend the size of the spreadsheet

Some formulas, like ROW & COLUMN need to know the coordinates of the cell on which  they are applied to be computed and we did not provide those in the evaluation context when evaluating isolated formulas.

Task: 5798610

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/composer/composer/cell_composer_store.ts
+++ b/src/components/composer/composer/cell_composer_store.ts
@@ -244,7 +244,11 @@ export class CellComposerStore extends AbstractComposerStore {
       return;
     }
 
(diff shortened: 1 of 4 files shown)


rrahir:
robodoo up to saas-18.2

PR #7854, opened 2026-01-22
[FIX] Evaluation: Provide cell position for isolated formula evaluation

rrahir:
Some formulas, like ROW & COLUMN need to know the coordinates of the cell on which they are applied to be computed and we did not provide those in the evaluation context when evaluating isolated formulas.

How to reproduce:

- in a spreadsheet, go to the last row
- in a cell of the last row, write the formula `=MUNIT(ROW())` -> the formula is in #SPILL error because we did not extend the size of the spreadsheet

This also cause issues in standalone composers in which the hovered result displays an error.

How to reproduce:

- in a standalone composer, write =ROW()
- hover the ROW function -> it displays #ERROR

Task: 5798610

## Description:

Task: [5798610](https://www.odoo.com/odoo/2328/tasks/5798610)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/composer/composer/abstract_composer_store.ts
+++ b/src/components/composer/composer/abstract_composer_store.ts
@@ -324,9 +324,13 @@ export abstract class AbstractComposerStore extends SpreadsheetStore {
       hoveredFormula = `=${hoveredFormula}`;
     }
     const canonicalFormula = canonicalizeNumberContent(hoveredFormula, this.getters.getLocale());
-    const result = this.getters.evaluateFormulaResult(this.sheetId, canonicalFormula);
+    const result = this.evaluateCanonicalFormula(canonicalFormula);
     this.hoveredTokens = hoveredContextTokens;
     this.hoveredContentEvaluation = this.evaluationResultToDisplayString(result);
+  }
+
+  protected evaluateCanonicalFormula(canonicalFormula: string) {
+    return this.getters.evaluateFormulaResult(this.sheetId, canonicalFormula);
   }
 
   private getRelatedTokens(tokens: EnrichedToken[], tokenIndex: number): EnrichedToken[] {
--- a/src/components/composer/standalone_composer/standalone_composer_store.ts
+++ b/src/components/composer/standalone_composer/standalone_composer_store.ts
@@ -84,4 +84,12 @@ export class StandaloneComposerStore extends AbstractComposerStore {
     }
     return super.getTokenColor(token);
   }
+
+  hoverToken() {
+    /**
+     * Some functions can only be evaluated in the context of the grid, which is not feasible
+     * in the standalone composer.
+     */
+    return;
+  }
 }
--- a/src/plugins/ui_core_views/cell_evaluation/evaluator.ts
+++ b/src/plugins/ui_core_views/cell_evaluation/evaluator.ts
@@ -125,6 +125,22 @@ export class Evaluator {
     }
   }
 
+  private updateCompilationParametersForIsolatedFormula(originCellPosition?: CellPosition) {
+    this.compilationParams = buildCompilationParameters(
+      this.context,
+      this.getters,
+      this.computeAndSave.bind(this)
+    );
+    this.compilationParams.evalContext.__originCellPosition = originCellPosition;
+    this.compilationParams.evalContext.updateDependencies = undefined;
+    this.compilationParams.evalContext.addDependencies = undefined;
+    this.compilationParams.evalContext.lookupCaches = this.compilationParams.evalContext
+      .lookupCaches || {
+      forwardSearch: new Map(),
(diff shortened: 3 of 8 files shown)


src/components/composer/composer/abstract_composer_store.ts
```diff
@@ -324,11 +324,15 @@ export abstract class AbstractComposerStore extends SpreadsheetStore {
       hoveredFormula = `=${hoveredFormula}`;
     }
     const canonicalFormula = canonicalizeNumberContent(hoveredFormula, this.getters.getLocale());
-    const result = this.getters.evaluateFormulaResult(this.sheetId, canonicalFormula);
+    const result = this.faisDesTrucs(canonicalFormula);
```
hokolomopo: We're an english repo, it should be `doSomething()` ! Or a real name 🙃

src/helpers/misc.ts
```diff
@@ -705,3 +705,12 @@ export function getUniqueText(
 export function isFormula(content: string): boolean {
   return content.startsWith("=") || content.startsWith("+");
 }
+
+export function doesCellContainFunction(cell: Cell, functionName: string): boolean {
```
hokolomopo: unused

src/components/composer/standalone_composer/standalone_composer_store.ts
```diff
@@ -84,4 +85,15 @@ export class StandaloneComposerStore extends AbstractComposerStore {
     }
     return super.getTokenColor(token);
   }
+
+  faisDesTrucs(canonicalFormula: string) {
+    const tokens = tokenize(canonicalFormula);
+    if (tokens.some((t) => t.type === "SYMBOL" && t.value.toUpperCase() === "ROW")) {
+      return { value: _t("(row)") };
+    }
```
hokolomopo: Shouldn't it be #ERROR ? And only if the function isn't passed a parameter, `=ROW(A1)` works fine. 

But why can't you simply give the evaluation result ? There's no `__originCellPosition` => the function returns an error. The tooltip showing an error is correct. 

Unless ROW() should work for CFs/DVs ? In that case we need to pass an  __originCellPosition in the formula of these plugin (but then again, it make no sense for charts/pivots, so we'd have a case-by-case behaviour for the standalone composer)

rrahir: the thing is, it makes no sense in the sidepanel as the originPosition will only be set during the actual evaluation and it will change positionally. "Row" and "Col" seem nicer than #ERROR or #BAD_EXPR but U just realise now that this can also occur with the computed measure/dimension references in the composer of a pivot...
So yeah, maybe leaving in error is okayish

rrahir: does it even make sense to have the hover result in standalonc composers? I mean, they re bound to end in those weird situations since they are not tied to the grid evaluation

hokolomopo: On one hand I'd argue that it's even more useful to debug what's going on since you don't have access to the result in the grid on the other end it doesn't work great. In CF/DV we don't take the offset into account so it's confusing, and in pivots it's totally broken since we don't support the named field in the hover ... So yeah maybe deleting the feature in the standalone composer isn't that bad.

rrahir:
robodoo r+

rrahir:
robodoo r-

wtf rémi

src/components/composer/standalone_composer/standalone_composer_store.ts
```diff
@@ -84,4 +84,12 @@ export class StandaloneComposerStore extends AbstractComposerStore {
     }
     return super.getTokenColor(token);
   }
+
+  hoverToken() {
+    /**
```
rrahir, line 89: Sorry about this, the thing is I wanted to limit the mutators of StandaloneComposerStore but the typing of the mutators prevents it :'(

hokolomopo: Shoudn't the `hover` method be moved from the abstractComposerStore to the cellComposerStore instead/also ? Or is it too annoying ?

hokolomopo (commented):
:+1:

src/plugins/ui_core_views/cell_evaluation/evaluator.ts
```diff
@@ -125,6 +125,22 @@ export class Evaluator {
     }
   }
 
+  private updateCompilationParametersForIsolatedFormula(originCellPosition?: CellPosition) {
```
LucasLefevre, line 128: third different name :p 

I don't really care, but at one point it'll conflict and you'll have to choose

PR #7855, opened 2026-01-22
[IMP] Regex Formulas

matho-odoo:
## Description:

REGEXTEST: Checks if any part of supplied text matches a regex pattern.
REGEXREPLACE: Searches for a regex pattern within supplied text and replaces it with different text.

Task: [5411453](https://www.odoo.com/odoo/2328/tasks/5411453)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


packages/o-spreadsheet-engine/src/functions/module_text.ts
```diff
@@ -280,6 +280,54 @@ export const PROPER = {
...
+    }
+    if (_caseSensitivity !== 0 && _caseSensitivity !== 1) {
+      return new EvaluationError(_t("The case_sensitivity (%s) must be 0 or 1.", _caseSensitivity));
+    }
+
+    const flags = _caseSensitivity === 1 ? "gi" : "g";
+    const regex = new RegExp(_pattern, flags);
+    const matches = [..._text.matchAll(regex)];
+
+    return matches.length !== 0;
```
fdamhaut:
```suggestion
    return _text.search(regex)] !== -1;
```
This will avoid looping over the whole string and stop at the first match found

packages/o-spreadsheet-engine/src/functions/module_text.ts
```diff
@@ -352,6 +400,82 @@ export const REGEXEXTRACT = {
...
+    if (_occurence !== 0) {
+      const i = _occurence > 0 ? _occurence - 1 : matches.length + _occurence;
+      const length = matches[i][0].length;
+      const position = matches[i]["index"];
+      return _text.substring(0, position) + _replacement + _text.substring(position + length);
+    }
+    for (const match of matches) {
+      _text = _text.split(match[0]).join(_replacement);
+    }
+    return _text;
```
fdamhaut:
```suggestion
    return _text.replace(regex, _replacement)
```
The original is wrong if you have multiple match overlapping (You can test that with `=REGEXREPLACE("ABC", "[A-Z]*", "123")`)

(If you want to write a smaller code, you can also use replace if pattern = "")

tests/functions/module_text.test.ts
```diff
@@ -474,6 +528,84 @@ describe("REGEXEXTRACT function", () => {
...
+      expectedResult: string
+    ) => {
+      expect(
+        evaluateCell("A1", {
+          A1: `=REGEXREPLACE("${arg0}", "${arg1}", "${arg2}", ${arg3}, ${arg4})`,
+        })
+      ).toBe(expectedResult);
+    }
+  );
+});
```
LucasLefevre, line 648: add test with capturing groups

Image

packages/o-spreadsheet-engine/src/functions/module_text.ts
```diff
@@ -352,6 +399,76 @@ export const REGEXEXTRACT = {
...
+    if (matches.length === 0 || Math.abs(_occurence) > matches.length) {
+      return _text;
+    }
+    if (_occurence !== 0) {
+      const i = _occurence > 0 ? _occurence - 1 : matches.length + _occurence;
+      const length = matches[i][0].length;
+      const position = matches[i]["index"];
+      return _text.substring(0, position) + _replacement + _text.substring(position + length);
+    }
+    return _text.replace(regex, _replacement);
```
LucasLefevre: executing a regexp can be very (very) expensive, depending on the regexp.
Try not to execute it twice. Execute `matchAll` only if `_occurence !== 0`

EDIT: maybe we should even also cache regexes. Compiling a regexp can also be expensive

packages/o-spreadsheet-engine/src/functions/module_text.ts
```diff
@@ -280,6 +280,53 @@ export const PROPER = {
...
+      return false;
+    }
+    if (_caseSensitivity !== 0 && _caseSensitivity !== 1) {
+      return new EvaluationError(_t("The case_sensitivity (%s) must be 0 or 1.", _caseSensitivity));
+    }
+
+    const flags = _caseSensitivity === 1 ? "gi" : "g";
+    const regex = new RegExp(_pattern, flags);
+
+    return _text.search(regex) !== -1;
```
LucasLefevre: more idiomatic and probably slightly faster since it doesn't need to track the index
```suggestion
    return regex.test(_text);
```

tests/functions/module_text.test.ts
```diff
@@ -391,6 +391,60 @@ describe("PROPER formula", () => {
...
+    ["Hello there", "i(.*)", 0],
+    ["Hello there", "[0-9]+", 0],
+    ["Hello there", "hello", 0],
+  ])(
+    "various function with false results =REGEXTEST(%s, %s, %s)",
+    (arg0: string, arg1: string, arg2: number) => {
+      expect(evaluateCell("A1", { A1: `=REGEXTEST("${arg0}", "${arg1}", ${arg2})` })).toBe(false);
+    }
+  );
+});
```
fdamhaut: Could you add one or two test with case sensivity turned off

tests/functions/module_text.test.ts
```diff
@@ -391,6 +391,60 @@ describe("PROPER formula", () => {
...
+    ["Hello there", "o(.*)", 0],
+    ["Hello there", "\\s.*", 0],
+    ["Hello 56 there 89", "[0-9]+", 0],
+    ["Hello there my guy", "\\s[a-z]+", 0],
+    ["word boundary", "\\b[a-z]+", 0],
+    ["Hello There My Guy", "(\\s[a-z]+)+", 1],
+    ["VAT:21% PRICE:200€", ".*:([0-9]+).*:([0-9]+)", 0],
+  ])(
+    "various function with true results =REGEXTEST(%s, %s, %s)",
+    (arg0: string, arg1: string, arg2: number) => {
```
fdamhaut: Could you rename the args so they correspond to their usage for clarity

tests/functions/module_text.test.ts
```diff
@@ -474,6 +528,84 @@ describe("REGEXEXTRACT function", () => {
...
+    ["ABC", "[A-Z]*", "123", 0, 0, "123123"],
+  ])(
+    "various function results =REGEXREPLACE(%s, %s, %s, %s, %s)",
+    (
+      arg0: string,
+      arg1: string,
+      arg2: string,
+      arg3: number,
+      arg4: number,
+      expectedResult: string
```
fdamhaut: same as above

tests/functions/module_text.test.ts
```diff
@@ -474,6 +528,84 @@ describe("REGEXEXTRACT function", () => {
...
+    expect(evaluateCell("A1", { A1: '=REGEXREPLACE("HellO", "[A-Z]+", "y", 0, 1)' })).toBe("y");
+  });
+
+  test.each([
+    ["Hello there", "there", "World", 0, 0, "Hello World"],
+    ["Hello there", "o.*", "u", 0, 0, "Hellu"],
+    ["Hello there", "\\s.*", "World", 0, 0, "HelloWorld"],
+    ["Hello 56 there 89", "[0-9]+", "you", 0, 0, "Hello you there you"],
+    ["Hello 56 there 89", "[0-9]+", "you", 1, 0, "Hello you there 89"],
+    ["Hello 56 there 89", "[0-9]+", "you", 2, 0, "Hello 56 there you"],
```
fdamhaut: One test where occurance > number of occurance in text

tests/functions/module_text.test.ts
```diff
@@ -391,6 +391,60 @@ describe("PROPER formula", () => {
...
+  test("case_sensitivity is either 0 or 1", () => {
+    expect(evaluateCell("A1", { A1: '=REGEXTEST("Hello", "lo", -1)' })).toBe("#ERROR");
+    expect(evaluateCell("A1", { A1: '=REGEXTEST("Hello", "lo", 0)' })).not.toBe("#ERROR");
+    expect(evaluateCell("A1", { A1: '=REGEXTEST("Hello", "lo", 1)' })).not.toBe("#ERROR");
+    expect(evaluateCell("A1", { A1: '=REGEXTEST("Hello", "lo", 2)' })).toBe("#ERROR");
+  });
+
+  test.each([
+    ["Hello there", "there", 0],
+    ["Hello there", "hello", 1],
```
LucasLefevre: In some tests, you use `test.each` and in others you use many `evaluateCell`. Try to be consistent.
Many `evaluateCell` is more readable IMO

tests/functions/module_text.test.ts
```diff
@@ -391,6 +391,60 @@ describe("PROPER formula", () => {
...
+    ["Hello there", "o.*", 0],
+    ["Hello there", "o(.*)", 0],
+    ["Hello there", "\\s.*", 0],
```
LucasLefevre: you can use `String.raw` to avoid the need to escape. The regexp would be more readable

```
String.raw`\s.*`
String.raw`=REGEXTEST("Hello there", "\s.*", 0)`
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/raw

tests/functions/module_text.test.ts
```diff
@@ -391,6 +391,60 @@ describe("PROPER formula", () => {
   });
 });
 
+describe("REGEXTEST function", () => {
```
LucasLefevre, line 394: add a test with an invalid regexp

tests/functions/module_text.test.ts
```diff
@@ -474,6 +528,84 @@ describe("REGEXEXTRACT function", () => {
   );
 });
 
+describe("REGEXREPLACE function", () => {
```
LucasLefevre, line 539: test with invalid regexp

LucasLefevre, line 539: test with escaped characters in the regexp

`=REGEXTEST(A1,A2)` where `A1` contains `Hello \there\` and `A2` contains `\\\w+\\`

packages/o-spreadsheet-engine/src/functions/module_text.ts
```diff
@@ -352,6 +404,82 @@ export const REGEXEXTRACT = {
...
+    }
+
+    if (_occurence !== 0) {
+      const matches = [..._text.matchAll(regex)];
+      if (matches.length === 0 || Math.abs(_occurence) > matches.length) {
+        return _text;
+      }
+      const i = _occurence > 0 ? _occurence - 1 : matches.length + _occurence;
+      const length = matches[i][0].length;
+      const position = matches[i]["index"];
```
hokolomopo:
```suggestion
      const position = matches[i].index
```
avoid bracket notation, it ignores typing

LucasLefevre: up

PR #7861, opened 2026-01-26
[IMP] annotation tool for charts

matho-odoo:
Possibility to add a note or a link in the chart side panel

Task: [5365254](https://www.odoo.com/odoo/2328/tasks/5365254)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/components/figures/chart/chart_dashboard_menu/chart_dashboard_menu.xml
```diff
@@ -12,11 +12,25 @@
             <t t-if="item.preview" t-call="{{item.preview}}"/>
           </div>
         </t>
+        <div
+          class="o-figure-info-icon d-flex flex-row align-items-center align-middle text-muted me-2"
```
anhe-odoo: We should add a `cursor: pointer` CSS rule on this div :)

src/components/side_panel/chart/building_blocks/annotation/annotation.xml
```diff
@@ -0,0 +1,28 @@
...
+            t-on-change="(ev) => this.updateAnnotationText(ev)"
+          />
+        </Section>
+        <Section class="'pt-0 px-4'" title.translate="See more link">
+          <input
+            class="o-input"
+            data-test-id="annotation-link"
+            type="text"
+            placeholder="Add a redirection link"
+            t-att-value="this.props.definition.annotationLink"
```
anhe-odoo: It will feel better with an `onKeyDown` method to blur when we press "enter" and to reset when the user press "Escape"

src/components/side_panel/chart/building_blocks/annotation/annotation.xml
```diff
@@ -0,0 +1,28 @@
+<templates>
+  <t t-name="o-spreadsheet.ChartAnnotation">
+    <SidePanelCollapsible isInitiallyCollapsed="false" title.translate="Annotation">
+      <t t-set-slot="content">
+        <Section class="'pt-0 px-4'" title.translate="Text">
+          <textarea
```
anhe-odoo: TBD with François, but the feeling of the textArea is not really good. We start with an empty line, and then if we have more than 2 lines we end with a very small text area and a scrollbar not really usable. Mayce you could see if there is a way to make the textArea automatically resize

packages/o-spreadsheet-engine/src/types/chart/funnel_chart.ts
anhe-odoo: This should extends `CommonChartDefinition` to make it shorter and easier to maintains :)

packages/o-spreadsheet-engine/src/types/chart/gauge_chart.ts
```diff
@@ -9,6 +9,8 @@ export interface GaugeChartDefinition {
   readonly sectionRule: SectionRule;
   readonly background?: Color;
   readonly humanize?: boolean;
+  readonly annotationText?: string;
+  readonly annotationLink?: string;
```
anhe-odoo: Maybe at this point we can imagine some `MinimalChartDefinition` containing the fields common to every chart (title, background, humanize, annotationText and annotationLink), and have every other base definition (Gauche, Scorecard, Sunburst, Treemap and Common) extends it ? What do you think ?

src/components/figures/chart/chart_dashboard_menu/chart_dashboard_menu.ts
```diff
@@ -25,13 +26,15 @@ interface MenuItem {
...
   static template = "o-spreadsheet-ChartDashboardMenu";
-  static components = { MenuPopover };
+  static components = { MenuPopover, InfoPopover };
   static props = { chartId: String, hasFullScreenButton: { type: Boolean, optional: true } };
   static defaultProps = { hasFullScreenButton: true };
 
   private fullScreenFigureStore!: Store<FullScreenFigureStore>;
 
   private menuState: MenuState = useState({ isOpen: false, anchorRect: null, menuItems: [] });
+  private infoState: InfoState = useState({ isOpen: false, anchorRect: null });
```
anhe-odoo: Leave this for the next reviewer to be sure, but I'll use only one state as we are not supposed to have the menu and the info opened at the same time. Si we can imagine something like

```js
interface ChartMenuState {
    isOpen: boolean;
    openedPopover?: "menu" | "info";
    anchorRect: null | Rect;
    menuItems: Action[];
};
```

and then, instead of `this.menuState.isOpen` and `this.infoState.isOpen`, we can use `this.state.openedPopover === "menu"` and `this.state.openedPopover === "info"`. But as I said, I'm not sure, so let's wait for Lucas, Adrien or someone else to go through this to have another opinon.

src/components/figures/chart/chart_dashboard_menu/chart_dashboard_menu.ts
```diff
@@ -53,6 +56,21 @@ export class ChartDashboardMenu extends Component<Props, SpreadsheetChildEnv> {
     this.menuState.menuItems = getChartMenuActions(figureId, this.env);
   }
 
+  showInfo(ev: MouseEvent) {
+    this.infoState.isOpen = true;
+    this.infoState.anchorRect = getBoundingRectAsPOJO(ev.currentTarget as HTMLElement);
+  }
+
+  getAnnotationText() {
+    const chart = this.env.model.getters.getChart(this.props.chartId);
```
anhe-odoo: It's better to get the information from the definition, as done a few line below : 

```js
const definition = this.env.model.getters.getChartDefinition(this.props.chartId);
return definition.annotationText;
```

You can then have something like this : 

```js

get chartDefinition() {
      return this.env.model.getters.getChartDefinition(this.props.chartId);
}
```

and then use `this.chartDefinition.annotationText`

src/components/figures/figure/figure.ts
```diff
@@ -280,4 +283,19 @@ export class FigureComponent extends Component<Props, SpreadsheetChildEnv> {
       !this.env.model.getters.isCurrentSheetLocked()
     );
   }
+
+  showInfo() {
```
anhe-odoo: These should not be handled by the `figure` component but by the `figure_chart` component, as there is not annotation for the image. Moreover, here, you have to use `getChartFromFigureId` while you are not sure that `this.props.figureUI.id` is an id of a chart.

src/components/figures/chart/chart_dashboard_menu/chart_dashboard_menu.xml
```diff
@@ -12,11 +12,25 @@
...
+          t-if="getAnnotationText() or getAnnotationLink()"
+          t-on-click="showInfo"
```
anhe-odoo: Clicking on the icon should toggle the Info bubble, not always showing it :)

src/components/figures/figure/figure.xml
```diff
@@ -25,6 +25,21 @@
           openContextMenu.bind="openContextMenu"
         />
         <div class="o-figure-menu position-absolute m-2" t-if="!env.isDashboard()">
+          <div
+            class="o-figure-info-icon d-flex flex-row align-items-center align-middle text-muted me-2"
```
anhe-odoo: What is the aim of using the `text-muted` class here ?

src/components/info_popover/info_popover.css
```diff
@@ -0,0 +1,21 @@
+.o-spreadsheet {
+  a.o-link-info {
```
anhe-odoo: `height`, `width` and `justify-content` properties seem useless here. Moreover, as you only display "See more" or some translated string, I don't think you need the `white-space`, `overflow` and `text-overflow` properties here :)

src/components/info_popover/info_popover.css
```diff
@@ -0,0 +1,21 @@
...
+    line-height: 13px;
+    color: var(--os-link-color);
+    text-decoration: none;
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
+  }
+
+  a.o-link-info:hover {
+    text-decoration: none;
```
anhe-odoo: `text-decoration: none` is useless here as it is inherited from the `a.o-link-info` classe

src/components/side_panel/chart/building_blocks/annotation/annotation.ts
```diff
@@ -0,0 +1,64 @@
...
+  ChartSidePanelProps<ChartDefinition>,
+  SpreadsheetChildEnv
+> {
+  static template = "o-spreadsheet.ChartAnnotation";
+  static components = { SidePanelCollapsible, Section };
+  static props = ChartSidePanelPropsObject;
+
+  setup() {
+    onMounted(() => this.autoResize());
+    onPatched(() => this.autoResize());
```
hokolomopo: The scroll is bugged, I'm pretty sure it's because of this.autoresize

https://drive.google.com/file/d/1sbSh4S6hkD7PwbPDYiQOAE2KemV8zDAX/view?usp=drive_link

Not sure how to make a resizeable textare w/o javascript tho. I'd use a `contenteditable` div like the composer, so you can handle the reisze with just a bit of css `max-height: 200px;height: fit-content;
`

src/components/info_popover/info_popover.xml
```diff
@@ -0,0 +1,22 @@
+<templates>
+  <t t-name="o-spreadsheet-Info-Popover">
+    <Popover t-props="popoverProps">
+      <div class="bg-white overflow-auto p-2" t-att-style="infoStyle" t-ref="info">
+        <div class="m-1" data-test-id="annotation-text" t-if="props.annotationText">
+          <t t-foreach="props.annotationText.split('\n')" t-as="line" t-key="line">
```
hokolomopo:
```suggestion
          <t t-foreach="props.annotationText.split('\n')" t-as="line" t-key="line_index">
```
if 2 lines have the same text it crashes (t-key cannot be duplicate)

packages/o-spreadsheet-engine/src/types/chart/common_chart.ts
```diff
@@ -4,15 +4,20 @@ import { AxesDesign, CustomizedDataSet, TitleDesign } from "./chart";
...
   readonly labelRange?: string;
-  readonly title: TitleDesign;
-  readonly background?: Color;
   readonly legendPosition: LegendPosition;
   readonly aggregated?: boolean;
   readonly axesDesign?: AxesDesign;
   readonly showValues?: boolean;
+}
+
+export interface MinimalChartDefinition {
```
hokolomopo: Not sure I'd add MinimalChartDefinition, it doesn't really improve the clarity of the typing. The name weren't very clear before (**Common**ChartDefinition isn't common to all the charts ??), and aren't much better now.

But probably all the charts except for gauge/scorecards can extend CommonChartDefinition? Then you could just add annotationText/Link to the CommonChartDefinitiion and in scorecard/gauge

src/components/figures/chart/chart_dashboard_menu/chart_dashboard_menu.xml
```diff
@@ -12,16 +12,30 @@
...
+      />
       <MenuPopover
-        t-if="menuState.isOpen"
-        anchorRect="menuState.anchorRect"
-        menuItems="menuState.menuItems"
-        onClose="() => this.menuState.isOpen=false"
+        t-if="state.isOpen and state.openedPopover === 'menu'"
+        anchorRect="state.anchorRect"
+        menuItems="state.menuItems"
+        onClose="() => this.state.isOpen=false"
```
hokolomopo: I'd change the onClose to a real method that would properly clean up the state. ATM if you open a menu then an info popover, you get a state with popover === "info", but state.menuItems that has some things that haven't beed cleaned

src/components/figures/figure/figure.css
```diff
@@ -13,6 +13,14 @@
     }
   }
 
+  .o-figure-info-icon {
+    margin-top: 2px;
+    width: 15px;
+    height: 15px;
+    cursor: pointer;
+    margin-right: 25px;
```
hokolomopo: add a TODO comment to say that we should imrpove that in the future

src/components/info_popover/info_popover.css
```diff
@@ -0,0 +1,14 @@
...
+    font-weight: bold;
+    line-height: 13px;
+    color: var(--os-link-color);
+    text-decoration: none;
+  }
+
+  a.o-link-info:hover {
+    color: var(--os-link-hover-color);
+    cursor: pointer;
+  }
```
hokolomopo: Use a div with the `o-button-link` class instead of an `a`, it has most of the style you want already

src/components/info_popover/info_popover.ts
```diff
@@ -0,0 +1,80 @@
...
+import { PopoverPropsPosition } from "../../types/cell_popovers";
+import { cssPropertiesToCss } from "../helpers";
+import { isChildEvent } from "../helpers/dom_helpers";
+import { Popover, PopoverProps } from "../popover";
+
+interface Props {
+  anchorRect: Rect;
+  popoverPositioning: PopoverPropsPosition;
+  maxHeight?: Pixel;
+  width?: number;
```
hokolomopo:
```suggestion
  width: number;
```
with is always defined with your defaultProps

src/components/info_popover/info_popover.ts
```diff
@@ -0,0 +1,80 @@
...
+import { SpreadsheetChildEnv } from "@odoo/o-spreadsheet-engine/types/spreadsheet_env";
+import { Component, useExternalListener, useRef } from "@odoo/owl";
+import { PopoverPropsPosition } from "../../types/cell_popovers";
+import { cssPropertiesToCss } from "../helpers";
+import { isChildEvent } from "../helpers/dom_helpers";
+import { Popover, PopoverProps } from "../popover";
+
+interface Props {
+  anchorRect: Rect;
+  popoverPositioning: PopoverPropsPosition;
```
hokolomopo: this props is never used

src/components/info_popover/info_popover.ts
hokolomopo: and why  is the width different in dashboard & in spreadsheet mode ? Is there a reason ?

tests/test_helpers/constants.ts
```diff
@@ -51,6 +54,16 @@ export const TEST_CHART_DATA = {
...
+    annotationLink: "https://www.odoo.com",
+  },
+  funnel: {
+    type: "funnel" as const,
+    keyValue: "B1:B4",
+    baseline: "A2:A4",
+    title: { text: "hello" },
+    baselineDescr: { text: "description" },
+    baselineMode: "difference" as const,
+    annotationLink: "javascript:alert('hello world')",
```
hokolomopo: ... does this really work ? It would be a problem if it does, would allow to execute an arbitrary code on a victim

hokolomopo (commented):
👋

tests/test_helpers/constants.ts
matho-odoo: Https:// is automatically added to avoid security problem, there is a test to show it

src/components/figures/chart/chart_dashboard_menu./chart_dashboard_menu.ts
```diff
@@ -23,15 +25,28 @@ interface MenuItem {
   preview?: string;
 }
 
+interface ChartMenuState {
+  isOpen: boolean;
+  openedPopover?: "menu" | "info";
```
anhe-odoo, line 28: Now that we have the `openedPopover`, I"m not sure that `isOpen` is still needed anymore (we can check if `openedPopover !== undefined` )

src/components/figures/chart/chart_dashboard_menu/chart_dashboard_menu.xml
```diff
@@ -12,16 +12,30 @@
...
+          t-ref="infoButton">
+          <t t-call="o-spreadsheet-Icon.CIRCLE_INFO"/>
+        </div>
         <button
           class="o-chart-dashboard-item btn text-muted lh-1 p-1 fa fa-ellipsis-v"
           t-on-click="openContextMenu"
         />
       </div>
+      <InfoPopover
+        t-if="state.isOpen and state.openedPopover === 'info'"
```
anhe-odoo: We don't need to check if `state.isOpen` if we already know that `state.openedPopover === "info"`

src/components/figures/chart/chart_dashboard_menu/chart_dashboard_menu.xml
```diff
@@ -12,16 +12,30 @@
...
+        onClose.bind="onClose"
+        annotationText="getAnnotationText()"
+        annotationLink="getAnnotationLink()"
+      />
       <MenuPopover
-        t-if="menuState.isOpen"
-        anchorRect="menuState.anchorRect"
-        menuItems="menuState.menuItems"
-        onClose="() => this.menuState.isOpen=false"
+        t-if="state.isOpen and state.openedPopover === 'menu'"
```
anhe-odoo: Same here

src/components/info_popover/info_popover.ts
```diff
@@ -0,0 +1,76 @@
...
+interface Props {
+  anchorRect: Rect;
+  maxHeight?: Pixel;
+  width: number;
+  onClose: () => void;
+  annotationText?: string;
+  annotationLink?: string;
+}
+
+export interface InfoState {
```
anhe-odoo: Move this to figure_chart.ts as it's defined here but only imported and used there

src/components/info_popover/info_popover.ts
```diff
@@ -0,0 +1,76 @@
...
+  get popoverProps(): PopoverProps {
+    return {
+      anchorRect: this.props.anchorRect,
+      maxHeight: this.props.maxHeight,
+      positioning: "bottom-left",
+      verticalOffset: 0,
+    };
+  }
+
+  private onExternalClick(ev: MenuMouseEvent) {
```
anhe-odoo, line 38: I'll go for having the external listener in the parent (figure_chart.ts for example) instead of here directly. This would avoid having to set the `ev.closedMenuId` and passing the `onClose` property. But I'm not sure about it. What do you think @hokolomopo ?

hokolomopo: We cannot do that in the figure component because we need the HTML element of the popover for the external click

src/components/figures/figure_chart/figure_chart.ts
```diff
@@ -13,6 +16,11 @@ interface Props {
   openContextMenu?: (anchorRect: Rect, onClose?: () => void) => void;
 }
 
+export interface InfoState {
```
anhe-odoo: This should not be exported as it's not used anywhere else outside this file :)

src/components/side_panel/chart/building_blocks/annotation/annotation.xml
```diff
@@ -0,0 +1,29 @@
...
+            class="o-input o-input-infotext"
+            t-ref="annotationTextarea"
+            data-placeholder="Add a description text"
+            data-test-id="annotation-text"
+            t-att-contenteditable="'plaintext-only'"
+            t-on-input="(ev) => this.updateAnnotationText(ev)"
+          />
+        </Section>
+        <Section class="'pt-0 px-4'" title.translate="See more link">
+          <input
```
anhe-odoo: Here you can use the `TextInput` component instead of an `<input type="text">`. We try to use theme as much as possible in the template now, and they handle the Enter/Escape part so that you can remove your onKeyDown handling in the annotation.ts file

src/components/side_panel/chart/building_blocks/annotation/annotation.xml
```diff
@@ -0,0 +1,30 @@
+<templates>
+  <t t-name="o-spreadsheet.ChartAnnotation">
+    <SidePanelCollapsible isInitiallyCollapsed="true" title.translate="Annotation">
+      <t t-set-slot="content">
+        <Section class="'pt-0 px-4'" title.translate="Text">
+          <div
+            class="o-input o-input-infotext"
+            t-ref="annotationTextarea"
+            data-placeholder="Add a description text"
```
anhe-odoo: shouldn't it be `placeholder.translate` instead of `data-placeholder` ? And it feels strange to have these in a div instead of an input

anhe-odoo: @hokolomopo : looks like it was your idea/suggestion. Can you explain why ? Thanks :)

hokolomopo: Erm ... was it ? `data-placeholder` seems wrong to me ...

src/components/side_panel/chart/building_blocks/annotation/annotation.ts
```diff
@@ -0,0 +1,53 @@
...
+        this.editorRef.el.innerText = this.props.definition.annotationText || "";
+      }
+    });
+  }
+
+  checkLength(ev: Event) {
+    const currentLength = (ev.target as HTMLInputElement).innerText.length;
+    if (currentLength >= 2000) {
+      ev.preventDefault();
+    }
```
VincentSchippefilt, line 36: if I paste a large text, it will be cut at 2000 chars (probably) but then I can't modify it anymore

src/components/side_panel/chart/building_blocks/annotation/annotation.xml
```diff
@@ -0,0 +1,30 @@
...
+          />
+        </Section>
+        <Section class="'pt-0 px-4'" title.translate="See more link">
+          <div data-test-id="annotation-link">
+            <TextInput
+              class="'o-input'"
+              value="props.definition.annotationLink ?? ''"
+              onChange="(label) => this.updateAnnotationLink(label)"
+              placeholder.translate="Add a redirection link"
+              maxLength="2000"
```
VincentSchippefilt, line 23: I can paste text here, and there is no limit of length (I pasted a string of 90176 characters)

tests/__snapshots__/top_bar_component.test.ts.snap
```diff
@@ -79,6 +79,7 @@ exports[`TopBar component simple rendering 1`] = `
             >
               <input
                 class="w-100 os-input"
+                maxlength="this.props.maxLength"
```
VincentSchippefilt: lol no. this probably explains why the max-length doesn't work on the link input... in the template of TextInput, it should be `t-att-maxlengh=this.props.maxLength`so it is evaluated

VincentSchippefilt:
could you also add a few annotations in the demo data?

LucasLefevre (commented):
Testing the review skill for Claude:

# PR Review: [IMP] charts: annotation tool

Commit: `085fb04eb2977126384ed4f427a4e6304f3353d6`

## Commit hygiene
Single commit, clean message, tests included.

## Findings

### src/components/figures/figure_chart/figure_chart.ts:34-85 — warning
Dead code: `infoState`, `infoButtonRef`, `showInfo`, `getAnnotationText`, `getAnnotationLink`, and the imports of `InfoPopover`, `useRef`, `useState`, `MenuMouseEvent`, `getRefBoundingRect` are unused — the new template only renders `<ChartMenu>` and none of these are referenced. Remove them; also drop `InfoPopover` from `static components`.

### src/components/figures/chart/chart_menu/chart_menu.ts:96-99 — warning
`showInfo` doesn't toggle: when the popover is open, clicking the info icon fires `onExternalClick` (capture phase, closes it) and then the bubble-phase `t-on-click` runs `showInfo` which reopens it. The previous `figure_chart` version checked `ev.closedMenuId === "info-popover"` to short-circuit this; do the same here, or stop propagation in `onExternalClick` when the click target is the info icon.

### src/components/info_popover/info_popover.ts:21 — nit
`maxHeight` is declared in `static props` but never used as a prop — `popoverProps` hardcodes `maxHeight: 300`. Drop it from the props schema or honor `this.props.maxHeight`.

### src/components/figures/chart/chart_menu/chart_menu.css:1 / .xml:3 — nit
Class name `o-dashboard-chart-select` is misleading now that the menu also shows outside dashboard mode. Consider `o-chart-menu` for consistency with the component rename.

### src/components/side_panel/chart/building_blocks/annotation/annotation.ts:14 — nit
`onMounted` reads from `this.props.definition.annotationText` once; if the definition changes externally (undo/redo with the panel open) the editor won't refresh. Use `useEffect` on the value if you want it to track.

## Verdict
COMMENT — solid feature, but the dead `InfoPopover` plumbing in `figure_chart.ts` and the non-toggling info button are worth fixing before merge.

src/components/info_popover/info_popover.xml
```diff
@@ -0,0 +1,23 @@
+<templates>
+  <t t-name="o-spreadsheet-Info-Popover">
+    <Popover t-props="this.popoverProps">
+      <div class="bg-white overflow-auto p-2" t-ref="this.infoRef">
+        <div class="m-1" data-test-id="annotation-text" t-if="this.props.annotationText">
+          <t t-foreach="this.props.annotationText.split('\n')" t-as="line" t-key="line_index">
```
hokolomopo, line 6: I have empty lines sometimes, probably missing a text.trim()

Image

src/components/figures/figure/figure.ts
```diff
@@ -296,4 +296,12 @@ export class FigureComponent extends Component<SpreadsheetChildEnv> {
...
     );
   }
+
+  get shouldShowMenuButton(): boolean {
+    return (
+      !this.env.model.getters.isReadonly() &&
+      this.props.figureUI.tag !== "carousel" &&
+      this.props.figureUI.tag !== "chart"
+    );
+  }
```
hokolomopo: It's only for image figures then ? the menu should be removed from figure and just added to Image at this point

src/components/figures/chart/chart_menu/chart_menu.xml
```diff
@@ -0,0 +1,45 @@
...
+            class="'o-chart-dashboard-item w-auto py-0 px-1 text-muted text-nowrap rounded border-0'"
+            popoverClass="'w-auto'"
+            values="this.regionOptions"
+            selectedValue="this.selectedRegion"
+            onChange.bind="this.onRegionSelected"
+          />
+        </t>
+        <t t-foreach="this.getMenuItems()" t-as="item" t-key="item.id">
+          <div
+            class="o-chart-item border-0 lh-1 p-1 d-flex align-items-center justify-content-center rounded"
```
hokolomopo: probably remove `o-chart-dashboard-item` in favor of `o-chart-item` ?

src/components/info_popover/info_popover.xml
```diff
@@ -0,0 +1,23 @@
...
+            <br/>
+          </t>
+        </div>
+        <div class="m-1" t-if="this.props.annotationLink">
+          <a
+            class="o-button-link"
+            t-att-title="this.props.annotationLink"
+            t-att-href="this.getAnnotationLinkWithHttps()"
+            target="_blank">
+            <t t-out="this.seeMoreLabel"/>
```
hokolomopo: A reason you use a t-out instead of just inlining the text in the HTML ?

src/components/side_panel/chart/building_blocks/annotation/annotation.css
```diff
@@ -0,0 +1,19 @@
+.o-spreadsheet {
+  .o-input-infotext {
+    height: fit-content;
+    max-height: 200px;
+    min-height: 1.2em;
+    overflow-y: auto;
+    outline: none;
+    position: relative;
+    cursor: text;
+    border-bottom: 1px solid var(--os-border-color);
```
hokolomopo: most of them can be done in bootstrap (and I don't think pointer: text is useful ?)

src/components/side_panel/chart/building_blocks/annotation/annotation.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  private editorRef = signal<HTMLElement | null>(null);
+
+  private state: annotationState = proxy({
+    annotationTextInput: this.props.definition.annotationText || "",
+  });
+
+  get annotationPlaceholder() {
+    return _t("Add a description text");
+  }
```
hokolomopo: nitpick: define after the setup so the setup is the first method of the component

src/components/side_panel/chart/zoomable_chart/design_panel.xml
```diff
@@ -31,5 +31,6 @@
         />
```
hokolomopo, line 31: The feature doesn't work on the newly added bubble charts :/

src/components/side_panel/pivot/pivot_filter/pivot_filter.xml
```diff
@@ -14,7 +14,7 @@
         </Popover>
         <div class="d-flex">
           <i class="o-pivot-filter-caption">
-            <t t-esc="this.filterCaption()"/>
+            <t t-out="this.filterCaption()"/>
```
hokolomopo, line 17: probably a change for another task/another commit, no ?

src/components/side_panel/chart/building_blocks/annotation/annotation.ts
```diff
@@ -0,0 +1,73 @@
...
+import { SpreadsheetChildEnv } from "../../../../../types/spreadsheet_env";
+import { TextInput } from "../../../../text_input/text_input";
+import { SidePanelCollapsible } from "../../../components/collapsible/side_panel_collapsible";
+import { Section } from "../../../components/section/section";
+import { ChartSidePanelProps, chartSidePanelPropsDefinition } from "../../common";
+
+interface annotationState {
+  annotationTextInput: string;
+}
+export class ChartAnnotation extends Component<SpreadsheetChildEnv> {
```
hokolomopo, line 16:
```suggestion
}

export class ChartAnnotation extends Component<SpreadsheetChildEnv> {
```
nitpick

src/types/chart/common_chart.ts
```diff
@@ -5,21 +5,25 @@ import { AxesDesign, ChartDataSource, DataSetStyle, TitleDesign } from "./chart"
...
+/* All charts have these properties. */
+export interface BaseChartDefinition {
   readonly title: TitleDesign;
   readonly background?: Color;
+  readonly annotationText?: string;
+  readonly annotationLink?: string;
+  readonly humanize?: boolean;
+}
+
+export interface CommonChartDefinition<T extends string | Range = Range>
```
hokolomopo: DataSourceChartDefinition maybe ?

src/types/chart/funnel_chart.ts
```diff
@@ -1,23 +1,21 @@
...
 import { Color } from "../misc";
 import { Range } from "../range";
-import { AxesDesign, ChartDataSource, DataSetStyle, TitleDesign } from "./chart";
-import { LegendPosition } from "./common_chart";
+import { AxesDesign, ChartDataSource, DataSetStyle } from "./chart";
+import { BaseChartDefinition, LegendPosition } from "./common_chart";
 
-export interface FunnelChartDefinition<T extends string | Range = Range> {
+export interface FunnelChartDefinition<T extends string | Range = Range>
+  extends BaseChartDefinition {
```
hokolomopo: Why not CommonChartDefinition ?

tests/figures/chart/charts_component.test.ts
```diff
@@ -227,21 +233,37 @@ describe("charts", () => {
...
       createTestChart(chartType);
-      model.updateMode("dashboard");
       await nextTick();
       expect(fixture.querySelector(".o-figure")).not.toBeNull();
-      expect(fixture.querySelector(".o-figure-menu-item")).toBeNull();
+      expect(fixture.querySelector(".o-chart-item")).toHaveAttribute("data-id", "chartInfo");
     }
   );
 
+  test("charts does not have an info button if there is no annotation text or link", async () => {
```
hokolomopo:
```suggestion
  test("charts do not have an info button if there is no annotation text or link", async () => {
```

tests/figures/chart/charts_component.test.ts
```diff
@@ -254,6 +276,77 @@ describe("charts", () => {
...
+          "This is an annotation text"
+        );
+        expect(fixture.querySelector(".o-button-link")).toHaveText("See more");
+        expect(fixture.querySelector(".o-button-link")).toHaveAttribute(
+          "href",
+          "https://www.odoo.com"
+        );
+      }
+    );
+    test.each(["normal", "readonly", "dashboard"] as const)(
```
hokolomopo, line 315:
```suggestion
    );
    
    test.each(["normal", "readonly", "dashboard"] as const)(
```
nitpick

tests/test_helpers/constants.ts
```diff
@@ -34,13 +34,16 @@ export const TEST_CHART_DATA = {
     stacked: false,
     legendPosition: "top" as const,
     humanize: true,
+    annotationText: "This is an annotation text",
+    annotationLink: "https://www.odoo.com",
```
hokolomopo: Instead of adding the annotationText text here, can you manually add it in each test with an updateChart ? Reading the tests it isn't really clear that the chart have an annotationText, even more so when different chart types have a different annotated text

tests/figures/figure_component.test.ts
```diff
@@ -955,27 +956,41 @@ describe("figures", () => {
...
+    expect(fixture.querySelector(".o-figure")).not.toBeNull();
+    expect(fixture.querySelector(".o-figure-menu-item")).toBeNull();
+  });
+
   describe.each(["image", "basicChart", "scorecard", "gauge"])(
     "common tests for chart & image",
     (type: string) => {
       let sheetId: UID;
       let figureId: UID;
+      let menuQuery: string;
```
hokolomopo:
```suggestion
      let menuSelector: string;
```
nitpick

tests/figures/chart/charts_component.test.ts
```diff
@@ -254,6 +276,77 @@ describe("charts", () => {
...
+
+  test("withHttps helper adds https if the url does not start with https", async () => {
+    await mountSpreadsheet();
+    createTestChart("funnel");
+    await nextTick();
+    await click(fixture.querySelector(".o-chart-item")!);
+    expect(fixture.querySelector(".o-button-link")).toHaveAttribute(
+      "href",
+      "https://javascript:alert('hello world')"
+    );
```
hokolomopo, line 359: maybe a small comment that says `https://javascript:alert` is not valid and that adding https prevent javascript injection :)

hokolomopo (commented):
👋 

Having a common menu for all charts instead of the separate dashboard menu is really cool :) 

Let's try to merge this PR, it has been forgotten for quite a while 🙈

src/components/info_popover/info_popover.xml
matho-odoo: So that it can be translated but maybe there is another way ?

hokolomopo (approved):

src/components/side_panel/chart/building_blocks/annotation/annotation.css
```diff
@@ -0,0 +1,13 @@
+.o-spreadsheet {
+  .o-input-infotext {
+    height: fit-content;
+    max-height: 200px;
+    min-height: 1.2em;
+  }
+
+  .o-input-infotext:empty:before {
+    content: attr(data-placeholder);
```
rrahir, line 9: here; you could add a media query that adds some border to the div, see my comment in the task. That's something that Frgi should validate though.

rrahir (changes requested):

hokolomopo:
@robodoo r+

rrahir:
for future reference, the task trailer needs to be explicitely 
`Task: taskId`
 there is a space between the colon and the taskId

rrahir:
I need to check with runbot team because it matters in the commit msg, not the PR message... the style check is incorrect

PR #7867, opened 2026-01-29
[FIX] find_and_replace_store: crash after sheet deletion

dhrp-odoo:
## Description:

Steps:
- Create 2 sheets
- Focus Sheet1
- Open Find & Replace side panel
- Select a specific range
- Type "Sheet1!A1" and search a term
- Delete Sheet1 -> crash

Fix:
- If a specific range targets the deleted sheet, clear the range.

Related issue:
- If "Search in current sheet" is enabled and that sheet is deleted, search is not marked dirty, so the new active sheet highlights old matches.

Fix:
- Mark search as dirty when a sheet is deleted and "Search in current sheet" is enabled.

Task: [5453311](https://www.odoo.com/odoo/2328/tasks/5453311)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/find_and_replace/find_and_replace_store.ts
+++ b/src/components/side_panel/find_and_replace/find_and_replace_store.ts
@@ -119,6 +119,15 @@ export class FindAndReplaceStore extends SpreadsheetStore implements HighlightPr
           });
         }
         break;
+      case "DELETE_SHEET":
+        if (
+          this.searchOptions.searchScope === "specificRange" &&
+          this.searchOptions.specificRange?.sheetId === cmd.sheetId
+        ) {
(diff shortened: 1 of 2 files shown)


tests/find_and_replace/find_replace_side_panel_component.test.ts
```diff
@@ -190,6 +190,38 @@ describe("find and replace sidePanel component", () => {
       expect(getMatchesCount()).toMatchObject({ specificRange: 1 });
     });
 
+    test("clears specific range when its sheet is deleted", async () => {
```
hokolomopo: I'd test everything in the store, components tests are slower and I don't think you test much DOM-specific behavior there

PR #7913, opened 2026-02-04
[IMP] pivot: truncate dimension display name and add title

rmbh-odoo:
## Description:

Current behavior before PR:
- Long dimension display names were cropped and not fully readable.

- Dimension display names are truncated using text-truncate and include
  a title attribute to show the full value on hover.
- Only the first and last fields are displayed in the format
  `first field > ... > last field`, as these are the most relevant.
  
Technical:
- Update the type of dimension props to include displayName to ensure proper
  type safety.

Task: [5895856](https://www.odoo.com/odoo/2328/tasks/5895856)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/pivot/pivot_layout_configurator/pivot_dimension/pivot_dimension.ts
+++ b/src/components/side_panel/pivot/pivot_layout_configurator/pivot_dimension/pivot_dimension.ts
@@ -1,15 +1,16 @@
 import { Component } from "@odoo/owl";
-import { PivotCoreDimension, PivotCoreMeasure } from "../../../../..";
+import { PivotDimension as PivotDimensionType, PivotMeasure } from "../../../../..";
 import { GRAY_300 } from "../../../../../constants";
+import { collapseHierarchicalDisplayName } from "../../../../../helpers/pivot/pivot_helpers";
 import { SpreadsheetChildEnv } from "../../../../../types";
 import { css } from "../../../../helpers";
 import { TextInput } from "../../../../text_input/text_input";
 import { CogWheelMenu } from "../../../components/cog_wheel_menu/cog_wheel_menu";
 
 interface Props {
-  dimension: PivotCoreDimension | PivotCoreMeasure;
-  onRemoved: (dimension: PivotCoreDimension | PivotCoreMeasure) => void;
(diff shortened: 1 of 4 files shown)


src/components/side_panel/pivot/pivot_layout_configurator/pivot_dimension/pivot_dimension.xml
```diff
@@ -4,7 +4,9 @@
       class="py-1 px-2 d-flex flex-column shadow-sm pivot-dimension"
       t-att-class="{'pivot-dimension-invalid': !props.dimension.isValid}">
       <div class="d-flex flex-row justify-content-between align-items-center">
-        <div class="d-flex align-items-center overflow-hidden text-nowrap">
+        <div
+          class="d-flex align-items-center overflow-hidden text-nowrap"
+          t-att-title="props.dimension.displayName">
```
hokolomopo, line 9: could you also maybe add a text ellipsis ?

src/components/side_panel/pivot/pivot_layout_configurator/pivot_dimension/pivot_dimension.ts
```diff
@@ -1,15 +1,19 @@
 import { Component } from "@odoo/owl";
-import { PivotCoreDimension, PivotCoreMeasure } from "../../../../..";
+import {
+  PivotDimension as PivotDimensionType,
+  PivotMeasure as PivotMeasureType,
```
LucasLefevre: Why this change ? is it necessary ?
It's easier to read when keeping the names (if they are not conflicting with anything else)

LucasLefevre (commented):
LGTM 👌 

Only a small detail

src/components/side_panel/pivot/pivot_layout_configurator/pivot_dimension/pivot_dimension.ts
rmbh-odoo: Here, the `dimension` should be typed as `PivotDimension` or `PivotMeasure` because at runtime it includes `displayName`, so the type needs to be updated. 
This caused a conflict with the existing PivotDimension name, which is why I initially considered renaming both types.
Never mind,  I’ve now only changed the conflicting type name. :smile:

PR #7924, opened 2026-02-06
[FIX] session: release memory kept for debounce

LucasLefevre:
The timeout of the debounced function `debouncedMove` keeps a reference to the session (and the history, and basically everything).

It means it cannot be GC'ed before at least 200ms. It's not a real issue in practice for users.

But it becomes an issue when instantiating a model with node.js for scripting purposes (like performance benchmark where you instantiate many models). For this use-case, you want the memory to be freed as soon as possible.

It's even worse if your script is synchronous: it blocks the main thread and never lets the even loop execute other tasks (including the timeout)

With this commit, we remove the debounce. It's now the responsibility of the transport service to debounce messages if it wants. (see odoo PR)

Task: [5916695](https://www.odoo.com/odoo/2328/tasks/5916695)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


rrahir:
We should consider updating the demo transport accordingly ^^

LucasLefevre:
do we care ?

rrahir:
not that much ,true

pro-odoo:
robodoo r+

PR #7925, opened 2026-02-06
[IMP] Format, Style: add default value on Sheet, Cols and Rows

fdamhaut:
## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/misc.ts
+++ b/src/helpers/misc.ts
@@ -267,7 +267,7 @@ export function isObjectEmptyRecursive<T extends object>(argument: T | undefined
     return true;
   }
   return Object.values(argument).every((value) =>
-    typeof value === "object" ? isObjectEmptyRecursive(value) : !value
+    typeof value === "object" ? isObjectEmptyRecursive(value) : value === undefined
   );
 }
 
@@ -804,4 +804,25 @@ export function getMissingHeadersForSpreadResult(
   const missingRows = row + evaluated[0].length - numberOfRows;
   const missingCols = col + evaluated.length - numberOfCols;
   return { missingRows, missingCols };
+}
+
+/*
+ * Simple object that mimic the defaultDict behaviour in python
+ * It is a dictionary with the get function setting returning a copy of the default value
+ * if when accessing previously unset values
+ */
+export function defaultDict<T>(def: T) {
+  const state: Record<string, T> = {};
+  return {
+    state,
+    get: (key: string): T => {
+      if (!(key in state)) {
+        state[key] = deepCopy(def);
+      }
+      return state[key];
+    },
+    set: (key: string, value: T) => {
+      state[key] = value;
+    },
+  };
 }
--- a/src/plugins/core/cell.ts
+++ b/src/plugins/core/cell.ts
@@ -1,11 +1,10 @@
-import { DEFAULT_NUMBER_STYLE, DEFAULT_STYLE } from "../../constants";
 import { PositionMap } from "../../helpers/cells/position_map";
 import {
   getItemId,
   groupItemIdsByZones,
   iterateItemIdsPositions,
 } from "../../helpers/data_normalization";
-import { deepEquals, range, replaceNewLines } from "../../helpers/misc";
+import { deepEquals, isObjectEmptyRecursive, range, replaceNewLines } from "../../helpers/misc";
 
 import { toXC } from "../../helpers/coordinates";
 import { CorePlugin } from "../core_plugin";
@@ -18,7 +17,6 @@ import {
   CommandResult,
   CoreCommand,
   PositionDependentCommand,
-  SetFormattingCommand,
   UpdateCellCommand,
 } from "../../types/commands";
 import { CellPosition, HeaderIndex, RangeAdapterFunctions, UID } from "../../types/misc";
@@ -29,10 +27,8 @@ import {
   createFormulaCellFromCompiledFormula,
 } from "../../helpers/cells/cell_evaluation";
 import { isExcelCompatible } from "../../helpers/format/format";
-import { isNumber } from "../../helpers/numbers";
 import { recomputeZones } from "../../helpers/recompute_zones";
 import { Format } from "../../types/format";
-import { DEFAULT_LOCALE } from "../../types/locale";
 import { Style, UpdateCellData, Zone } from "../../types/misc";
 import { Range, RangePart } from "../../types/range";
 import { ExcelWorkbookData, WorkbookData } from "../../types/workbook_data";
@@ -56,7 +52,6 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
     "zoneToXC",
     "getCells",
     "getTranslatedCellFormula",
-    "getCellStyle",
     "getCellById",
     "getFormulaString",
     "getFormulaMovedInSheet",
@@ -97,8 +92,6 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
         return !cmd.cellId || this.cells[cmd.sheetId]?.[cmd.cellId]
           ? CommandResult.Success
           : CommandResult.InvalidCellId;
-      case "SET_FORMATTING":
-        return this.checkUselessSetFormatting(cmd);
       default:
         return CommandResult.Success;
     }
@@ -106,12 +99,6 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
 
   handle(cmd: CoreCommand) {
     switch (cmd.type) {
-      case "SET_FORMATTING":
-        this.setStyleFormat(cmd.sheetId, cmd.target, cmd.style, cmd.format);
-        break;
-      case "CLEAR_FORMATTING":
-        this.clearFormatting(cmd.sheetId, cmd.target);
-        break;
       case "ADD_COLUMNS_ROWS":
         if (cmd.dimension === "COL") {
           this.handleAddColumnsRows(cmd, this.copyColumnStyle.bind(this));
@@ -130,7 +117,7 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
           row: cmd.row,
           content: "",
           style: null,
-          format: "",
+          format: null,
         });
         break;
 
@@ -166,25 +153,6 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
   }
 
   /**
-   * Clear the styles and format of zones
-   */
-  private clearFormatting(sheetId: UID, zones: Zone[]) {
-    for (const zone of recomputeZones(zones)) {
-      for (let col = zone.left; col <= zone.right; col++) {
-        for (let row = zone.top; row <= zone.bottom; row++) {
-          this.dispatch("UPDATE_CELL", {
-            sheetId,
-            col,
-            row,
-            style: null,
-            format: "",
-          });
-        }
-      }
-    }
-  }
-
-  /**
    * Clear the styles, the format and the content of zones
    */
   private clearCells(sheetId: UID, zones: Zone[]) {
@@ -197,7 +165,7 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
             row,
             content: "",
             style: null,
-            format: "",
+            format: null,
           });
         }
       }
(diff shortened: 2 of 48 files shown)


src/types/workbook_data.ts
```diff
@@ -50,7 +51,9 @@ export interface SheetData {
   rowNumber: number;
   cells: { [key: string]: string | undefined | SquishedCell };
   styles: { [zone: string]: number };
+  defaultStyle?: defaultValue<number>;
   formats: { [zone: string]: number };
+  defaultFormat?: defaultValue<number>;
```
VincentSchippefilt: using this data structure for the export is not a good idea: it will generate an element for each row/col that is null, until the last default is defined. So if I have 1000 rows and define the last one as bold, I will have 999 null, followed by the 1000th item with my style.
For the import/export to json, could we have a Dict of defaults please?

src/plugins/ui_core_views/custom_colors.ts
```diff
@@ -134,6 +134,7 @@ export class CustomColorsPlugin extends CoreViewPlugin<CustomColorState> {
   private getColorsFromCells(sheetId: UID): Color[] {
     const cells: Cell[] = Object.values(this.getters.getCells(sheetId));
     const colors: Set<Color> = new Set();
+    // TODO getDefault Values colors
```
VincentSchippefilt: c'est tout doux

src/plugins/ui_feature/autofill.ts
```diff
@@ -436,7 +436,7 @@ export class AutofillPlugin extends UIPlugin {
       content,
       style,
       border,
-      format,
+      format: format ?? undefined,
```
VincentSchippefilt, line 439: is it to covert null into undefined?

src/plugins/ui_feature/sort.ts
```diff
@@ -179,6 +179,7 @@ export class SortPlugin extends UIPlugin {
           } else {
             content = cell.content;
           }
+          // TODO default
```
VincentSchippefilt: c'est tout doux

src/plugins/ui_stateful/selection.ts
```diff
@@ -361,7 +361,7 @@ export class GridSelectionPlugin extends UIPlugin {
   getCurrentStyle(): Style {
     const zone = this.getters.getSelectedZone();
     const sheetId = this.getters.getActiveSheetId();
-    return this.getters.getCellStyle({ sheetId, col: zone.left, row: zone.top });
+    return this.getters.getCellComputedStyle({ sheetId, col: zone.left, row: zone.top });
```
VincentSchippefilt, line 364: I don't think it is correct: if would show the style of a CF as the background

ex.: select the B column, that is green because of conditional format, it will show as background color green in the toolbar

{"version":"19.1.0","sheets":[{"id":"Sheet1","name":"Sheet1","colNumber":2,"rowNumber":2,"rows":{},"cols":{},"merges":[],"cells":{},"styles":{},"formats":{},"borders":{},"conditionalFormats":[{"id":"47dea4ce-0435","rule":{"type":"CellIsRule","operator":"isEmpty","style":{"fillColor":"#b6d7a8"},"values":[]},"ranges":["B:B"]}],"dataValidationRules":[],"figures":[],"tables":[],"areGridLinesVisible":true,"isVisible":true,"headerGroups":{"ROW":[],"COL":[]}}],"styles":{},"formats":{},"borders":{},"revisionId":"6d19727f-f8fe-4f59-8a38-f7829764d2a5","uniqueFigureIds":true,"settings":{"locale":{"name":"English (US)","code":"en_US","thousandsSeparator":",","decimalSeparator":".","weekStart":7,"dateFormat":"m/d/yyyy","timeFormat":"hh:mm:ss a","formulaArgSeparator":","}},"pivots":{},"pivotNextId":1,"customTableStyles":{},"namedRanges":{}}

src/plugins/core/default.ts
```diff
@@ -0,0 +1,837 @@
...
+      for (let row = zone.top; row <= zone.bottom; row++) {
+        this.updateCellFormat({ sheetId, col, row }, format, option);
+      }
+    }
+  }
+
+  private updateCellFormat(
+    position: CellPosition,
+    format: Format | null,
+    option?: { cellPriority?: boolean; force?: boolean }
```
VincentSchippefilt: I don't understand the concept of 'cellPriority'. 
IMHO booleans should be named isSomething to give an indication of what the true value does. 
Also, I can't find any usage of cellPriority

src/plugins/core/default.ts
```diff
@@ -0,0 +1,837 @@
...
+        }
+      }
+    }
+  }
+
+  private getPartialDefaultStyleInCell(
+    sheetId: UID,
+    zones: Zone[],
+    newDefaultStyle: Style,
+    newHasPriorityOver: { col?: boolean; row?: boolean; sheet?: boolean }
```
VincentSchippefilt: I have a hard time understanding the usage here. 
Would this be correct : 

```suggestion
    priorities: { shouldUseDefaultCol?: boolean; shouldUseDefaultRow?: boolean; shouldUseDefaultSheet?: boolean }
```

src/plugins/core/cell.ts
```diff
@@ -368,16 +337,17 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
...
-    const cleanedStyle = { ...cell.style };
-    const defaultStyle =
-      !cell.isFormula && isNumber(cell.content, DEFAULT_LOCALE)
-        ? DEFAULT_NUMBER_STYLE
-        : DEFAULT_STYLE;
+  private extractCustomStyle(position: CellPosition, cell: Cell): Style {
+    const cleanedStyle = this.getters.getCellStyle(position, cell);
+    if (!cell) {
+      return {};
+    }
```
VincentSchippefilt:
```suggestion
       if (!cell) {
      return {};
    }
     const cleanedStyle = this.getters.getCellStyle(position, cell);

```

src/plugins/core/default.ts
```diff
@@ -0,0 +1,837 @@
...
+      }
+    }
+    return partialDefaults;
+  }
+
+  // ---------------------------------------------------------------------------
+  // Getters
+  // ---------------------------------------------------------------------------
+
+  getCellStyle(position: CellPosition, cell?: Cell): Style {
```
VincentSchippefilt, line 604: I don't like the variable geometry of this function, would removing the cell parameter cost a lot (in perf)?

src/plugins/core/default.ts
```diff
@@ -0,0 +1,837 @@
...
+    cell = cell || this.getters.getCell(position);
+    const style = { ...cell?.style };
```
VincentSchippefilt: can we return cell.style without making a copy in the case where there are no styleSheet?

src/plugins/core/default.ts
```diff
@@ -0,0 +1,837 @@
...
+    return (
+      styleSheet?.rowDefault?.[position.row] ??
+      styleSheet?.colDefault?.[position.col] ??
+      styleSheet?.sheetDefault ??
+      DEFAULT_STYLE[key]
+    );
+  }
+
+  getCellFormat(position: CellPosition, cell?: Cell): Format | undefined {
+    cell = cell || this.getters.getCell(position);
```
VincentSchippefilt, line 637: same as getCellStyle: I would remove the cell param unless it'a big cost in perf

src/helpers/misc.ts
```diff
@@ -805,3 +805,31 @@ export function getMissingHeadersForSpreadResult(
   const missingCols = col + evaluated.length - numberOfCols;
   return { missingRows, missingCols };
 }
+
+export function isObjectEmpty(obj: Object) {
```
VincentSchippefilt: Can we have a small comment to explain what is considered empty, please?

src/helpers/misc.ts
```diff
@@ -805,3 +805,31 @@ export function getMissingHeadersForSpreadResult(
...
+  }
+  for (const value of Object.values(obj)) {
+    if (value !== undefined && (typeof value !== "object" || !isObjectEmpty(value))) {
+      return false;
+    }
+  }
+  return true;
+}
+
+export function defaultDict<T>(def: T) {
```
VincentSchippefilt, line 814: Also a small comment for the poor soul that do not know Python?

src/types/commands.ts
```diff
@@ -371,7 +371,7 @@ export interface UpdateCellCommand extends PositionDependentCommand {
   type: "UPDATE_CELL";
   content?: string;
   style?: Style | null;
-  format?: Format;
+  format?: Format | null;
```
VincentSchippefilt, line 374: where does the null come from?

VincentSchippefilt (changes requested):
I still need to review
- cell_clipboard
- default_clipboard
- cell_evaluation
- cell Plugin
- all the tests

src/plugins/core/default.ts
fdamhaut: It seems these options were an artifact from a previous implementation 🤭

src/helpers/misc.ts
fdamhaut: after searching for the function to comment it, I found `isObjectEmptyRecursive` and use it instead

src/types/commands.ts
fdamhaut: null reprensent the "removal" of format on a cell to allow for the cell to use the default format of that position.
Previously we used `""` to signify no format but now `""` represent the value no format on a cell and not "use the default format for that cell"

src/plugins/ui_feature/autofill.ts
fdamhaut: yes, it's a side-effect of the change to the updateCell typing (as the cellData's type is dependent on it)

src/plugins/core/default.ts
fdamhaut: That function is used to get the default values in a zone that is going to be overwritten. Your wording is more understandable 👍.

src/plugins/ui_stateful/selection.ts
fdamhaut: IIRC this was a to fix a performance issue, as the function is called 1000 times by all the style-related components. 
This felt like a nice fix which use the computedStyle cache as well as align the component behaviour to be more wysiwyg. I could implement a cache for this if you strongly feel like the old behaviour should be kept but at the time it seemed like a overengineer solution to a non-problem.

VincentSchippefilt: I would rather keep the old behavior, and if there is a performance issue, we will fix it. 
This also shows that there is a test missing, that shows that the component doesn't react to the style of a CF (until we decide to change it)

tests/default/default_plugin_format.test.ts
```diff
@@ -0,0 +1,1012 @@
...
+    sheetId = model.getters.getActiveSheetId();
+  });
+
+  describe.each(TEST_FORMATS)("Format : %s", (format) => {
+    test("Can set format on sheet", () => {
+      model.dispatch("SET_FORMATTING", {
+        sheetId,
+        target: [model.getters.getSheetZone(sheetId)],
+        format,
+      });
```
VincentSchippefilt, line 42: in general, we dispatch to models using a helper, it reduces duplicate code and make it easy to modify if we modify the parameters of the command. 
The helper usually uses the active sheet as the default sheet

tests/default/default_plugin_format.test.ts
```diff
@@ -0,0 +1,1012 @@
...
+        target: [model.getters.getSheetZone(sheetId)],
+        format: DATE_FORMAT,
+      });
+      insertCells(model, "B2", "right");
+      expect(getCellFormat(model, "B2")).toEqual(DATE_FORMAT); // ??
+      expect(getCellFormat(model, "C2")).toEqual(DATE_FORMAT);
+      expect(model.getters.getCells(sheetId).length).toBe(1);
+    });
+  });
+
```
VincentSchippefilt, line 786: might add "sheet manipulation Merge Unmerge", when the cell that is merged as the dimension of a row/col/sheet, when it is unmerged, all the cells keep the same style

VincentSchippefilt: and probably the same for Merge Unmerge keep the format

VincentSchippefilt:
robodoo r+

VincentSchippefilt:
robodoo rebase-ff

PR #7928, opened 2026-02-10
[FIX] dynamic_tables: trim overlap on the correct side

dhrp-odoo:
## Description:

When reducing a dynamic table to avoid overlap with a static table, we always trimmed the right edge.

That is wrong when the static table overlaps the dynamic anchor column: the computed zone can become invalid (`right < left`), which later crashes table style computation.

Trim logic is now directional:
- trim `right` only when the static table starts to the right of the anchor
- otherwise trim `bottom`

Task: [5905900](https://www.odoo.com/odoo/2328/tasks/5905900)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


hokolomopo:
@robodoo r+

PR #7930, opened 2026-02-10
[FIX] charts: chart have wrong background at copy/download

hokolomopo:
## Description

The background of the charts are usually set in the CSS of the
`ChartJsComponent`. But for some features, we need to convert the chart
into an image (copy to clipboard, download as image, xlsx export, ...).

In this case, we added on the fly a `chartBackgroundPlugin` to the runtime that would draw the background color. But we were always drawing a white background.

With this commit, we always draw the correct background in chartJS, so it works in both the component and the exported image.

Task: [5925821](https://www.odoo.com/odoo/2328/tasks/5925821)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/figures/chart/chartJs/chartjs_background_plugin.ts
+++ b/src/components/figures/chart/chartJs/chartjs_background_plugin.ts
@@ -1,0 +1,25 @@
+import { BACKGROUND_CHART_COLOR } from "@odoo/o-spreadsheet-engine/constants";
+import { ChartType, Plugin } from "chart.js";
+
+export interface ChartBackgroundPluginOptions {
+  color: string | undefined;
+}
+
+declare module "chart.js" {
+  interface PluginOptionsByType<TType extends ChartType> {
+    background?: ChartBackgroundPluginOptions;
+  }
+}
+
+/** This is a chartJS plugin that will draw a background of a chart */
+export const chartBackgroundPlugin: Plugin = {
+  id: "background",
+  beforeDraw(chart: any, args, options: ChartBackgroundPluginOptions) {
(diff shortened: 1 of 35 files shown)


src/components/figures/chart/chartJs/chartjs_background_plugin.ts
```diff
@@ -0,0 +1,25 @@
...
+  color: string | undefined;
+}
+
+declare module "chart.js" {
+  interface PluginOptionsByType<TType extends ChartType> {
+    background?: ChartBackgroundPluginOptions;
+  }
+}
+
+/** This is a chartJS plugin that will draw the values of each data next to the point/bar/pie slice */
```
pro-odoo: Not really :D

src/components/figures/figure_carousel/figure_carousel.ts
```diff
@@ -113,9 +116,14 @@ export class CarouselFigure extends Component<Props, SpreadsheetChildEnv> {
     const cssProperties: CSSProperties = {};
     if (this.selectedCarouselItem?.type === "chart") {
       const chart = this.env.model.getters.getChartRuntime(this.selectedCarouselItem.chartId);
-      cssProperties["background-color"] = chart.background;
+      if ("background" in chart && chart.background) {
+        cssProperties["background-color"] = chart.background || BACKGROUND_CHART_COLOR;
```
pro-odoo:
```suggestion
        cssProperties["background-color"] = chart.background
```

hokolomopo:
@robodoo r+

PR #7935, opened 2026-02-11
Master fix repeat vsc

VincentSchippefilt:
## Description:

Task: 0

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/packages/o-spreadsheet-engine/src/plugins/core/cell.ts
+++ b/packages/o-spreadsheet-engine/src/plugins/core/cell.ts
@@ -70,7 +70,7 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
     "getFormulaString",
     "getFormulaMovedInSheet",
   ] as const;
-  readonly nextId = 1;
+  nextId = 1;
   public readonly cells: { [sheetId: string]: { [id: string]: Cell } } = {};
 
   adaptRanges({ applyChange }: RangeAdapterFunctions, sheetId: UID, sheetName: AdaptSheetName) {
@@ -556,9 +556,7 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
   }
 
   private getNextCellId(): number {
-    const id = this.nextId;
-    this.history.update("nextId", this.nextId + 1);
-    return id;
+    return this.nextId++;
   }
 
   private updateCell(sheetId: UID, col: HeaderIndex, row: HeaderIndex, after: UpdateCellData) {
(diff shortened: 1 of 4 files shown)


VincentSchippefilt:
robodoo rebase-ff

packages/o-spreadsheet-engine/src/plugins/core/cell.ts
```diff
@@ -70,7 +70,7 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
     "getFormulaString",
     "getFormulaMovedInSheet",
   ] as const;
-  readonly nextId = 1;
+  nextId = 1;
   public readonly cells: { [sheetId: string]: { [id: string]: Cell } } = {};
```
rrahir, line 74: I don't think it is true, the cell Id is shared to the Sheet plugin (through core commands which are never root commands but still) and are used in some getters and even used to compute specific UI values (https://github.com/odoo/o-spreadsheet/blob/19.0/src/plugins/ui_core_views/header_sizes_ui.ts#L211-L219)
Here is a collaborative test that will fail with this change

```typescript
test("coucou", () => {
  const { network, alice, bob, charlie } = setupCollaborativeEnv();
    network.concurrent(() => {
      setCellContent(alice, "A1", "hello");
      setCellContent(bob, "A2", "coucou");
    });
    addColumns(charlie, "before", "A", 1);
    expect([alice, bob, charlie]).toHaveSynchronizedValue((user) => user.getters.getCellById(1)?.content, "hello");
})
```

packages/o-spreadsheet-engine/src/plugins/ui_core_views/custom_colors.ts
```diff
@@ -75,7 +75,7 @@ interface CustomColorState {
  */
 export class CustomColorsPlugin extends CoreViewPlugin<CustomColorState> {
   private readonly customColors: Immutable<Record<Color, true>> = {};
-  private readonly shouldUpdateColors = true;
+  private shouldUpdateColors = true;
```
rrahir, line 78: we still update the history  here, you only removed a typing constraint. I think that `CustomColorsPlugin` never really created a history step change because if a command is rejected, then the handler never receives the command and cannot update the history.

AFAICS, the problem you mention in the commit message
> Imagine sorting an already
sorted range, then selecting another range and F4, you want this new range to be sorted.
Fix this too.

is already present in master. If I make an operation with change (like changing the background color of a cell) then try to sort a range and repeat the last command on another range, I just color the background of the latter.

rrahir (changes requested):
👋

PR #7937, opened 2026-02-11
[FIX] charts: hierarchical charts should show formatted labels instead of raw

rmbh-odoo:
## Description:

Current behavior before PR:
- Hierarchical charts interpret datasets as labels and the last column as values. dataset cells were read as raw values, which is correct for other charts but incorrect here because these values are rendered as labels.
- As a result, formatted values were displayed incorrectly in charts like Sunburst and Treemap.

- Dataset columns used as labels now use formatted cell values instead of raw values, ensuring dates and other formatted types display correctly in hierarchical charts.

Task: [5913296](https://www.odoo.com/odoo/2328/tasks/5913296)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/figures/charts/runtime/chart_data_extractor.ts
+++ b/src/helpers/figures/charts/runtime/chart_data_extractor.ts
@@ -13,6 +13,7 @@ import {
   CellValue,
   DEFAULT_LOCALE,
   Format,
+  FormattedValue,
   GenericDefinition,
   Getters,
   Locale,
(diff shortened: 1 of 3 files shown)


src/helpers/figures/charts/runtime/chart_data_extractor.ts
```diff
@@ -654,20 +655,34 @@ function fixEmptyLabelsForDateCharts(
...
+  const range = getDataRange(getters, ds);
+  if (!range) return [];
+
+  return getters.getRangeValues(range).map((value) => (value === "" ? undefined : value));
+}
+
+/**
+ * Get the formatted data from a dataSet
+ */
+function getFormattedData(getters: Getters, ds: DataSet): (string | undefined)[] {
```
dhrp-odoo:
```suggestion
function getFormattedData(getters: Getters, ds: DataSet): (FormattedValue | undefined)[] {
```

Nit: return type. Looks good otherwise.

src/helpers/figures/charts/runtime/chart_data_extractor.ts
```diff
@@ -654,20 +655,34 @@ function fixEmptyLabelsForDateCharts(
   return { labels: newLabels, dataSetsValues: newDatasets };
 }
 
+function getDataRange(getters: Getters, ds: DataSet): Range | undefined {
```
hokolomopo:
```suggestion
function getDatasetRange(getters: Getters, ds: DataSet): Range | undefined {
```
nitpick

rrahir:
just changed the one-line if statements so that we avoid conflit on fwp

@robodoo r+

rrahir:
@robodoo r+

PR #7938, opened 2026-02-11
[IMP] spreadsheet: implement a print wizard

hokolomopo:
### [IMP] spreadsheet: implement a print wizard

This commit implements a print wizard for spreadsheets, allowing users
to customize their print settings and preview the output before printing.

To achieve this, we need to be able to draw any part of a spreadsheet,
not limited to the active viewport of `SheetViewPlugin`. The rendering
logic now lose access to the SheetView/Selection getters, and are instead
passed a `GridRenderingContext` object containing all the necessary
information.

Task: [5891329](https://www.odoo.com/odoo/2328/tasks/5891329)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/spreadsheet_print/spreadsheet_print.ts
+++ b/src/components/spreadsheet_print/spreadsheet_print.ts
@@ -1,0 +1,82 @@
+import { SpreadsheetChildEnv } from "@odoo/o-spreadsheet-engine/types/spreadsheet_env";
+import { Component, onWillUnmount, useExternalListener } from "@odoo/owl";
+import { Store, useLocalStore } from "../../store_engine";
+import { cssPropertiesToCss } from "../helpers";
+import { Select } from "../select/select";
+import { BadgeSelection } from "../side_panel/components/badge_selection/badge_selection";
+import { Checkbox } from "../side_panel/components/checkbox/checkbox";
+import { Section } from "../side_panel/components/section/section";
+import { StandaloneGridCanvas } from "../standalone_grid_canvas/standalone_grid_canvas";
+import {
+  Orientation,
+  PrintPageLayout,
+  PrintScale,
+  PrintSelection,
+  SpreadsheetPrintStore,
+} from "./spreadsheet_print_store";
+
+interface Props {
+  onExitPrintMode: () => void;
+}
+
+export class SpreadsheetPrint extends Component<Props, SpreadsheetChildEnv> {
+  static template = "o-spreadsheet-SpreadsheetPrint";
+  static props = { onExitPrintMode: Function };
+  static components = { StandaloneGridCanvas, Section, Select, BadgeSelection, Checkbox };
+
+  printStore!: Store<SpreadsheetPrintStore>;
+
+  setup() {
+    this.printStore = useLocalStore(SpreadsheetPrintStore);
+    let styleElement: HTMLStyleElement | null = null;
+    useExternalListener(window, "beforeprint", () => {
+      styleElement = document.createElement("style");
+      styleElement.id = "o-spreadsheet-print-style";
+      const size = `${this.printStore.pageLayout} ${this.printStore.orientation}`;
+      styleElement.textContent = `@media print { @page { size: ${size}; margin: ${this.printStore.printMargin}px;}}`;
+      document.head.appendChild(styleElement);
+    });
+    const removePrintStyle = () => {
+      if (styleElement) {
+        document.head.removeChild(styleElement);
+        styleElement = null;
+      }
+    };
+    useExternalListener(window, "afterprint", () => removePrintStyle());
+    onWillUnmount(() => removePrintStyle());
+  }
+
+  get pageStyle(): string {
+    const { width, height } = this.printStore.pageDimensionsInPixels;
+    return cssPropertiesToCss({
+      width: `${width}px`,
+      height: `${height}px`,
+      padding: `${this.printStore.printMargin}px`,
+    });
+  }
+
+  onLayoutChange(value: PrintPageLayout) {
+    this.printStore.changePrintLayout(value);
+  }
+
+  onPrintSelectionChange(value: PrintSelection) {
+    this.printStore.changePrintSelection(value);
+  }
+
+  onPrintScaleChange(value: PrintScale) {
+    this.printStore.changePrintScale(value);
+  }
+
+  setGridLinesVisibility(value: boolean) {
+    this.printStore.setGridLinesVisibility(!value);
+  }
+
+  changeOrientation(value: Orientation) {
+    this.printStore.changePrintOrientation(value);
+  }
+
+  onPrint() {
+    window.print();
+    this.props.onExitPrintMode();
+  }
+}
--- a/src/components/spreadsheet_print/spreadsheet_print.xml
+++ b/src/components/spreadsheet_print/spreadsheet_print.xml
@@ -1,0 +1,96 @@
+<templates>
+  <t t-name="o-spreadsheet-SpreadsheetPrint">
+    <div class="o-spreadsheet-print d-flex flex-column h-100 w-100">
+      <t t-set="pages" t-value="printStore.printPages"/>
+      <div class="o-print-header flex-shrink-0 d-flex align-items-center d-print-none">
+        <h3 class="m-0 p-3 ps-4 text-white">Print Preview</h3>
+        <span class="text-white">
+          <t t-esc="pages.length"/>
+          pages
+        </span>
+        <button class="o-button ms-auto me-2 flex-grow-0" t-on-click="props.onExitPrintMode">
+          Cancel
+        </button>
+        <button class="o-button primary flex-grow-0 me-4" t-on-click="onPrint">Next</button>
+      </div>
+      <div class="o-print-container d-flex flex-grow-1 overflow-hidden">
+        <div class="o-print-preview w-100 overflow-auto d-flex flex-column">
+          <div class="mx-auto">
+            <div
+              t-if="pages.length === 0"
+              class="o-print-page o-empty-print-page m-4 shadow d-flex flex-column align-items-center justify-content-center flex-shrink-0"
+              t-att-style="pageStyle">
+              <h4 class="fst-italic text-muted">No content to print</h4>
+            </div>
+            <div
+              t-else=""
+              t-foreach="pages"
+              t-as="page"
+              t-key="page_index"
+              class="o-print-page m-4 shadow d-flex justify-content-center flex-shrink-0"
+              t-att-style="pageStyle">
+              <StandaloneGridCanvas
+                sheetId="page.sheetId"
+                zone="page.zone"
+                renderingCtx="page.renderingCtx"
+              />
+            </div>
+          </div>
+        </div>
+        <div class="o-sidePanel flex-shrink-0 bg-white border-top border-start d-print-none">
+          <div class="o-sidePanelBody pt-4">
+            <Section t-if="!this.env.isDashboard()">
+              <div class="o-section-title">Print content</div>
+              <Select
+                selectedValue="printStore.printSelection"
+                values="printStore.printSelectionOptions"
+                class="'o-print-selection'"
+                onChange.bind="this.onPrintSelectionChange"
+              />
+            </Section>
+
+            <Section>
+              <div class="o-section-title">Scale</div>
+              <Select
+                selectedValue="printStore.printScale"
+                values="printStore.printScaleOptions"
+                class="'o-print-scale'"
+                onChange.bind="this.onPrintScaleChange"
+              />
+            </Section>
+
+            <Section>
+              <div class="o-section-title">Paper</div>
+              <Select
+                selectedValue="printStore.pageLayout"
+                values="printStore.layoutOptions"
+                class="'o-print-layout'"
+                onChange.bind="this.onLayoutChange"
+              />
+            </Section>
+
+            <Section>
+              <div class="o-section-title">Orientation</div>
+              <BadgeSelection
+                choices="printStore.orientationChoices"
+                onChange.bind="this.changeOrientation"
+                selectedValue="printStore.orientation"
+              />
+            </Section>
+
+            <Section t-if="!this.env.isDashboard()">
+              <div class="o-section-title">Formatting</div>
+              <Checkbox
+                name="'showGridLines'"
+                value="!this.printStore.hideGridLines"
+                onChange.bind="this.setGridLinesVisibility"
+                className="'mt-2'"
+                label.translate="Show gridlines"
+              />
+            </Section>
+          </div>
+        </div>
+      </div>
+    </div>
+  </t>
+</templates>
--- a/src/components/spreadsheet_print/spreadsheet_print_store.ts
+++ b/src/components/spreadsheet_print/spreadsheet_print_store.ts
@@ -1,0 +1,340 @@
+import {
+  _t,
+  Dimension,
+  DOMDimension,
+  GridRenderingContext,
+  HeaderIndex,
+  UID,
+  ValueAndLabel,
+  Zone,
+} from "@odoo/o-spreadsheet-engine";
+import { ViewportCollection } from "@odoo/o-spreadsheet-engine/helpers/viewport_collection";
+import { cellPositions, intersection } from "../../helpers";
+import { SpreadsheetStore } from "../../stores";
+import { CellValueType, FigureUI } from "../../types";
+
+// Page layouts in millimeters
+const PRINT_PAGES_LAYOUTS = {
+  A3: { width: 297, height: 420, name: _t("A3 (297 x 420 mm)") },
+  A4: { width: 210, height: 297, name: _t("A4 (210 x 297 mm)") },
+  A5: { width: 148, height: 210, name: _t("A5 (148 x 210 mm)") },
+  B4: { width: 250, height: 353, name: _t("B4 (250 x 353 mm)") },
+  B5: { width: 176, height: 250, name: _t("B5 (176 x 250 mm)") },
+  Letter: { width: 216, height: 279, name: _t("Letter (216 x 279 mm)") },
+  Legal: { width: 216, height: 356, name: _t("Legal (216 x 356 mm)") },
+};
+
+export type PrintPageLayout = keyof typeof PRINT_PAGES_LAYOUTS;
+export type PrintSelection = "currentSheet" | "entireWorkbook" | "selection";
+export type PrintScale = "fitToWidth" | "fitToHeight" | "actualSize";
+export type Orientation = "portrait" | "landscape";
+
+const PX_PER_MM = 3.78;
+
+interface PagePrintInfo {
+  sheetId: UID;
+  zone: Zone;
+  renderingCtx: Partial<GridRenderingContext>;
+}
+
+export class SpreadsheetPrintStore extends SpreadsheetStore {
+  mutators = [
+    "changePrintLayout",
+    "changePrintSelection",
+    "changePrintScale",
+    "changePrintOrientation",
+    "setGridLinesVisibility",
+  ] as const;
+
+  pageLayout = "A4";
+  printSelection = "currentSheet";
+  printScale = "fitToWidth";
+  orientation = "portrait";
+  hideGridLines =
+    this.getters.isDashboard() ||
+    !this.getters.getGridLinesVisibility(this.getters.getActiveSheetId());
+
+  get printPages(): PagePrintInfo[] {
+    if (this.printSelection === "selection") {
+      const sheetId = this.getters.getActiveSheetId();
+      return this.splitZoneToPrintPages(sheetId, this.getters.getSelectedZone());
+    }
+    const sheetIds =
+      this.printSelection === "entireWorkbook"
+        ? this.getters.getSheetIds()
+        : [this.getters.getActiveSheetId()];
+    return sheetIds.flatMap((sheetId) => {
+      const { lastUsedCol, lastUsedRow } = this.getLastUsedHeaders(sheetId);
+      const zone = { left: 0, top: 0, right: lastUsedCol, bottom: lastUsedRow };
+      return this.splitZoneToPrintPages(sheetId, zone);
+    });
+  }
+
+  changePrintLayout(value: PrintPageLayout) {
+    this.pageLayout = value;
+  }
+
+  changePrintSelection(value: PrintSelection) {
+    this.printSelection = value;
+  }
+
+  changePrintScale(value: PrintScale) {
+    this.printScale = value;
+  }
+
+  setGridLinesVisibility(value: boolean) {
+    this.hideGridLines = value;
+  }
+
+  changePrintOrientation(value: Orientation) {
+    this.orientation = value;
+  }
+
+  private splitZoneToPrintPages(sheetId: UID, printZone: Zone): PagePrintInfo[] {
+    const zoom = this.getPrintZoom(sheetId, printZone);
+
+    const horizontalBreaks: HeaderIndex[] =
+      this.printScale === "fitToWidth"
+        ? [printZone.left, printZone.right + 1]
+        : this.getPageBreaks(sheetId, printZone, "COL", zoom);
+
+    const verticalBreaks: HeaderIndex[] =
+      this.printScale === "fitToHeight"
+        ? [printZone.top, printZone.bottom + 1]
+        : this.getPageBreaks(sheetId, printZone, "ROW", zoom);
+
+    const pages: PagePrintInfo[] = [];
+
+    for (let x = 0; x < horizontalBreaks.length - 1; x++) {
+      for (let y = 0; y < verticalBreaks.length - 1; y++) {
+        const zone = {
+          left: horizontalBreaks[x],
+          top: verticalBreaks[y],
+          right: horizontalBreaks[x + 1] - 1,
+          bottom: verticalBreaks[y + 1] - 1,
+        };
+        pages.push({
+          sheetId,
+          zone,
+          renderingCtx: this.getZoneRenderingContext(sheetId, zone, zoom),
+        });
+      }
+    }
+    return pages.filter((page) => this.pageHasContent(page));
+  }
+
+  private getPageBreaks(
+    sheetId: UID,
+    printZone: Zone,
+    dimension: Dimension,
+    zoom: number
+  ): number[] {
+    const { width: printWidth, height: printHeight } = this.pageDimensionWithMargins;
+
+    const start = dimension === "COL" ? printZone.left : printZone.top;
+    const end = dimension === "COL" ? printZone.right : printZone.bottom;
+    const getHeaderSize =
+      dimension === "COL" ? this.getters.getColDimensions : this.getters.getRowDimensions;
+    const max = dimension === "COL" ? printWidth : printHeight;
+
+    const breaks: number[] = [];
+    let current = 0;
+    for (let i = start; i <= end; i++) {
+      const headerSize = getHeaderSize(sheetId, i).size * zoom;
+      current += headerSize;
+      if (current >= max) {
+        breaks.push(i);
+        current = headerSize;
+      }
+    }
+    return [start, ...breaks, end + 1];
+  }
+
+  private getPrintZoom(sheetId: UID, printZone: Zone) {
+    if (this.printScale === "fitToWidth") {
+      const startX = this.getters.getColDimensions(sheetId, printZone.left).start;
+      const endX = this.getters.getColDimensions(sheetId, printZone.right).end;
+      const width = endX - startX;
+      return Math.min(this.pageDimensionWithMargins.width / width, 1);
+    } else if (this.printScale === "fitToHeight") {
+      const startY = this.getters.getRowDimensions(sheetId, printZone.top).start;
+      const endY = this.getters.getRowDimensions(sheetId, printZone.bottom).end;
+      const height = endY - startY;
+      return Math.min(this.pageDimensionWithMargins.height / height, 1);
+    }
+    return 1;
+  }
+
+  private getZoneRenderingContext(
+    sheetId: UID,
+    zone: Zone,
+    zoom: number
+  ): Omit<GridRenderingContext, "ctx" | "thinLineWidth"> {
+    const firstRowStart = this.getters.getRowDimensions(sheetId, zone.top).start;
+    const lastRowEnd = this.getters.getRowDimensions(sheetId, zone.bottom).end;
+    const firstColStart = this.getters.getColDimensions(sheetId, zone.left).start;
+    const lastColEnd = this.getters.getColDimensions(sheetId, zone.right).end;
+
+    const sheetViewWidth = lastColEnd - firstColStart;
+    const sheetViewHeight = lastRowEnd - firstRowStart;
+    const viewports = new ViewportCollection(
+      this.getters,
+      { [sheetId]: { xSplit: 0, ySplit: 0 } },
+      sheetViewWidth,
+      sheetViewHeight,
+      zoom
+    );
+    viewports.setSheetViewOffset(sheetId, firstColStart, firstRowStart);
+
+    return {
+      sheetId,
+      viewports,
+      hideGridLines: this.hideGridLines,
+      dpr: 3, // Increase the canvas' dpr on print so the image is of better quality
+      selectedZones: [],
+      activeCols: new Set(),
+      activeRows: new Set(),
+      activePosition: undefined,
+    };
+  }
+
+  get pageDimensionWithMargins(): DOMDimension {
+    const { width, height } = this.pageDimensionsInPixels;
+    return { width: width - this.printMargin * 2, height: height - this.printMargin * 2 };
+  }
+
+  get printMargin() {
+    return 50; // in pixels
+  }
+
+  private getLastUsedHeaders(sheetId: UID): { lastUsedCol: number; lastUsedRow: number } {
+    let lastUsedCol = 0;
+    let lastUsedRow = 0;
+
+    const coreCells = this.getters.getCells(sheetId);
+    for (const cell of coreCells) {
+      const position = this.getters.getCellPosition(cell.id);
+      const spreadZone = this.getters.getSpreadZone(position);
+      lastUsedCol = Math.max(lastUsedCol, spreadZone?.right ?? position.col);
+      lastUsedRow = Math.max(lastUsedRow, spreadZone?.bottom ?? position.row);
+    }
+
+    const tables = this.getters.getTables(sheetId);
+    for (const table of tables) {
+      lastUsedCol = Math.max(lastUsedCol, table.range.zone.right);
+      lastUsedRow = Math.max(lastUsedRow, table.range.zone.bottom);
+    }
+
+    for (const figure of this.getters.getFigures(sheetId)) {
+      const figureUI = this.getters.getFigureUI(sheetId, figure);
+      const figureZone = this.getZoneContainingFigure(sheetId, figureUI);
+      lastUsedCol = Math.max(lastUsedCol, figureZone.right);
+      lastUsedRow = Math.max(lastUsedRow, figureZone.bottom);
+    }
+
+    return { lastUsedCol, lastUsedRow };
+  }
+
+  private pageHasContent({ sheetId, zone }: PagePrintInfo): boolean {
+    if (
+      this.getters.getTablesOverlappingZones(sheetId, [zone]).length > 0 ||
+      this.areThereFiguresInZone(sheetId, zone)
+    ) {
+      return true;
+    }
+
+    for (const position of cellPositions(sheetId, zone)) {
+      const coreCell = this.getters.getCell(position);
+      if (coreCell && (coreCell.style || coreCell.isFormula || coreCell.content)) {
+        return true;
+      }
+      const evaluatedCell = this.getters.getEvaluatedCell(position);
+      if (evaluatedCell && evaluatedCell.type !== CellValueType.empty) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  get pageDimensionsInPixels(): DOMDimension {
+    const layout = PRINT_PAGES_LAYOUTS[this.pageLayout];
+    const widthInPixels = Math.floor(layout.width * PX_PER_MM);
+    const heightInPixels = Math.floor(layout.height * PX_PER_MM);
+    return this.orientation === "portrait"
+      ? { width: widthInPixels, height: heightInPixels }
+      : { width: heightInPixels, height: widthInPixels };
+  }
+
+  private areThereFiguresInZone(sheetId: UID, zone: Zone): boolean {
+    const figures = this.getters.getFigures(sheetId);
+    return figures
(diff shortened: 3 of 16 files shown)


tests/grid/__snapshots__/dashboard_grid_component.test.ts.snap
```diff
@@ -33,9 +33,9 @@ exports[`Grid component in dashboard mode simple dashboard rendering snapshot 1`
     </div>
     
     <canvas
-      height="985"
-      style="width:985px;height:985px;zoom:1;"
-      width="985"
+      height="1011"
+      style="width:1033px; height:1011px; "
+      width="1033"
```
hokolomopo, line 38: This snapshot was wrong before, we didn't wait for the resize observers. 

Before: 
Initial state: gridOffsetX = HEADER_WIDTH
1) First render. We update the mode to dashboard. We use `getSheetViewDimension` to get the canvas size, which is at the time wrong (it's equal to `spreadsheetSize - HEADER_WIDTH`). **We did the snapshot here**
2) The resize observer is triggered, changing `gridOffsetX` to 0 and triggering a new render. The canvas' size is now correct.

After:
We now always use `getSheetViewDimensionWithHeaders` for the canvas size.
Initial state: gridOffsetX = HEADER_WIDTH
1) First render. We update the mode to dashboard. We use `getSheetViewDimensionWithHeaders` to get the canvas size, which is the correct size. The canvas size if correct at that time, but since `gridOffsetX = HEADER_WIDTH` at this point we probably draw wrong things on a canvas with the correct size.
2) The resize observer is triggered, changing `gridOffsetX` to 0 and triggering a new render. The canvas' size & content are now correct.

src/components/helpers/draw_grid_hook.ts
```diff
@@ -1,36 +1,58 @@
...
+      ...(partialRenderingCtx?.() || {}),
     };
-    let { width, height } = canvasSize();
-    width = zoom * width;
-    height = zoom * height;
-    canvas.style.width = `${width}px`;
-    canvas.style.height = `${height}px`;
-    canvas.width = width * dpr;
-    canvas.height = height * dpr;
-    canvas.setAttribute("style", `width:${width}px;height:${height}px;zoom:${1 / zoom};`);
```
hokolomopo, line 33: Before if I understand the code correctly, we were increasing the size of the canvas based on the zoom, then using the `zoom` css property to scale the size down. We can just not change the size of the canvas in the first place, then the zoom is not needed ?

src/components/spreadsheet_print/spreadsheet_print.ts
```diff
@@ -0,0 +1,374 @@
...
+  });
+
+  setup() {
+    useExternalListener(window, "beforeprint", () => {
+      const style = document.createElement("style");
+      style.id = "print-style";
+      const size = `${this.state.pageLayout} ${this.state.orientation}`;
+      style.innerHTML = `@media print { @page { size: ${size}; margin: ${PRINT_MARGIN}px;}}`;
+      document.head.appendChild(style);
+    });
```
hokolomopo, line 38: 🙃

tests/test_helpers/renderer_helpers.ts
```diff
@@ -49,6 +58,37 @@ export class MockGridRenderingContext implements GridRenderingContext {
...
+
+  get viewports() {
+    const viewports = this.model.getters.getViewportCollection();
+    // FIXME: normally, the condition to determine if we want to draw the headers or not should be if gridOffsetX/Y are
+    // equal to 0 or not, (ie. if there is space above/left of the grid to draw them, which in practice is equivalent to `isDashboard()`)
+    // But in the renderer tests, we are testing with `isDashboard() === false` but `gridOffsetX/Y === 0`, which is
+    // not supposed to happen. Before this commit we were drawing the headers over the grid in the test, now without this hack
+    // we are not drawing the headers at all since `gridOffsetX/Y === 0`. We should fix the tests.
+    viewports.shouldDisplayHeaders = () => !this.model.getters.isDashboard();
+    return viewports;
```
hokolomopo: https://github.com/odoo/o-spreadsheet/pull/7943

packages/o-spreadsheet-engine/src/helpers/figures/charts/chart_ui_common.ts
```diff
@@ -188,3 +111,56 @@ async function canvasToObjectUrl(canvas: OffscreenCanvas): Promise<string | unde
...
+      config as ChartConfiguration
+    );
+    return () => {
+      chart.destroy();
+      if (!extensionsLoaded) {
+        unregisterChartJsExtensions();
+      }
+    };
+  }
+  // TODO: make a registry of chart types to their rendering functions
```
anhe-odoo, line 137: TODO. Do you want to adress it in this PR and forgot ?

packages/o-spreadsheet-engine/src/helpers/figures/charts/chart_ui_common.ts
```diff
@@ -188,3 +111,56 @@ async function canvasToObjectUrl(canvas: OffscreenCanvas): Promise<string | unde
...
+    config.plugins = [backgroundColorChartJSPlugin];
+    if (!globalThis.Chart.registry.controllers.get(config.type)) {
+      if (!extensionsLoaded) {
+        unregisterChartJsExtensions();
+      }
+      throw new Error(`Chart of type "${config.type}" is not registered in Chart.js library.`);
+    }
+
+    const chart = new globalThis.Chart(
+      canvas as unknown as HTMLCanvasElement,
```
anhe-odoo: `as unknown` ? We cannot force it to `HTMLCanvasElement` directly ? (I tried to remove it as everything seems OK :) )

packages/o-spreadsheet-engine/src/helpers/internal_viewport.ts
```diff
@@ -188,22 +188,22 @@ export class InternalViewport {
     const scrollDeltaY = this.snapCorrection.y;
     if (targetZone) {
       const x =
-        this.getters.getColRowOffset("COL", this.left, targetZone.left) +
+        this.getters.getColRowOffset("COL", this.left, targetZone.left, this.sheetId) +
```
anhe-odoo, line 191: Nitpick, but maybe these changes could have been done in a separate commit ? :) Feel free to ignore, I'm not sure about it

packages/o-spreadsheet-engine/src/types/getters.ts
```diff
@@ -37,7 +37,11 @@ import { SheetViewPlugin } from "../plugins/ui_stateful/sheetview";
 // Getters
 // -----------------------------------------------------------------------------
 
-export type Getters = {
+/**
+ * The getters that can be used in the rendering-related stores and helpers. The SheetView and Selection getters should
+ * not be used in those, they should use the values in the GridRenderingContext instead.
+ */
+export type RenderingGetters = {
```
anhe-odoo, line 44: Isn't there any other plugin's getters that should be out of this ? When I read the name "RenderingGetters", I feel like it's for all the rendering-related plugins, but there is the Clipboard and the LockSheet Plugins for example, which, at first, don't seem related to the rendering (even if the clipboard can be based on rendered things I suppose)

src/components/helpers/draw_grid_hook.ts
```diff
@@ -1,36 +1,58 @@
...
     const ctx = canvas.getContext("2d", { alpha: false })!;
-    const zoom = Math.max(model.getters.getViewportZoomLevel(), 1);
-    const thinLineWidth = 0.4 * dpr;
-    const renderingContext = {
+    dpr = dpr || window.devicePixelRatio || 1;
+    const renderingContext: GridRenderingContext = {
       ctx,
       dpr,
-      thinLineWidth,
+      thinLineWidth: 0.4,
```
anhe-odoo, line 32: Why not 0.4 * dpr ?

src/components/helpers/draw_grid_hook.ts
anhe-odoo: @fdamhaut : What do you think ?

src/components/spreadsheet_print/spreadsheet_print.ts
```diff
@@ -0,0 +1,374 @@
...
+import { Section } from "../side_panel/components/section/section";
+import { StandaloneGridCanvas } from "../standalone_grid_canvas/standalone_grid_canvas";
+
+interface Props {
+  onExitPrintMode: () => void;
+}
+
+// Page layouts in millimeters
+const PRINT_PAGES_LAYOUTS = {
+  A3: { width: 297, height: 420, name: _t("A3 (297 x 420 mm)") },
```
anhe-odoo: Nitpick: Not sure we need the dimensions in the `name`, as other printers only display A3, A4, ... . Moreover, when setting the orientation to landscape, these are "kind of" wrong.

src/components/spreadsheet_print/spreadsheet_print.ts
```diff
@@ -0,0 +1,374 @@
...
+  get pageDimensionWithMargins(): DOMDimension {
+    const { width, height } = this.pageDimensionsInPixels;
+    return { width: width - PRINT_MARGIN * 2, height: height - PRINT_MARGIN * 2 };
+  }
+
+  private getLastUsedHeaders(sheetId: UID): { lastUsedCol: number; lastUsedRow: number } {
+    let lastUsedCol = 0;
+    let lastUsedRow = 0;
+
+    const evaluatedCells = this.env.model.getters.getEvaluatedCellsPositions(sheetId);
```
anhe-odoo: Is this loop only for the spread values (as their should be a coreCells corresponding to any other cell) ? If so, isn't there any other way to check for the spread values than looping on all the evaluatedCells ?

src/components/spreadsheet_print/spreadsheet_print.ts
```diff
@@ -0,0 +1,374 @@
...
+    ) {
+      return true;
+    }
+
+    for (const position of cellPositions(sheetId, zone)) {
+      const coreCell = this.env.model.getters.getCell(position);
+      if (coreCell && (coreCell.style || coreCell.content)) {
+        return true;
+      }
+      const evaluatedCell = this.env.model.getters.getEvaluatedCell(position);
```
anhe-odoo: Same question here : Can we have an evaluatedCell (other than spread value) and no corresponding coreCell ?

src/components/spreadsheet_print/spreadsheet_print.ts
```diff
@@ -0,0 +1,374 @@
...
+      if (evaluatedCell && evaluatedCell.type !== CellValueType.empty) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  get pageDimensionsInPixels(): DOMDimension {
+    const layout = PRINT_PAGES_LAYOUTS[this.state.pageLayout];
+    const widthInPixels = Math.floor(layout.width * 3.78);
```
anhe-odoo: Nitpick : Maybe the 3.78 can be a constant with an obvious name. Here it looks like a magic number

src/stores/grid_renderer_store.ts
```diff
@@ -119,18 +135,21 @@ export class GridRenderer extends SpreadsheetStore {
         this.preventNewAnimationsInNextFrame = true;
         break;
       case "UPDATE_CELL":
-        if (cmd.sheetId === this.getters.getActiveSheetId()) {
-          const zones = this.getters.getCommandZones(cmd);
-          this.zonesWithPreventedAnimationsInNextFrame.push(...zones);
+        const zones = this.getters.getCommandZones(cmd);
+        if (!this.zonesWithPreventedAnimationsInNextFrame[cmd.sheetId]) {
```
anhe-odoo, line 140: When can we have multiple entries in this object ?

packages/o-spreadsheet-engine/src/helpers/figures/charts/chart_ui_common.ts
hokolomopo: Nah this was a TODO that was there before, I just copied it during refactoring

packages/o-spreadsheet-engine/src/helpers/figures/charts/chart_ui_common.ts
hokolomopo: Same, that's just something that I copy/paste, but the unknown doesn't seem to be needed indeed

packages/o-spreadsheet-engine/src/types/getters.ts
hokolomopo: Yeah the name isn't that great, I'm open to propositions .... The goal isn't to group all the rendering-related getters, but all the getters that _can_ be accessed in the rendering process.

src/components/helpers/draw_grid_hook.ts
hokolomopo: coz it's really ugly on high DPR

src/components/spreadsheet_print/spreadsheet_print.ts
hokolomopo: We can do that in the same loop as the core cells, you're right :)

src/components/spreadsheet_print/spreadsheet_print.ts
hokolomopo: Yes if the zone only take the bottom of an spread result w/o the root cell.

src/components/spreadsheet_print/spreadsheet_print.ts
hokolomopo: Francois like it :shrug:

src/stores/grid_renderer_store.ts
hokolomopo: If there is a bunch of update cells on different sheet in a revision (I don't think it happens in practice, but it theoretically could)

packages/o-spreadsheet-engine/src/helpers/viewport_collection.ts
```diff
@@ -0,0 +1,778 @@
...
+    });
+  }
+
+  getViewportZoomLevel(): number {
+    return this.zoomLevel;
+  }
+
+  // ---------------------------------------------------------------------------
+  //
+  // ---------------------------------------------------------------------------
```
LucasLefevre: nice section title :)

packages/o-spreadsheet-engine/src/plugins/ui_stateful/sheetview.ts
```diff
@@ -142,15 +81,17 @@ export class SheetViewPlugin extends UIPlugin {
...
   allowDispatch(cmd: LocalCommand): CommandResult | CommandResult[] {
     switch (cmd.type) {
-      case "SET_VIEWPORT_OFFSET":
+      case "SET_VIEWPORT_OFFSET": {
+        const sheetId = this.getters.getActiveSheetId();
         return this.chainValidations(
-          this.checkScrollingDirection,
-          this.checkIfViewportsWillChange
+          () => this.viewports.checkScrollingDirection.call(this.viewports, sheetId, cmd),
+          () => this.viewports.checkIfViewportsWillChange.call(this.viewports, sheetId, cmd)
```
LucasLefevre:
```suggestion
          () => this.viewports.checkScrollingDirection(sheetId, cmd),
          () => this.viewports.checkIfViewportsWillChange(sheetId, cmd)
```

packages/o-spreadsheet-engine/src/plugins/ui_stateful/sheetview.ts
```diff
@@ -142,15 +81,17 @@ export class SheetViewPlugin extends UIPlugin {
...
         )(cmd);
+      }
       case "RESIZE_SHEETVIEW":
         return this.chainValidations(
-          this.checkValuesAreDifferent,
-          this.checkPositiveDimension
+          () => this.viewports.checkValuesAreDifferent.call(this.viewports, cmd),
+          () => this.viewports.checkPositiveDimension.call(this.viewports, cmd)
```
LucasLefevre: same here

src/components/spreadsheet_print/spreadsheet_print.ts
```diff
@@ -0,0 +1,381 @@
...
+    orientation: "portrait",
+    hideGridLines:
+      this.env.model.getters.isDashboard() ||
+      !this.env.model.getters.getGridLinesVisibility(this.env.model.getters.getActiveSheetId()),
+  });
+
+  setup() {
+    useExternalListener(window, "beforeprint", () => {
+      const style = document.createElement("style");
+      style.id = "print-style";
```
LucasLefevre: just a little less conflict-prone
```suggestion
      style.id = "o-spreadsheet-print-style";
```

src/components/spreadsheet_print/spreadsheet_print.ts
```diff
@@ -0,0 +1,381 @@
...
+      const size = `${this.state.pageLayout} ${this.state.orientation}`;
+      style.textContent = `@media print { @page { size: ${size}; margin: ${PRINT_MARGIN}px;}}`;
+      document.head.appendChild(style);
+    });
+    useExternalListener(window, "afterprint", () => {
+      const style = document.getElementById("print-style");
```
LucasLefevre:
```suggestion
      const style = document.getElementById("o-spreadsheet-print-style");
```

src/components/spreadsheet_print/spreadsheet_print.ts
```diff
@@ -0,0 +1,381 @@
...
+      if (style) {
+        document.head.removeChild(style);
+      }
+    });
+  }
+
+  get printPages(): PagePrintInfo[] {
```
LucasLefevre: AFAICT, this could be extracted out of the component (to a store, or whatever), with the state. It would be more easily testable without the entire UI.
And since it looks like it's most of the business logic here, it's probably worth it to be able to easily write tests, which are fast and easily debuggable.

src/stores/grid_renderer_store.ts
```diff
@@ -69,22 +72,35 @@ interface Animation {
   animationTypes: string[];
 }
 
-export class GridRenderer extends SpreadsheetStore {
+export class GridRenderer extends DisposableStore {
```
LucasLefevre, line 76: why cannot it be a `SpreadsheetStore` anymore ?

src/stores/grid_renderer_store.ts
```diff
@@ -876,37 +921,37 @@ export class GridRenderer extends SpreadsheetStore {
     return box;
   }
 
-  private getGridBoxes(zone: Zone): Box[] {
+  private getGridBoxes(ctx: GridRenderingContext, zone: Zone): Box[] {
```
LucasLefevre: The idea behind boxes is to clearly decouple *what* to render (boxes) from *how* to render (on the canvas, but theoretically, it could be rendered in other ways).

To keep this, `getGridBoxes` should not be coupled with the canvas in any way and hence not have access to the canvas rendering context.
```suggestion
  private getGridBoxes(viewports: ViewportCollection, sheetId: UID, zone: Zone): Box[] {
```

tests/test_helpers/renderer_helpers.ts
```diff
@@ -49,6 +58,42 @@ export class MockGridRenderingContext implements GridRenderingContext {
     };
     this.ctx = new Proxy({}, handler);
   }
+
+  get getters(): RenderingGetters {
+    return this.model.getters;
+  }
```
LucasLefevre: could be assigned in the constructor, instead of adding one layer with a function call

tests/spreadsheet/spreadsheet_print_component.test.ts
```diff
@@ -0,0 +1,338 @@
...
+  });
+
+  test("Can print the entire workbook", async () => {
+    setCellContent(model, "A1", "Hello");
+    createSheet(model, { name: "Sheet2", sheetId: "Sheet2" });
+    setCellContent(model, "A1", "World", "Sheet2");
+    await mountSpreadsheetPrint();
+    expect(".o-print-page").toHaveCount(1);
+    expect(getLastZonesRendered()).toEqual([
+      { sheetId: "Sheet1", left: 0, top: 0, right: 0, bottom: 0 },
```
LucasLefevre: For readability:
if you change the return type of `getLastZonesRendered` to `{ sheetId, zone }`
```suggestion
      toBoundedRange("Sheet1", "A1")
```
or
```suggestion
      { sheetId: "Sheet1", ...toZone("A1") },
```

src/components/spreadsheet_print/spreadsheet_print.ts
```diff
@@ -0,0 +1,381 @@
...
+  }
+
+  changeOrientation(value: Orientation) {
+    this.state.orientation = value;
+  }
+
+  private splitZoneToPrintPages(sheetId: UID, printZone: Zone): PagePrintInfo[] {
+    const zoom = this.getPrintZoom(sheetId, printZone);
+
+    const xBreaks: number[] =
```
LucasLefevre: nitpick:
the "x" in `xBreaks` made me initially think it was pixels, but it's actually header indices.
maybe `horizontalBreaks` ?

At least it could be typed more precisely
```suggestion
    const xBreaks: HeaderIndex[] =
```

src/stores/grid_renderer_store.ts
hokolomopo: Spreadsheet stores are registering themselves in the global `RendererStore` in their constructor. Here I want to register to the `RendererStore` given as argument

tests/spreadsheet/spreadsheet_print_component.test.ts
hokolomopo: Maybe it's just me, but my first version had XCs and I found it less readable. I have an easier time reading "so some zones stop at 6, and th eother start at 7, make sense" than trying to parse letters in my head to check if the next one is correct and trying to check the the two parts of the xcs 

image

LucasLefevre: fair enough, for columns it's indeed probably easier

packages/o-spreadsheet-engine/src/helpers/viewport_collection.ts
```diff
@@ -0,0 +1,774 @@
...
+ *  and horizontally between rows 4 and 5.
+ *
+ *  One can see that we scrolled horizontally from column E to G and
+ *  vertically from row 5 to 7.
+ *
+ *     A  B  C  D   G  H  I  J  K  L  M  N  O  P  Q  R  S  T
+ *     _______________________________________________________
+ *  1 |           :                                           |
+ *  2 |           :                                           |
+ *  3 |           :        B   ↑                 6            |
```
pro-odoo, line 52: What does "B", "6" means ?

pro-odoo: Ok I see now, maybe add a little comment: "This schema will be used in comments below" or something like this

packages/o-spreadsheet-engine/src/helpers/viewport_collection.ts
```diff
@@ -0,0 +1,774 @@
...
+    return {
+      scrollX: viewport.offsetX,
+      scrollY: viewport.offsetY,
+    };
+  }
+
+  getSheetViewVisibleCols(sheetId: UID): HeaderIndex[] {
+    const viewports = this.getSubViewports(sheetId);
+
+    //TODO ake another commit to eimprove this
```
pro-odoo: Typos

packages/o-spreadsheet-engine/src/helpers/viewport_collection.ts
```diff
@@ -0,0 +1,774 @@
...
+  getters: RenderingGetters;
+  viewports: Record<UID, SheetViewports | undefined> = {};
+
+  /**
+   * The viewport dimensions are usually set by one of the components
+   * (i.e. when grid component is mounted) to properly reflect its state in the DOM.
+   * In the absence of a component (standalone model), is it mandatory to set reasonable default values
+   * to ensure the correct operation of this plugin.
+   */
+  sheetViewWidth: Pixel = getDefaultSheetViewSize();
```
pro-odoo: Can we keep these attributes private ? To not expose/mutate internal state directly

packages/o-spreadsheet-engine/src/helpers/viewport_collection.ts
```diff
@@ -0,0 +1,774 @@
...
+      height: this.sheetViewHeight,
+    };
+  }
+
+  getGridOffset(): DOMCoordinates {
+    return { x: this.gridOffsetX, y: this.gridOffsetY };
+  }
+
+  /** type as pane, not viewport but basically pane extends viewport */
+  getSheetMainViewport(sheetId: UID): Viewport {
```
pro-odoo: private

packages/o-spreadsheet-engine/src/helpers/viewport_collection.ts
```diff
@@ -0,0 +1,774 @@
...
+      width: this.sheetViewWidth,
+      height: this.sheetViewHeight,
+    };
+  }
+
+  getGridOffset(): DOMCoordinates {
+    return { x: this.gridOffsetX, y: this.gridOffsetY };
+  }
+
+  /** type as pane, not viewport but basically pane extends viewport */
```
pro-odoo: ?

src/components/spreadsheet_print/spreadsheet_print.ts
```diff
@@ -0,0 +1,83 @@
...
+  static template = "o-spreadsheet-SpreadsheetPrint";
+  static props = { onExitPrintMode: Function };
+  static components = { StandaloneGridCanvas, Section, Select, BadgeSelection, Checkbox };
+
+  printStore!: Store<SpreadsheetPrintStore>;
+
+  setup() {
+    this.printStore = useLocalStore(SpreadsheetPrintStore);
+    useExternalListener(window, "beforeprint", () => {
+      const style = document.createElement("style");
```
rrahir: nitpick, store the node in the component state, no need to query it then.

src/components/spreadsheet_print/spreadsheet_print.ts
```diff
@@ -0,0 +1,83 @@
...
+      style.id = "o-spreadsheet-print-style";
+      const size = `${this.printStore.pageLayout} ${this.printStore.orientation}`;
+      style.textContent = `@media print { @page { size: ${size}; margin: ${this.printStore.printMargin}px;}}`;
+      document.head.appendChild(style);
+    });
+    useExternalListener(window, "afterprint", () => {
+      const style = document.getElementById("o-spreadsheet-print-style");
```
rrahir: if we somehow destroy the component, we should also clear the style imho

src/components/spreadsheet_print/spreadsheet_print_store.ts
```diff
@@ -0,0 +1,331 @@
...
+export class SpreadsheetPrintStore extends SpreadsheetStore {
+  mutators = [
+    "changePrintLayout",
+    "changePrintSelection",
+    "changePrintScale",
+    "changePrintOrientation",
+    "setGridLinesVisibility",
+  ] as const;
+
+  pageLayout = "A4";
```
rrahir, line 49: Not a big deal right now (+ rmight require a migration so bof bof) but the paperformat depends per user country (typically in Odoo we have a specific model tied to the company_id)
```python
paperformat_id = fields.Many2one('report.paperformat', 'Paper format', default=lambda self: self.env.ref('base.paperformat_euro', raise_if_not_found=False))
...
format = fields.Selection([(ps['key'], ps['description']) for ps in PAPER_SIZES], 'Paper size', default='A4', help="Select Proper Paper size")
```
and see larger paper formats proposed in https://github.com/odoo/odoo/blob/19.0/odoo/addons/base/models/report_paperformat.py

I think we should add a usser-dependant default and a possibility  to extend that list (like for the currencies) but maybe not in this PR directly so we can go forward

src/components/spreadsheet_print/spreadsheet_print_store.ts
```diff
@@ -0,0 +1,331 @@
...
+    };
+  }
+
+  get pageDimensionWithMargins(): DOMDimension {
+    const { width, height } = this.pageDimensionsInPixels;
+    return { width: width - this.printMargin * 2, height: height - this.printMargin * 2 };
+  }
+
+  get printMargin() {
+    return 50; // in pixels
```
rrahir, line 207: ? should probably change from 1 format to the other?

src/components/spreadsheet_print/spreadsheet_print.xml
```diff
@@ -0,0 +1,96 @@
...
+            <Section>
+              <div class="o-section-title">Orientation</div>
+              <BadgeSelection
+                choices="printStore.orientationChoices"
+                onChange.bind="this.changeOrientation"
+                selectedValue="printStore.orientation"
+              />
+            </Section>
+
+            <Section t-if="!this.env.model.getters.isDashboard()">
```
rrahir:
```suggestion
            <Section t-if="!this.env.isDashboard()">
```

src/components/standalone_grid_canvas/figure_renderer_store.ts
```diff
@@ -0,0 +1,151 @@
...
+  GRAY_400,
+} from "@odoo/o-spreadsheet-engine/constants";
+import { computeTextFont } from "../../helpers";
+import { drawChartOnCanvas } from "../../helpers/figures/charts";
+import { DisposableStore, Get, Store } from "../../store_engine";
+import { ModelStore } from "../../stores";
+import { RendererStore } from "../../stores/renderer_store";
+import { FigureUI, RenderingGetters } from "../../types";
+
+export interface HighlightProvider {
```
rrahir: can be removed

src/components/spreadsheet_print/spreadsheet_print_store.ts
```diff
@@ -0,0 +1,331 @@
...
+    const firstColStart = this.getters.getColDimensions(sheetId, zone.left).start;
+    const lastColEnd = this.getters.getColDimensions(sheetId, zone.right).end;
+
+    const sheetViewWidth = lastColEnd - firstColStart;
+    const sheetViewHeight = lastRowEnd - firstRowStart;
+    const viewports = new ViewportCollection(this.getters, sheetViewWidth, sheetViewHeight, zoom);
+    viewports.setSheetViewOffset(sheetId, firstColStart, firstRowStart);
+
+    return {
+      selectedZones: [],
```
rrahir: you should override the full selectionState, this only works because you do not render the headers. It feels a bit brouillon, I think it'd be clearer if we had to provide the full rendering context to `useGridDrawing` to avoid "magic" behaviours

packages/o-spreadsheet-engine/src/helpers/figures/charts/chart_ui_common.ts
```diff
@@ -37,70 +38,79 @@ export async function chartToImageUrl(
...
+  type: ChartType
+): Promise<Blob | null> {
+  try {
+    const canvas = createRenderingSurface(figure.width, figure.height);
+    const cleanup = drawChartOnCanvas(canvas, runtime, figure, type);
+    const chartBlob = await canvasToBlob(canvas);
+    cleanup();
+    return chartBlob;
+  } catch (error) {
+    console.log("Error exporting chart to image file: " + error.message);
```
rrahir, line 65: what can happen?

packages/o-spreadsheet-engine/src/helpers/viewport_collection.ts
```diff
@@ -74,19 +74,15 @@ export class ViewportCollection {
...
-  private sheetViewWidth: Pixel = getDefaultSheetViewSize();
-  private sheetViewHeight: Pixel = getDefaultSheetViewSize();
   private gridOffsetX: Pixel = 0;
   private gridOffsetY: Pixel = 0;
-  private zoomLevel: number = 1;
 
-  constructor(getters: RenderingGetters) {
+  constructor(
+    getters: RenderingGetters,
+    private sheetViewWidth: Pixel = getDefaultSheetViewSize(),
```
rrahir, line 84: is it still realted to SheetView though ?

src/components/helpers/draw_grid_hook.ts
```diff
@@ -1,36 +1,58 @@
...
+      viewports: model.getters.getViewportCollection(),
+      ...model.getters.getSelectionState(),
+      ...(partialRenderingCtx?.() || {}),
```
rrahir: we should give a full GridRenderingContext as a parameter, the partial part is not clear imho.

src/components/standalone_grid_canvas/standalone_grid_canvas.ts
```diff
@@ -0,0 +1,68 @@
...
+    drawFiguresOnCanvas: { type: Boolean, optional: true },
+  };
+
+  canvasRef = useRef("canvas");
+
+  rendererStore!: Store<RendererStore>;
+  figureRendererStore!: Store<FigureRendererStore>;
+
+  setup() {
+    this.rendererStore = useLocalStore(RendererStore, ["Background", "Chart", "Headers"]);
```
rrahir: looks louche, you set selectedZones to an empty list but you still end up skipping the layer so it has 0 impact.
 Another point related to that situation;
we have a new typing that removed the sheetview and selection getters from the store but while we draw in gridOverlay, we still call  `GridSelectionPlugin.drawGrid` which has access to all the getters (and it's internal state) so the access limitation to the getters is bypassed.

packages/o-spreadsheet-engine/src/plugins/ui_stateful/selection.ts
```diff
@@ -48,6 +48,13 @@ interface SheetInfo {
   gridSelection: Selection;
 }
 
+export interface SelectionState {
+  sheetId: UID;
+  selectedZones: Zone[];
```
rrahir, line 53: seeing my previous comme,nt, I don't think selectedZones belong here anyway

src/components/standalone_grid_canvas/standalone_grid_canvas.ts
rrahir: Why not simply skipping the Header layer ?

src/components/spreadsheet_print/spreadsheet_print_store.ts
```diff
@@ -0,0 +1,331 @@
...
+  printScale = "fitToWidth";
+  orientation = "portrait";
+  hideGridLines =
+    this.getters.isDashboard() ||
+    !this.getters.getGridLinesVisibility(this.getters.getActiveSheetId());
+
+  get printPages(): PagePrintInfo[] {
+    if (this.printSelection === "selection") {
+      const sheetId = this.getters.getActiveSheetId();
+      const selectedZones = this.getters.getSelectedZones();
```
rrahir: the print is only active with a single zone
```suggestion
const selectedZones = this.getters.getSelectedZone();
```

src/components/spreadsheet_print/spreadsheet_print_store.ts
rrahir: is actually to repvent the printer to print on the very top of the page, avoid clipped content

src/components/standalone_grid_canvas/standalone_grid_canvas.ts
hokolomopo: Good idea, I removed `viewports.shouldDisplayHeaders` and now we just don't draw the header layer for standalone grid/dashboard

hokolomopo: As discussed, I'll change the `drawLayers` of the plugin to use the correct info from the renderingCtx instead of the getters. And add a FIXME to say it's really not solid and we should improve that

packages/o-spreadsheet-engine/src/helpers/figures/charts/chart_ui_common.ts
hokolomopo: I kept master behaviour, we have tests that check that we don't crash and have this console.log. Some were written by you 😅  `cannot export Chart.js chart when lib is not loaded`

packages/o-spreadsheet-engine/src/helpers/viewport_collection.ts
hokolomopo: I think it still make sense, it's not directly related to the `SheetViewPlugin` but it literally represent a view of a sheet. Changing  it to `width` is more confusing IMO, as it's less clear it refers to the size of the view and not the total size of a sheet or whatever.

src/components/spreadsheet_print/spreadsheet_print_store.ts
hokolomopo: As discussed, maybe for the future, but will need a layer of validation as some odoo page formats are not supported by the [CSS size property](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@page/size)

packages/o-spreadsheet-engine/src/plugins/ui_stateful/sheetview.ts
```diff
@@ -182,13 +186,16 @@ export class SheetViewPlugin extends UIPlugin {
...
       case "UNFREEZE_COLUMNS":
       case "FREEZE_COLUMNS":
       case "FREEZE_ROWS":
+        this.sheetsWithDirtyViewports.add(cmd.sheetId);
+        this.viewports.setPaneDivision(cmd.sheetId, this.getters.getPaneDivisions(cmd.sheetId));
+        break;
+      case "REMOVE_TABLE":
+      case "UPDATE_TABLE":
+      case "UPDATE_FILTER":
       case "UNFREEZE_COLUMNS_ROWS":
```
rrahir: this.viewports.setPaneDivision(cmd.sheetId, this.getters.getPaneDivisions(cmd.sheetId)); as well

src/components/spreadsheet_print/spreadsheet_print.xml
```diff
@@ -0,0 +1,95 @@
...
+                selectedValue="printStore.pageLayout"
+                values="printStore.layoutOptions"
+                class="'o-print-layout'"
+                onChange.bind="this.onLayoutChange"
+              />
+            </Section>
+
+            <Section>
+              <div class="o-section-title">Orientation</div>
+              <BadgeSelection
```
rrahir, line 74: badges don't have the same size, i'll open  a task for that

src/components/standalone_grid_canvas/standalone_grid_canvas.ts
```diff
@@ -0,0 +1,61 @@
...
+  };
+
+  canvasRef = useRef("canvas");
+
+  rendererStore!: Store<RendererStore>;
+  figureRendererStore!: Store<FigureRendererStore>;
+
+  setup() {
+    this.rendererStore = useLocalStore(RendererStore, ["Background", "Chart"]);
+    // Note: if we ever want to change the value of props.drawFiguresOnCanvas, we'd need to manually dispose the figureRendererStore
```
rrahir: Maybe add the willUpdateProps that does it or we can't, in which case we could throw an error if someone tries to change the props

hokolomopo: I'll just remove the props for the time being, it's currently useless. But yeah it's a bit annoying/ugly to handle it properly ...

rrahir:
robodoo rebase-ff

rrahir:
robodoo r+

PR #7943, opened 2026-02-12
[FIX] renderer: rendering tests are wrong

hokolomopo:
### [FIX] renderer: rendering tests are wrong

The rendering tests wre drawing a grid for which `gridOffsetX/y` were
equal to 0, but the headers were still drawn. So the headers were drawn
overlapping the cells, which obviously made no sense.

Now unless explicitly changed, the mock rendering context used in the
tests does not draw the headers, and simply draws the grid on a
1000*1000 area.

Task: 5933130
Task: [5933130](https://www.odoo.com/odoo/2328/tasks/5933130)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


tests/composer/standalone_composer_component.test.ts
```diff
@@ -66,6 +66,7 @@ describe("Spreadsheet integrations tests", () => {
   test("Can edit a standalone composer", async () => {
     await openSidePanelWithComposer({ composerContent: "Hello world" });
     expect(composerEl.textContent).toBe("Hello world");
+    composerEl.focus();
```
hokolomopo, line 69: The mocked DOM got "better", changing the selection on an element that doesn't have the focus no longer works

tests/setup/jest_extend.ts
```diff
@@ -60,7 +62,12 @@ function getPrettyEvaluatedCells(model: Model, sheetId: string, zone: Zone) {
   });
 }
 
+const toMatchImageSnapshot = configureToMatchImageSnapshot({
+  dumpDiffToConsole: false, // allows to copy-paste the diff from Runbot logs
+});
```
hokolomopo, line 67: TODO: how to enable this only for runbot ?

hokolomopo:
https://github.com/odoo/o-spreadsheet/pull/8251

PR #7956, opened 2026-02-17
[PERF] evaluation: replace r-tree with interval trees

LucasLefevre:
This commit changes the dependency graph implementation from a global R-tree
to a collection of interval trees.

An interval tree is just a binary search tree but with a special trick
allowing to query over an interval of values [10, 15] (1D segments)
We use one interval tree per column.

See https://en.wikipedia.org/wiki/Interval_tree
and https://en.wikipedia.org/wiki/Interval_tree#Augmented_tree ;)

Some key observations:

- building the entire dependency graph is often
  wasted time as only parts of the spreadsheet are typically updated (e.g.,
  array formula spills or manual updates from the user).
  With this, we only build trees for the specific columns that we need.

- spreadsheets are naturally biased toward columns.
  The previous R-tree often produced long, thin bounding boxes could lead to
  sub-optimal spatial partitioning. One interval tree per column maps
  more naturally to the data shape.

- Insertions are now O(1) via a buffer (instead of O(log n) and doesn't trigger
  any tree rebalancing;
  Trees are built (or rebuilt) only upon the first query for that specific
  column. Caveat: for rebuilds, the tree is rebuilt from (almost) scratch.
  We could just insert new insertions in the existing tree, but it requires
  a re-balancing algorithm. I kept it simple for now. It's very fast.

- balancing an R-tree requires non-trivial heuristics to split bounding boxes
  in an efficient/optimal way.

- Funnily, the current graph  partly relies on `RangeSet`, which is itself
  interval-based. The recursive retrival still relies on `RangeSet`. This
  move lays the groundwork for further optimizations to implement more
  efficient operations.

- trees are individually a lot cheaper to build and query because they are smaller.

- the delete operation is currently not implemented. I didn't find a real-world
  use-case where it really matter. If we were to implement deletion,
  it would require a re-balancing algorithm. I kept it simple for now.

Large formula data set
----------------------

Model created: no difference (the dependency graph isn't built)
  before:	Mean: 2754 ms, StdErr: 11 ms, n=10 (-0%)
  after:	Mean: 2757 ms, StdErr: 15 ms, n=10

The first UPDATE_CELL builds the dependency graph:
  before:	Mean: 2164 ms, StdErr: 12 ms, n=10
  after:	Mean: 142  ms, StdErr: 1  ms, n=10 (-93%)

Next UPDATE_CELL in a column where the interval tree was not built yet is
slower because it builds the tree for that column.
  before:	Mean: 20 ms, StdErr: 1 ms, n=10 (-66%)
  after:	Mean: 61 ms, StdErr: 0 ms, n=10

Second UPDATE_CELL in the same column are faster
  before:	Mean: 16 ms, StdErr: 0.5 ms, n=10
  after:	Mean: 13 ms, StdErr: 0.3 ms, n=10 (-16%)

Split VLOOKUP large dataset
---------------------------

evaluate all cells
  before:       Mean: 518 ms, StdErr: 2 ms, n=10
  after:	Mean: 300 ms, StdErr: 3 ms, n=10 (-42%)

Model created in
  before:	Mean: 1090 ms, StdErr: 3 ms, n=10
  after:	Mean: 865  ms, StdErr: 4 ms, n=10 (-21%)

MRR split (from odoo.com)
---------

Model created in
  before:	Mean=10709 ms, StdErr=180 ms, n=10
  after:	Mean=7201 ms,  StdErr=95  ms, n=10 (-33%)

evaluate all cells
  before:	Mean=16897 ms, StdErr=214 ms, n=10
  after:	Mean=8606  ms, StdErr=123 ms, n=10 (-49%)

Portfolio management
--------------------

EVALUATE_CELLS 1
  before:	Mean=2527.37 ms, StdErr=50.59 ms, n=15
  after:	Mean=2206.89 ms, StdErr=47.78 ms, n=15  (-13%)

disclaimer: AI assisted for the base implementation :)

Task: 5949419


packages/o-spreadsheet-engine/src/plugins/ui_core_views/cell_evaluation/formula_dependency_graph.ts
```diff
@@ -1,69 +1,68 @@
...
-    const ranges = this.dependencies.get(formulaPosition);
-    if (!ranges) {
-      return;
-    }
-    for (const range of ranges) {
-      this.rTree.remove(range);
-    }
-    this.dependencies.delete(formulaPosition);
+    // Not implemented.
+    // For simplicity, we do not have a self-reblancing tree algorithm which is required
```
anhe-odoo: self-rebalancing ?

packages/o-spreadsheet-engine/src/plugins/ui_core_views/cell_evaluation/formula_dependency_graph.ts
```diff
@@ -1,69 +1,68 @@
...
+    // to delete intervals efficiently.
+    // We accumulate stale intervals in the tree, but we didn't not find any real-world scenario
```
anhe-odoo: did not find ?

PR #7957, opened 2026-02-18
[IMP] LinkEditor: change UX with a suggestion/fuzzy search system

rrahir:
## Description:

Task: [5932139](https://www.odoo.com/odoo/2328/tasks/5932139)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/components/link/link_editor/link_editor.css
```diff
@@ -1,6 +1,7 @@
 .o-spreadsheet {
   .o-link-editor {
-    font-size: 13px;
+    /* width: 350px; */
+    width: 330px;
```
hokolomopo: comment

src/components/link/link_editor/link_editor.ts
```diff
@@ -27,19 +35,72 @@ export class LinkEditor extends Component<LinkEditorProps, SpreadsheetChildEnv>
...
+  suggestionListRef = useRef("suggestionList");
+  urlInputContainer = useRef("urlInputContainer");
+
+  private state: LinkState = useState(this.defaultState);
+
+  // private selection: ProposalsState = useState({
+  //   index: -1,
+  //   proposals: {},
+  //   proposalList: [],
+  // });
```
hokolomopo: comment

src/components/link/link_editor/link_editor.ts
```diff
@@ -27,19 +35,72 @@ export class LinkEditor extends Component<LinkEditorProps, SpreadsheetChildEnv>
...
   }
 
-  get defaultState(): State {
+  computeLinks() {
+    this.state.selectedIndex = -1;
+    this.state.linksByCategory = this.linkProposalByCategory;
+    this.state.linksList = Object.values(this.state.linksByCategory).flat();
+  }
+
+  get linkProposalByCategory(): Record<string, LinkProposal> {
```
hokolomopo, line 58: I'm not sure exactly what the fuzzy lookup is looking for. 

Image

Also maybe highlight the matched letters like in the composer if it's not too annoying.

hokolomopo: Edit: oh it's looking for title + item name. Not very intuitive, but maybe it's just me

src/components/link/link_editor/link_editor.xml
```diff
@@ -1,68 +1,84 @@
...
             />
           </t>
-          <button
-            t-if="link.url"
+          <button 
+            t-if="state.url"
             t-on-click="removeLink"
             class="o-remove-url o-button-icon"
             t-translation="off">
             ✖
```
hokolomopo: while you're changing this, change the cross to be an fa icon maybe

src/components/link/link_editor/link_editor.xml
```diff
@@ -1,68 +1,84 @@
...
-        menuItems="menuItems"
-        onMenuClicked="(ev) => this.onSpecialLink(ev)"
-        onClose="() => this.menu.isOpen=false"
-      />
+
+      <div class="suggestions border rounded" t-ref="suggestionList">
+        <t
+          t-foreach="Object.entries(this.state.linksByCategory)"
+          t-as="proposalCategory"
+          t-key="proposalCategory[0]">
```
hokolomopo: I'm pretty sure there are `_key` and `_values` utilities instead of using Object.entries

src/components/link/link_editor/link_editor.xml
```diff
@@ -1,68 +1,84 @@
...
       </div>
-      <MenuPopover
-        t-if="menu.isOpen"
-        anchorRect="menuButtonRect"
-        menuItems="menuItems"
-        onMenuClicked="(ev) => this.onSpecialLink(ev)"
-        onClose="() => this.menu.isOpen=false"
-      />
+
+      <div class="suggestions border rounded" t-ref="suggestionList">
```
hokolomopo: need a placeholder when the list is empty 

Image

src/components/link/link_editor/link_editor.css
```diff
@@ -13,6 +14,21 @@
...
+      --os-icon-edge-length: 15px;
+    }
+    .suggestions {
+      height: 150px;
+      overflow-y: auto;
+    }
+    .suggestion-item {
+      &:hover,
+      &.selected {
+        background-color: var(--os-button-hover-bg);
```
hokolomopo: note for rebase: popover color/hovered item color will change with this https://github.com/odoo/o-spreadsheet/pull/8114/changes#diff-36106881c5f5c9f7cc23d305e0604b854442bf8ba18abe4a05cb85c55ef6e133R18

src/components/link/link_editor/link_editor.ts
rrahir: I don't really know how to make it discoverable tbh but if we end up with lots of categories (currently only 2 though) it might come in handy

src/components/grid_popover/grid_popover.xml
```diff
@@ -4,7 +4,7 @@
       t-if="cellPopover.isOpen"
       positioning="cellPopover.cellCorner"
       maxHeight="cellPopover.Component.size and cellPopover.Component.size.maxHeight"
-      maxWidth="cellPopover.Component.size and cellPopover.Component.size.maxHidth"
```
LucasLefevre, line 7: 🤪

src/components/link/link_editor/link_editor.css
LucasLefevre: Since the other PR is merged, this needs to be updated then ?

src/components/link/link_editor/link_editor.ts
```diff
@@ -27,19 +35,66 @@ export class LinkEditor extends Component<LinkEditorProps, SpreadsheetChildEnv>
...
+  private state: LinkState = useState(this.defaultState);
 
   setup() {
+    this.computeLinks();
     onMounted(() => this.urlInput.el?.focus());
   }
 
-  get defaultState(): State {
+  computeLinks() {
+    this.state.selectedIndex = -1;
```
LucasLefevre: with `number | null`, types force us to handle this case, with `-1` typescript doesn't help us if we forget this special case.
```suggestion
    this.state.selectedIndex = null;
```

src/components/link/link_editor/link_editor.ts
LucasLefevre: Right now, it's pretty confusing I think. I was also wandering what was happening.
I'd remove the fuzzy lookup on title for now and we'll change it later if we add more categories

Image

src/components/link/link_editor/link_editor.ts
```diff
@@ -27,19 +35,66 @@ export class LinkEditor extends Component<LinkEditorProps, SpreadsheetChildEnv>
...
+    this.state.linksByCategory = this.linkProposalByCategory;
+    this.state.linksList = Object.values(this.state.linksByCategory).flat();
```
LucasLefevre, line 55: do we need to store `linksList` ? Even `linksByCategory` looks like it could be derived every time

src/components/link/link_editor/link_editor.ts
```diff
@@ -27,19 +35,66 @@ export class LinkEditor extends Component<LinkEditorProps, SpreadsheetChildEnv>
...
+  }
+
+  get linkProposalByCategory(): Record<string, LinkProposal> {
+    const proposals = {};
+    let counter = -1;
+    for (const category of urlRegistry.getKeys()) {
+      const spec = urlRegistry.get(category);
+      const inputVal = (this.urlInput?.el as HTMLInputElement)?.value || "";
```
LucasLefevre: nit: could be moved outside the loop

src/components/grid_popover/grid_popover.xml
rrahir: very used property up until now!

src/components/link/link_editor/link_editor.ts
rrahir: I didn't like the idea of a floating "selectedIndex" that might not be aligned with the actual number of values that are displayed and go through all the proposals all the time seemed a bit useless  but I can change this for a derived value.

LucasLefevre: > a floating "selectedIndex"

that's true. It doesn't matter too much either way TBH

rrahir: I changed it but it comes to a functional cos (at least I think so)

Before: if you typed a value, it would filter the results and then you could navigate through those results to select one proposal. If you typed enter on it, it would select the proposal without resetting the list of proposals

Now: it does reset the list of proposals and I think it's wrong

LucasLefevre: Let's keep it as it is then
(maybe with a test for the behavior you describe)

rrahir: => added the test `selecting a proposal does not reset the suggestions`

LucasLefevre:
robodoo r+

PR #7958, opened 2026-02-18
[IMP] chart: add geo chart region selector

pro-odoo:
This commit adds a region selector to the geo charts in dashboard mode.

Task-5048552

### DIFF
--- a/packages/o-spreadsheet-engine/src/plugins/ui_feature/geo_features.ts
+++ b/packages/o-spreadsheet-engine/src/plugins/ui_feature/geo_features.ts
@@ -1,4 +1,6 @@
-import { GeoChartRegion } from "../../types/chart/geo_chart";
+import { GeoChartDefinition, GeoChartRegion } from "../../types/chart/geo_chart";
+import { Command } from "../../types/commands";
+import { UID } from "../../types/misc";
 import { ModelConfig } from "../../types/model";
 import { UIPlugin, UIPluginConfig } from "../ui_plugin";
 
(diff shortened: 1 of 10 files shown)


src/components/figures/chart/chart_dashboard_menu/chart_dashboard_menu.xml
```diff
@@ -2,6 +2,15 @@
   <t t-name="o-spreadsheet-ChartDashboardMenu">
     <div class="o-dashboard-chart-select position-absolute top-0 end-0" t-on-click.stop="">
       <div class="d-flex align-items-center gap-1 p-1 rounded" t-att-style="backgroundColor">
+        <t t-if="regionOptions.length > 1">
+          <Select
```
hokolomopo, line 6: I'd try to tweak a bit the style of this Select to be a bit more similar to the chart granularity selector, which is a bit more pretty (no bottom border, rounded background, arrow not disappearing on hover, ...)

packages/o-spreadsheet-engine/src/plugins/ui_feature/geo_features.ts
```diff
@@ -26,6 +72,28 @@ export class GeoFeaturePlugin extends UIPlugin {
...
+    const chart = this.getters.getChart(chartId);
+    if (!chart || !chart.type.includes("geo")) {
+      return [];
+    }
+    const initialRegion = this.initialRegions[chartId];
+    if (!initialRegion) {
+      return [];
+    }
+    if (this.geoJsonService.getAlternativeRegions) {
+      return this.geoJsonService.getAlternativeRegions(initialRegion);
```
rrahir:
```suggestion
      return this.geoJsonService.getAlternativeRegions?.(initialRegion) || [];
```

packages/o-spreadsheet-engine/src/types/commands.ts
```diff
@@ -607,6 +608,12 @@ export interface DeleteChartCommand extends SheetDependentCommand {
   chartId: UID;
 }
 
+export interface UpdateChartRegionCommand {
+  type: "UPDATE_CHART_REGION";
+  chartId: UID;
+  region: string;
```
rrahir, line 614: don't know if it's worth listing the exact regions that we support ?

tests/test_helpers/helpers.ts
```diff
@@ -988,6 +991,13 @@ export const mockGeoJsonService: ModelExternalConfig["geoJsonService"] = {
     }
     return "";
   },
+  getAlternativeRegions: (initialRegion: string) => {
+    if (initialRegion !== "world") {
+      return [];
+    }
+    const allRegions = getAvailableRegions();
+    return allRegions.filter((r) => r.id !== "usa");
```
rrahir, line 999: random thought: this looks a bit magic, I got the idea that you either group by  countries or states but rejecting 'usa' looks weird out of context. In the future, we could think of categories (or trees) to filter out the kind of regions we can switch against.

packages/o-spreadsheet-engine/src/types/commands.ts
pro-odoo: I think no, the regions are given from the geoJsonService

tests/test_helpers/helpers.ts
pro-odoo: Yes we could, but it's inside the geoJsonService so I wouldn't do something generic for that

rrahir:
robodoo r+

PR #7974, opened 2026-02-20
[FIX] Model: reject data that postdate the library version

rrahir:
Currently, if we try to load data that postdate tjhe library version, the code will consider the data as "versionless" and will apply every migration step to it, which makes no sense as the data already went through those changes.

With this revision, we reject the data that postdate the library version.

Task: 5895572

## Description:

Task: [5895572](https://www.odoo.com/odoo/2328/tasks/5895572)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/migrations/data.ts
+++ b/src/migrations/data.ts
@@ -134,7 +134,14 @@ function compareVersions(v1: string, v2: string): number {
 function migrate(data: any): WorkbookData {
   const start = performance.now();
   const versions = getSortedVersions();
-  const index = versions.findIndex((v) => v === data.version);
+  const index = versions.findIndex((v) => compareVersions(v, data.version) >= 0);
+  if (index === -1) {
+    throw new Error(
+      `Data version ${
+        data.version
+      } postdates the current version of o-spreadsheet (version ${getCurrentVersion()}). It cannot be loaded.`
+    );
+  }
   for (let i = index + 1; i < versions.length; i++) {
     const nextVersion = versions[i];
     data = migrationStepRegistry.get(nextVersion).migrate(data);
--- a/tests/model/model_import_export.test.ts
+++ b/tests/model/model_import_export.test.ts
@@ -981,3 +981,15 @@ test("Can import spreadsheet with only version", () => {
   // We expect the model to be loaded without traceback
   expect(true).toBeTruthy();
 });
+
+test("Reject data import from data with a subsequent version", () => {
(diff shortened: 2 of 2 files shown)


src/migrations/data.ts
```diff
@@ -134,7 +134,14 @@ function compareVersions(v1: string, v2: string): number {
 function migrate(data: any): WorkbookData {
   const start = performance.now();
   const versions = getSortedVersions();
-  const index = versions.findIndex((v) => v === data.version);
+  const index = versions.findIndex((v) => v >= data.version);
```
LucasLefevre: you cannot compare string versions with `>=` because `"19.2"` would be after `"19.12"`

-> use `compareVersions`

src/migrations/data.ts
```diff
@@ -134,7 +134,14 @@ function compareVersions(v1: string, v2: string): number {
...
+  if (index === -1) {
+    throw new Error(
```
LucasLefevre, line 139: throwing an error is better then the current situation,
but it happens quite often that I freeze a spreadsheet from odoo to load it in o-spreadsheet.
I always need to manually force the version, I'll still need to do it :sad:

tests/model/model_import_export.test.ts
```diff
@@ -981,3 +981,15 @@ test("Can import spreadsheet with only version", () => {
...
+
+test("Reject data import from data with a subsequent version", () => {
+  const futureVersion = (Number(getCurrentVersion()) + 1).toString();
+  expect(() => new Model({ version: futureVersion })).toThrow(
+    `Data version ${futureVersion} postdates the current version of o-spreadsheet (version ${getCurrentVersion()}). It cannot be loaded.`
+  );
+});
+
+test("Accept data that predates the latest version while not being present in the migration steps", () => {
+  const futureVersion = "16.3.1";
```
LucasLefevre:
```suggestion
  const pastVersion = "16.3.1";
```

tests/model/model_import_export.test.ts
```diff
@@ -981,3 +981,15 @@ test("Can import spreadsheet with only version", () => {
   // We expect the model to be loaded without traceback
   expect(true).toBeTruthy();
 });
+
+test("Reject data import from data with a subsequent version", () => {
+  const futureVersion = (Number(getCurrentVersion()) + 1).toString();
```
LucasLefevre: `Number("19.1.2")` -> `NaN`

PR #7998, opened 2026-02-23
[IMP] error_tooltip: add button to automatically add missing headers

hokolomopo:
## Description

This commit adds a button to the error tooltip to automatically add missing rows/columns on a #SPILL! error.

Task: [5959884](https://www.odoo.com/odoo/2328/tasks/5959884)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/error_tooltip/error_tooltip.ts
+++ b/src/components/error_tooltip/error_tooltip.ts
@@ -1,7 +1,8 @@
+import { _t } from "@odoo/o-spreadsheet-engine";
 import { SpreadsheetChildEnv } from "@odoo/o-spreadsheet-engine/types/spreadsheet_env";
 import { Component } from "@odoo/owl";
-import { deepEquals, positionToZone } from "../../helpers";
-import { CellPosition, CellValueType } from "../../types";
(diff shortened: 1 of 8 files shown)


src/components/error_tooltip/error_tooltip.ts
```diff
@@ -61,6 +62,65 @@ export class ErrorToolTip extends Component<ErrorToolTipProps, SpreadsheetChildE
...
+    if (!cell || !cell.isFormula) {
+      return;
+    }
+    const formula = cell.compiledFormula.toFormulaString(this.env.model.getters);
+    return getMissingHeadersForSpreadResult(
+      this.env.model.getters,
+      this.props.cellPosition,
+      formula
+    );
+  }
```
hokolomopo, line 89: There's a world where instead of all these checks and re-evaluating the formula, we just add the info of the missing cols/rows directly in the evaluatedCell at evaluation (like errorOriginPosition).

But I'm not sure if we really want to add more crap to the evaluated cell ...

PR #8007, opened 2026-02-26
[PERF] clipboard: OS HTML parsing on internal paste

dhrp-odoo:
## Description:

Before this change, paste always parsed OS clipboard HTML (`DOMParser` + `JSON.parse`) to decide whether the paste was internal or external.

For internal paste, this is unnecessary because the spreadsheet clipboard data is already in memory after COPY/CUT. We only need a quick check to know if the OS clipboard still belongs to this spreadsheet instance.

This change uses a fast clipboard id check to detect internal paste and dispatch `PASTE` directly.

If the clipboard id does not match, we keep the existing fallback path and use `PASTE_FROM_OS_CLIPBOARD`.

Perf setup:
- `this.createModel(makeLargeDataset(26, 500, ["numbers"]))`
- 26x500 numeric dataset (no style/format)
- measured internal paste `A1:A500 -> A501:A1000` (`n=20`)

Paste (internal, A1:Z500 -> A501:Z1000)
- before: Mean: 983.98 ms, StdErr: 9.66 ms, n=20
- after:    Mean: 355.63 ms, StdErr: 4.63 ms, n=20 (-63.86%)

Task: [5959666](https://www.odoo.com/odoo/2328/tasks/5959666)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/packages/o-spreadsheet-engine/src/helpers/clipboard/clipboard_helpers.ts
+++ b/packages/o-spreadsheet-engine/src/helpers/clipboard/clipboard_helpers.ts
@@ -100,6 +100,15 @@ export function parseOSClipboardContent(content: OSClipboardContent): ParsedOSCl
   return osClipboardContent;
 }
 
+/**
+ * Fast-path extraction used to detect an internal o-spreadsheet paste without
+ * instantiating a DOMParser on large HTML payloads. Reads the dedicated
+ * `data-osheet-clipboard-id` marker instead of parsing the embedded JSON.
+ */
+export function getOSheetClipboardIdFromHTML(htmlContent: string | undefined): string | undefined {
+  return htmlContent?.match(/<div data-osheet-clipboard-id=(['"])([^'"]+)\1/)?.[2];
+}
+
 function getOSheetDataFromHTML(htmlDocument: Document) {
   const attributes = [...htmlDocument.documentElement.attributes];
   // Check if it's a Microsoft Office clipboard data (it will have some namespaces defined in the root element)
--- a/packages/o-spreadsheet-engine/src/plugins/ui_stateful/clipboard.ts
+++ b/packages/o-spreadsheet-engine/src/plugins/ui_stateful/clipboard.ts
@@ -532,7 +532,6 @@ export class ClipboardPlugin extends UIPlugin {
   private getSheetData(): SpreadsheetClipboardData {
     const data = {
       version: getCurrentVersion(),
-      clipboardId: this.clipboardId,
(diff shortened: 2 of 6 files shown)


packages/o-spreadsheet-engine/src/helpers/clipboard/clipboard_helpers.ts
```diff
@@ -100,6 +100,36 @@ export function parseOSClipboardContent(content: OSClipboardContent): ParsedOSCl
...
+  const quote = htmlContent[attrIndex + attr.length];
+  if (quote !== "'" && quote !== '"') {
+    return;
+  }
+
+  const idStartIndex = attrIndex + attr.length + 1;
+  const idEndIndex = htmlContent.indexOf(quote, idStartIndex);
+  if (idEndIndex === -1) {
+    return;
+  }
```
VincentSchippefilt: isn't this entire function the same as this regex: `/^<div data-osheet-clipboard-id=["'](\d+)["']/gi` ?

dhrp-odoo: Yes, you’re right about simplifying this, but UUIDs aren’t digits-only, so we need to tweak it a bit.

`htmlContent?.match(/\bdata-osheet-clipboard-id=(['"])([^'"]+)\1/)?.[2]`

IMO this should be fine. Or should we make it stricter for UUID?

VincentSchippefilt: it's fine for the matcher, but you could start the regex with `^<div data-osheet-clipboard-id=` so that it must *begin* with `<div data...`, no need to look inside the html further than the start of the html normally

dhrp-odoo: Updated the code to strictly check with `/<div data-.../`, but we cannot use `^` here.

Reason (I learned this during testing):
When we set clipboard HTML as `<div data-...>...</div>`, `navigator.clipboard.read()` can return the same fragment.  
But with `Ctrl+V` (`e.clipboardData.getData("text/html")`), browsers often normalize the content through the paste pipeline and expose it as a full HTML document (`<html><head><body>...`).

So an anchored pattern like `^<div ...` is not reliable for real paste events.

JSFiddle repro: https://jsfiddle.net/pv8cd5tg/2/

PR #8010, opened 2026-02-27
[IMP] export: squish consecutive numbers

VincentSchippefilt:
When snapshotting a spreadsheet, we collapse all the numbers that are
the same or the same increment into a single entry in the JSON.

Example:
Before:
{
    "A1": 1,
    "A2": 3,
    "A3": 5,
    "A4": 7,
    "A5": 9,
}
After:
{
    "A1": 1,
    "A2:A5": "+2",
}

Task: 5959334
## Description:

Task: [5959334](https://www.odoo.com/odoo/2328/tasks/5959334)

### DIFF
--- a/packages/o-spreadsheet-engine/src/plugins/core/squisher.ts
+++ b/packages/o-spreadsheet-engine/src/plugins/core/squisher.ts
@@ -42,6 +42,8 @@ export class Squisher {
   // whether the base formula was already transformed. Formulas that have already been transformed must continue to be transformed
   private baseFormulaWasTransformed: boolean = false;
 
+  private baseNumber: number | undefined = undefined;
+
   constructor(getters: CoreGetters) {
     this.getters = getters;
   }
@@ -84,10 +86,11 @@ export class Squisher {
     this.alreadyAppliedNumberOffsets = formula.literalValues.numbers.map((_) => 0);
     this.previousStrings = formula.literalValues.strings.map((x) => x.value);
     this.baseFormulaWasTransformed = false;
+    this.baseNumber = undefined;
   }
 
   /** Reset the base formula to undefined, resetting all offsets */
-  resetBase() {
+  resetBaseFormula() {
     if (this.baseFormula) {
       this.baseFormula = undefined;
       this.alreadyAppliedNumberOffsets = [];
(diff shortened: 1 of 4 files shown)


packages/o-spreadsheet-engine/src/plugins/core/squisher.ts
```diff
@@ -110,42 +113,71 @@ export class Squisher {
...
+        this.baseNumber = numberValue;
+        return { N: (numberOffset > 0 ? "+" : "") + numberOffset.toString() };
       }
-      numbers = this.squishNumbers(cell.compiledFormula.literalValues.numbers);
-      strings = this.squishStrings(cell.compiledFormula.literalValues.strings);
-      references = this.squishReferences(cell.compiledFormula.rangeDependencies, forSheetId);
-      this.baseFormulaWasTransformed = true;
     }
-    return this.buildResult(numbers, strings, references);
+    this.resetBase();
```
pro-odoo: `baseNumber` is not reset in case of decimal number I think.
Check with this test: 
```
[
  { A1: "1", A2: "1.5", A3: "2", A4: "3" },
  { A1: "1", A2: "1.5", A3: "2", A4: "3" }, // No squish, A2 split the sequence
],
```

packages/o-spreadsheet-engine/src/plugins/core/squisher.ts
```diff
@@ -110,42 +113,71 @@ export class Squisher {
...
+        numbers = this.squishNumbers(cell.compiledFormula.literalValues.numbers);
+        strings = this.squishStrings(cell.compiledFormula.literalValues.strings);
+        references = this.squishReferences(cell.compiledFormula.rangeDependencies, forSheetId);
+        this.baseFormulaWasTransformed = true;
+      }
+      return this.buildResult(numbers, strings, references);
+    }
+
+    if (
+      !cell.isFormula &&
```
pro-odoo: At this point, `!cell.isFormula` is always true

pro-odoo (commented):
The `Co-authored-by` section should be at the end of the commit, after the task-id to let github understand :)

I just found a little bug

packages/o-spreadsheet-engine/src/plugins/core/squisher.ts
```diff
@@ -110,42 +113,66 @@ export class Squisher {
    *     - for references: returns a relative change (+Ck or +Rk) if possible, else the full reference or "=" if unchanged
    * */
   squish(cell: Cell, forSheetId: UID): SquishedCell {
-    if (!cell.isFormula) {
+    if (!cell.isFormula && (typeof cell.parsedValue !== "number" || cell.parsedValue % 1 !== 0)) {
       this.resetBase();
+      this.baseNumber = undefined;
       return cell.content;
     }
```
fdamhaut: nitpick: this seems uselessly complex, both of the following if always return a value (so this condition is equivalent to the base case at the end 
(if we add the %1!=0 to the number condition))

on top of that this cut the condition for parsed Number at two != line

packages/o-spreadsheet-engine/src/plugins/core/squisher.ts
fdamhaut: nitpick: rename to resetBaseFormula or smth

packages/o-spreadsheet-engine/src/plugins/core/squisher.ts
VincentSchippefilt: good point ! 👍

rrahir:
robodoo r+

PR #8030, opened 2026-03-03
[FIX] chart: ensure chart values remain visible (remove clipping)

rmbh-odoo:
## Description:

Current behavior before PR:
- Chart value labels could be clipped and become hidden when rendered outside
the drawing area.
- yPosition was constrained within chart bounds using min/max limits.

- Clipping is removed so chart value labels remain visible even when rendered
outside the drawing area.
- yPosition is no longer clamped to chart min/max bounds.

Task: [5993132](https://www.odoo.com/odoo/2328/tasks/5993132)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/components/figures/chart/chartJs/chartjs_show_values_plugin.ts
```diff
@@ -82,12 +81,28 @@ function drawLineOrBarChartValues(
...
       const point = dataset.data[i];
-
-      const xPosition = point.x;
-
+      const valueToDisplay = options.callback(Number(value), dataset, i);
+      let xPosition = point.x;
       let yPosition = 0;
       if (chart.config.type === "line") {
-        yPosition = point.y - 10;
+        const textHeight = 12; // ChartJS default
```
rrahir: rely on the actual font value to be consistent.
```Chart.defaults.font.size ```
:-)

Comment (rmbh-odoo) [reply src/components/figures/chart/chartJs/chartjs_show_values_plugin.ts]: Great :+1: 
Also, since it can be `undefined`, I’ve added a fallback check so that it defaults back to a height of `12` in that case.

File: src/components/figures/chart/chartJs/chartjs_show_values_plugin.ts [hunk 2]
```diff
@@ -1,4 +1,4 @@
-import { ChartDataset, ChartType, Plugin } from "chart.js";
+import { Chart, ChartDataset, ChartType, Plugin } from "chart.js";
```
hokolomopo: IIRC @anhe-odoo deletes the `clip` thatmakes labels half-hidden in some PR (bubble chart?). Can you confirm @anhe-odoo ? If so, we could just do that in 18.0, and accept that the text sometimes goes beyond the chart area.

anhe-odoo: Sorry for the late answer. Indeed I do it in the bubble chart PR, but only for the bubble chart, not for all the other charts (as you suggested)

dhrp-odoo:
Should we also remove this or is it needed for some cases? https://github.com/odoo/o-spreadsheet/blob/19.0/src/components/figures/chart/chartJs/chartjs_show_values_plugin.ts#L109-L110

It clamps yPosition so it stays inside the chart’s vertical drawing area. but now values can be shown outside the drawing area as well :thinking:

rmbh-odoo:
Hello @dhrp-odoo :wave: 
Nice catch! I think we no longer need the additional position-based clipping since the edge clipping has already been removed. The rendering now looks more natural around the edges.

hokolomopo:
@robodoo r+

PR #8035, opened 2026-03-03
[FIX] selection_input: disable spill references

hokolomopo:
## Description

The selection input were replacing classical range strings with spilled references (eg. `A1#`), but those type of ranges are not supported in side panels/charts/pivots/etc..

This commit disable spilled references in selection inputs, we'll see in the future is we want to support them or not.

Task: [5945112](https://www.odoo.com/odoo/2328/tasks/5945112)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


rrahir:
robodoo r+

PR #8037, opened 2026-03-03
[FIX] xlsx: do not export dynamic tables to excel

hokolomopo:
## Description

Excel does not support array formulas inside tables. With this commit, if a table contains an array formula, it will not be exported to excel but be exported as individual cell styles instead.

Task: [5214240](https://www.odoo.com/odoo/2328/tasks/5214240)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_core_views/dynamic_tables.ts
+++ b/src/plugins/ui_core_views/dynamic_tables.ts
@@ -1,12 +1,15 @@
 import {
   areZonesContinuous,
   deepEquals,
+  getItemId,
   getZoneArea,
   isInside,
+  isObjectEmptyRecursive,
   overlap,
+  positions,
+  toXC,
   toZone,
   union,
-  zoneToXc,
 } from "../../helpers";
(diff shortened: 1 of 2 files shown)


src/plugins/ui_core_views/dynamic_tables.ts
```diff
@@ -221,19 +225,57 @@ export class DynamicTablesPlugin extends UIPlugin {
...
+
+          if (!isObjectEmptyRecursive(style)) {
+            const styleId = getItemId(style, data.styles);
+            sheet.styles[xc] = styleId;
+            cell.style = styleId;
+          }
+
+          if (border) {
+            const borderId = getItemId(border, data.borders);
+            sheet.borders[xc] = borderId;
```
rrahir, line 273: in later versions, the borders are stored by zone, so we will probably need to split the said zones with the new border definition. same goes for the style...

rrahir:
robodoo r+

PR #8038, opened 2026-03-04
[FIX] composer: composer hover in dark mode

hokolomopo:
### [FIX] composer: composer hover in dark mode

Add a dark mode variant to `os-composer-highlight-flag-color`. Also
change the light mode color to a standard gray color.

### [REV] composer: use opacity rather than color alpha

This reverts commit 9cf4b117a9941690ad92e21716464965e6f27556.

Using opacity rather than color alpha also impacts the background
color of the token, which we do not want. This is especially visible
on dark mode, with a black composer background.

With this commit, we go back to setting the color alpha rather than the
opacity of the token. This means that the `setColorAlpha` helper now
needs to handle `light-dark` colors.

Task: [5928811](https://www.odoo.com/odoo/2328/tasks/5928811)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/variables.css
+++ b/src/variables.css
@@ -37,7 +37,7 @@
   --os-composer-placeholder-color: light-dark(#bdbdbd, #757a89);
   --os-composer-assistant-color: #9b359b; /* COMPOSER_ASSISTANT_COLOR */
   --os-composer-assistant-background: light-dark(#f2f2f2, var(--os-gray-200));
-  --os-composer-highlight-flag-color: #e6edf3;
+  --os-composer-highlight-flag-color: light-dark(var(--os-gray-200), var(--os-gray-400));
   --os-composer-selector-flag-color: #a9a9a9;
 
   --os-selection-border-color: #3266ca; /* SELECTION_BORDER_COLOR */


packages/o-spreadsheet-engine/src/helpers/color.ts
```diff
@@ -4,6 +4,8 @@ import { concat } from "./misc";
 
 const RBA_REGEX = /rgba?\(|\s+|\)/gi;
 const HEX_MATCH = /^#([A-F\d]{2}){3,4}$/;
+const LIGHT_DARK_REGEX =
+  /^light-dark\(\s*(rgba?\([0-9.,\s]+\)|#[a-z0-9]+)\s*,\s*(rgba?\([0-9.,\s]+\)|#[a-z0-9]+)\s*\)$/i;
```
rrahir:
```suggestion
  /^light-dark\(\s*(rgba?\([0-9.,\s]+\)|#[a-f0-9]+)\s*,\s*(rgba?\([0-9.,\s]+\)|#[a-f0-9]+)\s*\)$/i;
```
or we could let pretty much everything pass and let the subsequent calls to rgba and toHex to raise the error ?

hokolomopo: I want a semi-solid regex because I'm using it to get the two colors of the `light-dark` expression. Splitting the two parts w/o a regex is a bit annoying with commas being both the light-dark separator, and being present in `rgb()` colors

rrahir:
robodoo rebase-ff r+

PR #8040, opened 2026-03-05
pivot : filter out empty rows

matho-odoo:
Task: [5947935](https://www.odoo.com/odoo/2328/tasks/5947935)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/pivot/spreadsheet_pivot/spreadsheet_pivot.ts
+++ b/src/helpers/pivot/spreadsheet_pivot/spreadsheet_pivot.ts
@@ -534,6 +534,7 @@ export class SpreadsheetPivot implements Pivot<SpreadsheetPivotRuntimeDefinition
       const zone = { top: row, bottom: row, left: range.zone.left, right: range.zone.right };
       const cells = this.getters.getEvaluatedCellsInZone(range.sheetId, zone);
       const entry: DataEntry = {};
+      let isEntryEmpty = true;
       for (const index in cells) {
         const cell = cells[index];
         const field = this.fields[this.metaData.fieldKeys[index]];
@@ -544,6 +545,9 @@ export class SpreadsheetPivot implements Pivot<SpreadsheetPivotRuntimeDefinition
           entry[field.name] = { value: null, type: CellValueType.empty, formattedValue: "" };
         } else {
           entry[field.name] = cell;
+          if (cell.type !== CellValueType.empty) {
+            isEntryEmpty = false;
+          }
         }
       }
       for (const customFieldName in this.definition.customFields || {}) {
@@ -565,8 +569,10 @@ export class SpreadsheetPivot implements Pivot<SpreadsheetPivotRuntimeDefinition
           value: group ? group.name : baseValue.value,
         };
       }
-      entry["__count"] = { value: 1, type: CellValueType.number, formattedValue: "1" };
-      dataEntries.push(entry);
+      if (!isEntryEmpty) {
+        entry["__count"] = { value: 1, type: CellValueType.number, formattedValue: "1" };
+        dataEntries.push(entry);
+      }
     }
     const dateDimensions = this.definition.columns
       .concat(this.definition.rows)
--- a/tests/pivots/spreadsheet_pivot/pivot_filter/pivot_filter_plugin.test.ts
+++ b/tests/pivots/spreadsheet_pivot/pivot_filter/pivot_filter_plugin.test.ts
@@ -162,6 +162,31 @@ describe("pivot table with filters", () => {
       ["Total", "10"]
     ]);
   });
+
+  test("Filter empty rows", () => {
+    // prettier-ignore
+    const grid = {
+      A1: "Customer", B1: "Price", C1: "=PIVOT(1)",
+      A2: "Alice",    B2: "10",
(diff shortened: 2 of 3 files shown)


src/helpers/pivot/spreadsheet_pivot/spreadsheet_pivot.ts
```diff
@@ -397,6 +397,14 @@ export class SpreadsheetPivot implements Pivot<SpreadsheetPivotRuntimeDefinition
...
       return [];
     }
+    dataEntries = dataEntries.filter((dataEntry) => {
+      for (const field in dataEntry) {
+        if (field !== "__count" && dataEntry[field]?.value !== null) {
+          return true;
+        }
+      }
+      return false;
+    });
```
hokolomopo: Note that filtering on the data entries will keep data entries that are only non-empty on values that we don't use in the pivot, so we still can get empty rows. 

Image

It's probably fine. Not sure how we could do it differently. Or if we even want to.

src/helpers/pivot/spreadsheet_pivot/spreadsheet_pivot.ts
```diff
@@ -397,6 +397,14 @@ export class SpreadsheetPivot implements Pivot<SpreadsheetPivotRuntimeDefinition
     if (!sheetId) {
       return [];
     }
+    dataEntries = dataEntries.filter((dataEntry) => {
```
hokolomopo: I would directly change the result in the main loop of `extractDataEntriesFromRange`, instead of looping another time on the data entries here.

hokolomopo (commented):
👋

tests/pivots/spreadsheet_pivot/pivot_filter/pivot_filter_plugin.test.ts
```diff
@@ -162,6 +162,31 @@ describe("pivot table with filters", () => {
...
+  test("Filter empty rows", () => {
+    // prettier-ignore
+    const grid = {
+      A1: "Customer", B1: "Price", C1: "=PIVOT(1)",
+      A2: "Alice", B2: "10",
+      A3: "", B3: "",
+      A4: "Olaf", B4: "30",
+    };
+    const model = createModelFromGrid(grid);
+    addPivot(model, "A1:B4", {
```
LucasLefevre: Since the main use-case is empty rows at the end to account for future data, let's test that 😉 
```suggestion
    addPivot(model, "A1:B5", {
```

tests/pivots/spreadsheet_pivot/spreadsheet_pivot.test.ts
```diff
@@ -2132,7 +2132,7 @@ describe("Spreadsheet Pivot", () => {
       columns: [{ fieldName: "Date", granularity: "month_number" }],
       measures: [{ id: "__count:sum", fieldName: "__count", aggregator: "sum" }],
     });
-    expect(getEvaluatedGrid(model, "B4:E4")).toEqual([["March", "(Undefined)", "Total", ""]]);
```
LucasLefevre, line 2135: I think this test really meant to test empty dates are working and give `(Undefined)`.
So this test should be adapted in another way: add data for a measure, such that the data entry is not fully empty

Image

tests/pivots/spreadsheet_pivot/pivot_filter/pivot_filter_plugin.test.ts
```diff
@@ -162,6 +162,31 @@ describe("pivot table with filters", () => {
...
     ]);
   });
+
+  test("Filter empty rows", () => {
+    // prettier-ignore
+    const grid = {
+      A1: "Customer", B1: "Price", C1: "=PIVOT(1)",
+      A2: "Alice", B2: "10",
+      A3: "", B3: "",
+      A4: "Olaf", B4: "30",
```
LucasLefevre: if you have `// prettier-ignore`, the goal is to format the grid in the shape of a grid 😉 
```suggestion
    // prettier-ignore
    const grid = {
      A1: "Customer", B1: "Price", C1: "=PIVOT(1)",
      A2: "Alice",    B2: "10",
      A3: "",         B3: "",
      A4: "Olaf",     B4: "30",
```

tests/pivots/spreadsheet_pivot/pivot_filter/pivot_filter_plugin.test.ts
```diff
@@ -162,6 +162,31 @@ describe("pivot table with filters", () => {
       ["Total", "10"]
     ]);
   });
+
+  test("Filter empty rows", () => {
+    // prettier-ignore
+    const grid = {
+      A1: "Customer", B1: "Price", C1: "=PIVOT(1)",
+      A2: "Alice", B2: "10",
+      A3: "", B3: "",
```
LucasLefevre: I'm not entirely sure if this actually sets the cell content to an empty string or if it's ignored.
with this, we are sure 👇 
```suggestion
      A3: '=""', B3: '=""', // all empty strings
```

pro-odoo:
robodoo r+

PR #8042, opened 2026-03-05
[FIX] xlsx: export formula that returns an error

LucasLefevre:
Steps to reproduce:
- in A1: "hello"
- in A2: =A1+1
- export to xlsx file
- import to excel

⮕ the formula in A2 is recognized as a string rather than a formula

Because the value was set to `undefined`, the export later treats it as string rather than a formula

Task: 6008904

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/tests/xlsx/xlsx_export.test.ts
+++ b/tests/xlsx/xlsx_export.test.ts
@@ -1066,6 +1066,20 @@ describe("Test XLSX export", () => {
       });
       expect(await exportPrettifiedXlsx(model)).toMatchSnapshot();
     });
+
+    test("#BAD_EXPR formula is exported as a string, not a formula", async () => {
+      const model = new Model();
+      setCellContent(model, "A1", "=THISISBAD())))");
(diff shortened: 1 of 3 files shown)


tests/xlsx/xlsx_export.test.ts
```diff
@@ -1066,6 +1066,17 @@ describe("Test XLSX export", () => {
...
+    test("formula whose value is an error", async () => {
+      const model = new Model({
+        sheets: [
+          {
+            cells: { A1: "hello", B2: "=A1 + 2" },
+          },
+        ],
+      });
+      expect(await exportPrettifiedXlsx(model)).toMatchSnapshot();
+    });
```
hokolomopo: If the fix was only on the result of `plugin.exportForExcel()`, i'd test the result of `getExportedExcelData` directly, Or queryselector the relevant data from the xml. Snapshot tests of xlsx are totally useless for fixes IMO

PR #8043, opened 2026-03-05
[REF] *: prepare to owl 3

pro-odoo:
## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/components/composer/formula_assistant/formula_assistant.ts
```diff
@@ -18,7 +19,7 @@ export class FunctionDescriptionProvider extends Component<Props, SpreadsheetChi
   };
   static components = { Collapse };
 
-  private state: { isCollapsed: boolean } = useState({
+  private state: { isCollapsed: boolean } = proxy({
```
dhrp-odoo, line 22: 👋 Just to understand: Since `isCollapsed` is a scalar value, would `isCollapsed = signal(true);` be a better fit here? Or are we using `proxy(...)` everywhere for consistency?

pro-odoo: Yes it would be a better fit :) 
This PR is only about converting "blindly" from owl2 to owl3, so to replace `useState` by `proxy`. When it will be merged in Odoo, we will have to convert the code to follow owl3 philosophy :)

pro-odoo:
robodoo rebase-ff

rrahir:
robodoo r+

PR #8044, opened 2026-03-05
[FIX] pivot cell style copied twice

matho-odoo:
If we copy the top left cell or the all pivot table, we should not copy the style of the cell

Task: [5909138](https://www.odoo.com/odoo/2328/tasks/5909138)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/tests/pivots/pivot_insert.test.ts
+++ b/tests/pivots/pivot_insert.test.ts
@@ -1,9 +1,9 @@
 import { PIVOT_INSERT_TABLE_STYLE_ID } from "@odoo/o-spreadsheet-engine/constants";
 import { Model } from "../../src";
 import { toZone } from "../../src/helpers";
-import { insertPivot } from "../test_helpers/commands_helpers";
-import { getCellText, getTable } from "../test_helpers/getters_helpers";
-import { setGrid } from "../test_helpers/helpers";
+import { copy, insertPivot, paste } from "../test_helpers/commands_helpers";
+import { getCell, getCellText, getTable } from "../test_helpers/getters_helpers";
+import { createModelFromGrid, setGrid } from "../test_helpers/helpers";
 
 describe("Insert pivot command", () => {
   test("Can insert a pivot in a cell", () => {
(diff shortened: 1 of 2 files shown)


tests/pivots/pivot_insert.test.ts
```diff
@@ -44,3 +44,42 @@ describe("Insert pivot command", () => {
...
+          cells: {
+            A1: "1",
+            A2: "2",
+          },
+        },
+      ],
+    });
+    insertPivot(model, "A1", "pivot1", "Sheet2");
+  });
+  test("copy/paste the top left cell", () => {
```
hokolomopo: nitpick: use createModelWithPivot

tests/pivots/pivot_insert.test.ts
```diff
@@ -44,3 +44,42 @@ describe("Insert pivot command", () => {
...
+    copy(model, "A1");
+    paste(model, "C1");
+    expect(getCell(model, "C1")?.style).toBe(undefined);
```
hokolomopo, line 42: also test that the resulting copied cell has a table

tests/pivots/pivot_insert.test.ts
```diff
@@ -44,3 +44,42 @@ describe("Insert pivot command", () => {
...
+        {
+          id: "Sheet1",
+          cells: {
+            A1: "1",
+            A2: "2",
+          },
+        },
+      ],
+    });
+    insertPivot(model, "A1", "pivot1", "Sheet2");
```
LucasLefevre: avoid `beforeEach` which setup data:

- when reading the test, you don't have the full picture to understand the test: you need to scroll and look somewhere else.
- Every test should be minimal. Sooner or later, someone will add a test and need other data, or not need all of the data, but use it anyway -> it makes the test harder to read, and you don't know what it's testing anymore.

Here it's not a lot of data, and there's only a few tests, but it's a bad practice in general.

bottom line: avoid `beforeEach` with setup data, even if it means copy-pasting a few line.

LucasLefevre: Let's be explicit, rather than relying on the fact that it selects the entire table.
```suggestion
          },
        },
      ],
    });
    insertPivot(model, "A1:A2", "pivot1", "Sheet2");
```

tests/pivots/pivot_insert.test.ts
LucasLefevre: With `createModelWithPivot`, tests are not minimal. It should be banned.
It's easy when writing the test initially, but it's a foot-gun when we need to later debug.

tests/pivots/pivot_insert.test.ts
```diff
@@ -44,3 +44,42 @@ describe("Insert pivot command", () => {
...
+  });
+  test("copy/paste the top left cell", () => {
+    copy(model, "A1");
+    paste(model, "C1");
+    expect(getCell(model, "C1")?.style).toBe(undefined);
+  });
+
+  test("copy/paste the all table", () => {
```
LucasLefevre:
```suggestion
  test("copy/paste the whole table", () => {
```

tests/pivots/pivot_insert.test.ts
```diff
@@ -44,3 +44,66 @@ describe("Insert pivot command", () => {
...
+      sheets: [
+        {
+          id: "Sheet1",
+          cells: {
+            A1: "1",
+            A2: "2",
+          },
+        },
+      ],
+    });
```
hokolomopo: use createModelFromGrid instead (same in other tests)

PR #8045, opened 2026-03-05
[IMP] rendering : add dark mode for grid and chart

anhe-odoo:
This commit adds the isDarkMode in the configuration of the model. We then use a getters isDarkMode() to know if we are in a dark mode rendering, and a new plugin ColorThemeUIPlugin to adapt the colors we want to draw/use according to the current color theme. This getters is then used in the color picker and in the grid renderer for cell text, background, border, DV, table and CF.

For the grid (border, background, headers, frozen pane), a theme has been fixed with colors depending on the current color scheme.

For the chart, we currently keep all the colors set by the user as is, and all default colors (scale, grid, title, legend, default background) are updated.

## Related Task:

- Task: [5245940](https://www.odoo.com/odoo/2328/tasks/5245940)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/demo/main.js
+++ b/demo/main.js
@@ -62,7 +62,7 @@ let start;
 
 class Demo extends Component {
   setup() {
-    this.state = useState({ key: 0, displayHeader: false, colorScheme: "light" });
+    this.state = useState({ key: 0, displayHeader: false });
     this.stateUpdateMessages = [];
     this.client = {
       id: UuidGenerator.uuidv4(),
@@ -107,9 +107,11 @@ class Demo extends Component {
       name: "Toggle dark mode",
       sequence: 12.5,
       isReadonlyAllowed: true,
-      execute: () =>
-        (this.state.colorScheme = this.state.colorScheme === "dark" ? "light" : "dark"),
-
+      execute: () => {
+        this.model.dispatch("UPDATE_COLOR_SCHEME", {
+          colorScheme: this.model.getters.isDarkMode() ? "light" : "dark",
+        });
+      },
       icon: "o-spreadsheet-Icon.DARK_MODE",
       isEnabledOnLockedSheet: true,
     });
@@ -406,11 +408,11 @@ Demo.template = xml/* xml */ `
   <div t-if="state.displayHeader" class="d-flex flex flex-column justify-content w-100 h-100">
     <div class="p-3 border-bottom">A header</div>
     <div class="flex-fill">
-      <Spreadsheet model="model" notifyUser="notifyUser" t-key="state.key" colorScheme="state.colorScheme"/>
+      <Spreadsheet model="model" notifyUser="notifyUser" t-key="state.key"/>
     </div>
   </div>
   <div t-else="" class="w-100 h-100">
-    <Spreadsheet model="model" t-key="state.key" notifyUser="notifyUser" colorScheme="state.colorScheme"/>
+    <Spreadsheet model="model" t-key="state.key" notifyUser="notifyUser"/>
   </div>
 `;
 Demo.components = { Spreadsheet };
--- a/src/components/grid_add_rows_footer/grid_add_rows_footer.ts
+++ b/src/components/grid_add_rows_footer/grid_add_rows_footer.ts
@@ -36,6 +36,7 @@ export class GridAddRowsFooter extends Component<Props, SpreadsheetChildEnv> {
 
     return cssPropertiesToCss({
       top: `${top}px`,
+      ["z-index"]: "1",
     });
   }
 
--- a/src/components/spreadsheet/spreadsheet.css
+++ b/src/components/spreadsheet/spreadsheet.css
@@ -193,20 +193,21 @@
   .o-grid {
     position: relative;
     overflow: hidden;
-    background-color: var(--os-background-gray-color);
+    background-color: light-dark(var(--os-background-gray-color), #1e1f29);
     &:focus {
       outline: none;
     }
 
     > canvas {
       box-sizing: content-box;
-      border-bottom: 1px solid var(--os-grid-border-color);
-      border-right: 1px solid var(--os-grid-border-color);
+      border-bottom: 1px solid var(--os-canvas-border-color);
+      border-right: 1px solid var(--os-canvas-border-color);
     }
 
     .o-grid-overlay {
       position: absolute;
       outline: none;
+      z-index: 1;
     }
   }
 
@@ -290,5 +291,10 @@
 
   .o-checkers-background {
     background-image: url("data:image/svg+xml;utf8,<svg xmlns='http://www.w3.org/2000/svg' width='10' height='10'><path fill='#d9d9d9' d='M5 5h5v5H5zH0V0h5'/></svg>");
+  }
+  &.dark {
+    .os-theme-dependant {
+      filter: var(--os-dark-mode-filter);
+    }
   }
 }
--- a/src/helpers/color.ts
+++ b/src/helpers/color.ts
@@ -402,6 +402,87 @@ export function lightenColor(color: Color, percentage: number): Color {
   return hslaToHex(hsla);
 }
 
+export const DARK_MODE_FILTER = {
+  invert: 1,
+  hueRotate: 170,
+  contrast: 0.85,
+  brightness: 1.2,
+};
+
+export const DARK_MODE_FILTER_STRING = `invert(${DARK_MODE_FILTER.invert}) hue-rotate(${DARK_MODE_FILTER.hueRotate}deg) contrast(${DARK_MODE_FILTER.contrast}) brightness(${DARK_MODE_FILTER.brightness})`;
+
+/**
+ * Computes the 3x3 matrix for a hue-rotate filter.
+ * https://www.w3.org/TR/filter-effects-1/#feColorMatrixElement
+ */
+function getHueRotateMatrix(degrees: number): number[][] {
+  const radians = (degrees * Math.PI) / 180;
+  const cos = Math.cos(radians);
+  const sin = Math.sin(radians);
+  return [
+    [
+      0.213 + cos * 0.787 - sin * 0.213,
+      0.715 - cos * 0.715 - sin * 0.715,
+      0.072 - cos * 0.072 + sin * 0.928,
+    ],
+    [
+      0.213 - cos * 0.213 + sin * 0.143,
+      0.715 + cos * 0.285 + sin * 0.14,
+      0.072 - cos * 0.072 - sin * 0.283,
+    ],
+    [
+      0.213 - cos * 0.213 - sin * 0.787,
+      0.715 - cos * 0.715 + sin * 0.715,
+      0.072 + cos * 0.928 + sin * 0.072,
+    ],
+  ];
+}
+
+/**
+ * Converts a desired "displayed" color into the pre-inverted color that must be
+ * stored in the theme so that, after the dark-mode CSS filter is applied, the
+ * desired color is actually shown.
+ *
+ * This function computes the exact inverse of the DARK_MODE_FILTER chain.
+ */
+export function adaptForDarkMode(displayedColor: Color): Color {
+  const { r, g, b, a } = colorToRGBA(displayedColor);
+
+  // Normalize RGB channels to [0, 1]
+  let rn = r / 255;
+  let gn = g / 255;
+  let bn = b / 255;
+
+  // Step 1 – reverse brightness: c = c / brightness
+  rn /= DARK_MODE_FILTER.brightness;
+  gn /= DARK_MODE_FILTER.brightness;
+  bn /= DARK_MODE_FILTER.brightness;
+
+  // Step 2 – reverse contrast: c = (c - 0.5) / contrast + 0.5
+  rn = (rn - 0.5) / DARK_MODE_FILTER.contrast + 0.5;
+  gn = (gn - 0.5) / DARK_MODE_FILTER.contrast + 0.5;
+  bn = (bn - 0.5) / DARK_MODE_FILTER.contrast + 0.5;
+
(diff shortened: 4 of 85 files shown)


packages/o-spreadsheet-engine/src/plugins/ui_feature/color_theme.ts
```diff
@@ -0,0 +1,68 @@
...
+    if (command.type === "UPDATE_COLOR_SCHEME") {
+      this.colorScheme = command.colorScheme;
+    }
+  }
+
+  isDarkMode(): boolean {
+    return this.colorScheme === "dark";
+  }
+
+  getAdaptedColor(color: Color): Color {
```
hokolomopo: Is it still needed ? Its' only used once in the color picker, could it be replaced with css ?

packages/o-spreadsheet-engine/src/types/commands.ts
```diff
@@ -1172,6 +1172,18 @@ export interface ToggleCheckboxCommand extends TargetDependentCommand {
   type: "TOGGLE_CHECKBOX";
 }
 
+export interface UpdateChartCommand extends SheetDependentCommand {
+  type: "UPDATE_CHART";
+  figureId: UID;
+  chartId: UID;
+  definition: ChartDefinition;
+}
```
hokolomopo: Not sure it's needed 🙃

src/components/figures/figure/figure.css
```diff
@@ -83,4 +83,7 @@
       }
     }
   }
+  .o-figure-wrapper.os-theme-dependant {
+    filter: invert(1) hue-rotate(180deg) contrast(0.9) brightness(1.5);
+  }
```
hokolomopo: maybe a small comment explaining how it's different than default `os-theme-dependant`, and why it's needed here

src/components/helpers/draw_grid_hook.ts
```diff
@@ -15,7 +15,7 @@ export function useGridDrawing(refName: string, model: Model, canvasSize: () =>
   function drawGrid() {
     const canvas = canvasRef.el as HTMLCanvasElement;
     const dpr = window.devicePixelRatio || 1;
-    const ctx = canvas.getContext("2d", { alpha: false })!;
+    const ctx = canvas.getContext("2d", { alpha: true })!;
```
hokolomopo: w/o that, we cannot draw transparent colors I'm guessing ? Is the `multiply` still needed int he selection then ?

src/stores/grid_renderer_store.ts
```diff
@@ -1,20 +1,12 @@
 import { getPath2D } from "@odoo/o-spreadsheet-engine/components/icons/icons";
```
hokolomopo: checkboxes are almost invisible

src/components/figures/figure/figure.css
hokolomopo: and it's applied also in light theme

Image

src/model.ts
```diff
@@ -442,6 +442,7 @@ export class Model extends EventBus<any> implements CommandDispatcher {
       defaultCurrency: this.config.defaultCurrency,
       customColors: this.config.customColors || [],
       external: this.config.external,
+      colorScheme: this.config.colorScheme,
```
hokolomopo, line 444: will need an Odoo PR

src/stores/grid_renderer_store.ts
```diff
@@ -212,8 +196,9 @@ export class GridRenderer extends SpreadsheetStore {
     const { ctx } = renderingContext;
     for (const box of boxes) {
       const style = box.style;
-      if (style.fillColor && style.fillColor !== "#ffffff") {
-        ctx.fillStyle = style.fillColor || "#ffffff";
+      const fillColor = toHex(style.fillColor || "#ffffff");
+      if (fillColor.toLowerCase() !== "#ffffff") {
```
hokolomopo:
```suggestion
      const fillColor = toHex(style.fillColor || "#ffffff");
      if (fillColor!== "#FFFFFF") {
```

toHex always uppercases the colors

tests/renderer/renderer_store.test.ts
```diff
@@ -1874,7 +1855,7 @@ describe("renderer", () => {
 
     function getCellOverflowingBackgroundDims() {
       // first draw of white rectangle is the spreadsheet's background
-      const instruction = fillWhiteRectInstructions[1];
+      const instruction = fillWhiteRectInstructions[0];
```
hokolomopo:
```suggestion
      const instruction = clearRectInstructions[0];
```

hokolomopo (commented):
👋 

It's a bit less pretty, but waaaaay easier. The code looks nice now :D

src/components/helpers/draw_grid_hook.ts
anhe-odoo: Yes, it's still needed. The alpha-true is there to show the css background of the canva

packages/o-spreadsheet-engine/src/components/icons/icons.ts
```diff
@@ -161,6 +161,31 @@ export const CHECKBOX_CHECKED: ImageSVG = {
   ],
 };
 
+// Dark mode variants of the ckeckbox icons, as the applied CSS filter make them alsmost invisible.
```
hokolomopo:
```suggestion
// Dark mode variants of the checkbox icons, as the applied CSS filter make them almost invisible.
```
typos

packages/o-spreadsheet-engine/src/components/icons/icons.ts
```diff
@@ -161,6 +161,31 @@ export const CHECKBOX_CHECKED: ImageSVG = {
...
+};
+
+export const CHECKBOX_UNCHECKED_HOVERED_DARK: ImageSVG = {
+  name: "CHECKBOX_UNCHECKED",
+  width: 512,
+  height: 512,
+  paths: [{ fillColor: "#82d1d5", path: "M45,45 h422 v422 h-422 v-422 m30,30 v362 h362 v-362" }],
+};
+
+export const CHECKBOX_CHECKED_DARK: ImageSVG = {
```
hokolomopo: Maybe instead of creating new icons with the same path but a different color, do something like `getPivotIconSvg` where we change the fill colo based on some arguments (isDark in this case)

packages/o-spreadsheet-engine/src/plugins/ui_feature/color_theme.ts
hokolomopo: up

hokolomopo (commented):
Some nitpicks, will need a runbot/odoo PR because I can green the task 🙂

src/components/spreadsheet/spreadsheet.css
```diff
@@ -190,6 +190,7 @@
       box-sizing: content-box;
```
hokolomopo, line 202: As we discussed we should remove the spreadsheet component `colorScheme` props, and only rely on the value of getter.isDarkTheme()

src/stores/grid_renderer_store.ts
```diff
@@ -179,15 +171,6 @@ export class GridRenderer extends SpreadsheetStore {
...
 
-  private drawGlobalBackground(renderingContext: GridRenderingContext) {
-    const { ctx } = renderingContext;
-    const { width, height } = this.getters.getSheetViewDimensionWithHeaders();
-
-    // white background
-    ctx.fillStyle = "#ffffff";
-    ctx.fillRect(0, 0, width + CANVAS_SHIFT, height + CANVAS_SHIFT);
-  }
-
```
hokolomopo, line 210: not drawing a global background make cell animations break everything. We should either do a clearRect or draw the correct background color

demo/main.js
```diff
@@ -108,9 +108,10 @@ class Demo extends Component {
       name: "Toggle dark mode",
       sequence: 12.5,
       isReadonlyAllowed: true,
-      execute: () =>
-        (this.state.colorScheme = this.state.colorScheme === "dark" ? "light" : "dark"),
-
+      execute: () => {
+        this.state.colorScheme = this.state.colorScheme === "dark" ? "light" : "dark";
```
hokolomopo: this.state.colorScheme is now useless

src/plugins/ui_feature/color_theme.ts
```diff
@@ -0,0 +1,58 @@
...
+    if (command.type === "UPDATE_COLOR_SCHEME") {
+      this.colorScheme = command.colorScheme;
+    }
+  }
+
+  isDarkMode(): boolean {
+    return this.colorScheme === "dark";
+  }
+
+  getSpreadsheetTheme(): GridRenderingTheme {
```
hokolomopo, line 68: nitpick: maybe a small docsting that those colors should only be used on elements w/ inverted css

src/variables.css
```diff
@@ -46,7 +46,7 @@
   --os-background-gray-color: light-dark(#f5f5f5, var(--os-white-bg));
   --os-background-gray-color-hover: light-dark(rgba(0, 0, 0, 0.08), #ffffff40);
   --os-frozen-pane-header-border-color: #bcbcbc; /* FROZEN_PANE_HEADER_BORDER_COLOR */
-  --os-grid-border-color: light-dark(#e2e3e3, var(--os-gray-200));
+  --os-grid-border-color: light-dark(#e2e3e3, var(--os-gray-400));
```
hokolomopo, line 51: color is wrong in dark mode (double inversion for canvas I'm guessing) 

Image

hokolomopo (commented):
Spreadsheet is dark in light mode 🙃 

Image

pro-odoo:
@anhe-odoo Sorry, rebase due to chartpocalypse :/

demo/main.js
```diff
@@ -352,6 +359,10 @@ class Demo extends Component {
     o_spreadsheet.__DEBUG__.model = this.model;
     this.model.joinSession();
     this.activateFirstSheet();
+    const prefersDark = window.matchMedia("(prefers-color-scheme: dark)").matches;
```
anhe-odoo, line 362: @VincentSchippefilt : the colorScheme can be set directly in the model config

src/components/grid_add_rows_footer/grid_add_rows_footer.ts
```diff
@@ -35,6 +35,7 @@ export class GridAddRowsFooter extends Component<Props, SpreadsheetChildEnv> {
 
     return cssPropertiesToCss({
       top: `${top}px`,
+      ["z-index"]: "1",
```
VincentSchippefilt, line 39: I don't understand why we need a z-index in this PR?

src/helpers/color.ts
```diff
@@ -402,6 +402,60 @@ export function lightenColor(color: Color, percentage: number): Color {
...
+ * The dark-mode CSS filter chain applied to theme-dependent elements is:
+ *   filter: invert(1) hue-rotate(180deg) contrast(0.9) brightness(1.5)
+ *
+ * This function computes the exact inverse of that filter chain:
+ *   1. Reverse brightness(1.2)  ->  c / 1.2
+ *   2. Reverse contrast(0.85)    ->  (c − 0.5) / 0.85 + 0.5
+ *   3. Reverse hue-rotate(180°) ->  apply the CSS hue-rotate matrix (self-inverse)
+ *   4. Reverse invert(1)        ->  1 − c (self-inverse)
+ */
+export function adaptForDarkMode(displayedColor: Color): Color {
```
VincentSchippefilt, line 448: this is not very robust and already wrong: https://github.com/odoo/o-spreadsheet/pull/8045/changes#diff-9a88aa8fb8a14e6c3a40f2dbabbc5b495c154e90785c2768ff7c1f09c3f8c15cR298 is not what is in the comments. 
If it is possible, it would be better that both the CSS and this function rely on the same parameters, so that if we change one, the other is also changed.
If it isn't the same single source of truth, we need a test that proves that the values used here (or the result) is really the opposite of the actual style defined in spreadsheet.css

src/stores/grid_renderer_store.ts
```diff
@@ -235,8 +229,9 @@ export class GridRenderer extends DisposableStore {
     const { ctx } = renderingContext;
     for (const box of boxes) {
       const style = box.style;
-      if (style.fillColor && style.fillColor !== "#ffffff") {
-        ctx.fillStyle = style.fillColor || "#ffffff";
+      const fillColor = toHex(style.fillColor || "#ffffff");
+      if (fillColor !== "#FFFFFF") {
```
VincentSchippefilt, line 229: lowercase `ffffff !== FFFFFF` uppercase
I guess the default should be uppercase, as it is the one used everywhere else.
I guess the check could also be case insensitive to be more robust, as well as accept the short version (#FFF) as the exceptional case too.

tests/spreadsheet/spreadsheet_component.test.ts
```diff
@@ -492,8 +492,8 @@ test("components take the small screen into account", async () => {
 });
 
 test("Spreadsheet color scheme can be set to dark", async () => {
-  const model = new Model();
-  await mountSpreadsheet({ model, colorScheme: "dark" });
+  const model = new Model({}, { colorScheme: "dark" });
+  await mountSpreadsheet({ model });
```
VincentSchippefilt, line 496: we should probably also call the getter to know the theme, it's not tested anywhere that I can see

src/stores/grid_renderer_store.ts
VincentSchippefilt: tohex does the uppercase, sorry

VincentSchippefilt:
robodoo delegate+

anhe-odoo:
@robodoo r+

PR #8056, opened 2026-03-06
Master replace live server vsc

VincentSchippefilt:
upgrade dev infra and live server


tools/live-server.cjs
```diff
@@ -0,0 +1,155 @@
...
+    res.write("data: reload\n\n");
+  }
+}
+
+function serveFile(filePath, res) {
+  fs.readFile(filePath, (err, data) => {
+    if (err) {
+      res.writeHead(500);
+      res.end("Internal server error");
+      return;
```
LucasLefevre, line 83: `console.error(err);` to see in the logs what's happening

tools/live-server.cjs
```diff
@@ -0,0 +1,155 @@
...
+
+  let urlPath = req.url.split("?")[0];
+  if (urlPath.endsWith("/")) {
+    urlPath += "index.html";
+  }
+
+  // Prevent path traversal
+  const filePath = path.normalize(path.join(ROOT, urlPath));
+  if (!filePath.startsWith(ROOT + path.sep) && filePath !== ROOT) {
+    res.writeHead(403);
```
LucasLefevre, line 119: j'allais justement tester ça avant de voir que c'était géré ici :p

But we should probably restrict to the `demo` and `build` directories ? Here anyone on the local network could read any file under `o-spreadsheet/`

There's probably nothing sensitive, but who knows what's inside all the git-ignored directories

tools/live-server.cjs
```diff
@@ -0,0 +1,155 @@
...
+    });
+    res.write("\n");
+    sseClients.add(res);
+    req.on("close", () => sseClients.delete(res));
+    return;
+  }
+
+  let urlPath = req.url.split("?")[0];
+  if (urlPath.endsWith("/")) {
+    urlPath += "index.html";
```
rrahir, line 113: 🪄  we lose the ability to display a folder. Dont think we care that much though.

VincentSchippefilt:
robodoo rebase-ff

tools/live_server/main.cjs
```diff
@@ -0,0 +1,155 @@
...
+
+// Watch for file changes and notify browsers
+watch(WATCH_PATHS, { recursive: true }, (evt, name) => {
+  process.stdout.write(`[live-server] changed: ${path.relative(ROOT, name)}\n`);
+  notifyClients();
+});
+
+server.listen(PORT, () => {
+  const url = `http://localhost:${PORT}${OPEN_PATH}`;
+  process.stdout.write(`[live-server] http://localhost:${PORT} — opening ${url}\n`);
```
LucasLefevre: useless to have localhost twice

Image

```
process.stdout.write(`[live-server] opening ${url}\n`);
```

tools/live-server.cjs
```diff
@@ -0,0 +1,155 @@
...
+
+server.listen(PORT, () => {
+  const url = `http://localhost:${PORT}${OPEN_PATH}`;
+  process.stdout.write(`[live-server] http://localhost:${PORT} — opening ${url}\n`);
+  const open =
+    process.platform === "darwin"
+      ? `open "${url}"`
+      : process.platform === "win32"
+      ? `start "" "${url}"`
+      : `xdg-open "${url}"`;
```
rrahir, line 155: @hokolomopo does this work on wsl ? :)

tools/live-server.cjs
LucasLefevre: Oh well...everything is also available in master...

tools/live-server.cjs
```diff
@@ -0,0 +1,155 @@
...
+  ".xml": "application/xml",
+  ".png": "image/png",
+  ".jpg": "image/jpeg",
+  ".jpeg": "image/jpeg",
+  ".svg": "image/svg+xml",
+  ".ico": "image/x-icon",
+  ".woff": "font/woff",
+  ".woff2": "font/woff2",
+  ".ttf": "font/ttf",
+  ".eot": "application/vnd.ms-fontobject",
```
rrahir, line 46: don't know if we need to support fonts ATM but anyhow, some mimetypes are missing. I don't know if we want to be thorough or if we only want a small support.

tools/live-server.cjs
LucasLefevre, line 1: we don't need a directory with a single file.
`/live_server/main.cjs` -> `live_server.cjs`

tools/live-server.cjs
rrahir: probably better to whitelist specific files/folders but we also load stuff from node_modules ><

tools/live-server.cjs
rrahir: example missing: webp, gif, Opentype fonts (.otf)

tools/live-server.cjs
VincentSchippefilt: for me, it was more a pain to have the folder view than a help. I guess there is no loss

tools/live-server.cjs
VincentSchippefilt: on my linux machine, there is a windows 10 VM. 
On that windows VM there is wsl2
On that wsl2 it doesn't work :'(

tools/bundle_iife_watch.cjs
```diff
@@ -0,0 +1,58 @@
...
+    });
+    proc.on("exit", (code) => process.exit(code ?? 0));
+  });
+} else {
+  // Dev mode: initial build already done — only rebuild on subsequent JS changes
+  const WATCH_PATH = path.join(ROOT, "build/js/src");
+  let building = false;
+
+  function rebuild() {
+    if (building) return;
```
rrahir, line 42: SINGLE LINE IF STATEMENT !!

PR #8064, opened 2026-03-09
[IMP] border: default

fdamhaut:
## Description:

Task: [5990542](https://www.odoo.com/odoo/2328/tasks/5990542)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/clipboard_handlers/borders_clipboard.ts
+++ b/src/clipboard_handlers/borders_clipboard.ts
@@ -1,36 +1,86 @@
+import { splitZoneForPaste } from "../helpers/clipboard/clipboard_helpers";
+import { getItemId } from "../helpers/data_normalization";
+import { groupConsecutive, range } from "../helpers/misc";
 import { recomputeZones } from "../helpers/recompute_zones";
 import { positionToZone } from "../helpers/zones";
+import { BorderDescrInternal } from "../plugins/core/borders";
+import { defaultValue } from "../plugins/core/default";
 import { ClipboardCellData, ClipboardOptions, ClipboardPasteTarget } from "../types/clipboard";
-import { Border, CellPosition, HeaderIndex, UID, Zone } from "../types/misc";
+import {
+  Border,
+  BorderDescr,
+  BorderOrNull,
+  CellPosition,
+  Column,
+  HeaderIndex,
+  UID,
+  Zone,
+} from "../types/misc";
 import { AbstractCellClipboardHandler } from "./abstract_cell_clipboard_handler";
 
 type ClipboardContent = {
-  borders: (Border | null)[][];
+  content: {
+    left: HeaderIndex;
+    top: HeaderIndex;
+    bordersTop: Column<BorderDescrInternal>[];
+    bordersLeft: Column<BorderDescrInternal>[];
+    defaultTop: defaultValue<BorderDescrInternal>;
+    defaultLeft: defaultValue<BorderDescrInternal>;
+    height: number;
+    width: number;
+  }[];
+  height: number;
+  width: number;
 };
+
+function toDescr(
+  border: BorderDescrInternal | undefined,
+  undefinedIf?: "external" | "internal"
+): BorderDescr | undefined | null {
+  if (border?.style === "empty") {
+    return null;
+  }
+  if (!border || border.internal === undefinedIf) {
+    return undefined;
+  }
+  return { color: border.color, style: border.style };
+}
 
 export class BorderClipboardHandler extends AbstractCellClipboardHandler<
   ClipboardContent,
   Border | null
 > {
-  private queuedBordersToAdd: Record<string, Zone[]> = {};
-
   copy(data: ClipboardCellData): ClipboardContent | undefined {
     const sheetId = data.sheetId;
     if (data.zones.length === 0) {
       return;
     }
-    const { rowsIndexes, columnsIndexes } = data;
-    const borders: (Border | null)[][] = [];
+    const content: ClipboardContent["content"] = [];
+    let width = 0;
+    let height = 0;
 
-    for (const row of rowsIndexes) {
-      const bordersInRow: (Border | null)[] = [];
-      for (const col of columnsIndexes) {
-        const position = { col, row, sheetId };
-        bordersInRow.push(this.getters.getCellBorder(position));
+    let topIndex = 0;
+    for (const row of groupConsecutive(data.rowsIndexes)) {
+      const top = row[0];
+      const bottom = row[row.length - 1];
+      let leftIndex = 0;
+      for (const col of groupConsecutive(data.columnsIndexes)) {
+        const left = col[0];
+        const right = col[col.length - 1];
+        content.push({
+          left: leftIndex,
+          top: topIndex,
+          height: row.length,
+          width: col.length,
+          ...this.getters.getBorderClipboardData(sheetId, { left, right, top, bottom }),
+        });
+        leftIndex += col.length;
+        width = Math.max(leftIndex, width);
       }
-      borders.push(bordersInRow);
+      topIndex += row.length;
+      height = Math.max(topIndex, height);
     }
-    return { borders };
+    return { content, height, width };
   }
 
   paste(target: ClipboardPasteTarget, content: ClipboardContent, options: ClipboardOptions) {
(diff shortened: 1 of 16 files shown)


src/plugins/core/borders.ts
```diff
@@ -131,6 +311,408 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
...
+    const horizontalZone = this.getters.getRowsZone(sheetId, zone.top, zone.bottom);
+    const externalHorizontalZones = recomputeZones([horizontalZone], [zone]);
+    const defaults = this.getDefaultBorderInCell(borderType, sheetId, externalHorizontalZones, {
+      sheet: true,
+      row: true,
+    });
+    const verticalZone = this.getters.getColsZone(sheetId, zone.left, zone.right);
+    const externalVerticalZones = recomputeZones([verticalZone], [zone]);
+    defaults.push(
+      ...this.getDefaultBorderInCell(borderType, sheetId, externalVerticalZones, {
```
pro-odoo: Could lead to maximum stack (See points 1 & 2 in the pad. Same for L361, 516, 523

src/plugins/core/borders.ts
```diff
@@ -131,6 +311,408 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
...
+      .map((k) => parseInt(k, 10))
+      .filter((header) => zone.top <= header && header <= zone.bottom);
+    for (const row of rowOverlap) {
+      for (let col = zone.left; col <= zone.right; col++) {
+        this.history.update(borderKey, sheetId, col, row, EMPTY_BORDER);
+      }
+    }
+
+    const colOverlap = Object.keys(this[defaultKey][sheetId]?.colDefault ?? {})
+      .map(parseInt)
```
pro-odoo: You cannot do that, callbackFn of map function can have up to 3 arguments (current element, index of the element, and the array on which map is called).
`parseInt` takes two arguments, the number and the radix. So here you give the index of the element as the radix.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map#callbackfn
```js
console.log(["1", "2", "3", "4"].map(parseInt));
// [ 1, NaN, NaN, NaN ]
console.log(["1", "2", "3", "4"].map((n) => parseInt(n, 10));
// [ 1, 2, 3, 4 ]
```

In addition, the radix of `parseInt` is not always 10, it's derived from the given string (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt#description). It's considered as a best practice to always specify it.

pro-odoo: See point 3 in the pad to reproduce

src/helpers/misc.ts
```diff
@@ -806,6 +806,18 @@ export function getMissingHeadersForSpreadResult(
   return { missingRows, missingCols };
 }
 
+export function isObjectEmpty(obj: any) {
+  if (!obj || Object.keys(obj).length === 0) {
+    return true;
+  }
+  for (const value of Object.values(obj)) {
+    if (value !== undefined && (typeof value !== "object" || !isObjectEmpty(value))) {
```
pro-odoo: null value should be considered as non-empty IMO:
```js
expect(isObjectEmpty({ top: undefined })).toBe(true);
expect(isObjectEmpty({ top: null })).toBe(false);
// Currently, the sentence above crash
```

pro-odoo: I wrote two tests to highlight the issue:
```ts
function sheetOfSize(cols: number, rows: number): Model {
  const model = new Model();
  const originalSheetId = model.getters.getActiveSheetId();
  createSheet(model, { sheetId: "sh", cols, rows, activate: true });
  deleteSheet(model, originalSheetId);
  return model;
}

test("a row cleared inside a sheet-wide default survives export/import", () => {
    const model = sheetOfSize(26, 20);
    setBordersOnTarget(model, ["A1:Z20"], { left: THIN });
    clearFormatting(model, "A3:Z3");
    expect(getBorder(model, "C3")).toBeNull();

    const imported = new Model(model.exportData());

    expect(getBorder(imported, "C3")).toBeNull(); // <-- Boom
    expect(getBorder(imported, "C4")).toEqual({ left: THIN });
  });

  test("a column cleared inside a sheet-wide default survives export/import", () => {
    const model = sheetOfSize(26, 20);
    setBordersOnTarget(model, ["A1:Z20"], { top: THIN });
    clearFormatting(model, "C1:C20");
    expect(getBorder(model, "C5")).toBeNull();

    const imported = new Model(model.exportData());

    expect(getBorder(imported, "C5")).toBeNull(); // <-- Boom
    expect(getBorder(imported, "D5")).toEqual({ top: THIN });
  });
```

src/plugins/core/borders.ts
```diff
@@ -233,438 +784,625 @@ export class BordersPlugin extends CorePlugin<BordersPluginState> implements Bor
...
+    return null;
+  }
+
+  getBordersColors(sheetId: UID): Color[] {
+    const colors: Color[] = [];
+    const tops = Object.values(this.bordersTop[sheetId] ?? {})
+      .filter(isDefined)
+      .flatMap((r) => Object.values(r).filter(isDefined))
+      .map((b) => b.color);
+    colors.push(...tops);
```
pro-odoo: this getter can throw with a huge amount of borders:
```ts
test("getBordersColors does not throw on a heavily bordered sheet", () => {
    const model = sheetOfSize(26, 20000);
    expect(() => setZoneBorders(model, { position: "all" }, ["A1:M9999"])).not.toThrow();
    expect(() => model.getters.getBordersColors("sh")).not.toThrow();
  });
```

src/clipboard_handlers/borders_clipboard.ts
```diff
@@ -40,47 +78,126 @@ export class BorderClipboardHandler extends AbstractCellClipboardHandler<
...
-
-    this.executeQueuedChanges(sheetId);
   }
 
-  pasteZone(sheetId: UID, col: HeaderIndex, row: HeaderIndex, borders: (Border | null)[][]) {
-    for (const [r, rowBorders] of borders.entries()) {
-      for (const [c, originBorders] of rowBorders.entries()) {
-        const position = { col: col + c, row: row + r, sheetId };
-        this.pasteBorder(originBorders, position);
+  pasteContent(sheetId: UID, zone: Zone, contents: ClipboardContent["content"]) {
```
pro-odoo, line 103: This function generates a bunch of commands :/

```ts
test("paste generates a bunch of commands", () => {
    const model = sheetOfSize(120, 120);
    setZoneBorders(model, { position: "all" }, ["A1:AN40"]);
    copy(model, "A1:AN40");

    const borderCommands: string[] = [];
    model.on("command-dispatched", null, (cmd: any) => {
      if (cmd.type.includes("BORDER")) {
        borderCommands.push(cmd.type);
      }
    });

    paste(model, "A60");

    expect(getBorder(model, "A60")).toEqual(ALL);
    // The number is very high :/ 
    expect(borderCommands.length).toBeLessThan(100);
  });
```

pro-odoo (changes requested):
I'm a bit afraid of the performance regression :/ We will need to test it with big sheets...

pro-odoo:
With a simple spreadsheet with not that much border, copy-paster a zone (something like B6:N24) generates the following payload, in master and in your branch (+/- 4300 char in master vs +/- 120K). It's a too big regression to be ignore :D

IMO we should be able to dispatch one command by edge type, something like that

[master-border.json](https://github.com/user-attachments/files/31210557/master-border.json)
[master.json](https://github.com/user-attachments/files/31210558/master.json)
[Untitled spreadsheet.osheet.json](https://github.com/user-attachments/files/31210561/Untitled.spreadsheet.osheet.json)

src/helpers/misc.ts
```diff
@@ -806,6 +806,18 @@ export function getMissingHeadersForSpreadResult(
   return { missingRows, missingCols };
 }
 
+export function isObjectFalsyRecursively(obj: any) {
+  if (!obj || Object.keys(obj).length === 0) {
+    return true;
+  }
+  for (const value of Object.values(obj)) {
+    if (value !== undefined && (typeof value !== "object" || !isObjectFalsyRecursively(value))) {
```
anhe-odoo, line 814: Not sure with the "isObjectFalsy" name while you only test empty and undefined object (not `false` or `0` value for example). But it's really a nitpick so feel free to ignore

tests/sheet/sheet_manipulation_plugin.test.ts
```diff
@@ -1060,7 +1060,7 @@ describe("Rows", () => {
     });
   });
 
-  describe("Correctly update border and style", () => {
+  describe("Update border and style correctly", () => {
```
anhe-odoo, line 1063: I personally prefer the previous title but whatever :D

fdamhaut: It's just because there is another test with the same name (the old one) and having both with the same name was annoying when testing them individually

PR #8107, opened 2026-03-13
[IMP] chart: add total line on stacked bar charts

dhrp-odoo:
## Description:

Add a "Show totals" option for stacked bar and column charts.

When enabled, the chart renders a derived line dataset showing the sum of
visible source datasets for each category. Empty/non-numeric points are
ignored, and all-empty totals are skipped.

The total line follows the Odoo graph view styling and uses a line legend marker.

Task: [5886277](https://www.odoo.com/odoo/2328/tasks/5886277)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/components/figures/chart/chartJs/chartjs_show_values_plugin.ts
```diff
@@ -10,11 +10,21 @@ import { Color } from "../../../../types";
 export interface ChartShowValuesPluginOptions {
   type: AllChartType;
   showValues: boolean;
+  showTotals?: boolean;
+  stacked?: boolean;
```
hokolomopo: having a `stacked` argument just to use it to determine whether we should ignore `showTotals` argument feels wrong. I'd remove this argument, and just set showTotals to false when generating a bar chart w/o stacked

src/components/figures/chart/chartJs/chartjs_show_values_plugin.ts
```diff
@@ -296,3 +318,107 @@ function drawPieChartValues(
...
+  if (dataset.hidden || isTrendLineAxis(dataset.xAxisID)) {
+    return false;
+  }
+
+  if (options.type === "combo") {
+    return dataset.type === "bar";
+  }
+
+  return options.type === "bar" && !!options.stacked;
+}
```
hokolomopo: can't it be simplified to `dataset.hidden && dataset.type === "bar"` ? In both combo &  bar chart we only take the bar datasets (+option.stacked if we keep the argument)

src/components/figures/chart/chartJs/chartjs_show_values_plugin.ts
```diff
@@ -296,3 +318,107 @@ function drawPieChartValues(
     }
   }
 }
+
+function drawStackedBarChartTotals(
```
hokolomopo: can't you draw them in the same methods as the showValues ? (eg. drawLineOrBarOrRadarChartValues). Drawing them separelty means that we loop yet again on every data point, we don't use the same rendering logic when I'm guessing it should be mostly the same, and we're missing the logic so that the totals & show values don't overlap (see Chloe Adams in the screenshot)

Image

tests/figures/chart/charts_component.test.ts
```diff
@@ -2097,6 +2097,27 @@ describe("charts", () => {
...
 
+  test.each<ChartType>(["bar", "combo"])("showTotals checkbox updates the chart", async (type) => {
+    createTestChart(type);
+    if (type === "bar") {
+      updateChart(model, chartId, { stacked: true });
+    }
+    await mountChartSidePanel();
+    await openChartDesignSidePanel(model, env, fixture, chartId);
+
+    expect(document.querySelector("input[name='showTotals']")).toBeTruthy();
```
hokolomopo:
```suggestion
    expect(("input[name='showTotals']")).toHaveCount(1);
```

src/components/side_panel/chart/combo_chart/combo_chart_design_panel.ts
```diff
@@ -38,4 +40,10 @@ export class ComboChartDesignPanel extends GenericZoomableChartDesignPanel<
     }
     return dataSets[index].type ?? "line";
   }
+
+  get hasBarDataSets() {
+    return this.props.definition.dataSets.some(
+      (_dataSet, index) => this.getDataSeriesType(index) === "bar"
+    );
+  }
 }
```
hokolomopo: this isn't tested :)

hokolomopo (commented):
👋

src/components/side_panel/chart/combo_chart/combo_chart_design_panel.ts
dhrp-odoo: Good catch! added :+1:

src/components/figures/chart/chartJs/chartjs_show_values_plugin.ts
dhrp-odoo: I first split it out because totals are category-level labels, not point-level labels like `showValues`, and it made the code easier to read.

But your point makes sense: once totals also need to participate in `textsPositions` to avoid overlapping with value labels

src/components/figures/chart/chartJs/chartjs_show_values_plugin.ts
```diff
@@ -78,20 +100,37 @@ function drawTextWithBackground(text: string, x: number, y: number, ctx: CanvasR
...
 function drawLineOrBarOrRadarChartValues(
   chart: any,
   options: ChartShowValuesPluginOptions,
   ctx: CanvasRenderingContext2D
 ) {
   const yMax = chart.chartArea.bottom;
   const yMin = chart.chartArea.top;
+  const textHeight = globalThis.Chart?.defaults.font.size ?? 12; // ChartJS default text height
   const textsPositions: Record<number, number[]> = {};
+  const stackedTotalsByIndex = new Map<number, StackedTotalLabel>();
```
anhe-odoo: Do we really need a map, or an array will be sufficient ?

src/components/figures/chart/chartJs/chartjs_show_values_plugin.ts
```diff
@@ -136,6 +196,32 @@ function drawLineOrBarOrRadarChartValues(
       drawTextWithBackground(valueToDisplay, xPosition, yPosition, ctx);
     }
   }
+
+  if (!options.showTotals) {
```
anhe-odoo: Here I would go for a `drawLineOrBarOrRadarChartTotals` containing the following lines and simply do

```ts
if (options.showTotals) {
    drawLineOrBarOrRadarChartTotals(...);
}
```

It will be easier if we need to change something in the future in this part so that we won't have to read all of this method to see where the "if showTotals" is and deduct that the part we are looking for is under it. What do you think ?

src/components/figures/chart/chartJs/chartjs_show_values_plugin.ts
```diff
@@ -250,6 +357,33 @@ function drawHorizontalBarChartValues(
       drawTextWithBackground(displayValue, xPosition, yPosition, ctx);
     }
   }
+
+  if (!options.showTotals) {
```
anhe-odoo: Same remarks here, use another helper :)

src/components/side_panel/chart/combo_chart/combo_chart_design_panel.ts
```diff
@@ -3,6 +3,7 @@ import { CustomizableSeriesChartRuntime, UID } from "../../../../types";
 import { ComboChartDefinition } from "../../../../types/chart/combo_chart";
 import { RadioSelection } from "../../components/radio_selection/radio_selection";
 import { ChartShowDataMarkers } from "../building_blocks/show_data_markers/show_data_markers";
+import { ChartShowTotals } from "../building_blocks/show_totals/show_totals";
```
anhe-odoo: I don't think we can have stacked combo chart (I'm quite sure but maybe I missed something ?). If I'm right, we shouldn't have this "show totals" in the combo chart panel

src/helpers/figures/charts/runtime/chartjs_show_values.ts
```diff
@@ -20,10 +20,16 @@ export function getChartShowValues(
   args: ChartRuntimeGenerationArgs
 ): ChartShowValuesPluginOptions {
   const { axisFormats, locale } = args;
+  // show totals only for combo charts or stacked bar charts
+  const showTotals =
+    "showTotals" in definition &&
+    !!definition.showTotals &&
+    (definition.type === "combo" || (definition.type === "bar" && !!definition.stacked));
```
anhe-odoo: You can remove "combo" here if I"m right about the fact a combo chart can't be stacked :)

src/components/side_panel/chart/combo_chart/combo_chart_design_panel.ts
dhrp-odoo: Yes, i was kind of confused when discussed with FRGI, but indeed we should not have totals for combo as those are not stacked as of now :)

src/components/figures/chart/chartJs/chartjs_show_values_plugin.ts
dhrp-odoo: Agreed!

src/components/figures/chart/chartJs/chartjs_show_values_plugin.ts
dhrp-odoo: yes, array of objects can do the work!

src/components/figures/chart/chartJs/chartjs_show_values_plugin.ts
```diff
@@ -136,6 +196,40 @@ function drawLineOrBarOrRadarChartValues(
...
+
+function drawVerticalStackedBarChartTotals(
+  stackedTotals: StackedTotalLabel[],
+  textsPositions: Record<number, number[]>,
+  textHeight: number,
+  options: ChartShowValuesPluginOptions,
+  ctx: CanvasRenderingContext2D
+) {
+  // Totals may extend above/right of the plot area. Restoring removes the chartArea clip.
+  restoreContextForTotals(ctx);
```
hokolomopo: note: won't be needed after https://www.odoo.com/odoo/2328/tasks/5993132

src/helpers/figures/charts/runtime/chartjs_dataset.ts
```diff
@@ -113,9 +113,57 @@ export function getBarChartDatasets(
...
+    pointHitRadius: 10,
+    pointRadius: 3,
+    pointHoverRadius: 6,
+    borderWidth: 2,
+    backgroundColor: color,
+    borderColor: color,
+    pointBackgroundColor: color,
+    xAxisID: "x",
+    yAxisID: "y",
+  };
```
hokolomopo, line 162: nitpick: Try to use the same values as our lien datasets in line charts (eg. LINE_DATA_POINT_RADIUS), not sure you need to change the defaults for pointHoverRadius/pointHitRadius

hokolomopo (commented):
I really like the idea of having a separate dataset for the totals, it makes the code way easier :)

tests/figures/chart/bar_chart_plugin.test.ts
```diff
@@ -345,4 +347,171 @@ describe("bar chart", () => {
     expect(scales.y?.grid?.display).toBe(false);
     expect(scales.y?.grid?.minor?.display).toBe(true);
   });
+
+  describe("total line", () => {
+    beforeEach(() => {
+      // prettier-ignore
+      model = createModelFromGrid({
```
LucasLefevre: could you remove this `beforeEach` and setup the minimal data in each tests ?

see [keep-tests-minimal](https://github.com/odoo/o-spreadsheet/blob/19.0/.claude/skills/o-spreadsheet-testing/SKILL.md#keep-tests-minimal
) and [avoid-beforeeach](https://github.com/odoo/o-spreadsheet/blob/19.0/.claude/skills/o-spreadsheet-testing/SKILL.md#avoid-beforeeach)

LucasLefevre (approved):
LGTM 👍 

Except a small comment for tests

tests/figures/chart/bar_chart_plugin.test.ts
dhrp-odoo: Okay, I have simplified the tests :+1: 

I have one question though: is it still better to call `setCellContent` in each test (6 out of 7 tests), rather than doing it in `beforeEach`? I’m thinking from a LOC point of view.

image

(we will always need two datasets for the total line to be drawn)

LucasLefevre: Thanks!
Yes, IMO it's better to have minimal and self-contained tests, even if it's more LoC (the extra lines are very simple and readable)

LucasLefevre (approved):
robodoo r+

PR #8158, opened 2026-03-20
[FIX] evaluation: fix IF vectorization for errors

anhe-odoo:
## Description:

Currently, if the first arguments of an if is an array with an error or a non-boolean value, the IF returns a scalar error. This PR aims to change this behavior by modifing the vectorization of the compute ot the IF method.

## Related Task

- Task: [6009255](https://www.odoo.com/odoo/2328/tasks/6009255)

### DIFF
--- a/src/functions/create_compute_function.ts
+++ b/src/functions/create_compute_function.ts
@@ -1,11 +1,183 @@
 import { CellValue } from "../types/cells";
-import { BadExpressionError, EvaluationError } from "../types/errors";
+import { BadExpressionError, EvaluationError, NotAvailableError } from "../types/errors";
 
 import { _t } from "../translation";
 import { ComputeFunction, EvalContext, FunctionDescription } from "../types/functions";
 import { Arg, FunctionResultObject, isMatrix, Matrix } from "../types/misc";
 import { argTargeting } from "./arguments";
-import { applyVectorization, isEvaluationError, matrixForEach, matrixMap } from "./helpers";
+import { generateMatrix, isEvaluationError, matrixForEach, matrixMap } from "./helpers";
+
+type VectorArgType = "horizontal" | "vertical" | "matrix";
+
+/**
+ * Enables a formula function to accept matrix or vector inputs instead of simple value, computing results across multiple dimensions.
+ *
+ * ```
+ *                    /         |‾                 ‾| \        |‾                                                    ‾|
+ *                   |          | [A]               |  |       | compute(A, D, E), compute(A, D, F), compute(A, D, G) |
+ * applyVectorization| compute, | [B], D, [E, F, G] |  |  <=>  | compute(B, D, E), compute(B, D, F), compute(B, D, G) |
+ *                   |          | [C]               |  |       | compute(C, D, E), compute(C, D, F), compute(C, D, G) |
+ *                    \         |_                 _| /        |_                                                    _|
+ * ```
+ *
+ * By default, all arguments are vectorized. To control which arguments are vectorized,
+ * pass an `acceptToVectorize` boolean array of the same length as `args`:
+ * - `true`  enables vectorization for that argument
+ * - `false` disables vectorization for that argument
+ *
+ * For example, with `[true, true, false]` on previous example you get:
+ *
+ * ```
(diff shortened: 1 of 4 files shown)


src/functions/module_logical.ts
```diff
@@ -72,14 +73,20 @@ export const IF = {
   ],
   compute: function (logicalExpression: Arg, valueIfTrue: Arg, valueIfFalse: Arg) {
     if (isMultipleElementMatrix(logicalExpression)) {
-      return applyVectorization(IF.compute, [logicalExpression, valueIfTrue, valueIfFalse]);
+      return applyVectorization(IF_errorHandlingCompute.bind(this), [
```
hokolomopo: can `createErrorHandlingCompute` be called automatically inside `applyVectorization` ? The only time we don't use it is for IFNA/IFERROR, but I'm not sure if using it would change anything

anhe-odoo: I tried to make something as clean as possible using your idea in the second commit :) Let me know what you think

src/functions/create_compute_function.ts
```diff
@@ -1,11 +1,195 @@
...
+ *   as ranges and invoke this helper directly within your `compute` implementation.
+ */
+export function applyVectorization(
+  context: EvalContext,
+  descr: FunctionDescription,
+  name: string,
+  args: Arg[],
+  acceptToVectorize: boolean[] | undefined = undefined
+): FunctionResultObject | Matrix<FunctionResultObject> {
+  const formula = createErrorHandlingCompute(descr, name).bind(context);
```
LucasLefevre:
```suggestion
  const formula = createErrorHandlingCompute(descr, name);
```
and then instead of `formula(...args)` -> `formula.apply(context, args)`

That way, we avoid creating a new binded function every time and spreading the args

src/functions/create_compute_function.ts
```diff
@@ -1,11 +1,195 @@
...
+ * - If an argument is declared **scalar** (not `"range"`), it is vectorized.
+ * - If **all** arguments are declared **ranges**, no vectorization occurs.
+ *   - e.g. `SUM(A1:B2)` returns a 1×1 sum over the range.
+ *   - e.g. `COS(A1:B2)` over `A1:B2` returns a 2×2 element-wise result.
+ * - For special behaviors (e.g. the `IF` function), you may declare all arguments
+ *   as ranges and invoke this helper directly within your `compute` implementation.
+ */
+export function applyVectorization(
+  context: EvalContext,
+  descr: FunctionDescription,
```
LucasLefevre, line 50: because that what you give when it's called manually inside the computes
```suggestion
  descr: AddFunctionDescription,
```

src/functions/create_compute_function.ts
```diff
@@ -35,7 +219,7 @@ export function createComputeFunction(
     }
 
     return replaceErrorPlaceholderInResult(
-      applyVectorization(errorHandlingCompute.bind(this), args, acceptToVectorize)
+      applyVectorization(this, descr, descr.name, args, acceptToVectorize)
```
LucasLefevre: `replaceErrorPlaceholderInResult` should also be part of `applyVectorization` isn't it ?

LucasLefevre (commented):
Could you check the performance with this commit ?

Before, we were creating all the wrapper functions like`errorHandlingCompute`/`computeFunctionToObject` once per function (~300-400 times, don't know how many functions we have, but that a low number).

Now, new ones are created at each function call (which can be millions)

src/functions/create_compute_function.ts
anhe-odoo: That would mean manually calling it in before every returns in `applyVectorization` no ? I'm not sure then, it's simpler to have only this call here

LucasLefevre:
no perf changes (if not slightly faster)

```
evaluate all cells
  before:  Mean: 606.19 ms, StdErr: 7.62 ms, n=15
  after:     Mean: 597.46 ms, StdErr: 7.25 ms, n=15 (-1%)
```

LucasLefevre:
robodoo r+

PR #8196, opened 2026-03-25
[IMP] Allows sorting filter menu in readonly mode

matho-odoo:
Sorting a table should be possible even in readonly mode

Task: [6047484](https://www.odoo.com/odoo/2328/tasks/6047484)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/doc/extending/command.md
+++ b/doc/extending/command.md
@@ -62,15 +62,7 @@ coreTypes.add("MY_COMMAND_NAME");
 
 ### Read-Only Mode
 
-In read-only mode, all core commands are cancelled with the `CommandResult` `Readonly` since the spreadsheet state cannot be modified.
-However, some locale commands still need to be executed, such as updating the active sheet.
-To allow a new local command in read-only mode, add its type to `readonlyAllowedCommands`:
-
-```ts
-import { readonlyAllowedCommands } from "@odoo/o-spreadsheet";
-
-readonlyAllowedCommands.add("MY_COMMAND_NAME");
-```
+All commands are allowed in readonly mode. Because the spreadsheet’s state cannot be changed, a specific transport service (ReadonlyTransportFilter) is used to prevent revisions from being sent to the server.
 
 ## Reserved keywords in commands
 
--- a/src/components/grid/grid.ts
+++ b/src/components/grid/grid.ts
@@ -251,203 +251,287 @@ export class Grid extends Component<SpreadsheetChildEnv> {
 
   // this map will handle most of the actions that should happen on key down. The arrow keys are managed in the key
   // down itself
-  private keyDownMapping: { [key: string]: Function } = {
-    Enter: () => this.editOrMoveInSelection("down"),
-    "Shift+Enter": () => this.editOrMoveInSelection("up"),
-    Tab: () => this.moveInSelection("right"),
-    "Shift+Tab": () => this.moveInSelection("left"),
-    F2: () => {
-      this.focusComposerFromActiveCell();
+  private keyDownMapping: { [key: string]: { fn: () => void; allowInReadonly: boolean } } = {
+    Enter: { fn: () => this.editOrMoveInSelection("down"), allowInReadonly: false },
+    "Shift+Enter": { fn: () => this.editOrMoveInSelection("up"), allowInReadonly: false },
+    Tab: { fn: () => this.moveInSelection("right"), allowInReadonly: true },
+    "Shift+Tab": { fn: () => this.moveInSelection("left"), allowInReadonly: true },
+    F2: {
+      fn: () => {
+        this.focusComposerFromActiveCell();
+      },
+      allowInReadonly: false,
     },
-    Delete: () => {
-      this.env.model.dispatch("DELETE_UNFILTERED_CONTENT", {
-        sheetId: this.env.model.getters.getActiveSheetId(),
-        target: this.env.model.getters.getSelectedZones(),
-      });
+    Delete: {
+      fn: () => {
+        this.env.model.dispatch("DELETE_UNFILTERED_CONTENT", {
+          sheetId: this.env.model.getters.getActiveSheetId(),
+          target: this.env.model.getters.getSelectedZones(),
+        });
+      },
+      allowInReadonly: false,
     },
-    Backspace: () => {
-      this.env.model.dispatch("DELETE_UNFILTERED_CONTENT", {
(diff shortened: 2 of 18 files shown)


doc/extending/command.md
```diff
@@ -42,15 +42,7 @@ coreTypes.add("MY_COMMAND_NAME");
...
 ### Read-Only Mode
 
-In read-only mode, all core commands are cancelled with the `CommandResult` `Readonly` since the spreadsheet state cannot be modified.
-However, some locale commands still need to be executed, such as updating the active sheet.
-To allow a new local command in read-only mode, add its type to `readonlyAllowedCommands`:
-
-```ts
-import { readonlyAllowedCommands } from "@odoo/o-spreadsheet";
-
-readonlyAllowedCommands.add("MY_COMMAND_NAME");
```
hokolomopo, line 72: readonlyAllowedCommands can be totally deleted then, no ?

hokolomopo: also need a PR in odoo to remove commands from readonlyAllowedCommands

hokolomopo (commented):
👋

doc/extending/command.md
```diff
@@ -42,15 +42,7 @@ coreTypes.add("MY_COMMAND_NAME");
...
-```
+All commands are allowed in readonly mode. Since the spreadsheet state cannot be modified , we use a specific transport service (ReadonlyTransportFilter) blocking revision to be send to the server.
```
rrahir:
```suggestion
All commands are allowed in readonly mode. Because the spreadsheet’s state cannot be changed, a specific transport service (ReadonlyTransportFilter) is used to prevent revisions from being sent to the server.
```

src/components/grid/grid.ts
```diff
@@ -446,6 +446,28 @@ export class Grid extends Component<Props, SpreadsheetChildEnv> {
...
+    "F4",
+    "Alt+Enter",
+    "Ctrl+Home",
+    "Ctrl+End",
+    "Shift+ ",
+    "Ctrl+ ",
+    "Ctrl+H",
+    "Ctrl+F",
+    "Ctrl+Shift+ ",
+    "Alt+Shift+ArrowRight",
```
rrahir: these are whitelisted but don't seem do do anything, or they open a menu in which every element is greyed out and therefore cannot be used

src/components/grid/grid.ts
```diff
@@ -446,6 +446,28 @@ export class Grid extends Component<Props, SpreadsheetChildEnv> {
...
+    "Ctrl+Z",
+    "Ctrl+Y",
+    "F4",
+    "Alt+Enter",
+    "Ctrl+Home",
+    "Ctrl+End",
+    "Shift+ ",
+    "Ctrl+ ",
+    "Ctrl+H",
+    "Ctrl+F",
```
rrahir: all sheet navigation shortcuts (arrowLeft/right/up/down) and their variants with Ctrl and Shift should be whitelisted as well

rrahir (changes requested):

VincentSchippefilt:
the commit **[IMP] Allows sorting filter menu in readonly mode** doesn't do that at all: it is done in the scope of that task, but it changes a bunch of fondamental stuff. The commit should explain what and why we do those changes. 
I guess the tests that actually test that we can use the sorting filter menu in readonly mode can stay in this commit

.claude/settings.local.json
```diff
@@ -0,0 +1,5 @@
+{
+  "permissions": {
+    "allow": ["Bash(gh pr *)"]
```
rrahir: 🙅‍♂️

tests/lock_sheet/lock_sheet_plugin.test.ts
```diff
@@ -52,8 +47,9 @@ describe("Lock Sheet plugin", () => {
     }
   });
 
-  test("read only commands bypass lock in dashboard mode", () => {
-    for (const cmdType of readonlyCommands) {
+  test.each(["UPDATE_CHART", "UPDATE_PIVOT"])(
```
rrahir, line 50: Don't think the test makes much sense anymore. it's even weird. if the sheet is locked, a user should not be able  to update a chart located in that sheet? @VincentSchippefilt

src/components/grid/grid.ts
```diff
@@ -441,6 +441,44 @@ export class Grid extends Component<SpreadsheetChildEnv> {
     "Alt+Shift+ArrowDown": () => this.processHeaderGroupingKey("down"),
   };
 
+  private readonlyAllowedShortcuts = new Set([
```
rrahir: are there any commands that are not contained here ? I removed the condition below and every test passes.

it looks like we could allow pretty much everything and add the readonly conditions in the elements called by those shortcuts.

src/components/headers_overlay/unhide_headers.ts
```diff
@@ -84,6 +87,9 @@ export class UnhideColumnHeaders extends Component<SpreadsheetChildEnv> {
   }
 
   unhide(hiddenElements: HeaderIndex[]) {
+    if (this.env.model.getters.isReadonly()) {
```
rrahir: Don't know what we aim for, but if we want to prevent to unhide elements, we should do the same for the groups (collapse/uncollapse). imho, we should let people unhide whatever they want

tests/collaborative/collaborative.test.ts
```diff
@@ -603,11 +603,6 @@ describe("Multi users synchronisation", () => {
       (user) => getCellContent(user, "A1"),
       "hello"
     );
-    setCellContent(david, "A1", "I'm David and I want access !");
```
rrahir, line 638: this element was relevant. Readonly users can no longer dispatch to the network but the change still affects them. it should be tested in another dedicated test

rrahir (changes requested):

tests/lock_sheet/lock_sheet_plugin.test.ts
matho-odoo: In dashboard mode you can update a chart by changing is granularity for example.

UPDATE_CHART and UPDATE_PIVOT were the only commands readonlyCommands here so my changes did not change the test.

src/components/grid/grid.ts
matho-odoo: I added a few tests that cannot pass without the condition (for exemple delete in readonly).

PR #8203, opened 2026-03-26
[FIX] format: number format with no digit

hokolomopo:
## Description

Since e8b38a a format with no date part and no digit is (correctly) detected as a number format. But the implementation of a format with no digit applied to a non-zero number was not correct. `123` formatted with `[$str]` was formatted as `str123` instead of `str`.

Task: [6068824](https://www.odoo.com/odoo/2328/tasks/6068824)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/tests/formats/format_helpers.test.ts
+++ b/tests/formats/format_helpers.test.ts
@@ -409,6 +409,14 @@ describe("formatValue on number", () => {
     expect(() => formatValue(1234, { format: "# @", locale })).toThrow();
     expect(() => formatValue(1234, { format: "@0", locale })).toThrow();
   });
+
+  test("Format with no digit in them", () => {
+    expect(formatValue(200, { format: "[$Positive]", locale })).toBe("Positive");
+    expect(formatValue(-123, { format: '0;,"Negative"', locale })).toBe("Negative");
+    expect(formatValue(0, { format: '0;0;"Zero"', locale })).toBe("Zero");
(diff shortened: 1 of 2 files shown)


tests/formats/format_helpers.test.ts
```diff
@@ -409,6 +409,14 @@ describe("formatValue on number", () => {
     expect(() => formatValue(1234, { format: "# @", locale })).toThrow();
     expect(() => formatValue(1234, { format: "@0", locale })).toThrow();
   });
+
+  test("Format with no digit in them", () => {
+    expect(formatValue(200, { format: "[$Positive]", locale })).toBe("Positive");
+    expect(formatValue(-123, { format: '0;,"Negative"', locale })).toBe("Negative");
+    expect(formatValue(0, { format: '0;0;"Zero"', locale })).toBe("Zero");
+    expect(formatValue("test", { format: '"Smile"', locale })).toBe("test");
+    expect(formatValue("test", { format: '@"Smile"', locale })).toBe("testSmile");
```
LucasLefevre, line 418: c'est marrant ça

hokolomopo: Yeah first one is a number format applied to a text, so it does nothing, second one is a text format applied to a text. Looks a bit strange indeed :sweat_smile:

LucasLefevre (approved):
robodoo r+

PR #8273, opened 2026-04-01
[IMP] node: add script to run the model in node

hokolomopo:
## Description:

This commit adds the `model-in-node` node script that runs the
spreadsheet model in a node environment.

Task: [6088515](https://www.odoo.com/odoo/2328/tasks/6088515)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
     "postdist": " npm run bundle:xml -- --outDir dist && npm run bundle:css -- --out dist",
     "test": "tsc --noEmit --project tsconfig.jest.json && jest",
     "monkey": "SPREADSHEET_MONKEY_COUNT=$npm_config_monkey_count jest 'tests/collaborative/collaborative_monkey_party.test.ts'",
+    "model-in-node": "npm run dist && node tools/run_in_node.cjs",
     "zipXlsx": "node tools/bundle_xlsx/zip_xlsx_demo.cjs",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -42,6 +42,7 @@
     "postdist": " npm run bundle:xml -- --outDir dist && npm run bundle:css -- --out dist",
     "test": "tsc --noEmit --project tsconfig.jest.json && jest",
     "monkey": "SPREADSHEET_MONKEY_COUNT=$npm_config_monkey_count jest 'tests/collaborative/collaborative_monkey_party.test.ts'",
+    "model-in-node": "npm run dist && node tools/run_in_node.cjs",
```
hokolomopo, line 45: Add this to the runbot config, and we should be good ?

LucasLefevre: mouais... not sure it's worth bothering runbot team for that. Not sure either what is their "per version" config for o-spreadsheet, of if it's a single config for all versions

Another way to do that is to run a jest test suite with another config without jsdom. I don't know if it's easy or a lot of setup.

(I still stand my point that we should have `npm run runbot` so that we control what it does, per version)

rrahir: Runbot team will still require some control over the parameters like the # of workers. 1 parameter is easy, 2+ not so much, maybe with npm-run-all and a pre-defined set or args

rrahir:
robodoo r+

PR #8275, opened 2026-04-01
[FIX] localize and canonicalize content

matho-odoo:
Task: [6089166](https://www.odoo.com/odoo/2328/tasks/6089166)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/composer/composer/abstract_composer_store.ts
+++ b/src/components/composer/composer/abstract_composer_store.ts
@@ -13,7 +13,6 @@ import {
   splitReference,
   zoneToDimension,
 } from "../../../helpers/index";
-import { canonicalizeNumberContent } from "../../../helpers/locale";
 import { cycleFixedReference } from "../../../helpers/reference_type";
 import {
   AutoCompleteProvider,
@@ -98,6 +97,7 @@ export abstract class AbstractComposerStore extends SpreadsheetStore {
   }
   protected abstract confirmEdition(content: string): void;
   protected abstract getComposerContent(position: CellPosition): string;
+  protected abstract getCurrentCanonicalContent(): string;
 
   abstract stopEdition(direction?: Direction): void;
 
@@ -361,10 +361,6 @@ export abstract class AbstractComposerStore extends SpreadsheetStore {
       }
       this.confirmEdition(content);
     }
-  }
-
-  protected getCurrentCanonicalContent(): string {
-    return canonicalizeNumberContent(this._currentContent, this.getters.getLocale());
   }
 
   protected cancelEditionAndActivateSheet() {
--- a/src/components/composer/composer/abstract_composer_store.ts
+++ b/src/components/composer/composer/abstract_composer_store.ts
@@ -97,6 +97,7 @@ export abstract class AbstractComposerStore extends SpreadsheetStore {
   }
   protected abstract confirmEdition(content: string): void;
   protected abstract getComposerContent(position: CellPosition): string;
+  // We need to use different canonicalize method in the standalone composer and the cell composer where dates should not be converted (they are not canonicals on UPDATE_CELL)
   protected abstract getCurrentCanonicalContent(): string;
 
   abstract stopEdition(direction?: Direction): void;
--- a/src/components/composer/composer/cell_composer_store.ts
+++ b/src/components/composer/composer/cell_composer_store.ts
@@ -10,7 +10,11 @@ import {
   updateSelectionOnDeletion,
   updateSelectionOnInsertion,
 } from "../../../helpers/index";
-import { getDateTimeFormat, localizeFormula } from "../../../helpers/locale";
+import {
+  canonicalizeNumberContent,
+  getDateTimeFormat,
+  localizeFormula,
+} from "../../../helpers/locale";
 import { dataValidationEvaluatorRegistry } from "../../../registries/data_validation_registry";
 import { _t } from "../../../translation";
 import {
@@ -295,5 +299,9 @@ export class CellComposerStore extends AbstractComposerStore {
       return false;
     }
     return true;
+  }
+
+  protected getCurrentCanonicalContent(): string {
+    return canonicalizeNumberContent(this._currentContent, this.getters.getLocale());
   }
 }
--- a/src/components/side_panel/conditional_formatting/cf_editor/cf_editor.ts
+++ b/src/components/side_panel/conditional_formatting/cf_editor/cf_editor.ts
@@ -13,7 +13,6 @@ import {
   isColorValid,
   rangeReference,
 } from "../../../../helpers";
-import { canonicalizeCFRule } from "../../../../helpers/locale";
 import { cycleFixedReference } from "../../../../helpers/reference_type";
 import { _t } from "../../../../translation";
 import {
@@ -252,11 +251,10 @@ export class ConditionalFormattingEditor extends Component<Props, SpreadsheetChi
       return [CommandResult.InvalidRange];
     }
     const sheetId = this.env.model.getters.getActiveSheetId();
-    const locale = this.env.model.getters.getLocale();
     const rule = newCf.rule || this.getEditedRule(this.state.currentCFType);
     const result = this.env.model.dispatch("ADD_CONDITIONAL_FORMAT", {
       cf: {
-        rule: canonicalizeCFRule(rule, locale),
+        rule,
         id: this.props.editedCf.id,
       },
       ranges: ranges.map((xc) => this.env.model.getters.getRangeDataFromXc(sheetId, xc)),
--- a/src/components/side_panel/conditional_formatting/cf_editor/cf_editor.ts
+++ b/src/components/side_panel/conditional_formatting/cf_editor/cf_editor.ts
@@ -13,6 +13,7 @@ import {
   isColorValid,
   rangeReference,
 } from "../../../../helpers";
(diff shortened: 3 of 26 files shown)


src/components/composer/composer/abstract_composer_store.ts
```diff
@@ -13,7 +13,6 @@ import {
   splitReference,
```
hokolomopo, line 13: commit title is wrongly formatted (should be [FIX]composer : machin) or something like that

missing commit message

tests/data_validation/data_validation_core_plugin.test.ts
```diff
@@ -227,16 +227,12 @@ describe("Data validation", () => {
...
-    "Deleting content of a cell doesn't remove  %s data validation rule",
-    async (criterion) => {
-      //@ts-ignore
-      addDataValidation(model, "A1", "id", criterion);
-      expect(model.getters.getDataValidationRules(sheetId)).toHaveLength(1);
-      deleteContent(model, ["A1"]);
-      expect(model.getters.getDataValidationRules(sheetId)).toHaveLength(1);
-    }
-  );
+  ])("Deleting content of a cell doesn't remove  %s data validation rule", async (criterion) => {
```
hokolomopo:
```suggestion
  ])("Deleting content of a cell doesn't remove  %s data validation rule",(criterion) => {
```
while you're changing the test, the async is useless

hokolomopo (commented):
Commit format is wrong, otherwise LGTM :)

src/components/composer/composer/abstract_composer_store.ts
LucasLefevre, line 1: include in the commit message the clear steps to reproduce the issue.

src/components/composer/composer/abstract_composer_store.ts
```diff
@@ -361,10 +361,6 @@ export abstract class AbstractComposerStore extends SpreadsheetStore {
     }
   }
 
-  protected getCurrentCanonicalContent(): string {
-    return canonicalizeNumberContent(this._currentContent, this.getters.getLocale());
```
LucasLefevre, line 365: why remove it from here and copy the exact same code in the children ?

src/components/side_panel/conditional_formatting/cf_editor/cf_editor.ts
```diff
@@ -252,11 +251,10 @@ export class ConditionalFormattingEditor extends Component<Props, SpreadsheetChi
       return [CommandResult.InvalidRange];
     }
     const sheetId = this.env.model.getters.getActiveSheetId();
-    const locale = this.env.model.getters.getLocale();
     const rule = newCf.rule || this.getEditedRule(this.state.currentCFType);
     const result = this.env.model.dispatch("ADD_CONDITIONAL_FORMAT", {
       cf: {
-        rule: canonicalizeCFRule(rule, locale),
```
LucasLefevre, line 259: `canonicalizeCFRule` is not used anymore

src/components/side_panel/data_validation/data_validation_panel.xml
```diff
@@ -4,18 +4,15 @@
       <t t-if="state.mode === 'list'">
         <div class="o-dv-preview-list">
           <t t-foreach="validationRules" t-as="rule" t-key="rule.id">
-            <DataValidationPreview
-              rule="localizeDVRule(rule)"
```
LucasLefevre, line 8: `localizeDVRule` is not used anymore.

And `localizeDataValidationRule` is still used in the "Dropdown list" menu item, but I think it's wrong now.

src/registries/data_validation_registry.ts
```diff
@@ -480,7 +506,11 @@ dataValidationEvaluatorRegistry.add("isGreaterThan", {
   criterionValueErrorString: DVTerms.CriterionError.numberValue,
   numberOfValues: () => 1,
   name: _t("Is greater than"),
-  getPreview: (criterion) => _t("Value is greater than %s", criterion.values[0]),
+  getPreview: (criterion: IsGreaterThanCriterion, getters: Getters) => {
```
LucasLefevre, line 509: I think I would call `localizeDataValidationRule` in `get descriptionString()`. It would be done once, instead of managed in each `getPreview`, which adds extra noise and a risk of forgetting

matho-odoo: After discussing it, we decided to leave it as is.

src/components/composer/composer/abstract_composer_store.ts
matho-odoo: One children use "canonicalizeContent" and the other "canonicalizeNumberContent"

src/components/composer/composer/cell_composer_store.ts
```diff
@@ -296,4 +300,8 @@ export class CellComposerStore extends AbstractComposerStore {
     }
     return true;
   }
+
+  protected getCurrentCanonicalContent(): string {
+    return canonicalizeNumberContent(this._currentContent, this.getters.getLocale());
```
anhe-odoo, line 305: Maybe add a comment to explain why you used two different canonicalize method (here and in standalone_composer_store)

src/components/composer/composer/abstract_composer_store.ts
LucasLefevre: it's very weird and it's impossible to understand and know why by reading the code.

I found out it's because of dates which are *not* canonical on UPDATE_CELL (and then core detects the format).
If we want to be consistent and to always use canonical content on the wire, shouldn't we change that ?

At the very least, it deserves a comment.

`canonicalizeNumberContent` and `canonicalizeContent` should not both exist, only one them. It's very confusing, and you don't know which one should be used. But that's a refactoring for master.

src/registries/auto_completes/data_validation_auto_complete.ts
```diff
@@ -20,7 +21,9 @@ autoCompleteProviders.add("dataValidation", {
     }));
   },
   selectProposal(tokenAtCursor, value) {
-    this.composer.setCurrentContent(value);
+    const locale = this.getters.getLocale();
+    const canonicalizedValue = canonicalizeContent(value, locale);
+    this.composer.setCurrentContent(canonicalizedValue);
```
LucasLefevre: `composer.setCurrentContent` is supposed to take localized content, since it later takes `this._currentContent` and run `canonicalizeNumberContent` on it

rrahir: Think it's what it does ^^

matho-odoo: I change it to localizedValue, Lucas is right :)
(and on top of that, value was already canonical)

rrahir: I read the sentence backwards, I think ><

LucasLefevre:
testing Claude review skill:

### Commits

`7c6ed6dd7` removes `.map(canonicalizeContent…)` but leaves the preceding `.filter(...)` without a trailing `;` (relying on ASI); `ef2dd7a63` adds it back. Cosmetic — not a blocker, but ideally squashed.

`ef2dd7a63` bundles unrelated `pivot_composer_helpers.ts` / `auto_complete_registry.ts` typing changes (adding `getters` to the `this` type) alongside CF localization work. Defensible since it's required by `data_validation_auto_complete.ts` in the same commit.

### Findings

#### src/registries/data_validation_registry.ts:747 — nit
`getNumberCriterionlocalizedValues` has a casing typo (`localizedValues` → `LocalizedValues`) and the "Number" name is misleading — it's also used for `isValueInList`, which stores dates/text too. Rename to `getCriterionLocalizedValues`.

#### src/registries/data_validation_registry.ts:147 — nit
`getPreview: (criterion: TextContainsCriterion, …)` is used for the `textIs` entry; the inner field type should be `TextIsCriterion`. Same shape, but the wrong-name annotation is confusing.

#### src/registries/data_validation_registry.ts:109,131,153 — nit
`criterion.values[0]?.toString()` — `values` is `string[]`, so both `?.` and `.toString()` are redundant. Just use `criterion.values[0]`.

#### src/registries/data_validation_registry.ts:685,696 — warning
`isValueInList` preview/error now uses `locale.formulaArgSeparator` as separator (e.g. `;` in FR). Display only, but it's a UX change from the old hardcoded `", "` — worth a one-line comment or confirming it matches the spec for the task.

#### src/components/side_panel/data_validation/dv_criterion_form/dv_input/dv_input.ts:63,70 — nit
`localizeContent(...) ?? ""` — `localizeContent` is typed `(string, Locale) => string` and never returns nullish. Drop the fallback or change the helper's contract.

#### src/components/side_panel/data_validation/dv_criterion_form/dv_input/dv_input.ts:60-65 — warning
`onWillUpdateProps` overwrites `state.textInput` whenever the parent's `props.value` changes. If the parent re-renders for any reason while the user has uncommitted input in the field, their text is lost. In practice the parent only re-renders the value on confirm, so it's fine today — but worth gating on `nextProps.value !== canonicalizeContent(this.state.textInput, locale)` to be safe.

#### tests — coverage gap (warning)
No test exercises the CF editor color-scale/icon-set localized input path added in `ef2dd7a63` (typing `1,5` in FR locale should canonicalize to `1.5` in the rule). Worth a focused component test.

rrahir:
dropped in favor of https://github.com/odoo/o-spreadsheet/pull/8773

PR #8276, opened 2026-04-01
[IMP] scrollbar: take whole available size with frozen panes

hokolomopo:
## Description

When there are frozen panes, the scrollbars are offset to start at the start of the main viewport. But there is no element on the side of the frozen panes, so components like Autofill and Table Resizer can overflow from the grid with nothing to cover them.

Task: [5891719](https://www.odoo.com/odoo/2328/tasks/5891719)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/scrollbar/scrollbar_horizontal.ts
+++ b/src/components/scrollbar/scrollbar_horizontal.ts
@@ -1,5 +1,6 @@
 import { Component, xml } from "@odoo/owl";
 import { SpreadsheetChildEnv } from "../../types/spreadsheet_env";
+import { cssPropertiesToCss } from "../helpers";
 import { isBrowserFirefox } from "../helpers/dom_helpers";
 import { ScrollBar } from "./scrollbar";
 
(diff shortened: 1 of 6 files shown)


src/components/scrollbar/scrollbar_horizontal.ts
```diff
@@ -40,12 +43,21 @@ export class HorizontalScrollBar extends Component<Props, SpreadsheetChildEnv> {
     return xRatio < 1;
   }
 
+  get containerStyle() {
+    const scrollbarWidth = this.env.model.getters.getScrollBarWidth();
+    return cssPropertiesToCss({
+      left: `${this.props.leftOffset}px`,
+      bottom: "0px",
+      height: `${scrollbarWidth}px`,
+      right: isBrowserFirefox() ? `${scrollbarWidth}px` : "0",
```
rrahir, line 52: dont thnk that's necessary in the container, same for the vertical component

PR #8318, opened 2026-04-07
[IMP] perf_profile: add performance analysis tool

LucasLefevre:
## Description:

Task: [6096737](https://www.odoo.com/odoo/2328/tasks/6096737)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/components/side_panel/perf_profile/perf_profile_panel.ts
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  selectEntry(index: number) {
+    this.state.selectedIndex = index;
+    const entry = this.perfProfile?.entries[index];
+    if (!entry) {
+      return;
+    }
+    const activeSheetId = this.env.model.getters.getActiveSheetId();
+    if (entry.range.sheetId !== activeSheetId) {
```
rrahir, line 70: if the range disappears ( sheet deleted or range invalid because the rows,cols were removed), it displays # REF and we can click itn it will try to move our selection in the invalid zone but it makes no sense

src/components/side_panel/perf_profile/perf_profile_panel.css
```diff
@@ -0,0 +1,14 @@
+.o-spreadsheet {
+  .o-perf-range-entry {
+    cursor: pointer;
+  }
+  .o-perf-bar {
+    background-color: color-mix(in srgb, var(--os-selection-border-color) 25%, transparent);
```
rrahir: I would have gone with a more "odoo-like" color than the blue personnally ^^

src/components/side_panel/perf_profile/perf_profile_panel.ts
```diff
@@ -0,0 +1,110 @@
...
+      this.env.model.dispatch("ACTIVATE_SHEET", {
+        sheetIdFrom: activeSheetId,
+        sheetIdTo: entry.range.sheetId,
+      });
+    }
+    const zone = entry.range.zone;
+    this.env.model.selection.selectCell(zone.right, zone.bottom); // Select the bottom right cell of the range to move the viewport there
```
rrahir: I don't understand why you don't just select the cell

src/components/side_panel/perf_profile/perf_profile_panel.ts
```diff
@@ -0,0 +1,110 @@
...
+    return [{ range: entry.range, color: HIGHLIGHT_COLOR, noFill: true }];
+  }
+
+  stringifyRange({ range }: RangeTiming) {
+    return this.env.model.getters.getRangeString(range, "forceSheetReference");
+  }
+
+  startProfiling() {
+    this.state.selectedIndex = undefined;
+    this.env.model.dispatch("EVALUATE_CELLS", { profiling: true });
```
rrahir, line 55: this probably should be throttled, either in the interface ,the model or both ? because right now I can spam the button and it will trigger tons of evaluations

src/components/side_panel/perf_profile/perf_profile_panel.xml
```diff
@@ -0,0 +1,78 @@
...
+                  class="o-perf-range-entry position-relative overflow-hidden rounded"
+                  t-att-class="{ 'o-perf-entry-selected': this.isSelected(entry_index) }"
+                  t-on-click="() => this.selectEntry(entry_index)">
+                  <div
+                    class="o-perf-bar position-absolute top-0 start-0 bottom-0"
+                    t-attf-style="width: {{ this.barWidth(entry) }}%"
+                  />
+                  <div
+                    class="position-relative d-flex justify-content-between py-1 px-2 text-nowrap">
+                    <span class="text-truncate me-2">
```
rrahir, line 59: nit: maybe a title with the full reference ?

rrahir (changes requested):

src/components/side_panel/perf_profile/perf_profile_panel.css
LucasLefevre: agreed, but I'm not a designer. Claude did this for me

LucasLefevre: Same for the highlight color btw, I'm totally open to better suggestions

LucasLefevre: Changed to use the green-ish odoo color
image
image

src/components/side_panel/perf_profile/perf_profile_panel.xml
LucasLefevre: Not sure it would be really helpful and it add noise to the UI

src/components/side_panel/perf_profile/perf_profile_panel.ts
LucasLefevre: Not super straight forward because the events are queued in the event queue and by the time `startProfiling` is called, the previous evaluation is already done and multiple seconds could have passed (-> debounce would be useless)
J'ai bricolé un truc, but it's a nightmare to test (one reason being jsdom does not use the same time reference as real browsers for event.timeStamp and performance.now())

src/components/side_panel/perf_profile/perf_profile_panel.ts
LucasLefevre: improved the comment. Let me know if it's better

rrahir: 👍  sorry I did not get the intention. I still have a question, if the computed cell is actually a spill, shouldn't we consider the zone of the spill?

src/components/side_panel/perf_profile/perf_profile_panel.ts
rrahir: I think the best effort is good enough ^^

src/components/side_panel/perf_profile/perf_profile_panel.ts
LucasLefevre: I asked myself the same question when developing. I had to choose and I probably picked the easiest choice. My reasoning also was: in this case, we are interested in the functions and not so much about the result.
But I don't have a strong opinion on that matter

rrahir: neither do I, it's a go. Let's try someting

@robodoo r+

rrahir:
robodoo rebase-ff

rrahir:
robodoo r+

PR #8330, opened 2026-04-08
[copy/paste cross spreadsheet] cross version paste should be prevented

hokolomopo:
## Description:

Task: [6095101](https://www.odoo.com/odoo/2328/tasks/6095101)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_stateful/clipboard.ts
+++ b/src/plugins/ui_stateful/clipboard.ts
@@ -17,6 +17,7 @@ import {
   ClipboardOptions,
   MinimalClipboardData,
   OSClipboardContent,
+  ParsedOSClipboardContent,
 } from "../../types/clipboard";
 import {
   Command,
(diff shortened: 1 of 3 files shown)


src/plugins/ui_stateful/clipboard.ts
```diff
@@ -127,9 +128,15 @@ export class ClipboardPlugin extends UIPlugin {
       case "PASTE_FROM_OS_CLIPBOARD": {
         this._isCutOperation = false;
 
+        let contentToPaste: ParsedOSClipboardContent = cmd.clipboardContent;
+
+        // Only paste the spreadsheet data in the clipboard if the versions match
+        if (contentToPaste.data?.version !== CURRENT_VERSION) {
+          contentToPaste = { text: cmd.clipboardContent.text };
```
anhe-odoo: Maybe it's not possible, but I'll find it clearer if we had something like

```js
if (contentToPaste.data?. ...) {
    contentToPaste.data = undefined;
}
```

It will be more consistent with the above comment. Here I had to read the code 3 times to understand the fact that this line was to remove data instead of updating text. But maybe it's me that hasn't slept enough ... :D

hokolomopo: I didn't want to change the command content in place, but I can make the code clearer :)

VincentSchippefilt:
robodoo r+

PR #8332, opened 2026-04-08
[FIX] side_panel: stabilize list criterion color sync

rmbh-odoo:
## Description:

The list criterion form did not handle colors correctly when values were edited.

Colors were mapped using the value as key, which caused:
- Changing a value kept the old key -> stale color entries
- Editing a value after assigning a color lost the color
- Empty values were synced -> inconsistent UI behavior

Example:
- Setting a color without value creates `{'': '#CFE2F3'}` -> all empty inputs show the same color
- Changing value 'ok' (with color) to 'bye' keeps color on 'ok' and loses it for 'bye'

We cannot change the schema as many spreadsheets already depend on it and it may break existing data.
The issue comes from directly syncing UI state with the schema.

This commit introduces an intermediate UI state:
- Initialize UI from schema
- Avoid direct sync while editing
- Clean and apply valid color mappings

This ensures stable behavior when editing values and assigning colors.

Task: [5418098](https://www.odoo.com/odoo/2328/tasks/5418098)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/criterion_form/value_in_list_criterion/value_in_list_criterion.ts
+++ b/src/components/side_panel/criterion_form/value_in_list_criterion/value_in_list_criterion.ts
@@ -55,9 +55,8 @@ export class ListCriterionForm extends CriterionForm<IsValueInListCriterion> {
     const values = this.state.items.map((item) => item.value);
     const colors: Record<string, Color> = {};
     for (const { value, color } of this.state.items) {
-      const trimmed = value?.trim();
-      if (trimmed && color) {
-        colors[trimmed] = color;
+      if (value?.trim() !== "" && color) {
+        colors[value] = color;
       }
     }
     this.updateCriterion({ values, colors });
--- a/src/components/side_panel/criterion_form/value_in_list_criterion/value_in_list_criterion.ts
+++ b/src/components/side_panel/criterion_form/value_in_list_criterion/value_in_list_criterion.ts
@@ -12,8 +12,13 @@ css/* scss */ `
   }
 `;
 
+interface ListItem {
+  value: string;
+  color?: Color;
+}
+
 interface State {
-  numberOfValues: number;
+  items: ListItem[];
   focusedValueIndex?: number;
 }
 
@@ -22,11 +27,21 @@ export class ListCriterionForm extends CriterionForm<IsValueInListCriterion> {
   static components = { CriterionInput, RoundColorPicker };
 
   state = useState<State>({
-    numberOfValues: Math.max(this.props.criterion.values.length, 2),
+    items: [],
   });
 
   setup() {
     super.setup();
+    const values = this.props.criterion.values || [];
+    const colors = this.props.criterion.colors || {};
+    this.state.items = Array.from({ length: Math.max(values.length, 2) }, (_, i) => {
+      const value = values[i] ?? "";
+      return {
+        value,
+        color: value ? colors[value] : undefined,
+      };
+    });
+
     const setupDefault = (props: this["props"]) => {
       if (props.criterion.displayStyle === undefined) {
         this.updateCriterion({ displayStyle: "chip" });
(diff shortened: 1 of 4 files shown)


src/components/side_panel/criterion_form/value_in_list_criterion/value_in_list_criterion.ts
```diff
@@ -12,21 +12,40 @@ css/* scss */ `
   }
 `;
 
+interface Item {
```
dhrp-odoo:
```suggestion
interface ListValue {
```

nitpick: ListValue or something more explicit

src/components/side_panel/criterion_form/value_in_list_criterion/value_in_list_criterion.xml
```diff
@@ -1,25 +1,25 @@
 <templates>
   <t t-name="o-spreadsheet-ListCriterionForm">
-    <t t-foreach="displayedValues" t-as="value" t-key="value_index">
+    <t t-foreach="this.state.items" t-as="item" t-key="item_index">
```
dhrp-odoo: Can we use id here for key?

dhrp-odoo (commented):
👋
LGTM otherwise

src/components/side_panel/criterion_form/value_in_list_criterion/value_in_list_criterion.xml
rmbh-odoo: For this use case, using the index should not cause any issues. However, for better reliability and safety, we can use the ID as the source of truth.

src/components/side_panel/criterion_form/value_in_list_criterion/value_in_list_criterion.ts
```diff
@@ -12,21 +12,39 @@ css/* scss */ `
   }
 `;
 
+interface CriterionValue {
+  id: number;
```
hokolomopo: id is only used once in the xml, it is useless I think. Just use _index

src/components/side_panel/criterion_form/value_in_list_criterion/value_in_list_criterion.ts
```diff
@@ -36,27 +54,35 @@ export class ListCriterionForm extends CriterionForm<IsValueInListCriterion> {
...
 
-  onValueChanged(value: string, index: number) {
-    const values = [...this.displayedValues];
-    values[index] = value;
-    this.updateCriterion({ values });
+  private syncCriterion() {
+    const values = this.state.values.map((item) => item.value);
+    const colors: Record<string, Color> = {};
+    for (const { value, color } of this.state.values) {
+      const trimmed = value?.trim();
```
hokolomopo: I'm pretty sure trimming is wrong, we could have values with spaces in the grid

src/components/side_panel/criterion_form/value_in_list_criterion/value_in_list_criterion.ts
```diff
@@ -36,27 +54,35 @@ export class ListCriterionForm extends CriterionForm<IsValueInListCriterion> {
...
+      if (trimmed && color) {
+        colors[trimmed] = color;
+      }
+    }
+    this.updateCriterion({ values, colors });
+  }
+
+  onValueChanged(item: CriterionValue, value: string) {
+    item.value = value;
```
hokolomopo: give the index as argument instead of the item, and change this.state.values|index] in the ts. Reading the code it's really not clear that item is the value of the state, it could be anything

hokolomopo (commented):
👋

src/components/side_panel/criterion_form/value_in_list_criterion/value_in_list_criterion.ts
rmbh-odoo: Yeah, I also thought about this. It’s now fixed here, and for the previously affected versions, I’ve created a PR: https://github.com/odoo/o-spreadsheet/pull/8423

LucasLefevre (approved):
robodoo rebase-ff r+

PR #8372, opened 2026-04-09
[FIX] sheetview: undo/redo should update frozen panes

dhrp-odoo:
## Description:

Undo and redo restore pane divisions in the sheet state, but sheetview kept
using cached pane divisions until a direct freeze command updated them.
This could leave frozen panes visually stuck after history replay.

Resync pane divisions from the core state before rebuilding dirty viewports.

Task: [6103593](https://www.odoo.com/odoo/2328/tasks/6103593)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_stateful/sheetview.ts
+++ b/src/plugins/ui_stateful/sheetview.ts
@@ -155,6 +155,7 @@ export class SheetViewPlugin extends UIPlugin {
         this.viewports.cleanViewports();
         for (const sheetId of this.getters.getSheetIds()) {
           this.sheetsWithDirtyViewports.add(sheetId);
+          this.syncPaneDivision(sheetId);
         }
         this.shouldAdjustViewports = true;
         break;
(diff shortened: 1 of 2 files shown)


src/helpers/viewport_collection.ts
```diff
@@ -770,13 +771,8 @@ export class ViewportCollection {
...
-  }
-
-  setPaneDivision(sheetId: UID, paneDivision: PaneDivision) {
-    // FIXME: the pane divisions that should be used are set here, but there is nothing preventing us from
-    // using getters.getPaneDivisions anyways. We should either fix the typing/getters access, or handle pane divisions for
-    // arbitrary viewports
-    this.paneDivision[sheetId] = paneDivision;
+    // Special rendering contexts such as print can provide custom pane divisions.
+    // Otherwise, use the sheet panes from the model getters.
+    return this.paneDivisionsBySheetId?.[sheetId] || this.getters.getPaneDivisions(sheetId);
```
hokolomopo: Using `getters.getPaneDivisions` here is something we explicitely don't want to do, we do'nt want to mix `this.paneDivision` and `getters.getPaneDivisions` and have multiple sources of truth.

Maybe we could instead use `viewports.setPaneDivision` at each finalize or resetViewports ?

src/plugins/ui_stateful/sheetview.ts
```diff
@@ -192,7 +192,6 @@ export class SheetViewPlugin extends UIPlugin {
       case "FREEZE_ROWS":
       case "UNFREEZE_COLUMNS_ROWS":
         this.sheetsWithDirtyViewports.add(cmd.sheetId);
-        this.viewports.setPaneDivision(cmd.sheetId, this.getters.getPaneDivisions(cmd.sheetId));
```
dhrp-odoo, line 195: IMO this isn’t necessary

Once `cmd.sheetId` is added to `sheetsWithDirtyViewports`, `finalize()` will
resync the pane division when it rebuilds the viewports anyway

src/helpers/viewport_collection.ts
dhrp-odoo: I thought there was a comment in the code which is about this

Updated the code 👍

src/plugins/ui_stateful/sheetview.ts
```diff
@@ -192,7 +192,6 @@ export class SheetViewPlugin extends UIPlugin {
...
         break;
```
hokolomopo, line 196: the switch case content can be removed so we use the same `case` as  `REMOVE_TABLE` and the other commands

hokolomopo: Actually can't we also set the frozen panes at UNDO/REDO instead of setting them (uselessly) at every finalize ?

dhrp-odoo: then we need to do it for REMOVE_COLUMNS_ROWS / ADD_COLUMNS_ROWS / HIDE_COLUMNS_ROWS ...

hokolomopo: Why ? Frozen panes don't change when you remove rows/cols

dhrp-odoo: no :)

paneDivision stored inside ViewportCollection, we have to update to see frozen panes at correct place in UI

dhrp-odoo: Alright, i did it with specific to command, i hope i didn't missed any commad

hokolomopo (commented):
@robodoo r+

PR #8387, opened 2026-04-13
|IMP] node: remove useless dependencies

hokolomopo:
### [IMP] node: remove `mockdate` dependency

We used the `mockdate` package in the tests to mock the current date.
Mocking the date is something that can be done with jest, the extra
dependency is not needed.

### [IMP] node: remove `cors` dependency

We use the `cors` package to handle Cross-Origin Resource Sharing (CORS)
in our Node.js server. But we don't really care about fine-tuned
CORS configuration for our demo server, we can set the headers
manually and remove the `cors` dependency.

In practice, `cors` is still installed as it's a dependency of
`live-server`, but this will make it easier to get rid of `liver-server`
in the future.

### [IMP] node: remove explicit `body-parser` dependency

We have a dev dependency on `body-parser` that is not used directly
in the codebase. It will still be installed since it's a dependency of
`express`, but we can remove it from our `package.json`.

 Task: 0

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,7 +23,6 @@
         "@types/rbush": "^3.0.3",
         "@typescript-eslint/eslint-plugin": "^8.30.1",
         "babel-eslint": "^10.1.0",
-        "body-parser": "^1.19.0",
         "canvas": "^3.0.0",
         "chart.js": "4.5.0",
         "chartjs-adapter-luxon": "^1.3.1",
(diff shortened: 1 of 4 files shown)


pro-odoo:
robodoo rebase-ff r+

PR #8389, opened 2026-04-13
[FIX] computed style: fix multi-user clear formatting

matho-odoo:
Before this commit:
When in multi-user, after one user uses clear formatting, other users still see the old format until they trigger a change on a cell.

After this commit:
CLEAR_FORMATTING command invalidate cell computed style cache.

UPDATE_CELL is not handled in the plugin UI because it's a sub-command of CLEAR_FORMATTING and sub-commands are not dispatched to UI plugins when they come from the network.

Task: [6086129](https://www.odoo.com/odoo/2328/tasks/6086129)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


VincentSchippefilt:
robodoo r+

PR #8390, opened 2026-04-13
[FIX] spreadsheet_pivot: avoid crash on invalid dimension values

dhrp-odoo:
## Description:

A spreadsheet pivot could crash when a grouped dimension contained an evaluation error such as "#N/A", or mixed boolean values.

Steps to reproduce:
- create a table:

| Dates | Revenue |
| -------- | -------- |
| 01/01/2026 | 100 |
| =NA() | 200 |

- insert a Spreadsheet Pivot
- add "Dates" as a column

Adding a grouped or ordered dimension could break the whole spreadsheet when the source field contained values that could not be normalized as numbers.

In that case the pivot inferred the field as a numeric-like dimension and tried to sort its grouped keys as numbers. This worked for valid date or numeric values, but crashed as soon as one grouped key was "#N/A" or another non-normalizable value. Once the pivot recomputation failed, the spreadsheet became unusable and the pivot could no longer be removed or recovered through undo.

Infer a field as numeric only when all its non-empty values are numbers. Mixed or error values now fall back to char, so they are grouped as visible values instead of being converted as numbers during pivot recomputation.

Task: [6111913](https://www.odoo.com/odoo/2328/tasks/6111913)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/pivot/spreadsheet_pivot/spreadsheet_pivot.ts
+++ b/src/helpers/pivot/spreadsheet_pivot/spreadsheet_pivot.ts
@@ -404,10 +404,10 @@ export class SpreadsheetPivot implements Pivot<SpreadsheetPivotRuntimeDefinition
     if (nonEmptyCells.every((cell) => cell.type === CellValueType.boolean)) {
       return "boolean";
     }
-    if (nonEmptyCells.some((cell) => cell.type === CellValueType.text)) {
-      return "char";
+    if (nonEmptyCells.every((cell) => cell.type === CellValueType.number)) {
+      return "integer";
     }
-    return "integer";
+    return "char";
   }
 
   private assertCellIsValidField(col: number, row: number, cell: EvaluatedCell) {
--- a/tests/pivots/spreadsheet_pivot/spreadsheet_pivot.test.ts
+++ b/tests/pivots/spreadsheet_pivot/spreadsheet_pivot.test.ts
@@ -236,6 +236,50 @@ describe("Spreadsheet Pivot", () => {
     expect(getEvaluatedGrid(model, "B26:D26")).toEqual([["2024", "Total", ""]]);
   });
 
+  test("can group pivot columns with values in error", () => {
+    // prettier-ignore
+    const grid = {
+      A1: "Dates",        B1: "Revenue", C1: "=PIVOT(1)",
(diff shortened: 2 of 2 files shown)


src/helpers/pivot/spreadsheet_pivot/spreadsheet_pivot.ts
```diff
@@ -404,7 +404,10 @@ export class SpreadsheetPivot implements Pivot<SpreadsheetPivotRuntimeDefinition
     if (nonEmptyCells.every((cell) => cell.type === CellValueType.boolean)) {
       return "boolean";
     }
-    if (nonEmptyCells.some((cell) => cell.type === CellValueType.text)) {
+    if (
+      nonEmptyCells.some((cell) => cell.type === CellValueType.text) ||
+      nonEmptyCells.some((cell) => cell.type === CellValueType.boolean)
+    ) {
```
hokolomopo: Why this change ? I'm not sure if mix a number and boolean being an integer field make sense or not, but I don't think it's related to your fix ? It seems more of a change of behaviour to me 🤔

dhrp-odoo: I included this change because the same error can occur with booleans too, not only with text.

Steps:

| Dates | Revenue |
| -------- | -------- |
| 01/01/2026 | 100 |
| TRUE | 200 |
| 01/03/2026 | 300 |

If we create a spreadsheet pivot from this data and use `Dates` as a column, the dimension is inferred as `integer`. Later, it tries to convert a string "true" to a number, which causes the same error.

So, IMO when a dataset contains booleans in this kind of mixed column, we should treat the dimension as char instead.

hokolomopo: Make sense. But if we want to change the default field type to string, we should change the default return of the function, and have a condition `nonEmptyCells.every(cell is number) => return "integer"`

dhrp-odoo: make sense, if we do this, then we don't need those conditions to check evaluation errors

hokolomopo (commented):
@robodoo r+

PR #8392, opened 2026-04-13
[IMP] icons: add scalability

anhe-odoo:
## Description

Most of the current icons won't scale with their container, as they has been made for a 18x18 area. If we change this (e.g. change --os-icon-edge-length to something else), the result will be ugly. This can be fixed by adding a viewbox to all the svg, ensuring they will be resized to fill the container.

## Related Task:
- Task: 6047628


hokolomopo (commented):
@robodoo r+

PR #8422, opened 2026-04-15
[FIX] pivot side panel: invalid computed measure

matho-odoo:
Before this commit:
When creating a computed measure with  "=1/0a" or "=NotAnExistingSheetName!a1", the measure is not shown as invalid in the composer but it's shown as a BAD_EXPR in the grid

After this commit:
The measure is shown as invalid in the composer

Task: [5095901](https://www.odoo.com/odoo/2328/tasks/5095901)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/pivot/pivot_layout_configurator/pivot_measure/pivot_measure.ts
+++ b/src/components/side_panel/pivot/pivot_layout_configurator/pivot_measure/pivot_measure.ts
@@ -1,8 +1,10 @@
 import { Component } from "@odoo/owl";
 import { compile } from "../../../../../formulas";
+import { InternalCompiledFormula } from "../../../../../formulas/compiler";
+import { functionRegistry } from "../../../../../functions";
 import { PivotRuntimeDefinition } from "../../../../../helpers/pivot/pivot_runtime_definition";
 import { createMeasureAutoComplete } from "../../../../../registries/auto_completes/pivot_dimension_auto_complete";
-import { PivotMeasure } from "../../../../../types";
+import { PivotMeasure, Range } from "../../../../../types";
 import { StandaloneComposer } from "../../../../composer/standalone_composer/standalone_composer";
 import { PivotDimension } from "../pivot_dimension/pivot_dimension";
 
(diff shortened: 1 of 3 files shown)


matho-odoo:
robodoo help

matho-odoo:
robodoo up to saas-19.2

tests/pivots/spreadsheet_pivot/spreadsheet_pivot_side_panel.test.ts
```diff
@@ -171,13 +171,16 @@ describe("Spreadsheet pivot side panel", () => {
...
   });
 
-  test("Invalid calculated measure formula have an invalid class on the composer", async () => {
-    await click(fixture.querySelectorAll(".add-dimension")[2]);
-    expect(fixture.querySelector(".o-popover")).toBeDefined();
-    await click(fixture, ".add-calculated-measure");
-    await editStandaloneComposer(".pivot-dimension .o-composer", "=abcdefg()");
-    expect(fixture.querySelector(".o-standalone-composer")).toHaveClass("o-invalid");
-  });
+  test.each(["=abcdefg()", "=1/0a", "=NotAnExistingSheetName!a1"])(
```
anhe-odoo, line 174: Maybe we can also test `=InvalidMeasureName`

src/components/side_panel/pivot/pivot_layout_configurator/pivot_measure/pivot_measure.ts
```diff
@@ -84,6 +86,31 @@ export class PivotMeasureEditor extends Component<Props> {
...
+    if (compiledFormula.isBadExpression) {
+      return true;
+    }
+    const measures = new Set(this.props.definition.measures.map((m) => m.id));
+    const dimensions = new Set([
+      ...this.props.definition.columns.map((d) => d.nameWithGranularity),
+      ...this.props.definition.rows.map((d) => d.nameWithGranularity),
+    ]);
+    for (const symbol of compiledFormula.symbols) {
+      if (!measures.has(symbol) && !dimensions.has(symbol) && !functionRegistry.contains(symbol)) {
```
LucasLefevre, line 101: The current approach works until other new types of symbol appear (*named range* 👀 ). It is not robust and isn't maintainable with future new features. No one is will ever think about fixing this here.

The only sane and future-proof way to know if the formula is correct is to evaluate it (in the correct context, which is different depending on the group level).

And even then, there would be no way to differentiate valid errors (result of a valid formula, expected by the users) from a "wrong" formula.

I think the current situation is good-enough/best effort and it doesn't need to be fixed.

VincentSchippefilt: @LucasLefevre While you are correct that every time we add a new type of symbol this method will need to be updated, it still provides added value while editing the pivots.

In the future, we should probably collect all the "symbol" related features in a registry, this validator, the resolver used in evaluation, etc. but this in not the point of this PR.

My opinion is that is is simple enough and provides enough value to be merged.

LucasLefevre: It's not that simple ^^
It already has 2 problems:

- function name check should be case-insensitive
- `=ADD` is *not* marked as invalid

Also, having a red underline is nice and fine, but it doesn't tell you what's actually wrong -> you still need to check the error message in the cell.

VincentSchippefilt: After discussing with Remi about this, here is my current opinion:
- indeed it's not enough to validate only the calculated measures of the pivot, we should also validate somehow the formulas in other contextes and side_panels (different types of CF too)
- the validation probably needs some kind of context to be done correctly: we don't validate a CF formula using the symbols of the calculated measures of a pivot. In the future, we also should not validate the calculated measures of a pivot with the one of another pivot or list.
- The validation should move to the StandaloneComposerStore. It's going to be weird, because validating a formula like `=pivot(3,"measure","field",pivot(4,"second_measure"))` should be correct: the `second_measure` is only valid within the pivot 4

I don't have a good design in mind for now. Better put this task in the fridge for a week or 2 to have the time to thing about it...

PR #8435, opened 2026-04-16
[IMP] clipboard: optimize copiedData

anhe-odoo:
## Task Description

This PR aims to simplify the copiedData and the clipboard handlers structure globally to reduce de copied data and the HTMLContent :
- Refactor clipboard handlers to use unified 2D array data model
- Copy only minimalist and needed data in each handlers
- Update copy/paste logic to pass origin zones and support new data structure
- Refactor clipboard data compaction/expansion for flat and sparse formats to
  avoid keeping empty cell/CF/DV/borders in the copiedData

## Related Task

- Task: [6052104](https://www.odoo.com/odoo/2328/tasks/6052104)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/plugins/ui_stateful/clipboard.ts
```diff
@@ -760,17 +792,97 @@ export class ClipboardPlugin extends UIPlugin {
...
+    for (const zone of zones as Zone[]) {
       const { x, y, width, height } = viewports.getVisibleRect(sheetId, zone);
       if (width > 0 && height > 0) {
         ctx.strokeRect(x, y, width, height);
       }
     }
   }
 }
+
+// ---------------------------------------------------------------------------
```
anhe-odoo, line 810: Should these be moved to `clipboard_helpers.ts` (it's only used here)

src/clipboard_handlers/abstract_cell_clipboard_handler.ts
```diff
@@ -4,12 +4,12 @@ import { ClipboardCellData, ClipboardCopyOptions, ClipboardOptions } from "../ty
...
 import { ClipboardHandler } from "./abstract_clipboard_handler";
 
-export class AbstractCellClipboardHandler<T, T1> extends ClipboardHandler<T> {
+export class AbstractCellClipboardHandler<T1> extends ClipboardHandler<T1[][]> {
   copy(
     data: ClipboardCellData,
     isCutOperation: boolean,
     mode: ClipboardCopyOptions = "copyPaste"
-  ): T | undefined {
+  ): T1[][] | undefined {
```
LucasLefevre:
```suggestion
export class AbstractCellClipboardHandler<T> extends ClipboardHandler<T[][]> {
copy(
data: ClipboardCellData,
isCutOperation: boolean,
mode: ClipboardCopyOptions = "copyPaste"
  ): T[][] | undefined {
```

src/helpers/clipboard/clipboard_helpers.ts
```diff
@@ -118,7 +118,103 @@ function getOSheetDataFromHTML(htmlDocument: Document) {
...
+          value: cell.evaluatedCell?.value ?? null,
+          format: cell.evaluatedCell?.format,
+          formattedValue: cell.evaluatedCell?.value?.toString?.() ?? cell.content ?? "",
+        },
+      };
+    })
+  );
+
+  // Expand sparse 2D arrays for handlers that use them
+  const sparse2DHandlers = ["border", "merge", "conditionalFormat", "dataValidation", "table"];
```
LucasLefevre: Don't hard-code clipboard handler names. It's error prone when we'll add a new one

LucasLefevre (commented):
Only scrolled really fast.
I only have two questions at this point:

- Do you have a measure of the size before/after ?
- Did you consider using the squisher/unsquisher ?

src/clipboard_handlers/cell_clipboard.ts
```diff
@@ -214,27 +202,63 @@ export class CellClipboardHandler extends AbstractCellClipboardHandler<
     cells: ClipboardCell[][],
     clipboardOptions?: ClipboardOptions
   ) {
-    // then, perform the actual paste operation
+    const { originRows, originCols } = this.getOriginIndices(clipboardOptions?.originZones);
     for (const [r, rowCells] of cells.entries()) {
+      const originRow = originRows?.[r];
+      const deltaRow = originRow !== undefined ? row + r - originRow : 0;
```
hokolomopo, line 274: This seems very wonky. There is no guarantee that `getOriginIndices` will return indices in the same order as `getClipboardDataPositions`. Can't you store the origin positions in `ClipboardCell` ? Or pass `rowsIndexes, columnsIndexes` as argument of the `paste`.

src/clipboard_handlers/conditional_format_clipboard.ts
```diff
@@ -54,54 +47,67 @@ export class ConditionalFormatClipboardHandler extends AbstractCellClipboardHand
...
     }
   }
 
+  private getOriginPosition(r: number, c: number, options?: ClipboardOptions): CellPosition {
+    return {
+      col: (options?.originZones?.[0]?.left ?? 0) + c,
+      row: (options?.originZones?.[0]?.top ?? 0) + r,
+      sheetId: options?.originSheetId ?? "",
+    };
+  }
```
hokolomopo: The logic here to find the origin position is totally different than the one in the cell clipboard. One of them is probably wrong (this one ? it doesn't account for multiple origin zones).

And if multiple clipboard have the same need to find the origin position from the copied content matrix, there should be a generic solution.

hokolomopo: also having everything optional kinda sucks, and the fallbacks make no sense. Surely this should crash if no originZone was provided ? Defaulting to `0,0` seems wrong. 

Can you make them mandatory in `ClipboardOptions` ?

src/clipboard_handlers/data_validation_clipboard.ts
```diff
@@ -51,43 +44,56 @@ export class DataValidationClipboardHandler extends AbstractCellClipboardHandler
...
-        const position = { col: col + c, row: row + r, sheetId };
-        this.pasteDataValidation(origin, position, clipboardOptions?.isCutOperation);
+        const target = { col: col + c, row: row + r, sheetId };
+        const originPosition = this.getOriginPosition(r, c, clipboardOptions);
+        this.pasteDataValidation(origin, target, originPosition, clipboardOptions?.isCutOperation);
       }
     }
   }
 
+  private getOriginPosition(r: number, c: number, options?: ClipboardOptions): CellPosition {
```
hokolomopo: same comment as CFs

src/clipboard_handlers/references_clipboard.ts
```diff
@@ -1,31 +0,0 @@
-import { ClipboardCellData, ClipboardOptions, ClipboardPasteTarget } from "../types/clipboard";
-import { UID, Zone } from "../types/misc";
-import { AbstractCellClipboardHandler } from "./abstract_cell_clipboard_handler";
-
-interface ClipboardContent {
-  zones: Zone[];
-  sheetId: UID;
-}
-
-export class ReferenceClipboardHandler extends AbstractCellClipboardHandler<ClipboardContent, {}> {
```
hokolomopo, line 10: not sure about renaming `ReferenceClipboardHandler` to `ZoneClipboardHandler`. Reference was more explicit for me. 

But for this and for the renamed `frozenPaneCLipboard`, can you do the rename in a separate commit ? git didn't detect the rename, we'll lose history

src/helpers/clipboard/clipboard_helpers.ts
```diff
@@ -118,7 +118,103 @@ function getOSheetDataFromHTML(htmlDocument: Document) {
...
+  }
+
+  return data;
+}
+
+function getSparse2DDefaultValue(handlerName: string, sheetId: string): any {
+  switch (handlerName) {
+    case "border":
+      return null;
+    case "merge":
```
hokolomopo: I do agree with lucas' comment, don't hard-code handledr name. Add static properties to the clipboard handlers instead, with `isSparseData/defaultValue` and whatnot

src/helpers/clipboard/clipboard_helpers.ts
```diff
@@ -118,7 +118,103 @@ function getOSheetDataFromHTML(htmlDocument: Document) {
...
+ * Handles two formats:
+ * - Old format: has top-level 'cells' key directly (won't be needed as soon as cross-version
+ *   compitiblity is droppen (see ADRM task))
+ * - New flat format: data.cell is an array (handler data is flat)
+ */
+function expandClipboardData(data: Record<string, any>): Record<string, any> {
+  // Old format: has top-level 'cells' key directly
+  if ("cells" in data && !data.cell) {
+    return data;
+  }
```
hokolomopo: why do we care about the old format at all ?

src/clipboard_handlers/abstract_cell_clipboard_handler.ts
```diff
@@ -4,16 +4,16 @@ import { ClipboardCellData, ClipboardCopyOptions, ClipboardOptions } from "../ty
 import { HeaderIndex, UID, Zone } from "../types/misc";
```
hokolomopo: I didn't read into much details your code to "compress" the data, but the promises of the commit message don't seem to be held (unless I messed something up). 

[copied_data.json](https://github.com/user-attachments/files/27698697/copied_data.json)

Here is some copied data. There is a bunch of empty cells, empty table cells, and the same merge/conditional format are repeated 20 times, and the nonempty table cells are not normalized (and it's probably not something you changed, but I'm not sure why individual table cells are copied when `isWholeTableCopied === true`)

hokolomopo (commented):
👋

src/clipboard_handlers/references_clipboard.ts
hokolomopo: Up

src/clipboard_handlers/abstract_cell_clipboard_handler.ts
```diff
@@ -1,34 +1,45 @@
...
+  ClipboardCellData,
+  ClipboardCopyOptions,
+  ClipboardOptions,
+  ClipboardPositions,
+} from "../types/clipboard";
+import { CellPosition, HeaderIndex, UID, Zone } from "../types/misc";
 import { ClipboardHandler } from "./abstract_clipboard_handler";
 
-export class AbstractCellClipboardHandler<T, T1> extends ClipboardHandler<T> {
+export class AbstractCellClipboardHandler<T, T1> extends ClipboardHandler<(T | null)[][]> {
```
rrahir:
```suggestion
export class AbstractCellClipboardHandler<T, U> extends ClipboardHandler<(T | null)[][]> {
```

src/clipboard_handlers/abstract_cell_clipboard_handler.ts
```diff
@@ -39,7 +50,23 @@ export class AbstractCellClipboardHandler<T, T1> extends ClipboardHandler<T> {
     sheetId: UID,
     col: HeaderIndex,
     row: HeaderIndex,
-    data: T1[][],
-    clipboardOptions?: ClipboardOptions
+    data: (T | null)[][],
+    clipboardOptions: ClipboardOptions,
```
rrahir, line 54: There is something fishy with the typing. sometimes the options &  positions are mandatory, sometimes not. I tried to fix it but I'm facing a wall because of the default plugin (I think) because it poses as a cell plugin even though it isn't really. I think it should become a SheetClipboard handler but can't make it work easily.

 I added a commit with the enforcing of types but it creates a weird structured of clipboard handers with 4 different types. Alternatively, we can considerate DefaultClipboardhandler as a cellhandler and define some ghost  `pasteZone` method that will nver be called and I find it kind of shitty but it's way easier than the commit I suggest.

src/clipboard_handlers/tables_clipboard.ts
```diff
@@ -130,42 +117,49 @@ export class TableClipboardHandler extends AbstractCellClipboardHandler<
...
 
   protected pasteZone(
     sheetId: UID,
     col: HeaderIndex,
     row: HeaderIndex,
-    tableCells: TableCell[][],
-    clipboardOptions?: ClipboardOptions
+    tableCells: ClipboardTableCell[][],
+    clipboardOptions?: ClipboardOptions,
+    _positions?: ClipboardPositions
```
rrahir: seems unused

rrahir:
robodoo r+

rrahir:
robodoo rebase-ff

rrahir:
robodoo delegate+

rrahir:
robodoo delegate-

@anhe-odoo we have issues with the squisher right now, I want to make sure that it does not make it worse

src/stores/clipboard_store.ts
```diff
@@ -384,18 +384,20 @@ export class ClipboardStore extends SpreadsheetStore {
...
+      return { handlerName, handler: new Handler(this.getters, this.model.dispatch) };
     });
-    const sheetEntries = clipboardHandlersRegistries.sheetHandlers.getKeys().map((handlerName) => {
-      const Handler = clipboardHandlersRegistries.sheetHandlers.get(handlerName);
+    const rangeEntries = clipboardHandlersRegistries.rangeHandlers.getKeys().map((handlerName) => {
+      const Handler = clipboardHandlersRegistries.rangeHandlers.get(handlerName);
       return { handlerName, handler: new Handler(this.getters, this.model.dispatch) };
     });
-    return [...cellEntries, ...sheetEntries];
+    1977;
```
anhe-odoo: @rrahir 🙄 😆

rrahir: wtf

PR #8463, opened 2026-04-21
[FIX] formulas: add IFERROR second argument when exporting data

anhe-odoo:
## Task Description

In our implementation of the IFERROR formula, the second argument (value if error) is optionnal, defaulting to an empty string, which results to a 0 value in the cell when evaluated. In the excel documentation, the second argument of IFERROR is required leading to issue when exporting from spreadsheet a formula with IFERROR and
then importing it to excel or Gsheet.

This PR fixes this issue by adding a second argument when needed in the exporting process.

## Related Task
- Task: [5993405](https://www.odoo.com/odoo/2328/tasks/5993405)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/xlsx/constants.ts
+++ b/src/xlsx/constants.ts
@@ -89,10 +89,10 @@ export const FIRST_NUMFMT_ID = 164;
 
 export const DEFAULT_DOUGHNUT_CHART_HOLE_SIZE = 50;
 
-interface functionDefaultArg {
+type functionDefaultArg = {
   type: "NUMBER";
   value: number;
-}
+};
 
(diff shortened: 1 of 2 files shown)


src/migrations/migration_steps.ts
```diff
@@ -639,6 +640,70 @@ migrationStepRegistry
       }
       return data;
     },
+  })
+  .add("19.3.3", {
```
LucasLefevre, line 644: we also need a python upgrade script for commands in revisions

anhe-odoo:
Closed in favor of https://github.com/odoo/o-spreadsheet/pull/8756

PR #8464, opened 2026-04-21
[IMP] dashboard: set white background color

matho-odoo:
Task: [6088803](https://www.odoo.com/odoo/2328/tasks/6088803)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/dashboard/dashboard.xml
+++ b/src/components/dashboard/dashboard.xml
@@ -1,7 +1,7 @@
 <templates>
   <t t-name="o-spreadsheet-SpreadsheetDashboard">
     <div
-      class="o-grid o-two-columns o-zoomable"
+      class="o-grid o-two-columns o-zoomable bg-white"
       t-ref="dashboard"
       tabindex="-1"
       t-on-wheel="onMouseWheel"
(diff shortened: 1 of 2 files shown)


src/components/dashboard/dashboard.ts
```diff
@@ -158,6 +158,6 @@ export class SpreadsheetDashboard extends Component<Props, SpreadsheetChildEnv>
 
   get dashboardStyle() {
     const zoomLevel = this.env.model.getters.getViewportZoomLevel();
-    return cssPropertiesToCss({ zoom: `${zoomLevel}` });
+    return cssPropertiesToCss({ zoom: `${zoomLevel}`, "background-color": "#FFFFFF" });
```
LucasLefevre: use the class `bg-white` in the xml instead.
Only dynamic style should be set with javascript. If the style is static: it should be bootstrap classes or in css files (if bootstrap cannot do what you want)

LucasLefevre (approved):
robodoo r+

PR #8465, opened 2026-04-22
[IMP] *: remove index files in sources

rrahir:
several index files in the sources are a generic source of circular dependencies as some of them (helpers/index for instance) export everything from their subfolders. In this case, IDE's will generally suggest to import from helpers/index which means that we create a dependency between the caller and every helpers files and we already had to tweak some import manually to avoid those circular dependencies.

While they are useful to ensure that some files are properly loaded, they don't need to export anything and be an open access to their subfolders.

Overall, the index files were essential to properly load some registries, which is done in standalone files to avoid circular dependencies in the first place.

This commit removes the 'open door' index files aggressively and only keeps the files meant to populate the registries. These files will have to be explicitely loaded in the root file of the library `src/index.ts` as it was already the case for the chart types.

Task: 6094655

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


LucasLefevre (approved):
robodoo r+

LucasLefevre:
fuck, c'est le UuidGenerator

LucasLefevre:
robodoo r+

rrahir:
robodoo r+

rrahir:
:troll:

PR #8466, opened 2026-04-22
[FIX] grid overlay: unhide buttons visibility

matho-odoo:
Before this commit:
When a row/column is frozen and hidden, the unhide buttons were not visible in the header overlay. There is no way for the user to unhide these rows/columns.

After this commit:
The unhide buttons are now visible in the header overlay when a frozen row/column is hidden.

Task: [6127335](https://www.odoo.com/odoo/2328/tasks/6127335)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


pro-odoo:
robodoo r+

PR #8469, opened 2026-04-22
[IMP] chart: export pie slice colors

anhe-odoo:
## Task Description
Now that we can have custom pie slice colors, this PR aims to export and import the colors from/to xlsx format.

## Related Task
- Task: [5969070](https://www.odoo.com/odoo/2328/tasks/5969070)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/xlsx/extraction/chart_extractor.ts
```diff
@@ -44,6 +44,29 @@ export class XlsxChartExtractor extends XlsxBaseExtractor {
...
+          if (ser) {
+            const dPtNodes = Array.from(this.querySelectorAll(ser, "c:dPt"));
+            const colorsByIdx: Record<number, string> = {};
+            for (const dPt of dPtNodes) {
+              const idx = this.extractChildAttr(dPt, "c:idx", "val", { required: true }).asNum();
+              const clr = this.querySelector(dPt, "a:solidFill a:srgbClr");
+              const val = clr?.getAttribute("val");
+              if (typeof idx === "number" && val && isColorValid(val)) {
+                colorsByIdx[idx] = toHex(val);
+              }
```
hokolomopo: we have specific method (`extractColor`) to extract colors from excel, because they are not always in hex format. I suspect this is why it's bugged (see my comment in the chatter)

tests/xlsx/xlsx_export.test.ts
```diff
@@ -1746,6 +1746,23 @@ describe("Test XLSX export", () => {
       });
       expect(await exportPrettifiedXlsx(model)).toMatchSnapshot();
     });
+    test("pie chart with custom colors", async () => {
```
hokolomopo, line 1845:
```suggestion
}); 

    test("pie chart with custom colors", async () => {
```
newline

tests/xlsx/xlsx_export.test.ts
```diff
@@ -1746,6 +1746,23 @@ describe("Test XLSX export", () => {
...
+            dataSets: [{ dataRange: "Sheet1!B2:B5" }],
+            labelRange: "Sheet1!A2:A5",
+            dataSetsHaveTitle: false,
+          }),
+          type: "pie",
+          slicesColors: ["#FF0000", "#00FF00", "#0000FF", "#FFFF00"],
+        },
+        "1"
+      );
+      expect(await exportPrettifiedXlsx(model)).toMatchSnapshot();
```
hokolomopo: if possible test the feature with querySelectors/regex, snapshots tests are mostly useless IMO

tests/xlsx/xlsx_import_export.test.ts
```diff
@@ -543,4 +544,31 @@ describe("Export data to xlsx then import it", () => {
     const newFigure = importedModel.getters.getFigures(sheetId)[0];
     expect(newFigure.height).toBe(figure.height);
   });
+
+  test("Pie chart slices colors are imported from xlsx", async () => {
```
hokolomopo, line 602: Don't test the xlsx import that way. You could be generating a totally wrong xlsx at export then test that this wrong xlsx is correctly imported.

We have the `xlsx_demo_data.xlsx` file, use it. Open it in excel, create a pie chart w/ custom colors (or modify the existing pie chart), then test the import of pie colors in `xlsx_import.test.ts`.

hokolomopo (commented):
👋

hokolomopo:
@anhe-odoo you forgot to address my comments 🙃

src/xlsx/extraction/chart_extractor.ts
```diff
@@ -139,6 +162,50 @@ export class XlsxChartExtractor extends XlsxBaseExtractor {
...
+  private resolveSchemeColor(schemeName: string): string | undefined {
+    const themeFile = this.xlsxFileStructure.theme;
+    if (!themeFile) {
+      return undefined;
+    }
+    const schemeEl = this.querySelector(themeFile.file.xml, `a:clrScheme a:${schemeName}`);
+    if (!schemeEl) {
+      return undefined;
+    }
+    const srgbClr = this.querySelector(schemeEl, "a:srgbClr");
```
hokolomopo, line 254: So in their very easy to follow spec, it seems that this is not totally compete for openXML. The color is sometime a theme color, but some time it should be translated in a theme color [via a colorMap](https://c-rex.net/samples/ooxml/e1/part4/OOXML_P4_DOCX_clrMap_topic_ID0ETDFMB.html?utm_source=chatgpt.com). But the colorMap is not always defined, so excel uses a default one  ...

Image

For example here `bg2` is not in the theme. But it is present in the default color map, so we should look at `bg2 => lt2` in the theme.

Image

But that's not all ! Here we don't use the color `accent2` as-it. We need to modify it with the `lumMod/lumOff` parameters to get a lighter shade of orange.
 
Talk about annoying ...

[Book1.xlsx](https://github.com/user-attachments/files/28469259/Book1.xlsx)

hokolomopo (commented):
The import is still a bit foireux, because OpenXML is very easy and simple. (Or is it only excel's version of openXML ?)

I guess we want to address these to do a best effort color conversion ...

hokolomopo:
@robodoo r+

PR #8499, opened 2026-04-27
[FIX] Charts: Do not cast dates x-values as numbers

rrahir:
### [FIX] charts: Do not cast empty values as 'count'

Following the recent refactoring, dataset entries with empty values were
mistakenly treated as text and therefore converted as 'counts' whenever
a part of the dataset contained some text values.

### [FIX] Charts: Do not cast dates x-values as numbers
the datapoint X-values of the datasets should be the same as the ones
provided as labels in order to Match and be properly positions in the
chart.
The refactoring changed that behaviour to respect the typing of Chart.js
but once again, the typing is off again goes against some examples
provided in their online documentation.

Task: 6146610


rrahir:
robodoo rebase-ff

package.json
```diff
@@ -164,6 +164,9 @@
       "node"
     ],
     "workerIdleMemoryLimit": "800MB",
+    "transformIgnorePatterns": [
+      "/node_modules/(?!chartjs-adapter-luxon)"
+    ],
```
hokolomopo, line 169: tests are running fine w/o this for me

rrahir: it crashes without it on my end ><

src/helpers/figures/charts/chart_data_sources.ts
```diff
@@ -293,7 +293,7 @@ function getChartDatasetValues(getters: Getters, dataSets: DataSet[]): DatasetVa
       data.filter(isTextResult).length > 1
     ) {
       // Convert categorical data into counts
-      data = data.map((cell) => (!isErrorResult(cell) ? ONE : EMPTY));
+      data = data.map((cell) => (cell.value && !!isErrorResult(cell) ? ONE : EMPTY));
```
pro-odoo: As discussed, add a test for error and check for double negation

rrahir: => a test already exists for the condition but the double negation was useless indeed

pro-odoo (approved):
robodoo rebase-ff r+

PR #8521, opened 2026-04-29
[IMP] find and replace: add option to include hidden values in search results

matho-odoo:
Task: [6171587](https://www.odoo.com/odoo/2328/tasks/6171587)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/find_and_replace/find_and_replace.css
+++ b/src/components/side_panel/find_and_replace/find_and_replace.css
@@ -3,7 +3,7 @@
     outline: none;
     height: 100%;
     .o-input-search-container {
-      display: flex;
+      padding: 4px 0;
       .o-input-with-count {
         flex-grow: 1;
         width: auto;
@@ -13,15 +13,13 @@
       }
       .o-input-count {
         width: fit-content;
-        padding: 4px 0 4px 4px;
         white-space: nowrap;
       }
     }
     .o-result-buttons {
       .o-button {
-        height: 19px;
-        width: 19px;
-        box-sizing: content-box;
+        min-height: 23px;
+        width: 23px;
         .o-icon {
           height: 14px;
           width: 14px;
--- a/src/components/side_panel/find_and_replace/find_and_replace.xml
+++ b/src/components/side_panel/find_and_replace/find_and_replace.xml
@@ -2,7 +2,7 @@
   <t t-name="o-spreadsheet-FindAndReplacePanel">
     <div class="o-find-and-replace" t-on-keydown="this.onKeydownPanel">
       <Section title.translate="Search">
-        <div class="o-input-search-container">
+        <div class="o-input-search-container d-flex align-items-center">
           <input
             type="text"
             t-ref="this.searchInputRef"
@@ -72,15 +72,47 @@
             value="this.searchOptions.searchFormulas"
             label="searchFormulasLabel"
             onChange.bind="this.searchFormulas"
+            className="'mb-1'"
+          />
+          <t t-set="includeHiddenLabel">Include hidden results</t>
+          <Checkbox
+            value="this.searchOptions.includeHidden"
+            label="includeHiddenLabel"
+            onChange.bind="this.searchIncludeHidden"
           />
         </div>
         <div class="o-matches-count mt-4" t-if="this.searchInfo.length">
-          <ValidationMessages msgType="'info'" messages="this.searchInfo" singleBox="true"/>
+          <ValidationMessages msgType="'info'">
+            <t t-set-slot="info-text">
+              <div class="ps-2" t-foreach="this.searchInfo" t-as="infoLine" t-key="infoLine">
+                <span t-out="infoLine"/>
+              </div>
+            </t>
+          </ValidationMessages>
+        </div>
+        <div
+          class="o-matches-count mt-4"
+          t-if="this.searchOptions.includeHidden and this.hiddenSheetsWithMatchesInfo.length">
+          <ValidationMessages msgType="'info'">
+            <t t-set-slot="info-text">
+              <div class="ps-2">
+                Matches in the following hidden sheets:
+                <ul class="mb-0">
+                  <li
+                    t-foreach="this.hiddenSheetsWithMatchesInfo"
+                    t-as="sheetName"
+                    t-key="sheetName">
+                    <span t-out="sheetName"/>
+                  </li>
+                </ul>
+              </div>
+            </t>
+          </ValidationMessages>
         </div>
       </Section>
       <t t-if="!this.env.model.getters.isReadonly()">
         <Section class="'pt-0'" title.translate="Replace">
-          <div class="o-input-search-container">
+          <div class="o-input-search-container d-flex align-items-center">
             <input
               type="text"
               class="o-input o-input-without-count o-replace"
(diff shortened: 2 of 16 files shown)


src/components/side_panel/find_and_replace/find_and_replace_store.ts
```diff
@@ -2,7 +2,7 @@ import { getSearchRegex } from "../../../helpers/misc";
 import { isInside, positionToZone } from "../../../helpers/zones";
```
hokolomopo, line 2: If possible, as the spec says, we should tweak the rendering of the highlight a bit so they are visible when they are hidden 

Image

src/components/side_panel/find_and_replace/find_and_replace_store.ts
```diff
@@ -330,7 +366,9 @@ export class FindAndReplaceStore extends SpreadsheetStore implements HighlightPr
     // we want grid selection to capture the selection stream
     this.model.selection.getBackToDefault();
     if (options.updateSelection) {
-      this.model.selection.selectCell(selectedMatch.col, selectedMatch.row);
+      this.model.selection.selectCell(selectedMatch.col, selectedMatch.row, {
+        byppassSelectionOutOfBound: true,
```
hokolomopo:
```suggestion
        bypassSelectionOutOfBound: true,
```

src/components/validation_messages/validation_messages.xml
```diff
@@ -1,22 +1,24 @@
...
-            <span
-              t-foreach="box"
-              t-as="msg"
-              t-key="msg_index"
-              class="ps-2"
-              t-att-class="{'text-truncate': props.singleBox }"
-              t-out="msg"
-            />
+    <div t-if="props.msgType !== 'info'">
+      <t t-foreach="this.props.messages" t-as="msg" t-key="msg_index">
```
hokolomopo: I don't think it's a good idea to have the condition be on the type of msg. What if we want to display a custom error message ? Or a standard info message ? 

Is it possible to have a condition `if a slot is given` instead ? Or just always require the slot, and remove the props message

src/selection_stream/selection_stream_processor.ts
```diff
@@ -441,7 +441,7 @@ export class SelectionStreamProcessorImpl implements SelectionStreamProcessor {
     const sheetId = this.getters.getActiveSheetId();
     const refCol = this.getters.findVisibleHeader(sheetId, "COL", left, right);
     const refRow = this.getters.findVisibleHeader(sheetId, "ROW", top, bottom);
-    if (refRow === undefined || refCol === undefined) {
+    if ((refRow === undefined || refCol === undefined) && !options?.byppassSelectionOutOfBound) {
       return CommandResult.SelectionOutOfBound;
     }
```
hokolomopo: Not sure it's a good idea, yes it works for hidden rows but now it also works if we give col/row indexes out of the sheet (and it probably make everything crash). We should at least always check that the col/rows are inside the sheet.

But did you investigate why this check was there in the first place ? Can we always allow events on hidden rows/cols ?

src/components/side_panel/find_and_replace/find_and_replace.xml
```diff
@@ -71,9 +71,36 @@
...
+          class="o-matches-count mt-4"
+          t-if="searchOptions.includeHidden and hiddenSheetsWithMatchesInfo.length">
+          <ValidationMessages msgType="'info'">
+            <t t-set-slot="info-text">
+              <t t-out="hiddenSheetsWithMatchesTooltip"/>
+              <ul>
+                <li t-foreach="hiddenSheetsWithMatchesInfo" t-as="sheetName" t-key="sheetName">
+                  <span t-out="sheetName"/>
+                </li>
+              </ul>
```
hokolomopo: bottom margin is a bit too big

Image

hokolomopo (commented):
👋

src/components/side_panel/find_and_replace/find_and_replace.css
```diff
@@ -28,5 +28,8 @@
         }
       }
     }
+    .o-hidden-sheets-list {
+      margin-bottom: 0;
+    }
```
hokolomopo: can be done with bootstrap (mb-0)

hokolomopo (commented):
👋

src/components/side_panel/find_and_replace/find_and_replace.css
```diff
@@ -20,6 +20,7 @@
     .o-result-buttons {
       .o-button {
         height: 19px;
```
rrahir:
```suggestion
        height: 19px !important;
```
the cahnge below conflicts with this rule (there already was a conflit before) 
I think the buttons were meant to be square (height and width set to 19px) so its better to force the height to 19 px

src/components/side_panel/find_and_replace/find_and_replace.css
```diff
@@ -20,6 +20,7 @@
...
+        min-height: 28px;
```
rrahir: to remove :)

src/components/validation_messages/validation_messages.css
```diff
@@ -7,6 +7,8 @@
 
     .o-icon {
       margin-right: 5px;
+      min-width: 1.2em;
```
rrahir: do you need this? there already is a rule just below so it probably means that another css rule is posing problems

EDIT: ok i get it, the icon gets crushed by the content beside it.

it would be better to wrap the `<t t-if="this.props.msgType === 'info'" t-call="o-spreadsheet-Icon.CIRCLE_INFO"/>` of the template inside a div with the class `flex-shrink-0` which prevents the element to be crushed by its neighbors :)

src/components/validation_messages/validation_messages.ts
```diff
@@ -8,9 +8,9 @@ export class ValidationMessages extends Component<SpreadsheetChildEnv> {
   static template = "o-spreadsheet-ValidationMessages";
 
   protected props = props({
-    messages: types.array(types.string()),
+    "messages?": types.array(types.string()),
     msgType: types.or([types.literal("warning"), types.literal("error"), types.literal("info")]),
-    "singleBox?": types.boolean(),
+    "slots?": types.object(),
```
rrahir: this will require another rebase :/
```suggestion
    "slots": types.object().optional(),
```

src/components/validation_messages/validation_messages.xml
```diff
@@ -1,19 +1,25 @@
 <templates>
   <t t-name="o-spreadsheet-ValidationMessages">
-    <t t-foreach="this.alertBoxes" t-as="box" t-key="'box' + box_index">
+    <t t-if="this.props.messages">
+      <t t-foreach="this.props.messages" t-as="msg" t-key="msg_index">
+        <div t-att-class="this.divClasses" class="d-flex flex-column p-3 m-1 o-validation">
+          <div class="d-flex align-items-center">
+            <t t-if="this.props.msgType === 'info'" t-call="o-spreadsheet-Icon.CIRCLE_INFO"/>
```
rrahir:
```suggestion
           <div class="flex-shrink-0"> <t t-if="this.props.msgType === 'info'" t-call="o-spreadsheet-Icon.CIRCLE_INFO"/> </div>
```
same below

rrahir (changes requested):

src/components/side_panel/find_and_replace/find_and_replace.css
```diff
@@ -19,7 +19,7 @@
     }
     .o-result-buttons {
       .o-button {
-        height: 19px;
+        height: 19px !important;
```
rrahir: apparently I was wrong,; could have sworn that forcing the height would do the trick but we need to override the behaviour of  the button class and the "min-height:32px" rule... sorry

rrahir (approved):
I updated the css to avoid jumps when a result appears.

robodoo r+

PR #8524, opened 2026-04-30
[FIX] Pivots: cache `getPivotIdFromPosition` for performance

rrahir:
the getter `getPivotIdFromPosition` can be heavily called, notably when a pivot sidepanel is open. This can become costy as we parse the tokens every time to find PIVOT functions.

This revision caches the values, which are only invalidated when a cell content is altered.

Task: 6159213

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_core_views/pivot_ui.ts
+++ b/src/plugins/ui_core_views/pivot_ui.ts
@@ -3,6 +3,7 @@ import { astToFormula } from "../../formulas/formula_formatter";
 import { toScalar } from "../../functions/helper_matrices";
 import { toBoolean } from "../../functions/helpers";
 import { deepCopy, deepEquals, getUniqueText, isDefined } from "../../helpers";
+import { PositionMap } from "../../helpers/cells/position_map";
 import {
   getNumberOfPivotFunctions,
   getPivotFunctions,
@@ -62,6 +63,8 @@ export class PivotUIPlugin extends CoreViewPlugin {
   private pivots: Record<UID, Pivot> = {};
   private unusedPivots?: UID[];
   private custom: UIPluginConfig["custom"];
+  private pivotPositionCache: PositionMap<UID[]> = new PositionMap();
+  private shouldInvalidateCache: boolean = false;
 
   constructor(config: CoreViewPluginConfig) {
     super(config);
@@ -79,9 +82,13 @@ export class PivotUIPlugin extends CoreViewPlugin {
 
   handle(cmd: Command) {
     if (invalidateEvaluationCommands.has(cmd.type)) {
+      this.shouldInvalidateCache = true;
       for (const pivotId of this.getters.getPivotIds()) {
         this.setupPivot(pivotId, { recreate: true });
       }
+    }
+    if (cmd.type === "UPDATE_CELL") {
+      this.shouldInvalidateCache = true;
     }
     switch (cmd.type) {
       case "REFRESH_PIVOT":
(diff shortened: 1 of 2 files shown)


rrahir:
robodoo rebase-ff

src/plugins/ui_core_views/pivot_ui.ts
```diff
@@ -70,10 +73,14 @@ export class PivotUIPlugin extends CoreViewPlugin {
...
   handle(cmd: Command) {
     if (invalidateEvaluationCommands.has(cmd.type)) {
+      this.pivotPositionCache = new PositionMap();
       for (const pivotId of this.getters.getPivotIds()) {
         this.setupPivot(pivotId, { recreate: true });
       }
     }
+    if (cmd.type === "UPDATE_CELL" && !this.shouldInvalidateCache) {
+      this.shouldInvalidateCache = true;
+    }
```
hokolomopo, line 92: why a different behaviour for update cells & for other invalidateEvaluationCommands ? Why do one at the finalize but not the other ? 

And not sur how much a problem it is (and not specific to this task), but if we call `getPivotIdFromPosition` between the invalidation of the cache and the finalize (so before the end of the evalution), we will fill the cache with wrong values. Might be worth to throw in those case.

src/plugins/ui_core_views/pivot_ui.ts
```diff
@@ -129,15 +143,22 @@ export class PivotUIPlugin extends CoreViewPlugin {
...
-        return pivotId && this.getters.getPivotId(pivotId);
+        const pivotFormulaId = pivotFunction.args[0]?.toString();
+        const pivotId = pivotFormulaId && this.getters.getPivotId(pivotFormulaId);
+        if (pivotId) {
+          this.pivotPositionCache.set(position, pivotId);
+        }
+        return pivotId;
       }
     }
     return undefined;
```
hokolomopo: the cache should be filled no matter the result, no ? Calling `getPivotIdFromPosition` on a non-pivot cell multiple time will not be cached there.

hokolomopo (commented):
Finally a cache 🙏 Do you think it'd be worth to cache other things in this plugin ? eg. getPivotCellFromPosition

I have the same exact comments for the Odoo PR with list :)

rrahir:
@hokolomopo 
> Do you think it'd be worth to cache other things in this plugin ?

in theory yes but I just went with the most problematic one at first.

src/plugins/ui_core_views/pivot_ui.ts
rrahir: I think it's more of a mistake to have different behaviours. just wanted to avoid recreating the map over and over again (I think?) as update_cell commands can be plenty back to back.

I agree that there can be issues while calling that getter before te evaluation is done but one could argue that calling anything that depends on the evaluation while it's dirty is incorrect. So maybe the getEvaluatedCell getter should throw ?

hokolomopo: > maybe the getEvaluatedCell getter should throw

we should try that sometime, we'll probably find some incorrect code 🙃

src/plugins/ui_core_views/pivot_ui.ts
```diff
@@ -79,10 +82,14 @@ export class PivotUIPlugin extends CoreViewPlugin {
...
   handle(cmd: Command) {
     if (invalidateEvaluationCommands.has(cmd.type)) {
+      this.shouldInvalidateCache = true;
       for (const pivotId of this.getters.getPivotIds()) {
         this.setupPivot(pivotId, { recreate: true });
       }
     }
+    if (cmd.type === "UPDATE_CELL") {
+      this.shouldInvalidateCache = true;
+    }
```
hokolomopo, line 92: not directly linked with this PR, but we don't call `setupPivot` on UPDATE_CELL ? Is that correct ?

rrahir: update_cell does not imply the creation of a pivot, so I think it's fair

hokolomopo:
@robodoo r+

PR #8527, opened 2026-04-30
[FIX] chart: cannot change chart type with gauge/scorecard

hokolomopo:
## Description

The chart type picker was using `chart.getSupportedChartTypes()` to check which charts the current chart can switch to. But this didn't work for gauge/scorecards, as they don't have a dataSource and `getSupportedChartTypes` would return an empty set.

Note: In master we should improve the architecture so this type of issue do not happens again. Either the gauge/scorecards should not have a dataSource key, and calls to that should crash, or they should use a functional dataSource.

Task: [6179286](https://www.odoo.com/odoo/2328/tasks/6179286)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/chart/chart_type_picker/chart_type_picker.ts
+++ b/src/components/side_panel/chart/chart_type_picker/chart_type_picker.ts
@@ -1,10 +1,12 @@
 import { Component, useExternalListener, useRef, useState } from "@odoo/owl";
+import { chartDataSourceRegistry } from "../../../../registries/chart_data_source_registry";
 import { chartSubtypeRegistry } from "../../../../registries/chart_subtype_registry";
+import { CHART_TYPES, ChartDefinition, ChartType } from "../../../../types/chart/chart";
 import {
   chartCategories,
   ChartSubtypeProperties,
 } from "../../../../types/chart_subtype_properties";
-import { ChartDefinition, ChartType, UID } from "../../../../types/index";
+import { UID } from "../../../../types/index";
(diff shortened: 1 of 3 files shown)


pro-odoo:
> Note: In master we should improve the architecture so this type of issue do not happens again. Either the gauge/scorecards should not have a dataSource key, and calls to that should crash, or they should use a functional dataSource.

+1000

src/components/side_panel/chart/chart_type_picker/chart_type_picker.ts
```diff
@@ -52,6 +55,24 @@ export class ChartTypePicker extends Component<Props, SpreadsheetChildEnv> {
...
 
+  private getSupportedChartTypes(definition: ChartDefinition): Set<ChartType> {
+    let supportedTypes: Set<ChartType>;
+
+    if (definition.dataSource) {
+      const dataSourceBuilder = chartDataSourceRegistry.get(definition.dataSource.type);
+      supportedTypes = new Set(dataSourceBuilder.supportedChartTypes);
+    } else if (
+      !definition.dataSource &&
+      (definition.type === "scorecard" || definition.type === "gauge")
```
dhrp-odoo, line 66: 👋
ICYDK, in master, you’ll need to add `bubble chart` as well :)

pro-odoo:
robodoo r+

PR #8533, opened 2026-05-01
[FIX] range: correctly handle unbounded ranges on row/col changes

rmbh-odoo:
## Description:

Steps to reproduce:
- Create a sheet with 3 rows
- Set `A1 = 1`, `A2 = 2`, and `C1 = SUM(A:A)`
- Delete row 3
- Add a new row at the bottom
- Set `A3 = 6`

Issue:
- Unbounded ranges (e.g. `A:A`) are not properly updated on row/col changes.
  The changeType can be incorrectly computed as `NONE` or `MOVE` instead
  of `RESIZE`, causing formulas to miss newly added cells (e.g. `A3` in `SUM`).

Cause:
- Incomplete condition handling for unbounded ranges caused inconsistent
  changeType detection across different scenarios.

Fix:
- Refine changeType computation for unbounded ranges by covering all edge
  cases and ensuring correct `RESIZE` behavior when the range should expand.

Task: [6167358](https://www.odoo.com/odoo/2328/tasks/6167358)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/range.ts
+++ b/src/helpers/range.ts
@@ -437,8 +437,20 @@ function getApplyRangeChangeAddColRow(cmd: AddColumnsRowsCommand): ApplyRangeCha
     if (range.sheetId !== cmd.sheetId) {
       return { changeType: "NONE", range };
     }
+
+    const isUnboundedAtEnd = range.unboundedZone[end] === undefined;
+
+    // Full column/row ranges (e.g. A:A) are anchored at the sheet origin,
+    // so any insertion only expands them (RESIZE), never moves them.
+    if (isUnboundedAtEnd && !range.unboundedZone.hasHeader) {
+      return {
+        changeType: "RESIZE",
+        range: createAdaptedRange(range, dimension, "RESIZE", cmd.quantity),
+      };
+    }
+
     if (cmd.position === "after") {
-      if (range.zone[start] <= cmd.base && cmd.base < range.zone[end]) {
+      if (range.zone[start] <= cmd.base && (cmd.base < range.zone[end] || isUnboundedAtEnd)) {
         return {
           changeType: "RESIZE",
           range: createAdaptedRange(range, dimension, "RESIZE", cmd.quantity),
--- a/tests/range_plugin.test.ts
+++ b/tests/range_plugin.test.ts
@@ -1,5 +1,5 @@
 import { CorePlugin, coreTypes, Model } from "../src";
-import { duplicateRangeInDuplicatedSheet } from "../src/helpers";
+import { duplicateRangeInDuplicatedSheet, numberToLetters } from "../src/helpers";
 import { corePluginRegistry } from "../src/plugins";
 import { Command, Range, RangeAdapterFunctions } from "../src/types";
 import { CellErrorType } from "../src/types/errors";
@@ -35,7 +35,7 @@ coreTypes.add("USE_RANGE");
 coreTypes.add("USE_TRANSIENT_RANGE");
 
 class PluginTestRange extends CorePlugin {
-  static getters = ["getUsedRanges", "getRanges"];
+  static getters = ["getUsedRanges", "getRanges", "getRangeZoneEdge"];
 
   ranges: Range[] = [];
 
@@ -77,6 +77,10 @@ class PluginTestRange extends CorePlugin {
 
   getRanges() {
     return this.ranges;
+  }
+
+  getRangeZoneEdge(edge, index = 0) {
+    return this.ranges[index].zone[edge];
   }
 }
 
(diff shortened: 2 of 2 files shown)


src/helpers/range.ts
```diff
@@ -59,21 +59,29 @@ interface RangeXcArgs {
...
     parts = [args.parts[0], args.parts[0]];
-  } else if (args.parts.length === 2 && getZoneArea(zone) === 1) {
+  } else if (args.parts.length === 2 && getZoneArea(initialZone) === 1) {
     parts = [args.parts[0]];
   }
-  return {
+  const range = {
     unboundedZone,
-    zone,
     parts,
```
rmbh-odoo, line 72: I'm not entirely sure whether `parts` should also be dynamic. Since `parts` are derived from the `zone`, they logically should update when the `zone` changes.
However, at the moment, I’m not observing any side effects or issues with `parts` being static.

src/helpers/range.ts
```diff
@@ -59,21 +59,29 @@ interface RangeXcArgs {
...
     prefixSheet: args.prefixSheet,
     invalidSheetName: args.invalidSheetName,
     sheetId: args.sheetId,
-  };
+  } as Range;
+
+  Object.defineProperty(range, "zone", {
```
dhrp-odoo: I like that idea but there are still few bugs.

`createRange()` now makes `range.zone` dynamic, but `createAdaptedRange()` still rebuilds ranges with object spread (and there are other places as well)

To reproduce here are the steps:

- Start with a sheet that has 3 rows.
- Put =1 in A1.
- Put =2 in A2.
- Put =SUM(A:A) in C1.
- Delete row 3.
- Add 1 row at the bottom.
- Put =6 in A3.

dhrp-odoo (commented):
👋

src/helpers/range.ts
rmbh-odoo: I've now introduced a cloneRange helper and replaced object spread usage, so the zone getter is preserved.

src/helpers/range.ts
```diff
@@ -178,7 +186,7 @@ export function duplicateRangeInDuplicatedSheet(
   range: Range
 ): Range {
   const sheetId = range.sheetId === sheetIdFrom ? sheetIdTo : range.sheetId;
-  return { ...range, sheetId };
+  return cloneRange(range, { sheetId });
```
hokolomopo: So maybe switching from a `Range` class to a POJO was not a good idea ... We're back at the start with `clone` methods, but now we aren't type-safe, and nothing prevents us from spreading a range to copy it ...

cc @LucasLefevre

tests/range_plugin.test.ts
```diff
@@ -667,6 +667,20 @@ describe("full column range", () => {
     addRows(m, "before", 0, 1);
     expect(m.getters.getUsedRanges()[1]).toEqual("B2:C");
   });
+  test("full-column range updates correctly on row insertions and deletions", () => {
```
hokolomopo:
```suggestion
});

  test("full-column range updates correctly on row insertions and deletions", () => {
```
nitpick. same for the other test

hokolomopo (commented):
@LucasLefevre

src/helpers/range.ts
LucasLefevre: This is magic, like it was before with the `Range` facade and `RangeImpl`.
Code shouldn't be magic 😉 

The POJO has no magic at all... at the cost of having two sources of truth (`unboundedZone` and `zone`) that can lead to bugs such as this one.

I'd much rather prefer to fix it with something like this 👇 easy, more explicit, no magic.

(other cases needs to be fixed in similar ways)

```patch
diff --git a/src/helpers/range.ts b/src/helpers/range.ts
index bd37d1321..8a2c2b762 100644
--- a/src/helpers/range.ts
+++ b/src/helpers/range.ts
@@ -438,7 +438,7 @@ function getApplyRangeChangeAddColRow(cmd: AddColumnsRowsCommand): ApplyRangeCha
       return { changeType: "NONE", range };
     }
     if (cmd.position === "after") {
-      if (range.zone[start] <= cmd.base && cmd.base < range.zone[end]) {
+      if (range.zone[start] <= cmd.base && (cmd.base < range.zone[end] || range.unboundedZone[end] === undefined)) {
         return {
           changeType: "RESIZE",
           range: createAdaptedRange(range, dimension, "RESIZE", cmd.quantity),
```

src/helpers/range.ts
rmbh-odoo: We remove the dynamic zone and find another way to consider the POJO approach :smile:

tests/range_plugin.test.ts
rmbh-odoo: In this describe block, none of the test cases have spacing, so I kept it consistent and didn’t add any new line :sob: 
Do you think it’s worth adding new lines for all test cases in a separate commit, wdyt?

LucasLefevre: Not worth it IMO ;)
Keep it consistent 👍

src/helpers/range.ts
rmbh-odoo: I think I’ve covered all possible edge cases, so the ranges shouldn’t break anymore :slightly_smiling_face:

rrahir:
robodoo r+

PR #8539, opened 2026-05-04
[FIX] pivot: unused pivot detection 

hokolomopo:
### [FIX] pivot: unused pivot detection with composed formula

If a formula has two pivots (eg. `=PIVOT(1) + PIVOT(2)`), the
second pivot will not be detected as used.

### [FIX] pivot: unused pivot detection with calculated measure

If a pivot has a calculated measure that refers to a second pivot,
the second pivot is not detected as used. This commit fixes that.

Note: the fix does not fix 100% of the issues. A pivot only referenced
in other places (eg. in a CF rule formula) will still not be
detected as used. But checking every other place the pivot can be
referenced is expensive and error-prone. That may be done in master.

Task: [6105894](https://www.odoo.com/odoo/2328/tasks/6105894)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/pivot/pivot_highlight.ts
+++ b/src/helpers/pivot/pivot_highlight.ts
@@ -15,8 +15,8 @@ function getVisiblePivotCellPositions(getters: Getters, pivotId: UID) {
   for (const col of getters.getSheetViewVisibleCols()) {
     for (const row of getters.getSheetViewVisibleRows()) {
       const position = { sheetId, col, row };
-      const cellPivotId = getters.getPivotIdFromPosition(position);
-      if (pivotId === cellPivotId) {
+      const cellPivotIds = getters.getPivotIdsFromPosition(position);
+      if (cellPivotIds.includes(pivotId)) {
         positions.push(position);
       }
     }
--- a/src/plugins/ui_core_views/pivot_ui.ts
+++ b/src/plugins/ui_core_views/pivot_ui.ts
@@ -2,10 +2,10 @@ import { Token } from "../../formulas";
 import { astToFormula } from "../../formulas/parser";
 import { toScalar } from "../../functions/helper_matrices";
 import { toBoolean } from "../../functions/helpers";
-import { deepCopy } from "../../helpers";
+import { deepCopy, isDefined } from "../../helpers";
 import {
-  getFirstPivotFunction,
   getNumberOfPivotFunctions,
+  getPivotFunctions,
 } from "../../helpers/pivot/pivot_composer_helpers";
 import withPivotPresentationLayer from "../../helpers/pivot/pivot_presentation";
 import { pivotRegistry } from "../../helpers/pivot/pivot_registry";
@@ -15,9 +15,11 @@ import { _t } from "../../translation";
 import {
   AddPivotCommand,
   CellPosition,
+  CellValue,
   Command,
   CoreCommand,
   FunctionResultObject,
+  Matrix,
   PivotCoreMeasure,
   PivotTableCell,
   RangeCompiledFormula,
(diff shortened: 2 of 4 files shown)


tests/pivots/pivot_plugin.test.ts
```diff
@@ -417,5 +417,23 @@ describe("Pivot plugin", () => {
...
+      const model = createModelWithPivot("A1:I5");
+      addPivot(model, "A1:I5", { name: "Unused Pivot" }, "2");
+      setCellContent(model, "A40", "=PIVOT(1) + PIVOT(2)");
+
+      expect(model.getters.isPivotUnused("1")).toBe(false);
+      expect(model.getters.isPivotUnused("2")).toBe(false);
+    });
+  });
+
+  test("Pivot highlights works with multiple pivot formulas in the same cell", () => {
```
LucasLefevre, line 432: this test is duplicated with the one above

src/plugins/ui_core_views/pivot_ui.ts
```diff
@@ -120,24 +129,37 @@ export class PivotUIPlugin extends UIPlugin {
...
-        position.sheetId,
-        cell.compiledFormula.tokens
-      );
-      if (pivotFunction) {
-        const pivotId = pivotFunction.args[0]?.toString();
-        return pivotId && this.getters.getPivotId(pivotId);
-      }
+      return this.getPivotIdsFromPosition(position)[0];
     }
     return undefined;
```
LucasLefevre: can be simplified to 

```ts
getPivotIdFromPosition(position: CellPosition): UID | undefined {
  return this.getPivotIdsFromPosition(position)[0];
}
```

tests/pivots/pivot_plugin.test.ts
hokolomopo: I wrote the title of the test, then forgot to actually write the test AND to fix the highlights 🤦

src/helpers/pivot/pivot_highlight.ts
```diff
@@ -15,8 +15,8 @@ function getVisiblePivotCellPositions(getters: Getters, pivotId: UID) {
   for (const col of getters.getSheetViewVisibleCols()) {
     for (const row of getters.getSheetViewVisibleRows()) {
       const position = { sheetId, col, row };
-      const cellPivotId = getters.getPivotIdFromPosition(position);
-      if (pivotId === cellPivotId) {
+      const cellPivotId = getters.getPivotIdsFromPosition(position);
+      if (cellPivotId.includes(pivotId)) {
```
LucasLefevre:
```suggestion
      const cellPivotIds = getters.getPivotIdsFromPosition(position);
      if (cellPivotIds.includes(pivotId)) {
```

src/plugins/ui_core_views/pivot_ui.ts
```diff
@@ -328,6 +354,25 @@ export class PivotUIPlugin extends UIPlugin {
         }
       }
     }
+
+    for (const pivotId of this.getters.getPivotIds()) {
+      const pivot = this.getters.getPivot(pivotId);
+      for (const measure of pivot.definition.measures) {
+        if (measure.computedBy) {
+          const { sheetId } = measure.computedBy;
+          const formula = this.getters.getMeasureCompiledFormula(pivotId, measure);
```
LucasLefevre, line 363: We'll hit a limit someday and it'll need another approach to 'standalone' formulas. Calculated measures are checked here, fine, but what about custom formulas in CF/DV ? calculated field for odoo lists in master?

LucasLefevre (commented):
robodoo r+

LucasLefevre:
robodoo rebase-ff

src/plugins/ui_core_views/pivot_ui.ts
hokolomopo: Yeah it's in master https://github.com/odoo/o-spreadsheet/pull/8556

PR #8553, opened 2026-05-04
[FIX] color_picker: prevent gradient from opening when picking a cust…

pro-odoo:
…om color

Clicking a custom color swatch bubbled up to the parent toggler and opened the gradient picker instead of just selecting the color. Stop the click propagation on the custom color item so it only triggers the color pick.

Task: 6186050

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/color_picker/color_picker.xml
+++ b/src/components/color_picker/color_picker.xml
@@ -39,7 +39,7 @@
             class="o-color-picker-line-item"
             t-att-data-color="color"
             t-attf-style="background-color:{{color}};"
-            t-on-click="() => this.onColorClick(color)">
+            t-on-click.stop="() => this.onColorClick(color)">
             <div
               t-if="isSameColor(props.currentColor, color)"
               align="center"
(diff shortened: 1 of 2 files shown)


src/components/color_picker/color_picker.xml
LucasLefevre, line 29: shouldn't `t-on-click.stop="toggleColorPicker"` be moved on the PLUS icon div instead ?

pro-odoo: There are two `t-on-click` to open the color picker, one on the section name (L26) which handle the entire line of `Custom`, and this one to handle the entire zone below. Currently, if you click on the right of the last color icon, the color picker is opened. If I moved it on the PLUS icon, it will not be the same behavior anymore.
Image

Personally, I don't care at all, anyway I still need to add `.stop` L42 as the click should not be propagated.

LucasLefevre (approved):
robodoo r+

PR #8556, opened 2026-05-05
[IMP] pivot: add side panel for unused pivots

hokolomopo:
## Description

This commit replace the `Unused pivot` warning in the `Data` topbar menu by a dedicated side panel. Checking which pivot is unused is an expansive operation, and is not frankly important, removing the pivots only cleans up the JSON. Doing it every time we open the menu can make the menu very slow on big spreadsheets.

There was also a problem on how we were checking which pivot was unused. Only the formulas in the cells were checked, not the formulas of cfs/charts/... This commit introduce a new `FormulaProvider` to the core plugins, so they can register their formulas.

Task: [6185119](https://www.odoo.com/odoo/2328/tasks/6185119)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/data_sources_cleanup/data_sources_cleanup.ts
+++ b/src/components/side_panel/data_sources_cleanup/data_sources_cleanup.ts
@@ -1,0 +1,53 @@
+import { proxy } from "@odoo/owl";
+import { Component } from "../../../owl3_compatibility_layer";
+import { useLocalStore } from "../../../store_engine/store_hooks";
+import { UID } from "../../../types/misc";
+import { SpreadsheetChildEnv } from "../../../types/spreadsheet_env";
+import { Store } from "../../../types/store_engine";
+import { Checkbox } from "../components/checkbox/checkbox";
+import { Section } from "../components/section/section";
+import { DataSourcesCleanupStore } from "./data_sources_cleanup_store";
+
+interface State {
+  uncheckedDataSources: Record<UID, boolean>;
+}
+
+export class DataSourcesCleanup extends Component<SpreadsheetChildEnv> {
+  static template = "o-spreadsheet-DataSourcesCleanup";
+  static components = { Section, Checkbox };
+
+  state = proxy<State>({
+    uncheckedDataSources: {},
+  });
+
+  store!: Store<DataSourcesCleanupStore>;
+
+  setup() {
+    this.store = useLocalStore(DataSourcesCleanupStore);
+  }
+
+  onToggleDataSource(type: string, id: UID, checked: boolean) {
+    this.state.uncheckedDataSources[`${type}-${id}`] = !checked;
+  }
+
+  isDataSourceChecked(type: string, id: UID) {
+    return !this.state.uncheckedDataSources[`${type}-${id}`];
+  }
+
+  get dataSourcesToDelete() {
+    const dataSourcesToDelete: { type: string; id: UID }[] = [];
+    for (const category of this.store.unusedDataSourcesCategories) {
+      for (const dataSource of category.dataSources) {
+        const stateKey = `${category.type}-${dataSource.id}`;
+        if (!this.state.uncheckedDataSources[stateKey]) {
+          dataSourcesToDelete.push({ type: category.type, id: dataSource.id });
+        }
+      }
+    }
+    return dataSourcesToDelete;
+  }
+
+  onRemoveUnusedDataSources() {
+    this.env.model.dispatch("DELETE_DATA_SOURCES", { dataSources: this.dataSourcesToDelete });
+  }
+}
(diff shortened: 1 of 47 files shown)


src/plugins/core_plugin.ts
```diff
@@ -30,15 +33,23 @@ export interface CorePluginConstructor {
...
+    getters,
+    stateObserver,
+    range,
+    dispatch,
+    canDispatch,
+    formulasPlugin,
+  }: CorePluginConfig) {
     super(stateObserver);
     range.addRangeProvider(this.adaptRanges.bind(this));
+    formulasPlugin.addFormulaProvider(this.getFormulas.bind(this));
```
hokolomopo, line 52: maybe we should have something generic ?

src/registries/data_source_registry.ts
```diff
@@ -0,0 +1,25 @@
...
+import { UID } from "../types/misc";
+import { Registry } from "./registry";
+
+interface DataSourceType {
+  type: string;
+  unusedLabel: string;
+  getAllIds: (getters: Getters) => UID[];
+  isDataSourceUnused: (getters: Getters, id: UID) => boolean;
+  getDisplayName: (getters: Getters, id: UID) => string;
+  deleteItem: (dispatch: Model["dispatch"], id: UID) => void;
```
rrahir:
```suggestion
  deleteDataSource: (dispatch: Model["dispatch"], id: UID) => void;
```

src/registries/data_source_registry.ts
```diff
@@ -0,0 +1,25 @@
+import { Model } from "../model";
+import { _t } from "../translation";
+import { Getters } from "../types/getters";
+import { UID } from "../types/misc";
+import { Registry } from "./registry";
+
+interface DataSourceType {
+  type: string;
+  unusedLabel: string;
+  getAllIds: (getters: Getters) => UID[];
```
rrahir: there's no need to have that many properties. a simple "getUnusedInstances" that returns all the
 necessary information would probably do the trick. You already have an "unusedLabel" that clearly orients the use of that registry anyway.
I think I feel quite odd with this registry being so similar to globalFieldMatchingRegistry in odoo

src/model.ts
```diff
@@ -184,6 +186,8 @@ export class Model extends EventBus<any> implements CommandDispatcher {
     this.coreGetters.copyFormulaStringForSheet = this.range.copyFormulaStringForSheet.bind(
       this.range
     );
+    this.formulasPlugin = new FormulaPlugin();
```
rrahir: doesn't need to be a plugin at all. it should only expose the register method and the getter but it does not participate in the model lifecycle in any way. For all we care, the list of formula providers could be stored in model directly when we register them

src/components/side_panel/data_sources_cleanup/data_sources_cleanup.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  onToggleItem(type: string, id: UID, checked: boolean) {
+    this.state.uncheckedItems[`${type}-${id}`] = !checked;
+  }
+
+  isItemChecked(type: string, id: UID) {
+    return !this.state.uncheckedItems[`${type}-${id}`];
+  }
+
+  get itemsToDelete() {
```
rrahir: items is a bit too generic (I know I introduced some), here it clearly targets datasources so we might call them that way?

src/registries/data_source_registry.ts
```diff
@@ -0,0 +1,25 @@
...
+  isDataSourceUnused: (getters: Getters, id: UID) => boolean;
+  getDisplayName: (getters: Getters, id: UID) => string;
+  deleteItem: (dispatch: Model["dispatch"], id: UID) => void;
+}
+
+export const dataSourceRegistry = new Registry<DataSourceType>();
```
rrahir:
```suggestion
export const unsuedDataSourceRegistry = new Registry<DataSourceType>();
```

rrahir (changes requested):

src/plugins/ui_core_views/pivot_ui.ts
```diff
@@ -50,6 +51,7 @@ export class PivotUIPlugin extends CoreViewPlugin {
     "getPivot",
     "getFirstPivotFunction",
     "getPivotCellSortDirection",
+    "getPivotIdsFromFormula",
```
rrahir: does not seem used as a getter

rrahir (commented):
Should we consider removing the secondary icons code ? maybe in another commit so that we can add it back if necessary

src/plugins/ui_core_views/pivot_ui.ts
hokolomopo: True. I also removed the secondary icons in another commit

pro-odoo (approved):
robodoo delegate+

hokolomopo:
@robodoo r+

hokolomopo:
@robodoo rebase-ff

PR #8559, opened 2026-05-05
[FIX] named range selector: select hidden named range

matho-odoo:
Before this commit:
If the zone is hidden, nothing happens when selecting a named range from the dropdown. And if the sheet is hidden, the zone is selected in the active sheet, which is not the expected behavior.

After this commit:
When selecting a named range from the dropdown, the zone is selected even if it's hidden. And if the sheet is hidden, nothing is selected and the user is notified that the sheet is hidden.

Task: [6171849](https://www.odoo.com/odoo/2328/tasks/6171849)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/internal_viewport.ts
+++ b/src/helpers/internal_viewport.ts
@@ -177,6 +177,10 @@ export class InternalViewport {
     this.adjustViewportZoneY();
   }
 
+  isZoneVisible(zone: Zone): boolean {
+    return intersection(zone, this) !== undefined;
+  }
+
   /**
    *
    * Computes the visible coordinates & dimensions of a given zone inside the viewport
--- a/src/helpers/viewport_collection.ts
+++ b/src/helpers/viewport_collection.ts
@@ -231,6 +231,10 @@ export class ViewportCollection {
     return this.getSubViewports(sheetId).some((pane) => pane.isVisible(col, row));
   }
 
+  isZoneVisibleInViewport(sheetId: UID, zone: Zone): boolean {
+    return this.getSubViewports(sheetId).some((pane) => pane.isZoneVisible(zone));
+  }
+
   getScrollBarWidth(): Pixel {
     return SCROLLBAR_WIDTH / this.zoomLevel;
   }
--- a/src/plugins/ui_stateful/selection.ts
+++ b/src/plugins/ui_stateful/selection.ts
@@ -894,11 +894,27 @@ export class GridSelectionPlugin extends UIPlugin {
     ctx.strokeStyle = SELECTION_BORDER_COLOR;
     ctx.lineWidth = 1.5 * thinLineWidth;
     for (const zone of zones) {
+      if (!viewports.isZoneVisibleInViewport(sheetId, zone)) {
+        continue;
+      }
       const { x, y, width, height } = viewports.getVisibleRect(sheetId, zone);
       ctx.globalCompositeOperation = "multiply";
-      ctx.fillRect(x, y, width, height);
-      ctx.globalCompositeOperation = "source-over";
-      ctx.strokeRect(x, y, width, height);
+      const currentLineWidth = ctx.lineWidth;
+      if (height === 0 || width === 0) {
+        ctx.lineWidth = 3 * thinLineWidth;
+      }
+      if (height === 0 && width === 0) {
+        ctx.beginPath();
+        ctx.arc(x, y, 3, 0, 2 * Math.PI);
+        ctx.fill();
+        ctx.globalCompositeOperation = "source-over";
+        ctx.stroke();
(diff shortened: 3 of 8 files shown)


src/plugins/ui_stateful/selection.ts
```diff
@@ -896,9 +896,20 @@ export class GridSelectionPlugin extends UIPlugin {
     for (const zone of zones) {
       const { x, y, width, height } = viewports.getVisibleRect(sheetId, zone);
       ctx.globalCompositeOperation = "multiply";
-      ctx.fillRect(x, y, width, height);
-      ctx.globalCompositeOperation = "source-over";
-      ctx.strokeRect(x, y, width, height);
+      if (height === 0 || width === 0) {
+        ctx.lineWidth = 3 * thinLineWidth;
```
fdamhaut, line 904: This context change is not reset.
If there is multiple zones the first one might affect the following ones

fdamhaut (changes requested):

src/plugins/ui_stateful/selection.ts
```diff
@@ -896,9 +896,21 @@ export class GridSelectionPlugin extends UIPlugin {
...
+      }
+      if (height === 0 && width === 0) {
```
rrahir, line 906: both the change of lineWidth & drawing can be done in the same if statement.

Also, there is a subtlety,   you assume that the value of lineWidth is `1.5 * thinLineWidth;` when you first enter this function but this is not something that is set in stone.

Generally, we can use https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/save to save the state (the different property values) then cahnge the state as we want, for instance here changing the lineWidth, then use `ctx.restore()` to reset the state as it was before we changed everything.

Ctx.save() saves the WHOLE state though. Since we only change 2 properties here (globalCompositeOperation and lineWidth, we can simply set them to their first value once we are done drawing.

src/plugins/ui_stateful/selection.ts
```diff
@@ -896,9 +896,21 @@ export class GridSelectionPlugin extends UIPlugin {
     for (const zone of zones) {
       const { x, y, width, height } = viewports.getVisibleRect(sheetId, zone);
       ctx.globalCompositeOperation = "multiply";
-      ctx.fillRect(x, y, width, height);
-      ctx.globalCompositeOperation = "source-over";
-      ctx.strokeRect(x, y, width, height);
+      if (height === 0 || width === 0) {
+        ctx.lineWidth = 3 * thinLineWidth;
```
rrahir, line 904:
```suggestion
        const currentLineWidth = ctx.lineWidth;
        ctx.lineWidth = 3 * thinLineWidth;
```

src/plugins/ui_stateful/selection.ts
```diff
@@ -896,9 +896,21 @@ export class GridSelectionPlugin extends UIPlugin {
...
+        ctx.arc(x, y, 3, 0, 2 * Math.PI);
+        ctx.fill();
+        ctx.globalCompositeOperation = "source-over";
+        ctx.stroke();
+      } else {
+        ctx.fillRect(x, y, width, height);
+        ctx.globalCompositeOperation = "source-over";
+        ctx.strokeRect(x, y, width, height);
+      }
+      ctx.lineWidth = 1.5 * thinLineWidth;
```
rrahir:
```suggestion
      ctx.lineWidth = currentLineWidth
```

rrahir (changes requested):

src/helpers/internal_viewport.ts
```diff
@@ -177,6 +177,10 @@ export class InternalViewport {
     this.adjustViewportZoneY();
   }
 
+  isZoneInsideViewport(zone: Zone): boolean {
```
rrahir: the zone is alsways inside the viewport ^^ 
```suggestion
  isZoneVisible(zone: Zone): boolean {
```

src/helpers/viewport_collection.ts
```diff
@@ -231,6 +231,10 @@ export class ViewportCollection {
     return this.getSubViewports(sheetId).some((pane) => pane.isVisible(col, row));
   }
 
+  isZoneVisible(sheetId: UID, zone: Zone): boolean {
```
rrahir:
```suggestion
  isZoneVisibleInViewport(sheetId: UID, zone: Zone): boolean {
```

rrahir (commented):
small nitpick to align the naming with the methods that check the visibility of a single position

rrahir:
robodoo r+

PR #8560, opened 2026-05-05
[IMP] Statistics: Add first format to bottombar

fdamhaut:
Task: 6186747

## Description:

Task: [6189449](https://www.odoo.com/odoo/2328/tasks/6189449)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/bottom_bar/bottom_bar_statistic/bottom_bar_statistic.ts
+++ b/src/components/bottom_bar/bottom_bar_statistic/bottom_bar_statistic.ts
@@ -46,8 +46,8 @@ export class BottomBarStatistic extends Component<Props, SpreadsheetChildEnv> {
     ) {
       return undefined;
     }
-    if (this.state.selectedStatisticFn === "") {
-      this.state.selectedStatisticFn = Object.keys(this.store.statisticFnResults)[0];
+    if (!this.store.statisticFnResults[this.state.selectedStatisticFn]) {
+      return this.getComposedFnName(Object.keys(this.store.statisticFnResults)[0]);
     }
     return this.getComposedFnName(this.state.selectedStatisticFn);
   }
@@ -74,10 +74,9 @@ export class BottomBarStatistic extends Component<Props, SpreadsheetChildEnv> {
   private getComposedFnName(fnName: string): string {
     const locale = this.env.model.getters.getLocale();
     const fnValue = this.store.statisticFnResults[fnName];
-    return (
-      fnName +
-      ": " +
-      (fnValue?.value !== undefined ? formatValue(fnValue.value(), { locale }) : "__")
-    );
+    if (!fnValue?.value) {
+      return "";
+    }
+    return fnName + ": " + formatValue(fnValue.value(), { locale, format: fnValue.format() });
   }
 }
--- a/src/components/side_panel/column_stats/column_stats_store.ts
+++ b/src/components/side_panel/column_stats/column_stats_store.ts
@@ -1,6 +1,6 @@
 import { sum } from "../../../functions/helper_math";
 import { average, max, median, min } from "../../../functions/helper_statistical";
-import { formatValue } from "../../../helpers/format/format";
+import { formatValue, isDateTimeFormat } from "../../../helpers/format/format";
 import { isDefined } from "../../../helpers/misc";
 import {
   computeStatisticFnResults,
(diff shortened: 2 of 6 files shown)


src/helpers/selection_statistic_functions.ts
```diff
@@ -16,7 +17,8 @@ export interface SelectionStatisticFunction {
   name: string;
   compute: (data: EvaluatedCell[], locale: Locale) => number | string;
   types: CellValueType[];
-  format?: string;
+  visible?: (data: EvaluatedCell[], locale: Locale) => boolean;
+  computeFormat: (data: EvaluatedCell[], locale: Locale) => Format;
```
LucasLefevre: nitpick: `data` -> `cells` or `evaluatedCells`

src/components/side_panel/column_stats/column_stats_store.ts
```diff
@@ -32,32 +32,40 @@ const columnStatisticFunctions: SelectionStatisticFunction[] = [
...
     },
-    format: "0",
+    computeFormat: (values, locale) => "0",
   },
   {
     name: _t("Sum"),
     types: [CellValueType.number],
     compute: (values, locale) => sum([[values]], locale),
+    visible: (values, locale) =>
+      values.some((cell) => !cell.format || !isDateTimeFormat(cell.format)),
```
LucasLefevre, line 42: Should both sides (bottom bar and stats side panel) be unified/factorized?

src/components/side_panel/column_stats/column_stats_store.ts
```diff
@@ -32,32 +32,40 @@ const columnStatisticFunctions: SelectionStatisticFunction[] = [
...
     name: _t("Median"),
     types: [CellValueType.number],
     compute: (values, locale) => median([[values]], locale) ?? "",
+    computeFormat: (values, locale) => values[0]?.format ?? "0.00",
   },
   {
     name: _t("Minimum value"),
     types: [CellValueType.number],
     compute: (values, locale) => min([[values]], locale).value,
+    computeFormat: (values, locale) => values[0]?.format ?? "0.00",
```
LucasLefevre, line 62: why "0.00" here and no default format in the bottom bar ? Is it intended ?

src/components/bottom_bar/bottom_bar_statistic/bottom_bar_statistic.ts
```diff
@@ -77,7 +79,9 @@ export class BottomBarStatistic extends Component<Props, SpreadsheetChildEnv> {
     return (
       fnName +
       ": " +
-      (fnValue?.value !== undefined ? formatValue(fnValue.value(), { locale }) : "__")
+      (fnValue?.value !== undefined
+        ? formatValue(fnValue.value(), { locale, format: fnValue.format() })
+        : "__")
```
LucasLefevre: can this be simplified since we no longer display "__" ?

src/components/side_panel/column_stats/column_stats_store.ts
fdamhaut: Yes, it's the current behaviour and I wanted to keep it as is.

src/components/side_panel/column_stats/column_stats_store.ts
fdamhaut: Same reasons as the column format (keeping distinct formats)

rrahir (changes requested):
I think we are missing a boolean "defaultStatisticsSeclected"  that would indicate to recompute the valid measure to display when we change the selection, because right now, we will select the visible one based on the very first selection and it ends up as the selected statistics.
E.g.
I have "coucou" as the content of A1, when I load the spreadsheet, the bottom statistics now shows the "Count" measure. but then if I move my selection to the full column B and that column contains numbers, the measure stays on "Count" but I would expect to see the "Sum" measure.

fdamhaut:
> I think we are missing a boolean "defaultStatisticsSeclected" that would indicate to recompute the valid measure to display when we change the selection, because right now, we will select the visible one based on the very first selection and it ends up as the selected statistics. E.g. I have "coucou" as the content of A1, when I load the spreadsheet, the bottom statistics now shows the "Count" measure. but then if I move my selection to the full column B and that column contains numbers, the measure stays on "Count" but I would expect to see the "Sum" measure.

I just removed setting the state if there is no currently selected value (we now always display the 1st value and don't set it) this should act as the same behaviour as you proposed

rrahir (approved):
robodoo r+

PR #8568, opened 2026-05-06
[FIX] grid: wrong z-index for grid elements

hokolomopo:
## Description

With the new dark mode for the grid, the canvas is applied a css `filter` property. This creates a new stacking context for the canvas, so elements that were previously above it were now below it. Some `z-index` values were added, but they were missing for some elements.

- Added a css variable for the grid overlay `z-index`
- Added `z-index` to autofill and table resizer so they are at the same level as the grid overlay.
- `Add row footer` button `z-index` was useless since it's a child of the grid overlay
- Highlight had a custom `z-index` ... which was not applied since the highlight element had `position: static`. The `z-index` should be applied to its children.

Task: [0](https://www.odoo.com/odoo/2328/tasks/0)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/autofill/autofill.css
+++ b/src/components/autofill/autofill.css
@@ -11,6 +11,7 @@
     position: absolute;
     height: var(--os-autofill-edge-length);
     width: var(--os-autofill-edge-length);
+    z-index: var(--os-components-importance-grid-overlay);
     &:hover {
       cursor: crosshair;
     }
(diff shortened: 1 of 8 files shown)


rrahir:
robodoo r+

PR #8579, opened 2026-05-07
[FIX] composer: remove forced reflow in content editable

hokolomopo:
## Description

With the introduction of dark mode, we removed the `toHex` in the color comparison of `compareContentToSpanElement`, because it didn't handle `light-dark()` colors. But using simple color string comparison is wrong, because the browser will often change color formats (e.g. hex values will be converted to rgb strings).

The method `compareContentToSpanElement` would also use `innerText` to compare text, but `innerText` is style-awayre and might trigger a reflow. Which can be a performance bottleneck.

This commit:
- make `toHex` handle `light-dark()` colors, so we can use it in the content editable helper
- change `innerText` to `textContent` to avoid forced reflow when the content actually changed

Benchmark:
---------------

Performance in Odoo (that has an expansive style/layout computation) on a slow-ish CPU when hovering an autocomplete pivot item in the formula `=1+1+1+1+1+1+1+1+1+1+1+PIVOT()`:

Before:
- Animation frame: ~92ms
- compareContentToSpanElement: ~79 ms

After:
- Animation frame: ~4ms
- compareContentToSpanElement: ~0.35

Task: [6199661](https://www.odoo.com/odoo/2328/tasks/6199661)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/color.ts
+++ b/src/helpers/color.ts
@@ -47,6 +47,8 @@ export function colorToNumber(color: Color | number): number {
  * Converts any CSS color value to a standardized hex6 value.
  * Accepts: hex3, hex6, hex8, rgb[1] and rgba[1].
  *
+ * In the case of `light-dark(...)` colors, it will convert both light and dark colors to hex.
+ *
  * [1] under the form rgb(r, g, b, a?) or rgba(r, g, b, a?)
  * with r,g,b ∈ [0, 255] and a ∈ [0, 1]
  *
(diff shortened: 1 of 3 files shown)


src/helpers/color.ts
rrahir, line 66: add a comment that light dark is supported

rrahir (approved):
robodoo r+

PR #7536, opened 2025-11-26
Master add shortcuts matho

matho-odoo:
## Description:

Add some shortcuts

Task: [5231802](https://www.odoo.com/odoo/2328/tasks/5231802)

## review checklist

- [ ] feature is organized in plugin, or UI components
- [ ] support of duplicate sheet (deep copy)
- [ ] in model/core: ranges are Range object, and can be adapted (adaptRanges)
- [ ] in model/UI: ranges are strings (to show the user)
- [ ] undo-able commands (uses this.history.update)
- [ ] multiuser-able commands (has inverse commands and transformations where needed)
- [ ] new/updated/removed commands are documented
- [ ] exportable in excel
- [ ] translations (\_t("qmsdf %s", abc))
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes
- [ ] doc is rebuild (npm run doc)
- [ ] status is correct in Odoo

PR #8593, opened 2026-05-07
[IMP] action menu: replace Ctrl by Cmd for Mac users

fdamhaut:
Task: 6200859

## Description:

Task: [6200859](https://www.odoo.com/odoo/2328/tasks/6200859)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/actions/action.ts
+++ b/src/actions/action.ts
@@ -1,3 +1,4 @@
+import { isMacOS } from "../components/helpers/dom_helpers";
 import { Color } from "../types/misc";
 import { SpreadsheetChildEnv } from "../types/spreadsheet_env";
 
@@ -111,13 +112,17 @@ export function createActions(menuItems: ActionSpec[]): Action[] {
   return menuItems.map(createAction).sort((a, b) => a.sequence - b.sequence);
 }
 
+function adaptShortcutMacOs(shortcut: string) {
+  return shortcut.replace("Ctrl", "⌘").replace("Alt", "⌃");
+}
+
 let nextItemId = 1;
 
 export function createAction(item: ActionSpec): Action {
   const name = item.name;
   const children = item.children;
   const description = item.description;
-  const shortcut = item.shortcut;
+  const shortcut = item.shortcut && isMacOS() ? adaptShortcutMacOs(item.shortcut) : item.shortcut;
   const icon = item.icon;
   const secondaryIcon = item.secondaryIcon;
   const itemId = item.id || nextItemId++;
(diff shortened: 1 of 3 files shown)


src/actions/action.ts
```diff
@@ -111,13 +112,17 @@ export function createActions(menuItems: ActionSpec[]): Action[] {
   return menuItems.map(createAction).sort((a, b) => a.sequence - b.sequence);
 }
 
+function replaceCtrlMacOs(shortcut: string) {
+  return shortcut.replace("Ctrl", "Cmd").replace("Alt", "Ctrl");
+}
```
VincentSchippefilt:
```suggestion
function adaptShortcutMacOs(shortcut: string) {
  return shortcut.replace("Ctrl", "Cmd").replace("Alt", "Ctrl");
}
```

src/actions/action.ts
```diff
@@ -111,13 +112,17 @@ export function createActions(menuItems: ActionSpec[]): Action[] {
   return menuItems.map(createAction).sort((a, b) => a.sequence - b.sequence);
 }
 
+function replaceCtrlMacOs(shortcut: string) {
+  return shortcut.replace("Ctrl", "Cmd").replace("Alt", "Ctrl");
```
rrahir:
```suggestion
  return shortcut.replace("Ctrl", "⌘").replace("Alt", "⌃");
```
I think the key icons are way more useful for a standard user

rrahir: Actually, we should go even further

Ctrl > `⌃`
Alt > actually options so `⌥` (don't think we use use for now)

src/actions/action.ts
```diff
@@ -111,13 +112,17 @@ export function createActions(menuItems: ActionSpec[]): Action[] {
...
+}
+
 let nextItemId = 1;
 
 export function createAction(item: ActionSpec): Action {
   const name = item.name;
   const children = item.children;
   const description = item.description;
-  const shortcut = item.shortcut;
+  const shortcut = item.shortcut && isMacOS() ? replaceCtrlMacOs(item.shortcut) : item.shortcut;
```
rrahir: Some items (like the one to insert links) are not adapted. It seems to come from https://github.com/odoo/o-spreadsheet/pull/7536 where the shortcut key did not exist yet and the shortcut is set as a `descriptiion`

rrahir (changes requested):

pro-odoo:
robodoo r+

PR #8596, opened 2026-05-11
[FIX] xlsx: preserve chart title font size on xlsx export

dhrp-odoo:
## Description:

When exporting a chart title to XLSX, the font size was only set on the
text run properties (`a:rPr`). However, some spreadsheet editors also rely
on the paragraph default run properties (`a:defRPr`) to determine the
selected title font size.

As a result, importing the exported file in Google Sheets could reset the
title size to automatic, and Excel Online could render the title with the
right visual size while still showing the default size in the font size
selector.

Set the font size on both `a:defRPr` and `a:rPr` so the exported title
style is explicit and consistently preserved across spreadsheet editors.

Task: [6170954](https://www.odoo.com/odoo/2328/tasks/6170954)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/xlsx/functions/charts.ts
+++ b/src/xlsx/functions/charts.ts
@@ -170,6 +170,11 @@ function insertText(
   fontsize: number = CHART_TITLE_FONT_SIZE,
   style: { bold?: boolean; italic?: boolean } = {}
 ): XMLString {
+  const textProperties: XMLAttributes = [
+    ["b", style.bold ? "1" : "0"],
+    ["i", style.italic ? "1" : "0"],
+    ["sz", fontsize * 100],
+  ];
   return escapeXml/*xml*/ `
     <c:tx>
       <c:rich>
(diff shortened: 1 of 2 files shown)


src/helpers/figures/charts/runtime/chartjs_dataset.ts
```diff
@@ -620,7 +621,7 @@ function getTextStyle(design: TitleDesign | undefined, defaultDesign: TitleDesig
     font: {
       weight: design?.bold ?? defaultDesign?.bold ? "bold" : "normal",
       style: design?.italic ?? defaultDesign?.italic ? "italic" : "normal",
-      size: design?.fontSize ?? defaultDesign?.fontSize,
+      size: fontSizeInPixels(design?.fontSize ?? defaultDesign.fontSize!),
```
LucasLefevre: better to avoid `!`
Sooner or later, it'll lead to a crash

src/helpers/figures/charts/scorecard_chart_config_builder.ts
```diff
@@ -369,7 +369,7 @@ class ScorecardChartConfigBuilder {
         this.baselineArrow === "neutral" || this.runtime.progressBar
           ? undefined
           : {
-              size: this.keyValue ? 0.8 * baselineValueFontSize : 0,
+              size: this.keyValue ? 0.8 * fontSizeInPixels(baselineValueFontSize) : 0,
```
LucasLefevre: is it intentional to have  `0.8 * px` now instead of `0.8 * pt` before ?

LucasLefevre (commented):
The first commit should target master IMO to avoid changing sizes on a stable release (users already have set sizes for the look they want, no matter the number. They don't care if 24 != 24 as long as it visually matches what they want). Let's avoid breaking that.
And it's also 1) not important at all -> master 2) easier because no fw ports

src/helpers/figures/charts/scorecard_chart_config_builder.ts
dhrp-odoo: Yes, intentional.

`baselineArrow.size` is a canvas pixel size, and it is derived from the baseline text size. Since the baseline text now renders from pt to px, the arrow should use the converted px size too, otherwise it becomes smaller relative to the text.

LucasLefevre (approved):
robodoo r+

PR #8637, opened 2026-05-11
Master cf dv side panel bis flda matho

fdamhaut:
## Description:

Task: [4072640](https://www.odoo.com/odoo/2328/tasks/4072640)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/filters/filter_menu_criterion/filter_menu_criterion.xml
+++ b/src/components/filters/filter_menu_criterion/filter_menu_criterion.xml
@@ -15,6 +15,7 @@
       onCriterionChanged.bind="this.onCriterionChanged"
       disableFormulas="true"
       autofocus="true"
+      sheetId="this.env.model.getters.getActiveSheetId()"
     />
   </t>
 </templates>
--- a/src/components/helpers/formula.ts
+++ b/src/components/helpers/formula.ts
@@ -1,0 +1,23 @@
+import { Getters, UID } from "../..";
+import { rangeTokenize } from "../../formulas/range_tokenizer";
+
+/*
+ * Adapt string `formula` representation on `fromSheetId`
+ * to ensure they are correctly represented on `toSheetId`.
+ */
+export function adaptFormulaToSheet(
+  getters: Getters,
+  formula: string,
+  toSheetId: UID,
+  fromSheetId: UID
+) {
+  return rangeTokenize(formula)
+    .map((token) => {
+      if (token.type === "REFERENCE") {
+        const range = getters.getRangeFromSheetXC(fromSheetId, token.value);
+        return getters.getRangeString(range, toSheetId);
+      }
+      return token.value;
+    })
+    .join("");
+}
--- a/src/components/side_panel/conditional_formatting/cf_editor/cf_editor.ts
+++ b/src/components/side_panel/conditional_formatting/cf_editor/cf_editor.ts
@@ -33,28 +33,25 @@ export class ConditionalFormattingEditor extends Component<SpreadsheetChildEnv> 
     cf: types.ConditionalFormat(),
     isNewCf: types.boolean(),
     onCloseSidePanel: types.function(),
+    sheetId: types.UID(),
   });
 
-  private activeSheetId!: UID;
   private store!: Store<ConditionalFormattingEditorStore>;
 
   setup() {
-    this.activeSheetId = this.env.model.getters.getActiveSheetId();
     this.store = useLocalStore(
       ConditionalFormattingEditorStore,
       deepCopy(this.props.cf),
-      this.props.isNewCf
+      this.props.isNewCf,
+      this.props.sheetId
     );
     useLayoutEffect(
-      (sheetId, isCfRemoved) => {
-        if (this.activeSheetId !== sheetId || isCfRemoved) {
-          this.env.replaceSidePanel(
-            "ConditionalFormatting",
-            `ConditionalFormattingEditor_${this.props.cf.id}`
-          );
+      (isCfRemoved) => {
+        if (isCfRemoved) {
+          this.closeEditor();
         }
       },
-      () => [this.env.model.getters.getActiveSheetId(), this.isEditedCfRemoved]
+      () => [this.isEditedCfRemoved]
     );
     useListener(window as any, "click", () => this.store.closeMenus());
   }
@@ -62,7 +59,7 @@ export class ConditionalFormattingEditor extends Component<SpreadsheetChildEnv> 
   get isEditedCfRemoved() {
     return !Boolean(
       this.env.model.getters
-        .getConditionalFormats(this.activeSheetId)
+        .getConditionalFormats(this.props.sheetId)
         .find((cf) => cf.id === this.props.cf.id)
     );
   }
@@ -80,10 +77,7 @@ export class ConditionalFormattingEditor extends Component<SpreadsheetChildEnv> 
     this.store.updateConditionalFormat({});
     const isSuccessful = this.store.state.errors.length === 0;
     if (isSuccessful) {
-      this.env.replaceSidePanel(
-        "ConditionalFormatting",
-        `ConditionalFormattingEditor_${this.props.cf.id}`
-      );
+      this.closeEditor();
     }
   }
 
@@ -91,19 +85,27 @@ export class ConditionalFormattingEditor extends Component<SpreadsheetChildEnv> 
     if (this.store.state.hasEditedCf) {
       if (this.props.isNewCf) {
         this.env.model.dispatch("REMOVE_CONDITIONAL_FORMAT", {
-          sheetId: this.activeSheetId,
+          sheetId: this.props.sheetId,
           id: this.props.cf.id,
         });
       } else {
         this.env.model.dispatch("ADD_CONDITIONAL_FORMAT", {
           cf: this.props.cf,
           ranges: this.props.cf.ranges.map((range) =>
-            this.env.model.getters.getRangeDataFromXc(this.activeSheetId, range)
+            this.env.model.getters.getRangeDataFromXc(this.props.sheetId, range)
           ),
-          sheetId: this.activeSheetId,
+          sheetId: this.props.sheetId,
         });
       }
     }
+    this.closeEditor();
+  }
+
+  closeEditor() {
+    this.env.model.dispatch("ACTIVATE_SHEET", {
+      sheetIdTo: this.props.sheetId,
+      sheetIdFrom: this.env.model.getters.getActiveSheetId(),
+    });
     this.env.replaceSidePanel(
       "ConditionalFormatting",
       `ConditionalFormattingEditor_${this.props.cf.id}`
--- a/src/components/side_panel/conditional_formatting/cf_editor/cf_editor.ts
+++ b/src/components/side_panel/conditional_formatting/cf_editor/cf_editor.ts
@@ -1,9 +1,9 @@
 import { useListener, useProps } from "@odoo/owl";
 import { deepCopy } from "../../../../helpers/misc";
+import { splitReference } from "../../../../helpers/references";
 import { Component, useLayoutEffect } from "../../../../owl3_compatibility_layer";
 import { useLocalStore } from "../../../../store_engine/store_hooks";
 import { _t } from "../../../../translation";
-import { UID } from "../../../../types/misc";
 import { SpreadsheetChildEnv } from "../../../../types/spreadsheet_env";
 import { Store } from "../../../../types/store_engine";
 import { types } from "../../../props_validation";
@@ -51,17 +51,9 @@ export class ConditionalFormattingEditor extends Component<SpreadsheetChildEnv> 
           this.closeEditor();
         }
       },
-      () => [this.isEditedCfRemoved]
+      () => [this.store.isEditedCfRemoved]
     );
     useListener(window as any, "click", () => this.store.closeMenus());
-  }
-
-  get isEditedCfRemoved() {
-    return !Boolean(
-      this.env.model.getters
-        .getConditionalFormats(this.props.sheetId)
-        .find((cf) => cf.id === this.props.cf.id)
-    );
   }
 
   get cfTypesValues() {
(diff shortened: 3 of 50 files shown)


src/plugins/core/data_validation.ts
```diff
@@ -367,9 +367,8 @@ export class DataValidationPlugin
 
   private checkValidRange(cmd: AddDataValidationCommand): CommandResult {
     const ranges = cmd.ranges.map((range) => this.getters.getRangeFromRangeData(range));
-    const stringRanges = ranges.map((range) => this.getters.getRangeString(range, cmd.sheetId));
-    if (stringRanges.some((xc) => !this.getters.isRangeValid(xc))) {
-      return CommandResult.InvalidRange;
+    if (ranges.some((range) => range.sheetId !== cmd.sheetId)) {
```
anhe-odoo, line 380: Why don't we have to check here anymore if the range is still valid ?

src/components/side_panel/conditional_formatting/cf_editor/cf_editor.ts
```diff
@@ -77,34 +74,49 @@ export class ConditionalFormattingEditor extends Component<Props, SpreadsheetChi
...
           ),
-          sheetId: this.activeSheetId,
+          sheetId: this.props.sheetId,
         });
       }
     }
+    this.closeSidePanel();
+  }
+
+  closeSidePanel() {
```
anhe-odoo: I'm not a fan of the name. From my understanding, it will change the switch to the "detailed" side panel of a CF, not really close a side panel. Or maybe the name should suggest that if closes and open another ?

src/components/side_panel/data_validation/dv_editor/dv_editor.ts
```diff
@@ -1,14 +1,18 @@
 import { Component, ComponentConstructor, useState } from "@odoo/owl";
```
anhe-odoo: Typo in the commit : The use is allowed to witch sheet (missing the "S" of SWITCH)

src/components/side_panel/data_validation/dv_editor/dv_editor.ts
```diff
@@ -115,8 +125,29 @@ export class DataValidationEditor extends Component<Props, SpreadsheetChildEnv>
     return _t("Apply to ranges:");
   }
 
-  get dispatchPayload(): Omit<AddDataValidationCommand, "type"> {
+  get dispatchPayload(): Omit<AddDataValidationCommand, "type">[] | undefined {
     const rule = { ...this.state.rule, ranges: undefined };
+    let errors = [] as CancelledReason[];
```
anhe-odoo: `const errors: CancelledReason[] = [];`

src/components/side_panel/data_validation/dv_editor/dv_editor.ts
```diff
@@ -115,8 +125,29 @@ export class DataValidationEditor extends Component<Props, SpreadsheetChildEnv>
...
+    const ranges = this.state.rule.ranges.map((xc) => {
+      const { sheetName } = splitReference(xc);
+      const sheetId = this.env.model.getters.getSheetIdByName(sheetName);
+      if (sheetName && !sheetId) {
+        errors = [CommandResult.InvalidRange];
+      }
+      return this.env.model.getters.getRangeDataFromXc(this.props.sheetId, xc);
+    });
+    if (errors.length) {
```
anhe-odoo: There is no need to split the case here as if errors.length is falsy, then errors == [], so you can do directly `this.state.errors = errors`

src/components/side_panel/data_validation/dv_editor/dv_editor.ts
```diff
@@ -115,8 +125,29 @@ export class DataValidationEditor extends Component<Props, SpreadsheetChildEnv>
...
 
-  get dispatchPayload(): Omit<AddDataValidationCommand, "type"> {
+  get dispatchPayload(): Omit<AddDataValidationCommand, "type">[] | undefined {
     const rule = { ...this.state.rule, ranges: undefined };
+    let errors = [] as CancelledReason[];
+    const ranges = this.state.rule.ranges.map((xc) => {
+      const { sheetName } = splitReference(xc);
+      const sheetId = this.env.model.getters.getSheetIdByName(sheetName);
+      if (sheetName && !sheetId) {
+        errors = [CommandResult.InvalidRange];
```
anhe-odoo: Is it wanted that if you have a sheetName and no sheetId, you will still call the `this.env.model.getters.getRangeDataFromXc` ? Shouldn't we return `undefined` instead, as we already know that we will have an InvalidRange ?

src/components/side_panel/conditional_formatting/cf_editor/cf_editor.ts
fdamhaut: > From my understanding, it will change the switch to the "detailed" side panel of a CF, not really close a side panel.

It's the opposite, we are leaving the detailed view to go back to the "list" view. 
But as it's unclear I will switch to smth more explicit :D

src/plugins/core/data_validation.ts
fdamhaut: Fixed 😄

src/components/helpers/formula.ts
```diff
@@ -0,0 +1,22 @@
+import { Getters, UID } from "../..";
+import { rangeTokenize } from "../../formulas/range_tokenizer";
+
+export function adaptFormulaToSheet(
+  getters: Getters,
+  formula: string,
+  toSheetId: UID,
+  fromSheetId?: UID
```
hokolomopo: shouldn't be optional if the whole point of the helper is to adapt the sheet. And I think you never not give the argument

src/components/side_panel/conditional_formatting/cf_editor/cf_editor.xml
```diff
@@ -10,6 +10,7 @@
             onSelectionChanged="(ranges) => this.store.onRangeUpdate(ranges)"
             onSelectionConfirmed="() => this.store.onRangeConfirmed()"
             required="true"
+            prefixSheet="true"
```
hokolomopo: Do we really want to prefix the sheet name if we are on the same sheet as the CF ? Seem very strange that the composer is "smart" and only prefix the sheet if we are on another sheet, but the selection input always prefix the sheet

src/components/side_panel/conditional_formatting/cf_editor/cf_editor.ts
```diff
@@ -85,30 +82,35 @@ export class ConditionalFormattingEditor extends Component<Props, SpreadsheetChi
...
       }
     }
+    this.closeEditor();
+  }
+
+  closeEditor() {
+    this.env.model.dispatch("ACTIVATE_SHEET", {
+      sheetIdTo: this.props.sheetId,
+      sheetIdFrom: this.env.model.getters.getActiveSheetId(),
+    });
```
hokolomopo: not sure it's needed, if I created CFs on a different sheet than the one where I opened the panel, I don't want to go back to the sheet where I opened the panel

src/components/side_panel/conditional_formatting/cf_editor/cf_editor_store.ts
```diff
@@ -94,31 +96,44 @@ export class ConditionalFormattingEditorStore extends SpreadsheetStore {
   }
 
   updateConditionalFormat(newCf: Partial<ConditionalFormat> & { suppressErrors?: boolean }) {
```
hokolomopo: Seems a big bugged: if I open a CF on a sheet, change the range to another sheet then confirm, I get 2 cfs, instead of only one on the other sheet

src/components/side_panel/conditional_formatting/cf_preview_list/cf_preview_list.ts
```diff
@@ -80,13 +81,16 @@ export class ConditionalFormatPreviewList extends Component<Props, SpreadsheetCh
...
     return this.env.replaceSidePanel("ConditionalFormattingEditor", "ConditionalFormatting", {
       cf: {
         ...cf,
-        ranges: zones.map((zone) =>
-          zoneToXc(this.env.model.getters.getUnboundedZone(sheetId, zone))
-        ),
+        ranges,
       },
+      sheetId,
       isNewCf: true,
```
hokolomopo: are these change useful? The preview are only for the CFs of the current sheet, no ?

src/components/side_panel/data_validation/dv_editor/dv_editor.ts
```diff
@@ -87,20 +90,49 @@ export class DataValidationEditor extends Component<Props, SpreadsheetChildEnv>
 
   onCancel() {
     this.props.onCancel?.();
+    this.env.model.dispatch("ACTIVATE_SHEET", {
+      sheetIdTo: this.props.sheetId,
+      sheetIdFrom: this.env.model.getters.getActiveSheetId(),
+    });
```
hokolomopo: same comment as Cfs

src/components/side_panel/data_validation/dv_editor/dv_editor.ts
```diff
@@ -87,20 +90,49 @@ export class DataValidationEditor extends Component<Props, SpreadsheetChildEnv>
...
+      const result = this.env.model.dispatch("ADD_DATA_VALIDATION_RULE", payload);
+      if (!result.isSuccessful) {
+        this.state.errors = result.reasons;
+        return;
+      }
+    }
+    this.env.model.dispatch("ACTIVATE_SHEET", {
+      sheetIdTo: this.props.sheetId,
+      sheetIdFrom: this.env.model.getters.getActiveSheetId(),
+    });
```
hokolomopo: here also, not sure we want to activate the sheet

src/components/side_panel/data_validation/dv_editor/dv_editor.ts
```diff
@@ -87,20 +90,49 @@ export class DataValidationEditor extends Component<Props, SpreadsheetChildEnv>
...
+    if (!dispatchPayload) {
       return;
     }
+    for (const payload of dispatchPayload) {
+      const result = this.env.model.dispatch("ADD_DATA_VALIDATION_RULE", payload);
+      if (!result.isSuccessful) {
+        this.state.errors = result.reasons;
+        return;
+      }
+    }
```
hokolomopo: probably should call `canDispatch` beforehand. you could have half of the call sucessfull, the other not, leading to a strange state.

hokolomopo: Also this will create multiple history steps :/ (probably same issues for CFs)

hokolomopo (commented):
👋

src/components/side_panel/conditional_formatting/cf_editor/cf_editor.xml
matho-odoo: It is because you can now switch sheet while editing a CF, so the user need that information to be visible.

hokolomopo: Yeah but in the composer, the sheetName is only displayed if we are one a different sheet than the one we opened the panel in, but the selection input always displays the sheetName

src/components/helpers/formula.ts
anhe-odoo: The argument is not given in criterion_input.ts. But I agree with ADRM's comment, this shouldn't be optional.

src/components/side_panel/conditional_formatting/cf_editor/cf_editor.xml
fdamhaut: I am not sure I understand the point, If you switch sheet the name of the range will not change, so if you are on sheet3 there is no way to know whether the A1 refer to sheet1!A1 or sheet2!A1 (Other than remembering where you open the side panel)

src/components/side_panel/conditional_formatting/cf_preview_list/cf_preview_list.ts
fdamhaut: it's for https://github.com/odoo/o-spreadsheet/pull/8637#discussion_r3665333222

src/components/side_panel/conditional_formatting/cf_editor/cf_editor.xml
hokolomopo: My point is that when you _don't_ switch sheet, adding a range in the selection input will prefix it w/ the sheet name, and adding a range in the standalone composer will not prefix it. That's a bit strange. Both should have the same level of smartness when deciding when to prefix the range or not.

src/components/side_panel/criterion_form/value_in_list_criterion/value_in_list_criterion.xml
```diff
@@ -9,6 +9,7 @@
           />
         </div>
         <CriterionInput
+          sheetId="props.sheetId"
```
anhe-odoo: Shouldn't it be `this.props.sheetId` now with the new version of owl ? If I'm not wrong, this will lead to an undefined sheetId, meaning that this props is unused ?

src/plugins/core/data_validation.ts
```diff
@@ -367,6 +367,9 @@ export class DataValidationPlugin
 
   private checkValidRange(cmd: AddDataValidationCommand): CommandResult {
     const ranges = cmd.ranges.map((range) => this.getters.getRangeFromRangeData(range));
+    if (ranges.some((range) => range.sheetId !== cmd.sheetId)) {
+      return CommandResult.TargetOutOfSheet;
```
anhe-odoo: Why do you return an outOfSheet if the sheetId is different ? Shouldn't we use another error then ?

src/components/side_panel/data_validation/dv_editor/dv_editor.ts
```diff
@@ -76,30 +78,62 @@ export class DataValidationEditor extends Component<SpreadsheetChildEnv> {
...
   }
 
   onSave() {
-    const result = this.env.model.dispatch("ADD_DATA_VALIDATION_RULE", this.dispatchPayload);
+    const dispatchPayload = this.dispatchPayload;
+    if (!dispatchPayload) {
+      return;
+    }
+    const result = this.env.model.dispatch("ADD_DATA_VALIDATION_RULES", {
+      sheetIdsToAdd: dispatchPayload,
```
anhe-odoo: The `dispatchPayload` thing shouldn't be in another object, as the payload dispatch is supposed to the the arguments that are dispatched in the command. Moreover, the name "dispatchPayload" used as the value of the `sheetIdsToAdd` field is feels weird ...

anhe-odoo: And the name of the property `sheetIdsToAdd` give the feeling that we will get an array of sheet Ids, while here we give a dictionary of rules by sheetId

src/plugins/ui_feature/datavalidation_insertion.ts
```diff
@@ -1,11 +1,36 @@
...
+              sheetId,
+              rule: command.sheetIdsToAdd[sheetId].rule,
+              ranges: command.sheetIdsToAdd[sheetId].ranges,
+            })
+          );
+        }
+        const reasons = results
+          .map((result) => result.reasons)
+          .flat()
+          .filter((r) => r !== CommandResult.NoChanges);
```
anhe-odoo, line 25: Shouldn't we also filter `CommandResult.Succes` ?

src/types/commands.ts
```diff
@@ -1204,6 +1204,11 @@ export interface UpdateColorSchemeCommand {
   colorScheme: "light" | "dark";
 }
 
+export interface AddDataValidationsCommand {
+  type: "ADD_DATA_VALIDATION_RULES";
+  sheetIdsToAdd: { [key: UID]: { rule: Omit<DataValidationRule, "ranges">; ranges: RangeData[] } };
```
anhe-odoo: Maybe something like `rulesBySheetId` ? What do you think ?

src/plugins/ui_feature/datavalidation_insertion.ts
matho-odoo: result.reasons are "CancelledReason" type wich does not include CommandResult.Success

src/helpers/range.ts
```diff
@@ -446,19 +446,19 @@ function getApplyRangeChangeDeleteSheet(cmd: DeleteSheetCommand): ApplyRangeChan
 function getApplyRangeChangeRenameSheet(cmd: RenameSheetCommand): ApplyRangeChange {
   return (range: Range) => {
     if (range.sheetId === cmd.sheetId) {
-      return { changeType: "CHANGE", range };
```
pro-odoo, line 449: There are occurences of "CHANGE" in odoo that are not replaced. (It probably means that some tests are missing in odoo)

tests/pivots/spreadsheet_pivot/spreadsheet_pivot_side_panel.test.ts
```diff
@@ -237,12 +237,6 @@ describe("Spreadsheet pivot side panel", () => {
       "=A1+Sheet2!A1"
     );
     await keyDown({ key: "Enter" });
-
-    activateSheet(model, sheet2Id);
-    // close the side panel and reopen it while the second sheet is active
-    await click(fixture.querySelector(".o-sidePanelClose")!);
-    env.openSidePanel("PivotSidePanel", { pivotId: "3" });
-    await nextTick();
```
pro-odoo, line 255: I do not understand this change

tests/pivots/spreadsheet_pivot/spreadsheet_pivot_side_panel.test.ts
```diff
@@ -253,7 +247,7 @@ describe("Spreadsheet pivot side panel", () => {
     env.openSidePanel("PivotSidePanel", { pivotId: "3" });
     await nextTick();
     expect(fixture.querySelector(".pivot-dimension .o-composer")?.textContent).toEqual(
-      "=A1+Sheet2!A1"
+      "=Sheet1!A1+Sheet2!A1"
```
pro-odoo: Why ?

src/components/helpers/formula.ts
```diff
@@ -0,0 +1,19 @@
+import { Getters, UID } from "../..";
+import { rangeTokenize } from "../../formulas/range_tokenizer";
+
+export function adaptFormulaToSheet(
```
pro-odoo, line 8: Would require a little docstring, especially to explain what toSheetId and fromSheetId means.
Also, you can add tests to test this helper specifically

src/components/filters/filter_menu_criterion/filter_menu_criterion.xml
```diff
@@ -15,6 +15,7 @@
       onCriterionChanged.bind="this.onCriterionChanged"
```
pro-odoo, line 15:
```
Make the criterion input sheet dependent, this mean that when saved,
their content will always show the value based on their sheetId
 ```
The commit message misses the "why we do that". It's not that clear

**Review by pro-odoo — COMMENTED** [review]
Some comments while scrolling, commit messages would need some love, they are not that clear :/ 
Also, it needs an adaptation in Odoo

File: src/helpers/range.ts (hunk 20 above)
Comment (anhe-odoo) [reply src/helpers/range.ts:449]: @fdamhaut Can this comment be resolved now ? :)

File: src/plugins/core/data_validation.ts [hunk 25]
```diff
@@ -229,7 +229,17 @@ export class DataValidationPlugin
       adaptedRules[ruleIndex] = newRule;
       this.history.update("rules", sheetId, adaptedRules);
     } else {
-      this.history.update("rules", sheetId, [...adaptedRules, newRule]);
+      const similarRuleIndex = adaptedRules.findIndex(
```
anhe-odoo, line 232: Is there any scenario where we can end with more than one matching rule here ?

PR #8646, opened 2026-05-12
[IMP] sheet: add background color to sheet

hokolomopo:
## Description

This commit adds the possibility to set the background color of a sheet.

There was also a few issue with the sheet tab color, the color was not added to the custom colors and the color picker didn't have a color selected.

Task: [6209932](https://www.odoo.com/odoo/2328/tasks/6209932)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/stores/grid_renderer_store.ts
```diff
@@ -209,13 +215,15 @@ export class GridRenderer extends DisposableStore {
     const inset = areGridLinesVisible ? 0.1 * thinLineWidth : 0;
 
     if (areGridLinesVisible) {
-      const theme = this.getters.getSpreadsheetTheme();
-      ctx.strokeStyle = theme.gridBorderColor;
+      const baseGridColor = "#CECFCF";
+      const background = this.getBackgroundColor(renderingContext);
+      const blendedGridColor = blendColors(baseGridColor + "aa", background + "50");
+      ctx.strokeStyle = setColorAlpha(blendedGridColor, 1);
+      ctx.lineWidth = thinLineWidth;
```
hokolomopo: Bend the color a bit with the background so it's less shocking of non-white background color. Color was picked so it stays the same on a white background.

Still not great on colored background tho. Probably should tweak it a bit, and have a different color on high-contrast background. 

But for obvious reasons, I'm not great with colors 😔

anhe-odoo:
@hokolomopo : need a rebase (probably because of the dark mode, as the grid doesn't seem to work for the dark mode ...)

hokolomopo:
I wanted to use the new `DefaultPlugin` to set the sheet color, but those are cell styles that override table colors, which we probably don't want 😢

rrahir:
robodoo r+

PR #8694, opened 2026-05-13
[IMP] compiler: force validated strings

LucasLefevre:
The formula compiler builds JS source by string concatenation and feeds it to `new Function(...)`. Several of those strings came from user input (function names, operator symbols), so any value that slipped past the parser's validation could end up executed as code.

This commit introduces a `JsString` branded type plus a `jsStr` tagged-template helper:
generated code can only be assembled from values explicitly marked trusted, and untrusted strings must go through `dangerouslyCreateJsStr`, which makes the trust decision auditable.

Task: 6185314

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/formulas/code_builder.ts
+++ b/src/formulas/code_builder.ts
@@ -2,40 +2,90 @@
  * Block of code that produces a value.
  */
 export interface FunctionCode {
-  readonly returnExpression: string;
+  readonly returnExpression: JsString;
   /**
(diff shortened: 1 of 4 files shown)


rrahir:
robodoo r+

PR #8747, opened 2026-05-19
Master fix core order rar

rrahir:
## Description:

Task: [4908195](https://www.odoo.com/odoo/2328/tasks/4908195)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/helpers/pivot/pivot_registry.ts
```diff
@@ -38,7 +38,7 @@ export interface PivotRegistryItem {
   canHaveCustomGroup: (field: PivotField) => boolean;
   isPivotUnused: (getters: Getters, pivotId: UID) => boolean;
   adaptRanges?: (
-    getters: CoreGetters,
+    getters: Pick<CoreGetters, "getRangeFromZone">,
```
rrahir: should have the getter typing of pivot plugin

src/plugins/core/cell.ts
```diff
@@ -60,6 +60,7 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
     "getCellById",
```
hokolomopo, line 69: If all the commits area meant to be kept, they should maybe have a small commit msg + task id

src/plugins/core/cell.ts
```diff
@@ -64,6 +66,7 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
   ] as const;
   readonly nextId = 1;
   public readonly cells: { [sheetId: string]: { [id: string]: Cell } } = {};
+  previousLocale: Locale = this.getters.getLocale();
```
hokolomopo, line 78: shouldn't it be in a beforeHandle instead of always storing this ?

src/plugins/plugin_registries.ts
```diff
@@ -69,6 +70,7 @@ export const corePluginRegistry = new Registry<CorePluginConstructor>()
   .add("settings", SettingsPlugin)
   .add("header grouping", HeaderGroupingPlugin)
   .add("header visibility", HeaderVisibilityPlugin)
+  .add("headerSvisibility global", HeaderGlobalVisibilityPlugin)
```
hokolomopo: Snake case has the wrong letter uppercased.  `HeaderVisibilityPlugin` don't use snake case, maybe be consistent

src/plugins/core/pivot.ts
```diff
@@ -478,3 +484,5 @@ export class PivotCorePlugin extends CorePlugin<CoreState> implements CoreState
     data.pivotNextId = this.nextFormulaId;
   }
 }
+
+export type PivotCoreGetters = DepsGetters<typeof PivotCorePlugin>;
```
hokolomopo: types should probably exported from types/getters.ts

src/plugins/core/squisher.ts
```diff
@@ -46,7 +49,12 @@ export class Squisher {
 
   private baseNumber: number | undefined = undefined;
 
-  constructor(getters: CoreGetters) {
+  constructor(
+    getters: Pick<
+      CoreGetters,
+      "getCell" | "getSheetName" | "getRangeString" | "getLocale" | "getRangeFromSheetXC"
+    >
```
hokolomopo: could all these `Pick<someGetters>` be transformed to `CellGetters` & such ?

src/model.ts
```diff
@@ -706,3 +709,25 @@ function createCommand(type: string, payload: any = {}): Command {
   command.type = type;
   return command;
 }
+
+export function sortByDependencies(plugins: CorePluginConstructor[]): CorePluginConstructor[] {
```
hokolomopo: i'd either move the function to some helper file, or make it a private method of model

hokolomopo: (oh it's deleted in another commit, nvm then)

tests/model/model.test.ts
```diff
@@ -423,74 +422,74 @@ describe("Model", () => {
...
-        }
-      }
-    }
-  });
-});
+// describe("sortByDependencies", () => {
+//   function makePlugin(name: string, deps: CorePluginConstructor[] = []): CorePluginConstructor {
+//     return class {
+//       static getters = [] as const;
+//       static readonly dependencies = deps;
```
hokolomopo: a commit introducing the tests, another commenting them, and yet another uncommenting them doesn't seems great 🙃

src/model.ts
```diff
@@ -83,6 +83,8 @@ const enum Status {
  */
 export class Model extends EventBus<any> implements CommandDispatcher {
   private corePlugins: CorePlugin[] = [];
+  private corePluginsByConstructor = new Map<CorePluginConstructor, CorePlugin>();
```
hokolomopo: not sure how relevant the commit "[IMP] setupCorePlugin: restrict getters to transitive dependency closure" is, almost every single line of it ended up changing in the final version

src/plugins/core_plugin.ts
```diff
@@ -90,14 +90,14 @@ export class CorePlugin<State = any, Self extends AnyPluginConstructor = any>
 {
   static readonly dependencies: readonly AnyPluginConstructor[] = [];
 
-  protected getters: DepsGetters<Self>;
+  protected getters: DepsGetters<Self> & Partial<CoreGetters>;
```
hokolomopo: You sure we want to do this ? I feel like it just helps the dev make mistakes

src/plugins/plugin_registries.ts
```diff
@@ -65,16 +65,14 @@ import { HeaderPositionsUIPlugin } from "./ui_stateful/header_positions";
 import { GridSelectionPlugin } from "./ui_stateful/selection";
 import { SheetViewPlugin } from "./ui_stateful/sheetview";
 
-class CorePluginRegistry extends Registry<CorePluginConstructor> {
+export class CorePluginRegistry extends Registry<CorePluginConstructor> {
+  // private _entries: [string, CorePluginConstructor][] = [];
```
hokolomopo: comment

src/plugins/plugin_registries.ts
```diff
@@ -89,31 +87,24 @@ class CorePluginRegistry extends Registry<CorePluginConstructor> {
...
-      visiting.add(p);
-      for (const dep of p.dependencies as CorePluginConstructor[]) {
-        if (all.has(dep)) {
-          dfs(dep, [...path, p]);
-        }
-      }
-      visiting.delete(p);
-      visited.add(p);
-    };
+  // override getAll(): CorePluginConstructor[] {
```
hokolomopo: comment

tests/model/__snapshots__/model.test.ts.snap
```diff
@@ -0,0 +1,26 @@
+// Jest Snapshot v1, https://jestjs.io/docs/snapshot-testing
+
+exports[`sortByDependencies Core plugin order is unchanged 1`] = `
+[
+  "SheetPlugin",
+  "SettingsPlugin",
+  "HeaderGroupingPlugin",
+  "HeaderVisibilityPlugin",
+  "HeaderGlobalVisibilityPlugin",
```
hokolomopo: Idea: another snapshot that lists all of the plugin and for each plugin show their dependency tree. Might be useful to see if what we did was correct, and how to integrate a new feature somewhere.

hokolomopo (commented):
👋 

Seems very stylé 😄 

Having a bagillion commits makes it so some of them are not relevant by the time we get to the end tho 🙃

src/plugins/core/cell.ts
LucasLefevre, line 78: should be in `beforeHandle`?

src/plugins/core/data_validation.ts
```diff
@@ -33,6 +35,7 @@ export class DataValidationPlugin
   ] as const;
 
   readonly rules: { [sheet: string]: DataValidationRule[] } = {};
+  coordinateWithContent = new Set<UID>();
```
LucasLefevre: 🧦 
```suggestion
  coordinatesWithContent = new Set<UID>();
```

src/plugins/core/data_validation.ts
```diff
@@ -120,6 +123,25 @@ export class DataValidationPlugin
...
+    switch (cmd.type) {
+      case "DELETE_CONTENT":
+        const zones = recomputeZones(cmd.target);
+        const sheetId = cmd.sheetId;
+        for (const zone of zones) {
+          for (let row = zone.top; row <= zone.bottom; row++) {
+            for (let col = zone.left; col <= zone.right; col++) {
+              const cell = this.getters.getCell({ sheetId, col, row });
+              if (!cell?.isFormula && cell?.content) {
+                this.coordinateWithContent.add(`${col}-${row}`);
```
LucasLefevre: actually, it excludes formulas
```suggestion
              if (!cell?.isFormula && cell?.content) {
                this.coordinatesWithLiteralValue.add(`${col}-${row}`);
```

src/plugins/core/data_validation.ts
```diff
@@ -165,7 +185,8 @@ export class DataValidationPlugin
                 dataValidation.criterion.type === "isBoolean" ||
                 (dataValidation.criterion.type === "isValueInList" &&
                   !cell?.isFormula &&
-                  !cell?.content)
+                  !cell?.content &&
+                  !this.coordinateWithContent.has(`${col}-${row}`))
```
LucasLefevre: does `!cell?.isFormula && !cell?.content` still makes sense here then ?

and it doesn't match the new condition in `beforeHandle`, so not sure. I think the condition above should include formulas

src/plugins/core/tables.ts
```diff
@@ -107,6 +108,16 @@ export class TablePlugin extends CorePlugin<TableState> implements TableState {
     return CommandResult.Success;
   }
 
+  beforeHandle(cmd: CoreCommand): void {
+    if (cmd.type === "UPDATE_CELL") {
+      for (const table of this.getCoreTables(cmd.sheetId)) {
+        if (table.type === "static") {
```
LucasLefevre: `table.type` can also be `"forceStatic"`. I guess those should also be taken into account here

src/plugins/core/tables.ts
```diff
@@ -166,7 +177,7 @@ export class TablePlugin extends CorePlugin<TableState> implements TableState {
           if (table.type === "dynamic") {
             continue;
           }
```
LucasLefevre: can this `if` be removed and rely on `this.extendDirection[table.id];` instead ?
'cause right now, the two `if`s needs to be synced (and they aren't)

src/plugins/core/header_visibility.ts
```diff
@@ -1,25 +1,21 @@
...
 
 export class HeaderVisibilityPlugin extends CorePlugin {
   static getters = [
-    "checkElementsIncludeAllVisibleHeaders",
     "getHiddenColsGroups",
     "getHiddenRowsGroups",
-    "isHeaderHiddenByUser",
-    "isRowHiddenByUser",
-    "isColHiddenByUser",
+    "isRowExplicitlyHidden",
```
LucasLefevre: by looking at the names, I can't know the difference between `isRowHiddenByUser` and `isRowExplicitlyHidden`

maybe 
- `isRowHiddenByUser` -> `isRowHiddenOrFoldedByUser` ?
- `isRowExplicitlyHidden` -> `isRowHiddenByUser` ?

🤷

src/plugins/core/header_visibility.ts
LucasLefevre, line 1: maybe rename this plugin/file to "hidden_headers" or "hide_headers" or "hide something", and then
let `header_visibility_global` be `header_visibility` ?

src/plugins/plugin_registries.ts
LucasLefevre: typo in `headerSvisibility`

src/plugins/core/header_visibility.ts
```diff
@@ -1,25 +1,23 @@
...
-  range,
 } from "../../helpers/misc";
 import { Command, CommandResult } from "../../types/commands";
 import { ConsecutiveIndexes, Dimension, HeaderIndex, UID } from "../../types/misc";
-import { ExcelWorkbookData, WorkbookData } from "../../types/workbook_data";
 import { CorePlugin } from "../core_plugin";
+import { SheetPlugin } from "./sheet";
 
-export class HeaderVisibilityPlugin extends CorePlugin {
+export class HeaderVisibilityPlugin extends CorePlugin<any, typeof HeaderVisibilityPlugin> {
```
LucasLefevre: `any` 😢 
needs to be fixed later

src/plugins/core_plugin.ts
```diff
@@ -28,18 +75,20 @@ export interface CorePluginConstructor {
  * persisted state.
  * They should not be concerned about UI parts or transient state.
  */
-export class CorePlugin<State = any>
+export class CorePlugin<State = any, Self extends CorePluginConstructor = any>
```
LucasLefevre:
```suggestion
export class CorePlugin<State = any, Self extends CorePluginConstructor>
```
(and move it at the first position otherwise the compiler complains)

src/plugins/core/header_visibility_global.ts
```diff
@@ -1,10 +1,15 @@
...
 import { Command, CommandResult } from "../../types/commands";
 import { Dimension, HeaderGroup, HeaderIndex, UID } from "../../types/misc";
 import { CorePlugin } from "../core_plugin";
-
-export class HeaderGlobalVisibilityPlugin extends CorePlugin {
+import { HeaderGroupingPlugin } from "./header_grouping";
+import { HeaderVisibilityPlugin } from "./header_visibility";
+
+export class HeaderGlobalVisibilityPlugin extends CorePlugin<
+  any,
```
LucasLefevre: any -> never

src/plugins/core/squisher.ts
```diff
@@ -34,7 +34,10 @@ export const SEPARATOR = "|";
 export const NO_CHANGE = "=";
 
 export class Squisher {
-  private readonly getters: CoreGetters;
+  private readonly getters: Pick<
+    CoreGetters,
+    "getCell" | "getSheetName" | "getRangeString" | "getLocale" | "getRangeFromSheetXC"
```
LucasLefevre: there's a `UnsquishGetters`, there could be a `SquishGetters`

src/plugins/core_plugin.ts
```diff
@@ -1,12 +1,58 @@
...
+ *   4. RangeAdapterGetters & …                                           (always included)
+ */
+
+/** Converts a union `A | B | C` into an intersection `A & B & C` via function-parameter contravariance. */
+type UnionToIntersection<U> = (U extends any ? (x: U) => void : never) extends (x: infer I) => void
+  ? I
+  : never;
+
+/** Transitive closure of the dependency DAG: the plugin itself plus every dependency, recursively. Depth-capped at 8. */
+type _Reach<P extends CorePluginConstructor, D extends unknown[] = []> = D extends { length: 8 }
```
LucasLefevre: 🧦 
no reason to only prefix this one but no other ?
```suggestion
type Reach<P extends CorePluginConstructor, D extends unknown[] = []> = D extends { length: 8 }
```

src/plugins/core_plugin.ts
```diff
@@ -1,12 +1,67 @@
...
+export type AnyPluginConstructor = {
+  new (...args: unknown[]): any;
+  getters: readonly string[];
+  dependencies: readonly AnyPluginConstructor[];
+};
+
+/** Transitive closure of the dependency DAG: the plugin itself plus every dependency, recursively. Depth-capped at 8. */
+type _Reach<P extends AnyPluginConstructor, D extends unknown[] = []> = D extends { length: 8 }
+  ? P
+  : P | _Reach<P["dependencies"][number], [...D, unknown]>;
```
LucasLefevre: worth a comment to explain `D`'s purpose it only to cap recursivity, and without cap, it doesn't resolve (or rather resolves to `any`).
(at least that's what I think I understand)

src/plugins/core_plugin.ts
```diff
@@ -1,12 +1,67 @@
...
+ *
+ *   2. DistGetters<typeof MergePlugin | typeof CellPlugin | typeof SheetPlugin>
+ *        → MergeGetters | CellGetters | SheetGetters                     (map each ctor to its getter type)
+ *        The distributive conditional `T extends … ? PluginGetters<T> : never` is what makes
+ *        TypeScript apply PluginGetters to each union member separately rather than to the whole union.
+ *
+ *   3. UnionToIntersection<MergeGetters | CellGetters | SheetGetters>
+ *        → MergeGetters & CellGetters & SheetGetters                     (flip union → intersection)
+ *        Achieved via contravariance of function parameters: a union of `(x: A) => void` types
+ *        can only be inferred as `(x: A & B & C) => void`, so `infer I` gives the intersection.
```
LucasLefevre: I'd move this part to `UnionToIntersection`'s docstring

src/plugins/core_plugin.ts
```diff
@@ -1,12 +1,67 @@
...
+ *
+ *   4. RangeAdapterGetters & …                                           (always included)
+ */
+
+/** Converts a union `A | B | C` into an intersection `A & B & C` via function-parameter contravariance. */
+type UnionToIntersection<U> = (U extends any ? (x: U) => void : never) extends (x: infer I) => void
```
LucasLefevre: I'd move this type to `misc.ts` or a new `utility.ts`. Could be useful in other situations, and it's not related to plugins

src/plugins/core_plugin.ts
```diff
@@ -1,12 +1,67 @@
...
+/** Distributive map: turns a union of plugin constructors into a union of their getter types. */
+type DistGetters<T extends AnyPluginConstructor> = T extends AnyPluginConstructor
+  ? PluginGetters<T>
+  : never;
+
+/**
+ * The type of `this.getters` inside a core plugin.
+ * Automatically includes the getters of every plugin declared in `Class.dependencies`, transitively.
+ */
+export type DepsGetters<Plugin extends AnyPluginConstructor> = RangeAdapterGetters &
```
LucasLefevre: You have to pause and focus for a while to understand all those types, but it's actually understandable

src/plugins/core/image.ts
```diff
@@ -8,14 +8,15 @@ import { HeaderIndex, PixelPosition, UID } from "../../types/misc";
...
+import { CarouselPlugin } from "./carousel";
 import { FigurePlugin } from "./figures";
 
 interface ImageState {
   readonly images: Record<UID, Record<UID, Image | undefined> | undefined>;
 }
 
 export class ImagePlugin extends CorePlugin<ImageState, typeof ImagePlugin> implements ImageState {
-  static readonly dependencies = [FigurePlugin] as const;
+  static readonly dependencies = [FigurePlugin, CarouselPlugin] as const;
```
LucasLefevre: Looks like ImagePlugin doesn't actually depends on CarouselPlugin

src/plugins/plugin_registries.ts
```diff
@@ -65,7 +65,59 @@ import { HeaderPositionsUIPlugin } from "./ui_stateful/header_positions";
...
+
+    // Insert right after the last direct dependency already in the registry (O(n)).
+    // Use max() so we never move the insertion point backward: same-dep siblings
+    // stay in registration order rather than being reversed.
+    const entries = Object.entries(this.content);
+    const depSet = new Set(plugin.dependencies as CorePluginConstructor[]);
+    let insertAfter = entries.length - 1; // default: append at end
+    for (let i = 0; i < entries.length; i++) {
+      if (depSet.has(entries[i][1])) {
+        insertAfter = Math.max(insertAfter, i);
```
LucasLefevre: this does nothing because `insertAfter` is initialized at the max value `i` will ever take (`entries.length`)

which means every plugin is just always inserted at the end.

src/model.ts
LucasLefevre, line 1: I think the last few commits introducing the dependency system could be squashed, they rewrite each other

src/model.ts
```diff
@@ -83,6 +83,8 @@ const enum Status {
  */
 export class Model extends EventBus<any> implements CommandDispatcher {
   private corePlugins: CorePlugin[] = [];
+  private corePluginGetters = new Map<CorePluginConstructor, CoreGetters>();
+  private baseGetters: CoreGetters = {} as CoreGetters;
```
LucasLefevre, line 87: that's what they are, isn't it ?
```suggestion
  private rangeGetters: CoreGetters = {} as CoreGetters;
```
if yes, is it needed to have an additional attribute ? could we use `this.corePluginGetters.get(RangePlugin)` at the only place where `baseGetters` is used ?

tests/model/__snapshots__/model.test.ts.snap
LucasLefevre: could be nice indeed

src/plugins/core_plugin.ts
LucasLefevre: if we want a better "error", it could be something like that:

Don't know if it's worth the more complex type.

```diff
diff --git a/src/plugins/core_plugin.ts b/src/plugins/core_plugin.ts
index 558687e5c..30374562d 100644
--- a/src/plugins/core_plugin.ts
+++ b/src/plugins/core_plugin.ts
@@ -69,6 +69,12 @@ export interface CorePluginConstructor {
   readonly dependencies: readonly CorePluginConstructor[];
 }
 
+type MissingDependency = { readonly brand: unique symbol };
+
+type ScopedGetters<Plugin extends CorePluginConstructor> = DepsGetters<Plugin> & {
+    [key in Exclude<keyof CoreGetters, keyof DepsGetters<Plugin>>]: MissingDependency;
+  };
+
 /**
  * Core plugins handle spreadsheet data.
  * They are responsible to import, export and maintain the spreadsheet
@@ -81,14 +87,14 @@ export class CorePlugin<State = any, Self extends CorePluginConstructor = any>
 {
   static readonly dependencies: readonly CorePluginConstructor[] = [];
 
-  protected getters: DepsGetters<Self> & Partial<CoreGetters>;
+  protected getters: ScopedGetters<Self>;
   protected dispatch: CoreCommandDispatcher["dispatch"];
   protected canDispatch: CoreCommandDispatcher["dispatch"];
 
   constructor({ getters, stateObserver, range, dispatch, canDispatch }: CorePluginConfig) {
     super(stateObserver);
     range.addRangeProvider(this.adaptRanges.bind(this));
-    this.getters = getters as unknown as DepsGetters<Self> & Partial<CoreGetters>;
+    this.getters = getters as unknown as ScopedGetters<Self>;
     this.dispatch = dispatch;
     this.canDispatch = canDispatch;
   }
```

tests/model/model.test.ts
```diff
@@ -422,74 +427,102 @@ describe("Model", () => {
...
+function makePlugin(name: string, deps: CorePluginConstructor[] = []): CorePluginConstructor {
+  return class {
+    static getters = [] as const;
+    static readonly dependencies = deps;
+    static readonly name = name;
+  } as unknown as CorePluginConstructor;
+}
+
+describe("sortByDependencies", () => {
+  test("dependencies sorting", () => {
```
LucasLefevre, line 439: also test with insertion order different than order in the dependency graph

LucasLefevre (commented):
nice work!

a few things that are less easy than before (`getters: Pick<CoreGetters, ...>`, more `beforeHandle`), but I think it'll help us move forward and improve on that later.
The general state of things is more sound than it was before 👍

src/plugins/core/cell.ts
```diff
@@ -224,6 +238,22 @@ export class CellPlugin extends CorePlugin<CoreState> implements CoreState {
     fn(cmd.sheetId, styleReference, insertedElements);
   }
 
+  private duplicateSheet(sheetIdFrom: UID, sheetIdTo: UID) {
+    for (const cell of Object.values(this.getters.getCells(sheetIdFrom))) {
```
LucasLefevre: `getCells` already returns `Cell[]`
```suggestion
    for (const cell of this.getters.getCells(sheetIdFrom)) {
```

src/plugins/core/data_validation.ts
```diff
@@ -120,6 +123,25 @@ export class DataValidationPlugin
     return CommandResult.Success;
   }
 
+  beforeHandle(cmd: CoreCommand): void {
+    switch (cmd.type) {
+      case "DELETE_CONTENT":
```
LucasLefevre, line 127: Actually...we could just do the updates here, instead of storing temporary data, and iterate over everything twice, etc...
We never used `beforeHandle` much but nothing prevents updating the state

tests/model/model.test.ts
rrahir: yup, I wanted to make sure that having the full picture in a single commit was ok and hoped to merged them together.

src/model.ts
rrahir: I need to check the code but it could contain other getters (like the "giveMeAllTheFOrmulas" that ADRM introduces in another PR to track used datasources accross plugins.

src/plugins/core/cell.ts
rrahir: this or historize the entry as I did ...

src/plugins/core_plugin.ts
rrahir: I do think it's good to avoid having people reinvent the wheel, but lucas suggestion to raise an explicit error seems like a good compromise

PR #8759, opened 2026-05-20
[FIX] Unhide hearders in dark mode

matho-odoo:
CSS issue, they are only visible when hovered
But should be visible all the time.

Task: [6233467](https://www.odoo.com/odoo/2328/tasks/6233467)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/headers_overlay/headers_overlay.css
+++ b/src/components/headers_overlay/headers_overlay.css
@@ -1,5 +1,6 @@
 .o-spreadsheet {
   .o-col-resizer {
+    z-index: calc(var(--os-components-importance-grid) + 1);
     position: absolute;
     top: 0;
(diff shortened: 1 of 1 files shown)


src/components/headers_overlay/headers_overlay.css
```diff
@@ -1,5 +1,6 @@
 .o-spreadsheet {
```
hokolomopo, line 1: The icons also seem to have the wrong color in versions before master. probably needs a color-scheme: light; in those versions.

(can create a separata task if you prefer)

Image

LucasLefevre (approved):
robodoo r+

PR #8771, opened 2026-05-21
[IMP/FIX] charts: render chart font sizes consistently with cells

dhrp-odoo:
## Description:

Cell and chart font sizes are both stored as point values, but some chart renderers passed them directly to Chart.js, canvas, or CSS APIs that interpret them as pixels. As a result, chart text could render smaller than grid cells for the same configured font size.

Convert chart font sizes from pt to px only at render time so chart titles, axis titles, gauge, scorecard, treemap, sunburst, and carousel texts remain visually consistent with cells.

Task: [6170954](https://www.odoo.com/odoo/2328/tasks/6170954)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


LucasLefevre (approved):
robodoo r+

PR #8772, opened 2026-05-21
[PERF] charts: avoid useless evaluation

anhe-odoo:
## Description

This task aims to avoid useless reevaluation of charts when a cell not related to the data of the chart is updated. Here, when a cell is updated (through UPDATE_CELL), we compute the list of the cells depending on this one, and then check for each chart if there is some updated cells in the dataranges of the chart.

## Related Task

- Task: [6226590](https://www.odoo.com/odoo/2328/tasks/6226590)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/figures/charts/gauge_chart.ts
+++ b/src/helpers/figures/charts/gauge_chart.ts
@@ -3,6 +3,7 @@ import {
   DEFAULT_GAUGE_MIDDLE_COLOR,
   DEFAULT_GAUGE_UPPER_COLOR,
 } from "../../../constants";
+import { CompiledFormula } from "../../../formulas/compiler";
 import { isMultipleElementMatrix, toScalar } from "../../../functions/helper_matrices";
 import { tryToNumber } from "../../../functions/helpers";
 import { BasePlugin } from "../../../plugins/base_plugin";
@@ -228,6 +229,31 @@ export const GaugeChart: ChartTypeBuilder<"gauge"> = {
   },
 
   getDefinitionForExcel: () => undefined,
+
+  getRanges(definition, getters, sheetId) {
+    const ranges: Range[] = [definition.dataRange].filter(
+      (r): r is Range => r !== undefined && !r.invalidXc && !r.invalidSheetName
+    );
+    if (definition.sectionRule) {
+      const formulaValues = [
+        definition.sectionRule.rangeMin,
+        definition.sectionRule.rangeMax,
+        definition.sectionRule.lowerInflectionPoint.value,
+        definition.sectionRule.upperInflectionPoint.value,
+      ];
+      for (const formula of formulaValues) {
+        if (formula.startsWith("=")) {
+          for (const range of CompiledFormula.Compile(formula, sheetId, getters)
+            .rangeDependencies) {
+            if (!range.invalidXc && !range.invalidSheetName) {
+              ranges.push(range);
+            }
+          }
+        }
+      }
+    }
+    return ranges;
+  },
 
   getContextCreation(definition) {
     return {
--- a/src/plugins/ui_core_views/cell_evaluation/evaluation_plugin.ts
+++ b/src/plugins/ui_core_views/cell_evaluation/evaluation_plugin.ts
@@ -22,7 +22,7 @@ import {
   UID,
   Zone,
 } from "../../../types/misc";
-import { Range } from "../../../types/range";
+import { BoundedRange, Range } from "../../../types/range";
 import { ExcelWorkbookData } from "../../../types/workbook_data";
 import { SquishedFormula } from "../../core/squisher";
 import { CoreViewPlugin, CoreViewPluginConfig } from "../../core_view_plugin";
@@ -159,6 +159,7 @@ export class EvaluationPlugin extends CoreViewPlugin {
     "isArrayFormulaSpillBlocked",
     "isEmpty",
     "getPerfProfile",
+    "getCellsDependingOn",
   ] as const;
 
   private shouldRebuildDependenciesGraph = true;
@@ -309,6 +310,19 @@ export class EvaluationPlugin extends CoreViewPlugin {
 
   getPerfProfile(): PerfProfile | undefined {
     return this.evaluator.getPerfProfile();
+  }
+
+  /**
+   * Returns ranges of formula cells that (transitively) depend on the given cell position.
+   */
+  getCellsDependingOn(position: CellPosition): Iterable<BoundedRange> {
+    const zone = {
+      left: position.col,
+      right: position.col,
+      top: position.row,
+      bottom: position.row,
+    };
+    return this.evaluator.getCellsDependingOn([{ sheetId: position.sheetId, zone }]);
   }
 
   /**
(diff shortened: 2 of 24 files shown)


tests/figures/chart/chart_animations.test.ts
```diff
@@ -72,10 +73,11 @@ describe("Chart animations in dashboard", () => {
 
     expect(mockedChart.config.options.animation).toEqual({ animateRotate: true });
 
-    // Dispatch a command that doesn't change the chart data
+    // Dispatch a command that doesn't change the chart data (cell outside chart range)
+    // The chart is not updated at all, so animation config is unchanged
     setCellContent(model, "A50", "6");
     await nextTick();
-    expect(mockedChart.config.options.animation).toBe(false);
+    expect(mockedChart.config.options.animation).toEqual({ animateRotate: true });
```
anhe-odoo, line 80: @hokolomopo I'm not sure about this change. Any opinion ?

src/helpers/figures/charts/gauge_chart.ts
```diff
@@ -229,6 +230,31 @@ export const GaugeChart: ChartTypeBuilder<"gauge"> = {
 
   getDefinitionForExcel: () => undefined,
 
+  getRanges(definition, getters, sheetId) {
```
hokolomopo, line 233: Instead of adding a getRange to the charts, I'd base myself on this PR https://github.com/odoo/o-spreadsheet/pull/8556/changes#diff-b3e31e50ca41843ddbaccd214f3f575524867f87b28d2ac4702074a3dcddba4a and do a getFormula. 

getRange goes in the opposite direction of lul's refractoring, which made the chart unaware of ranges. 

And then a helper to combine data source ranges + ranges in formulas

src/plugins/ui_core_views/cell_evaluation/evaluation_plugin.ts
```diff
@@ -311,6 +312,19 @@ export class EvaluationPlugin extends CoreViewPlugin {
...
+  /**
+   * Returns ranges of formula cells that (transitively) depend on the given cell position.
+   */
+  getCellsDependingOn(position: CellPosition): Iterable<BoundedRange> {
+    const zone = {
+      left: position.col,
+      right: position.col,
+      top: position.row,
+      bottom: position.row,
+    };
```
hokolomopo, line 324: we have a positionToZone (or something like that) helper somewhere

src/plugins/ui_core_views/evaluation_chart.ts
```diff
@@ -33,14 +35,17 @@ export class EvaluationChartPlugin extends CoreViewPlugin<EvaluationChartState>
...
     ) {
       for (const chartId in this.charts) {
         this.charts[chartId] = undefined;
       }
     }
 
     switch (cmd.type) {
+      case "UPDATE_CELL":
+        this.invalidateChartsAffectedByCell(cmd.sheetId, cmd.col, cmd.row);
+        break;
```
hokolomopo, line 48: a bit strange that the optimization is only for UPDATE_CELL, and not for all the other commands invalidating the evaluation. Better than nothing I guess 🤷 See with pro, he tried to do something similar

src/plugins/ui_core_views/evaluation_chart.ts
```diff
@@ -33,14 +35,17 @@ export class EvaluationChartPlugin extends CoreViewPlugin<EvaluationChartState>
...
-      invalidateChartEvaluationCommands.has(cmd.type)
+      (invalidateChartEvaluationCommands.has(cmd.type) && cmd.type !== "UPDATE_CELL")
     ) {
       for (const chartId in this.charts) {
         this.charts[chartId] = undefined;
       }
     }
 
     switch (cmd.type) {
+      case "UPDATE_CELL":
```
hokolomopo, line 46: also since it's a PERF commit, try a small benchmark of the cost of of doing this, VS the cost os re-drawing a chart

tests/figures/chart/chart_evaluation_invalidation.test.ts
```diff
@@ -0,0 +1,411 @@
...
+      setCellContent(model, "B1", "10");
+      const chartId = "chartId";
+      createScorecardChart(model, { keyValue: "A1", baseline: "B1" }, chartId);
+
+      model.getters.getChartRuntime(chartId);
+      const spy = spyOnGetRuntime();
+      setCellContent(model, "B1", "20");
+      model.getters.getChartRuntime(chartId);
+
+      expect(spy).toHaveBeenCalledTimes(1);
```
hokolomopo, line 288: can't you just store the runtime and check that it's not the same object returned by getChartRuntime anymore instead of the spy stuff ?

tests/figures/chart/chart_evaluation_invalidation.test.ts
```diff
@@ -0,0 +1,411 @@
...
+        lowerInflectionPoint: { type: "number", value: "33", operator: "<=" },
+        upperInflectionPoint: { type: "number", value: "66", operator: "<=" },
+        ...sectionRuleOverride,
+      },
+    },
+    chartId
+  );
+}
+
+describe("Chart runtime cache invalidation", () => {
```
hokolomopo, line 43: nitpick: i'd put the test inside the different chart_plugin.test.ts files instead of a new file

tests/figures/chart/chart_animations.test.ts
hokolomopo: most likely fine.

hokolomopo (commented):
👋

src/helpers/figures/charts/gauge_chart.ts
anhe-odoo: But apart from the gauge there are no formula in any other chart, no ? I'm not for the idea of a getformula in a scatter plot that don't use formula. Maybe we can find a "juste milieu" if you do not want the word "range" appears in the chart, like "dataSource" ?

tests/figures/chart/chart_evaluation_invalidation.test.ts
anhe-odoo: Yes, I can, but I though that is was better to check if the method has been called instead of checking if it's the same object when it shouldn't have been reconstructed (for other tests)

src/helpers/figures/charts/gauge_chart.ts
hokolomopo: Yeah I wanted to add getFormula only for gauge, but we did want to define properties on every chart and not have defaults.

That being said it looks like we already have range stull in the charts with `updateRanges`, so `getRange` isn't that bad I guess (those should really disappear once every chart have data sources, but bubble chart make it very annoying).

Otherwise as a more general question: how worth is this PR do you think ?  It's a +600 lines across 24 files, just to improve UPDATE_CELL with charts inside the viewport ? Is it that impactful ?

anhe-odoo: I'll make some benchmark, but I'm not sure it's really worth it, as with the other commands I added today, we will be around 1000+ lines at the end :/

PR #8774, opened 2026-05-22
[PERF] backport vectorization improvements

rrahir:
## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/functions/helpers.ts
+++ b/src/functions/helpers.ts
@@ -559,27 +559,32 @@ export function applyVectorization(
 
   // Resolve each arg's access pattern once, outside the inner loop.
   type ArgGetter = (i: number, j: number) => Arg;
-  const argGetters: ArgGetter[] = new Array(args.length);
+  const argGetters: ArgGetter[] = [];
+  const vectorizedIndices: number[] = []; // tracks which slots need updating each iteration.
   for (let k = 0; k < args.length; k++) {
     const arg = args[k];
     switch (vectorArgsType?.[k]) {
       case "matrix": {
-        argGetters[k] = (i, j) => arg![i][j];
+        argGetters.push((i, j) => arg![i][j]);
+        vectorizedIndices.push(k);
         break;
       }
       case "horizontal": {
-        argGetters[k] = (i) => arg![i][0];
+        argGetters.push((i) => arg![i][0]);
+        vectorizedIndices.push(k);
         break;
       }
       case "vertical": {
-        argGetters[k] = (_i, j) => arg![0][j];
+        argGetters.push((_i, j) => arg![0][j]);
+        vectorizedIndices.push(k);
         break;
       }
       case undefined:
-        argGetters[k] = () => arg;
+        argsBuffer[k] = arg;
         break;
     }
   }
+  const nbVectorized = vectorizedIndices.length;
 
   return generateMatrix(countVectorizedCol, countVectorizedRow, (col, row) => {
     if (col > vectorizedColLimit - 1 || row > vectorizedRowLimit - 1) {
@@ -587,8 +592,8 @@ export function applyVectorization(
         _t("Array arguments to [[FUNCTION_NAME]] are of different size.")
       );
     }
-    for (let k = 0; k < argGetters.length; k++) {
-      argsBuffer[k] = argGetters[k](col, row);
+    for (let k = 0; k < nbVectorized; k++) {
+      argsBuffer[vectorizedIndices[k]] = argGetters[k](col, row);
     }
     const singleCellComputeResult = formula(...argsBuffer);
     // In the case where the user tries to vectorize arguments of an array formula, we will get an
--- a/src/functions/helpers.ts
+++ b/src/functions/helpers.ts
@@ -557,25 +557,29 @@ export function applyVectorization(
   // Reused across every vectorized cell to avoid allocating a new args array per call.
   const argsBuffer: Arg[] = new Array(args.length);
 
-  const fillArgsBuffer = (i: number, j: number): void => {
-    for (let k = 0; k < args.length; k++) {
-      const arg = args[k];
-      switch (vectorArgsType?.[k]) {
-        case "matrix":
-          argsBuffer[k] = arg![i][j];
-          break;
-        case "horizontal":
-          argsBuffer[k] = arg![i][0];
-          break;
-        case "vertical":
-          argsBuffer[k] = arg![0][j];
-          break;
-        case undefined:
-          argsBuffer[k] = arg;
-          break;
+  // Resolve each arg's access pattern once, outside the inner loop.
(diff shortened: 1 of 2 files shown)


rrahir:
robodoo rebase-ff

src/functions/helpers.ts
```diff
@@ -560,27 +554,37 @@ export function applyVectorization(
...
   }
 
-  const getArgOffset: (i: number, j: number) => Arg[] = (i, j) =>
-    args.map((arg, index) => {
-      switch (vectorArgsType?.[index]) {
+  // Reused across every vectorized cell to avoid allocating a new args array per call.
+  const argsBuffer: Arg[] = new Array(args.length);
+
+  const fillArgsBuffer = (i: number, j: number): void => {
+    for (let k = 0; k < args.length; k++) {
```
hokolomopo, line 561: can you use real names for the variables? very hard to understand what this does. k is the argIndex, is i the col ? Or the colOffset ? It wasn't very clear before, but even worse now 😅 

Talking about a buffer also sounds very esoteric IMO. It's just `functionArgs`/`computeArgsAtPosition` no ?

src/functions/index.ts
```diff
@@ -111,14 +113,19 @@ function createComputeFunction(
   descr: FunctionDescription,
   functionName: string
 ): ComputeFunction<Matrix<FunctionResultObject> | FunctionResultObject> {
+  // Shared across vectorizedCompute→errorHandlingCompute within a single (synchronous) call.
+  let currentArgDefinitions: ArgDefinition[] = [];
```
hokolomopo, line 117: Seems a bit wonky to have `currentArgDefinitions` in this scope, and rely on the fact that we call the methods in the correct order so it is correctly defined in `errorHandlingCompute`. Can't you scope the array to `vectorizedCompute`, and pass it as argument of  `errorHandlingCompute` ? 

The variable could then also be named `argDefinitions`, w/o the `current` which I'm not sure what it refers to (current evaluation cycle ? current function call ?)

src/functions/helpers.ts
```diff
@@ -559,36 +559,41 @@ export function applyVectorization(
...
     if (col > vectorizedColLimit - 1 || row > vectorizedRowLimit - 1) {
       return new NotAvailableError(
         _t("Array arguments to [[FUNCTION_NAME]] are of different size.")
       );
     }
-    for (let k = 0; k < argGetters.length; k++) {
-      argsBuffer[k] = argGetters[k](col, row);
+    for (let k = 0; k < nbVectorized; k++) {
+      argsBuffer[vectorizedIndices[k]] = argGetters[k](col, row);
     }
```
hokolomopo, line 597: nitpick: I'd inline `vectorizedIndices.length` instead of creating another variables which adds even more confusion to this complex code

src/functions/helpers.ts
```diff
@@ -559,36 +559,41 @@ export function applyVectorization(
 
   // Resolve each arg's access pattern once, outside the inner loop.
   type ArgGetter = (i: number, j: number) => Arg;
-  const argGetters: ArgGetter[] = new Array(args.length);
+  const argGetters: ArgGetter[] = [];
+  const vectorizedIndices: number[] = []; // tracks which slots need updating each iteration.
```
hokolomopo, line 563: Can you do a single array `{argGetter; argIndex}[]` instead ? Maybe it'd be a bit easier to undersand

src/functions/helpers.ts
```diff
@@ -586,29 +586,36 @@ export function applyVectorization(
...
 
-  return generateMatrix(countVectorizedCol, countVectorizedRow, (col, row) => {
-    if (col > vectorizedColLimit - 1 || row > vectorizedRowLimit - 1) {
-      return new NotAvailableError(
-        _t("Array arguments to [[FUNCTION_NAME]] are of different size.")
-      );
-    }
-    for (let k = 0; k < nbVectorized; k++) {
-      argsBuffer[vectorizedIndices[k]] = argGetters[k](col, row);
+  const result: Matrix<FunctionResultObject> = new Array(countVectorizedCol);
```
hokolomopo, line 589: If inlining a loop really has such an impact on perf, we should investigate other places too ... Maybe `computeFunctionToObject` for example

src/functions/helpers.ts
```diff
@@ -586,6 +586,24 @@ export function applyVectorization(
   }
   const nbVectorized = vectorizedIndices.length;
 
+  // Specialize the call for common arities — argsBuffer.length is constant for the whole call,
```
hokolomopo, line 589: Not sure if "arities" is a real word (it's not recognized by my spell checker), and if it is I'm not sure if a comment that require a google search to understand is really useful ... (or maybe it's common knowledge, and I'm just bad at computer english)

src/functions/helpers.ts
```diff
@@ -586,6 +586,24 @@ export function applyVectorization(
...
+      break;
+    case 2:
+      callFormula = () => formula(argsBuffer[0], argsBuffer[1]);
+      break;
+    case 3:
+      callFormula = () => formula(argsBuffer[0], argsBuffer[1], argsBuffer[2]);
+      break;
+    default:
+      callFormula = () => formula(...argsBuffer);
+  }
```
hokolomopo, line 605: I mean if we stop caring that much about readability, and it's faster, why stop there at 3 ? Vlookup takes 4 arguments, Xlookup takes 6. Create a helper that returns a callback from 1 to 10 arguments, bind it, and call it a day.

hokolomopo (commented):
👋 

My comments aren't about whether the code works or not, just readability. This part of the code is already complex, and becomes even more so. Performance is fine and all, but we should at least do the maximal effort to keep the code somewhat readable.

Additionally, these optimization are very low-level, and look very dependent of the JS engine. Sure V8 is good and all, but should our benchmark really be only "what happens on the current version of node" ? What about older/more recent V8 versions ? Firefox ? Safari ? Or even what happens on a more realistic test dataset.

hokolomopo:
Oh it's a backport of master commits .... well my comments still stand. @LucasLefevre

src/functions/helpers.ts
LucasLefevre: it's a real word: https://en.wikipedia.org/wiki/Arity even though I agree it's not super common knowledge, it's exactly what it's used for here.

LucasLefevre:
@hokolomopo can be improved, but then I'd merge it as it is in 18.0, then directly refactor in 18.0 and forward-port, that would be probably easier

rrahir:
@hokolomopo I agree with Lucas that we can first benefit from the improvement and apply your suggestions afterwards.

pro-odoo:
robodoo up to saas-19.3

pro-odoo:
robodoo r+

PR #8777, opened 2026-05-22
[FIX] Shortcuts: fix Macos Alt shortcuts

rrahir:
keyboard shortcuts involving the "Alt" (names Options) key do not work properly in macos because the options key alters the output of other alphanumeric keys (e.g. Options + T outputs the character †, an the browser then detects a keyboard event for which the key is `†`.

this revision replaces the shortcut modifier from Alt to Ctrl for mac os and iOs.

Task: 6240362

## Description:

Task: [6240362](https://www.odoo.com/odoo/2328/tasks/6240362)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/helpers/dom_helpers.ts
+++ b/src/components/helpers/dom_helpers.ts
@@ -215,7 +215,7 @@ export function keyboardEventToShortcutString(
     if (isCtrlKey(ev)) {
       keyDownString += "Ctrl+";
     }
-    if (ev.altKey) {
+    if (isAltKey(ev)) {
       keyDownString += "Alt+";
     }
     if (ev.shiftKey) {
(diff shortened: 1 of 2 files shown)


src/components/helpers/dom_helpers.ts
```diff
@@ -240,6 +240,10 @@ export function isCtrlKey(ev: KeyboardEvent | MouseEvent): boolean {
   return isMacOS() || isIOS() ? ev.metaKey : ev.ctrlKey;
 }
 
+function isAltKey(ev: KeyboardEvent | MouseEvent): boolean {
+  return isMacOS() || isIOS() ? ev.ctrlKey : ev.altKey;
+}
```
hokolomopo, line 245: Some questions: 

1) So in mac, ctrl is command, and alt is ctrl ? the hell ... 
2) I see we have a test `Mac user have ⌘ in menu instead of Ctrl`, we should do the same for alt menu descriptions since they don't have an alt key
3) If I understand correctly, on mac the "alt" equivalent would be the "option" key ? But they have native shortcuts "option + something" right ? We should make sure we don't collide with those

rrahir: 1. The rule of thumb is indeed that _**Ctrl+any**_ shortcuts have a mac equivalent of _**Cmd+any**_ 
and the equivalent "alt" command is officially called Options but in browsers, it is still considered as the AltKey modifier.
2. will do
3. there already are some shortcuts that conflict right now with native stuff. Ctrl +r comes to mind. From what I can tell, the ALt+F11 and alt shift stuff do not conflict with native shortcuts.

PR #8791, opened 2026-05-27
[FIX] table: correctly insert table on static pivots

hokolomopo:
## Description

On a dynamic pivot, the topbar button to create/edit a table opens the pivot design side panel.

But on static pivot formulas, it should create/edit a normal table, as static pivots formulas are not affected by pivot table style.

Task: [6204984](https://www.odoo.com/odoo/2328/tasks/6204984)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/tables/table_dropdown_button/table_dropdown_button.ts
+++ b/src/components/tables/table_dropdown_button/table_dropdown_button.ts
@@ -2,7 +2,7 @@ import { Component, useState } from "@odoo/owl";
 import { ActionSpec } from "../../../actions/action";
 import { DEFAULT_TABLE_CONFIG } from "../../../helpers/table_presets";
 import { interactiveCreateTable } from "../../../helpers/ui/table_interactive";
-import { positions } from "../../../helpers/zones";
+import { cellPositions } from "../../../helpers/zones";
 import { _t } from "../../../translation";
 import { UID } from "../../../types";
 import { SpreadsheetChildEnv } from "../../../types/spreadsheet_env";
@@ -52,7 +52,7 @@ export class TableDropdownButton extends Component<Props, SpreadsheetChildEnv> {
       this.closePopover();
       return;
     }
-    const pivotId = this.pivotIdInSelection;
+    const pivotId = this.dynamicPivotIdInSelection;
     if (pivotId) {
       this.env.openSidePanel("PivotSidePanel", { pivotId, openTab: "design" });
       return;
@@ -78,7 +78,7 @@ export class TableDropdownButton extends Component<Props, SpreadsheetChildEnv> {
   }
 
   get action(): ActionSpec {
-    const pivotId = this.pivotIdInSelection;
+    const pivotId = this.dynamicPivotIdInSelection;
     if (pivotId) {
       return {
         name: _t("Edit pivot style"),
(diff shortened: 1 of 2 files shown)


src/components/tables/table_dropdown_button/table_dropdown_button.ts
```diff
@@ -104,14 +104,22 @@ export class TableDropdownButton extends Component<Props, SpreadsheetChildEnv> {
...
-        const sheetId = this.env.model.getters.getActiveSheetId();
-        const pivotId = this.env.model.getters.getPivotIdFromPosition({ sheetId, ...position });
-        if (pivotId) {
-          return pivotId;
+    const dynamicPivotPositions = this.env.model.getters
+      .getCellsWithTrackedFormula("PIVOT")
+      .map((cellId) => this.env.model.getters.getCellPosition(cellId));
+
+    for (const position of dynamicPivotPositions) {
+      if (position.sheetId !== this.env.model.getters.getActiveSheetId()) {
```
rrahir:
```suggestion
      if (position.sheetId !== activeSheetId) {
```
store the activeSheetId in  a const maybe

src/components/tables/table_dropdown_button/table_dropdown_button.ts
```diff
@@ -104,14 +104,22 @@ export class TableDropdownButton extends Component<Props, SpreadsheetChildEnv> {
...
+  get dynamicPivotIdInSelection(): UID | undefined {
     const selection = this.env.model.getters.getSelectedZones();
-    for (const zone of selection) {
-      for (const position of positions(zone)) {
-        const sheetId = this.env.model.getters.getActiveSheetId();
-        const pivotId = this.env.model.getters.getPivotIdFromPosition({ sheetId, ...position });
-        if (pivotId) {
-          return pivotId;
+    const dynamicPivotPositions = this.env.model.getters
+      .getCellsWithTrackedFormula("PIVOT")
```
rrahir: if you have a metric ton of formulas; it's not the best :/  you'll have to convert everything and then pass through every position even after you have consumed all the positions of the selection.

I would invert the 2 sets of positions in the loops and/or consume the positions of the other to have an early return condition.

rrahir: this getter tracks the cells that  contain a pivot formula but not the cells that contain a spilled result. with your PR, if I select the range Pivots!D25:D26, the button title is now "edit table" instead of "edit pivot table" but when I click it, nothing occurs

rrahir (changes requested):

src/components/tables/table_dropdown_button/table_dropdown_button.ts
hokolomopo: Note that there shouldn't be that many spreads pivot formulas in a spreadsheet, and the main perf hit is probably `getPivotIdFromPosition`.

But you're right the can be improved.

src/components/tables/table_dropdown_button/table_dropdown_button.ts
```diff
@@ -104,17 +104,26 @@ export class TableDropdownButton extends Component<Props, SpreadsheetChildEnv> {
     return this.env.model.getters.getTableStyles();
   }
 
-  get pivotIdInSelection(): UID | undefined {
+  get dynamicPivotIdInSelection(): UID | undefined {
     const selection = this.env.model.getters.getSelectedZones();
+    const pivotCellIds = new Set(this.env.model.getters.getCellsWithTrackedFormula("PIVOT"));
```
rrahir, line 109: chti nitpick
```suggestion
    const pivotCellIds = new Set(this.env.model.getters.getCellsWithTrackedFormula("PIVOT"));
      if (pivotCellIds.size === 0) {
          return undefined;
        }
```

hokolomopo: sure :)

rrahir:
robodoo r+

PR #8794, opened 2026-05-27
[PERF] menu: avoid quadratic complexity with proxies

hokolomopo:
In the menu component, we would compute whether to display the icons for each menu item. But to compute that, we need to loop on each menu items, leading to a quadratic complexity.

This could have been good enough as menus aren't too large, but the menu items are proxified by Owl, making the loop very slow.

Task: [6250318](https://www.odoo.com/odoo/2328/tasks/6250318)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


LucasLefevre (approved):
robodoo r+

PR #8800, opened 2026-05-28
[FIX] Tables: invalidate computed style on `DELETE_CONTENT`

rrahir:
The command `DELETE_CONTENT` can sometimes dispatch a subcommand that deletes a table, which implies that the table style and therefore the cell computed style need to be invalidated. Since the subcommand does not reach Ui feature plugins when replaying commands, we need to explicitely invalidate the table & cell styles on `DELETE_CONTENT`.

Task: 6253199

## Description:

Task: [6253199](https://www.odoo.com/odoo/2328/tasks/6253199)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


LucasLefevre (approved):
robodoo r+

PR #8829, opened 2026-06-01
[IMP] carousel: open chart panel on chart double click

hokolomopo:
## Description

When double clicking in a chart that is inside a carousel, we should open the chart side panel. Double clicking on the carousel header should open the carousel side panel.

Task: [6264291](https://www.odoo.com/odoo/2328/tasks/6264291)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/figures/figure_carousel/figure_carousel.ts
+++ b/src/components/figures/figure_carousel/figure_carousel.ts
@@ -85,6 +85,15 @@ export class CarouselFigure extends Component<Props, SpreadsheetChildEnv> {
     this.env.openSidePanel("CarouselPanel", { figureId: this.props.figureUI.id });
   }
 
+  onCarouselChartDoubleClick() {
+    if (this.selectedCarouselItem?.type !== "chart") {
+      return;
+    }
+    const chartId = this.selectedCarouselItem.chartId;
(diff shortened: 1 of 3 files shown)


src/components/figures/figure_carousel/figure_carousel.ts
```diff
@@ -85,6 +85,15 @@ export class CarouselFigure extends Component<Props, SpreadsheetChildEnv> {
     this.env.openSidePanel("CarouselPanel", { figureId: this.props.figureUI.id });
   }
 
+  onCarouselChartDoubleClick() {
+    if (this.selectedCarouselItem?.type !== "chart") {
+      return;
+    }
+    const chartId = this.selectedCarouselItem.chartId;
+    this.env.model.dispatch("SELECT_FIGURE", { figureId: this.props.figureUI.id });
```
rrahir, line 93: nitpick I missed in the first pass: is it mandatory?  I think the figure is already selected from the first click on the figure itself

hokolomopo: probably not mandatory no, I just added it to make sure and because it cost nothing

rrahir:
@robodoo r+

PR #8835, opened 2026-06-02
[IMP] Model: plugins only see the grid selection

rrahir:
## Description:

Except from the grid selection, the other actors of the
SelectionProcessor only live in the stores, the plugins don't know about
them.
Since the plugins sometimes call the processor, it means they always
assume that they will impact the grid selection.

With this revision, every call to the selection processor coming from
the plugins will ensure that the grid selection is active.

Task: [6268168](https://www.odoo.com/odoo/2328/tasks/6268168)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/grid/grid.ts
+++ b/src/components/grid/grid.ts
@@ -910,4 +910,12 @@ export class Grid extends Component<SpreadsheetChildEnv> {
   get displaySelectionHandler() {
     return this.env.isMobile() && this.composerFocusStore.activeComposer.editionMode === "inactive";
   }
+
+  get displayTableResizer() {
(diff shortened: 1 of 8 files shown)


rrahir:
robodoo rebase-ff

hokolomopo (commented):
Not 100% sure about the second commit. If we change the selection into a store, I'm not sure if we can even achieve that ? And i'm not sure it's something plugin specific, 80% of stores don't want to touch SelectionInput's selection.

PR #8840, opened 2026-06-03
[FIX] Pivot: runtime definition might not be loaded

rrahir:
the fix in #8539 relies on the runtime definition of a Pivot to access its measures. Unfortunately, such definition is lazy loaded through the evaluation process so its not loaded (and will throw a loading error) if called when the pivot is not actually used.. which corresponds to the situation we try to fix in #8539.

Since the measures exist on the core definition of the pivot, we can use the core definition safely.

Note that an integration test is added in Odoo is added as well.

Task: 6267596

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


LucasLefevre (approved):
robodoo r+ rebase-ff

PR #8842, opened 2026-06-03
[FIX] pivot: prevent drag & drop in pivot side panel

hokolomopo:
## Description

In the pivot side panel, a pointerdown event on a button/input/select of an dimension should not trigger the drag & drop of the dimension.

The way we were handling that was all over the place. Some elements in the children had a `pointerdown.stop`, the `<select>` were handled directly before starting a drag & drop, and some elements didn't prevent a d&d where they should have.

We will now handle everything in the layout configurator. Before starting the d&d, we check that the event target is not an input/select, or has a class that indicates that it is interactive and should not trigger a d&d.

The solution is not the greatest, as we might forget some classes, but it beats having the children having to stop pointer events because of their parents.

Task: [6268403](https://www.odoo.com/odoo/2328/tasks/6268403)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/helpers/dom_helpers.ts
+++ b/src/components/helpers/dom_helpers.ts
@@ -288,3 +288,25 @@ function isOtherMobileOS() {
 export function isMobileOS() {
   return isAndroid() || isIOS() || isOtherMobileOS();
 }
+
+const INTERACTIVE_ELEMENT_SELECTOR =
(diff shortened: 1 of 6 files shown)


pro-odoo:
robodoo r+

PR #8844, opened 2026-06-03
[FIX] Figure: only focus when the selected figure changes

rrahir:
The condition of the useEffect to re-focus a figure were removed by mistake. This meant that every render, if there was a selected figure, its DOM element would be automatically focus.

There was a race condition with the component SelectionInput that sometimes let the latter have the focus, sometimes not, which lead to the weird behaviour described in the task:

- Insert a scorecard
- Add a key value
- Edit the key value and select it
- Hit backspace twice

=> The first press deletes the value, the second deletes the chart

Task: 6267606

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/figures/figure/figure.ts
+++ b/src/components/figures/figure/figure.ts
@@ -114,22 +114,25 @@ export class FigureComponent extends Component<SpreadsheetChildEnv> {
   setup() {
     const borderWidth = figureRegistry.get(this.props.figureUI.tag).borderWidth;
     this.borderWidth = borderWidth !== undefined ? borderWidth : BORDER_WIDTH;
-    useLayoutEffect(() => {
-      const selectedFigureIds = this.env.model.getters.getSelectedFigureIds();
-      const thisFigureId = this.props.figureUI.id;
-      const el = this.figureRef();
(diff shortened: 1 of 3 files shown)


src/components/figures/figure/figure.ts
```diff
@@ -114,22 +114,25 @@ export class FigureComponent extends Component<SpreadsheetChildEnv> {
...
+           * NOTE: `preventScroll`does not work on mobile but then again,
+           * mobile is not really supported ATM.
+           *
+           * TODO: When implementing proper mobile, we will need to scroll the viewport
+           * correctly (and render?) before focusing the element.
+           */
+          el?.focus({ preventScroll: true });
+        }
+      },
+      () => [this.env.model.getters.getSelectedFigureIds(), this.props.figureUI.id]
```
LucasLefevre, line 134: not super robust since it relies on the selected figures to be referentially equal, but the test fails if a reference to a new array if returned by `getSelectedFigureIds`, so 👍

LucasLefevre (approved):
robodoo r+

PR #8846, opened 2026-06-04
[IMP] headers: resize rows and columns from context menu

dhrp-odoo:
## Description:

Before this PR, the row and column context menus did not provide
any resizing actions.

This PR allows users to fit selected rows or columns to their data,
or set a custom size through a floating editor displayed next to the
clicked header. An empty custom size restores the default width or
height.

Task: [6234780](https://www.odoo.com/odoo/2328/tasks/6234780)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/grid/grid.ts
+++ b/src/components/grid/grid.ts
@@ -1,4 +1,5 @@
 import { onMounted, providePlugins, proxy, signal, useListener, useProps } from "@odoo/owl";
+import { Action, createAction } from "../../actions/action";
 import { insertSheet, insertTable } from "../../actions/insert_actions";
 import {
   CREATE_IMAGE,
@@ -21,7 +22,7 @@ import {
   interactivePaste,
   interactivePasteFromOS,
 } from "../../helpers/ui/paste_interactive";
-import { isInside } from "../../helpers/zones";
+import { isInside, positionToZone } from "../../helpers/zones";
 import { Component, useLayoutEffect } from "../../owl3_compatibility_layer";
 import { cellMenuRegistry } from "../../registries/menus/cell_menu_registry";
 import { colMenuRegistry } from "../../registries/menus/col_menu_registry";
@@ -40,6 +41,7 @@ import { ClipboardStore } from "../../stores/clipboard_store";
 import { HighlightStore } from "../../stores/highlight_store";
 import { ViewportsStore } from "../../stores/viewports_store";
 import { ZoomStore } from "../../stores/zoom_store";
+import { _t } from "../../translation";
 import { CellValueType } from "../../types/cells";
 import { ClipboardMIMEType } from "../../types/clipboard";
 import { Client } from "../../types/collaborative/session";
@@ -64,6 +66,7 @@ import { GridComposer } from "../composer/grid_composer/grid_composer";
 import { FiguresContainer } from "../figures/figure_container/figure_container";
 import { GridOverlay } from "../grid_overlay/grid_overlay";
 import { GridPopover } from "../grid_popover/grid_popover";
+import { HeaderResizeEditor } from "../header_resize_editor/header_resize_editor";
 import { HeadersOverlay } from "../headers_overlay/headers_overlay";
 import { cssPropertiesToCss } from "../helpers/css";
 import { getElBoundingRect, keyboardEventToShortcutString } from "../helpers/dom_helpers";
@@ -117,6 +120,11 @@ const registries = {
   UNGROUP_HEADERS: unGroupHeadersMenuRegistry,
 };
 
+interface HeaderResizeEditorTarget {
+  dimension: Dimension;
+  index: HeaderIndex;
+}
+
 // -----------------------------------------------------------------------------
 // JS
 // -----------------------------------------------------------------------------
@@ -126,6 +134,7 @@ export class Grid extends Component<SpreadsheetChildEnv> {
     GridComposer,
     GridOverlay,
     GridPopover,
+    HeaderResizeEditor,
     HeadersOverlay,
     MenuPopover,
     Autofill,
@@ -147,6 +156,7 @@ export class Grid extends Component<SpreadsheetChildEnv> {
   readonly HEADER_HEIGHT = HEADER_HEIGHT;
   readonly HEADER_WIDTH = HEADER_WIDTH;
   private menuState!: MenuState;
+  private headerResizeEditorTarget = signal<HeaderResizeEditorTarget | null>(null);
   private gridRef = signal.ref();
   private canvasRef = signal.ref(HTMLCanvasElement);
   private highlightStore!: Store<HighlightStore>;
@@ -687,8 +697,8 @@ export class Grid extends Component<SpreadsheetChildEnv> {
 
   onInputContextMenu(ev: MouseEvent) {
     ev.preventDefault();
-    const lastZone = this.env.model.getters.getSelectedZone();
-    const { left: col, top: row } = lastZone;
+    const position = this.env.model.getters.getActivePosition();
+    const { col, row, sheetId } = position;
     let type: ContextMenuType = "CELL";
     this.composerFocusStore.activeComposer.stopEdition();
     if (this.env.model.getters.getActiveCols().has(col)) {
@@ -696,12 +706,18 @@ export class Grid extends Component<SpreadsheetChildEnv> {
     } else if (this.env.model.getters.getActiveRows().has(row)) {
       type = "ROW";
     }
-    const { x, y, width } = this.viewStore.viewports.getVisibleRectWithZoom(
-      this.env.model.getters.getActiveSheetId(),
-      lastZone
-    );
+
+    let anchorZone = positionToZone(position);
+    const sheetZone = this.env.model.getters.getSheetZone(sheetId);
+    // Full-height/full-width header zones remain visible when the active cell is offscreen.
+    if (type === "COL") {
+      anchorZone = { ...sheetZone, left: col, right: col };
+    } else if (type === "ROW") {
+      anchorZone = { ...sheetZone, top: row, bottom: row };
+    }
+    const { x, y } = this.viewStore.viewports.getVisibleRectWithZoom(sheetId, anchorZone);
     const gridRect = this.getGridRect();
-    this.toggleContextMenu(type, gridRect.x + x + width, gridRect.y + y);
+    this.toggleContextMenu(type, gridRect.x + x, gridRect.y + y);
   }
 
   onCellRightClicked(col: HeaderIndex, row: HeaderIndex, { x, y }: DOMCoordinates) {
@@ -721,6 +737,60 @@ export class Grid extends Component<SpreadsheetChildEnv> {
     this.toggleContextMenu(type, x, y);
   }
 
+  private getMenuItems(type: ContextMenuType, col: HeaderIndex, row: HeaderIndex): Action[] {
+    const menuItems = registries[type].getMenuItems();
+    const headerIndex = type === "COL" ? col : type === "ROW" ? row : -1;
+    if (headerIndex === -1) {
+      return menuItems;
+    }
+
+    // Build the resize action on demand so it captures the target header index.
+    const isColumn = type === "COL";
+    const unhideAction = menuItems.find(
+      ({ id }) => id === (isColumn ? "unhide_columns" : "unhide_rows")
+    );
+    if (!unhideAction) {
+      throw new Error("Missing unhide action in the header menu");
+    }
+    menuItems.push(
+      createAction({
+        id: isColumn ? "resize_columns" : "resize_rows",
+        name: isColumn ? _t("Resize column") : _t("Resize row"),
+        sequence: unhideAction.sequence + 1,
+        separator: true,
+        icon: isColumn
+          ? "o-spreadsheet-Icon.RESIZE_HORIZONTAL"
+          : "o-spreadsheet-Icon.RESIZE_VERTICAL",
+        children: [
+          {
+            name: _t("Fit to data"),
+            execute: (env) => {
+              const sheetId = env.model.getters.getActiveSheetId();
+              return isColumn
+                ? env.model.dispatch("AUTORESIZE_COLUMNS", {
+                    sheetId,
+                    cols: [...env.model.getters.getActiveCols()],
+                  })
+                : env.model.dispatch("AUTORESIZE_ROWS", {
+                    sheetId,
+                    rows: [...env.model.getters.getActiveRows()],
+                  });
+            },
+          },
+          {
+            name: _t("Custom size"),
+            execute: () =>
+              this.headerResizeEditorTarget.set({
+                dimension: isColumn ? "COL" : "ROW",
+                index: headerIndex,
+              }),
+          },
+        ],
+      })
+    );
+    return menuItems.sort((a, b) => a.sequence - b.sequence);
+  }
+
   /**
    * expects x and y coordinates in true pixels (not zoomed)
    */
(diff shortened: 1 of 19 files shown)


src/components/side_panel/header_resize_panel/header_resize_panel.ts
```diff
@@ -0,0 +1,123 @@
...
+    inputValue: String(this.currentSize),
+  });
+
+  get minSize(): number {
+    return this.props.dimension === "COL" ? MIN_COL_WIDTH : MIN_ROW_HEIGHT;
+  }
+
+  get maxSize(): number {
+    return 2000;
+  }
```
hokolomopo: Why 2000 ? Not sure we want to bother doing a max size if we don't have a reason to, 2000 is not really different from 8000 or some other dumb value

src/components/side_panel/header_resize_panel/header_resize_panel.ts
```diff
@@ -0,0 +1,123 @@
...
+    }
+    return [];
+  }
+
+  onModeChanged(mode: ResizeMode) {
+    this.state.mode = mode;
+  }
+
+  onSizeChanged(inputValue: string) {
+    this.state.inputValue = inputValue;
```
hokolomopo: wouldn't it be simpler to parse the input value here (and update the error messages) ? So we do'nt have to parse it everywhere.

tests/menus/menu_items_registry.test.ts
```diff
@@ -1885,6 +1921,41 @@ describe("Menu Item actions", () => {
       });
     });
   });
+  describe("Resize Rows", () => {
```
hokolomopo:
```suggestion
});

  describe("Resize Rows", () => {
```
nitpick

tests/headers/header_resize_panel_component.test.ts
```diff
@@ -0,0 +1,169 @@
...
+
+    await click(fixture, SELECTORS.applyButton);
+    expect(dispatch).not.toHaveBeenCalled();
+    expect(onCloseSidePanel).not.toHaveBeenCalled();
+  });
+
+  test("Focusing the size input selects exact size mode", async () => {
+    await mountHeaderResizePanel();
+    await simulateClick(SELECTORS.fitToDataRadio);
+    expect(document.querySelector<HTMLInputElement>(SELECTORS.fitToDataRadio)!.checked).toBe(true);
```
hokolomopo:
```
    expect(SELECTORS.fitToDataRadio).toHaveValue(true)
```

hokolomopo (commented):
Some small issues, otherwise LTGM :)

src/components/side_panel/header_resize_panel/header_resize_panel.ts
dhrp-odoo: I just hard-coded from GSheet, But i will check this with FRGI :+1:

src/components/header_resize_editor/header_resize_editor.ts
```diff
@@ -0,0 +1,138 @@
...
+    };
+  }
+
+  get minSize(): number {
+    return this.props.dimension === "COL" ? MIN_COL_WIDTH : MIN_ROW_HEIGHT;
+  }
+
+  get maxSize(): number {
+    // Oversized headers can exceed browser scroll limits and make adjacent cells unreachable.
+    return 2000;
```
rrahir: Shouldn't this become the hard limit for every cell? if I write a huge text and use the autoresize,  it will set a size without any kindof limit. But if I open the resize editor afterwards, then I am bound by the limit of 2000px ?

rrahir: in anycase, it should become a constant

src/components/header_resize_editor/header_resize_editor.ts
```diff
@@ -0,0 +1,138 @@
...
+    if (!input) {
+      return;
+    }
+    if (!input.value && !input.validity.badInput) {
+      // A null size removes the custom size and restores the default.
+      return null;
+    }
+    const size = input.valueAsNumber;
+    if (!Number.isInteger(size)) {
+      this.errorMessage.set(_t("Size must be an integer."));
```
rrahir:
```suggestion
      this.errorMessage.set(_t("Size must be an integer"));
```
it's not really a full sentence, in odoo ,we generally do not punctuate short error message like these

src/components/header_resize_editor/header_resize_editor.ts
```diff
@@ -0,0 +1,138 @@
...
+      return;
+    }
+    if (size < this.minSize || size > this.maxSize) {
+      this.errorMessage.set(
+        _t(
+          "Size must be between %s and %s pixels.",
```
rrahir:
```suggestion
          "Size must be between %s and %s pixels",
```

src/components/header_resize_editor/header_resize_editor.ts
```diff
@@ -0,0 +1,138 @@
...
+    return this.props.dimension === "COL" ? MIN_COL_WIDTH : MIN_ROW_HEIGHT;
+  }
+
+  get maxSize(): number {
+    // Oversized headers can exceed browser scroll limits and make adjacent cells unreachable.
+    return 2000;
+  }
+
+  get defaultSize(): number {
+    return this.props.dimension === "COL" ? DEFAULT_CELL_WIDTH : DEFAULT_CELL_HEIGHT;
```
rrahir: the size of the rows being dynamic, you could have a header with a non default value because we set a text over several lines inside of it. In that case, the placeholder will show 23 px but if we leave it empty and press ok, then it will actually go back to the dynamic size of the row

src/components/grid/grid.ts
```diff
@@ -147,6 +155,7 @@ export class Grid extends Component<SpreadsheetChildEnv> {
   readonly HEADER_HEIGHT = HEADER_HEIGHT;
   readonly HEADER_WIDTH = HEADER_WIDTH;
   private menuState!: MenuState;
+  private headerResizeEditorState = signal<HeaderResizeEditorState | null>(null);
```
rrahir: i'm not yet super at ease with the new signal stuff but if it's a state, you should probably use `proxy` and adapt the state properties accordingly

src/components/grid/grid.ts
```diff
@@ -714,21 +728,69 @@ export class Grid extends Component<SpreadsheetChildEnv> {
...
+      this.headerResizeEditorState.set(null);
+    }
     this.menuState.isOpen = true;
     this.menuState.anchorRect = { x, y, width: 0, height: 0 };
     this.menuState.menuItems = registries[type].getMenuItems();
   }
 
+  onMenuClicked(ev: CustomEvent) {
+    const state = this.headerResizeEditorState();
+    if (ev.detail?.type === "OPEN_HEADER_RESIZE_EDITOR" && state) {
```
rrahir: This channel should have been removed when I removed the last known use. I don't think we use it anymore, even in Odoo.
If you want interactions between different components, using a store is probably the better way to go.
or maybe use the CellPopover, it' would be a bit hackish but way less than the current implementation that highjacks an DOM event payload.

rrahir (changes requested):
the current implementation relies on an old (and now unused) channel to have components interact w/ each others. We should either rely on existing store logic, create a new store, or even declare the menu items of the col/row context menus to be  binded to grid (not sure about the last one, I still think stores a a better approach but the bind would ease the state management inside the `Grid` component.

src/components/grid/grid.ts
dhrp-odoo: Okay, I explored suggested alternatives:

I looked into `CellPopoverStore`, but it is tied to cell positions and computes its anchor from a cell rectangle. Supporting headers would mean faking a cell position or extending cell-specific logic, which feels awkward.

Binding the action to `Grid` would simplify the state, but the row/column actions come from global registries. We would need to build this action separately for each menu opening, making it different from the other registry actions.

A small dedicated store seems cleaner: `Grid` provides the clicked header, the action opens the editor through the store, and `Grid` keeps responsibility for positioning it. WDYT?

rrahir:
The cleaner approach is probably better even if it means more state
management

Le lun. 10 août 2026, 09:33, Dhrutik Patel ***@***.***> a
écrit :

src/components/header_resize_editor/header_resize_editor.ts
dhrp-odoo: Should we enforce the same limit in the core resize logic, including autoresize? (GSheets does the same). 
I’ll define it as a shared constant as well :+1:

src/components/grid/grid.ts
dhrp-odoo: Alright, will proceed with the dedicated store approach then. thanks!

src/components/header_resize_editor/header_resize_editor.ts
rrahir: I think we can add the limit everywhere, yes :)

src/components/header_resize_editor/header_resize_editor.ts
dhrp-odoo: I actually removed placeholder, after thinking bit about it, since in column, we set `96` and in rows `Auto/Automatic` which seems not correct. If you think we can add small label below input when empty.

src/components/grid/grid.ts
dhrp-odoo: want to set null value, but proxy does not allow that

rrahir:
i'm not really convinced by the implementation. it is too much intertwined between the grid & the header overlay. One element adds context to the store, the other tells it to open, but you need both interactions for it to work and you have pretty much no way to guarantee it works except with good testing, which means that corner cases will be missed, and there is 0 reusability in other situations.

 To me, it looks like we are missing crucial information to pass through if we want to perform actions that require a position without relying on the actual active position defined in the selection.
There is a similar issue with the drilldown actions in Odoo. if you select multiple cells that contain an odoo formula, right-click on another cell than the anchor and try to drilldown, it will still apply to the anchor.

same goes for the 'insert link' shortcut, or the 'insert comment', etc.


I think we should rethink how the context menu works:
- should there be a generic store that is available to all to tell what cell was just clicked? Something similar to `useCellHovered` in  `GridOverlay` but it would be available to all stores/components. allows to use that information of the last clicked cell pretty much everywhere we want. this is basically extract what you store in grid/HeaderResizeStore in a separate structure. To be precise, it should extend the concept to something more than a cell, it should also probably react to action related to a cell (inputContextMenu, all clicks, etc) but you get the idea
- Should the execute of the actions have additional context on which cell was clicked ? (looks like a worse version than the previous suggested)
- Should we simply have the selection be adapted in https://github.com/odoo/o-spreadsheet/blob/master/src/components/headers_overlay/headers_overlay.ts#L341 so that the overall selection is not altered but the anchor is?
then the `resizeCols` spec could rely on activePosition.col to tell HeaderResizeEditor where to open. This would mean that we always trust getActivePosition and adapt it when necessary.  but the good news is that it's probably super easy to implement :shrug:

dhrp-odoo:
@rrahir :wave:, Agreed with your point, I explored both approaches more concretely:

1. Move the selection anchor on right-click

Is changing the active cell on right-click the intended product behavior?

As you said, this is simple when right-clicking inside the current anchor zone using `updateAnchorCell()`. However, it does not handle disjoint selections. Activating another selected zone while preserving all zones would require a **new** selection operation/event 😞 (which I’m not fully confident about)

Do you see another way to reuse the existing selection APIs here?

2. Introduce a generic grid interaction store

I prototyped a `GridInteractionStore` extending `SpreadsheetStore` that tracks the latest grid position. It is updated by grid clicks/context menus, header interactions, and `inputContextMenu`, and cleared when the sheet changes. (LMK if anything's missing)

When “Custom size” is executed, `HeaderResizeEditorStore` snapshots the corresponding row or column index. The main point to be careful about is ensuring every relevant interaction path updates the store.

To me, this approach seems cleaner and more reusable later. What do you think?

(Sorry for the back and forth. Lets properly finalize the approach before I rework the new design)

rrahir:
@dhrp-odoo  :wave: 
 
 Crap, I did not even think about disjoint selected zones :facepalm:  The selectionProcessor approach is probably a dead end ATM as we did not want manage several zones :/
 
 Let's try the GridInteractionStore idea :) I'll run it with the team as well

rrahir:
on second thought, I would really explore the idea to generate specific actions on the spot to add to the contextmenu.
i might mean a new kind of  menuItemRegistry but I think it's way easier to just add a "resize columns" section on the spot when handling the "onContextMenu"; which is already there and has limited entry points (that are already handled)  than to have to maintain the state altogether. I
I will still make a task with the idea to have a "last cell actually clicked by the client regardless of the state of the selection processor" but  it seems overkill right now ...

src/components/grid/grid.ts
```diff
@@ -714,19 +729,99 @@ export class Grid extends Component<SpreadsheetChildEnv> {
...
+      return menuItems;
+    }
+
+    // Build the resize action on demand so it captures the clicked header index used to anchor the editor.
+    const isColumn = type === "COL";
+    menuItems.push(
+      createAction({
+        id: isColumn ? "resize_columns" : "resize_rows",
+        name: isColumn ? _t("Resize column") : _t("Resize row"),
+        sequence: isColumn ? 107 : 87,
```
dhrp-odoo: This looks a bit static. Maybe we could find a specific action in the registry and insert this one after it? Not sure :/

@rrahir I tried this approach, and it seems much better than the previous implementation.

Could you please have a look when you have some time? Thanks :)

rrahir: I think its way better this way indeed. I think we could get rid of the headerIndex in the toggleContextMenu by inferring the col,row from the values of x and y, more like we do in the rightClick handler of gridOverlay. We could then pass col,row all the time and  have a "break" condition when their value is "-1". It feels a bit better than relying on  the combination of headerIndex & contextMenuType to know waht we are targetting exactly, imho.

It also could be worth changing the behaviour of onInputContextMenu so that the helper appears above (or  at least next to) the anchor of the selection and not at the far right of the whole selected zone. Currently, it seems really weird that we show the context menu at one place and then the resize header helper potentially very far on its left

EDIT: maybe have `getCartesianCoordinates` moved to a store or smth

rrahir: the sequence feels a bit static indeed... we should have sections in the menu to position stuff, not those sequence + separator stuff >< I don't have a good idea now, maybe target the unhide action and set the sequence right after it 🤷

dhrp-odoo: > I think we could get rid of the headerIndex in the toggleContextMenu by inferring the col,row from the values of x and y, more like we do in the rightClick handler of gridOverlay.

I actually tried that approach but dropped it because we can open the context menu with Fn + Shift + F10, and in that case we don't have client x/y. But yes, I think we can change onInputContextMenu so it opens near the anchor, and open the editor in the same place.

src/components/grid/grid.ts
```diff
@@ -717,16 +727,111 @@ export class Grid extends Component<SpreadsheetChildEnv> {
     this.toggleContextMenu(type, x, y);
   }
 
+  private getMenuItems(type: ContextMenuType, col: HeaderIndex, row: HeaderIndex): Action[] {
+    const menuItems = registries[type].getMenuItems();
+    const index = type === "COL" ? col : type === "ROW" ? row : -1;
```
rrahir:
```suggestion
    const headerIndex = type === "COL" ? col : type === "ROW" ? row : -1;
```

src/components/grid/grid.ts
```diff
@@ -717,16 +727,111 @@ export class Grid extends Component<SpreadsheetChildEnv> {
...
+    }
+    const { dimension, index } = target;
+    const sheetId = this.env.model.getters.getActiveSheetId();
+    const zoom = this.zoomStore.zoomLevel;
+    const gridRect = this.getGridRect();
+    if (dimension === "COL") {
+      const { start, size } = this.viewStore.viewports.getColDimensionsInViewport(sheetId, index);
+      return {
+        x: gridRect.x + (HEADER_WIDTH + start) * zoom,
+        y: gridRect.y,
```
rrahir: 🔥  why can't we rely on getZoomedRect ?

src/components/grid/grid.ts
```diff
@@ -683,21 +693,21 @@ export class Grid extends Component<SpreadsheetChildEnv> {
...
     this.composerFocusStore.activeComposer.stopEdition();
     if (this.env.model.getters.getActiveCols().has(col)) {
       type = "COL";
     } else if (this.env.model.getters.getActiveRows().has(row)) {
       type = "ROW";
     }
-    const { x, y, width } = this.viewStore.viewports.getVisibleRectWithZoom(
-      this.env.model.getters.getActiveSheetId(),
-      lastZone
+    const { x, y } = this.viewStore.viewports.getVisibleRectWithZoom(
```
rrahir: if the anchor is not visible in the viewport, this will return {x:0, y:0} which is not too good :'(

src/components/grid/grid.ts
```diff
@@ -717,16 +727,111 @@ export class Grid extends Component<SpreadsheetChildEnv> {
...
+                : env.model.dispatch("AUTORESIZE_ROWS", {
+                    sheetId,
+                    rows: [...env.model.getters.getActiveRows()],
+                  });
+            },
+          },
+          {
+            name: _t("Custom size"),
+            execute: () =>
+              this.headerResizeEditorTarget.set({
```
rrahir, line 779: Do you have a reason to use a signal here rather than have a store? In my head, the idea was to de-correlate grid from the HeaderResizeEditor, so its goes a bit against the general idea that I imagines but I don't really know what's best? What do you think ?

src/components/grid/grid.ts
dhrp-odoo: I now use the active position and build a full-height zone for a column selection (or full-width for a row). This keeps the active header position resolvable even when its active cell is offscreen. and this will probably works i guess

src/components/grid/grid.ts
dhrp-odoo: I’ve applied all the other changes, but I’m unsure about this one.

`Grid` owns this transient state, receives all relevant clicks, opens the context menu, and positions/renders the editor. A store would decouple the action from the `Grid` instance, but `Grid` would still consume it.

A store seems useful if we add other entry points later, but for the current context-menu flow, the local signal feels simpler. Am I missing another advantage?

rrahir:
robodoo rebase-ff

rrahir:
robodoo r+

PR #8874, opened 2026-06-05
[FIX] charts: bubble legend hover colors

dhrp-odoo:
## Description:

Bubble charts with multiple colors store per-bubble colors as arrays. The legend hover callbacks reused line chart highlighting, which expects single color values and crashes when resetting highlights.

Use bubble-specific highlighting based on the original hover colors so legend hover works with per-bubble colors and preserves transparency.

Task: [6269018](https://www.odoo.com/odoo/2328/tasks/6269018)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/figures/charts/runtime/chart_highlight.ts
+++ b/src/helpers/figures/charts/runtime/chart_highlight.ts
@@ -69,6 +69,32 @@ export function resetLineChartHighlights(dataSets: ChartDataset<"line" | "radar"
   }
 }
 
+export function highlightBubbleChartItem(item: LegendItem, dataSets: ChartDataset<"line">[]) {
+  const dataset = dataSets[0];
+  const backgroundColors = dataset.hoverBackgroundColor;
+  const borderColors = dataset.hoverBorderColor;
+  if (!Array.isArray(backgroundColors) || !Array.isArray(borderColors)) {
+    return;
+  }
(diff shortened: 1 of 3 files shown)


src/helpers/figures/charts/runtime/chart_highlight.ts
```diff
@@ -69,6 +69,32 @@ export function resetLineChartHighlights(dataSets: ChartDataset<"line" | "radar"
   }
 }
 
+export function highlightBubbleChartItem(item: LegendItem, dataSets: ChartDataset<"line">[]) {
```
rrahir, line 72:
```suggestion
export function highlightBubbleChartItem(item: LegendItem, dataSets: ChartDataset<"bubble">[]) {
```

the typing of chartJs is not helping at all ... It even suggests that we could end up with arrays of colors for any kind of chart ><

I feel a bit confused. we use the line chart type for the bubble chart so everything is typed as line. COuld we think of a intermediate layer that maps our chart type to the chartjs equivalent ? so that we could refer to our own 'dataSetType["bubble"] ' and not have the chartjs choice leak everywhere.

@anhe-odoo

rrahir (approved):
robodoo r+

PR #8888, opened 2026-06-09
[IMP] owl 3: improve props typing

hokolomopo:
## Description

The typing of the function props in owl 3 is very hard to understand while reading the code. The complexity is needed to make is work in JS. But we use TypeScript in o-spreadsheet, so we can re-write them into something more ts-like.

Also changes all the imports of `@owl/types` into our own extension of it. The main difference for function is that in `@owl/types`, functions are typed as any if not specified, while we type them as `() => void`, forcing the dev to explicitly type them if there are any arguments.

Task: [6289832](https://www.odoo.com/odoo/2328/tasks/6289832)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


pro-odoo:
robodoo r+

PR #8889, opened 2026-06-10
[FIX] charts: pie chart shows empty chart for text entries

rmbh-odoo:
## Description:

Current behavior before PR:
- Commit 233051c accidentally changed the condition for counting text values.
  As a result, counting failed for valid text data, causing pie charts with only
  text entries to render empty.

- Fixed the condition for counting text values. Pie charts now correctly
  represent counts of distinct text entries.

Task: [6275525](https://www.odoo.com/odoo/2328/tasks/6275525)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/tests/figures/chart/chart_plugin.test.ts
+++ b/tests/figures/chart/chart_plugin.test.ts
@@ -1429,7 +1429,7 @@ describe("datasource tests", function () {
       "1"
     );
     const { chartJsConfig } = model.getters.getChartRuntime("1") as BarChartRuntime;
-    expect(chartJsConfig.data.datasets[0].data.length).toBe(0);
+    expect(chartJsConfig.data.datasets[0].data.length).toBe(3);
   });
 });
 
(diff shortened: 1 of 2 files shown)


tests/figures/chart/chart_plugin.test.ts
```diff
@@ -1429,7 +1429,7 @@ describe("datasource tests", function () {
       "1"
     );
     const { chartJsConfig } = model.getters.getChartRuntime("1") as BarChartRuntime;
-    expect(chartJsConfig.data.datasets[0].data.length).toBe(0);
+    expect(chartJsConfig.data.datasets[0].data.length).toBe(3);
```
hokolomopo, line 1432: Don't we have a test for a pie chart with count values ? If not we should write one. If we did, why did it pass ?

rmbh-odoo: I couldn’t find any pie chart test case that specifically verifies the count of text values, so I added one.

rrahir:
robodoo r+

PR #8891, opened 2026-06-11
[REF] scrollbars: explicitly add scrollbar end margin

hokolomopo:
## Description

In the grid, we have fake a scrollbar div that spans the entirety of the sheetView height. But we also want a small empty corner bottom-right where the two scrollbar meet. How does it work ? Black magic.

We spawn a tall div inside a smaller div, which creates a vertical scrollbar. But. On chrome, I'm 90% sure that since the div is not wide enough, the browser also spawns a horizontal scrollbar. Which is not visible because the div is too narrow. But since it's there, it takes space and pushes the scrollbar up, leaving a gap. If we change `SCROLLBAR_WIDTH` from 15px to 16px, the gap disappear. Obviously, it doesn't work the same way in firefox 🙃.

With this commit, we explicitly set the scrollbar bottom/right position to leave a gap, and remove the overflow in the axis where we don't want a scrollbar.

Task: [6295886](https://www.odoo.com/odoo/2328/tasks/6295886)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/helpers/dom_helpers.ts
+++ b/src/components/helpers/dom_helpers.ts
@@ -257,13 +257,6 @@ export function downloadFile(dataUrl: string, fileName: string) {
   document.body.removeChild(a);
 }
 
-/**
- * Detects if the current browser is Firefox
- */
-export function isBrowserFirefox() {
-  return /Firefox/i.test(navigator.userAgent);
-}
-
 // Mobile detection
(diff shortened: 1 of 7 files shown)


rrahir:
robodoo r+

PR #8893, opened 2026-06-11
[REV] charts: revert font size conversion

dhrp-odoo:
## Description:

Revert o-spreadsheet #8771 because the complexity is higher than the value it brings.

If we keep the change and want to preserve the same font sizes as before, we would need to reduce all chart default font sizes and update dashboard JSON fontSize values accordingly. Existing customer dashboards would still face the same issue.

It is safer to restore the previous chart rendering behavior.

Task: [6295636](https://www.odoo.com/odoo/2328/tasks/6295636)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


LucasLefevre (approved):
robodoo r+

PR #8916, opened 2026-06-15
Refactor and fix chart show value plugin

hokolomopo:
## Description:

Task: [6123207](https://www.odoo.com/odoo/2328/tasks/6123207)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/components/figures/chart/chartJs/chartjs_show_values_plugin.ts
```diff
@@ -39,307 +60,348 @@ export const chartShowValuesPlugin: Plugin = {
...
 
     switch (options.type) {
       case "pie":
-        drawPieChartValues(chart, options, ctx);
+        drawPieValues(chart, options);
         break;
       case "line":
       case "scatter":
+        drawLineValues(chart, options);
+        break;
```
hokolomopo, line 72: There's a world where instead of this switch, all the chart specific logic is inside `chartjs_show_values.ts` and the callbacks are created when creating the runtime rather than here in the chartJs plugin.

It'd be probably a bit better ? I'll wait for another opinion before moving everything.

src/components/figures/chart/chartJs/chartjs_show_values_plugin.ts
```diff
@@ -1,10 +1,30 @@
...
+}
+
+interface TextStyle {
+  strokeColor?: string;
+  textColor: string;
+}
+
+interface CallbackArgs {
+  dataset: any;
+  chartElement: any;
```
hokolomopo, line 21: typing this probably isn't worth, we keep using "private" properties and having to ts-ignore

src/components/figures/chart/chartJs/chartjs_show_values_plugin.ts
```diff
@@ -1,10 +1,30 @@
-import type { ChartMeta, ChartType, Plugin } from "chart.js";
```
hokolomopo, line 1: Also some of the fix commits maybe should be backported

src/components/figures/chart/chartJs/chartjs_show_values_plugin.ts
rrahir: might be worth adding that as a commentary  then. so the next person that tries to type it knows what they're in for

src/components/figures/chart/chartJs/chartjs_show_values_plugin.ts
rrahir: idon't have a strong opinion on this one. kilometric files are not that good in general I'd say ?

src/components/figures/chart/chartJs/chartjs_show_values_plugin.ts
rrahir: I think that *all* the fix commits should be backported :o

tests/figures/chart/__image_snapshots__/chart-show-values-test-ts-bar-chart-show-value-values-close-to-0-are-not-displayed-over-the-x-axis-1-snap.png
rrahir, line 1: I'd argue that style should be defined based on the background of the chart.
0n this snashot, the values that are not displayed above the filled bar are fugly :/

Is it something we could achieve ?

hokolomopo: Would be really annoyting, would need to make the difference "display on the element"/"display in another element (and find out which one is on top)"/"display over chart background". Maybe later sometime ...

rrahir:
robodoo rebase-ff

rrahir:
robodoo r+

PR #8917, opened 2026-06-15
[IMP] figure container: multi figure resize

matho-odoo:
When resizing a figure, all the selected figure will be resized proportionally.

Task: [6302784](https://www.odoo.com/odoo/2328/tasks/6302784)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


tests/figures/figure_component.test.ts
```diff
@@ -570,6 +570,18 @@ describe("figures", () => {
     }
   );
 
+  test("Resize all selected figures", async () => {
```
pro-odoo: You should add tests with and without aspect ratio

src/components/figures/figure_container/figure_container.ts
```diff
@@ -465,51 +470,75 @@ export class FiguresContainer extends Component<SpreadsheetChildEnv> {
...
       const currentMousePosition = { x: ev.clientX / zoom, y: ev.clientY / zoom };
-      const draggedFigure = dragFigureForResize(
-        figureUI,
+      const keepRatio =
+        isSelectionMultiple || ev.shiftKey
+          ? true
+          : figureRegistry.get(selectedFiguresUI[0].tag).keepRatio || false;
+      const minSize = isSelectionMultiple
+        ? MIN_FIG_SIZE
+        : figureRegistry.get(selectedFiguresUI[0].tag).minFigSize || MIN_FIG_SIZE;
```
pro-odoo: In multi selection, figures can be shrunk below MIN_FIG_SIZE. Exemple, a small figure inside a large bounding box, try to resize it, it can be resized to only a few pixel...
It should respect the same rule as when a figure is resized alone.

src/components/figures/figure_container/figure_container.ts
```diff
@@ -465,51 +470,85 @@ export class FiguresContainer extends Component<SpreadsheetChildEnv> {
...
     const onMouseMove = (ev: MouseEvent) => {
       const currentMousePosition = { x: ev.clientX / zoom, y: ev.clientY / zoom };
-      const draggedFigure = dragFigureForResize(
-        figureUI,
+      const keepRatio =
+        isSelectionMultiple || ev.shiftKey
+          ? true
+          : figureRegistry.get(selectedFiguresUI[0].tag).keepRatio || false;
+
+      let minAggregateSize = MIN_FIG_SIZE;
```
pro-odoo: Useless assignment

src/components/figures/figure_container/figure_container.ts
```diff
@@ -465,51 +470,85 @@ export class FiguresContainer extends Component<SpreadsheetChildEnv> {
...
+        const minWidth = Math.min(...selectedFiguresUI.map((f) => f.width));
+        const minHeight = Math.min(...selectedFiguresUI.map((f) => f.height));
+        const scale = Math.max(MIN_FIG_SIZE / minWidth, MIN_FIG_SIZE / minHeight);
+        minAggregateSize = Math.max(
+          Math.round(initialAggregateRect.width * scale),
+          Math.round(initialAggregateRect.height * scale)
+        );
+      } else {
+        minAggregateSize = figureRegistry.get(selectedFiguresUI[0].tag).minFigSize || MIN_FIG_SIZE;
+      }
```
pro-odoo: This is recomputed each time the mouse move, it's a bit useless. It can be move before the mouse move

pro-odoo (changes requested):
I added a video of a bug in the task pad

src/components/figures/figure_container/figure_container.ts
```diff
@@ -465,51 +470,84 @@ export class FiguresContainer extends Component<SpreadsheetChildEnv> {
...
+            this.env.model.getters.getActiveSheetId(),
+            figureUI
+          );
+        if (dirX) {
+          update.width = figureUI.width;
+        }
+        if (dirY) {
+          update.height = figureUI.height;
+        }
+        this.env.model.dispatch("UPDATE_FIGURE", {
```
rrahir: every dispatch will create a new history step so if you loop for every figure, its means that we will need to undo <# of figures> times to revert the change

probably requires an `UPDATE_FIGURES` local command, there are already other occurences for this situation, look at MOVE_FIGURES, DELETE_FIGURES

And a test ;-)

src/components/figures/figure_container/figure_container.ts
```diff
@@ -465,51 +470,84 @@ export class FiguresContainer extends Component<SpreadsheetChildEnv> {
         this.env.model.getters.getNumberRows(sheetId) - 1
       ).end,
     };
+    const initialAggregateRect = rectUnion(...selectedFiguresUI);
+    let minAggregateWidth: number;
+    let minAggregateHeight: number;
+    if (isSelectionMultiple) {
+      const minWidth = Math.min(...selectedFiguresUI.map((f) => f.width));
+      const minHeight = Math.min(...selectedFiguresUI.map((f) => f.height));
+      minAggregateWidth = Math.round((initialAggregateRect.width * MIN_FIG_SIZE) / minWidth);
```
rrahir: the min_fig_size of an image is 20 px, the one of a chart is 80px. I guess here we should select the max of all the types of figures. 

ATM since you only rely on `MIN_FIG_SIZE (=80)`, it means that a single image can be resized to 20px but if I select 2 of them, I'm limited to a size of 80px

src/components/figures/figure_container/figure_container.ts
```diff
@@ -465,51 +470,84 @@ export class FiguresContainer extends Component<SpreadsheetChildEnv> {
...
+      const minFigSize = figureRegistry.get(selectedFiguresUI[0].tag).minFigSize || MIN_FIG_SIZE;
+      minAggregateWidth = minFigSize;
+      minAggregateHeight = minFigSize;
+    }
 
     const onMouseMove = (ev: MouseEvent) => {
       const currentMousePosition = { x: ev.clientX / zoom, y: ev.clientY / zoom };
-      const draggedFigure = dragFigureForResize(
-        figureUI,
+      const keepRatio =
```
rrahir, line 508: the new "shift" behaviour is not tested

src/components/figures/figure_container/figure_container.ts
```diff
@@ -465,51 +470,84 @@ export class FiguresContainer extends Component<SpreadsheetChildEnv> {
...
         maxDimensions
       );
 
-      const otherFigures = this.getOtherFigures([figureUI.id]);
-      const snapResult = snapForResize(this.env, dirX, dirY, draggedFigure, otherFigures);
-      this.dnd.draggedFigure = snapResult.snappedFigures[0];
-      this.dnd.selectedFigures = [this.dnd.draggedFigure];
+      const otherFigures = this.getOtherFigures(selectedFiguresIds);
+
+      // snapForResize mutates resizedAggregateRect in-place
```
rrahir: really not an ideal way to work. I realize it was already present before but since it is an improvement, I think we could/should change the `snapForResize` behaviour so that it copies the rect.

I even wonder why we call the 2 functions back 2 back, dragFigureForResize could probably call snapForResize

rrahir (changes requested):

src/components/figures/figure_container/figure_container.ts
```diff
@@ -439,25 +438,30 @@ export class FiguresContainer extends Component<SpreadsheetChildEnv> {
...
+  resizeAllSelectedFigures(dirX: ResizeDirection, dirY: ResizeDirection, ev: MouseEvent) {
     ev.stopPropagation();
     const initialScrollPosition = this.viewStore.activeSheetScrollInfo;
-
-    const keepRatio = figureRegistry.get(figureUI.tag).keepRatio || false;
-    const minFigSize = figureRegistry.get(figureUI.tag).minFigSize || MIN_FIG_SIZE;
-    const zoom = this.zoomStore.zoomLevel;
     const sheetId = this.env.model.getters.getActiveSheetId();
+    const selectedFiguresIds = this.env.model.getters.getSelectedFigureIds();
+    const isSelectionMultiple = selectedFiguresIds.length > 1;
```
hokolomopo:
```suggestion
    const mutlipleFiguresSelected = selectedFiguresIds.length > 1;
```
nitpick

src/plugins/ui_stateful/figure.ts
```diff
@@ -41,6 +42,7 @@ export class FigureUIPlugin extends UIPlugin {
   handle(cmd: Command) {
     switch (cmd.type) {
       case "MOVE_FIGURES":
+      case "UPDATE_FIGURES":
```
hokolomopo: I'd try to remove `MOVE_FIGURES`, and handle everything in `UPDATE_FIGURES` which is more generic sounding, and could handle both moving & resizing

src/components/helpers/figure_drag_helper.ts
```diff
@@ -25,27 +27,30 @@ export function dragFigureForMove(
...
+  env: SpreadsheetChildEnv,
+  initialRect: Rect,
   dirX: -1 | 0 | 1,
   dirY: -1 | 0 | 1,
   { x: mouseX, y: mouseY }: PixelPosition,
   { x: mouseInitialX, y: mouseInitialY }: PixelPosition,
   keepRatio: boolean,
-  minFigSize: number,
+  minFigWidth: number,
+  minFigHeight: number,
```
hokolomopo:
```suggestion
  minFigSize: DOMDimension,
```
nitpick

src/components/helpers/figure_drag_helper.ts
```diff
@@ -90,5 +95,5 @@ export function dragFigureForResize(
     height = maxY - y;
   }
 
-  return { ...initialFigure, x, y, width, height };
+  return snapForResize(env, dirX, dirY, { x, y, width, height }, otherFigures);
```
hokolomopo: The initial intention was for snap and move/resize to be separate (we would first resize then snap). 

Any reason you need to change that now ?

src/components/helpers/figure_snap_helper.ts
```diff
@@ -99,49 +102,50 @@ export function snapForResize(
   env: SpreadsheetChildEnv,
   resizeDirX: -1 | 0 | 1,
   resizeDirY: -1 | 0 | 1,
-  figureToSnap: FigureUI,
+  figureToSnap: Rect,
   otherFigures: FigureUI[]
```
hokolomopo: I'd try to be consistent between snapForMove/snapForReisze. Either they all take a single rect, and a list of otherRects, or they both take an argument a list of figures and a list of other figures.

hokolomopo (commented):
👋 

Also added some testing notes in the pad :)

src/components/helpers/figure_drag_helper.ts
rrahir: they are both always called together and the function snapForResize changes the figure in place, which looked magical

hokolomopo:
@robodoo r+

PR #8933, opened 2026-06-17
[locale] Issue with decimal separator

matho-odoo:
## Description:

Task: [6232818](https://www.odoo.com/odoo/2328/tasks/6232818)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/functions/module_text.ts
+++ b/src/functions/module_text.ts
@@ -3,7 +3,15 @@ import { _t } from "../translation";
 import { AddFunctionDescription, Arg, FunctionResultObject, Matrix, Maybe } from "../types";
 import { CellErrorType } from "../types/errors";
 import { arg } from "./arguments";
-import { assert, reduceAny, toBoolean, toNumber, toString, transposeMatrix } from "./helpers";
+import {
+  assert,
+  reduceAny,
+  toBoolean,
+  toLocaleString,
+  toNumber,
+  toString,
+  transposeMatrix,
+} from "./helpers";
 
 const DEFAULT_STARTING_AT = 1;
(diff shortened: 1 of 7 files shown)


src/functions/module_text.ts
```diff
@@ -3,7 +3,15 @@ import { _t } from "../translation";
 import { AddFunctionDescription, Arg, FunctionResultObject, Matrix, Maybe } from "../types";
```
anhe-odoo, line 3: I'm not sure but I think TEXTJOIN should also be updated (only the last `toString`). What do you think ? 

Image

anhe-odoo (changes requested):

src/functions/module_text.ts
hokolomopo: We have the same issue here for the delimiter. In fact, we have the same issue in a lot of places (PROPER, SPLIT, most of the function of module_text, even stuff like HYPERLINK). 

Maybe we should do the opposite approach. Always using toLocaleString instead of toString, except when we explicitly don't want something localized (eg. for pivotId or DECIMAL). What do you think ?

matho-odoo: okay :) I didn't added a test for each function, I don't think it's necessary ?

hokolomopo: I'd add a small test for all the relevant functions, using toLocaleString vs toString is a very deliberate choice now, it should be tested

rrahir:
good  luck with the potential conflicts :muscle: 

robodoo r+

PR #8961, opened 2026-06-19
[FIX] chart: zoomable chart height issue with rjsmin minification

rmbh-odoo:
Current behavior before PR:
- In Odoo, when `debug=assets` is not enabled, the spreadsheet bundle is served in a minified form using rjsmin.
- 'rjsmin' does not preserve spaces inside nested template literals.
- When using nested backticks, CSS values like: `calc(100% - 10px)` become `calc(100%-10px)` after minification.
- This breaks CSS expectations and can lead to incorrect rendering.
- As a result, the zoomable chart fails to compute and apply the correct height.

- Replace nested template literal CSS generation with`cssPropertiesToCss`.
- This avoids using nested template literals, preventing incorrect space removal during minification.
- Fixes incorrect height calculation in zoomable charts in production mode.

Task: [6306092](https://www.odoo.com/odoo/2328/tasks/6306092)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs.ts
+++ b/src/components/figures/chart/chartJs/zoomable_chart/zoomable_chartjs.ts
@@ -8,7 +8,7 @@ import {
 } from "../../../../../helpers/figures/charts/chart_common";
 import { Store, useStore } from "../../../../../store_engine";
 import { ChartJSRuntime } from "../../../../../types";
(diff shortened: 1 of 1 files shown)


rrahir:
good catch !

pro-odoo:
robodoo r+

PR #8994, opened 2026-06-23
[FIX] svg: avoid rjsmin issue caused by nested template literals

rmbh-odoo:
## Description:

Current behavior before PR:
- In Odoo, when `debug=assets` is disabled, bundles are minified using `rjsmin`.
- Some SVGs were written as template literals directly inside the `css` templates
  that use them, creating nested backticks.
- `rjsmin` uses regexes (not a real JS parser) and gets confused by these nested backticks,
  breaking the generated CSS.
- As a result, some SVGs were not rendering correctly.

- All raw SVGs are moved to a dedicated file (`raw_svgs.ts`) and exported as constants.
- The templates now reference these constants, so there are no more nested backticks.
- Rolldown keeps them as separate variables instead of inlining them back,
  so the problem does not come back after bundling.
- `rjsmin` can now minify the bundle correctly and all SVGs render properly.

Task: [6317134](https://www.odoo.com/odoo/2328/tasks/6317134)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/composer/top_bar_composer/top_bar_composer.ts
+++ b/src/components/composer/top_bar_composer/top_bar_composer.ts
@@ -9,19 +9,13 @@ import {
 import { Store, useStore } from "../../../store_engine";
 import { CSSProperties, ComposerFocusType, SpreadsheetChildEnv } from "../../../types/index";
 import { css, cssPropertiesToCss } from "../../helpers/css";
+import { FX_SVG } from "../../icons/raw_svgs";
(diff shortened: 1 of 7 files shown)


hokolomopo:
Note for forward port: in 19.1+ we inlined all of those svg directly in the css, but kept the original SVGs strings as comment. We could take this opportunity to remove those comment, they are useless.

@robodoo r+

PR #9012, opened 2026-06-24
[FIX] Fonts: Add default font for Linux

rrahir:
We currently have a font family set for the canvas and the composers. However, both fonts are proprietary:
- Roboto -> Google
- Arial -> Microsoft

For users using an 'open' operating system without any of those proprietary fonts, the font family has 0 matches and defaults on a 'Times new Roman' like font which is quite ugly.

This revision adds a fallback to 'Liberation Sans' which is the opensource equivalent to Arial [1].

[1]: https://wiki.archlinux.org/title/Metric-compatible_fonts

Task: 6328646

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


pro-odoo (commented):
robodoo r+

PR #9016, opened 2026-06-24
[FIX] typing: exclude non-exported symbols in type resolution

rrahir:
Owl's internal prop types use `unique symbol` constants (`hasDefault`, `isOptional`, `typeBrand`, `isProps`) that are declared but not exported. When `PropsOf<SidePanel>` is used as an explicit annotation, TypeScript emits it by name in `.d.ts` without expanding it. But without the annotation, TypeScript infers and expands the full type, which references those unexportable symbols.

This causes an issue in the component `SidePanels` where the getter `panelList` is typed implicitely, which has no effect at ts compilation time but will cause the aforementioned issue when generetin a types file.

Task: 0

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/types/props_of.ts
+++ b/src/types/props_of.ts
@@ -1,5 +1,9 @@
 import { GetPropsWithOptionals } from "@odoo/owl";
 
+// Strips symbol-keyed properties so the expanded type doesn't reference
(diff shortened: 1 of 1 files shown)


pro-odoo:
robodoo r+

PR #9026, opened 2026-06-24
[FIX] conditional_formatting: handle zero color scale midpoint

dhrp-odoo:
## Description:

A color scale midpoint value of 0 was treated as absent because the evaluation logic checked the parsed midpoint with truthiness.

As a result, three-point color scales with a midpoint at 0 were evaluated as two-point scales, interpolating directly from minimum to maximum color. This made the midpoint cell use a mixed color instead of the configured midpoint color.

Task: [6312961](https://www.odoo.com/odoo/2328/tasks/6312961)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


rrahir:
robodoo r+

PR #9027, opened 2026-06-24
[IMP] tools: run esm version is node

rrahir:
The script `run_in_node` does not work with owl3 because they changed the default type of the lib to 'module', which defines how nodeJs should interprete `.js` files [1]. The script relied on the common js version which unfortunately ends with the suffix `.js` and not `.cjs`. Therefore, we nodeJs tries to interprete the commonJs file as a module, which does not work quite well.

Since the lib is mainly used (i.e. in Odoo) with the module js version and that it seems to become the norm, we adapt the script to run as moduleJs. We could consider dropping the common js version in the future but one thing at a time...

[1]: https://docs.npmjs.com/cli/v11/configuring-npm/package-json#type

Task: 0

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


pro-odoo:
robodoo r+

PR #9036, opened 2026-06-25
[IMP] header_size: make getters consistent

hokolomopo:
## Description

Using the getter `getRowSize()` on an hidden row would return the size of the row, while calling `getHeaderSize()` on the same row would return 0. The different behaviour is confusing and not expected.

This commit make `getHeaderSize()` ignore hidden rows, we have to handle them manually when calling the getter.

Note: One could argue that `getCol/RowSize()` should also return 0 for hidden headers instead. But this crashed with the current plugin architecture on `DUPLICATE_SHEET`. The SheetPlugin dispatches `UPDATE_CELL` on sheet duplication, which is handled by all the plugins before every core plugin has received the `DUPLICATE_SHEET` and added the new sheet in its state. This causes some getters to crash, including `isColHidden()`.

Task: [6332479](https://www.odoo.com/odoo/2328/tasks/6332479)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/internal_viewport.ts
+++ b/src/helpers/internal_viewport.ts
@@ -261,7 +261,9 @@ export class InternalViewport {
     while (start <= end && start !== headers && end !== -1) {
       const mid: HeaderIndex = Math.floor((start + end) / 2);
       const offset = this.getters.getColRowOffset(dimension, startIndex, mid, this.sheetId);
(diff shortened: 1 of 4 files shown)


pro-odoo (commented):
robodoo r+

PR #9037, opened 2026-06-25
[FIX] print: handle hidden headers

hokolomopo:
## Description

To determine the page break, the print store would use the header size without taking into account if the header is hidden or not.

Task: [6332587](https://www.odoo.com/odoo/2328/tasks/6332587)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/spreadsheet_print/spreadsheet_print_store.ts
+++ b/src/components/spreadsheet_print/spreadsheet_print_store.ts
@@ -134,14 +134,16 @@ export class SpreadsheetPrintStore extends SpreadsheetStore {
 
     const start = dimension === "COL" ? printZone.left : printZone.top;
     const end = dimension === "COL" ? printZone.right : printZone.bottom;
-    const getHeaderSize =
-      dimension === "COL" ? this.getters.getColDimensions : this.getters.getRowDimensions;
+    const getHeaderSize = (index: HeaderIndex) =>
+      this.getters.isHeaderHidden(sheetId, dimension, index)
+        ? 0
+        : this.getters.getHeaderDimensions(sheetId, dimension, index).size;
     const max = dimension === "COL" ? printWidth : printHeight;
 
     const breaks: number[] = [];
     let current = 0;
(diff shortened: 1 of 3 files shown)


src/components/spreadsheet_print/spreadsheet_print_store.ts
```diff
@@ -134,14 +134,16 @@ export class SpreadsheetPrintStore extends SpreadsheetStore {
...
+      this.getters.isHeaderHidden(sheetId, dimension, index)
+        ? 0
+        : this.getters.getHeaderDimensions(sheetId, dimension, index).size;
     const max = dimension === "COL" ? printWidth : printHeight;
 
     const breaks: number[] = [];
     let current = 0;
     for (let i = start; i <= end; i++) {
-      const headerSize = getHeaderSize(sheetId, i).size * zoom;
+      const headerSize = getHeaderSize(i) * zoom;
```
fdamhaut:
```suggestion
      const headerSize = this.getters.getColRowOffset(dimension, i, i + 1, sheetId) * zoom;
```

fdamhaut (changes requested):

src/components/spreadsheet_print/spreadsheet_print_store.ts
hokolomopo: I can just use `getters.getHeaderSize` ...

src/plugins/ui_stateful/header_positions.ts
```diff
@@ -96,6 +101,12 @@ export class HeaderPositionsUIPlugin extends UIPlugin {
     };
   }
 
+  getHeaderDimensions(sheetId: UID, dim: Dimension, index: HeaderIndex): HeaderDimensions {
```
pro-odoo: Useless

pro-odoo (approved):
robodoo delegate+

hokolomopo:
@robodoo r+

PR #9041, opened 2026-06-26
[IMP] Charts: add new chart suggestions

matho-odoo:
## Description:

Task: [6209080](https://www.odoo.com/odoo/2328/tasks/6209080)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/components/side_panel/data_analysis/data_analysis_panel.ts
```diff
@@ -0,0 +1,19 @@
+import { props, proxy, types } from "@odoo/owl";
+import { Component } from "../../../owl3_compatibility_layer";
+import { useLocalStore } from "../../../store_engine/store_hooks";
+import { SpreadsheetChildEnv } from "../../../types/spreadsheet_env";
+import { Store } from "../../../types/store_engine";
+import { DataAnalysisStore } from "./data_analysis_store";
+
+export class DataAnalysisPanel extends Component<SpreadsheetChildEnv> {
+  static template = "o-spreadsheet-DataAnalysisPanel";
+  protected props = props({ onCloseSidePanel: types.function() });
```
pro-odoo: Should use `useProps` instead of `props` now.

pro-odoo: In owl 3, you should only declare props that you use in your component, it's not necessary anymore to declare all props that the component receive.
So, `onCloseSidePanel` can be removed

src/components/side_panel/data_analysis/data_analysis_panel.ts
```diff
@@ -0,0 +1,31 @@
...
+  static components = {
+    Section,
+    ChartSuggestionPreview,
+  };
+
+  store!: Store<DataAnalysisStore>;
+  selectedCol = proxy({ index: 0 });
+  openDescriptionKey = proxy({ value: "" });
```
pro-odoo: The two variables seems unused

src/components/side_panel/data_analysis/data_analysis_panel.xml
```diff
@@ -0,0 +1,31 @@
+<templates>
+  <t t-name="o-spreadsheet-DataAnalysisPanel">
+    <div class="o-data-analysis-panel">
```
pro-odoo:
```suggestion
  <div t-name="o-spreadsheet-DataAnalysisPanel" class="o-data-analysis-panel">
```
Nitpick

src/components/side_panel/data_analysis/data_analysis_store.ts
```diff
@@ -0,0 +1,74 @@
...
+  mutators = [] as const;
+  hasData: boolean = false;
+  chartSuggestions: ChartSuggestion[] = [];
+  private isDirty = false;
+  ranges?: string[];
+
+  constructor(get: Get) {
+    super(get);
+    this.model.selection.observe(this, {
+      handleEvent: () => this.refreshStatistics(),
```
pro-odoo:
```suggestion
      handleEvent: () => this.refreshSuggestions(),
```

src/components/side_panel/data_analysis/data_analysis_store.ts
```diff
@@ -0,0 +1,74 @@
...
+    switch (cmd.type) {
+      case "HIDE_COLUMNS_ROWS":
+      case "UNHIDE_COLUMNS_ROWS":
+      case "GROUP_HEADERS":
+      case "UNGROUP_HEADERS":
+      case "ACTIVATE_SHEET":
+      case "ACTIVATE_NEXT_SHEET":
+      case "ACTIVATE_PREVIOUS_SHEET":
+      case "EVALUATE_CELLS":
+        this.isDirty = true;
```
pro-odoo, line 47:
```suggestion
        this.isDirty = true;
        break;
```

src/components/side_panel/data_analysis/data_analysis_store.ts
```diff
@@ -0,0 +1,74 @@
...
+    }
+  }
+
+  finalize() {
+    if (this.isDirty) {
+      this.isDirty = false;
+      this.refreshStatistics();
```
pro-odoo:
```suggestion
      this.refreshStatistics();
      this.isDirty = false;
```
It's more natural to do the job before saying it's not dirty anymore, let's say we have an error in the job, the state is still dirty

src/components/side_panel/data_analysis/data_analysis_store.ts
```diff
@@ -0,0 +1,74 @@
...
+    }
+  }
+
+  private refreshStatistics() {
+    const getters = this.getters;
```
pro-odoo: It's a bit useless, I prefer to keep this.getters, it's easier to grep...

src/components/side_panel/data_analysis/data_analysis_store.ts
```diff
@@ -0,0 +1,74 @@
...
+    this.ranges = zones.map(zoneToXc);
+
+    this.hasData = zones.some((zone) =>
+      getters
+        .getEvaluatedCellsInZone(sheetId, zone)
+        .some((cell) => cell.type !== CellValueType.empty)
+    );
+
+    const suggestions = this.hasData ? getChartSuggestions(zones, getters) : [];
+    if (!deepEquals(suggestions, this.chartSuggestions)) {
```
pro-odoo: Why not simply re-assign the new computed value ? deepEquals can be costy...

src/components/side_panel/data_analysis/chart_suggestion_preview.ts
```diff
@@ -0,0 +1,185 @@
...
+import { ScorecardChartRuntime } from "../../../types/chart/scorecard_chart";
+import { SpreadsheetChildEnv } from "../../../types/spreadsheet_env";
+import { types } from "../../props_validation";
+
+const PREVIEW_BUBBLE_RADIUS_RATIO = 0.3;
+const MIN_PREVIEW_BUBBLE_RADIUS = 1;
+
+export class ChartSuggestionPreview extends Component<SpreadsheetChildEnv> {
+  static template = "o-spreadsheet-ChartSuggestionPreview";
+  protected props = props({
```
pro-odoo: useProps

src/components/side_panel/data_analysis/chart_suggestion_preview.ts
```diff
@@ -0,0 +1,185 @@
...
+    definition: types.ChartDefinition(),
+    title: types.string(),
+    rationale: types.string(),
+    isRecommended: types.boolean(),
+    onPointerDown: types.function(),
+  });
+
+  private chartCanvasRef = signal<HTMLCanvasElement | null>(null);
```
pro-odoo:
```suggestion
  private chartCanvasRef = signal.ref();
```

src/components/bottom_bar/bottom_bar_statistic/bottom_bar_statistic.ts
```diff
@@ -81,4 +85,15 @@ export class BottomBarStatistic extends Component<SpreadsheetChildEnv> {
...
   }
+
+  get isDataAnalysisOpen(): boolean {
+    return (
+      this.sidePanelStore.mainPanel?.componentTag === "DataAnalysisPanel" &&
+      this.sidePanelStore.isMainPanelOpen
+    );
+  }
+
+  showDataAnalysis() {
```
pro-odoo:
```suggestion
  toggleDataAnalysisPanel() {
```

src/helpers/ui/side_panel_interactive.ts
```diff
@@ -0,0 +1,21 @@
+import {
+  SidePanelComponentProps,
+  SidePanelStore,
+} from "../../components/side_panel/side_panel/side_panel_store";
+import { SpreadsheetChildEnv } from "../../types/spreadsheet_env";
+
+export function togglePinnedSidePanel(
```
pro-odoo: I would move it inside the side panel store

src/helpers/figures/charts/chart_suggestion_engine.ts
```diff
@@ -0,0 +1,1457 @@
+import {
```
pro-odoo: This whole file is pretty big and mix different concepts:
- the "framework"
- chart-specific stuff
- suggestion constants (`SINGLE_NUMBER_COLUMN_SUGGESTIONS`
- ...

I think it would be easier to maintain, read, understand, ... in different files

pro-odoo (changes requested):
ChartSuggestionPreview seems not tested, same for dataAnalysisPanel (only the store is tested). Same for allowDispatch of the two new commands

src/helpers/figures/charts/chart_suggestion_engine.ts
```diff
@@ -0,0 +1,602 @@
...
+// Rules are divided in two categories:
+//  * EXACT_PATTERNS: a fixed-length array of column types (e.g. "number + number" or "categorical + date + percentage").
+//  * EXTENDABLE_PATTERNS: a fixed-length set of leading types followed by one or more columns of a fixed type (e.g. "many-numbers" or "label + multiple numbers").
+// -----------------------------------------------------------------------------
+
+type ColumnTypeSpec = ExtendedColumnType | readonly ExtendedColumnType[];
+
+interface ShapePattern {
+  readonly leadingTypes: readonly ColumnTypeSpec[];
+  //Present only for open-ended rules: every column after leadingTypes must match `type`,
```
pro-odoo: for extendable_patters ?

tests/test_helpers/canvas_component_screenshot.ts
```diff
@@ -0,0 +1,156 @@
...
+  };
+  return () => {
+    NodeCanvasRenderingContext2D.prototype.fillText = original;
+  };
+}
+
+/**
+ * Mount a component containing a `<canvas>` and take a real screenshot of what it renders.
+ *
+ * jsdom has no canvas rasterizer and no layout engine: `HTMLCanvasElement` is patched by default
```
hokolomopo: rasterizer ? rasta ? No clue what it means, is my english bad ? 😅 

Image

hokolomopo: Ok it's just the most AI-written docstring I ever read

src/helpers/data_analysis.ts
```diff
@@ -0,0 +1,124 @@
...
+    rest.some((c) => c.type !== CellValueType.text && c.type !== CellValueType.empty);
+
+  const dataCells: EvaluatedCell[] = hasHeader ? rest : cells;
+  const nonEmpty = dataCells.filter((c) => c.type !== CellValueType.empty);
+
+  const type = computeColumnType(nonEmpty);
+
+  const numVals = nonEmpty
+    .filter((c) => c.type === CellValueType.number)
+    .map((c) => c.value as number);
```
pro-odoo:
```ts
.map((c) => v.value);
``` 
`as` can hide a future issue

src/helpers/data_analysis.ts
```diff
@@ -0,0 +1,124 @@
...
+
+  // Header: first cell is text AND rest has at least one non-text, non-empty cell
+  const hasHeader =
+    firstCell.type === CellValueType.text &&
+    rest.some((c) => c.type !== CellValueType.text && c.type !== CellValueType.empty);
+
+  const dataCells: EvaluatedCell[] = hasHeader ? rest : cells;
+  const nonEmpty = dataCells.filter((c) => c.type !== CellValueType.empty);
+
+  const type = computeColumnType(nonEmpty);
```
pro-odoo: Move it directly in the `return`:
```ts
return { 
    type: computeColumnType(nonEmpty)
    ...
```
It's easier to read, one less variable name to find...

src/helpers/data_analysis.ts
```diff
@@ -0,0 +1,124 @@
...
+
+  const numVals = nonEmpty
```
pro-odoo:
```ts
const numericValues = ...
```

src/helpers/data_analysis.ts
```diff
@@ -0,0 +1,124 @@
...
+    .filter((c) => c.type === CellValueType.number)
+    .map((c) => c.value as number);
+
+  const allVals = nonEmpty.map((c) => String(c.value ?? ""));
+  const uniqueCount = new Set(allVals).size;
+
+  return {
+    zone,
+    type,
+    header: hasHeader ? String(firstCell.value ?? "") : undefined,
```
pro-odoo: Why casting to `String` ? It's already a string

tests/test_helpers/canvas_component_screenshot.ts
```diff
@@ -0,0 +1,156 @@
...
+ * The sprite-based text renderer patched onto node-canvas (`tests/setup/canvas.mock.ts`, used by
+ * every image-snapshot test for OS-independent font rendering) only positions text correctly for
+ * `textBaseline: "middle"`: any other baseline — including "alphabetic", the real canvas default —
+ * is drawn as if `y` were the *top* of the glyph. Real canvas draws "alphabetic" text mostly
+ * *above* `y` (only descenders dip below), so text placed close to the bottom of a canvas (e.g. a
+ * scorecard's baseline value) renders fine in a browser but gets clipped here. This corrects that
+ * one case, scoped to this call only — the shared mock (and every other test's snapshots) is left
+ * untouched.
+ */
+function patchAlphabeticTextBaseline(): () => void {
```
hokolomopo: ... you're patching the patch ? just fix the first patch in canvas_mock.ts to handle all baselines

src/helpers/data_analysis.ts
```diff
@@ -0,0 +1,124 @@
...
+    rowCount: nonEmpty.length,
+    uniqueCount,
+    uniqueRatio: allVals.length > 0 ? uniqueCount / allVals.length : 0,
+    maxValue: numVals.length ? Math.max(...numVals) : undefined,
+    nonEmpty,
+  };
+}
+
+function computeColumnType(cells: EvaluatedCell[]): ExtendedColumnType {
+  if (cells.length > 0) {
```
pro-odoo: Instead of:
```ts
if (A) {
   ... 
}
return "empty";
```
do:
```ts
if (!A) {
    return "empty";
}
...
```
It's easier to read, easier to see what the condition imply, makes the code inside the if 1-less intented

tests/test_helpers/canvas_component_screenshot.ts
```diff
@@ -0,0 +1,156 @@
...
+ * process, not just the test). The initial size/draw is already correct from the synchronous
+ * `getBoundingClientRect` above, so this observer only needs to exist without crashing — it never
+ * has to actually report a resize for a single static screenshot.
+ */
+class NoopResizeObserver {
+  constructor(_cb: any) {}
+  observe() {}
+  unobserve() {}
+  disconnect() {}
+}
```
hokolomopo: I don't get this patch at all. Or the docstring is wrong. Why would this patch be any different than the one in `resize_observer.mock.ts` ? The docstring said that it crashed because reading `entries[0].contentRect` but 1) which entries ? and 2) this mock do not mock any entries so I'm really not sure why it helps

src/helpers/data_analysis.ts
```diff
@@ -0,0 +1,124 @@
...
+        }
+        if (c.format?.includes("%")) {
+          return true;
+        }
+        return false;
+      })
+    ) {
+      return "percentage";
+    } else if (cells.every((c) => c.type === CellValueType.number)) {
+      return "number";
```
pro-odoo:
```ts
else if (cells.every((c) => c.type === CellValueType.number) {
    if (cells.every((c) => c.format?.includes("%") {
        return "percentage";
    }
    return "number";
}
```

src/helpers/data_analysis.ts
```diff
@@ -0,0 +1,124 @@
...
+    } else {
+      const textVals = cells
+        .filter((c) => c.type === CellValueType.text)
+        .map((c) => c.value as string);
```
pro-odoo: remove as string

src/helpers/data_analysis.ts
```diff
@@ -0,0 +1,124 @@
...
+
+      if (textVals.length > 0) {
+        const unique = new Set(textVals).size;
+        const ratio = unique / textVals.length;
+        return ratio < 0.75 && unique <= 20 ? "categorical" : "label";
```
pro-odoo: Worth a little comment why we consider 0.75 and 20 as treshold

tests/test_helpers/canvas_component_screenshot.ts
```diff
@@ -0,0 +1,156 @@
...
+    width: size.width,
+    height: size.height,
+    top: 0,
+    left: 0,
+    right: size.width,
+    bottom: size.height,
+    x: 0,
+    y: 0,
+    toJSON: () => {},
+  } as DOMRect);
```
hokolomopo: why not use extendMockGetBoundingClientRect like all the other tests ?

src/helpers/figures/charts/chart_suggestion_engine.ts
```diff
@@ -0,0 +1,602 @@
...
+// A column "shape" (the tuple of each selected column's ExtendedColumnType) is matched
+// against a table of rules to decide which chart suggestions to propose.
+// Rules are divided in two categories:
+//  * EXACT_PATTERNS: a fixed-length array of column types (e.g. "number + number" or "categorical + date + percentage").
+//  * EXTENDABLE_PATTERNS: a fixed-length set of leading types followed by one or more columns of a fixed type (e.g. "many-numbers" or "label + multiple numbers").
+// -----------------------------------------------------------------------------
+
+type ColumnTypeSpec = ExtendedColumnType | readonly ExtendedColumnType[];
+
+interface ShapePattern {
```
pro-odoo: Explain what is a ShapePattern

src/helpers/figures/charts/chart_suggestion_engine.ts
```diff
@@ -0,0 +1,602 @@
...
+  readonly leadingTypes: readonly ColumnTypeSpec[];
+  //Present only for open-ended rules: every column after leadingTypes must match `type`,
+  readonly extension?: { readonly type: ColumnTypeSpec; readonly minSize: number };
+}
+
+function matchesColumnType(spec: ColumnTypeSpec, type: ExtendedColumnType): boolean {
+  return Array.isArray(spec) ? spec.includes(type) : spec === type;
+}
+
+function matchesShape(pattern: ShapePattern, shape: ExtendedColumnType[]): boolean {
```
pro-odoo: Mmm, object is named ShapePattern for argument `pattern`, and we have an argument named `shape` => not really clear

tests/test_helpers/canvas_component_screenshot.ts
```diff
@@ -0,0 +1,156 @@
...
+ *
+ * All of this is scoped to the single call: every patched global is restored before returning.
+ */
+export async function mountAndScreenshotCanvas<Props extends { [key: string]: any }>(
+  component: ComponentConstructor<any>,
+  props: Props,
+  model: Model,
+  size: { width: number; height: number } = DEFAULT_CANVAS_SIZE
+): Promise<Buffer> {
+  const nodeCanvasMap = new WeakMap<HTMLCanvasElement, ReturnType<typeof createCanvas>>();
```
hokolomopo: doesn't make sense that the helper maintains a map of canvas, so it's supposed to be able to handle multiple canvas, but then does `const canvas = fixture.querySelector("canvas");` at the end and only screeenshot a single random canvas. 

"Mount a componennt & screenshot a random canvas" also don't make much sense conceptually, should have 2 helpers (mount & screenshot ? Or just patch, normal mount, screenshot)

src/helpers/figures/charts/chart_suggestion_engine.ts
```diff
@@ -0,0 +1,602 @@
...
+  }
+  if (extension) {
+    for (let i = leadingTypes.length; i < shape.length; i++) {
+      if (!matchesColumnType(extension.type, shape[i])) {
+        return false;
+      }
+    }
+  }
+  return true;
+}
```
pro-odoo, line 119: This function is not really clear, the cyclomatic complexity is quite high :/ 
What do you think about 
```ts
  const { leadingTypes, extension } = pattern;
  const expectedLength = leadingTypes.length + (extension?.minSize ?? 0);
  if (extension ? shape.length < expectedLength : shape.length !== expectedLength) {
    return false;
  }
  return shape.every((type, i) => {
    const spec = i < leadingTypes.length ? leadingTypes[i] : extension?.type;
    return spec !== undefined && matchesColumnType(spec, type);
  });
```

src/helpers/figures/charts/chart_suggestion_engine.ts
```diff
@@ -0,0 +1,602 @@
...
+  suggestions: Suggestion<T>[];
+}
+
+function isPyramidLike(numCol1: ColumnAnalysis, numCol2: ColumnAnalysis): boolean {
+  const h1 = (numCol1.header ?? "").toLowerCase();
+  const h2 = (numCol2.header ?? "").toLowerCase();
+  return PYRAMID_HEADER_KEYWORDS.some((k) => h1.includes(k) || h2.includes(k));
+}
+
+function interestingCellsXc(col: ColumnAnalysis): {
```
pro-odoo, line 134: Worth some doc on what's an interesting cell

tests/test_helpers/canvas_component_screenshot.ts
```diff
@@ -0,0 +1,156 @@
...
+      throw new Error("mountAndScreenshotCanvas: the canvas was never drawn to");
+    }
+    return nodeCanvas.toBuffer("image/png");
+  } finally {
+    getContextSpy.mockRestore();
+    rectSpy.mockRestore();
+    window.ResizeObserver = originalResizeObserver;
+    globalThis.OffscreenCanvas = originalOffscreenCanvas;
+    restoreFillText();
+  }
```
hokolomopo: registerCleanup instead of finally stuff

pro-odoo (changes requested):
I didn't read all the patterns, but the feature is cool, nice work :)

Don't forget to add both of you as co-author of the commit.

tests/test_helpers/canvas_component_screenshot.ts
```diff
@@ -0,0 +1,156 @@
+import { CanvasRenderingContext2D as NodeCanvasRenderingContext2D, createCanvas } from "canvas";
+import { Model } from "../../src";
+import { ComponentConstructor } from "../../src/owl3_compatibility_layer";
+import { mountComponent } from "./helpers";
+
+const DEFAULT_CANVAS_SIZE = { width: 192, height: 120 };
+
+/**
```
hokolomopo: Remove all AI generated docstring, they are WAY too long. Write some yourself if need

hokolomopo (commented):
👋 

I just read `canvas_component_screenshot.ts`. It feels very much like an AI file, it's functional but probably can be improved a lot

It's probably harder with chartJs stuff, but I did a very simple mock canvas screenshot for print https://github.com/odoo/o-spreadsheet/pull/9294/changes#diff-6c59166a1148ed1ac95f699ee3bf81875ee577d54b48670c9f80d1164cd333b5

src/components/helpers/chart_drag_and_drop.ts
```diff
@@ -0,0 +1,195 @@
...
+/**
+ * Start dragging a floating preview of the given chart definition, following the mouse.
+ * On drop, creates the chart on the grid (or inside a carousel if dropped on one).
+ */
+export function startChartDragAndDrop(
+  env: SpreadsheetChildEnv,
+  definition: ChartDefinition,
+  ev: MouseEvent
+) {
+  const gridOverlay = document.querySelector(".o-grid-overlay") as HTMLElement | null;
```
hokolomopo: is now wrong, we can have multiple grid overlay (check gridOverlayPosition)

src/components/helpers/chart_drag_and_drop.ts
```diff
@@ -0,0 +1,195 @@
...
+  const spreadsheet = document.querySelector(".o-spreadsheet") as HTMLElement | null;
+  if (!gridOverlay || !spreadsheet) {
+    return;
+  }
+
+  const zoom = env.getStore(ZoomStore).zoomLevel;
+  const startX = ev.clientX / zoom;
+  const startY = ev.clientY / zoom;
```
hokolomopo: use withZoom no ?

src/components/helpers/chart_drag_and_drop.ts
```diff
@@ -0,0 +1,195 @@
...
+import { ScorecardChartRuntime } from "../../types/chart/scorecard_chart";
+import { FigureSize } from "../../types/figure";
+import { PixelPosition } from "../../types/misc";
+import { SpreadsheetChildEnv } from "../../types/spreadsheet_env";
+import { getCarouselOverlappingChart } from "../figures/figure_container/figure_container";
+import { startDnd } from "./drag_and_drop";
+
+const DRAG_MOVE_THRESHOLD = 5;
+
+function getChartFigureSize(type: ChartDefinition["type"]): FigureSize {
```
hokolomopo: nitpick
```suggestion
function getDefaultChartFigureSize(type: ChartDefinition["type"]): FigureSize {
```

src/components/helpers/chart_drag_and_drop.ts
```diff
@@ -0,0 +1,195 @@
...
+  let chart: Chart | undefined;
+  switch (definition.type) {
+    case "scorecard": {
+      canvas.style.width = `${figureWidth}px`;
+      canvas.style.height = `${figureHeight}px`;
+      const design = getScorecardConfiguration({ width, height }, runtime as ScorecardChartRuntime);
+      drawScoreChart(design, canvas, zoom);
+      break;
+    }
+    case "gauge": {
```
hokolomopo: can you use drawChartOnCanvas ?

src/components/helpers/chart_drag_and_drop.ts
```diff
@@ -0,0 +1,195 @@
...
+  }
+
+  let highlightedCarouselEl: Element | null = null;
+  const highlightCarousel = (figureId: string | undefined) => {
+    highlightedCarouselEl?.classList.remove("o-add-to-carousel");
+    highlightedCarouselEl = figureId
+      ? document.querySelector(`[data-id="${figureId}"].o-figure`)
+      : null;
+    highlightedCarouselEl?.classList.add("o-add-to-carousel");
+  };
```
hokolomopo: Let's avoid direct DOM handling when possible, maybe create a dedicated store ?

src/components/helpers/chart_drag_and_drop.ts
```diff
@@ -0,0 +1,195 @@
...
+
+  /** Grid coordinates (in sheet pixels) of a mouse position, or undefined if outside the grid. */
+  const getGridPosition = (clientX: number, clientY: number): PixelPosition | undefined => {
```
hokolomopo, line 99: why can'tyou use getColIndex/getRowIndex ? seems strange to have to do it manually

src/components/helpers/chart_drag_and_drop.ts
```diff
@@ -0,0 +1,195 @@
...
+import { UuidGenerator } from "../../helpers/uuid";
+import { ViewportsStore } from "../../stores/viewports_store";
+import { ZoomStore } from "../../stores/zoom_store";
+import { ChartDefinition } from "../../types/chart/chart";
+import { GaugeChartRuntime } from "../../types/chart/gauge_chart";
+import { ScorecardChartRuntime } from "../../types/chart/scorecard_chart";
+import { FigureSize } from "../../types/figure";
+import { PixelPosition } from "../../types/misc";
+import { SpreadsheetChildEnv } from "../../types/spreadsheet_env";
+import { getCarouselOverlappingChart } from "../figures/figure_container/figure_container";
```
hokolomopo: a helper file should not import a component file, it's an easy way tog et circular imports. Move getCarouselOverlappingChart to a helper file

src/components/helpers/chart_drag_and_drop.ts
```diff
@@ -0,0 +1,195 @@
+import type { Chart } from "chart.js/auto";
```
hokolomopo: maybe add the menu item in the "data" topbar menu next to col statistics

hokolomopo: statistic should probably ignore empty rows ? 

Image

src/components/helpers/chart_drag_and_drop.ts
```diff
@@ -0,0 +1,195 @@
...
+  if (!gridOverlay || !spreadsheet) {
+    return;
+  }
+
+  const zoom = env.getStore(ZoomStore).zoomLevel;
+  const startX = ev.clientX / zoom;
+  const startY = ev.clientY / zoom;
+  const { width, height } = getChartFigureSize(definition.type);
+  const figureWidth = width * zoom;
+  const figureHeight = height * zoom;
```
hokolomopo, line 86: not sure about that, trying to drag a suggestiuon on zoom *2 is ... interesting. The figure takes the whole screen size, and is resized on drop

src/components/helpers/chart_drag_and_drop.ts
```diff
@@ -0,0 +1,195 @@
...
+  );
+
+  const container = document.createElement("div");
+  container.className = "o-chart-drag-preview os-theme-dependant";
+  container.style.display = "none";
+  container.style.width = `${figureWidth}px`;
+  container.style.height = `${figureHeight}px`;
+  const canvas = document.createElement("canvas");
+  container.appendChild(canvas);
+  spreadsheet.appendChild(container);
```
hokolomopo: create the contaienr after the threshold is reached instead of always creating a container and a chart, possibly never displaying it

src/components/helpers/chart_drag_and_drop.ts
```diff
@@ -0,0 +1,195 @@
...
+import { ChartDefinition } from "../../types/chart/chart";
+import { GaugeChartRuntime } from "../../types/chart/gauge_chart";
+import { ScorecardChartRuntime } from "../../types/chart/scorecard_chart";
+import { FigureSize } from "../../types/figure";
+import { PixelPosition } from "../../types/misc";
+import { SpreadsheetChildEnv } from "../../types/spreadsheet_env";
+import { getCarouselOverlappingChart } from "../figures/figure_container/figure_container";
+import { startDnd } from "./drag_and_drop";
+
+const DRAG_MOVE_THRESHOLD = 5;
```
hokolomopo: we already have a DRAG_THRESHOLD constant

src/helpers/figures/charts/chart_ui_common.ts
```diff
@@ -30,6 +31,59 @@ export const CHART_COMMON_OPTIONS = {
   events: ["mousemove", "mouseout", "click", "touchstart", "touchmove", "mouseup"],
 } satisfies ChartOptions;
 
+/**
+ * Sample down a Chart.js configuration's data points to at most `maxPoints`, evenly spaced.
+ * Used for lightweight previews (drag preview, chart suggestions) where rendering every point
+ * of a large dataset would make Chart.js slow without any visible benefit on a small canvas.
+ */
+export function limitChartConfigDataPoints(
```
hokolomopo, line 39: Does that really works ? 

Image

src/actions/edit_actions.ts
```diff
@@ -73,6 +74,16 @@ export const pasteSpecialFormat: ActionSpec = {
   execute: ACTIONS.PASTE_FORMAT_ACTION,
 };
 
+export const dataAnalysis: ActionSpec = {
+  name: _t("Data analysis"),
```
hokolomopo: wrong file ? why edit_action.ts ?

src/components/bottom_bar/bottom_bar_statistic/bottom_bar_statistic.css
```diff
@@ -1,10 +1,30 @@
...
+      padding: 4px;
+    }
+    .o-data-analysis-button {
+      cursor: pointer;
+      border-radius: 3px;
+      &:hover {
+        background-color: var(--os-background-gray-color-hover);
+      }
+      &.active {
+        background-color: var(--os-button-active-bg);
```
hokolomopo, line 17: I cannot see the background color TBH

src/components/bottom_bar/bottom_bar_statistic/bottom_bar_statistic.css
```diff
@@ -1,10 +1,30 @@
...
+      }
+    }
+    .o-icon {
+      width: 30px;
+      height: 30px;
+      padding: 4px;
+    }
+    .o-data-analysis-button {
+      cursor: pointer;
+      border-radius: 3px;
```
hokolomopo: can use bootstrap rounded

src/components/bottom_bar/bottom_bar_statistic/bottom_bar_statistic.ts
```diff
@@ -81,4 +84,15 @@ export class BottomBarStatistic extends Component<SpreadsheetChildEnv> {
...
+  get isDataAnalysisOpen(): boolean {
+    return (
+      this.sidePanelStore.mainPanel?.componentTag === "DataAnalysisPanel" &&
+      this.sidePanelStore.isMainPanelOpen
+    );
+  }
+
+  toggleDataAnalysisPanel() {
+    this.sidePanelStore.togglePinnedSidePanel("DataAnalysisPanel");
+  }
```
hokolomopo, line 99: If the side panel is open as a secondary side panel, the button is shown as hovered but not active, and click does nothing

src/components/bottom_bar/bottom_bar_statistic/bottom_bar_statistic.xml
```diff
@@ -1,15 +1,23 @@
...
+          t-on-click="this.listSelectionStatistics">
+          <t t-out="selectedStatistic"/>
+          <span class="ms-2">
+            <t t-call="o-spreadsheet-Icon.CARET_DOWN"/>
+          </span>
+        </div>
+      </Ripple>
+      <span
+        class="o-data-analysis-button me-4"
+        t-att-class="{ 'ms-auto': selectedStatistic === undefined, 'active': this.isDataAnalysisOpen }"
```
hokolomopo: ms-auto is already on the parent, no ?

src/components/figures/figure_container/figure_container.css
```diff
@@ -46,4 +46,14 @@
     pointer-events: none;
     font-weight: bold;
   }
+
+  .o-chart-drag-preview {
```
hokolomopo: probably move this on the side panel css where startChartDragAndDrop is actually called and it make sense

hokolomopo: most of the css can be bootrap also (position-fixed, border, ...).

src/components/helpers/chart_drag_and_drop.ts
```diff
@@ -0,0 +1,195 @@
...
+    const position = getGridPosition(clientX, clientY);
+    if (!position) {
+      return undefined;
+    }
+    const figureUI = { tag: "chart", ...position, width, height };
+    const otherFigures = env.getStore(ViewportsStore).visibleFigures;
+    return getCarouselOverlappingChart(figureUI, otherFigures, ["carousel", "chart"]);
+  };
+
+  const onMouseMove = (e: MouseEvent) => {
```
hokolomopo, line 127: maybe make the cursor stay "grabbing" during the d&d ?

src/components/side_panel/data_analysis/chart_suggestion_preview.css
```diff
@@ -0,0 +1,27 @@
...
+    background: white;
+    position: relative;
+  }
+  .o-suggestion-canvas-wrap:hover {
+    border-color: var(--o-action-color, #3d9be9);
+  }
+
+  .o-suggestion-thumb-title {
+    font-size: 0.7rem;
+    color: var(--os-text-muted, #555);
```
hokolomopo: text-muted class ?

src/components/side_panel/data_analysis/chart_suggestion_preview.css
```diff
@@ -0,0 +1,27 @@
...
+    cursor: grabbing;
+  }
+
+  .o-suggestion-canvas-wrap {
+    height: 120px;
+    border: 1px solid var(--os-border-color);
+    border-radius: 3px;
+    overflow: hidden;
+    background: white;
+    position: relative;
```
hokolomopo: can all be bootstrap. The color is either useless or wrong in dark mode.

src/components/side_panel/data_analysis/chart_suggestion_preview.css
```diff
@@ -0,0 +1,27 @@
+.o-spreadsheet {
+  .o-chart-suggestion-thumb {
+    cursor: grab;
+    min-width: 0;
```
hokolomopo: useful ?

src/components/side_panel/data_analysis/chart_suggestion_preview.ts
```diff
@@ -0,0 +1,184 @@
...
+  setup() {
+    onMounted(() => registerChartJSExtensions());
+    onWillUnmount(() => this.destroyChart());
+    useLayoutEffect(
+      () => {
+        this.updateChart();
+      },
+      () => [this.props.definition]
+    );
+  }
```
hokolomopo, line 38: Just an idea (maybe for a follow up), but should the Chart component take the chart runtime/definition as a props instead of a chart id ? That way we could maybe use a `ChartFigure` here, and not duplicate the logic

src/components/side_panel/data_analysis/data_analysis_panel.css
```diff
@@ -0,0 +1,12 @@
+.o-spreadsheet {
+  .o-chart-suggestions-grid {
+    display: grid;
+    grid-template-columns: 1fr 1fr;
+    gap: 6px;
+  }
+
+  .o-data-analysis-chart {
+    cursor: pointer;
+    border: 1px solid var(--os-border-color);
```
hokolomopo: border class

src/helpers/figures/charts/smart_chart_engine.ts
```diff
@@ -116,8 +109,7 @@ function isDatasetTitled(getters: Getters, column: ColumnInfo): boolean {
 function buildSingleColumnChart(column: ColumnInfo, getters: Getters): ChartDefinition {
   const { type, zone } = column;
   const sheetId = getters.getActiveSheetId();
-  const dataSetsHaveTitle = isDatasetTitled(getters, column);
-  const dataRange = getUnboundRange(getters, zone);
+  const dataSetsHaveTitle = isDatasetTitled(getters, column.zone);
```
hokolomopo, line 112: can the smart chart engine use your data anlysis stuff instead of a spearate logic ?

src/types/commands.ts
```diff
@@ -690,6 +690,26 @@ export interface MergeIntoCarouselCommand extends SheetDependentCommand {
...
+  figureId: UID;
+  offset?: PixelPosition;
+  size?: FigureSize;
+  col?: number;
+  row?: number;
+}
+
+export interface CreateChartAndMergeIntoCarouselCommand extends SheetDependentCommand {
+  type: "CREATE_CHART_AND_MERGE_INTO_CAROUSEL";
+  baseFigureId: UID;
```
hokolomopo: reading only the types, i'd expect both to do the exact same thing

src/constants.ts
```diff
@@ -239,6 +241,10 @@ export const LINE_FILL_TRANSPARENCY = 0.4;
 export const LINE_DATA_POINT_RADIUS = 3;
 export const DEFAULT_WINDOW_SIZE = 2;
 
+// Maximum number of data points rendered by Chart.js in lightweight previews (drag preview,
+// chart suggestions), sampled down from the full dataset for performance.
+export const MAX_CHART_PREVIEW_DATA_POINTS = 150;
```
hokolomopo, line 246: oh that's why it didn't look great. 150 points is a wild limit for a 100px wide canvas

hokolomopo (commented):
Didn't read the test, and didn't read the actual data analysis 🙈

src/components/side_panel/data_analysis/chart_suggestion_preview.ts
anhe-odoo: Agree with you :)

src/types/commands.ts
anhe-odoo: No, `CREATE_CHART_INTO_CAROUSEL` add a chart in an existing carousel while `CREATE_CHART_AND_MERGE_INTO_CAROUSEL` create a chart and merge it with another chart into a carousel containing both

src/constants.ts
anhe-odoo: Not that much, as the same limit is also used for the dragged/dropped chart before creating the real chart so going to 15/20 points will make a huge difference when between the dragged chart and the real chart created after the drop.

src/components/bottom_bar/bottom_bar_statistic/bottom_bar_statistic.css
anhe-odoo: Probably because you're color blind ? It's the same color as the other "active" button, the only difference is the default (not active) background which is not white but kind of gray in the bottom bar

src/helpers/figures/charts/chart_suggestion_engine.ts
anhe-odoo: Changed it to RulePattern with a small comment :)

src/helpers/figures/charts/smart_chart_engine.ts
matho-odoo: The analyze column type doesn't work the same way in smart engine, it counts the cells of each type and returns the one with the bigger count. While we want all the cells to have the same type.

The smart chart engine can be re thinked in another task tough :) because indeed the smart chart inserted is not always the best match

src/components/helpers/chart_drag_and_drop.ts
anhe-odoo: Not really, with the "halfWidth" décalage to have the figure centered on the mouse cursor, we have to make other computation that are less intuitive if we use `withZoom` on the events

src/components/helpers/chart_drag_and_drop.ts
anhe-odoo: Why do you want to use getColIndex here ? ou getGridPosition gives us the position (in pixel) in the grid with the scrolling offset and the zoom

src/types/commands.ts
hokolomopo: Yes I did end up reading the code, my point was that the type names weren't clear in the distinction. Maybe `CREATE_CHART_AND_MERGE_WITH_EXISTING_CHART` or something

src/components/helpers/chart_drag_and_drop.ts
hokolomopo: getColIndex/getRowIndex is what we use in gridOverlay to convert pixel position into col/row position. It take into account frozen panes, scroll & whatnot
image

src/types/commands.ts
anhe-odoo: Yes but the fact that we merge it into a carousel is also important IMO, o we would need something like `CREATE_CHART_AND_MERGE_WITH_EXISTING_CHART_INTO_CAROUSEL` ... I think there is not really a perfect name for this one. We had a talk yesterday with Remi and he also ended up with exactly the name `CREATE_CHART_AND_MERGE_INTO_CAROUSEL`

src/components/helpers/chart_drag_and_drop.ts
anhe-odoo: Yes, I know that ... But here we don't want the column and row index, we want the position (IN PIXEL) of the grid

src/constants.ts
hokolomopo: Yeah but if you go from 1000 points to 150 the difference will be huge as well....  So If we want fidelity we shouldn't have this limit at all.

If the limit is here only for performance and not for readability, how relevant is it ? 1) is the perf hit that large and 2) do we really expect the feature to be used with selection with 150+ points ? I And in that case is the chart drawing the bottleneck ?

src/components/helpers/chart_drag_and_drop.ts
hokolomopo: Oh yeah I'm stupid nevermind

src/constants.ts
anhe-odoo: Yes it's only there for performance, and yes it's really needed. Vincent make some tests on huge database and without this part it's was really slow at first time. Some others improvements has also been added so I can try again (I would personnally prefer to revert these changes and draw all the points, the code will be easier to read).

src/components/helpers/chart_drag_and_drop.ts
```diff
@@ -0,0 +1,216 @@
+import {
```
hokolomopo, line 1: Kinda hard to not have the chart appear on the grid now, even when i'm clearly trying to drop it in the side panel

https://drive.google.com/file/d/1K8gXK6wTJhh0H4ehwt7G6y7lJEje3KMX/view?usp=drive_link

src/components/helpers/chart_drag_and_drop.ts
```diff
@@ -0,0 +1,216 @@
...
+      col,
+      row,
+      offset,
+    };
+    const overlappingFigure = getOverlappingFigure(
+      mouseEvent.clientX - halfWidth,
+      mouseEvent.clientY - halfHeight
+    );
+    if (overlappingFigure?.tag === "carousel") {
+      env.model.dispatch("CREATE_CHART_INTO_CAROUSEL", {
```
hokolomopo: Don't we have the `ADD_NEW_CHART_TO_CAROUSEL` command ? Is this new command necessary ? It's also kinda strange how it create a new figure, then instantly deletes it ...

src/components/side_panel/data_analysis/chart_suggestion_preview.ts
```diff
@@ -0,0 +1,183 @@
...
+            : {}),
+          ...("r" in existingScales
+            ? {
+                r: {
+                  ...(existingScales as any).r,
+                  ticks: { display: false },
+                  pointLabels: { display: false },
+                },
+              }
+            : {}),
```
hokolomopo: nitpick: for readability maybe create a separate function that does that, with real if/else instead of this ternary mess

tests/figures/chart/chart_suggestion_preview.test.ts
```diff
@@ -0,0 +1,120 @@
...
+  labelRange: "A1:A4",
+  dataSetsHaveTitle: false,
+});
+const TWO_SERIES = toChartDataSource({
+  dataSets: [{ dataRange: "B1:B4" }, { dataRange: "C1:C4" }],
+  labelRange: "A1:A4",
+  dataSetsHaveTitle: false,
+});
+
+async function previewSnapshot(definition: ChartDefinition<string>) {
```
hokolomopo: inline the function, it's only used once and having the model creation & the expect in a separate function hurt the readability

tests/test_helpers/canvas_component_screenshot.ts
```diff
@@ -0,0 +1,38 @@
...
+import { mountComponent } from "./helpers";
+import { extendMockGetBoundingClientRect } from "./mock_helpers";
+
+export const DEFAULT_CANVAS_SIZE = { width: 192, height: 120 };
+
+export async function mountAndScreenshotCanvas<Props extends { [key: string]: any }>(
+  component: ComponentConstructor<any>,
+  props: Props,
+  model: Model,
+  size: { width: number; height: number } = DEFAULT_CANVAS_SIZE
```
hokolomopo: 192\*120 is a very strange default size ... I'd make the size a mandatory argument (you never not define it anyway) and inline your 192\*120 size

tests/setup/canvas.mock.ts
```diff
@@ -219,16 +223,36 @@ function drawColoredChar(
...
+function adjustYForBaseline(
+  ctx: NodeCanvasRenderingContext2D,
+  text: string,
+  baseline: CanvasTextBaseline,
+  y: number
+): number {
+  if (baseline === "middle") {
+    return y - fontHeight / 2;
+  }
+  if (baseline && baseline !== "top" && baseline !== "hanging") {
```
hokolomopo: nitpick: put in the condition what baseline it does work with, not what baseline it does **not** work with

tests/test_helpers/canvas_component_screenshot.ts
```diff
@@ -0,0 +1,38 @@
+import { createCanvas } from "canvas";
+import { Model } from "../../src";
+import { ComponentConstructor } from "../../src/owl3_compatibility_layer";
+import { registerCleanup } from "../setup/jest.setup";
+import { mountComponent } from "./helpers";
+import { extendMockGetBoundingClientRect } from "./mock_helpers";
+
+export const DEFAULT_CANVAS_SIZE = { width: 192, height: 120 };
+
+export async function mountAndScreenshotCanvas<Props extends { [key: string]: any }>(
```
hokolomopo: I think I would change the helper into "mockHtmlCanvas" or something, not mount the component, and return the canvas. The test would call it, then mount, then screenshot.

My issue ATM is that we are missing a lot of arguments of `mount` (what if we want to pass a specific env ?), and cannot get back the fixture/env/stores, it's not really generic. Either do something very specific for your use case, and keep it as it but put it in the `chart_suggestion.test.ts` file, or do something a bit more generic.

hokolomopo (commented):
small things, otherwise LGTM

tests/test_helpers/mock_helpers.ts
```diff
@@ -17,7 +28,12 @@ export function mockGetBoundingClientRect() {
     .spyOn(HTMLElement.prototype, "getBoundingClientRect")
     .mockImplementation(function (this: HTMLElement) {
       const mockedClasses = Object.keys(persistentClassesWithMocks);
-      const mockedClass = mockedClasses.find((className) => this.classList.contains(className));
+      const mockedClass = mockedClasses.find(
+        (key) =>
+          key === "*" ||
```
hokolomopo: do you really need the `*` wildcard ? Can't you add a class to your canvas, or use the `tagName` option you just defined ?

hokolomopo:
@robodoo r+

PR #9046, opened 2026-06-29
[FIX] charts: handle undefined values in treemap chart

rmbh-odoo:
## Description:

Current behavior before PR:
- In b559bfd, undefined label values were handled for sunburst charts
but not for treemap charts.
- When using ranges of different sizes (for example A2 and B1),
treemap computations could raise a traceback.

- Handle undefined label values in treemap charts the same way
as in previous versions.
- This prevents the traceback. Label mapping for hierarchical charts is still
incorrect when ranges do not match and will be improved later if needed.

Task: [6329404](https://www.odoo.com/odoo/2328/tasks/6329404)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


rrahir:
robodoo r+

PR #9048, opened 2026-06-30
[FIX] calendar chart: skip missing value tooltips

dhrp-odoo:
## Description:

Calendar chart datasets use `data` as a Chart.js rendering payload, where every bucket is drawn with a constant value. The actual cell values are stored separately in `values`, which is only used by our tooltips and show-values plugin.

Before this commit, missing calendar values were stored as `NaN` in `values`. This is the right sentinel for Chart.js data points, but not for this metadata: the tooltip filter only skips `undefined`, so missing buckets still displayed a tooltip and formatted `NaN` as `1.00e-NaN`.

This commit stores missing calendar metadata as `undefined`, keeping empty buckets out of tooltips/show-values while preserving real numeric values.

Task: [6276800](https://www.odoo.com/odoo/2328/tasks/6276800)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/figures/charts/runtime/chartjs_dataset.ts
+++ b/src/helpers/figures/charts/runtime/chartjs_dataset.ts
@@ -144,7 +144,7 @@ export function getCalendarChartDatasetAndLabels(
       borderWidth: 1,
       barPercentage: 1,
       categoryPercentage: 1,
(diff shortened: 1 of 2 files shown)


hokolomopo:
@robodoo r+

PR #9065, opened 2026-06-30
[IMP] charts: adapt the behavior for boolean column

matho-odoo:
## Description:

A single-column of boolean values is now detected as a categorical dataset, producing an aggregated pie chart with TRUE/FALSE as labels — same behaviour as text columns, where boolean values in the dataset are mapped to 1 so the chart can count occurrences.
If the boolean values are in the data serie with something else as label, we treat them as 0 and 1 and use them as normal numerical values.

## Related Task(s):
- Task: [6344909](https://www.odoo.com/odoo/2328/tasks/6344909)

### DIFF
--- a/src/helpers/figures/charts/chart_data_sources.ts
+++ b/src/helpers/figures/charts/chart_data_sources.ts
@@ -17,12 +17,17 @@ import { CellErrorType } from "../../../types/errors";
 import { Getters } from "../../../types/getters";
 import { FunctionResultObject } from "../../../types/misc";
 import { Range } from "../../../types/range";
-import { isErrorResult, isNumberResult, isTextResult } from "../../cells/cell_evaluation";
+import {
+  isBooleanResult,
+  isErrorResult,
+  isNumberResult,
+  isTextResult,
+} from "../../cells/cell_evaluation";
 import { formatValue } from "../../format/format";
 import { isDefined } from "../../misc";
 import { createValidRange, duplicateRangeInDuplicatedSheet } from "../../range";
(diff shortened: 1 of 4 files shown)


src/helpers/figures/charts/chart_data_sources.ts
```diff
@@ -293,7 +302,18 @@ function getChartDatasetValues(getters: Getters, dataSets: DataSet[]): DatasetVa
...
+    if (!data.some((cell) => isNumberResult(cell)) && data.some((cell) => isBooleanResult(cell))) {
+      // When the labels are the boolean values themselves (categorical chart), each boolean
+      // occurrence is counted as 1 so it can be aggregated per TRUE/FALSE label. Otherwise, the
+      // booleans are used as regular numeric values: TRUE = 1, FALSE = 0.
+      const isCategorical =
+        !labelRange ||
+        (labelRange.sheetId === ds.dataRange.sheetId &&
+          isEqual(labelRange.zone, ds.dataRange.zone));
+      data = data.map((cell) =>
+        !isBooleanResult(cell) ? EMPTY : isCategorical ? ONE : { value: cell.value ? 1 : 0 }
```
fdamhaut:
```suggestion
      data = data.map((cell) => {
        if(!isBooleanResult(cell)) {
          return EMPTY
        } else if (isCategorical) {
          return ONE
        }else {
          return cell.value ? 1 : 0
        }
      }
```
nitpick

pro-odoo:
robodoo r+

PR #9092, opened 2026-07-01
[IMP] charts: allow to create calendar chart with only dates

anhe-odoo:
## Task Description

Just as we do with the pie chart when we only have one column of strings, we should be able to create a calendar chart with only a column of dates by counting the appearance by granularity. This PR aims to add this possibility, needed for the task 6209080

## Related Task

- Task: [6352279](https://www.odoo.com/odoo/2328/tasks/6352279)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


pro-odoo:
robodoo r+

PR #9096, opened 2026-07-02
[FIX] DefaultPlugin: no format should be preserved

rrahir:
When applying a default format to a dimension, we "pin" the format of the pre-existing cells so that it does not get overriden. However, a cell with a content an no format at all (the "General" case) was not handled. For those cells, we need to set an explicit ̀`""` format.

In future work, the format should probably always be defined on a cell, and we could keep the optimisation to not export it.

Task: 6345022

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/core/default.ts
+++ b/src/plugins/core/default.ts
@@ -300,29 +300,26 @@ export class DefaultPlugin extends CorePlugin<defaultState> implements defaultSt
     const defaults: [CellPosition, Format][] = [];
     for (const position of zones.flatMap((zone) => cellPositions(sheetId, zone))) {
       const cellFormat = this.getters.getCell(position)?.format;
-      if (cellFormat) {
+      if (cellFormat !== undefined) {
         continue;
       }
       const rowDefault = this.format[sheetId]?.rowDefault?.[position.row];
-      if (rowDefault) {
+      if (rowDefault !== undefined) {
         if (priorities.shouldUseDefaultRow) {
           defaults.push([position, rowDefault]);
         }
         continue;
       }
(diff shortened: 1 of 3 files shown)


src/plugins/core/default.ts
fdamhaut:
```suggestion
      if (rowDefault !== undefined) {
```

fdamhaut:
```suggestion
      if (colDefault !== undefined) {
```

src/plugins/core/default.ts
```diff
@@ -324,6 +324,17 @@ export class DefaultPlugin extends CorePlugin<defaultState> implements defaultSt
...
+      // If the default level is going to change, we want an explicit empty "" format
+      // to prevent the cell from inheriting the new default
+      if (
+        (priorities.shouldUseDefaultRow ||
+          priorities.shouldUseDefaultCol ||
+          priorities.shouldUseDefaultSheet) &&
+        cell
+      ) {
+        defaults.push([position, ""]);
+      }
```
fdamhaut:
```suggestion
```

src/plugins/core/default.ts
fdamhaut:
```suggestion
      const sheetDefault = this.format[sheetId]?.sheetDefault ?? "";
      if (priorities.shouldUseDefaultSheet) {
        defaults.push([position, sheetDefault]);
      }
```

pro-odoo (approved):
robodoo delegate+

rrahir:
robodoo r+

PR #9147, opened 2026-07-14
[FIX] tables: fix overlapping table

fdamhaut:
Overlapping table should not exists as they, for example,
cause multiple similar key issue with filter in tables.

Task: 6347755

## Description:

Task: [6347755](https://www.odoo.com/odoo/2328/tasks/6347755)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_stateful/selection.ts
+++ b/src/plugins/ui_stateful/selection.ts
@@ -5,6 +5,7 @@ import {
   clip,
   deepCopy,
   isEqual,
+  overlap,
   positionToZone,
   range,
   uniqueZones,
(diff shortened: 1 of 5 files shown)


src/plugins/ui_stateful/selection.ts
```diff
@@ -624,9 +625,75 @@ export class GridSelectionPlugin extends UIPlugin {
     if (!isCol && !this.isTableRowMoveAllowed(id, cmd.elements)) {
       return CommandResult.CannotMoveTableHeader;
     }
+    if (this.willMoveOverlapTables(cmd)) {
+      return CommandResult.WillOverlapTable;
+    }
     return CommandResult.Success;
   }
 
+  private willMoveOverlapTables(cmd: MoveColumnsRowsCommand): boolean {
```
hokolomopo, line 634: I don't quite get this fix:

1) AFAIK the only thing move_columns does is a copy/paste. So can't we have the same problem on actual copy/paste ? Can it really only happen here? 
2) if some combination of commands can make tables overlap, why isn't it fixed in the table plugin itself ?

hokolomopo: Oh 2) is done in master. But why in master ? A sheet with ovelapping tables will just crash, so why not fix it in 18 ?

tests/sheet/selection_plugin.test.ts
```diff
@@ -987,6 +987,57 @@ describe("move elements(s)", () => {
...
+    createTable(model, "B1:D3");
+    expect(moveColumns(model, "C", ["F"])).toBeSuccessfullyDispatched();
+  });
+
+  test("allows moving a whole table to an empty area", () => {
+    const model = new Model();
+    createTable(model, "A1:B3");
+    createTable(model, "D1:E3");
+    expect(moveColumns(model, "H", ["D", "E"], "after")).toBeSuccessfullyDispatched();
+  });
```
hokolomopo, line 1027: tests do'nt test anything, the command is dispatched sure but what happens to the table ?

hokolomopo (commented):
👋

src/plugins/ui_stateful/selection.ts
fdamhaut: 1) You already cannot copy a table on another
2) Didn't want to change existing spreadsheet as the crash only happens when overlapping headers
2bis) We talked about it with VSC but i don't remember why we made this exact functional choice 😕

hokolomopo: Ok for master then. Maybe a small data migration to remove overlapping table as discussed

rrahir: @fdamhaut what was the conclusion  regarding the migration idea ? we keep it as is, right ? otherwise we also need to migrate the commands :(

rrahir: bump

PR #9153, opened 2026-07-15
[FIX] squisher: adapted ranges that become cells

VincentSchippefilt:
Steps to reproduce the error:
from a new spreadsheet
in A1, =sum(B1:C1)
in A2, =sum(B1:D1)
delete column C
export the data (it's already corrupted, you lost) reimport it as a new spreadsheet (everything seems to work) try to re-export it

This fix does 2 things: on the squisher side it never simplifies ranges into cells, on the un-squisher side it still supports when ranges become cells.

Task: 6390727

### DIFF
--- a/src/formulas/compiler.ts
+++ b/src/formulas/compiler.ts
@@ -261,6 +261,23 @@ export class CompiledFormula implements Omit<ICompiledFormula, "tokens" | "depen
       params.execute
     );
   }
+
+  /*
+   * If you create a formula with a range as parameter, then delete all the dimensions of this range except one by removing columns and rows,
+   * it is possible that the formula AST and normalized formula says this range is a |R| kind of parameter while it is actually a |C| kind of parameter.
+   * This is useful because when squishing formulas, the kind of parameter must be correct according to the actual shape of the range to be compacted
+   * */
+  areRangeShapesInLineWithNormalizedFormula() {
+    const matches = Array.from(this.normalizedFormula.matchAll(/\|C\||\|R\|/g));
+    return matches.every((value, index) => {
+      return (
+        (this.rangeDependencies[index].parts.length === 1 && value[0] === "|C|") ||
+        (this.rangeDependencies[index].parts.length === 2 && value[0] === "|R|") ||
+        this.rangeDependencies[index].invalidXc ||
+        this.rangeDependencies[index].invalidSheetName
+      );
+    });
+  }
 }
 
 /**
--- a/tests/model/squish.test.ts
+++ b/tests/model/squish.test.ts
@@ -1,6 +1,7 @@
+import { FormulaCell } from "../../src";
 import { createRangeFromXc } from "../../src/helpers/range";
 import { Model } from "../../src/model";
-import { createSheet } from "../test_helpers";
+import { createSheet, deleteColumns, getCell, getCellContent, getCellText } from "../test_helpers";
 import { createModelFromGrid } from "../test_helpers/helpers";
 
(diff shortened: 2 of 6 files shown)


tests/model/squish.test.ts
```diff
@@ -82,6 +83,87 @@ describe("squish - unsquish", () => {
...
+      ],
+    };
+
+    const model = new Model(squishedData);
+    expect(getCellContent(model, "A1")).toEqual("FALSE");
+    expect(getCellContent(model, "A2")).toEqual("FALSE");
+    expect(getCellContent(model, "A3")).toEqual("FALSE");
+    expect(getCellContent(model, "A4")).toEqual("FALSE");
+    expect(getCellContent(model, "B4")).toEqual("FALSE");
+    expect(getCellContent(model, "B3")).toEqual("FALSE");
```
hokolomopo: Maybe test the formulas instead, the result could be FALSE with a completely wrong formula.

src/formulas/compiler.ts
```diff
@@ -261,6 +261,23 @@ export class CompiledFormula implements Omit<ICompiledFormula, "tokens" | "depen
       params.execute
     );
   }
+
+  /*
+   * If you create a formula with a range as parameter, then delete all the dimensions of this range except one by removing columns and rows,
+   * it is possible that the formula AST and normalized formula says this range is a |R| kind of parameter while it is actually a |C| kind of parameter.
+   * This is useful because when squishing formulae, the kind of parameter must be correct according to the actual shape of the range to be compacted
+   * */
```
hokolomopo:
```suggestion
   * This is useful because when squishing formulas, the kind of parameter must be correct according to the actual shape of the range to be compacted
   * */
```

hokolomopo (commented):
So if I understand correctly the issue is that sometime we have ranges that becomes cells in the formulas ? Should the unsquisher even make a difference between SUM(A3) and SUM(A3:A3) ? Both are the exact same functionally no ?

tests/model/squish.test.ts
```diff
@@ -82,6 +83,119 @@ describe("squish - unsquish", () => {
...
+    const newModel = new Model(squished);
+    expect((getCell(newModel, "A1") as FormulaCell).compiledFormula.normalizedFormula).toEqual(
+      "=isnumber(|R|)"
+      // it was explicitly saved with a range, it stays a range (that is the origin of the bug)
+    );
+    expect((getCell(newModel, "A2") as FormulaCell).compiledFormula.normalizedFormula).toEqual(
+      "=isnumber(|R|)"
+    );
+    expect((getCell(newModel, "A3") as FormulaCell).compiledFormula.normalizedFormula).toEqual(
+      "=isnumber(|C|)" // this updated reference becomes a range
```
fdamhaut:
```suggestion
      "=isnumber(|C|)" // this updated reference becomes a cell
```

VincentSchippefilt:
> So if I understand correctly the issue is that sometime we have ranges that becomes cells in the formulas ? Should the unsquisher even make a difference between SUM(A3) and SUM(A3:A3) ? Both are the exact same functionally no ?

not completely: reference cells get compiled with direct cell access, while ranges get compiled with 2D array, so while building the compiled code from the AST, there is a difference.

That being said, there is also a potential for bugs in the current compilation: a shared compiled formula might have the AST for a cell and receive a range at that position, or be compiled as a range and receive a cell at that position

hokolomopo (commented):
@robodoo r+

PR #9160, opened 2026-07-16
Master insert pivot store flda

fdamhaut:
## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/pivot/pivot_title_section/pivot_title_section.ts
+++ b/src/components/side_panel/pivot/pivot_title_section/pivot_title_section.ts
@@ -2,9 +2,12 @@ import { useProps } from "@odoo/owl";
 import { ActionSpec } from "../../../../actions/action";
 import { UuidGenerator } from "../../../../helpers/uuid";
 import { Component } from "../../../../owl3_compatibility_layer";
+import { useStore } from "../../../../store_engine/store_hooks";
+import { InsertPivotStore } from "../../../../stores/insert_pivot";
 import { _t } from "../../../../translation";
 import { CommandResult } from "../../../../types/commands";
 import { SpreadsheetChildEnv } from "../../../../types/spreadsheet_env";
+import { Store } from "../../../../types/store_engine";
 import { types } from "../../../props_validation";
 import { TextInput } from "../../../text_input/text_input";
 import { CogWheelMenu } from "../../components/cog_wheel_menu/cog_wheel_menu";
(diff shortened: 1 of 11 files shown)


src/components/side_panel/pivot/pivot_title_section/pivot_title_section.ts
hokolomopo: result is useless now

hokolomopo, line 90: the side panel will open even if the pivot wasn't created

src/stores/insert_pivot.ts
```diff
@@ -4,36 +4,25 @@ import { createPivotFormula } from "../../helpers/pivot/pivot_helpers";
...
-import { Command, CommandResult } from "../../types/commands";
+import { Command } from "../../types/commands";
 import { CellPosition, HeaderIndex, UID } from "../../types/misc";
 import { PivotTableData } from "../../types/pivot";
-import { UIPlugin } from "../ui_plugin";
-
-export class InsertPivotPlugin extends UIPlugin {
-  static getters = [] as const;
-
-  allowDispatch(cmd: Command) {
```
hokolomopo, line 16: In the others tasks I kept the the method allowDispatch but renamed it into canDuplicatePivot. That way you could use the same logic in both the dispatch and in the component. Not 100% convinced it's the right way of doing things, but probably a bit better than duplicating the logic.

It could also be used in the test then.

tests/pivots/pivot_insert.test.ts
```diff
@@ -4,10 +4,12 @@ import { toZone } from "../../src/helpers/zones";
 import { copy, insertPivot, paste } from "../test_helpers/commands_helpers";
 import { getCell, getCellText, getTable } from "../test_helpers/getters_helpers";
 import { createModelFromGrid, setGrid } from "../test_helpers/helpers";
+import { makeGlobalStoreWithModel } from "../test_helpers/stores";
 
 describe("Insert pivot command", () => {
   test("Can insert a pivot in a cell", () => {
     const model = new Model();
+    makeGlobalStoreWithModel(model);
```
hokolomopo: nitpick: maybe rename to `instanciateGlobalStoresWithModel` ? or keep the `make` but add the plural

hokolomopo (commented):
👋

the commit messages are a bit foireux FYI

src/stores/insert_pivot.ts
fdamhaut: The issue i have here is that the logic would need to be duplicated anyway to know which error message to show.
And returning an error message/dispatch result would feel wrong as well.

tests/pivots/pivot_insert.test.ts
fdamhaut: Indeed forgot s 😁, I like make more as it is consistent with others makeStore functions + it create+inject the notification store

pro-odoo:
First commit merged in https://github.com/odoo/o-spreadsheet/pull/9207

PR #9162, opened 2026-07-16
[FIX] table computed style: fix render table

VincentSchippefilt:
In table_computed_style.ts, the function `getTableComputedStyle` was modified to ensure that the computed styles for table elements are correctly retrieved and applied.
This fix addresses issues with rendering tables of Odoo pivots while they are loading, and have computed measures that fail when Loading.

Task: 6396322


pro-odoo (commented):
robodoo r+

PR #9195, opened 2026-07-22
[FIX] ui_sheet: `getCellWidth` getter for non-active sheet

hokolomopo:
## Description

The getter `getCellWidth` takes a cell position (with a sheetId) as argument, but inside the getter it uses the active sheet to get the column width instead. This make `AUTORESIZE_COLUMNS` command resize the column to the wrong with if called on a non-active sheet with text wrapping style.

Note that we never actually use `AUTORESIZE_COLUMNS` on a non-active sheet in practice.

Task: [6409426](https://www.odoo.com/odoo/2328/tasks/6409426)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


pro-odoo (commented):
robodoo r+

PR #9251, opened 2026-07-28
[FIX] chart: geoChart uses same d3 projection for main and full screen chart

rmbh-odoo:
## Description:

Current behavior before PR:
- For North America, we pass a custom d3 projection to rotate the 
`conicConformal` projection.
- However, since functions are not deep copied, the same projection object is
shared between the main and full screen charts.
- ChartJS updates this projection in place, causing the charts to
reuse each other's scale and resulting in an incorrect chart size.
- Other regions are not affected as they pass the projection as a string,
allowing ChartJS to create a fresh d3 projection for each chart.

- We now pass `conicConformal` as a string as well and use a ChartJS plugin
to apply the required 100-degree rotation.
- This ensures each chart gets its own d3 projection instance and
prevents the projection state from being shared between the main and full screen charts.

Task: [6395379](https://www.odoo.com/odoo/2328/tasks/6395379)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/figures/chart/chartJs/chartjs_geo_projection_plugin.ts
+++ b/src/components/figures/chart/chartJs/chartjs_geo_projection_plugin.ts
@@ -1,0 +1,20 @@
+import { Plugin } from "chart.js";
+
+/**
+ * ChartJS plugin to apply custom changes to a d3 projection.
+ *
+ * We pass the projection as a string instead of a customized d3 object so
+ * Chart.js creates a fresh projection instance. Custom changes (e.g. rotation)
+ * are then applied in `beforeUpdate`.
+ *
+ * This is important because Chart.js mutates the projection instance at runtime,
+ * and a customize d3 projection object would not be copied during deepCopy.
+ */
+export const geoProjectionPlugin: Plugin = {
+  id: "geoProjection",
+  beforeUpdate(chart: any) {
+    if (chart.options?.scales?.projection?.projection === "conicConformal") {
+      chart.scales?.projection?.projection?.rotate([100, 0]); // Centered on the US
+    }
+  },
+};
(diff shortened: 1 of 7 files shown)


src/components/figures/chart/chartJs/chartjs_geo_projection_plugin.ts
```diff
@@ -0,0 +1,17 @@
...
+ * We pass the projection as a string instead of a customized d3 object so
+ * Chart.js creates a fresh projection instance. Custom changes (e.g. rotation)
+ * are then applied in `beforeUpdate`.
+ */
+export const geoProjectionPlugin: Plugin = {
+  id: "geoProjection",
+  beforeUpdate(chart: any) {
+    if (chart.options?.scales?.projection?.projection === "conicConformal") {
+      chart.scales?.projection?.projection?.rotate([100, 0]);
+    }
```
hokolomopo, line 18: nitpick: add back the small comment that said we center on the US

src/components/figures/chart/chartJs/chartjs_geo_projection_plugin.ts
```diff
@@ -0,0 +1,17 @@
+import { Plugin } from "chart.js";
+
+/**
+ * ChartJS plugin to apply custom changes to a d3 projection.
+ *
+ * We pass the projection as a string instead of a customized d3 object so
+ * Chart.js creates a fresh projection instance. Custom changes (e.g. rotation)
+ * are then applied in `beforeUpdate`.
```
hokolomopo, line 8: nitpick: add it's important because chartjs modifies the runtime in place, and the object isn't coped on deepcopy

hokolomopo (commented):
Small nitpicks, otherwise LGTM :)

src/components/figures/chart/chartJs/chartjs_geo_projection_plugin.ts
rmbh-odoo: Done :+1:

hokolomopo (commented):
@robodoo r+

PR #9253, opened 2026-07-28
[REF] automatic evaluation: move it to a core view plugin

pro-odoo:
Evaluation should only depend on the core and core view plugins. This commit moves the automatic evaluation feature to a core view plugin.

Note that this commit introduces a little regression: re-enabling automatic evaluation does not trigger evaluation automatically. This is because the automatic evaluation plugin is a core view plugin, and core view plugins cannot dispatch a command. We could add a ui_feature plugin with a beforeHandle that checks the state of the automatic evaluation plugin and dispatches the evaluation command, but this would be a bit overkill. Instead, we can just add a dispatch to the two places where automatic evaluation is re-enabled (the top bar and the data action).

Task: 6425453

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/plugins/ui_feature/ui_options.ts
+++ b/src/plugins/ui_feature/ui_options.ts
@@ -2,9 +2,8 @@ import { Command } from "../../types/commands";
 import { UIPlugin } from "../ui_plugin";
 
 export class UIOptionsPlugin extends UIPlugin {
-  static getters = ["shouldShowFormulas", "isAutomaticEvaluationEnabled"] as const;
+  static getters = ["shouldShowFormulas"] as const;
   private showFormulas: boolean = false;
-  private automaticEvaluation: boolean = true;
 
   // ---------------------------------------------------------------------------
   // Command Handling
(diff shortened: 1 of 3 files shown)


tests/evaluation/evaluation.test.ts
```diff
@@ -1577,7 +1577,7 @@ describe("Automatic evaluation", () => {
     expect(getEvaluatedCell(model, "A2").value).toBe(2);
   });
 
-  test("Re-enabling automatic evaluation triggers evaluation", () => {
+  test("Re-enabling automatic evaluation does not trigger evaluation", () => {
```
pro-odoo: See commit message

rrahir:
I think it's worth merging it with the evaluation  plugin

src/plugins/ui_feature/ui_options.ts
```diff
@@ -15,15 +14,6 @@ export class UIOptionsPlugin extends UIPlugin {
       case "SET_FORMULA_VISIBILITY":
         this.showFormulas = cmd.show;
         break;
-      case "SET_AUTOMATIC_EVALUATION":
-        const wasDisabled = !this.automaticEvaluation;
```
rrahir, line 19: I think that the "was disabled" is still meaningful. If I am already in automatic mode, "activating" it again should not trigger an evaluation. But it really feels like an allowDispatch situation...

rrahir (changes requested):

src/plugins/ui_feature/ui_options.ts
pro-odoo: Done

rrahir (commented):
robodoo r+

PR #9254, opened 2026-07-28
[REF] evaluation: drop UIPlugin dependencies from formula evaluation

pro-odoo:
Move FilterEvaluationPlugin and PivotPresencePlugin to core view plugins
so that cell evaluation no longer reads from any UIPlugin getter.

- FilterEvaluationPlugin (isRowFiltered) moves to ui_core_views/, extends
  CoreViewPlugin. Its getFirstTableInSelection helper — which relied on
  UIPlugin getters (getActiveSheetId, getSelectedZones) — moves to
  SheetUIPlugin instead.
- PivotPresencePlugin (getPivotPresenceTracker) moves to ui_core_views/,
  extends CoreViewPlugin. PIVOT_START_PRESENCE_TRACKING now carries the
  sheetId to track; the tracker exposes itself only when both pivotId and
  sheetId match, replacing the getActiveSheetId gate previously used in
  module_lookup.ts by PIVOT.VALUE / PIVOT.HEADER.

Add a new EvaluationGetters type restricted to CoreGetters and core view
plugin getters (no UIPlugin), and use it in the evaluation layer:
EvalContext, Evaluator, buildCompilationParameters, and the
onIterationEndEvaluationRegistry. Formulas and evaluation code now fail
to type-check if they try to reach a UIPlugin getter.

Task: 6428159

### DIFF
--- a/src/components/pivot_html_renderer/pivot_html_renderer.ts
+++ b/src/components/pivot_html_renderer/pivot_html_renderer.ts
@@ -68,7 +68,8 @@ export class PivotHTMLRenderer extends Component<SpreadsheetChildEnv> {
   }
 
   get tracker() {
-    return this.env.model.getters.getPivotPresenceTracker(this.props.pivotId);
+    const sheetId = this.env.model.getters.getActiveSheetId();
+    return this.env.model.getters.getPivotPresenceTracker(this.props.pivotId, sheetId);
   }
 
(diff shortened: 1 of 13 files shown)


pro-odoo:
https://github.com/odoo/o-spreadsheet/pull/9439

PR #9257, opened 2026-07-30
[FIX] autofill: restore drag highlight

dhrp-odoo:
## Description:

When dragging the autofill handle, the target zone should be outlined before dropping.

This stopped working when the autofill plugin was converted into a store: the rendering layer was still declared as a static plugin layer, so it was ignored by the renderer store and the autofill drawLayer was never called.

Expose the Autofill layer through the store renderingLayers getter.

Task: [6424354](https://www.odoo.com/odoo/2328/tasks/6424354)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/autofill/autofill_store.ts
+++ b/src/components/autofill/autofill_store.ts
@@ -78,8 +78,6 @@ class AutofillGenerator {
 }
 
 export class AutofillStore extends SpreadsheetStore {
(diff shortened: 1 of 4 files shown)


rrahir:
robodoo r+

PR #9274, opened 2026-08-03
Stop throwing EvaluationError in getters

matho-odoo:
## Description:

Task: [6306250](https://www.odoo.com/odoo/2328/tasks/6306250)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/functions/module_text.ts
+++ b/src/functions/module_text.ts
@@ -3,7 +3,7 @@ import { escapeRegExp, trimContent } from "../helpers/misc";
 import { _t } from "../translation";
 import { CellErrorType, EvaluationError, NotAvailableError } from "../types/errors";
 import { AddFunctionDescription } from "../types/functions";
-import { Arg, FunctionResultNumber, FunctionResultObject, Maybe } from "../types/misc";
+import { Arg, FunctionResultObject, Maybe } from "../types/misc";
 import { arg } from "./arguments";
 import {
   matrixMap,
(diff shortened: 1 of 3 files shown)


src/functions/module_text.ts
```diff
@@ -168,6 +168,10 @@ export const FORMAT_LARGE_NUMBER = {
     value: Maybe<FunctionResultObject>,
     unite: Maybe<FunctionResultObject>
   ): FunctionResultNumber {
+    const uniteValue = unite?.value;
```
pro-odoo, line 169: Crash with `=FORMAT.LARGE.NUMBER(1000, True)`

Probably just:
```suggestion
    const uniteValue = toString(unite?.value);
```

Add a test for this case, to avoid the same issue later :)

src/functions/module_text.ts
```diff
@@ -169,6 +169,13 @@ export const FORMAT_LARGE_NUMBER = {
...
+    if (
+      uniteValue !== undefined &&
+      (typeof uniteValue !== "string" || !["k", "m", "b"].includes(uniteValue))
+    ) {
+      throw new EvaluationError(_t("The formatting unit should be 'k', 'm' or 'b'."));
```
rrahir:
```suggestion
      return new EvaluationError(_t("The formatting unit should be 'k', 'm' or 'b'."));
```

pro-odoo:
robodoo r+

PR #9289, opened 2026-08-05
[IMP] store: move zoom level to its own shared store

pro-odoo:
Zoom level was owned by ViewportsStore, so each viewport store had its own zoom, forcing the carousel to forward the main grid's zoom through a prop and a setZoom in a useEffect.

This commit extract the zoom level to its own store.

Task: 6448259


rrahir:
robodoo r+

PR #9311, opened 2026-08-10
[FIX] functions: omit functions from the squisher

rrahir:
Some functions do not react nicely to the squisher as the squished file becomes harder to post-process. In Odoo, for instance, we need to extract translatable terms and if it is squished, the extraction method then needs a case-by-case handling (squished or not) and requires a lot of extra code (basically some unsquishing logic) to extract the terms.

This revision adds a registry for the squisher to ignore the formulas including those problematic functions.

Task: 6452075

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -140,7 +140,7 @@ import { HoveredTableStore } from "./components/tables/hovered_table_store";
 import { TextInput } from "./components/text_input/text_input";
 import { ChartTerms } from "./components/translations_terms";
 import { arg } from "./functions/arguments";
-import { functionRegistry } from "./functions/function_registry";
+import { NonSquishableFunctionRegistry, functionRegistry } from "./functions/function_registry";
 import { PositionMap } from "./helpers/cells/position_map";
 import * as CHART_HELPERS from "./helpers/figures/charts";
 import { chartJsExtensionRegistry } from "./helpers/figures/charts/chart_js_extension";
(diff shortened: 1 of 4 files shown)


src/index.ts
```diff
@@ -325,6 +325,7 @@ export const registries = {
   pivotToFunctionValueRegistry,
   migrationStepRegistry,
   chartJsExtensionRegistry,
+  NonSquishableFunctionsRegistry: NonSquishableFunctionRegistry,
```
pro-odoo: Should be the same everywhere IMO

src/plugins/core/squisher.ts
```diff
@@ -105,7 +106,8 @@ export class Squisher {
    * The result of this method is:
    * - if the cell is not a formula, returns the content as is and resets the base formula
    * - if the cell is a formula:
-   *   - if there is no previous formula, or the normalized formula is different from the previous one, resets the base formula to this one and returns the full formula string
+   *   - if there is no previous formula, or the normalized formula is different from the previous one or it contains blacklisted functions (see ShittyFunctions),
```
pro-odoo: Not ShittyFunctions anymore :D

rrahir:
it's fixed and reexported in odoo

pro-odoo (approved):
robodoo r+

PR #9323, opened 2026-08-12
[FIX] Session: deactivate command squisher

rrahir:
Some issues were reported with the squished revisions where there is a loss of information and even command corruption when handling edge-cases like the presence of update_cell with an empty content or content with inlined format (e.g. `100%`, `2/2/2022`, etc).

Since these issues are nocive and can simply break a spreadsheet through their corrupted revisions, we deactivate the squisher on the session for the time being until we can come up with a satifying and robust solution.

reference task with the issues 6462247

Task: [6462247](https://www.odoo.com/odoo/2328/tasks/6462247)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/collaborative/session.ts
+++ b/src/collaborative/session.ts
@@ -431,7 +431,9 @@ export class Session extends EventBus<CollaborativeEvent> {
       }
       message = {
         ...message,
-        commands: this.commandSquisher.squish(revision.commands),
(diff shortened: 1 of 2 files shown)


pro-odoo:
robodoo r+

PR #9333, opened 2026-08-12
[REF] print: use an iframe to print the spreadsheet

hokolomopo:
## Description

When printing a spreadsheet, we would display a preview of the print, then use `@media print` rules to try to get a good print result. But the rules needed depend on the the web page where the spreadsheet is embedded.

That meant that every spreadsheet embedding would need to have its own print rules, which is annoying the first time, but even more so considering that any change in the CSS/DOM anywhere could break spreadsheet printing. And it's untestable.

With this commit, we create an iframe to display the spreadsheet print preview, and we print this iframe. This allow us full control of what's printed, independent of the embedding page.

Task: [6466136](https://www.odoo.com/odoo/2328/tasks/6466136)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/helpers/draw_grid_hook.ts
+++ b/src/components/helpers/draw_grid_hook.ts
@@ -1,4 +1,3 @@
-import type { Signal } from "@odoo/owl";
 import { CANVAS_SHIFT } from "../../constants";
 import { useLayoutEffect } from "../../owl3_compatibility_layer";
 import { useLocalStore, useStore } from "../../store_engine/store_hooks";
@@ -9,7 +8,7 @@ import { Store } from "../../types/store_engine";
 import { cssPropertiesToCss } from "./css";
 
 interface GridDrawingArgs {
-  canvasRef: Signal<HTMLElement | null>;
+  canvasRef: () => HTMLElement | null;
   renderingCtx: () => Omit<GridRenderingContext, "ctx" | "thinLineWidth">;
   rendererStore?: Store<RendererStore>;
   changeCanvasSizeOnZoom?: boolean;
@@ -26,7 +25,10 @@ export function useGridDrawing({
   useLocalStore(GridRenderer, renderer);
 
   function drawGrid() {
-    const canvas = canvasRef() as HTMLCanvasElement;
+    const canvas = canvasRef() as HTMLCanvasElement | null;
+    if (!canvas) {
+      return;
+    }
     const ctx = canvas.getContext("2d", { alpha: false })!;
     const renderingContext: GridRenderingContext = {
       ctx,
--- a/src/components/helpers/draw_grid_hook.ts
+++ b/src/components/helpers/draw_grid_hook.ts
@@ -49,6 +49,7 @@ export function useGridDrawing({
     if (width === 0 || height === 0) {
       return;
     }
+    ctx.resetTransform();
     // Imagine each pixel as a large square. The whole-number coordinates (0, 1, 2…)
     // are the edges of the squares. If you draw a one-unit-wide line between whole-number
     // coordinates, it will overlap opposite sides of the pixel square, and the resulting
--- a/src/components/spreadsheet_print/print_iframe.ts
+++ b/src/components/spreadsheet_print/print_iframe.ts
@@ -1,0 +1,193 @@
+import { onMounted, Signal, signal } from "@odoo/owl";
+import { useLayoutEffect } from "../../owl3_compatibility_layer";
+import { _t } from "../../translation";
+import { keyboardEventToShortcutString } from "../helpers/dom_helpers";
+
+const PRINT_IFRAME_CSS = /* css */ `
+  html, body {
+    margin: 0;
+    padding: 0;
+  }
+
+  body {
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    background-color: #e7e9ed;
+    font-family: "Roboto", "Arial";
+    min-width: min-content; /* Ensure the body does not shrink smaller than its content */
+  }
+
+  .o-print-page {
+    display: flex;
+    justify-content: center;
+    flex-shrink: 0;
+    box-sizing: border-box;
+    margin: 1rem;
+    background-color: #ffffff;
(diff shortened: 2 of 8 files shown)


src/components/spreadsheet_print/print_iframe.ts
```diff
@@ -0,0 +1,192 @@
...
+  function initializeIframe() {
+    const iframe = args.iframeRef();
+    if (!iframe || !iframe?.contentDocument) {
+      throw new Error("No iframe element found for print preview");
+    }
+
+    const doc = iframe.contentDocument;
+    styleElement = doc.createElement("style");
+    doc.head.appendChild(styleElement);
+    doc.body.replaceChildren();
```
hokolomopo, line 113: Since the iframe is empty (no src/srcDoc), it should be synchronously loaded with an about:blank document and we don't have to wait for a load event.

There was a Firefox bug with that (iframe document would be overriden on load event), but it seems to be fixed https://bugzilla.mozilla.org/show_bug.cgi?id=543435

src/components/spreadsheet_print/print_iframe.ts
```diff
@@ -0,0 +1,171 @@
...
+    margin: 0;
+    padding: 0;
+  }
+
+  body {
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    background-color: #e7e9ed;
+    font-family: "Roboto", "Arial";
```
rrahir, line 17:
```suggestion
    font-family: "Roboto", "Arial", 'Liberation Sans'
```

can't you use variables like --os-default-font ? will work with https://github.com/odoo/o-spreadsheet/pull/9297/changes

src/components/spreadsheet_print/print_iframe.ts
```diff
@@ -0,0 +1,171 @@
...
+  .o-empty-print-page {
+    align-items: center;
+    color: #5f636f;
+    font-style: italic;
+  }
+
+  @media print {
+    body {
+      display: block;
+      background-color: transparent;
```
rrahir, line 40: not sure but you probably need something like
`width: fit-content` such that the body takes the width of its larger child.

Currently, if we print in landscape on a large sheet and we reduce de width of the browser, some parts of the pages cannot be seen ,nor scrolled to

src/components/spreadsheet_print/print_iframe.ts
rrahir: are you sure it's fixed ? I have trouble navigating bugzilla, but we probably have to make sure that not only it's fixed (seems ok on my end) but it has to be fixed in the latest LTS of firefox as well. it's called firefox ESR and I think it's currenty in version 140, which is quite old :'(

rrahir: it's broken on Firefox ESR

Image

rrahir (changes requested):

src/components/spreadsheet_print/print_iframe.ts
hokolomopo: Well css variables of parent document don't work in child iframe AFAIK

src/components/spreadsheet_print/print_iframe.ts
hokolomopo: Yeah you're right, It's fixed in latests firefox versions but not 140... Added some code to handle that

src/components/helpers/draw_grid_hook.ts
```diff
@@ -55,6 +55,7 @@ export function useGridDrawing({
     // line will be drawn two pixels wide. To draw a line that is only one pixel wide,
     // you need to shift the coordinates by 0.5 perpendicular to the line's direction.
     // http://diveintohtml5.info/canvas.html#pixel-madness
+    ctx.resetTransform();
```
rrahir: the comment is there for the line below,  i would swap the lines if possible. 

Did you monitor if this explicit call had a cost? we could imagine doing it in the tests only w/ an explanation if it costs a lot 🤷

src/components/spreadsheet_print/print_iframe.ts
```diff
@@ -0,0 +1,192 @@
...
+    const iframe = args.iframeRef();
+    if (!iframe || !iframe?.contentDocument) {
+      throw new Error("No iframe element found for print preview");
+    }
+
+    iframe.addEventListener("load", () => {
+      // In older firefox versions with an `about:blank` iframe, the iframe would have a readyState "complete" at first
+      // render, marking it ready to use. But then an async navigation event would occur, replacing the whole iframe document.
+      // And for those `about:blank` iframes, other browsers might not dispatch a `load` event at all.
+      // https://bugzilla.mozilla.org/show_bug.cgi?id=543435
```
rrahir, line 94: nitpick
```suggestion
      // https://bugzilla.mozilla.org/show_bug.cgi?id=543435
      // FIXME: can be decommissioned when Firefox ESR reaches version > 148
```

src/components/helpers/draw_grid_hook.ts
hokolomopo: I asked claude to run a benchmark, and resetTransform seems free. But per this benchmark, we could save a bit by not calling canvas.width at each render if the width didn't change

image

rrahir:
robodoo rebase-ff

src/components/helpers/draw_grid_hook.ts
rrahir: ok, let's keep the resetTransform at this time, and come up with an improvement to avoid noisy changes on the canvas dimensions (I think we can monitor  the zoom/pixelRatio and for the change of width/height, I would also debounce the rendering over a change of browser viewport dimension but it's another discussion in itself.


Since we merged the print task yesterday, there is now a small (hopefully) rebase, 

@robodoo delegate+

hokolomopo:
@robodoo r+

rrahir:
robodoo r+

PR #9339, opened 2026-08-13
[FIX] xlsx: fix geo chart xlsx export

hokolomopo:
## Description

If a geo chart was present when exporting a spreadsheet to xlsx, more often than not the exported chart would be empty. It was because we didn't wait for the geo json to be loaded before exporting the xlsx.

This commit introduces a new helper `waitForSpreadsheetDataLoaded` that can be called before any spreadsheet export to ensure that all the data is loaded.

Task: [4632983](https://www.odoo.com/odoo/2328/tasks/4632983)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/demo/main.js
+++ b/demo/main.js
@@ -48,6 +48,7 @@ topbarMenuRegistry.addChild("xlsx", ["file"], {
   name: "Save as XLSX",
   sequence: 20,
   execute: async (env) => {
+    await env.model.getters.loadUsedGeoJsonFeatures();
     const doc = await env.model.exportXLSX();
     const zip = new JSZip();
     for (const file of doc.files) {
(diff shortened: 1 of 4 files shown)


demo/main.js
```diff
@@ -48,6 +48,7 @@ topbarMenuRegistry.addChild("xlsx", ["file"], {
   name: "Save as XLSX",
   sequence: 20,
   execute: async (env) => {
+    await env.model.getters.loadUsedGeoJsonFeatures();
```
rrahir, line 51: add this directly to export xlsx ? it's already async

hokolomopo: I don't want that a call to exportXlsc() to secretly dispatch commands & change the plugin states, it seems wrong to me :/

rrahir: getter that dispatches? 

Anyway, the current implementation means that any person (@nseinlet ) that uses the library will never see their geochart export work because there is no way to know you have to call that getter.

rrahir: and again, if getter alters a state, then it's not a getter

hokolomopo:
@robodoo up to saas-19.4

rrahir:
robodoo r+

PR #9365, opened 2026-08-17
[FIX] chart_menu : icons

matho-odoo:
Updated the class of the menu icon so that it is the same as the other icons.

Task: [6469005](https://www.odoo.com/odoo/2328/tasks/6469005)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/figures/chart/chart_menu/chart_menu.xml
+++ b/src/components/figures/chart/chart_menu/chart_menu.xml
@@ -13,7 +13,7 @@
         </t>
         <t t-foreach="this.getMenuItems()" t-as="item" t-key="item.id">
           <div
-            class="o-chart-menu-item border-0 lh-1 p-1 d-flex align-items-center justify-content-center rounded"
+            class="o-chart-menu-item lh-1 p-1 d-flex align-items-center justify-content-center rounded"
             t-att-title="item.label"
             t-on-click="item.onClick"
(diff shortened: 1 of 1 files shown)


src/components/figures/chart/chart_menu/chart_menu.xml
```diff
@@ -22,7 +22,7 @@
         </t>
         <t t-if="this.isMenuAvailable()">
           <button
-            class="o-chart-menu-item btn text-muted lh-1 p-1 fa fa-ellipsis-v"
```
hokolomopo, line 25: others icons are \<div\> this one is a \<button\>, we should be consistent if we want the classes to be consistent. border-0 is also probably useless for a div

src/components/figures/chart/chart_menu/chart_menu.xml
```diff
@@ -22,7 +22,7 @@
         </t>
         <t t-if="this.isMenuAvailable()">
```
hokolomopo, line 23: The icon now always have an hover-like style event when not hovered

Image

hokolomopo (commented):
👋 

Maybe be slightly more explicit in your commit title 🙃

hokolomopo:
@robodoo r+

PR #9366, opened 2026-08-17
[FIX] print: figure border wrong render

hokolomopo:
## Description

When printing, the figure border was drawn with a `strokeRect`. But its argument were wrong, `strokeRect` draw a border centered of the rectangle given as args, which mean that if we want a border to be drawn inside the rectangle we have to offset it by half a pixel.

Task: [6458624](https://www.odoo.com/odoo/2328/tasks/6458624)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


pro-odoo (commented):
robodoo r+

PR #9376, opened 2026-08-18
[IMP] topbar: improve banner style

hokolomopo:
## Description

This commit improves a bit the style of the topbar banner, adding a bit of padding and increasing its size when its content is too large to fit on one line. It also removes a double border inside the banner that appeared because we used the bootstrap alert class.

Task: [6478877](https://www.odoo.com/odoo/2328/tasks/6478877)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/top_bar/top_bar.xml
+++ b/src/components/top_bar/top_bar.xml
@@ -84,23 +84,25 @@
       </div>
       <div
         t-if="this.fingerprints.isEnabled"
-        class="topbar-banner irregularity-map d-flex align-items-center justify-content-between border-top">
+        class="topbar-banner irregularity-map d-flex align-items-stretch justify-content-between border-top">
         <div
           t-on-click="() => this.fingerprints.disable()"
           role="button"
           title="This tool analyzes spreadsheet formulas for patterns and highlights inconsistencies. Irregularities may indicate potential errors in formula structures, references, or arguments. (Click to turn off)"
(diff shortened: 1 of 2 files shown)


src/components/top_bar/top_bar.xml
```diff
@@ -84,42 +84,44 @@
       </div>
       <div
         t-if="this.fingerprints.isEnabled"
-        class="topbar-banner irregularity-map d-flex align-items-center justify-content-between border-top">
+        class="topbar-banner irregularity-map d-flex align-items-stretch justify-content-between border-top">
```
pro-odoo, line 87: Should we create a component topbar-banner ?
Not sure it's a good idea, and probably not in this task

hokolomopo: I did try for a bit to make something more generic, but the banners were just different enough that I didn't bother:
 - one has a clickable header the other not
 - one has a tooltip on title
 - the buttons have different labels
 - the colors are handled with 2 different BS classes & some o-spreadsheet colors (which is probably a problem in itself, but the BS colors are a bit different & prettier than spreadsheet colors, but again different in odoo).

My conclusion was: flemme. But we probably might want to remove all those differences and do something standard sometime.

pro-odoo:
robodoo r+

PR #9377, opened 2026-08-18
[FIX] evaluation: wrong dependencies invalidation with spread formulas

hokolomopo:
## Description

- Write =MUNIT(3) in A1
- Write =B2 in B3
- A1 will now be #SPILL!, but B3 will be 1 instead of 0

We didn't properly invalidate the spill dependencies if the cell we changed the value of was inside a spread formula.

Task: [5103328](https://www.odoo.com/odoo/2328/tasks/5103328)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


pro-odoo:
robodoo r+

PR #9384, opened 2026-08-19
[IMP] viewports_store: do not jump viewport on UNDO/REDO

hokolomopo:
On UNDO/REDO we would scroll all the viewports to the position of the selection. But only if no figure was selected.

I have no idea why we did it, it was introduced in seemingly unrelated commit 7b9055e, never tested, and didn't make much sense. We can most likely remove it.

Task: [6481352](https://www.odoo.com/odoo/2328/tasks/6481352)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


rrahir:
robodoo r+

PR #9389, opened 2026-08-19
[FIX] grid: hide AddRowFooter when the mainViewport is too small

rrahir:
How to reproduce:
- insert a list
- freeze pane to the lowest visible row
- start scrolling to get to the bottom of the page

=> the full layout is broken

Task: 6103620

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/internal_viewport.ts
+++ b/src/helpers/internal_viewport.ts
@@ -87,8 +87,12 @@ export class InternalViewport {
     if (this.canScrollVertically) {
       height = Math.max(height, this.viewportHeight); // if the viewport grid size is smaller than its client height, return client height
 
-      if (lastRowEnd + FOOTER_HEIGHT > height && !this.getters.isReadonly()) {
-        height += FOOTER_HEIGHT;
+      if (
+        lastRowEnd + FOOTER_HEIGHT > height &&
+        !this.getters.isReadonly() &&
+        this.viewportHeight > FOOTER_HEIGHT
+      ) {
+        height += FOOTER_HEIGHT; // if the footer is visible, we add its height to the viewport height
       }
     }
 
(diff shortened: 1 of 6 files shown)


src/helpers/internal_viewport.ts
```diff
@@ -87,8 +87,13 @@ export class InternalViewport {
...
       height = Math.max(height, this.viewportHeight); // if the viewport grid size is smaller than its client height, return client height
 
-      if (lastRowEnd + FOOTER_HEIGHT > height && !this.getters.isReadonly()) {
-        height += FOOTER_HEIGHT;
+      if (
+        lastRowEnd + FOOTER_HEIGHT > height &&
+        !this.getters.isReadonly() &&
+        this.viewportHeight > FOOTER_HEIGHT
+      ) {
+        //  && this.viewportHeight > FOOTER_HEIGHT
```
hokolomopo: comment

hokolomopo (commented):
👋

src/helpers/internal_viewport.ts
```diff
@@ -87,8 +87,12 @@ export class InternalViewport {
     if (this.canScrollVertically) {
       height = Math.max(height, this.viewportHeight); // if the viewport grid size is smaller than its client height, return client height
 
-      if (lastRowEnd + FOOTER_HEIGHT > height && !this.getters.isReadonly()) {
-        height += FOOTER_HEIGHT;
+      if (
+        lastRowEnd + FOOTER_HEIGHT > height &&
+        !this.getters.isReadonly() &&
+        this.viewportHeight > FOOTER_HEIGHT
```
hokolomopo, line 93: Conceptually it's kinda strange that we wouldn't show the footer if it's bigger than the viewport. From the viewport POV, it should just increase its scrollable height so it includes the footer, no ? Probably a harder fix tho

rrahir: the fix is indeed way easier this way and the footer is not usable in any way if the viewport is too small. we talk about a less than 50 px viewport. you get a warning in the interface when the viewport is smaller than 1/3 (or 1/4 or something like that) than the full height of the canvas so I don't think it's really an issue.

hokolomopo:
@robodoo r+

PR #9395, opened 2026-08-19
[IMP] menu_registry: ensure children unicity

hokolomopo:
## Description

The ids in a menu registry are supposed to be unique, but if the children are computed at runtime from a function, they could have duplicate ids.

Task: [5224033](https://www.odoo.com/odoo/2328/tasks/5224033)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/actions/action.ts
+++ b/src/actions/action.ts
@@ -140,11 +140,19 @@ export function createAction(item: ActionSpec): Action {
       : undefined,
     children: children
       ? (env) => {
-          return children
-            .map((child) => (typeof child === "function" ? child(env) : child))
-            .flat()
-            .map(createAction)
-            .sort((a, b) => a.sequence - b.sequence);
+          const uniqueChildren: Record<string, Action> = {};
+          children
+            .flatMap((child) => (typeof child === "function" ? child(env) : child))
+            .forEach((childSpec) => {
+              const childAction = createAction(childSpec);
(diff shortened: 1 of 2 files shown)


src/actions/action.ts
```diff
@@ -140,11 +140,16 @@ export function createAction(item: ActionSpec): Action {
...
-            .map((child) => (typeof child === "function" ? child(env) : child))
-            .flat()
-            .map(createAction)
-            .sort((a, b) => a.sequence - b.sequence);
+          const uniqueChildren: Record<string, Action> = {};
+          children
+            .flatMap((child) => (typeof child === "function" ? child(env) : child))
+            .forEach((childSpec) => {
+              const childAction = createAction(childSpec);
+              if (!uniqueChildren[childAction.id]) {
```
pro-odoo: Maybe we should throw if the id already exists ? It will avoid silent fail...

hokolomopo: I thought about doing, that and I'm not sure if we want to throw at runtime (vs the "normal" addchild that throws when inserting in the registry). But I'm 50/50 on hat, I can change

pro-odoo: I would do it, at least we will see if there's an error :)

pro-odoo (commented):
robodoo r+

PR #9409, opened 2026-08-21
[IMP] chart: render invalid stacked area values as zero

dhrp-odoo:
## Description:

Stacked area charts pass non-numeric values to Chart.js as NaN, which breaks the line and filled area. Google Sheets instead renders these values at the baseline.

Normalize non-numeric values to zero when extracting data for stacked area charts. The normalization happens after invalid points without a label have been filtered out, so no empty categories are introduced.

Other chart types keep their existing behavior: invalid values remain gaps in regular area and line charts.

Task: [6458796](https://www.odoo.com/odoo/2328/tasks/6458796)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/figures/charts/runtime/chart_data_extractor.ts
+++ b/src/helpers/figures/charts/runtime/chart_data_extractor.ts
@@ -277,7 +277,9 @@ export function getLineChartData(
           formatValue(value, { format, locale: getters.getLocale() })
         );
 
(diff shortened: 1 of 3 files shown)


hokolomopo (commented):
@robodoo r+

PR #9422, opened 2026-08-24
[FIX] chart suggestions: fix header for category

anhe-odoo:
## Task Description

When creating a chart from the chart suggestion with a category column, we currently decide that there is no header in the column and count the first row as an entry. This PR aims to change this behavior to use the first entry as an header in a more smart way :
 * If we have only category, we use the first entry as header
 * If we have a column of another type, we use the first entry as header if the other column has a header.

## Related Task

Task: 0

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -242,7 +242,7 @@ export const DEFAULT_WINDOW_SIZE = 2;
 
 // Maximum number of data points rendered by Chart.js in lightweight previews (drag preview,
 // chart suggestions), sampled down from the full dataset for performance.
-export const MAX_CHART_PREVIEW_DATA_POINTS = 150;
+export const MAX_CHART_PREVIEW_DATA_POINTS = 20;
 
 // session
 export const MESSAGE_VERSION = 1;
--- a/src/helpers/figures/charts/chart_suggestion_engine.ts
+++ b/src/helpers/figures/charts/chart_suggestion_engine.ts
@@ -150,6 +150,18 @@ function interestingCellsXc(col: ColumnAnalysis): {
   return { firstCellXC, lastCellXC, prevCellXC };
 }
 
+function getCategoryHeader(getters: Getters, catCol: ColumnAnalysis): string {
+  let catHeader = catCol.header ?? "";
+  if (!catHeader && catCol.rowCount > 0) {
+    const { left, top } = catCol.zone;
+    catHeader = (
+      getters.getEvaluatedCell({ sheetId: getters.getActiveSheetId(), col: left, row: top })
+        ?.value || _t("Category")
+    ).toString();
+  }
+  return catHeader;
+}
+
 /** Pattern A — Single numeric column */
 function buildSingleNumberContext([col]: ColumnAnalysis[], getters: Getters): SingleNumberContext {
   const title = col.header ?? _t("Value");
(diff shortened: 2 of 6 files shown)


src/constants.ts
```diff
@@ -243,7 +243,7 @@ export const DEFAULT_WINDOW_SIZE = 2;
 
 // Maximum number of data points rendered by Chart.js in lightweight previews (drag preview,
 // chart suggestions), sampled down from the full dataset for performance.
-export const MAX_CHART_PREVIEW_DATA_POINTS = 150;
+export const MAX_CHART_PREVIEW_DATA_POINTS = 20;
```
pro-odoo, line 245: Is it intended ? If yes, should be done in a separate commit

src/helpers/figures/charts/chart_suggestion_engine.ts
```diff
@@ -150,6 +150,17 @@ function interestingCellsXc(col: ColumnAnalysis): {
   return { firstCellXC, lastCellXC, prevCellXC };
 }
 
+function getCategoryHeader(getters: Getters, catCol: ColumnAnalysis): string {
+  let catHeader = catCol.header ?? "";
+  if (!catHeader && catCol.rowCount > 0) {
+    catHeader = (
+      getters.getEvaluatedCellsInZone(getters.getActiveSheetId(), catCol.zone)[0].value ||
```
pro-odoo: For a performance concern, we should give to `getEvaluatedCellsInZone` the zone of the top-left cell, and not the entire zone

pro-odoo (approved):
Create a task in Odoo, and put the task id in the commit. So François can test :)
(Put it in RD Validation green ;) )

src/helpers/figures/charts/chart_suggestion_engine.ts
```diff
@@ -150,6 +150,18 @@ function interestingCellsXc(col: ColumnAnalysis): {
...
+function getCategoryHeader(getters: Getters, catCol: ColumnAnalysis): string {
+  let catHeader = catCol.header ?? "";
+  if (!catHeader && catCol.rowCount > 0) {
+    const { left, top } = catCol.zone;
+    catHeader = (
+      getters.getEvaluatedCell({ sheetId: getters.getActiveSheetId(), col: left, row: top })
+        ?.value || _t("Category")
+    ).toString();
+  }
+  return catHeader;
```
hokolomopo, line 162: I don't really agree with using it when we have a single category column. There's _nothing_ to say that the first value is a header vs just a random value of the dataset 

See the screenshot, those chart do not make sense given the data
Image

src/helpers/figures/charts/chart_suggestion_engine.ts
```diff
@@ -182,18 +194,16 @@ function buildSingleCategoricalContext(
   [col]: ColumnAnalysis[],
```
hokolomopo, line 194: Not something you changed in this commit, but this selection has no suggested chart, is it normal ? Do the order of the column really matters ? 

Image

src/helpers/figures/charts/chart_suggestion_engine.ts
```diff
@@ -203,10 +213,11 @@ function buildCategoricalVsNumberContext(
 ): CategoricalVsNumberContext {
   const labelRange = getUnboundRange(getters, catCol.zone);
   const hasTitle = numCol.hasHeader;
+  const categoryHeader = hasTitle ? getCategoryHeader(getters, catCol) : _t("Category");
```
hokolomopo, line 217: Not something you changed in this commit, but do the default name (no header) really make sense ? I think having no title would be better than a "By category" or "Value" title that doesn't mean much

Image

hokolomopo (commented):
👋

src/helpers/figures/charts/chart_suggestion_engine.ts
matho-odoo: Yes you have to select the categories first and then the numbers :)

pro-odoo (approved):
robodoo rebase-ff r+

PR #9430, opened 2026-08-24
Saas 19.3 fix command squisher vsc

VincentSchippefilt:
## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/collaborative/session.ts
+++ b/src/collaborative/session.ts
@@ -431,9 +431,7 @@ export class Session extends EventBus<CollaborativeEvent> {
       }
       message = {
         ...message,
-        commands: revision.commands,
-        // deactivated squishing for now, as it is causing issues with empty cells & literals with inlined formats
-        //this.commandSquisher.squish(revision.commands),
+        commands: this.commandSquisher.squish(revision.commands),
       };
     }
     if (this.isReplayingInitialRevisions) {
--- a/src/collaborative/session.ts
+++ b/src/collaborative/session.ts
@@ -1,5 +1,5 @@
 import { DEFAULT_REVISION_ID, MESSAGE_VERSION } from "../constants";
-import { UuidGenerator } from "../helpers";
+import { deepEquals, UuidGenerator } from "../helpers";
 import { EventBus } from "../helpers/event_bus";
 import { isDefined } from "../helpers/misc";
 import { SelectiveHistory as RevisionLog } from "../history/selective_history";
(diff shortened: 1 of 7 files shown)


VincentSchippefilt:
robodoo rebase-ff

src/collaborative/session.ts
```diff
@@ -429,12 +431,28 @@ export class Session extends EventBus<CollaborativeEvent> {
...
       }
-      message = {
-        ...message,
-        commands: revision.commands,
-        // deactivated squishing for now, as it is causing issues with empty cells & literals with inlined formats
-        //this.commandSquisher.squish(revision.commands),
-      };
+
+      const squishedCommands = this.commandSquisher.squish(revision.commands);
+      if (this.shouldVerifySquish) {
```
pro-odoo: We should skip the verification if squishedCommands does not contains `SQUISHED_UPDATE_CELL`.

For exemple, write `=sum(a1:b1)` in a cell => squished failed because a1:b1 !== A1:B1

src/plugins/core/unsquisher.ts
```diff
@@ -197,6 +209,14 @@ export class Unsquisher {
...
             strategy = "COMBINE_OFFSET";
             break;
+          case "COMBINE_OFFSET":
+          case "OFFSET_NUMBER":
+            break;
+          default:
+            // e.g. a stale "NOT_A_FORMULA" surviving from a previous, unrelated cell: there is no
+            // formula/number to apply this offset on top of. Fail loudly instead of silently
+            // forwarding the raw offset object as if it were literal cell content.
+            throw new Error(`Cannot unsquish an offset without a preceding base cell: ${strategy}`);
```
pro-odoo: I'm not 100% sure, but throwing here will break existing spreadsheet with buggy squish command IMO

src/collaborative/session.ts
```diff
@@ -429,12 +431,28 @@ export class Session extends EventBus<CollaborativeEvent> {
...
+        const unsquishedCommands = this.commandSquisher.unsquish(squishedCommands);
+        if (!deepEquals(message.commands, unsquishedCommands)) {
```
pro-odoo: `revision.commands`
message.commands are the non-transformed ones

src/collaborative/session.ts
```diff
@@ -79,12 +79,14 @@ export class Session extends EventBus<CollaborativeEvent> {
...
    * @param commandSquisher used to squish and unsquish commands to reduce the size of messages sent to the server
+   * @param shouldVerifySquish either the commands should be unsquished and verified before being sent to the server
    */
   constructor(
     private revisions: RevisionLog<Revision>,
     private transportService: TransportService<CollaborationMessage>,
     private serverRevisionId: UID = DEFAULT_REVISION_ID,
-    private commandSquisher: ICommandSquisher
+    private commandSquisher: ICommandSquisher,
+    private shouldVerifySquish: boolean = true
```
pro-odoo, line 89: I'm not sure it's worth to add it to the constructor as we cannot set it from the model or a config. It's only used in a test that check that the squish is not verified if this param is set to false, I think it's a little bit useless.
We could define it at a private property

VincentSchippefilt: i'll keep it so it is configurable from the outside...

pro-odoo (approved):
robodoo r+

PR #9431, opened 2026-08-24
[FIX] calendar chart: wrong groupBy choice filtering

hokolomopo:
## Description

In the side panel of the calendar chart, we would filter the available granularities based on the chart data. There were a bunch of issues with this filtering:

- It's functionally limiting what the user can do. He cannot select a year granularity if the data is only for one year, which he might want to do to take future data into account.
- The side panel code would use different logic than the actual calendar runtime. Eg. calendar runtime accept non-date-formatted number values, the side panel would not.
- It wouldn't display the current granularity if it was not in the filtered list, showing a blank select.
- It would use private-ish chart helpers that we don't really want to use everywhere
- It would only work for range data, but in future version the side panel should be chart datasource-agnostic.
- It wasn't tested

This commit plainly removes the filtering logic, and instead show all available granularity. It's also consistent with what we do with pivot date granularities.

Task: [5358625](https://www.odoo.com/odoo/2328/tasks/5358625)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/components/side_panel/chart/calendar_chart/calendar_chart_config_panel.ts
+++ b/src/components/side_panel/chart/calendar_chart/calendar_chart_config_panel.ts
@@ -1,9 +1,4 @@
-import { toJsDate } from "../../../../functions/helpers";
-import { createValidRange, isDateTime } from "../../../../helpers";
-import { createDataSets } from "../../../../helpers/figures/charts";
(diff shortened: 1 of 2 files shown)


rrahir:
robodoo r+

PR #9438, opened 2026-08-25
[IMP] topbar: improve clear formatting action

hokolomopo:
## Description

With this commit, the clear formatting action does a bit more than clearing the format/style: it also clears merges, conditional formats and table styles.

We kept the old `CLEAR_FORMATTING` as it was so is doesn't break existing spreadsheet (and the command is still used once), and created a new `CLEAR_ALL_STYLING` command that clears everything else and dispatches a `CLEAR_FORMATTING`.

Task: [5391157](https://www.odoo.com/odoo/2328/tasks/5391157)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/actions/format_actions.ts
+++ b/src/actions/format_actions.ts
@@ -398,7 +398,7 @@ export const clearFormat: ActionSpec = {
   name: _t("Clear formatting"),
   shortcut: "Ctrl+<",
   execute: (env) =>
-    env.model.dispatch("CLEAR_FORMATTING", {
+    env.model.dispatch("CLEAR_ALL_STYLING", {
       sheetId: env.model.getters.getActiveSheetId(),
       target: env.model.getters.getSelectedZones(),
     }),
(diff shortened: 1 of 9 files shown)


src/actions/format_actions.ts
```diff
@@ -399,7 +398,7 @@ export const clearFormat: ActionSpec = {
   name: _t("Clear formatting"),
   shortcut: "Ctrl+<",
   execute: (env) =>
-    env.model.dispatch("CLEAR_FORMATTING", {
+    env.model.dispatch("CLEAR_ALL_STYLING", {
```
anhe-odoo: Will it need an upgrade PR then ?

anhe-odoo: Oh no, nevermind, it's a new command

PR #9443, opened 2026-08-26
[IMP] owl3: remove `env.getPopoverContainerRect`

hokolomopo:
## Description

This commit replaces the usage of `env.getPopoverContainerRect()` by an owl plugin `PopoverContainerPlugin`.

Some notes:
- there was an unused props `containerRect` for the Popover component
- the plugin is not reactive, we still need the `usePopoverContainer` hook. We'd need to make the zoom store and the viewport store reactive, and change Popover to use useEffect instead of useLayoutEffect.

Task: [6510584](https://www.odoo.com/odoo/2328/tasks/6510584)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


pro-odoo (approved):
robodoo r+

PR #9445, opened 2026-08-26
[REF] charts: resolve the default chart background at rendering time

pro-odoo:
The chart runtime used to read getSpreadsheetTheme() to default the chart background, which made an evaluation plugin depend on a UI one.

The runtime is theme-agnostic: it only carries the background coming from the chart definition (or, for scorecards and gauges, from the cell it is based on), and is undefined otherwise. The default is a rendering concern: the components substitute the current spreadsheet theme, while the rendering helpers fall back to the new DEFAULT_CHART_BACKGROUND_COLOR (white).

Charts are drawn inside a `os-theme-dependant` figure, so the theme colors they need are the ones pre-compensated for the dark mode CSS filter. The image and xlsx exports are not affected by that filter: they now export on white instead of an ugly one.

Task: 6507087

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/components/figures/chart/chartJs/chartjs.ts
```diff
@@ -105,6 +107,32 @@ export class ChartJsComponent extends Component<SpreadsheetChildEnv> {
...
+    return this.env.model.getters.getSpreadsheetTheme().backgroundColor;
+  }
+
+  /**
+   * Copy the runtime and resolve its theme-dependent colors. The runtime is theme-agnostic: it only
+   * carries the background explicitly set on the chart definition.
+   *
+   * Note: chartJS modifies the runtime in place, so it's important to give it a copy anyway.
+   */
+  private prepareRuntime(runtime: ChartJSRuntime): ChartJSRuntime {
```
hokolomopo, line 120: nitpick: put the method somewhere below the setup

src/components/figures/chart/gauge/gauge_chart_component.ts
```diff
@@ -30,6 +31,15 @@ export class GaugeChartComponent extends Component<SpreadsheetChildEnv> {
     return this.env.model.getters.getChartRuntime(this.props.chartId) as GaugeChartRuntime;
   }
 
+  get themedRuntime(): GaugeChartRuntime {
```
hokolomopo, line 34: same in here, leave the setup at the top

src/helpers/figures/charts/runtime/chart_data_extractor.ts
```diff
@@ -441,7 +442,7 @@ export function getHierarchalChartData(
     axisFormats: { y: getChartLabelFormat(labels) },
     labels: labels.map(({ value }) => String(value ?? "")),
     locale: getters.getLocale(),
-    background: getChartBackgroundColor(definition, getters),
+    background: definition.background || DEFAULT_CHART_BACKGROUND_COLOR,
```
hokolomopo, line 445: Well this is wrong isn't it ? the actual background is theme dependent, this isn't. And we use this mostly to generate bar/calendar/someOthers border colors, so now we have the wrong border colors in dark mode.

src/components/figures/chart/scorecard/chart_scorecard.ts
```diff
@@ -25,6 +27,20 @@ export class ScorecardChart extends Component<SpreadsheetChildEnv> {
     return this.env.model.getters.getChartRuntime(this.props.chartId) as ScorecardChartRuntime;
   }
 
+  get themedRuntime(): ScorecardChartRuntime {
```
hokolomopo, line 30: I feel this should be a getter in the color theme plugin, not the responsibility of every component/store that would want to draw a chart.

hokolomopo (commented):
👋 doens't work unfortunately 😔

PR #9450, opened 2026-08-27
[IMP] chart: evaluation do not rely on UI getter

pro-odoo:
Before this commit, evaluation chart relied on the UI getter to obtain the theme (dark or light).
After this commit, the theme is directly passed to the runtime creation function, so the chart can be evaluated independently of the UI.

Before this commit, downloading or copying as image a chart in dark mode would copy the chart with an ugly background, the inversed background. Now, it uses the current theme to render the chart correctly.

Task: 6514275

## Description:

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/helpers/color_themes.ts
+++ b/src/helpers/color_themes.ts
@@ -1,0 +1,41 @@
+import { ColorThemeName, GridRenderingTheme } from "../types/rendering";
+import { adaptForDarkMode } from "./color";
+
+const FROZEN_PANE_HEADER_BORDER_COLOR = "#BCBCBC";
+const FROZEN_PANE_BORDER_COLOR = "#DADFE8";
+const HEADER_BORDER_COLOR = "#C0C0C0";
+const TEXT_HEADER_COLOR = "#666666";
+const BACKGROUND_HEADER_COLOR = "#F8F9FA";
+const BACKGROUND_HEADER_SELECTED_COLOR = "#E8EAED";
+const BACKGROUND_HEADER_ACTIVE_COLOR = "#595959";
+
+export const COLOR_THEMES: Record<ColorThemeName, GridRenderingTheme> = {
+  light: {
+    colorThemeName: "light",
+    backgroundColor: "#FFFFFF",
+    gridBorderColor: "#CECFCF",
+    headerBackgroundColor: BACKGROUND_HEADER_COLOR,
+    headerActiveBackgroundColor: BACKGROUND_HEADER_ACTIVE_COLOR,
+    headerSelectedBackgroundColor: BACKGROUND_HEADER_SELECTED_COLOR,
+    headerTextColor: TEXT_HEADER_COLOR,
+    headerBorderColor: HEADER_BORDER_COLOR,
+    frozenPaneBorderColor: FROZEN_PANE_BORDER_COLOR,
+    frozenPaneHeaderBorderColor: FROZEN_PANE_HEADER_BORDER_COLOR,
+    singleCellSelectionBackgroundColor: "#F3F7FE",
+    multipleCellsSelectionBackgroundColor: "#E9F0FF",
+  },
+  dark: {
+    colorThemeName: "dark",
+    backgroundColor: adaptForDarkMode("#1A1C2E"),
+    gridBorderColor: adaptForDarkMode("#6B706F"),
+    headerBackgroundColor: adaptForDarkMode("#262A36"),
+    headerActiveBackgroundColor: adaptForDarkMode("#3A4052"),
+    headerSelectedBackgroundColor: adaptForDarkMode("#4E566E"),
+    headerTextColor: adaptForDarkMode("#A1A6B3"),
+    headerBorderColor: adaptForDarkMode("#7A7F91"),
+    frozenPaneBorderColor: adaptForDarkMode("#7A7F91"),
+    frozenPaneHeaderBorderColor: adaptForDarkMode("#9FA5BD"),
+    singleCellSelectionBackgroundColor: adaptForDarkMode("#696E8044"),
+    multipleCellsSelectionBackgroundColor: adaptForDarkMode("#828AA044"),
+  },
+};
--- a/src/helpers/figures/charts/chart_common.ts
+++ b/src/helpers/figures/charts/chart_common.ts
@@ -13,12 +13,13 @@ import {
 import { CommandResult } from "../../../types/commands";
 import { CoreGetters } from "../../../types/core_getters";
 import { LocaleFormat } from "../../../types/format";
-import { Getters } from "../../../types/getters";
 import { Locale } from "../../../types/locale";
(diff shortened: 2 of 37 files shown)


tests/plugins/color_theme_plugin.test.ts
```diff
@@ -18,7 +18,7 @@ describe("ColorThemeUIPlugin via Model getters", () => {
     );
     const theme = model.getters.getSpreadsheetTheme();
     for (const key of THEME_KEYS) {
-      expect(theme[key].toUpperCase()).toBe(expectedLightTheme[key]);
```
hokolomopo, line 21: nit: while you change it, you could change it to use toBeSamecolorAs instead of upperCase

src/types/rendering.ts
```diff
@@ -93,7 +93,10 @@ export interface SheetDOMScrollInfo {
   scrollY: Pixel;
 }
 
+export type ColorTheme = "light" | "dark";
```
hokolomopo: use this type everywhere (command, model config,...)

src/helpers/figures/charts/chart_common.ts
```diff
@@ -367,11 +368,10 @@ export function isTrendLineAxis(axisID: string) {
 
 export function getChartBackgroundColor(
   { background }: { background?: Color },
-  getters: Getters
+  colorTheme: ColorTheme
 ): Color {
-  const defaultColor = getters.getSpreadsheetTheme().backgroundColor;
   if (!background) {
-    return defaultColor;
+    return COLOR_THEMES[colorTheme].backgroundColor;
```
hokolomopo: I don't think we want to use the constant `COLOR_THEMES` anywhere (except in the tests). The source of truth is the theme returned by the plugin, not this constant. We should give the whole theme as argument of the getters.

Might also want to rename `ColorTheme` to `ColorThemeId` or something, I don't expect a string reading the type.

src/helpers/color_themes.ts
```diff
@@ -0,0 +1,41 @@
...
+
+const FROZEN_PANE_HEADER_BORDER_COLOR = "#BCBCBC";
+const FROZEN_PANE_BORDER_COLOR = "#DADFE8";
+const HEADER_BORDER_COLOR = "#C0C0C0";
+const TEXT_HEADER_COLOR = "#666666";
+const BACKGROUND_HEADER_COLOR = "#F8F9FA";
+const BACKGROUND_HEADER_SELECTED_COLOR = "#E8EAED";
+const BACKGROUND_HEADER_ACTIVE_COLOR = "#595959";
+
+export const COLOR_THEMES: Record<ColorTheme, GridRenderingTheme> = {
```
hokolomopo: I think it should stay in the plugin then (see my previous comment)

hokolomopo (commented):
👋

src/helpers/figures/charts/runtime/chart_data_extractor.ts
```diff
@@ -56,7 +57,8 @@ const ZERO = Object.freeze({ value: 0 });
 export function getBarChartData(
   definition: GenericDefinition<BarChartDefinition>,
   { labelValues, dataSetsValues }: ChartData,
-  getters: Getters
+  getters: Getters,
+  colorThemeId: ColorThemeId
```
VincentSchippefilt: you already receive the getters here, why do we need to also give the result of a call to the getter?

src/helpers/figures/charts/runtime/chart_data_extractor.ts
```diff
@@ -191,7 +193,8 @@ function computeValuesAndLabels(
 export function getCalendarChartData(
   definition: GenericDefinition<CalendarChartDefinition>,
   { labelValues, dataSetsValues }: ChartData,
-  getters: Getters
+  getters: Getters,
+  colorThemeId: ColorThemeId
```
VincentSchippefilt: Same as previous comment, and all the other functions that also receive the getters should not change their API

src/helpers/figures/charts/chart_ui_common.ts
```diff
@@ -101,15 +105,34 @@ export async function chartToImageUrl(
...
+ * through the `os-theme-dependant` CSS filter. An exported image doesn't get that filter, so it
+ * has to be applied to the pixels to obtain what is actually displayed on screen.
+ */
+function applyDarkMode(canvas: OffscreenCanvas): OffscreenCanvas {
+  const filtered = createRenderingSurface(canvas.width, canvas.height);
+  const ctx = filtered.getContext("2d");
+  if (!ctx || !("filter" in ctx)) {
+    // canvas filters are not supported in this environment: an unfiltered image is still better
+    // than no image at all
+    return canvas;
```
VincentSchippefilt:
```suggestion
  const isFilterSupportedInCanvas = !ctx || !("filter" in ctx);
  if (isFilterSupportedInCanvas) {
    return canvas;
```

src/types/rendering.ts
```diff
@@ -93,7 +93,10 @@ export interface SheetDOMScrollInfo {
   scrollY: Pixel;
 }
 
+export type ColorThemeId = "light" | "dark";
```
VincentSchippefilt:
```suggestion
export type ColorThemeName = "light" | "dark";
```
I wonder...

tests/figures/chart/chart_conversion.test.ts
```diff
@@ -28,11 +28,11 @@ describe("chart conversion", () => {
     createChart(model, TEST_CHART_DATA.basicChart, chartId, undefined, { figureId });
     const figure = model.getters.getFigure(sheetId, figureId)!;
     const runtime = model.getters.getChartRuntime(chartId);
-    const imageUrl = await chartToImageUrl(runtime, figure, "bar");
+    const imageUrl = await chartToImageUrl(runtime, figure, "bar", "light");
```
VincentSchippefilt: I guess chartToImageUrl could default to the first color theme or default to "light" so we don't have to mess with everything in the tests

VincentSchippefilt (changes requested):
summary, a lot of changes are accidental because the API changed. Adding a default value to the theme parameter of the function that creates the canvas will reduce the noise in the tests, and using the getters that we already have in the getChartXXX will reduce the noise on the other files

src/helpers/figures/charts/runtime/chart_data_extractor.ts
pro-odoo: as discussed on discord, color theme comes from UI getters, and in the next PR it's the EvaluationGetters that will be given

VincentSchippefilt:
robodoo delegate+

VincentSchippefilt:
robodoo r+

pro-odoo:
robodoo r+

PR #9461, opened 2026-08-27
[IMP] data_analysis: detection of headers

matho-odoo:
## Description:

Task: [6516198](https://www.odoo.com/odoo/2328/tasks/6516198)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


src/helpers/figures/charts/chart_suggestion_engine.ts
```diff
@@ -150,23 +150,29 @@ function interestingCellsXc(col: ColumnAnalysis): {
   return { firstCellXC, lastCellXC, prevCellXC };
 }
 
-function getCategoryHeader(getters: Getters, catCol: ColumnAnalysis): string {
-  let catHeader = catCol.header ?? "";
-  if (!catHeader && catCol.rowCount > 0) {
+function getCategoryHeader(getters: Getters, catCol: ColumnAnalysis, hasTitle: boolean): string {
```
anhe-odoo, line 153: Now we always have a title, while before your changes we sometimes had "" as title, leading to no title for the scorecard, which was intended

src/helpers/figures/charts/chart_suggestion_engine.ts
```diff
@@ -150,23 +150,29 @@ function interestingCellsXc(col: ColumnAnalysis): {
...
-  return catHeader;
+  return _t("Category");
 }
 
 /** Pattern A — Single numeric column */
 function buildSingleNumberContext([col]: ColumnAnalysis[], getters: Getters): SingleNumberContext {
   const title = col.header ?? "";
   const { firstCellXC, lastCellXC, prevCellXC } = interestingCellsXc(col);
-  const source = rangeSource([dataset(col.zone, getters)], col.hasHeader);
+  const source = rangeSource([dataset(col.zone, getters)], col.headerInZone);
```
anhe-odoo, line 175: Not sure of the change of this field name :/

src/helpers/figures/charts/chart_suggestion_engine.ts
```diff
@@ -283,8 +289,8 @@ function buildCategoricalVsPercentageContext(
   getters: Getters
 ): CategoricalVsPercentageContext {
   const labelRange = getUnboundRange(getters, catCol.zone);
-  const hasTitle = pctCol.hasHeader;
-  const categoryHeader = hasTitle ? getCategoryHeader(getters, catCol) : _t("Category");
+  const hasTitle = pctCol.headerInZone;
+  const categoryHeader = getCategoryHeader(getters, catCol, hasTitle);
```
anhe-odoo, line 293: The advantage of the ternary was that sometimes we wanted to have "Category" and sometimes we wanted an empty title, "au cas par cas"

src/helpers/data_analysis.ts
```diff
@@ -51,11 +52,27 @@ function analyzeColumn(zone: Zone, getters: Getters): ColumnAnalysis {
...
-  // Header: first cell is text AND rest has at least one non-text, non-empty cell
-  const hasHeader =
-    firstCell.type === CellValueType.text && rest.some((c) => c.type !== CellValueType.text);
+  let header: string | undefined;
+  let dataCells: EvaluatedCell[] = [];
+  let headerInZone = false;
+  const tableHeaderPosition = getColumnTableHeaderPosition(sheetId, zone, getters);
 
-  const dataCells: EvaluatedCell[] = hasHeader ? rest : cells;
+  if (tableHeaderPosition) {
```
anhe-odoo, line 60: This part and the `getColumnTableHeaderPosition` function are probably the only useful parts of this PR :) I think we can keep these and revert all the other changes

src/helpers/table_helpers.ts
```diff
@@ -324,3 +327,35 @@ function getTableElementZones(
...
+    return undefined;
+  }
+  const tables = getters.getTablesOverlappingZones(sheetId, [zone]);
+  if (tables.length !== 1) {
+    return undefined;
+  }
+  const table = tables[0];
+  const firstCell = { sheetId, col: zone.left, row: zone.top };
+  const lastCell = { sheetId, col: zone.left, row: zone.bottom };
+  if (firstCell.row < table.range.zone.top || lastCell.row > table.range.zone.bottom) {
```
anhe-odoo, line 346: Shouldn't we use the header only in the strict case where the first cell in zone is exactly the first of the tale col ?

src/helpers/figures/charts/chart_suggestion_engine.ts
matho-odoo: Only when we call this function which is not all the time

src/helpers/figures/charts/chart_suggestion_engine.ts
matho-odoo: I don't see wich case would be wrong

src/helpers/figures/charts/chart_suggestion_engine.ts
anhe-odoo: Tu as raison

src/helpers/table_helpers.ts
```diff
@@ -324,3 +327,35 @@ function getTableElementZones(
 
   return zones;
 }
+
+export function getColumnTableHeaderPosition(
```
hokolomopo: Nitpick: Mayble a smll docstring explaining what it does (zone need to be inside a single table & all non empty cells should be in table).

I'd also move the helper to dataAnalysis (and not export it), the behavior is very specific to the data analysis feature, not really a generic table behaviour.

src/helpers/data_analysis.ts
```diff
@@ -51,11 +52,27 @@ function analyzeColumn(zone: Zone, getters: Getters): ColumnAnalysis {
...
+    header = getters.getCellText(tableHeaderPosition) || undefined;
+    dataCells = cells.filter((c) => c.position && c.position.row > tableHeaderPosition.row);
+    headerInZone =
+      !!header && tableHeaderPosition.row >= zone.top && tableHeaderPosition.row <= zone.bottom;
```
hokolomopo: I think it's a bit wrong. We use headerInZone to determine whether the chart dataset have a header row or not. But for tables with multiple header rows, the first x rows might need to be ignored. Or even with a single table header row, we might select empty cells above the table.

At the same time taking the sub-header as a title is a bit strange if we select the whole table column as the chart dataset.  

We probably want to change the zone in those case, to only take data starting at the table sub-header and ending at the bottom of the table.

Image

src/helpers/figures/charts/chart_suggestion_engine.ts
```diff
@@ -175,7 +181,7 @@ function buildSinglePercentageContext(
   [col]: ColumnAnalysis[],
   getters: Getters
 ): SinglePercentageContext {
-  const hasTitle = col.hasHeader;
+  const hasTitle = col.headerInZone;
```
hokolomopo: In the ColumnAnalysis I'd change the name to `dataHasHeader(row)` to be clearer. 

Also in all those helper, the variable `hasTitle` is confusing; it having nothing to do with the actual chart title. I'd also rename it to have something to do with the datasets.

tests/helpers/data_analysis_helpers.test.ts
```diff
@@ -65,19 +70,122 @@ describe("analyzeColumns", () => {
...
+    const model = createModelFromGrid({ A1: "100", A2: "200", A3: "300" });
+    createTable(model, "A1:A3", { numberOfHeaders: 0 });
+    const col = analyzeColumn(model, "A1:A3");
+    expect(col.headerInZone).toBe(false);
+    expect(col.header).toBeUndefined();
+    expect(col.rowCount).toBe(3);
+  });
+
+  test("the analyzed zone extends past the table", () => {
+    const model = createModelFromGrid({
```
hokolomopo: nitpick: improve the test names to say what should happen when the table has no header/the zone extend past the bale.

Also maybe a test that we can expand past the table if the rows are all empty

hokolomopo (commented):
👋

hokolomopo:
@matho-odoo small rebase then we can r+ :)

hokolomopo:
@robodoo r+

PR #9530, opened 2026-09-01
[FIX] charts: show value do not work for combo chart

hokolomopo:
## Description

The show value chartjs plugin didn't work for combo charts.
- The runtime had the `combo` type, which isn't a valid chartJs
chart type (but it still mostly worked somehow).
- We would check the type of the chart to decide whether to draw values
for bar or lines instead of the type of the dataset

Task: [6528059](https://www.odoo.com/odoo/2328/tasks/6528059)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


pro-odoo (commented):
robodoo r+

PR #9560, opened 2026-09-04
[PERF] evaluation: clip range to sheet

LucasLefevre:
Some users do stupid things like:

`INDEX($B$2:$ZZZ:$10000)`... repeated thousands of time.

The current data structure for the dependency graph is optimized for row-heavy spreadsheets, which are much more common compared to column-heavy spreadsheets. (`IntervalTree`)

However, if a formula references a range which has many many columns (e.g. A1:ZZZ10), the range is still added to the dependency graph.

With this commit, we clip the ranges added to the dependency graph to the actual sheet size, cutting extra-columns that we don't need to track.

| | Synthetic benchmark [1] | spreadsheet 4756563 | spreadsheet 4185418|
|--------|--------|--------|--------|
| Before | ~10s | cannot open | cannot open |
| After | ~80ms | ~5s | ~14s [2] | 

[1] =INDEX(Sheet1!$A$1:$ZZZ$202, 1, 1) repeated 5200 times
[2] ~14s for initial eval (then ~1 minute, but it's unrelated)

Ticket: [6483083](https://www.odoo.com/odoo/2328/tasks/6483083)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes

### DIFF
--- a/src/model.ts
+++ b/src/model.ts
@@ -175,6 +175,8 @@ export class Model extends EventBus<any> implements CommandDispatcher {
     this.range = new RangeAdapterPlugin(this.coreGetters);
     this.coreGetters.getRangeString = this.range.getRangeString.bind(this.range);
     this.coreGetters.getRangeFromSheetXC = this.range.getRangeFromSheetXC.bind(this.range);
+    this.coreGetters.clipRangesToSheet = this.range.clipRangesToSheet.bind(this.range);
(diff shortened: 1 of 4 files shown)


pro-odoo:
robodoo r+

PR #9573, opened 2026-09-07
 [REF] spreadsheet: add `NotificationPlugin`

hokolomopo:
### [REF] spreadsheet: add `NotificationPlugin`

This commit replace the store `NotificationStore` with an owl plugin
`NotificationPlugin`.

This is the first store replaced by an owl plugin, thus requiring
many changes across the code base.

The action need to be able to access the owl plugins. They are
provided a special env, provided by the `OSComponent` that contain
a scoped method to get a plugin.

The tests changed quite a bit. We need to create an owl app instance,
even for the tests that only test the stores. It's done in `makeTestEnv`.
But it also means that we cannot create a test env before mounting a
component (we need the right scope for the plugins owl). All of the
component tested are now mounted inside a `TestParent` component,
which handles all the necessary setup.

Task: [6542121](https://www.odoo.com/odoo/2328/tasks/6542121)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


pro-odoo:
robodoo rebase-ff

pro-odoo:
robodoo r+

PR #9599, opened 2026-09-08
[FIX] sheetview: prevent blank grid when zoom value is empty

rmbh-odoo:
## Description:

Current behavior before PR:
- Clearing the zoom value sets the zoom level to NaN, causing
the grid to become completely blank.
- Typing a value outside `[50, 200]` and pressing Enter did nothing;
  the popover just closed with no feedback, which was confusing.
- Zoom bounds were `25-300`, inconsistent with the intended `50%-200%` range.

- Entering an empty value keeps the list open and ignores the change,
  instead of silently failing.
- Entering a numeric value clips it to `[50, 200]` and applies it (e.g., `2 -> 50%`, `300 -> 200%`).
- Zoom bounds are now 50-200, and values are rounded instead of floored,
  so `199.6` becomes `200` instead of `199`.

Task: [6515617](https://www.odoo.com/odoo/2328/tasks/6515617)

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


tests/sheet/sheetview_plugin.test.ts
```diff
@@ -1784,3 +1752,18 @@ describe("Partially Scrolled Viewport", () => {
...
+  test("cannot set zoom outside of the [0.5, 2] range", () => {
+    model = new Model();
+    expect(setZoom(model, 0.4)).toBeCancelledBecause(CommandResult.InvalidZoomLevel);
+    expect(setZoom(model, 2.1)).toBeCancelledBecause(CommandResult.InvalidZoomLevel);
+  });
+
+  test("setting an empty/falsy zoom value is cancelled and keeps the current zoom level", () => {
+    model = new Model();
+    setZoom(model, 1.5);
+    expect(setZoom(model, 0)).toBeCancelledBecause(CommandResult.InvalidZoomLevel);
```
dhrp-odoo: This does not reproduce the regression: `0` was already rejected by the previous `cmd.zoom < 0.5` validation, so this test passes without the fix. 

Clearing the zoom input produces `NaN` through `parseFloat("")`. Could you test that flow?

src/components/top_bar/zoom_editor/zoom_editor.ts
```diff
@@ -22,11 +22,11 @@ export class ToolBarZoom extends Component<Props, SpreadsheetChildEnv> {
 
   get currentFontSize(): number {
```
rrahir: nitpick but it could be the occasion to correct this getter name ><
=> getZoomLevel

src/plugins/ui_stateful/sheetview.ts
```diff
@@ -98,7 +98,7 @@ export class SheetViewPlugin extends UIPlugin {
           () => this.viewports.checkPositiveDimension(cmd)
         )(cmd);
       case "SET_ZOOM":
-        if (cmd.zoom > 2 || cmd.zoom < 0.5) {
+        if (!cmd.zoom || cmd.zoom > 2 || cmd.zoom < 0.5) {
```
rrahir: the typing suggests that it cannot be undefined, I guess that the issue is we dispatched `{ zoom: NaN}`? 

maybe !isNaN(cmd.zoom)  would be more accurate ?

tests/__snapshots__/top_bar_component.test.ts.snap
```diff
@@ -197,8 +197,8 @@ exports[`TopBar component simple rendering 1`] = `
               >
                 <input
                   class="o-font-size o-number-input bg-transparent border-0"
-                  max="300"
-                  min="25"
+                  max="200"
```
rrahir, line 200: 🤦   I think it's worth setting variables for the lower/upper boundary so the component and plugins a synced

tests/components/number_editor.test.ts
```diff
@@ -0,0 +1,108 @@
...
+    getInput().focus();
+    await nextTick();
+    await click(document.body, '[data-size="200"]');
+    expect(onValueChange).toHaveBeenCalledWith(200);
+    expect(".o-text-options").toHaveCount(0);
+  });
+
+  test.each([
+    ["150", 150], // within range, unchanged
+    ["149.6", 150], // rounded to the nearest integer
```
rrahir: 👍

rrahir (changes requested):

src/components/number_editor/number_editor.ts
```diff
@@ -130,21 +143,29 @@ export class NumberEditor extends Component<Props, SpreadsheetChildEnv> {
...
+        // In the case of a ESCAPE key, we get the previous value back
+        target.value = `${this.props.currentValue}`;
+        this.closeList();
+        break;
+      default:
+        return;
     }
+    this.skipNextInputChange = true;
+    this.DOMFocusableElementStore.focus();
+    this.skipNextInputChange = false;
```
rmbh-odoo, line 169: Adding this flag to prevent `setValue` from being called a second time by `setValueFromInput`.

rrahir (commented):
Thanks a lot :)

robodoo rebase-ff r+

PR #9718, opened 2026-09-21
[PERF] pivot: index the spreadsheet pivot data entries

LucasLefevre:
Create a spreadsheet pivot from a range with many rows
and many distinct values you're gouping on such that
 the resulting pivot table has many cells.

Let's say you group values in a range by partner, with
lots of distinct partners.

|  | Before | After | Speedup |
|--------|--------|--------|--------|
| 10k rows, 2.4k partners | 2.2s | 1.2s | x1.8 |
| 50k rows, 7.7k partners | 64s | 1.4s | x45 |
| 100k rows, 10k partners | 334s | 1.8s | x185 | 

Every call to `getPivotCellValueAndFormat` and
 `getPivotHeaderValueAndFormat` filtered the whole list of data entries to
    find the ones matching the given domain, once per domain node. Those are
    called once per pivot cell, which means the cost of evaluating a pivot
    was `O(number of pivot cells x number of data entries)`.

The data entries are now indexed in a tree, built lazily.

Each entry is inserted once per couple (row prefix, column prefix) it
    belongs to, which is exactly the number of pivot cells it contributes to.

Note that the domain nodes are not necessarily ordered with the rows
    first (`getDomainOfParentRow` and `getDomainOfParentCol` build domains
    with the columns first). We first classify domain dimensions into
    rows/columns before descending the tree.

Idea to go further
------------------
Store the aggregates (min, max, sum, count, etc.) at each node
instead of all entries. It will allow to get the aggregate in O(1)
instead scanning all matching entries.

## review checklist

- [ ] exportable in excel
- [ ] unit tested
- [ ] clean commented code
- [ ] track breaking changes


pro-odoo:
robodoo r+
Thanks :)

